PR #346, opened 2026-03-07
feat:Add inline user message editing

### LINKED ISSUE #331: Feat: Editing messages

6crucified:
## Summary
- add an inline edit action for user chat bubbles
- send edited content directly from the bubble editor instead of staging it in the main composer
- wait for checkpoint revert state to land before resending so the edited message replaces the original row

Fixes #331

> [!NOTE]
> ### Add inline editing of user messages in the chat timeline
> - Adds an `EditableUserMessageBubble` component in [ChatView.tsx](https://github.com/pingdotgg/t3code/pull/346/files#diff-4b49e092ccd43be0f0de24abe85ba522e09f04288a5d84253b0263e1a389400e) that lets users edit previously sent messages (text and images) directly in the timeline.
> - On edit submission, the app silently reverts to the message's checkpoint (skipping the confirmation dialog), waits up to 3 seconds for the message to disappear from the thread, then resends the updated content.
> - Extracts a `submitUserTurn` helper to share send logic between normal sends and edit resends; edit resends preserve the composer draft while normal sends clear it.
> - Image editing supports adding (via button or paste), removing, and previewing attachments; keyboard shortcuts Ctrl/Cmd+Enter to submit and Escape to cancel are supported.
> - Risk: the 3-second timeout (`EDIT_REVERT_SYNC_TIMEOUT_MS`) means a slow store update could cause the edited message to be sent before the reverted message is fully removed, producing a race in the timeline.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d458ed4.</sup>
> 
>


jparta:
This is an essential feature. I don't know how Theo and Julius don't find this annoying enough to ship a fix.

@juliusmarminge do you have time to look at this?

juliusmarminge:
Essential feature? Just hit the revert button and then type in your revised message. 

Your attitude doesn't help here, I am basically soloing maintenance + engineering and have 400 open issues, 300 open PRs and a product vision I need to fulfill so excuse me if stuff gets lost in the noise

jparta:
@juliusmarminge 

Sorry.

I can see the revert button now, but didn't before.

I wonder if this PR is relevant anymore if revert does the job.

Igloczek:
@juliusmarminge can you please re-open this? the revert feature doesn't really solve the problem

for example - when I stop the execution seconds after sending the prompt to correct it, there is no way to do anything with it, revert doesn't allow me to send new prompt in the same place, it just removes what AI generated in the few seconds, which isn't really helpful tbh

it would work fine-ish if there would be a way to revert to the end of the AI response, so it will remove my prompt, but it doesn't work that way

also reverting doesn't work at all with grok cli, it just throw errors

the edit is what plenty of other AI apps using, it's just good UX, simply because people are used to doing it that way

PR #595, opened 2026-03-09
fix: checkpoint diffs never resolve (shared PubSub subscription)

### LINKED ISSUE #585: Checkpoint diffs never resolve: ProviderRuntimeIngestion placeholder preempts CheckpointReactor git capture
## Bug

Turn diffs ("Loading checkpoint diff…" / "Filesystem checkpoint is unavailable for turn N") never resolve after a Codex turn completes, even when the project is a valid git repository.

## Root cause

There is a race between two independent event handlers that both react to turn completion:

1. **`ProviderRuntimeIngestion`** handles `turn.diff.updated` events and immediately dispatches `thread.turn.diff.complete` with a **placeholder** checkpoint entry:
   - `checkpointRef: "provider-diff:<eventId>"` (not a real git ref)
   - `status: "missing"`
   - `files: []`

   (`apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts`, lines 1064-1083)

2. **`CheckpointReactor`** handles `turn.completed` events and is supposed to capture a real git-ref-based checkpoint via `CheckpointStore.captureCheckpoint`, then dispatch `thread.turn.diff.complete` with the actual diff data.

   (`apps/server/src/orchestration/Layers/CheckpointReactor.ts`, lines 150-298)

**The problem:** `ProviderRuntimeIngestion` wins the race and inserts the placeholder into `thread.checkpoints` first. When `CheckpointReactor` later processes `turn.completed`, it checks:

```typescript
if (thread.checkpoints.some((checkpoint) => checkpoint.turnId === turnId)) {
  return; // ← early exit, real capture never happens
}
```

(line 169)

Since the placeholder already has the same `turnId`, the real git checkpoint is **never captured**. The DiffPanel then tries to resolve `provider-diff:UUID` via `git rev-parse`, which fails because it's not an actual git ref.

## Observed behavior

- Server logs show `thread.turn-diff-completed` events with `checkpointRef: "provider-diff:..."` and `status: "missing"`
- DiffPanel shows "Loading checkpoint diff..." indefinitely, then "Filesystem checkpoint is unavailable for turn N"
- No `refs/t3/checkpoints/` git refs are created for the affected turns

## Reproduction

1. Add any git-initialized project
2. Send a message that triggers file changes (e.g., "add a comment to README")
3. Wait for turn to complete
4. Open the diff panel → stuck on "Loading checkpoint diff..."

## Suggested fix

The dedup check in `CheckpointReactor` (line 169) should either:
- Only skip if an existing checkpoint has a **non-placeholder** status (i.e., skip `"missing"` entries), or
- Replace/upgrade placeholder checkpoints with the real git-based checkpoint when one becomes available

Additionally, there is a secondary CWD inconsistency worth noting: `CheckpointReactor` turn-completion capture prefers `sessionRuntime.cwd` while `CheckpointDiffQuery` only uses `resolveThreadWorkspaceCwd`. If these ever diverge, refs would be written to one repo but looked up in another.

## Environment

- T3 Code dev (main branch)
- Codex provider with GPT-5.4
- Linux (Arch)

eggfriedrice24:
## Summary

Fixes #585

- `Stream.fromPubSub()` stored as a property meant all consumers shared one subscription queue. Only the first reader got events; CheckpointReactor received nothing and never captured git checkpoints.
- Changed `streamDomainEvents` and `streamEvents` to getters so each consumer gets its own subscription.
- Added a domain event handler that replaces placeholder checkpoints (status `"missing"`) with real git captures.
- Extracted shared CWD resolution and capture+diff+dispatch logic to reduce duplication in CheckpointReactor.

## Files changed

- `OrchestrationEngine.ts`: `streamDomainEvents` property → getter
- `ProviderService.ts`: `streamEvents` property → getter
- `CheckpointReactor.ts`: placeholder handler, dedup fix, shared helpers extracted

## Test plan

- [x] `bun lint` passes
- [x] `bun typecheck` passes
- [x] Manual: create thread, prompt file change, verify diff panel shows actual file diffs

> [!NOTE]
> ### Fix checkpoint diffs never resolving due to shared PubSub subscriptions
> - Changes `streamDomainEvents` in [OrchestrationEngine.ts](https://github.com/pingdotgg/t3code/pull/595/files#diff-996de6bdce67b7926493255cf29e93653d211eca31023cf4c476b42f7dc59570) and `streamEvents` in [ProviderService.ts](https://github.com/pingdotgg/t3code/pull/595/files#diff-b9b9384b78856ff56cff5dd0ccb106054fb5d2d1b7500fcd30d8938ebfec5775) from pre-created Stream fields to getters, so each consumer gets an independent subscription.
> - Adds a `captureCheckpointFromPlaceholder` handler in [CheckpointReactor.ts](https://github.com/pingdotgg/t3code/pull/595/files#diff-79dc41615f79575ae1404bd102df915797fb24ba34989557f80e229d305521fb) that replaces placeholder (`status: 'missing'`) checkpoints with real git-ref checkpoints when `thread.turn-diff-completed` events are received, even if no `turn.completed` runtime event arrived.
> - Fixes `captureCheckpointFromTurnCompletion` to proceed when only a placeholder checkpoint exists for a turn, reusing its `checkpointTurnCount` rather than incrementing past it.
> - Guards the `thread.turn.diff.complete` projector in [projector.ts](https://github.com/pingdotgg/t3code/pull/595/files#diff-3da4865ecd7a117ebdca42c25e9526bc69b87696dc8209029ecb2809f8e6a937) so placeholder checkpoints cannot overwrite an already-captured real checkpoint.
> - Fixes placeholder dispatch in [ProviderRuntimeIngestion.ts](https://github.com/pingdotgg/t3code/pull/595/files#diff-22a6c9818b956463b848a73a5b3e787b44a144bd003b1cc25fa2b1b3110cdea7) to skip if a real checkpoint already exists and to compute `checkpointTurnCount` from the existing max rather than array length.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 420c3fa.</sup>
> 
>

before (took like ~2-3 minutes to load checkpoint diff and return error):
image

after fix:
image

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts
@@ -1064,22 +1064,34 @@ const make = Effect.gen(function* () {
       if (event.type === "turn.diff.updated") {
         const turnId = toTurnId(event.turnId);
         if (turnId && (yield* isGitRepoForThread(thread.id))) {
-          const assistantMessageId = MessageId.makeUnsafe(
-            `assistant:${event.itemId ?? event.turnId ?? event.eventId}`,
-          );
-          yield* orchestrationEngine.dispatch({
-            type: "thread.turn.diff.complete",
-            commandId: providerCommandId(event, "thread-turn-diff-complete"),
-            threadId: thread.id,
-            turnId,
-            completedAt: now,
-            checkpointRef: CheckpointRef.makeUnsafe(`provider-diff:${event.eventId}`),
-            status: "missing",
-            files: [],
-            assistantMessageId,
-            checkpointTurnCount: thread.checkpoints.length + 1,
-            createdAt: now,
-          });
+          // Skip if a checkpoint already exists for this turn. A real
+          // (non-placeholder) capture from CheckpointReactor should not
+          // be clobbered, and dispatching a duplicate placeholder for the
+          // same turnId would produce an unstable checkpointTurnCount.
+          if (thread.checkpoints.some((c) => c.turnId === turnId)) {
+            // Already tracked; no-op.
+          } else {
+            const assistantMessageId = MessageId.makeUnsafe(
+              `assistant:${event.itemId ?? event.turnId ?? event.eventId}`,
+            );
+            const maxTurnCount = thread.checkpoints.reduce(
+              (max, c) => Math.max(max, c.checkpointTurnCount),
+              0,
+            );
+            yield* orchestrationEngine.dispatch({
+              type: "thread.turn.diff.complete",
+              commandId: providerCommandId(event, "thread-turn-diff-complete"),
+              threadId: thread.id,
+              turnId,
+              completedAt: now,
+              checkpointRef: CheckpointRef.makeUnsafe(`provider-diff:${event.eventId}`),
+              status: "missing",
+              files: [],
+              assistantMessageId,
+              checkpointTurnCount: maxTurnCount + 1,
+              createdAt: now,
+            });
+          }
         }
       }
 
--- a/apps/server/src/orchestration/projector.ts
+++ b/apps/server/src/orchestration/projector.ts
@@ -500,6 +500,16 @@ export function projectEvent(
           "checkpoint",
         );
 
+        // Do not let a placeholder (status "missing") overwrite a checkpoint
+        // that has already been captured with a real git ref (status "ready").
+        // ProviderRuntimeIngestion may fire multiple turn.diff.updated events
+        // per turn; without this guard later placeholders would clobber the
+        // real capture dispatched by CheckpointReactor.
+        const existing = thread.checkpoints.find((entry) => entry.turnId === checkpoint.turnId);
+        if (existing && existing.status !== "missing" && checkpoint.status === "missing") {
+          return nextBase;
+        }
+
         const checkpoints = [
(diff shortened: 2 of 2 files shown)


juliusmarminge:
turn diffs still broken (i can see the full thread diff, but not the individual turn diffs and i don't see the tree of changed files in the thread activities), you up to debug or should I?

CleanShot 2026-03-10 at 16 17 26@2x

CleanShot 2026-03-10 at 16 17 43@2x

eggfriedrice24:
Found and fixed the individual turn diff issue. Two problems:

1. **ProviderRuntimeIngestion** fires multiple `turn.diff.updated` per turn. Each dispatched a new placeholder that the projector replaced by turnId, overwriting CheckpointReactor's real git capture. Fixed: skip placeholder dispatch if a checkpoint already exists for that turnId.

2. **checkpointTurnCount** was computed as `thread.checkpoints.length + 1` which produces unstable values after dedup. Fixed: use `max(checkpointTurnCount) + 1` instead.

Also added a guard in the projector so a "missing" placeholder can never overwrite a "ready" checkpoint.

Tested locally, both "All turns" and individual turn diffs now work correctly on fresh threads.

@juliusmarminge 

image

eggfriedrice24:
create a new thread the old thread has stale placeholder checkpoints from before, should work 100%

PR #609, opened 2026-03-09
fix(server): soften Codex CLI version enforcement

### LINKED ISSUE #604: Version check
Image

T3 Code is doing a hard version check, but the in-house version of codex in company shows an internal version number that differs from the open source codex one.

Is it possible to not rely on this behavior? T3 code 0.0.4 works for our codex (with a bit of warning), but T3 code 0.0.5 shows this version error.

binbandit:
## Summary

- Remove the hard `codex --version` startup gate added in 0.0.5 that blocks sessions for any parsable Codex version below `0.37.0`.
- Keep the minimum-version check as a provider health warning so users still get compatibility guidance without losing access to working internal Codex forks.

Closes #604

## Why

Issue #604 reports a real regression from `0.0.4` -> `0.0.5`: internal/company Codex builds can report a different version number than the open-source CLI, and `0.0.5` started hard-failing session startup before even attempting `codex app-server`.

I validated the regression in the code:

- PR #404 introduced a synchronous `assertSupportedCodexCliVersion()` preflight in `apps/server/src/codexAppServerManager.ts`
- that check runs `codex --version` and throws before startup when the parsed version is `< 0.37.0`
- provider health also marked those versions as `status: "error"` / `available: false`

That is too strict for enterprise forks where version numbers diverge but runtime compatibility can still be fine.

## Changes

- `apps/server/src/codexAppServerManager.ts`
  - remove the synchronous startup-time version gate
  - let `codex app-server` be the source of truth for whether the provider can actually start
- `apps/server/src/provider/Layers/ProviderHealth.ts`
  - keep the minimum-version check, but downgrade it to a `warning` when auth is otherwise healthy
  - preserve the existing warning message so users still see compatibility guidance
- tests updated accordingly

## Validation

Passed:
- `bun lint`
- `bun typecheck`
- `bun run test src/codexAppServerManager.test.ts` in `apps/server`
- `bun run test src/provider/Layers/ProviderHealth.test.ts` in `apps/server`

Repo baseline issue still present:
- full `bun run test` still fails on pre-existing unrelated web storage test regressions in `apps/web/src/terminalStateStore.test.ts` and `apps/web/src/composerDraftStore.test.ts`

> [!NOTE]
> ### Downgrade Codex CLI version enforcement from error to warning in provider health checks
> - Removes the pre-launch version check from `CodexAppServerManager.startSession`, so an unsupported CLI version no longer causes `startFailed` before the process spawns.
> - Changes `checkCodexProviderStatus` in [ProviderHealth.ts](https://github.com/pingdotgg/t3code/pull/609/files#diff-9cda1355058a2e518b1c0afc4b22752fbcfc06e99f56ca1527c622d6d2ad3448) to return `status: 'warning'` (available: true) instead of `status: 'error'` (available: false) when the CLI is below the minimum supported version.
> - Auth command timeouts now also return `'warning'` instead of `'error'`.
> - Behavioral Change: users on unsupported Codex CLI versions will see a warning rather than a hard error, and the server session will proceed.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 82c37ad.</sup>
> 
>
>


Da1sypetals:
😢 Anyone on this

PR #740, opened 2026-03-09
fix(web): only use VITE_WS_URL in dev mode

chirino:
In production, the page is served by the WS server on the same host:port, so the WebSocket URL should be derived from window.location. VITE_WS_URL is baked in at build time and would override the window.location fallback, pointing the browser to the wrong host (e.g. localhost instead of the actual server host).

> [!NOTE]
> ### Fix `WsTransport` to ignore `VITE_WS_URL` in production builds
> In [wsTransport.ts](https://github.com/pingdotgg/t3code/pull/740/files#diff-2129bf8b884f9eac0205e2de5672354ffe759743b6538d8dee077fb1944d1506), `VITE_WS_URL` is now only read when `import.meta.env.DEV` is true. In production, the WebSocket URL falls back to the provided `url`, the desktop bridge URL, or the default `ws(s)://host:port`. This prevents dev-only environment variables from leaking into production builds.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1ca0261.</sup>
> 
>

### DIFF
--- a/apps/web/src/wsTransport.ts
+++ b/apps/web/src/wsTransport.ts
@@ -37,7 +37,7 @@ export class WsTransport {
     const bridgeUrl = window.desktopBridge?.getWsUrl();
     // In dev mode, VITE_WS_URL points to the server's WebSocket endpoint.
     // In production, the page is served by the WS server on the same host:port.
-    const envUrl = import.meta.env.VITE_WS_URL as string | undefined;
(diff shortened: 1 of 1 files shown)


juliusmarminge:
Closing because its only target file was removed by the current connection architecture.

Issue #769, opened 2026-03-10
UI Overflow when Git Action Shows 'Commit, Push & Create PR'


dbalders:
When on a new branch with uncommitted changes, the Git action in the top menubar adjusts to 'Commit, Push & Create PR'. This becomes too long and overflows underneath the + button making you unable to click the dropdown arrow unless you resize. 

<img width="413" height="128" alt="Image" src="https://github.com/user-attachments/assets/2244e5cb-1aff-4bae-b0c8-dbeebf593bbb" />

PR #771, opened 2026-03-10
Fix UI Overflow when Github Action is 'Commit, Create & Push PR'

dbalders:
## What Changed

[edited post comments]

Updated the label for 'Commit, create & push PR' to be 'Commit, create & PR' to allow the overlap to no longer happen, but still fit within the default window size without force collapsing all the icons by default. Brings the text closer to the size of the other labels and I believe is still clear instructions.

## Why

Fixes: #769, the problem with the 'Commit, push & create PR' overflowing behind the + in the title bar.

## UI Changes

Before:
Screenshot 2026-03-09 at 10 17 09 AM

After:
Screenshot 2026-03-09 at 10 09 27 PM

## Checklist

> [!NOTE]
> ### Shorten 'Commit, push & create PR' label to 'Commit, push & PR' to fix UI overflow
> The quick action label in [GitActionsControl.logic.ts](https://github.com/pingdotgg/t3code/pull/771/files#diff-5e0942b75e4be5ac9ffadabc7d500f512b883da1c98aaca0baa2b959a54985c0) is shortened to prevent text overflow in the UI when the action is to commit, push, and create a PR.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7750bc8.</sup>
> 
>
>

### DIFF
--- a/apps/web/src/components/GitActionsControl.logic.test.ts
+++ b/apps/web/src/components/GitActionsControl.logic.test.ts
@@ -344,7 +344,7 @@ describe("when: working tree has local changes", () => {
     assert.deepInclude(quick, {
       kind: "run_action",
       action: "commit_push_pr",
-      label: "Commit, push & create PR",
+      label: "Commit, push & PR",
     });
   });
 
@@ -440,7 +440,7 @@ describe("when: working tree has local changes and branch is behind upstream", (
     assert.deepInclude(quick, {
       kind: "run_action",
       action: "commit_push_pr",
-      label: "Commit, push & create PR",
+      label: "Commit, push & PR",
     });
   });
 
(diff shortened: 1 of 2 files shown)


dbalders:
@Noojuno Ah, good find! Thanks. Been messing around with it for a while now, and kinda seems like the 2 solutions are to have it collapse earlier, or change the text to be something like 'Commit, push & PR' (removing the 'create'). 

Collapsing earlier would me changing the buttons from `@sm/header-actions` to `@[27rem]/header-actions`.

Both decisions kinda feel like maintainer decisions as one is a language change, and the other means they will be collapsed on the default opening window size in order to fit.

I'm going to go with what I would probably do if it was my app, and remove the 'create' as this keeps the logic the same, ui the same on default opening, keeps the text not too long, and people should kind of inherently understand what this means without the 'create'.

Can switch to something else if needed or feel free to close if that isn't a decision that is wanted.

juliusmarminge (approved):

Issue #677, opened 2026-03-09
Hyperlinks in response message do not work


klinki:
Hyperlinks in response messages are highlighted as blue but are not clickable. At least not on Windows.

Context menu on right click doesn't allow to copy link URL.

<img width="1461" height="1039" alt="Image" src="https://github.com/user-attachments/assets/1faadf51-346b-40ea-b7e4-3b1d2680d04a" />

PR #599, opened 2026-03-09
fix: fix external link opening in desktop

Noojuno:
## What Changed

Added electron link handling for safe links so they open in an external browser.

I have validated that file links still open in Cursor, and that external links work in both the browser and desktop environment.

## Why

Clicking on external links in the agent response was doing nothing rather than opening in an external browser.

Fixes #434

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] I included before/after screenshots for any UI changes
- [x] I included a video for animation/interaction changes

> [!NOTE]
> ### Validate and open http/https external links from desktop by routing `main.createWindow` window-open requests and `main.registerIpcHandlers` `OPEN_EXTERNAL` through `main.getSafeExternalUrl` in [main.ts](https://github.com/pingdotgg/t3code/pull/599/files#diff-31a471f6ef958ceff6e87ee910f4bc4f7bbc4986f797f159353b328a5916f7cb)
> Add `main.getSafeExternalUrl` to normalize and guard URLs; update `OPEN_EXTERNAL` IPC handler and `webContents.setWindowOpenHandler` to use it and open validated links via `shell.openExternal` in [main.ts](https://github.com/pingdotgg/t3code/pull/599/files#diff-31a471f6ef958ceff6e87ee910f4bc4f7bbc4986f797f159353b328a5916f7cb).
>
> #### 📍Where to Start
> Start with `getSafeExternalUrl` and its use in the `OPEN_EXTERNAL` IPC handler and `webContents.setWindowOpenHandler` in [main.ts](https://github.com/pingdotgg/t3code/pull/599/files#diff-31a471f6ef958ceff6e87ee910f4bc4f7bbc4986f797f159353b328a5916f7cb).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 015ee56.</sup>
> 
>

PR #772, opened 2026-03-10
Feature: add link actions to context menu

dbalders:
## What Changed

Added 2 new link-specific elements in the context menu when right-clicking a link in chat:

- Open Link
- Copy Link Address

Image:
Screenshot 2026-03-09 at 9 08 20 AM

This was part of the request from #677. I confirmed that the main issue for #677 of links not working on Windows was resolved previously by #599, but added the quality of life improvement of the extra attributes to the context menu.

Added an if statement to the context menu event to check if what is being right clicked on is a link or not, and if so, add the 2 elements and a separator. This required adding 'clipboard' to the electron import as well. 

## Why

This makes link behavior more closely resemble standard UX.

## UI Changes

This is a UI change specifically. 

Before:
Screenshot 2026-03-09 at 9 10 37 AM

After:
Screenshot 2026-03-09 at 9 08 20 AM

## Checklist

> [!NOTE]
> ### Add 'Open Link' and 'Copy Link Address' entries to the desktop context menu
> When right-clicking a link in the desktop app, two new items appear in the context menu: 'Open Link' (opens via `shell.openExternal`) and 'Copy Link Address' (writes to clipboard). Both items are only shown when `getSafeExternalUrl` returns a valid URL, followed by a separator before the standard edit items.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 956c18d.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/main.ts
+++ b/apps/desktop/src/main.ts
@@ -4,7 +4,7 @@ import * as FS from "node:fs";
 import * as OS from "node:os";
 import * as Path from "node:path";
 
-import { app, BrowserWindow, dialog, ipcMain, Menu, nativeImage, protocol, shell } from "electron";
+import { app, BrowserWindow, dialog, ipcMain, Menu, nativeImage, protocol, shell, clipboard } from "electron";
 import type { MenuItemConstructorOptions } from "electron";
 import * as Effect from "effect/Effect";
 import type { DesktopUpdateActionResult, DesktopUpdateState } from "@t3tools/contracts";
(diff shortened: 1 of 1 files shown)


apps/desktop/src/main.ts
```diff
@@ -1186,6 +1171,15 @@ function createWindow(): BrowserWindow {
       menuTemplate.push({ type: "separator" });
     }
 
+    const externalUrl = getSafeExternalUrl(params.linkURL);
+    if (externalUrl) {
+      menuTemplate.push(
+        { label: "Open Link", click: () => void shell.openExternal(externalUrl) },
```
juliusmarminge: don't think this should be `shell.openExternal` as that opens it on the host machine. if you run the server on a separate server from the frontend, i'd expect clikcing the link to open it on the current machien, not the host

juliusmarminge (changes requested):

PR #776, opened 2026-03-10
fix(git): preserve full target path for brace-style renames in numstat parsing

dipeshbabu:
## Summary

Fixes Git numstat path parsing for brace-style renames such as:

src/{old => new}/file.ts

Previously the parser sliced everything after ` => `, which produced an invalid path like:

new}/file.ts

This patch preserves the full target path by expanding brace-style renames first, while keeping the existing behavior for simple rename outputs like:

old.ts => new.ts

## What changed

- Added `normalizeNumstatPath()` to normalize git numstat paths
- Handle brace-style renames before the simple `old => new` fallback
- Added an integration test covering staged brace-style renames in `statusDetails()`

## Why this matters

`statusDetails()` uses numstat output to build the working tree file list and per-file insertions/deletions. For brace-style renames, the old parsing could attribute stats to a bogus truncated path.

> [!NOTE]
> ### Fix numstat parsing to preserve full target path for brace-style git renames
> Git numstat output for renamed files can use brace-style notation (e.g. `src/{old => new}/file.ts`), which was not handled correctly, causing the reported path to be `new}/file.ts` instead of `src/new/file.ts`.
>
> - Adds `normalizeNumstatPath` in [GitCore.ts](https://github.com/pingdotgg/t3code/pull/776/files#diff-9e2e5027bebfaa9721be501cd8057c4a36400119ad0ad4797a8d6aca7f3c7865) to expand brace-style `{old => new}` segments into the full post-rename path, with a fallback for plain ` => ` arrow format.
> - Updates `parseNumstatEntries` to call `normalizeNumstatPath` instead of the previous inline ` => `-only normalization.
> - Adds a regression test in [GitCore.test.ts](https://github.com/pingdotgg/t3code/pull/776/files#diff-49e9b5a653566cf07288b6f6d553d942f05d7f26eba74c4e0f94a9e2a144f986) that performs a `git mv` across subdirectories and asserts the correct path is returned by `statusDetails`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 908f3e0.</sup>
> 
>

### DIFF
--- a/apps/server/src/git/Layers/GitCore.test.ts
+++ b/apps/server/src/git/Layers/GitCore.test.ts
@@ -1,4 +1,4 @@
-import { existsSync } from "node:fs";
+import { existsSync, mkdirSync } from "node:fs";
 import path from "node:path";
 
 import * as NodeServices from "@effect/platform-node/NodeServices";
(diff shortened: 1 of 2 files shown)


dipeshbabu:
@juliusmarminge could you review this?

PR #836, opened 2026-03-10
Move user message metadata outside the chat bubble

### LINKED ISSUE #832: Move the time and action buttons outside of chat background
Right now, the time and the action buttons that show up on hover are inside the gray background. I feel like this looks a little bulky when there is a large gray space below every message. I believe best course of action would be to just move it outside the gray area and let it live on the white background.

Now:
Image

Proposed solution:
Image

dbalders:
## What Changed

Added a wrapper div around user chat messages so only the message content keeps the gray background, while the timestamp and action buttons render outside the bubble on the white background.

## Why

Fixes #832 - Just feel this has a better look, especially for longer messages so there isn't a huge gray area below every message.

## UI Changes

Before:
Screenshot 2026-03-10 at 12 30 05 PM

After:
Screenshot 2026-03-10 at 12 30 13 PM

## Checklist

> [!NOTE]
> ### Move user message metadata outside the chat bubble in `MessagesTimeline`
> Restructures the user message layout in [ChatView.tsx](https://github.com/pingdotgg/t3code/pull/836/files#diff-4b49e092ccd43be0f0de24abe85ba522e09f04288a5d84253b0263e1a389400e) so the action toolbar sits below the bubble as a sibling element rather than inside it. The bubble is now wrapped in a right-aligned flex column container, sized to its content with `w-fit`, and the toolbar gets slight right padding (`pr-1`) to align with the bubble edge.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized afa678c.</sup>
> 
>
>

### DIFF
--- a/apps/web/src/components/chat/MessagesTimeline.tsx
+++ b/apps/web/src/components/chat/MessagesTimeline.tsx
@@ -340,50 +340,52 @@ export const MessagesTimeline = memo(function MessagesTimeline({
           const canRevertAgentWork = revertTurnCountByUserMessageId.has(row.message.id);
           return (
             <div className="flex justify-end">
-              <div className="group relative max-w-[80%] rounded-2xl rounded-br-sm border border-border bg-secondary px-4 py-3">
-                {userImages.length > 0 && (
-                  <div className="mb-2 grid max-w-[420px] grid-cols-2 gap-2">
-                    {userImages.map(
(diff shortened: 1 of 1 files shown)


maria-rcks (approved):

dbalders:
Looks like this got fixed with something else

PR #934, opened 2026-03-11
fix(web): defer diff worker startup until diff opens

binbandit:
## Summary
- defer diff worker pool mounting until the diff UI is opened
- keep the diff panel mounted after first open within a thread so close and reopen behavior stays stable
- remove the eager chat-route worker pool mount that was triggering browser worker initialization during non-diff flows

## Validation
- bun run test:browser -- --maxWorkers=1
- bun fmt
- bun lint
- bun typecheck

> [!NOTE]
> ### Defer diff worker pool startup until the diff panel is first opened
> - The `DiffWorkerPoolProvider` is moved out of the chat route layout and into a new `LazyDiffPanel` component, so the worker pool is only instantiated when the diff panel is first opened.
> - `ChatThreadRouteView` tracks a `hasOpenedDiff` flag so the diff panel stays mounted (warm) after first open, avoiding re-initialization on close or thread switch.
> - Introduces `DiffPanelShell`, `DiffPanelHeaderSkeleton`, and `DiffPanelLoadingState` components to show a structured skeleton UI while the diff panel and worker pool load, replacing plain text fallbacks.
> - `DiffLoadingFallback` is updated to use these new skeleton components for both inline and sheet modes.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a532709.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/DiffPanel.tsx
+++ b/apps/web/src/components/DiffPanel.tsx
@@ -19,7 +19,6 @@ import { cn } from "~/lib/utils";
 import { readNativeApi } from "../nativeApi";
 import { resolvePathLinkTarget } from "../terminal-links";
 import { parseDiffRouteSearch, stripDiffSearchParams } from "../diffRouteSearch";
-import { isElectron } from "../env";
 import { useTheme } from "../hooks/useTheme";
 import { buildPatchCacheKey } from "../lib/diffRendering";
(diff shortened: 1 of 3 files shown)


binbandit:
@juliusmarminge  #2 -- this fixes a flaky browser test

PR #972, opened 2026-03-12
fix(desktop): backfill SSH_AUTH_SOCK from login shell on macOS

### LINKED ISSUE #971: [BUG] macOS desktop app does not inherit SSH_AUTH_SOCK from login shell when launched outside Terminal
## Problem

On macOS, the desktop app repairs `PATH` from the user’s login shell during startup, but it does not hydrate other shell-provided environment variables.

That means if T3 Code is launched from Finder or the Dock, it can miss `SSH_AUTH_SOCK` even though the same machine and repo work correctly from Terminal. In practice, Git operations that rely on an SSH agent can fail inside the desktop app while succeeding in a login shell.

Relevant code paths:
- `apps/desktop/src/fixPath.ts`
- `packages/shared/src/shell.ts`

## Reproduction

1. On macOS, configure a repo to use a Git remote over SSH.
2. Use an SSH agent setup that exposes auth through `SSH_AUTH_SOCK` in the login shell.
3. Launch T3 Code from Finder or the Dock, not from Terminal.
4. Trigger a Git operation from the desktop app that needs SSH auth.

## Actual

The desktop app process can be missing `SSH_AUTH_SOCK`, so Git/SSH auth fails even though the same repo works from Terminal.

## Expected

The macOS desktop app should sync the relevant shell environment needed for Git/SSH behavior, not just `PATH`, so Git operations behave consistently with a normal login shell.

## Proposed Scope

Small, focused bug fix:
- Extend the existing macOS shell-env bootstrap to read `SSH_AUTH_SOCK` alongside `PATH`
- Preserve the current fallback behavior if shell environment lookup fails
- Keep the change desktop-scoped and avoid unrelated shell/bootstrap refactors

## Why This Matters

This is a reliability issue in normal desktop usage. Launching from Finder is a standard macOS path, and Git auth should not depend on whether the app was started from Terminal.

stickerdaniel:
Closes #971.

## What Changed

- replace the desktop-only `fixPath()` startup hook with `syncShellEnvironment()`
- extend the shared login-shell probe so desktop startup can read multiple environment variables instead of only `PATH`
- continue refreshing `PATH` on macOS startup, but only backfill `SSH_AUTH_SOCK` when it is missing
- add desktop tests covering macOS hydration, non-macOS no-op behavior, and the regression case where an inherited `SSH_AUTH_SOCK` must be preserved

## Why

The macOS desktop app currently repairs `PATH` from the login shell, but Finder and Dock launches can still miss `SSH_AUTH_SOCK`. That breaks Git-over-SSH flows in the desktop app even when the same repo works correctly from Terminal.

This keeps the fix small and reliability-focused:
- Finder/Dock launches get the missing SSH agent socket they need
- Terminal-launched sessions keep their inherited `SSH_AUTH_SOCK` instead of being overwritten by a different login-shell value
- the server-side startup path is left unchanged so this PR stays desktop-scoped

## UI Changes

None.

## Checklist

> [!NOTE]
> ### Backfill SSH_AUTH_SOCK from login shell on macOS desktop startup
> - Replaces `fixPath` with a new `syncShellEnvironment` module that reads both `PATH` and `SSH_AUTH_SOCK` from the user's login shell on macOS at startup.
> - `SSH_AUTH_SOCK` is only written if it is not already set in the inherited environment; `PATH` is always updated from the shell.
> - Extends `readEnvironmentFromLoginShell` in `packages/shared/src/shell.ts` to capture multiple named variables in a single shell invocation using bracketed markers and `printenv <NAME> || true`.
> - Errors from the shell invocation are swallowed so a failure leaves the inherited environment intact.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 90d9c4f.</sup>
> 
>

### DIFF
--- a/packages/shared/src/shell.test.ts
+++ b/packages/shared/src/shell.test.ts
@@ -79,12 +79,10 @@ describe("readEnvironmentFromLoginShell", () => {
       ].join("\n"),
     );
 
-    expect(readEnvironmentFromLoginShell("/bin/zsh", ["PATH", "SSH_AUTH_SOCK"], execFile)).toEqual(
-      {
-        PATH: "/a:/b",
-        SSH_AUTH_SOCK: "/tmp/secretive.sock",
-      },
-    );
+    expect(readEnvironmentFromLoginShell("/bin/zsh", ["PATH", "SSH_AUTH_SOCK"], execFile)).toEqual({
+      PATH: "/a:/b",
+      SSH_AUTH_SOCK: "/tmp/secretive.sock",
+    });
     expect(execFile).toHaveBeenCalledTimes(1);
   });
 
@@ -105,11 +103,9 @@ describe("readEnvironmentFromLoginShell", () => {
       ].join("\n"),
     );
 
-    expect(readEnvironmentFromLoginShell("/bin/zsh", ["PATH", "SSH_AUTH_SOCK"], execFile)).toEqual(
-      {
-        PATH: "/a:/b",
-      },
-    );
+    expect(readEnvironmentFromLoginShell("/bin/zsh", ["PATH", "SSH_AUTH_SOCK"], execFile)).toEqual({
+      PATH: "/a:/b",
+    });
   });
 
   it("preserves surrounding whitespace in captured values", () => {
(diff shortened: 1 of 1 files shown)


apps/desktop/src/fixPath.ts
juliusmarminge: how bout this one? https://github.com/stickerdaniel/t3code/blob/2be451fd586270e42d90bf610fa3e389ae508504/apps/server/src/os-jank.ts#L5-L17

stickerdaniel: The Electron app fixes its env first and the bundled backend inherits that via `...process.env` when spawned so the desktop path is covered

Issue #830, opened 2026-03-10
refactor: split ChatView to smaller subcomponents


juliusmarminge:
extremely long. should be broken down to smaller pieces, both for maintainability, readability and React compiler efficiency

don't have time right now (bigger fish to fry) but good first issue for someone looking to contribute!

would be good with a couple smaller PRs (feel free to stack!) instead of one large one!

Ymit24:
Went ahead and made a PR to split out all of the components, that way we can start refactoring them independently without entering merge conflict hell.

LikoKiko:
Hi! I’d like to make my first contribution and PR #860 already covered much of #830 is there a small remaining task I can take?

juliusmarminge:
Closing as completed by merged PR #860, which split `ChatView` into focused components. Any current decomposition need should be filed as fresh tech debt against today's file rather than keeping the original task open.

PR #985, opened 2026-03-12
ChatView refactor split out ChatComposer component and some scrolling hooks

Ymit24:
## What Changed

1. Extracted hooks from ChatView for handling scrolling
2. Cut out composer UI and the purely internal to composer hooks into new ChatComposer component.

relates to #830 

## Why

In an ongoing effort to refactor ChatView, I pulled out the ChatComposer UI. This doesn't change the behavior of these hooks, just move them. There is still a lot of composer state intertwined in chatview. I wanted to do another structural move first before actually, finally, reworking the state flow. I think it will be easier to review+merge by doing this move first then the refactor, otherwise its hard to see what's changing.

As a quick win, i also pulled out a bunch of scroll related hooks and moved to a custom hook, this just reduces noise in ChatView without changing much. I think there may be more like this we can do in the future. Number 1 goal is just get stuff out of ChatView so more people can independently work on refactoring those sections without conflicting.

Feedback is very much welcomed!

## UI Changes

No UI changes here, should be purely technical.

## Checklist

- [~] This PR is small and focused (technically this does two things so i wouldn't be against splitting it into two PRs if that was preferable)

## Future Work

With this split out, it is now possible to work up and down, so rework data flow between ChatView<->ChatComposer but also work to break out more stuff inside ChatComposer into more sub components.

> [!NOTE]
> ### Extract auto-scroll and composer logic from `ChatView` into dedicated hook and component
> - Moves message list auto-scroll logic into a new [`useMessageListAutoScroll`](https://github.com/pingdotgg/t3code/pull/985/files#diff-afe984b6ce55154a2b4abaf5512dc8f3a19640a90d8908327024cb089220225e) hook that pins scroll to bottom, cancels stickiness on user scroll-up, and reschedules on thread changes or new messages.
> - Extracts the entire composer UI and state into a new [`ChatComposer`](https://github.com/pingdotgg/t3code/pull/985/files#diff-c968a393420901e312c24fcfdef204d8969fa91d9c2df9c52d32b6612bcc8d9a) component, accepting provider/model selection, image attachments, pending approvals, and send/interrupt/implement callbacks as props.
> - [`ChatView`](https://github.com/pingdotgg/t3code/pull/985/files#diff-4b49e092ccd43be0f0de24abe85ba522e09f04288a5d84253b0263e1a389400e) is simplified by replacing inline scroll refs, event handlers, and the composer form block with the new hook and component.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 88f950d.</sup>
> 
>

### DIFF
--- a/.github/workflows/pr-size.yml
+++ b/.github/workflows/pr-size.yml
@@ -6,26 +6,20 @@ on:
 
 permissions:
   contents: read
-  issues: write
-  pull-requests: read
 
 jobs:
-  label:
-    name: Label PR size
+  prepare-config:
+    name: Prepare PR size config
     runs-on: ubuntu-24.04
-    concurrency:
-      group: pr-size-${{ github.event.pull_request.number }}
-      cancel-in-progress: true
+    outputs:
+      labels_json: ${{ steps.config.outputs.labels_json }}
     steps:
-      - name: Sync PR size label
+      - id: config
+        name: Build PR size label config
         uses: actions/github-script@v7
         with:
+          result-encoding: string
           script: |
-            const issueNumber = context.payload.pull_request.number;
-            const additions = context.payload.pull_request.additions ?? 0;
-            const deletions = context.payload.pull_request.deletions ?? 0;
-            const changedLines = additions + deletions;
-
             const managedLabels = [
               {
                 name: "size:XS",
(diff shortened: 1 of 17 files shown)


Ymit24:
Going to close this actually. I have been reading through the code and considering some other options for refactoring. Definitely harder than i anticipated to get it in a nice state 😅.

PR #1027, opened 2026-03-13
Guard desktop startup against legacy provider state

Amer-alsayed:
## What Changed
- Desktop preload fetches the authenticated WS URL via IPC so the renderer always connects to the correct backend endpoint.
- ProviderSessionDirectory drops unknown persisted providers (e.g., legacy `gemini`) instead of failing startup.
- OrchestrationEventStore skips legacy/unreadable events and logs a warning rather than crashing the server.

## Why
Persisted legacy provider data can cause orchestration event decode failures at startup, which restarts the backend and leaves the UI stuck (e.g., Add project hangs). This keeps startup resilient and preserves UX while still logging the issue.

## UI Changes
None.

## Checklist
- [x] No UI changes
- [x] Tests run: `bun fmt`, `bun lint`, `bun typecheck`

> [!NOTE]
> ### Guard desktop startup against legacy provider state and undecodable events
> - `readFromSequence` in [OrchestrationEventStore.ts](https://github.com/pingdotgg/t3code/pull/1027/files#diff-6bb6ea3b72cbe6e78aa809efd497a676913960aaf3e73fd0d10a07ffd20ddd6d) now skips individual decode failures (logging a warning) rather than failing the stream, and advances pagination based on the last row sequence instead of the last decoded event.
> - `getBinding` in [ProviderSessionDirectory.ts](https://github.com/pingdotgg/t3code/pull/1027/files#diff-c811ed073701a673dc819f148e806c2381adade21beb4b0405ea7d712475f968) deletes and drops invalid provider sessions instead of propagating a decode error, preventing stale/legacy state from blocking startup.
> - The desktop preload now fetches `wsUrl` synchronously from the main process via a new `desktop:get-ws-url` IPC channel instead of reading from an environment variable, returning `null` on failure.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c038d0c.</sup>
> 
>
>


Amer-alsayed:
@juliusmarminge Yes, that’s what I ran into.

The failure mode here was older persisted local state containing legacy provider/runtime data and orchestration events that the current startup path no longer decoded successfully, which caused bootstrap to fail and the backend to restart repeatedly.

I agree the right long-term fix is to keep persisted state backwards compatible through explicit migrations in `apps/server/src/persistence/Migrations` whenever we introduce breaking persistence changes. I scoped this PR as a small defensive reliability fix so already-incompatible state does not leave the app unusable, but I think the proper follow-up would be a migration that normalizes or removes legacy provider-specific persisted rows/events during upgrade.

juliusmarminge:
> @juliusmarminge Yes, that’s what I ran into.
> 
> The failure mode here was older persisted local state containing legacy provider/runtime data and orchestration events that the current startup path no longer decoded successfully, which caused bootstrap to fail and the backend to restart repeatedly.
> 
> I agree the right long-term fix is to keep persisted state backwards compatible through explicit migrations in `apps/server/src/persistence/Migrations` whenever we introduce breaking persistence changes. I scoped this PR as a small defensive reliability fix so already-incompatible state does not leave the app unusable, but I think the proper follow-up would be a migration that normalizes or removes legacy provider-specific persisted rows/events during upgrade.

you know when we broke your state? were you using some fork maybe? we did make some breaking changes in the 0.0.0-alpha releases but since 0.0.1 every change should have been stable

juliusmarminge:
Closing because this compatibility guard came from unsupported fork data, targets an old provider architecture, and risks hiding corrupt events.

PR #1078, opened 2026-03-14
Stabilize checkpoint diffs for sparse turns and improve unavailable-diff UX

Meetpatel006:
## What Changed

This PR fixes checkpoint diff behavior for sparse turn histories and improves Diff Panel handling when checkpoints are unavailable.

- Added fallback checkpoint resolution at-or-before requested turn count to avoid hard failures when exact turn checkpoints are missing.
- Added dedupe for repeated placeholder checkpoint events (`turn.diff.updated`) for the same thread+turn.
- Removed the guard that skipped checkpoint capture when `activeTurnId` pointed to a different turn, so completed turns are still captured.
- Updated Diff Panel behavior to avoid endless loading in all-turn mode when missing checkpoints are present.
- Added explicit unavailable-checkpoint messaging and improved empty/loading/error states.
- Added/updated tests for fallback resolution, dedupe behavior, and Diff Panel availability logic.

## Why

Users could hit unreliable or confusing diff behavior:
- Sparse checkpoint timelines could fail when exact turn checkpoints were not available.
- Repeated provider diff updates could create unstable placeholder checkpoint state.
- All-turn Diff Panel could appear stuck loading when one or more checkpoints were missing.
- Error/empty states in the Diff Panel were not clear enough.

This approach keeps behavior predictable under partial/missing checkpoint data, avoids duplicate placeholder writes, and provides clearer UI feedback without expanding scope beyond bug fixes.

### Before
1. All-turn Diff Panel kept loading (stuck loading state).
Screenshot 2026-03-14 005501

2. Error state shown before the fix.
Screenshot 2026-03-14 005507

### After
1. All-turn Diff Panel with better UX and no stuck loading.
Screenshot 2026-03-14 004035

2. Explicit state when a checkpoint is missing.
Screenshot 2026-03-14 004042

3. Normal Diff Panel state.
Screenshot 2026-03-14 004028

## Checklist

- [x] I included a images for interaction changes

> [!NOTE]
> ### Stabilize checkpoint diffs for sparse turns and improve unavailable-diff UX
> - `getTurnDiff` in [CheckpointDiffQuery.ts](https://github.com/pingdotgg/t3code/pull/1078/files#diff-a7f6357814d1957a562de3ea25f26bd4dc2f18e4ccf790082ae04e2bc484da51) now resolves to the nearest checkpoint at or before the requested turn count instead of throwing when an exact match is missing.
> - The checkpoint reactor in [CheckpointReactor.ts](https://github.com/pingdotgg/t3code/pull/1078/files#diff-79dc41615f79575ae1404bd102df915797fb24ba34989557f80e229d305521fb) now captures completed turns even when the session's `activeTurnId` references a different turn.
> - [ProviderRuntimeIngestion.ts](https://github.com/pingdotgg/t3code/pull/1078/files#diff-22a6c9818b956463b848a73a5b3e787b44a144bd003b1cc25fa2b1b3110cdea7) deduplicates placeholder checkpoints so repeated `turn.diff.updated` events for the same turn only enqueue one placeholder.
> - [DiffPanel.tsx](https://github.com/pingdotgg/t3code/pull/1078/files#diff-432b64436ab58eaa9098a7db8382164d4d30470236045f6b3844eef9f8a257f3) skips the diff query when a selected turn's checkpoint is missing or lacks a ref, showing a descriptive message instead of a loading/error state.
> - `isCheckpointTemporarilyUnavailable` now treats `'checkpoint ref is unavailable for turn'` as a retriable condition.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b0fa258.</sup>
> 
>

### DIFF
--- a/apps/server/src/checkpointing/Layers/CheckpointDiffQuery.test.ts
+++ b/apps/server/src/checkpointing/Layers/CheckpointDiffQuery.test.ts
@@ -20,8 +20,11 @@ function makeSnapshot(input: {
   readonly threadId: ThreadId;
   readonly workspaceRoot: string;
   readonly worktreePath: string | null;
-  readonly checkpointTurnCount: number;
-  readonly checkpointRef: CheckpointRef;
+  readonly checkpoints: ReadonlyArray<{
+    readonly turnId: TurnId;
+    readonly checkpointTurnCount: number;
+    readonly checkpointRef: CheckpointRef;
+  }>;
 }): OrchestrationReadModel {
   return {
     snapshotSequence: 0,
@@ -62,17 +65,15 @@ function makeSnapshot(input: {
         messages: [],
         activities: [],
         proposedPlans: [],
-        checkpoints: [
-          {
-            turnId: TurnId.makeUnsafe("turn-1"),
-            checkpointTurnCount: input.checkpointTurnCount,
-            checkpointRef: input.checkpointRef,
-            status: "ready",
-            files: [],
-            assistantMessageId: null,
-            completedAt: "2026-01-01T00:00:00.000Z",
-          },
-        ],
+        checkpoints: input.checkpoints.map((checkpoint) => ({
+          turnId: checkpoint.turnId,
+          checkpointTurnCount: checkpoint.checkpointTurnCount,
+          checkpointRef: checkpoint.checkpointRef,
+          status: "ready" as const,
+          files: [],
+          assistantMessageId: null,
+          completedAt: "2026-01-01T00:00:00.000Z",
+        })),
         session: null,
       },
     ],
@@ -96,8 +97,13 @@ describe("CheckpointDiffQueryLive", () => {
       threadId,
       workspaceRoot: "/tmp/workspace",
       worktreePath: null,
-      checkpointTurnCount: 1,
-      checkpointRef: toCheckpointRef,
+      checkpoints: [
+        {
+          turnId: TurnId.makeUnsafe("turn-1"),
+          checkpointTurnCount: 1,
+          checkpointRef: toCheckpointRef,
+        },
+      ],
     });
 
     const checkpointStore: CheckpointStoreShape = {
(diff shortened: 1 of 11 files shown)


juliusmarminge (commented):
the root cause of this issue is that we're not handling subagents very well yet (they emit unique turn ids so a single user-assistant pass emits multiple turns but we are not capturing these), and codex is constantly 'improving' them. checkpointing worked flawlessly before we shipped, never had issues.

This seems like another bandaid to me(?), I was working on a real and proper fix earlier this week but got distracted.

Really like the UI improvements though!

Meetpatel006:
> the root cause of this issue is that we're not handling subagents very well yet (they emit unique turn ids so a single user-assistant pass emits multiple turns but we are not capturing these), and codex is constantly 'improving' them. checkpointing worked flawlessly before we shipped, never had issues.
> 
> This seems like another bandaid to me(?), I was working on a real and proper fix earlier this week but got distracted.
> 
> Really like the UI improvements though!


Thanks for the context on the subagent root cause that's genuinely helpful to have documented.

I dug into the codex-rs app-server some of the source to understand the event model better before putting this together. A few things that stood out: `turn/diff/updated` is emitted as a full snapshot keyed on `threadId + turnId`, not incremental fragments so the dedupe logic here is actually consistent with how codex expects clients to handle it (replace-in-place on same thread+turn). The `turn/completed` boundary is what codex treats as the finalization signal, and durable `turn/history` events are what replay rebuilds stable state from  transient deltas are intentionally ephemeral on codex's side.

The subagent fan-out case is documented in the app-server model too  spawn/collab flows carry sender/receiver thread relationships, meaning one user-visible exchange can fan out into multiple turns across multiple threads. So you're correct that the root fix needs to be thread-aware first, then turn-aware, not flattened. This PR doesn't go there.

That said, I'd push back slightly on "bandaid" for the PR. The surface-area is really two separate things:

1. **Structural mitigations** (fallback resolution, dedupe, activeTurnId guard removal) agreed these don't fix root cause. If your proper subagent-aware checkpoint model lands, some of these get replaced and that's fine.

2. **UI/UX fixes** (explicit unavailable states, no stuck loading, clear messaging) these are independently correct regardless of what the checkpoint model looks like underneath. Per how codex handles it, missing refs should produce an explicit unavailable state, not endless loading. That part isn't a bandaid, it's just correct defensive UI.

Happy to split the PR if that unblocks review  UI changes fast-tracked, structural mitigations held until your fix lands. Or if you want to share what you had in mind for the subagent fix, I can rebase onto that approach instead.

PR #1083, opened 2026-03-14
feat(web): add arrow-key navigation for sidebar 

Xanacas:
## Summary
- Add keyboard arrow navigation (Up/Down/Left/Right) for the sidebar project/thread tree
- ArrowLeft collapses a project or jumps to its parent; ArrowRight expands or focuses the first child thread
- Skips collapsed project threads when navigating with Up/Down
- Uses `aria-expanded` for accessibility instead of a custom `data-expanded` attribute
- Extracted navigation logic into `Sidebar.logic.ts` with 12 unit tests

> [!NOTE]
> ### Add arrow-key navigation for sidebar projects and threads
> - Adds `handleSidebarArrowNavigation` in [Sidebar.logic.ts](https://github.com/pingdotgg/t3code/pull/1083/files#diff-529ae8997a33774d7ec3514d7abdb655942eaa993f71e6ec6728c892285d251a) to handle `ArrowUp`/`ArrowDown`/`ArrowLeft`/`ArrowRight` key events on the sidebar menu.
> - `ArrowUp`/`ArrowDown` move focus between projects and threads, skipping threads under collapsed projects.
> - `ArrowLeft` on a thread focuses its parent project; on an expanded project, collapses it. `ArrowRight` expands a collapsed project or focuses its first thread.
> - Navigation is wired in [Sidebar.tsx](https://github.com/pingdotgg/t3code/pull/1083/files#diff-12c9f3c619e6ab8dda238e965323ac96e62f3084e3adfb0671b1a6459aa61ba1) via an `onKeyDown` handler on `SidebarMenu`, with `data-nav-item` and `aria-expanded` attributes added to project and thread buttons.
> - Navigation ignores events when focus is on an `input` or `textarea`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ce62140.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/Sidebar.logic.ts
+++ b/apps/web/src/components/Sidebar.logic.ts
@@ -1,8 +1,11 @@
+import type React from "react";
+import type { ProjectId } from "@t3tools/contracts";
 import type { Thread } from "../types";
 import { cn } from "../lib/utils";
 import { findLatestProposedPlan, isLatestTurnSettled } from "../session-logic";
 
 export const THREAD_SELECTION_SAFE_SELECTOR = "[data-thread-item], [data-thread-selection-safe]";
+export const SIDEBAR_NAV_ITEM_SELECTOR = "[data-nav-item]";
 export type SidebarNewThreadEnvMode = "local" | "worktree";
 
 export interface ThreadStatusPill {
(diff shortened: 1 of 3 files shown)


Xanacas:
@juliusmarminge as you already merged #1079 and mostlikely did read the comment in that PR: This PR is the implementation of the arrow keys, but without the change of the tab order /without limiting the sidebar to only be focusable as one element. 

>My preference would be:
>Tab just focuses the project-list, not an individual entry. switching "focus" between projects happens then by standard WAI-ARIA Treeview pattern (up,down,left,right arrow keys). This would make tab navigation much more practical especially if there are a lot of projects/threads in the sidebar.
Color wise I then would prefer to have the current focused projects/threads ring being the grey tone it currently had accidentally.
>Following this concept, there is a minor change that could matter for some users. currently I can tab to the "new thread" butten behind the project name. this would work anymore. My perspective: If I am using keyboard shortcuts to navigate, I can easily learn (or would have learned already) that mod+shift+n does the same thing. I'd probably just add a "n" keyboard shortcut that works when on a project to create a new thread even if I'm not on a thread of the current focused project.


So this one is really only adding arrow-key navigation following best practices for tree view navigation.
If you'd be willing to also accept the mentioned change towards the tree view only be one focusable entry, please let me know.

apps/web/src/components/Sidebar.logic.ts
```diff
@@ -145,3 +148,73 @@ export function resolveThreadStatusPill(input: {
...
+  const currentIndex = navItemEl ? items.indexOf(navItemEl) : -1;
+  if (currentIndex === -1) return;
+
+  event.preventDefault();
+
+  const currentItem = items[currentIndex]!;
+  const navType = currentItem.dataset.navItem;
+  const projectId = currentItem.dataset.projectId as ProjectId | undefined;
+
+  if (key === "ArrowUp") {
```
juliusmarminge: Should these be configurable and integrated into the centralized keybindigns system?

Xanacas: from my perspective not. I've never come across a different set of key bindings for walking along a tree view and as the arrow keys are bond to just the tree view, I currently can't come up with a scenario where I would want to bind my arrow keys to something other globally.

But what I'm thinking of:
Add some more global actions for key bindings:
- sidebar.treeView.focus
- sidebar.treeView.up
- sidebar.treeView.down
- sidebar.treeView.threadNext 
- sidebar.treeView.threadPrev

For those I would immediately activate/load the thread (denounced). But I wanted to see if you're ok with this one 😅

Also I saw PR #69, what are your thoughts on Tanstack keyboard?

juliusmarminge:
Closing because this implementation conflicts with a heavily rewritten grouped sidebar; the accessibility idea should be reconsidered as a fresh focused change.

PR #1102, opened 2026-03-15
feat(editor): add system editor option from VISUAL/EDITOR

### LINKED ISSUE #1068: [Feature]: Support custom code editors via VISUAL/EDITOR environment variables
### Before submitting

### Area

apps/desktop

### Problem or use case

The list of supported editors is hardcoded. When users need a different editor (e.g., WebStorm, #428), it requires a code change. As adoption grows, individual editor requests won't scale.

### Proposed solution

Read the `VISUAL` / `EDITOR` environment variable and offer it as an additional option in the editor picker — not as a replacement for the current default. The built-in list and its default (Cursor) stay as-is. If the env var points to a known editor, reuse its icon/config. If unknown, show it as a generic "System Editor" option that launches with the directory path as the argument.

### Why this matters

Users of less common editors can use t3code without waiting for their editor to be explicitly added. Contributors don't need to maintain an ever-growing editor list.

### Smallest useful scope

Read `VISUAL` (falling back to `EDITOR`), check if the command is on PATH, and surface it as one extra option in the picker. No settings UI needed.

### Alternatives considered

Adding editors one by one to the hardcoded list (current approach, see #428). A dedicated settings UI for custom editors — viable but heavier for what environment variables already solve.

### Risks or tradeoffs

Some editors may not follow the convention of accepting a directory path as an argument. The generic fallback can't support editor-specific flags like `--goto`.

### Examples or references

Image

### Contribution

binbandit:
## Summary

- add a `System Editor` picker option backed by `VISUAL`/`EDITOR`
- preserve quoted executable paths and known-editor `--goto` behavior for env-based launches
- pass `VISUAL` and `EDITOR` through Turbo so the server actually sees them in `dev`/`start` flows

## Why

Issue #1068 asks for one extra editor option sourced from standard shell editor env vars instead of hardcoding more editors one by one.

The implementation also needs to account for this repo's Turbo setup: without forwarding `VISUAL`/`EDITOR`, the feature can pass unit tests while failing in real `bun run dev` / `start` flows because the server process never receives those env vars.

## Changes

- add a shared `system-editor` id so the existing picker and IPC flow can expose one extra editor choice without widening the protocol shape
- resolve `VISUAL` first, then `EDITOR`, parsing quoted commands and flags without breaking executable paths that include spaces
- reuse known editor launch behavior when the env command resolves to VS Code or Cursor, so `--goto` still works for line/column targets
- only surface `system-editor` when the resolved command is actually available
- add the picker entry in the web UI
- forward `VISUAL` and `EDITOR` through `turbo.json`

## Validation

- `bun run --cwd apps/server test src/open.test.ts`
- `bun fmt`
- `bun lint`
- `bun typecheck`

Closes #1068.

> [!NOTE]
> ### Add 'system-editor' option that resolves the editor from VISUAL/EDITOR environment variables
> - Adds `system-editor` as a recognized editor id in [editor.ts](https://github.com/pingdotgg/t3code/pull/1102/files#diff-952d09e8dcc92b201e44f8f0b5edc71af952e9af9479b5853c41ce6e93d636f7) with no fixed command, and surfaces it as a "System Editor" option in the [OpenInPicker](https://github.com/pingdotgg/t3code/pull/1102/files#diff-e2a98d212ae98eb3ec0902f11d6e8f0daeb2b1729c80a65a8792371f7605e802) when available.
> - `resolveAvailableEditors` in [open.ts](https://github.com/pingdotgg/t3code/pull/1102/files#diff-67f3655f84064d89de7cec58f5e8d2b9d406dd21986dcefc77c335e3036b65be) resolves the system editor command from `VISUAL` or `EDITOR` env vars (in that order) and checks availability with `isCommandAvailable`.
> - On launch, the resolved command is matched against known editors so that `--goto` line/column behavior is reused when the command maps to a known editor (e.g. `code`).
> - Quoted executable paths (e.g. Windows paths with spaces) are preserved via a new `splitCommandString` helper.
> - `EDITOR` and `VISUAL` are added to the Turborepo global env passthrough in [turbo.json](https://github.com/pingdotgg/t3code/pull/1102/files#diff-f8de965273949793edc0fbfe249bb458c0becde39b2e141db087bcbf5d4ad5e3).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b2a2804.</sup>
> 
>
>


juliusmarminge:
Closing as part of an open-PR triage — duplicate of #1301, and both predate the editor-subsystem rework on main (#2659, #2475, #1974), so neither rebases cleanly anymore. A fresh take against current main would be welcome. Thanks!

PR #1133, opened 2026-03-16
fix: consume plans on implementation

### LINKED ISSUE #770: Plan mode support only one plan
Plan mode always falls back to refining the latest existing plan, with no visible way to start a fresh plan in the same thread.

Description When a thread already contains an older proposed plan, switching the thread back into Plan mode automatically enters the “refine latest plan” flow. The composer/banner treats the latest settled proposed plan as the active target by default, so sending in plan mode reuses that old plan instead of letting the user start a new planning turn.

This makes plan mode sticky to the most recent prior plan in the thread. Users who want to create a brand-new plan in the same thread cannot do that explicitly and are pushed into refining the previous one.

Expected If a thread already has a prior plan, plan mode should still offer an explicit choice:

refine/use latest plan
start a new plan
The chosen mode should be visible in the composer UI and should affect what gets dispatched.

Actual Entering plan mode on a thread with an older plan immediately behaves like “refine latest plan,” with no visible control to switch to “new plan.”

Repro

Open a thread and generate a proposed plan.
Leave that thread and later return to it.
Switch the thread into Plan mode again.
Try to create a completely new plan.
Observed result The composer/banner assumes the existing latest plan is the active target and treats the next send as a follow-up/refinement of that plan.

Impact This makes long-running threads awkward to use for multiple planning cycles and can cause users to accidentally refine an outdated plan when they intended to start over.

Image

UtkarshUsername:
You can see my PRs and issues at https://pr-navigator.pages.dev/t3code-utkarsh.
Fixes #770

## What Changed

This changes the primary same-thread `Implement` flow so a proposed plan is consumed when its implementation turn actually starts.

The PR includes the shared plan-consumption plumbing needed to support that behavior end to end:
- proposed plans now carry implementation metadata
- `thread.turn.start` can reference a source proposed plan
- orchestration persists that source-plan reference through turn start projection
- provider runtime ingestion marks the source plan implemented only when the target turn is accepted as started
- the web app treats only unimplemented plans as actionable

On the web side, the same-thread `Implement` action now passes `sourceProposedPlan`. As a result, once the implementation turn starts successfully, the thread no longer treats that plan as actionable and the plan follow-up UI disappears.

This branch intentionally does **not** change `Implement in new thread`. That will follow in a stacked PR reusing the same plumbing.

## Why

Before this change, same-thread implementation switched the thread out of `plan` mode, which hid `Plan Ready` temporarily, but it never consumed the underlying proposed plan in domain state.

That meant the plan could become actionable again later if the thread returned to `plan` mode or another client reloaded from the read model. This also prevented users from making multiple plans in the same thread (see #770). The fix is to make plan consumption explicit and tie it to the accepted start of the implementation turn, not to a UI mode switch.

## UI Changes

Before:
(Plan has already been implemented but it still shows the implementation/refine UI.)
image

After:
(Same-thread `Implement` consumes the plan once the implementation turn actually starts. The thread no longer shows the follow-up plan actions for that consumed plan.)
image

## Checklist

> [!NOTE]
> ### Mark proposed plans as implemented when a follow-up turn starts on an implementation thread
> - When a `thread.turn.start` command includes a `sourceProposedPlan` reference, the decider validates the source plan exists on the same project and embeds it in the `thread.turn-start-requested` event.
> - On an accepted `turn.started` event that matches the provider's expected active turn, `ProviderRuntimeIngestion` dispatches a `thread.proposed-plan.upsert` with `implementedAt` and `implementationThreadId` set on the source plan.
> - Two new DB migrations add `implemented_at`/`implementation_thread_id` to `projection_thread_proposed_plans` and `source_proposed_plan_thread_id`/`source_proposed_plan_id` to `projection_turns`.
> - The web client attaches `sourceProposedPlan` when sending a follow-up turn start and uses the new `hasActionableProposedPlan` helper to hide the follow-up prompt and 'Plan Ready' status for already-implemented plans.
> - Behavioral Change: threads whose plan has been implemented no longer show 'Plan Ready' in the sidebar; they fall through to other status rules.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7939acb.</sup>
> 
>
>


UtkarshUsername:
https://github.com/UtkarshUsername/t3code/pull/2 is stacked on top of this. After this PR is merged, will retarget that to main.

PR #964, opened 2026-03-12
fix(web): surface fatal bootstrap snapshot failure

copypasteitworks:
Refs #961

## Summary

- surface fatal initial `orchestration.getSnapshot()` bootstrap failure instead of swallowing it
- render a blocking recovery view with `Retry snapshot`, `Reload app`, and expandable error details
- keep later snapshot resync failures unchanged

## Why

Today the first snapshot load can fail silently, which can leave the app unusable with no diagnosis or recovery path.

This PR keeps the fix tightly scoped to the initial web bootstrap path only.

## Non-goals

- no reset or clear-state UX
- no provider runtime hardening
- no Codex resume fallback widening
- no server contract changes

## UI Change

When the first snapshot load fails before any successful snapshot has loaded, the app now shows a blocking recovery view that mentions persisted state broadly (`T3CODE_STATE_DIR`, `CODEX_HOME`, or other saved app state), exposes the backend error details, and allows retrying without reload.

## Validation

- `bun run test:browser src/routes/-__root.browser.tsx`
- `bun fmt`
- `bun lint`
- `bun typecheck`

> [!NOTE]
> ### Show blocking recovery view when initial orchestration snapshot fails to load
> - `EventRouter` in [`__root.tsx`](https://github.com/pingdotgg/t3code/pull/964/files#diff-9cf47cc915a8ae96883f175e43540fd214a10fa797409a128613573c92b47100) now tracks whether the first `getSnapshot` call succeeded; on failure it sets an error state and defers bootstrap navigation until a successful retry.
> - A new `InitialSnapshotRecoveryView` component renders a blocking UI with the error details and buttons to retry the snapshot or reload the app.
> - Adds browser tests in [`-__root.browser.tsx`](https://github.com/pingdotgg/t3code/pull/964/files#diff-e6205b04803147fc68e6eedc6bb41baeb23fa3a51794c9e10848b41e0f1443a3) using MSW WebSocket/HTTP mocks to verify the recovery view appears on failure and disappears after a successful retry.
> - Behavioral Change: before this change, a failed initial snapshot would leave the app in a broken/partial state; it now renders a dedicated recovery screen instead of the main UI.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b196590.</sup>
> 
>

Issue #961, opened 2026-03-12
Persisted state corruption can leave T3 Code unusable with no clear recovery path


copypasteitworks:
## Summary

Persisted state corruption can leave T3 Code unusable with no clear recovery path.

This should be treated as a separate reliability track, distinct from the already-open diff and version fixes.

## Symptom

- the app can become broken or unusable on startup
- current bootstrap behavior does not clearly explain that persisted state load failed
- one malformed persisted runtime row can poison list-based recovery paths
- some Codex resume failures caused by state corruption are still treated as fatal even though a fresh thread start would be an acceptable fallback

## Known workaround

In at least one repro, renaming `state.sqlite` allowed the app to recover.

That suggests persisted corruption can poison startup, but the mental model here should be broader than a single DB file.

## Important framing

For this track, `state` should include all persisted surfaces that can poison startup or recovery:

- `T3CODE_STATE_DIR`
- `state.sqlite`
- files under the state dir such as keybindings, logs, attachments, and runtime metadata
- `CODEX_HOME`
- persisted browser/Electron client state

Version drift may sometimes be a downstream symptom of corrupted persisted state, not necessarily a separate root cause.

## Proposed split

1. surface fatal initial bootstrap failure instead of swallowing it
2. harden malformed persisted provider runtime rows so one bad row does not poison listing/startup
3. widen recoverable Codex `thread/resume` fallback for narrow, clearly state-related corruption signatures
4. consider reset/clear-state UX separately only if smaller reliability slices are insufficient

## Explicit non-goals

- do not bundle this with the diff-toggle fix
- do not bundle this with the version-display fix
- do not bundle reset/clear-state UX into the first reliability PR
- do not treat `state.sqlite` as the entire problem statement

## Acceptance criteria for this issue

- maintainers can agree or redirect on the slice order before broader recovery UX is proposed
- follow-up PRs can stay narrowly scoped and upstreamable

PR #1231, opened 2026-03-20
fix(server): auto-recover from corrupted SQLite database on startup

eggfriedrice24:
Refs #961

## Problem

When `state.sqlite` becomes corrupted (truncated, overwritten, or contains non-SQLite data), the backend server crashes immediately on startup. The desktop app then restarts the backend in a loop, producing endless `backend exited unexpectedly (code=1)` messages with no recovery path. Users are stuck unless they manually find and delete the database file.

The root cause is two-fold:

1. `makeSqlitePersistenceLive` in `Sqlite.ts` blindly passes the database path to the SQLite client with no pre-flight validation
2. `NodeSqliteClient.ts` calls `openDatabase()` as a bare synchronous call inside `Effect.gen` - when it throws, the error becomes an unhandled defect that crashes the process

### Corruption detection and auto-recovery (`Sqlite.ts`)

Before opening the database, `makeSqlitePersistenceLive` now validates the file header. Every valid SQLite file starts with the 16-byte magic string `SQLite format 3\0`. If the file exists but has an invalid header:

- The corrupted file is renamed to `state.sqlite.corrupted.<timestamp>` (preserving it for debugging)
- Stale WAL and SHM journal files are cleaned up
- A warning is logged explaining what happened
- A fresh database is created and migrations run normally

### Safety net for database open errors (`NodeSqliteClient.ts`)

The bare `openDatabase()` call is now wrapped in `Effect.try` with `Effect.orDie`, so any throw during database construction produces a properly reported defect with a clear `"Failed to open database"` message instead of an opaque crash.

## Testing

Manually verified with a corrupted database file:

```bash
echo "not a db" > ~/.t3/userdata/state.sqlite
t3code
```

**Before:** crash loop, `file is not a database` repeated every ~500ms, process killed  
**After:**
```
WARN: Corrupted database detected at /path/state.sqlite. Backing up to /path/state.sqlite.corrupted.1774012943326 and creating a fresh database. Session history will be reset.
INFO: Running migrations...
INFO: Migrations ran successfully
```

Header validation was also tested against:
- Corrupted file (garbage text) → detected, recovered
- Empty file (0 bytes) → detected, recovered
- Valid SQLite database → passes, opened normally
- Non-existent file → skipped, fresh database created

## Scope

This is complementary to #964 (draft), which handles the web/UI recovery view for bootstrap snapshot failures. This PR fixes the server side so the backend never gets stuck in a crash loop.

## Test plan

- [ ] Corrupt `state.sqlite` with `echo "not a db" > ~/.t3/userdata/state.sqlite`, start app, verify recovery
- [ ] Delete `state.sqlite` entirely, start app, verify fresh database is created
- [ ] Start app with a valid existing database, verify normal startup (no false positives)
- [ ] Verify backed up `.corrupted.<timestamp>` file is created alongside the new database

> [!NOTE]
> ### Auto-recover from corrupted SQLite database on server startup
> - On startup, `makeSqlitePersistenceLive` in [Sqlite.ts](https://github.com/pingdotgg/t3code/pull/1231/files#diff-47b4fe6e24808f5d59f5b97892cce0d736c01e7a8323b77e0c6a3751ff7ba7fa) validates the existing database file by checking the 16-byte SQLite magic header before initializing.
> - If the header is invalid, the file is renamed to a timestamped `.corrupted` backup and any `-wal`/`-shm` sidecar files are removed (best-effort), then a fresh database is created and a warning is logged.
> - In [NodeSqliteClient.ts](https://github.com/pingdotgg/t3code/pull/1231/files#diff-cf61d9fd2d48cc7f6a5d0ebaaa66f8c39d982f536913a6fb92196591465087aa), `openDatabase()` is now wrapped in `Effect.try` so failures produce a `SqlError` with the message `'Failed to open database'` before terminating the fiber.
> - Behavioral Change: On first startup after corruption, session history is lost and replaced with a fresh database.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 33283db.</sup>
> 
>

### DIFF
--- a/apps/server/src/persistence/Layers/Sqlite.ts
+++ b/apps/server/src/persistence/Layers/Sqlite.ts
@@ -1,8 +1,30 @@
 import { Effect, Layer, FileSystem, Path } from "effect";
 import * as SqlClient from "effect/unstable/sql/SqlClient";
+import { openSync, readSync, closeSync } from "node:fs";
 
 import { runMigrations } from "../Migrations.ts";
 import { ServerConfig } from "../../config.ts";
+
+// First 16 bytes of every valid SQLite database file
+const SQLITE_MAGIC = "SQLite format 3\0";
+
+/** Check whether the file at `dbPath` starts with the SQLite magic header. */
+const isSqliteFileValid = (dbPath: string): Effect.Effect<boolean> =>
+  Effect.sync(() => {
+    try {
+      const fd = openSync(dbPath, "r");
+      try {
+        const buf = Buffer.alloc(16);
+        const bytesRead = readSync(fd, buf, 0, 16, 0);
+        if (bytesRead < 16) return false;
+        return buf.toString("ascii", 0, 16) === SQLITE_MAGIC;
+      } finally {
+        closeSync(fd);
+      }
+    } catch {
+      return false;
+    }
+  });
 
 type RuntimeSqliteLayerConfig = {
   readonly filename: string;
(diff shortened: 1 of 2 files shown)


juliusmarminge:

PR #1301, opened 2026-03-22
Support configured system editors

Jeccoman:
## What Changed

This updates the open-in-editor flow to respect VISUAL and EDITOR when resolving available editors and launching files. If the configured command matches a known editor, the existing editor option is reused. If it points to another runnable editor, a System Editor option is shown instead. The change also adds tests covering configured known editors and unknown configured commands.

## Why

Users with a preferred editor configured through VISUAL or EDITOR could still be blocked by the current hardcoded editor detection flow. This keeps the existing behavior for known editors while allowing configured system editors to work without requiring a dedicated built-in entry first.

## Checklist

- [1 ] This PR is small and focused
- [1 ] I explained what changed and why

> [!NOTE]
> ### Add system editor support via VISUAL/EDITOR environment variables
> - Reads `VISUAL` and `EDITOR` env vars to detect a user-configured system editor, tokenizing the command with basic quote handling and matching it against known built-in editor IDs.
> - Adds `'system-editor'` as a recognized editor ID in the contracts and surfaces it as a "System Editor" option in the editor picker when available.
> - `resolveAvailableEditors` now includes the configured editor even when its default binary is not on PATH, and lists `'system-editor'` for unknown-but-available commands.
> - `resolveEditorLaunch` uses the configured command and args when launching `'system-editor'`, applies `--goto` for file:line:col targets, and returns an explicit error if `'system-editor'` is requested but no env var is configured.
> - Behavioral Change: known editors (e.g. VS Code) can now launch via the configured `VISUAL`/`EDITOR` path instead of their default binary when the default is absent from PATH.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f9c2f48.</sup>
> 
>

### DIFF
--- a/apps/server/src/open.test.ts
+++ b/apps/server/src/open.test.ts
@@ -121,6 +121,54 @@ it.layer(NodeServices.layer)("resolveEditorLaunch", (it) => {
       });
     }),
   );
+
+  it.effect("uses the configured system editor from VISUAL or EDITOR", () =>
+    Effect.gen(function* () {
+      const fs = yield* FileSystem.FileSystem;
+      const path = yield* Path.Path;
+      const dir = yield* fs.makeTempDirectoryScoped({ prefix: "t3-open-configured-editor-" });
+      const editorPath = path.join(dir, "custom editor.CMD");
+      yield* fs.writeFileString(editorPath, "@echo off\r\n");
+      const launch = yield* resolveEditorLaunch(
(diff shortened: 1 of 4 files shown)


juliusmarminge:
Closing as part of an open-PR triage — same feature as #1102, and both predate the editor-subsystem rework on main (#2659, #2475, #1974), so neither rebases cleanly anymore. A fresh take against current main would be welcome. Thanks!

PR #1706, opened 2026-04-03
fix(web): narrow stale send spinner recovery to composer state

hhharryyyyy:
## Root cause

The stale send-spinner symptom is local to the composer, but the session lifecycle is not.

Today the composer stop/running state is driven by `phase === "running"`, which comes from the thread session state. In stale cases, that session state can remain `running` after the latest turn is actually done, so the composer keeps showing the stop state and blocks the next send.

However, `latestTurn.completedAt` is not a safe global replacement for running state. That was the core mistake in `#1700`: assistant completion can happen before the turn is fully settled. `#1704` reverted that broader semantic change because it overcompensated for legitimate running cases. The earlier discussion in `#1393` also correctly identified that follow-up sends and delayed turn-start windows must not clear busy state early.

## Why `#1700` regressed behavior

`#1700` changed shared running semantics by treating a completed `latestTurn` as no longer running, and then reused that across the composer spinner/stop button, footer layout, auto-scroll, and timer behavior.

That was too broad. It made the composer stop showing running state as soon as `latestTurn.completedAt` existed, including cases where assistant output had finished but the turn itself was still legitimately running.

## Why this fix is narrower and safer

This change leaves shared session semantics alone:

- no change to `derivePhase`
- no change to `isLatestTurnSettled`
- no change to broader `Working` / running indicators outside the composer

Instead, it narrows the stale-spinner recovery to the composer primary action only. The composer now stops showing the running/stop state only when:

- the session still says `running`, and
- the latest turn has completed, and
- that same turn already has a finalized checkpoint / turn-diff summary

That final checkpoint is a stronger proof of true turn settlement than `latestTurn.completedAt` alone, because it occurs downstream of full turn completion. This preserves the legitimate-running case that `#1700` broke while still recovering the stale composer state when the session lags behind.

## Exact verification performed

- `bun fmt`
- `bun lint`
- `bun typecheck`
- `bun run --cwd apps/web test src/components/ChatView.logic.test.ts`
- `bun run --cwd apps/web test src/session-logic.test.ts`
- `bun run --cwd apps/web test:browser src/components/ChatView.browser.tsx`

## Regression scenarios covered

- stale spinner after a completed turn:
  - browser test covers a `running` session whose latest turn already has a finalized checkpoint; composer returns to send state instead of staying on stop
- legitimately running session must still show running state:
  - browser test covers a `running` session with assistant completion but without a finalized checkpoint; composer still shows stop/running
- follow-up send must not prematurely clear busy state:
  - browser test covers a completed prior turn followed by a delayed `thread.turn.start`; composer stays in `Sending`
- long pre-send / delayed turn-start behavior:
  - covered by the delayed `thread.turn.start` browser case and existing local-dispatch logic tests

> [!NOTE]
> ### Fix stale send spinner by narrowing composer running state to session phase and checkpoint presence
> - Adds `shouldShowComposerRunningState` in [ChatView.logic.ts](https://github.com/pingdotgg/t3code/pull/1706/files#diff-464e876afd6be3800453dd8cc94d805066953bed62ffd825bbd6fdfd86f4d114) to determine composer running state based on session phase, latest turn completion, and presence of a finalized checkpoint summary for that turn.
> - Updates [ChatView.tsx](https://github.com/pingdotgg/t3code/pull/1706/files#diff-4b49e092ccd43be0f0de24abe85ba522e09f04288a5d84253b0263e1a389400e) to use this function so the Stop button is only shown while the session is running and the latest turn either has no `completedAt` or lacks a finalized checkpoint; otherwise the Send button is restored.
> - Behavioral Change: previously the composer stayed in running/stop state as long as `phase === "running"`; now it reverts to idle/send once a finalized checkpoint exists for the latest turn, even if the session is still running.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1228dab.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/ChatView.browser.tsx
+++ b/apps/web/src/components/ChatView.browser.tsx
@@ -2,6 +2,7 @@
 import "../index.css";
 
 import {
+  CheckpointRef,
   EventId,
   ORCHESTRATION_WS_METHODS,
   type MessageId,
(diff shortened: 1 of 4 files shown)


juliusmarminge:
Closing because the spinner and session behavior was subsequently fixed and refactored through newer changes.

PR #1757, opened 2026-04-05
fix: close claude query in probe finalizer

StayBlue:
## What Changed

`ClaudeProvider.ts` uses `q.close()` for cleanup instead of `abort.abort()`.

## Why

`abort.abort()` could throw `AbortError` during shutdown and crash startup. `q.close()` uses the SDK's own disposal path, so the probe still gets `initializationResult()` and no longer crashes on cleanup.

## Checklist

- [X] I included a video for animation/interaction changes

---

> [!NOTE]
> **Low Risk**
> Small, localized change to probe teardown in Claude provider startup; no auth or data-path logic changes.
> 
> **Overview**
> **`probeClaudeCapabilities`** now keeps the `claudeQuery` handle in outer scope so the `Effect.ensuring` finalizer can dispose it with **`q?.close()`**, not only **`abort.abort()`**.
> 
> Cleanup wraps both **abort** and **close** in **try/catch** so thrown errors (e.g. **`AbortError`** during shutdown) do not bubble up and crash startup while the probe still reads **`initializationResult()`** before teardown.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 088eb5ab52acc953c111e9bd60448601647f6ea5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `probeClaudeCapabilities` to close the Claude query in its finalizer
> The cleanup block in `probeClaudeCapabilities` now calls `q?.close()` on the query object if it was initialized, preventing lingering query sessions. `abort.abort()` is also wrapped in a try/catch so errors during cleanup are swallowed rather than propagated as unhandled exceptions.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 088eb5a.</sup>
> 
>

---

<a href="https://app.devin.ai/review/pingdotgg/t3code/pull/1757" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/server/src/provider/Layers/ClaudeProvider.ts
+++ b/apps/server/src/provider/Layers/ClaudeProvider.ts
@@ -589,11 +589,13 @@ const probeClaudeCapabilities = (
     });
   }).pipe(
     Effect.ensuring(
-      Effect.gen(function* () {
-        yield* Effect.sync(() => {
+      Effect.sync(() => {
+        try {
           if (!abort.signal.aborted) abort.abort();
-        }).pipe(Effect.ignoreCause({ log: false }));
-        yield* Effect.sync(() => q?.close()).pipe(Effect.ignoreCause({ log: false }));
+        } catch {}
+        try {
+          q?.close();
+        } catch {}
       }),
     ),
     Effect.timeoutOption(CAPABILITIES_PROBE_TIMEOUT_MS),
--- a/apps/server/src/provider/Layers/ClaudeProvider.ts
+++ b/apps/server/src/provider/Layers/ClaudeProvider.ts
@@ -551,10 +551,11 @@ const probeClaudeCapabilities = (
   environment?: NodeJS.ProcessEnv,
 ) => {
   const abort = new AbortController();
+  let q: ReturnType<typeof claudeQuery> | undefined;
   return Effect.gen(function* () {
     const claudeEnvironment = yield* makeClaudeEnvironment(claudeSettings, environment);
     return yield* Effect.tryPromise(async () => {
-      const q = claudeQuery({
+      q = claudeQuery({
         // Never yield — we only need initialization data, not a conversation.
         // This prevents any prompt from reaching the Anthropic API.
         // oxlint-disable-next-line require-yield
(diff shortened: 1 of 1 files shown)


apps/server/src/provider/Layers/ClaudeProvider.ts
```diff
@@ -396,14 +396,13 @@ const CAPABILITIES_PROBE_TIMEOUT_MS = 8_000;
  * subscription type information.
  */
 const probeClaudeCapabilities = (binaryPath: string) => {
-  const abort = new AbortController();
+  let q: ReturnType<typeof claudeQuery> | undefined;
```
gameroman, line 554: Should this use [`Effect.acquireUseRelease`](https://effect.website/docs/resource-management/scope/#acquirerelease)?

```ts
const probeClaudeCapabilities = (binaryPath: string) => {
  return Effect.acquireUseRelease(
    Effect.sync(() =>
      claudeQuery({
        prompt: ".",
        options: {
          persistSession: false,
          pathToClaudeCodeExecutable: binaryPath,
          maxTurns: 0,
          settingSources: [],
          allowedTools: [],
          stderr: () => {},
        },
      })
    ),
    (q) =>
      Effect.tryPromise(async () => {
        const init = await q.initializationResult();
        return { subscriptionType: init.account?.subscriptionType };
      }),
    (q) => Effect.sync(() => q.close())
  ).pipe(
    Effect.timeoutOption(CAPABILITIES_PROBE_TIMEOUT_MS),
    Effect.result,
    Effect.map((result) => {
      if (Result.isFailure(result)) return undefined;
      return Option.isSome(result.success) ? result.success.value : undefined;
    })
  );
};
```

StayBlue: Honestly, I don't know. I don't have any experience using Effect. I'd appreciate if someone else could chime in as well.

StayBlue: I think it should be good now. Let me know if it isn't.

shivamhwp:

Thanks for investigating this. We're closing this older patch because main now uses Claude Agent SDK 0.3.260, compared with 0.3.170 on this branch. The current SDK's abort handler already invokes its process-close routine, and newer releases include shutdown-related fixes.

A focused check using the current SDK with a simulated CLI subprocess completed initialization and abort cleanup without an explicit query.close() or an unhandled rejection. This does not establish that every shutdown case is fixed, but we could not reproduce the reported failure through that path.

If this still crashes with current main and its SDK version, please provide a fresh reproduction and a regression test so we can address the remaining failure directly.

PR #1761, opened 2026-04-05
Git Sync Settings - Credential Prompt Control

### LINKED ISSUE #1467: Background git status refresh triggers SSH passphrase prompt and cancels it
## Summary

When a repository remote uses SSH and the SSH key is locked, t3code's passive git status refresh can trigger an SSH key passphrase prompt. The prompt steals focus, then disappears before the user can finish typing because the background git process times out.

This makes normal editing disruptive and makes the app feel like it is interrupting the user unexpectedly.

## Repro

1. Open a repo in t3code whose `origin` uses SSH.
2. Make sure the SSH key is not already loaded into `ssh-agent`.
3. Open the app and wait for git status polling, or focus/reconnect the window.
4. Observe an SSH passphrase prompt appear briefly and then disappear.

## Actual

- Passive git status polling triggers interactive SSH auth UI.
- The passphrase window steals focus.
- The prompt is canceled before the user can finish typing.

## Expected

- Passive/background status refresh should never trigger an interactive auth prompt.
- If ahead/behind counts require remote access, the background refresh should fail quietly when credentials are unavailable.
- Interactive auth should only happen on explicit user actions such as pull/push/fetch.

## Suspected Cause

The git status path appears to do a background upstream refresh with `git fetch --quiet --no-tags ...` before computing status, and that fetch is on a short timeout. If SSH needs a passphrase, the auth dialog is opened by `ssh`, but the fetch process is then canceled before the user can complete it.

Relevant code paths in the current tree:

- `apps/web/src/lib/gitReactQuery.ts`
  - git status polling every 15s and on focus/reconnect
- `apps/server/src/git/Layers/GitCore.ts`
  - `statusDetails()` calls upstream refresh first
  - `fetchUpstreamRefForStatus()` runs a background `git fetch`
  - that background fetch uses a 5s timeout

## Suggested Fix

Make the background upstream refresh non-interactive, for example by setting non-interactive git/ssh env for that status-refresh fetch only, or by skipping the remote refresh during passive polling when credentials are not already available.

That keeps ahead/behind refresh working when auth is already available, without stealing focus or popping up passphrase windows during normal UI polling.

mrinc:
Closes #1467

## Problem

t3code performs a background `git fetch` every ~15 seconds to compute ahead/behind counts.
When a repository uses SSH remotes and the user's SSH agent (e.g. 1Password) requires
interactive confirmation, this causes repeated authentication dialogs that steal focus
and get killed by the 5-second fetch timeout before the user can respond.

## Solution

Add **server settings** under a new "Git Sync" section in Settings.
All defaults preserve the current behaviour so existing users are unaffected.

| Setting | Type | Default | What it does |
|---|---|---|---|
| **Background git polling** | Select | `"all"` | `All sessions` — all sessions poll (current behaviour). `Active session only` — only the focused session polls; sidebar background-thread queries pause. `Disabled` — no automatic git sync at all. |
| **Quiet background fetches** | Toggle | OFF | Suppresses credential prompts during automatic git fetches by setting `GIT_TERMINAL_PROMPT=0`, `SSH_ASKPASS=""`, `GIT_ASKPASS=""`, `GCM_INTERACTIVE=never` on the fetch subprocess. Useful with SSH agents like 1Password. |

### Polling Mode (`gitPollingMode`)

- **`"all"`** (default) — Every session and the sidebar poll for git status on their normal intervals (15s for active, 60s for sidebar threads). Server-side upstream fetch runs every 15s. No change from current behaviour.
- **`"active-session"`** — The active/focused session polls normally. Sidebar thread status queries stop polling (`refetchInterval: false`). Server-side upstream fetch still runs when triggered by the active session's status query.
- **`"disabled"`** — All client-side polling intervals are disabled (`refetchInterval: false`, `refetchOnWindowFocus: false`). Server-side `refreshStatusUpstreamIfStale` early-returns without executing `git fetch`. Manual operations (push, pull, checkout) still work.

### Quiet Background Fetches (`gitQuietBackgroundChecks`)

When enabled, the `fetchUpstreamRefForStatus` function passes environment variables to the `git fetch` subprocess that suppress all interactive credential prompts:

| Variable | Purpose |
|---|---|
| `GIT_TERMINAL_PROMPT=0` | Prevents git from prompting on the terminal |
| `SSH_ASKPASS=""` | Prevents SSH from spawning an askpass helper (e.g. 1Password dialog) |
| `GIT_ASKPASS=""` | Prevents git's own askpass mechanism |
| `GCM_INTERACTIVE=never` | Prevents Git Credential Manager from showing dialogs (Windows) |

If credentials are already cached/loaded in the agent, the fetch succeeds silently. If not, it fails silently (the fetch already has `allowNonZeroExit: true` and the caller wraps it in `Effect.ignoreCause`).

This toggle is visually disabled when polling mode is `"disabled"` since there are no background fetches to quiet.

### Data Flow

```
Settings UI (Select / Switch)
  → useUpdateSettings() → ServerSettingsService.updateSettings(patch)
  → settings.json on disk (sparse: only non-default values stored)
  → PubSub → WebSocket subscription → client atom update → component re-render
```

### Server-Side (Effect layers)

`GitCoreLive` now depends on `ServerSettingsService` (yielded in `makeGitCore`). Settings are read via `serverSettings.getSettings` with `Effect.catch(() => Effect.succeed(null))` fallback so settings errors don't break git operations.

### Client-Side (React Query)

`gitStatusQueryOptions` and `gitBranchSearchInfiniteQueryOptions` accept an optional `{ pollingEnabled }` parameter that controls `refetchInterval`, `refetchOnWindowFocus`, and `refetchOnReconnect`. Each consumer component derives `pollingEnabled` from `settings.gitPollingMode`.

### Schema & Contracts
- **`packages/contracts/src/settings.ts`** — Add `GitPollingMode` literal type (`"all" | "active-session" | "disabled"`), `gitPollingMode` field (default `"all"`), and `gitQuietBackgroundChecks` boolean (default `false`) to `ServerSettings` + `ServerSettingsPatch`

### Server — Git Fetch Control
- **`apps/server/src/git/Layers/GitCore.ts`** (+52 / -14)
  - Add `env` field to `ExecuteGitOptions` interface and forward it through `executeGit` → `execute` → subprocess spawn
  - Import `ServerSettingsService`, yield it in `makeGitCore`
  - `refreshStatusUpstreamIfStale`: early-return when `gitPollingMode === "disabled"`
  - `fetchUpstreamRefForStatus`: read `gitQuietBackgroundChecks`, conditionally set quiet env vars on the fetch subprocess

### Client — Polling Interval Control
- **`apps/web/src/lib/gitReactQuery.ts`** — Add optional `pollingEnabled` param to `gitStatusQueryOptions` and `gitBranchSearchInfiniteQueryOptions`; controls `refetchInterval`, `refetchOnWindowFocus`, `refetchOnReconnect`
- **`apps/web/src/components/Sidebar.tsx`** — Sidebar thread polling enabled only when `gitPollingMode === "all"`
- **`apps/web/src/components/ChatView.tsx`** — Derive `pollingEnabled` from `gitPollingMode !== "disabled"`
- **`apps/web/src/components/DiffPanel.tsx`** — Same pattern
- **`apps/web/src/components/GitActionsControl.tsx`** — Same pattern (added `useSettings` import)
- **`apps/web/src/components/BranchToolbarBranchSelector.tsx`** — Same pattern for both status and branch search queries (added `useSettings` import)

### Settings UI
- **`apps/web/src/components/settings/SettingsPanels.tsx`** (+80)
  - New `GIT_POLLING_MODE_LABELS` constant
  - New "Git Sync" `SettingsSection` between Providers and Advanced with:
    - **Background git polling** — `Select` with three options (All sessions / Active session only / Disabled)
    - **Quiet background fetches** — `Switch` toggle, disabled when polling mode is "Disabled"
  - Both rows have reset buttons and descriptive helper text  
image

### Test Files — New Layer Dependency
Adding `ServerSettingsService` as a dependency of `GitCoreLive` requires providing `ServerSettingsService.layerTest()` in test layer stacks:
- `apps/server/src/git/Layers/GitCore.test.ts`
- `apps/server/src/git/Layers/GitManager.test.ts`
- `apps/server/src/checkpointing/Layers/CheckpointStore.test.ts`
- `apps/server/src/orchestration/Layers/CheckpointReactor.test.ts`
- `apps/server/src/workspace/Layers/WorkspaceEntries.test.ts`
- `apps/server/src/workspace/Layers/WorkspaceFileSystem.test.ts`

## Verification

- **Build**: `npx turbo run build` — all 5 tasks pass
- **Server startup**: `node dist/bin.mjs` — starts cleanly on `:3773`, no errors
- **Runtime**: Verified `Effect.catch` (not `Effect.catchAll`) is the correct API for Effect 4.0 beta

> [!NOTE]
> ### Add Git Sync settings to control background polling mode and credential prompts
> - Adds two new server settings: `gitPollingMode` (`all` | `active-session` | `disabled`) and `gitQuietBackgroundChecks` (boolean), configurable via a new Git Sync section in the General Settings panel.
> - When `gitPollingMode` is `disabled`, background upstream status refresh is skipped server-side in `GitCore`, and all client-side git status/branch polling (Sidebar, ChatView, DiffPanel, GitActionsControl, BranchToolbarBranchSelector) is suspended.
> - When `gitQuietBackgroundChecks` is enabled, background git fetches suppress interactive credential prompts by setting `GIT_TERMINAL_PROMPT=0`, `GIT_ASKPASS`, `SSH_ASKPASS` empty, and `GCM_INTERACTIVE=never`.
> - Behavioral Change: `gitPollingMode` defaults to `all` and `gitQuietBackgroundChecks` defaults to `false`, preserving existing behavior for users who have not changed settings.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a789d30.</sup>
> 
>
>

### DIFF
--- a/apps/server/src/checkpointing/Layers/CheckpointStore.test.ts
+++ b/apps/server/src/checkpointing/Layers/CheckpointStore.test.ts
@@ -12,6 +12,7 @@ import { GitCoreLive } from "../../git/Layers/GitCore.ts";
 import { GitCore } from "../../git/Services/GitCore.ts";
 import { GitCommandError } from "@t3tools/contracts";
 import { ServerConfig } from "../../config.ts";
+import { ServerSettingsService } from "../../serverSettings.ts";
 import { ThreadId } from "@t3tools/contracts";
 
 const ServerConfigLayer = ServerConfig.layerTest(process.cwd(), {
@@ -19,6 +20,7 @@ const ServerConfigLayer = ServerConfig.layerTest(process.cwd(), {
 });
 const GitCoreTestLayer = GitCoreLive.pipe(
   Layer.provide(ServerConfigLayer),
+  Layer.provide(ServerSettingsService.layerTest()),
   Layer.provide(NodeServices.layer),
 );
 const CheckpointStoreTestLayer = CheckpointStoreLive.pipe(
--- a/apps/server/src/git/Layers/GitCore.test.ts
+++ b/apps/server/src/git/Layers/GitCore.test.ts
@@ -11,12 +11,14 @@ import { GitCore, type GitCoreShape } from "../Services/GitCore.ts";
 import { GitCommandError } from "@t3tools/contracts";
 import { type ProcessRunResult, runProcess } from "../../processRunner.ts";
 import { ServerConfig } from "../../config.ts";
+import { ServerSettingsService } from "../../serverSettings.ts";
 
 // ── Helpers ──
 
 const ServerConfigLayer = ServerConfig.layerTest(process.cwd(), { prefix: "t3-git-core-test-" });
 const GitCoreTestLayer = GitCoreLive.pipe(
   Layer.provide(ServerConfigLayer),
+  Layer.provide(ServerSettingsService.layerTest()),
   Layer.provide(NodeServices.layer),
 );
 const TestLayer = Layer.mergeAll(NodeServices.layer, GitCoreTestLayer);
--- a/apps/server/src/git/Layers/GitCore.ts
+++ b/apps/server/src/git/Layers/GitCore.ts
@@ -36,6 +36,7 @@ import {
   parseRemoteRefWithRemoteNames,
 } from "../remoteRefs.ts";
 import { ServerConfig } from "../../config.ts";
+import { ServerSettingsService } from "../../serverSettings.ts";
 import { decodeJsonResult } from "@t3tools/shared/schemaJson";
 
 const DEFAULT_TIMEOUT_MS = 30_000;
@@ -74,6 +75,7 @@ interface ExecuteGitOptions {
   maxOutputBytes?: number | undefined;
   truncateOutputAtMaxBytes?: boolean | undefined;
   progress?: ExecuteGitProgress | undefined;
+  env?: Record<string, string> | undefined;
 }
 
 function parseBranchAb(value: string): { ahead: number; behind: number } {
@@ -632,6 +634,7 @@ export const makeGitCore = Effect.fn("makeGitCore")(function* (options?: {
   const fileSystem = yield* FileSystem.FileSystem;
   const path = yield* Path.Path;
   const { worktreesDir } = yield* ServerConfig;
+  const serverSettings = yield* ServerSettingsService;
 
   let executeRaw: GitCoreShape["execute"];
 
@@ -770,6 +773,7 @@ export const makeGitCore = Effect.fn("makeGitCore")(function* (options?: {
       cwd,
       args,
       ...(options.stdin !== undefined ? { stdin: options.stdin } : {}),
+      ...(options.env !== undefined ? { env: options.env } : {}),
       allowNonZeroExit: true,
       ...(options.timeoutMs !== undefined ? { timeoutMs: options.timeoutMs } : {}),
       ...(options.maxOutputBytes !== undefined ? { maxOutputBytes: options.maxOutputBytes } : {}),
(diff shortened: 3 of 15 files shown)


juliusmarminge:
The credential-prompt control is built on an older polling path, and current fetch/connection work has superseded the linked use case. Keeping this branch open would mean maintaining a second settings path for behavior now handled elsewhere.

PR #1896, opened 2026-04-10
Fix archived-only project deletion (superseded)

SAKETH11111:
## What changed

This fixes the sidebar bug where a project that only contains archived threads appears empty but still cannot be removed.

The project deletion guard in `Sidebar.tsx` now checks only visible project threads, matching the same archived-thread filtering the sidebar already uses when rendering per-project thread lists.

To keep that logic consistent, I extracted a small shared helper in `Sidebar.logic.ts` and used it in the relevant sidebar call sites.

I also added focused regression coverage for the archived-thread filtering helper.

## Why

The root cause was a mismatch between two sidebar code paths:

- project rendering filtered archived threads out, so archived-only projects looked empty
- project deletion checked raw `threadIdsByProjectId`, which still included archived threads

That made users hit a misleading `Project is not empty` warning even when the UI showed no threads under the project.

## Impact

Projects that only contain archived threads can now be removed successfully from the sidebar.

## Validation

- `bun fmt`
- `bun lint`
- `bun typecheck`
- `bun run --cwd apps/web test -- src/components/Sidebar.logic.test.ts`
- `bun run --cwd apps/server test -- src/provider/Layers/ClaudeAdapter.test.ts`
- `git diff --check`

> [!NOTE]
> ### Fix project deletion check to correctly handle archived-only projects
> - Introduces `getSidebarThreadsByIds` in [Sidebar.logic.ts](https://github.com/pingdotgg/t3code/pull/1896/files#diff-529ae8997a33774d7ec3514d7abdb655942eaa993f71e6ec6728c892285d251a) to map thread IDs to thread objects, filtering out missing entries and archived threads by default.
> - Fixes a bug in [Sidebar.tsx](https://github.com/pingdotgg/t3code/pull/1896/files#diff-12c9f3c619e6ab8dda238e965323ac96e62f3084e3adfb0671b1a6459aa61ba1) where deleting a project with only archived threads was incorrectly blocked; the deletion guard now uses `getSidebarThreadsByIds` which excludes archived threads from the active thread count.
> - Also extends Claude adapter in [ClaudeAdapter.ts](https://github.com/pingdotgg/t3code/pull/1896/files#diff-d7dbffa9ad1a6f15e826a38ade915e1641f1b345d06da87d48cfc1f99b9ccd8a) to handle `server_tool_use` and `mcp_tool_use` content blocks alongside `tool_use`, validated via a new `getClaudeToolStartBlock` helper.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7d3763d.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
+++ b/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
@@ -965,6 +965,89 @@ describe("ClaudeAdapterLive", () => {
     );
   });
 
+  it.effect("tracks server tool blocks as runtime tool items", () => {
+    const harness = makeHarness();
+    return Effect.gen(function* () {
+      const adapter = yield* ClaudeAdapter;
+
+      const runtimeEventsFiber = yield* Stream.take(adapter.streamEvents, 8).pipe(
+        Stream.runCollect,
+        Effect.forkChild,
+      );
+
+      const session = yield* adapter.startSession({
+        threadId: THREAD_ID,
+        provider: "claudeAgent",
+        runtimeMode: "full-access",
+      });
+
+      const turn = yield* adapter.sendTurn({
+        threadId: session.threadId,
+        input: "hello",
+        attachments: [],
+      });
+
+      harness.query.emit({
+        type: "stream_event",
+        session_id: "sdk-session-server-tool",
+        uuid: "stream-server-tool-start",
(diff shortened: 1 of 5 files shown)


SAKETH11111:
Closing and replacing this PR because the branch/title naming needed cleanup.

PR #1901, opened 2026-04-10
Fix Windows source startup script runners

RonWelWork:
## Summary
This PR is now narrowed to the Windows `apps/server` dev startup issue.

On Windows, the current `apps/server` dev script runs `bun run src/bin.ts`. That path fails before normal app use because Bun does not support the PTY-backed server runtime there.

## Changes
- switch `apps/server` `dev` from `bun run src/bin.ts` to `tsx src/bin.ts`
- keep `build` unchanged
- add a focused regression test covering the server dev runtime expectation

## Verification
- `bunx vitest run scripts/package-scripts.test.ts`
- `bun fmt`
- `bun lint`
- `bun typecheck`
- confirmed separately that `scripts/dev-runner.ts` runs under supported Node 24, so the root package-script changes from the earlier revision were removed

## Context
This is intended as a narrow Windows source/dev fix for the server runtime path only.

> [!NOTE]
> ### Fix Windows compatibility of the apps/server dev startup script
> - Replaces `bun run src/bin.ts` with `tsx src/bin.ts` in the `dev` script of [apps/server/package.json](https://github.com/pingdotgg/t3code/pull/1901/files#diff-fafe51f68c0409784dd523adb04b10cb09fe3b7d48c7646c325fecab5fb4ebd2), fixing startup on Windows where `bun` is not reliably available.
> - Adds `tsx ^4.20.6` as a dependency and updates the lockfile accordingly.
> - Adds a test in [scripts/package-scripts.test.ts](https://github.com/pingdotgg/t3code/pull/1901/files#diff-961075058470c820fbd966c620a459b470bae9e12cc2d519905fab53cfa2c289) that asserts the `dev` script uses `tsx` to prevent regressions.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5c38686.</sup>
> 
>
> 

---

> [!NOTE]
> **Low Risk**
> Low risk: changes only the dev-time entrypoint runner and adds a small regression test; production build/start paths are unchanged.
> 
> **Overview**
> Switches the `apps/server` `dev` script from `bun run src/bin.ts` to `tsx src/bin.ts` to use a Node-compatible (Windows-safe) TypeScript runner.
> 
> Adds `tsx` as a dev dependency (and updates `bun.lock`) and introduces `scripts/package-scripts.test.ts` to assert the server `dev` script stays on the expected `tsx` command.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c38686bdb324ccb28d2088b53810e2a5f724237. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


juliusmarminge:
we don't use node 22 so the running ts files doens't apply

RonWelWork:
Good catch. I checked this against supported Node 24, and you're right that the root `node ...ts` part is not a strong argument here: `scripts/dev-runner.ts` does run under Node 24, so I removed those root/package-level changes from the PR.

I narrowed the patch to the Windows `apps/server` dev path only:
- keep root scripts unchanged
- keep `apps/server` `build` unchanged
- switch only `apps/server` `dev` from `bun run src/bin.ts` to `tsx src/bin.ts`
- keep a focused regression test for that server dev runtime expectation

The remaining issue I was targeting is that the current Windows server dev path goes through Bun, and that fails before normal app use because Bun PTY support is unavailable there.

RonWelWork:
Addressed both Cursor Bugbot notes in the latest push:
- restored `configVersion: 1` in `bun.lock` so the PR no longer changes Bun linker behavior
- strengthened the regression test to assert the exact expected dev command (`tsx src/bin.ts`) instead of only rejecting the old Bun command

Re-ran:
- `vitest run scripts/package-scripts.test.ts`
- `fmt`
- `lint`
- `typecheck`

PR #1903, opened 2026-04-10
Fix archived-only project deletion

### LINKED ISSUE #1880: Cannot delete project with only archived threads
## Bug

Right-clicking a project in the sidebar and selecting **Remove project** shows:

> **Project is not empty**
> Delete all threads in this project before removing it.

…even when the sidebar shows **no threads** under that project.

## Root Cause

The deletion guard in `Sidebar.tsx` checks `threadIdsByProjectId[projectId]` which includes **all** non-deleted threads — including archived ones. But the sidebar rendering filters archived threads out (`.filter((thread) => thread.archivedAt === null)`), so they're invisible to the user.

Projects that only contain archived threads appear empty but can't be deleted.

## Suggested Fix

In `handleProjectContextMenu`, filter out archived threads before the emptiness check so it matches what the sidebar displays:

```diff
       const projectThreadIds = threadIdsByProjectId[projectId] ?? [];
-      if (projectThreadIds.length > 0) {
+      const visibleThreadIds = projectThreadIds.filter(
+        (threadId) => sidebarThreadsById[threadId]?.archivedAt === null,
+      );
+      if (visibleThreadIds.length > 0) {
```

And add `sidebarThreadsById` to the `useCallback` dependency array.

SAKETH11111:
Closes #1880

## What was broken

Projects with only archived threads could not be deleted, even though the sidebar showed them as empty.

## Root cause

The delete guard checked all project thread IDs, while the sidebar filtered archived threads out of the visible project state.

## What changed

- reuse shared sidebar thread filtering before the delete check
- keep the delete guard aligned with the visible thread list
- add regression coverage for archived-only project deletion

## Validation

- reproduced the bug locally before the fix
- verified archived-only projects can now be removed in the running app
- ran `bun fmt`
- ran `bun lint`
- ran `bun typecheck`
- ran `bun run --cwd apps/web test -- src/components/Sidebar.logic.test.ts`
- ran `bun run --cwd apps/server test -- src/provider/Layers/ClaudeAdapter.test.ts`
- ran `git diff --check`

---

> [!NOTE]
> **Medium Risk**
> Changes the project deletion guard to use the same non-archived thread filter as the rendered sidebar, which could affect when project removal is permitted. Scope is limited to sidebar thread visibility/deletion logic and adds a small regression test.
> 
> **Overview**
> Fixes a sidebar inconsistency where projects containing only archived threads appeared empty but could not be removed.
> 
> This introduces `getVisibleSidebarThreads()` (non-archived thread filter) and reuses it across sidebar thread list derivations and the *Remove project* delete guard (now keyed off `visibleProjectThreads.length`). Adds a unit test covering archived thread filtering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47b668058b4569600803ffab07e168c9aec76979. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix project deletion to allow removing projects with only archived threads
> Previously, the project deletion handler checked the total thread count (including archived threads), blocking deletion even when all threads were archived. The fix introduces `getVisibleSidebarThreads` in [Sidebar.logic.ts](https://github.com/pingdotgg/t3code/pull/1903/files#diff-529ae8997a33774d7ec3514d7abdb655942eaa993f71e6ec6728c892285d251a) to filter threads to only non-archived ones, and the deletion check in [Sidebar.tsx](https://github.com/pingdotgg/t3code/pull/1903/files#diff-12c9f3c619e6ab8dda238e965323ac96e62f3084e3adfb0671b1a6459aa61ba1) now uses this filtered list.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 47b6680.</sup>
> 
>


SAKETH11111:
Before removal:
t3code-issue1880-before-remove


After removal:
t3code-issue1880-after-remove

juliusmarminge:
Closing as superseded: project deletion behavior has since been fixed through the newer force/archive deletion flows.

PR #1945, opened 2026-04-12
fix: Isolate thread reactor workers to prevent head-of-line blocking

ashvinnihalani:
## What Changed

Users can keep working in healthy threads even if another thread gets stuck starting a session. This replaces the single provider command reactor worker with per-thread drainable workers so provider intents are queued independently by threadId while preserving per-thread ordering, and it keeps the hanging-thread regression coverage with a test that verifies a blocked thread-1 start does not prevent thread-2 from reaching sendTurn.

## Why

A hung or slow thread start could block the shared reactor queue and make unrelated threads look broken. The failure scenario here is one thread stalling during session startup or resume, followed by another thread trying to start a turn and getting stuck behind the same worker. Isolating reactor workers by thread removes that head-of-line blocking, keeps unaffected threads responsive, and preserves serialized processing within each individual thread.

## Checklist

- [X] I included a video for animation/interaction changes

---

> [!NOTE]
> **Medium Risk**
> Moderate risk: changes the reactor’s event-queueing/concurrency model, which could affect ordering, draining behavior, and resource usage across threads if worker lifecycle isn’t handled correctly.
> 
> **Overview**
> Prevents head-of-line blocking in `ProviderCommandReactor` by replacing the single `DrainableWorker` with a per-thread worker map keyed by `threadId`, so provider-intent events are queued/serialized per thread but processed independently across threads.
> 
> Updates `drain` to wait on all active thread workers, and extends the test harness to allow custom `startSession` behavior plus a new regression test proving a hung session start in `thread-1` does not stop `thread-2` from reaching `sendTurn`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bde0e2a6bf097134cf3c1320e2ef600b5b2804ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Isolate `ProviderCommandReactor` thread workers to prevent head-of-line blocking
> - Replaces a single global `DrainableWorker` with a per-thread worker map in [ProviderCommandReactor.ts](https://github.com/pingdotgg/t3code/pull/1945/files#diff-0630e9ad70d5dcd1c8586d29297f9b7d5f5436cc8e10fc88dc22acc511f1450d), keyed by `ThreadId`.
> - Workers are lazily created and memoized via a `workerForThread` effect; the `drain` method now drains all per-thread workers concurrently.
> - Adds a test in [ProviderCommandReactor.test.ts](https://github.com/pingdotgg/t3code/pull/1945/files#diff-24590cda9b5d40bcb845f474304592ddf62ae6d252d40be60e19435aa75347c4) that verifies a hung session start on one thread does not block turn processing on another.
> - Behavioral Change: events that previously queued behind a blocked thread will now process independently.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bde0e2a.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts
@@ -12,7 +12,7 @@ import {
   type TurnId,
 } from "@t3tools/contracts";
 import { Cache, Cause, Duration, Effect, Equal, Layer, Option, Schema, Stream } from "effect";
-import { makeDrainableWorker } from "@t3tools/shared/DrainableWorker";
+import { type DrainableWorker, makeDrainableWorker } from "@t3tools/shared/DrainableWorker";
 
 import { resolveThreadWorkspaceCwd } from "../../checkpointing/Utils.ts";
 import { GitCore } from "../../git/Services/GitCore.ts";
@@ -787,7 +787,19 @@ const make = Effect.gen(function* () {
       }),
     );
 
-  const worker = yield* makeDrainableWorker(processDomainEventSafely);
+  const workersByThreadId = new Map<ThreadId, DrainableWorker<ProviderIntentEvent>>();
+
+  const workerForThread = (threadId: ThreadId) =>
+    Effect.gen(function* () {
+      const existing = workersByThreadId.get(threadId);
+      if (existing) {
+        return existing;
+      }
+
+      const created = yield* makeDrainableWorker(processDomainEventSafely);
+      workersByThreadId.set(threadId, created);
+      return created;
+    });
 
   const start: ProviderCommandReactorShape["start"] = Effect.fn("start")(function* () {
     const processEvent = Effect.fn("processEvent")(function* (event: OrchestrationEvent) {
@@ -799,6 +811,7 @@ const make = Effect.gen(function* () {
         event.type === "thread.user-input-response-requested" ||
         event.type === "thread.session-stop-requested"
       ) {
+        const worker = yield* workerForThread(event.payload.threadId);
         return yield* worker.enqueue(event);
       }
     });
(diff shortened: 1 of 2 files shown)


apps/server/src/orchestration/Layers/ProviderCommandReactor.ts
```diff
@@ -787,7 +787,19 @@ const make = Effect.gen(function* () {
       }),
     );
 
-  const worker = yield* makeDrainableWorker(processDomainEventSafely);
+  const workersByThreadId = new Map<ThreadId, DrainableWorker<ProviderIntentEvent>>();
```
juliusmarminge, line 790: this map is ever growing, workers are never torn down?

ashvinnihalani: Yeah good point, I have never reached a point where the workers threads have grown that much since I regulalry rebuild and clear the thread history. Let me make sure when either we send a session_close id or something similar we remove delete the thread form the map

ashvinnihalani: Maybe this isn't the right abstraction here. I made this because I ran into a very specific use case where I wanted to start a new thread on the existing thread was still starting/hanging. Given that we have a DurableWorker for the ProviderCommandReactor, ProviderRuntimeIngestion, CheckpointReactor it may make more sense to instead make DurableWorker non-blocking for seperate thread events

juliusmarminge: Honestly not sure how much punch the workers are pulling anymore. They did a good job guaranteeing order before when stuff were more async. Now we're more bought into the effect runtime so I think some of the issues the worker aimed to solve went away so I don't think we need these workers in every place we currently have them anymore.

I have some branches with some performance work that I got sidetracked from finishing. Will see if I can dig those up

juliusmarminge:
The head-of-line blocking problem is real, but the per-thread worker map has no lifecycle and grows without bound; discussion also concluded the worker abstraction itself should be redesigned. This should return as a keyed worker/lifecycle change or worker removal.

PR #1960, opened 2026-04-12
fix(server): set session to error state when sendTurn fails

sachssem:
## What Changed

When `sendTurnForThread` fails inside `ProviderCommandReactor`, the `catchCause` handler now also calls `setThreadSession` to transition the session into a terminal state, in addition to the existing failure activity.

- If an active provider binding exists: sets `status: "error"` (preserves binding for recovery).
- If no binding exists (e.g. provider mismatch before session start, or a previously stopped session): sets `status: "stopped"` to avoid creating a synthetic session without a backing `ProviderService` directory entry.

Both paths set `lastError` so the client can detect the failure.

Single-file production change in `ProviderCommandReactor.ts`. Four regression tests added covering existing-session, null-session, stopped-session failure paths, and recovery after a provider-switch error.

## Why

After the client dispatches `thread.turn.start`, the HTTP call succeeds so `turnStartSucceeded = true`. The client then waits for a server-side signal via `hasServerAcknowledgedLocalDispatch`, which checks for `phase === "running"`, `threadError`, or a session snapshot change.

When `sendTurn` fails server-side, the existing handler only appended a failure activity — it never updated the session. If the session was already bound and unchanged, none of the acknowledgment conditions were met: `phase` stayed `"ready"`, `threadError` was `null`, and the session snapshot was identical. This left `isSendBusy` permanently stuck, the send button disabled, and "Working for …" hanging until `Cmd+R`.

Previous attempts to fix this (#1348, #1393, #1542) added client-side workarounds. This change addresses the root cause on the server by ensuring the session always transitions to a detectable state on failure.

## Validation

- `bun fmt` — clean
- `bun lint` — clean
- 26/26 `ProviderCommandReactor` tests pass (4 new)
- Full orchestration + provider suite: 300/300 pass

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes error-handling in `ProviderCommandReactor` to mutate persisted session state on turn-start failures, which can affect client behavior and session reuse/recovery paths. Covered by new regression tests across existing-session and no-session scenarios, reducing risk but still touching orchestration state transitions.
> 
> **Overview**
> Ensures `thread.turn-start-requested` failures no longer only append a `provider.turn.start.failed` activity: the reactor now also calls `setThreadSession` to move the thread session into a detectable terminal state and populate `lastError`.
> 
> On failure, sessions with an existing non-`stopped` binding are marked `status: "error"` (preserving binding for recovery), while cases with no real binding (including previously `stopped` sessions) are marked `status: "stopped"` to avoid creating a synthetic bound session. Tests are updated/added to assert these statuses, `lastError`, and that a subsequent correct-provider turn can recover after a provider-switch failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aee8e69d5a44182080d10521de1f5de73a9b9449. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Set thread session to error state when `sendTurn` fails in `ProviderCommandReactor`
> - When a `thread.turn.start` fails, the reactor now updates the thread session: if an active session existed it transitions to `status: 'error'`, otherwise it creates/updates a session with `status: 'stopped'`.
> - In both cases, `activeTurnId` is cleared and `lastError` is set to `'Provider turn start failed: …'`, alongside the existing `provider.turn.start.failed` activity.
> - New and updated tests in [`ProviderCommandReactor.test.ts`](https://github.com/pingdotgg/t3code/pull/1960/files#diff-24590cda9b5d40bcb845f474304592ddf62ae6d252d40be60e19435aa75347c4) cover the error, stopped, and recovery scenarios.
> - Behavioral Change: previously a failed turn start left the session in an ambiguous state; it now always reflects the failure explicitly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized aee8e69.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
@@ -757,7 +757,9 @@ describe("ProviderCommandReactor", () => {
 
     const readModel = await Effect.runPromise(harness.engine.getReadModel());
     const thread = readModel.threads.find((entry) => entry.id === ThreadId.make("thread-1"));
-    expect(thread?.session).toBeNull();
+    expect(thread?.session).not.toBeNull();
+    expect(thread?.session?.status).toBe("stopped");
+    expect(thread?.session?.lastError).toContain("Provider turn start failed");
     expect(
       thread?.activities.find((activity) => activity.kind === "provider.turn.start.failed"),
     ).toMatchObject({
@@ -1129,6 +1131,8 @@ describe("ProviderCommandReactor", () => {
     expect(thread?.session?.threadId).toBe("thread-1");
     expect(thread?.session?.providerName).toBe("codex");
     expect(thread?.session?.runtimeMode).toBe("approval-required");
+    expect(thread?.session?.status).toBe("error");
+    expect(thread?.session?.lastError).toContain("Provider turn start failed");
     expect(
       thread?.activities.find((activity) => activity.kind === "provider.turn.start.failed"),
     ).toMatchObject({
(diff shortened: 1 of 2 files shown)


sachssem:
Closing this because the change is no longer applicable in its original form. The underlying behavior has since been covered by later changes on main.

PR #2009, opened 2026-04-13
Cursor/harness routing scheduled UI

coah80:
## Checklist

- [ ] This PR is small and focused
- [ ] I explained what changed and why

---

> [!NOTE]
> **High Risk**
> High risk because it introduces a large new harness runtime that can execute shell/filesystem/network tools and adds new auth probing paths, plus changes projection persistence for project icons/thread metadata.
> 
> **Overview**
> Introduces a new CoahCode harness runtime in `apps/server/src/harness`, including a streaming agent loop with **parallel tool execution**, MCP/LSP integration, skill/instruction loading (incl. OpenCode interop), large tool-result spilling, and automatic pre-mutation checkpoints/rollback support.
> 
> Adds scheduled-task tooling (create/list/update/toggle/delete) with natural-language schedule parsing, plus model switching and follow-up steering/queue state helpers. Also adds CLI auth probing helpers/tests for Claude/Codex subscription flows.
> 
> Updates git text-generation to pass `--skip-git-repo-check`, tightens provider typing in routing text generation, and extends orchestration projections/snapshots to persist and query `project.icon` and allow `thread.meta-updated` to change `projectId`. Rebrands desktop/README naming to **CoahCode** and adds a `HANDOFF.md` handover doc.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f98eda9092b17c0d371bea57ced9db80730f5cc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add Harness provider with scheduled tasks, workspace management, and steering UI
> - Introduces a new `harness` provider that routes `codex` and `claudeAgent` sessions through an internal agent loop, supporting Anthropic, OpenAI, and CLI subscription turn execution, with MCP/LSP tool integration and skill discovery.
> - Adds scheduled task infrastructure end-to-end: DB migration, `ScheduledTasksService`, a cron runner that executes agent loops per task, WebSocket RPCs, React Query hooks, and a settings panel at `/settings/scheduled`.
> - Adds a follow-up steering system in the chat composer: users can queue follow-ups while a turn is running or interrupt to steer, with localStorage persistence and drag-to-reorder in a new `ComposerQueuedFollowUpsPanel`.
> - Adds workspace discovery and project management: sidebar auto-creates a Home project, threads can be dragged between projects using a physics-based drag ghost, and a `WorkspacePicker` in settings lets users switch or create workspaces.
> - Adds project icon support with preset and custom image options via `ProjectIconDialog`, stored in `icon_json` and propagated through projections, events, and the sidebar.
> - Renames the app from "T3 Code" to "CoahCode" across display names, branding, and electron manifests.
> - Risk: the scheduled task runner executes agent loops server-side on a timer; tasks without upstream auth are silently skipped, and concurrent duplicate runs are prevented in-memory (state is lost on restart).
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 8f98eda. 51 files reviewed, 40 issues evaluated, 13 issues filtered, 15 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> <details>
> <summary>apps/server/src/harness/engine/checkpoint.ts — 1 comment posted, 5 evaluated, 1 filtered</summary>
>
> - [line 78](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/server/src/harness/engine/checkpoint.ts#L78): On Windows, `destPath.lastIndexOf("/")` returns -1 because `join()` uses `\` as the path separator. `substring(0, -1)` then returns an empty string, causing `fs.mkdir("", { recursive: true })` to fail or create files in the wrong location. Use `path.dirname(destPath)` instead of manual string manipulation. <b>[ Out of scope (triage) ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/harness/engine/home.ts — 1 comment posted, 2 evaluated, 1 filtered</summary>
>
> - [line 97](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/server/src/harness/engine/home.ts#L97): In `createProject`, if all name variants from `baseName` through `baseName 1000` already exist, the loop exits after setting `projectPath` to the 1000th variant without verifying it's actually available. The subsequent `fs.mkdir` with `recursive: true` succeeds silently on the existing directory, and the function returns a `WorkspaceInfo` pointing to a pre-existing project rather than a newly created one. <b>[ Out of scope (triage) ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/harness/index.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 29](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/server/src/harness/index.ts#L29): `getNextRunTime` and `describeCron` are imported into `scheduler.ts` from `@t3tools/shared/scheduler` but are NOT re-exported from `scheduler.ts`. This export statement will fail at compile/runtime because these names do not exist in the `./engine/scheduler` module's exports. Either add `export { describeCron, getNextRunTime }` to `scheduler.ts`, or change this export to import directly from `@t3tools/shared/scheduler`. <b>[ Out of scope (triage) ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/harness/skills/loader.ts — 1 comment posted, 2 evaluated, 1 filtered</summary>
>
> - [line 144](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/server/src/harness/skills/loader.ts#L144): `getSkillSource` uses `dir.startsWith(workspaceRoot)` which incorrectly matches sibling directories. For example, if `workspaceRoot` is `/home/user/project` and `dir` is `/home/user/project-backup/skills`, `startsWith` returns `true` even though the directory is outside the workspace. This would misclassify an external skill directory as `"project"` when it should be `"config"`. The check should ensure the match is at a directory boundary, e.g., `dir === workspaceRoot || dir.startsWith(workspaceRoot + "/")`. <b>[ Cross-file consolidated ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/harness/tools/index.ts — 1 comment posted, 6 evaluated, 4 filtered</summary>
>
> - [line 12](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/server/src/harness/tools/index.ts#L12): `toolHandlers` is missing handlers for `AskQuestion` and `SwitchMode` tools that are defined in `TOOL_DEFINITIONS`. When `getToolDefinitions()` advertises these tools to the model and the model invokes them, they are not handled by the special cases in `loop.ts` (MCP, LSP, Skill, ScheduledTask), so they fall through to `executeTool()` which returns `Unknown tool` error because they are not in `toolHandlers`. <b>[ Exceeded comment limit ]</b>
> - [line 68](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/server/src/harness/tools/index.ts#L68): Path validation bypass via directory prefix collision. The check `resolved.startsWith(workspaceRoot)` passes for paths that share a prefix but are outside the workspace. For example, if `workspaceRoot` is `/home/user/workspace`, a path resolving to `/home/user/workspaceevil/secret.txt` passes the check. Same issue with `/tmp` allowing `/tmpevil/...`. The check should be `resolved.startsWith(workspaceRoot + "/") || resolved === workspaceRoot` (and similarly for `/tmp/`). <b>[ Exceeded comment limit ]</b>
> - [line 90](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/server/src/harness/tools/index.ts#L90): The `spawn` timeout option (line 90) and the manual `setTimeout` (lines 103-109) both use `blockUntilMs`, causing contradictory behavior: when the timeout fires, the manual timer resolves the promise with a "Command running in background" message, but the spawn's `timeout` option simultaneously kills the process. The command is reported as running in background while actually being terminated. <b>[ Exceeded comment limit ]</b>
> - [line 177](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/server/src/harness/tools/index.ts#L177): Using `substring(0, resolved.lastIndexOf("/"))` to extract the parent directory fails when writing to a root-level file. For example, if `resolved` is `/file.txt`, then `lastIndexOf("/")` returns `0`, making `substring(0, 0)` return an empty string `""`. Subsequently, `fs.mkdir("", { recursive: true })` will throw an error. This is reachable because `resolvePath` returns absolute paths unchanged, so passing `path: "/somefile.txt"` in args triggers this. The correct fix is to use `dirname` from the `path` module instead of manual string manipulation. <b>[ Cross-file consolidated ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/harness/tools/scheduledTasks.ts — 0 comments posted, 4 evaluated, 3 filtered</summary>
>
> - [line 102](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/server/src/harness/tools/scheduledTasks.ts#L102): The function `naturalLanguageScheduleToCron` does not handle `"every weekday"` or `"every weekdays"` inputs, even though `extractSchedulePhrase` explicitly recognizes these as valid schedule phrases. When `schedule_text` is provided directly (e.g., `schedule_text: "every weekday"`), no match is found and the function returns `null`, causing an error. However, when the same phrase is extracted from the prompt text, it goes through a special normalization step (`extractedScheduleText.replace(/^every\s+weekdays?/i, "weekdays")`) in `resolveCronExpression` before being passed to this function. This inconsistency means users providing `schedule_text: "every weekday"` will get an error while users writing "every weekday" in their prompt will succeed. <b>[ Exceeded comment limit ]</b>
> - [line 211](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/server/src/harness/tools/scheduledTasks.ts#L211): `extractSchedulePhrase` normalizes whitespace via `normalizeWhitespace(value)` before matching patterns, so the returned match string has collapsed whitespace. However, in `resolveCronExpression` (lines 211, 231-232, 256), this normalized match is used with `prompt.replace(schedulePhrase, " ")` against the original unnormalized prompt. If the original prompt contains irregular whitespace like `"every   5   minutes"`, the replacement will fail because the normalized `"every 5 minutes"` won't be found, leaving the schedule phrase in the cleaned prompt. <b>[ Exceeded comment limit ]</b>
> - [line 400](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/server/src/harness/tools/scheduledTasks.ts#L400): In the `update` action, providing `cron_expression` or `schedule_text` without a `prompt` silently ignores the schedule update. Line 400-401 requires `rawPrompt.length > 0` before calling `resolveCronExpression`, so if a user calls update with `{action: "update", id: "foo", cron_expression: "0 9 * * *"}` (no prompt), `nextSchedule` will be `null` and the new cron expression is never passed to `manager.update()`. This defeats the user's intent to update only the schedule. <b>[ Exceeded comment limit ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/ProjectFavicon.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 24](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/web/src/components/ProjectFavicon.tsx#L24): The `useState` hook on line 24 is called conditionally — after the early return on lines 15-17. This violates React's rules of hooks, which require hooks to be called in the same order on every render. If `input.projectIcon` changes from falsy to truthy (or vice versa) between renders, React will either crash or produce incorrect behavior because the hook call count changes. <b>[ Exceeded comment limit ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/ScheduledTasks.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 260](https://github.com/pingdotgg/t3code/blob/8f98eda9092b17c0d371bea57ced9db80730f5cc/apps/web/src/components/ScheduledTasks.tsx#L260): In the `onSubmit` handler (lines 251-261), `setIsEditing(false)` is called immediately after `onUpdate`, but `onUpdate` triggers an async mutation. If the mutation fails, the form closes anyway, losing the user's edits with no error indication. Compare with the create flow in `ScheduledTasks` which correctly uses `onSuccess` to close the form only after the mutation succeeds. The fix would be to move `setIsEditing(false)` into an `onSuccess` callback passed to the mutation, similar to how `handleCreate` uses `onSuccess: () => { resetCreateForm(); }` <b>[ Exceeded comment limit ]</b>
> </details>
>
>
> </details>
>


coah80:
Wrong target repo; work lives on coah80/coahcode.

PR #1774, opened 2026-04-06
feat: added terminal docking to right

atmikshetty:
## What Changed

 - Added a terminal dock toggle button next to the existing split terminal action in the thread terminal toolbar.
 - Terminals can now be docked either:
     - below the chat view, or
     - on the right side of the chat view.
- Terminal dock position and right-dock width are persisted per thread.

## Why

- The terminal could previously only be docked below the chat view, which limited layout flexibility and made it harder to use alongside the conversation in wider desktop layouts.
- The change fixes that by allowing the terminal to be docked to the right. 
- Also, the dock state persists, so users won't have to manually do it every time they open the app. 

## UI Changes

Terminal Docked to the right: 
image

Works when the app is minimised as well: 
image

Redocking back to below the chat: 
image

Small Video showing the full change: 
https://github.com/user-attachments/assets/3a6f0f84-5040-4dd1-b139-7c2e94650050

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] I included before/after screenshots for any UI changes
- [x] I included a video for animation/interaction changes

---

> [!NOTE]
> **Medium Risk**
> Moderate UI/state changes: introduces persisted terminal dock/width with a storage migration and alters terminal resize/open behavior, which could regress layout or terminal sizing across existing users.
> 
> **Overview**
> Adds a *dock toggle* to the thread terminal drawer so terminals can be displayed either **below** the chat view or **docked to the right**, including right-side resize handling and updated split-pane layout when docked right.
> 
> Persists the new per-thread `terminalDock` and `terminalWidth` in the terminal Zustand store (with a storage version bump + migration/normalization), updates `ChatView` to render right-docked drawers alongside the main content while keeping bottom-docked drawers below, and extends tests to cover dock/width behavior.
> 
> Additionally, the desktop `electron-launcher` now attempts to automatically run Electron’s `install.js` when `require('electron')` fails with the known “failed to install correctly” error, and `package.json` marks `electron` as a trusted dependency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f328b1dc6a6c662c1f42da98ba8a0d2a8ed5cca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add right-side docking to the terminal drawer
> - Adds a dock toggle button to `ThreadTerminalDrawer` so users can switch the terminal between bottom and right-side docking.
[...]

PR #2089, opened 2026-04-16
fix(desktop): recover when Electron runtime is missing

itsklimov:
## What Changed

This PR fixes desktop startup when the `electron` package is present but its runtime binary was never installed.

The change is intentionally small and limited to the existing Electron launcher path:

- add `electron` to `trustedDependencies` so Bun can run Electron's install lifecycle during normal installs
- teach `apps/desktop/scripts/electron-launcher.mjs` to recover automatically when `require("electron")` fails with Electron's known "failed to install correctly" error
- run Electron's own `install.js` script from the launcher as a one-time recovery path
- retry `require("electron")` after installation succeeds, then continue through the normal launcher flow

## Related PRs

- Extracts and isolates the Electron runtime recovery fix that also appears in #1774.

#1774 is primarily a terminal docking and UI state PR. The Electron recovery change inside it is logically separate and useful on its own.

This PR pulls that desktop startup fix into a small standalone change so it can be reviewed and merged independently, without coupling it to unrelated terminal layout, persistence, and UI behavior changes.

## Why this standalone PR is still useful

- smaller review surface
- no dependency on the broader terminal docking work
- directly addresses a desktop startup failure with a focused fix

## Why

Desktop development currently fails hard in environments where Bun does not execute Electron's install lifecycle, leaving the package metadata in place but omitting the actual Electron runtime binary.

In that state, startup dies before the app can launch:

```text
Electron failed to install correctly, please delete node_modules/electron and try installing again
```

That failure mode is especially awkward because the repository otherwise looks installed correctly. The package resolves, but the runtime it expects on disk is missing.

## Root Cause

The `electron` package resolves its executable path from installation artifacts created by its postinstall step.

When that lifecycle does not run, files such as `path.txt` and the unpacked Electron runtime are missing. The existing launcher called `require("electron")` directly, so desktop startup crashed immediately before any app-specific recovery logic could run.

## How It Works

The new launcher behavior stays narrow and only activates for Electron's specific missing-runtime failure:

1. attempt to resolve the Electron binary with `require("electron")`
2. if resolution succeeds, continue exactly as before
3. if resolution fails with Electron's known missing-install error, run Electron's bundled `install.js`
4. clear the module cache entry and retry `require("electron")`
5. if the retry succeeds, continue through the normal desktop launcher path
6. if installation fails, surface an explicit error instead of silently continuing

A few implementation details matter here:

- the recovery path uses `process.execPath` instead of assuming `node` is available on `PATH`
- the retry path validates success by re-resolving `require("electron")`, not by checking a surrogate file only
- fresh installs should avoid the recovery path entirely because `electron` is now included in `trustedDependencies`

## Impact

This change improves two cases:

- fresh clones and reinstalls are more reliable because Bun is explicitly allowed to run Electron's install step
- already-broken local installs can self-heal on first desktop launch instead of requiring manual deletion and reinstall work

The normal success path is unchanged when Electron is already installed correctly.

## Validation

- `bun run fmt`
- `bun run lint` (passes with the repository's current pre-existing warnings outside this diff)
- `bun run typecheck`
- `bun run dev:desktop` and verified startup progressed past the previous Electron crash into normal desktop boot and server listen

## UI Changes

None.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because desktop startup behavior now conditionally runs Electron’s `install.js` via a synchronous subprocess, which can fail or behave differently across environments and affects local installs/CI.
> 
> **Overview**
> Desktop startup is updated to *recover automatically* when `require('electron')` fails due to Electron’s known “failed to install correctly” missing-runtime error.
> 
> The launcher now detects that specific failure, runs Electron’s bundled `install.js` (via `process.execPath`), clears the module cache, and retries resolving the Electron binary before continuing. Separately, `electron` is added to root `trustedDependencies` so Bun will execute Electron’s install lifecycle during normal installs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1991c61fccfa57b830f10f627fa8a7a3e328689. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix Electron desktop launcher to recover when Electron binary is missing
> When `require('electron')` fails with the 'Electron failed to install correctly' error, the launcher now automatically runs the package's `install.js` via a synchronous child process, clears the module cache, and re-requires the binary path. Non-matching errors are rethrown unchanged. This logic is added to [`electron-launcher.mjs`](https://github.com/pingdotgg/t3code/pull/2089/files#diff-8a117bf8144a87af3b3928053a08e442ffe3b14884df08757b77d55907d596b4) via three new helpers: `isMissingElectronBinaryError`, `ensureElectronInstalled`, and `loadElectronBinaryPath`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c1991c6.</sup>
> 
>
>


juliusmarminge:
Closing as superseded by #2861, which ensures the Electron runtime is installed in release workflows.

PR #2091, opened 2026-04-16
Insert non-image file references on drop/paste

itsklimov:
## Summary

This changes composer file ingestion so that dropping or pasting non-image files inserts file references into the prompt instead of rejecting them.

Images keep the existing attachment flow.
Non-image files now resolve to local file references:

- on desktop, the composer inserts the native file system path
- outside the desktop shell, it falls back to the file name

## Related issues

- Fixes #557
- Fixes #2139

## Prior art / related PRs

- Supersedes #885 for the current codebase shape.

#885 is directionally the same user-facing fix, but it targets the older composer structure. The current implementation path now runs through `ChatComposer`, the current `LocalApi` surface, and the current `ComposerPromptEditor` handle rather than the older `ChatView`-centric flow.

This PR keeps the same product intent, but re-implements it against the current architecture so it is actually mergeable on top of today's code.

- Complementary to #2108, not a duplicate.

#2108 expands the drag/drop surface for URL-backed image drags from other browser tabs. This PR covers the `File`-backed non-image path and paste/drop insertion of local file references. They touch adjacent composer code, but solve different cases.

### Before

Dropping a non-image file into the composer is rejected with:

> Please attach image files only.

image

### After: non-image file drop

Dropping a non-image file inserts its local file reference into the composer.

image

### After: mixed drop

Mixed drops keep image attachments and insert non-image file references into the prompt.

image

## Problem

Today the composer only accepts `image/*` files.
If a user drops or pastes a `.json`, `.csv`, `.txt`, or other non-image file into the chat box, the action is rejected with:

> Please attach image files only.

That is a poor fit for the actual developer workflow here.
In practice, users often want to reference a local file in their prompt rather than upload it as binary message content.

Examples:

- `look at /Users/alice/project/output.json`
- `compare this against ./fixtures/request.txt`
- `inspect the CSV I just dropped`

## Why this approach

I intentionally did **not** implement arbitrary non-image attachments.

That would expand the product contract significantly:

- new attachment semantics for non-image files
- persistence and preview rules
- payload size and MIME policy questions
- provider/runtime expectations for how local files should be interpreted

For this UX, that is unnecessary.
The cleaner behavior is:

- keep `image/*` as true attachments
- treat non-image files as prompt references

This matches how coding-agent users actually work and keeps the existing send contract unchanged.

### Desktop bridge support

Added `getPathForFile(file)` to the desktop bridge contract and preload layer.

[...]

PR #885, opened 2026-03-11
fix: insert native file path when dropping or pasting non-image files

mfazekas:
## What Changed

When dropping or pasting non-image files (e.g. `.csv`, `.json`, `.txt`, `.riv`) onto the chat composer, their native filesystem path is now inserted at the cursor position instead of being silently ignored. Fixes #557.

- **ChatView**: New `addComposerAttachments` helper partitions files into images and non-images. Both `onComposerDrop` and `onComposerPaste` delegate to it. Non-image paths are auto-quoted when containing spaces.
- **ComposerPromptEditor**: Added `insertText(text)` using Lexical's `selection.insertText()` to insert at the real cursor position, correctly handling @mention nodes.
- **preload.ts**: Exposed `webUtils.getPathForFile(file)` via contextBridge (required since Electron's sandbox blocks `file.path`).
- **contracts/ipc.ts**: Added `getPathForFile` to `DesktopBridge` interface.

## Why

Dropping or pasting any non-image file silently failed with an "unsupported file type" error

## UI Changes

| | |
|-|-|
|*before*|<video src="https://github.com/user-attachments/assets/bd96d57b-f17b-44b8-99ba-1c5925a6212f" width=600>|
|*after*|<video src="https://github.com/user-attachments/assets/cf4f83f6-8da6-44f3-8949-d9a589cf0c4c" width=600>|

## Checklist
- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] I included before/after screenshots for any UI changes
- [x] I included a video for animation/interaction changes

> [!NOTE]
> 
> 
> [!NOTE]
> ### Insert native file paths when dropping or pasting non-image files into the composer
> - Adds `addComposerAttachments` in [ChatView.tsx](https://github.com/pingdotgg/t3code/pull/885/files#diff-4b49e092ccd43be0f0de24abe85ba522e09f04288a5d84253b0263e1a389400e) to split dropped/pasted files into images (handled as before) and non-images (inserted as paths).
> - Non-image file paths are resolved via `window.desktopBridge?.getPathForFile(file)` (Electron) or fall back to `file.name`; paths containing spaces are quoted and joined with a space separator.
[...]

PR #2108, opened 2026-04-17
feat(web): attach images dragged from another browser tab

roni-estein:
## What Changed

Dropping an image straight from another browser tab onto the composer now works. Today, dragging an image from e.g. the ChatGPT web UI onto the composer silently does nothing — you have to drop the image into Finder first, then drop the file from Finder into T3 Code.

## Why

The composer's drop handler in [`ChatComposer.tsx`](https://github.com/pingdotgg/t3code/blob/main/apps/web/src/components/chat/ChatComposer.tsx) short-circuits on the very first line of every drag handler:

```ts
if (!event.dataTransfer.types.includes("Files")) return;
```

That gate is correct for Finder drops (which expose a real `File`), but wrong for cross-tab browser drags. When a browser is the drag source, it puts the image into `DataTransfer` as `text/uri-list` + `text/html` (an ``). There is no `File` — the destination has to fetch the bytes itself.

The drop zone never even highlighted, so the user couldn't tell whether the gesture was recognized.

## Why not existing PRs

- **#2091** (Insert non-image file references on drop/paste) — extends non-image file drops to insert native paths; still gated on `dataTransfer.files`.
- **#885** (Insert native file path when dropping or pasting non-image files) — same mechanism, earlier attempt.

Neither touches the "Files" gate, and neither helps a drag that contains no `File` at all. The two surfaces are complementary.

## How it works

Three small pure helpers in `composer-logic.ts`, one async helper in `lib/fetchDroppedImage.ts`, and a ~30-line change to the existing drop handler.

1. **`isComposerAttachmentDrag(types)`** — accepts `Files` (as before) **or** `text/uri-list`. Gating on `text/uri-list` rather than any text type is deliberate: plain-text drags (e.g. selecting and dragging text from another input) don't include `text/uri-list`, so the Lexical editor's native text-drop behavior is preserved.

2. **`extractDraggedImageUrls(dataTransfer)`** — extracts candidate URLs in priority order:
   - `text/uri-list` (the standard; strips `#` comment lines per RFC 2483)
   - `text/html` (`` parsed for the URL)
   - `text/plain` (last-resort fallback, only used if the two above yielded nothing)

   Only `http(s):`, `data:image/*`, and `blob:` URLs pass — everything else (`file:`, `about:`, random text) is dropped so the caller can safely `fetch()` the result.

3. **`deriveImageFilenameFromUrl(url, mime)`** — strips query strings, takes the last path segment, falls back to `dropped-image.<ext>` when the URL has no usable name, and infers the extension from the MIME type when missing.

4. **`fetchDroppedImageAsFile(url)`** — downloads the URL and returns it as a `File` suitable for the existing attachment pipeline. Explicitly handles every failure mode (network/CORS reject, non-2xx status, non-image MIME, body read error) by returning `null` so a single bad URL can't poison a batch.

5. **Drop handler update** — when `dataTransfer.files` is empty but image-shaped URLs are present, fetch them asynchronously and feed the results through the existing `addComposerImages(...)` path. If every fetch fails (typically a CORS block on the source host), a single toast tells the user to fall back to the Finder round-trip that already works.

Order of preservation:

- Finder / app drop → unchanged sync path.
- Text-only drag → still handled by Lexical as before (no `text/uri-list`, handler bails).
- Link / image drag from browser → new path.

## Defensive coding notes

This PR was written with explicit attention to three patterns that have bitten me (and been caught by bots) on recent PRs:

1. **Optional fallthrough in ternary / guarded fallback chain** — the `text/plain` fallback only fires when both `text/uri-list` and `text/html` produced zero URLs. There's a dedicated test (`does not fall back to text/plain when uri-list/html already yielded URLs`) that would fail if the guard regressed to something like `if (uris.length === 0 || ...)`.
2. **Check return values before signaling success** — `fetchDroppedImageAsFile` explicitly checks `response.ok` (fetch resolves on 4xx/5xx; treating a 403 HTML error page as image bytes would otherwise wrap it as a "file"), explicitly checks `blob.type.startsWith("image/")`, and the caller gates the toast + `addComposerImages` on `fetched.length > 0`. A dedicated test covers the 403 case.
3. **DOM cleanup** — not applicable here (no temporary DOM elements created), but noted in case future reviewers expect it.

## Tests

**`apps/web/src/composer-logic.test.ts`** — new coverage for all three pure helpers:

- uri-list single URL
- uri-list with RFC 2483 `#` comment lines
- `` extraction from html, both `"` and `'` quoted
- dedup across uri-list and html
- `data:image/` and `blob:` URL acceptance
- rejection of `file:`, `about:`, and `data:text/plain`
- text/plain last-resort fallback
- no-fall-through guard (see above)
- filename derivation: passthrough, query-string stripping, extension inference, generic fallback, percent-decoding, invalid URL

**`apps/web/src/lib/fetchDroppedImage.test.ts`** — all four failure modes covered:

- happy path (returns File with derived name/MIME)
- fetch rejects (network/CORS)
- response.ok false (403)
- non-image blob type
- blob read throws

Full suite: 897/897 tests pass.

## UI Changes

None when no image is dragged. New behavior only activates when `dataTransfer.files` is empty and image-shaped URLs are present.

Error path (image URL but fetch blocked, typically CORS on the source host):

> **Couldn't attach dragged image.**
> The source blocked a direct download. Save the image to your computer first, then drop the file.

## Validation

- `bun fmt` — clean
- `bun lint` — zero new warnings from the changed files (pre-existing warnings unrelated)
- `bun typecheck` — clean across `@t3tools/scripts`, `@t3tools/web`, `@t3tools/desktop`, root
- `bun run test` — 897/897

## Checklist

- [x] Tests cover both pure logic and the async fetch helper
- [x] No overlap with #2091 or #885 (different surface — they extend the Files-present path, this handles the Files-absent path)

---

> [!NOTE]
> **Medium Risk**
> Adds new drag/drop behavior that triggers client-side `fetch()`es of dropped URLs and alters event gating, which could introduce edge-case UX regressions or unexpected network requests (though failures are handled and surfaced to users).
> 
> **Overview**
> Enables attaching images by dragging them from another browser tab into the chat composer, not just dropping local files.
> 
> Updates composer drag handlers to recognize URL-based drags (`text/uri-list`), extract image URLs from `DataTransfer`, and asynchronously download them into `File`s via a new `fetchDroppedImageAsFile` helper; if all downloads fail (commonly due to CORS), the user gets an error toast.
> 
> Adds tested helpers in `composer-logic.ts` for drag-type gating, URL extraction/deduping, and deriving filenames from URLs, plus unit tests for the new fetch helper’s success and failure modes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68e330f5957f29604b03b1ec26b4361faefe8418. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Support attaching images dragged from another browser tab in the chat composer
> - Extends `onComposerDrop` in [ChatComposer.tsx](https://github.com/pingdotgg/t3code/pull/2108/files#diff-c968a393420901e312c24fcfdef204d8969fa91d9c2df9c52d32b6612bcc8d9a) to handle URL-based drags (e.g. dragging an image from another tab) in addition to file drags.
> - Adds `extractDraggedImageUrls` to parse image URLs from `text/uri-list`, `text/html` ``, and `text/plain` in the drag payload, and `isComposerAttachmentDrag` to gate drag-over highlighting on both `Files` and `text/uri-list` drags.
> - Adds `fetchDroppedImageAsFile` in [lib/fetchDroppedImage.ts](https://github.com/pingdotgg/t3code/pull/2108/files#diff-be7e157a57cc34640d19974794877c9141422e9f5ac9f996016f861aa4a3fcbb) to fetch each extracted URL and convert it to a `File`; non-image, failed, or non-OK responses return `null`.
> - If all URL fetches fail, a toast error is shown suggesting the user save the image locally first.
> - Behavioral Change: drag-over highlighting and copy `dropEffect` now activates for `text/uri-list` drags but not for plain-text drags.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 68e330f.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/chat/ChatComposer.tsx
+++ b/apps/web/src/components/chat/ChatComposer.tsx
@@ -37,8 +37,11 @@ import {
   collapseExpandedComposerCursor,
   detectComposerTrigger,
   expandCollapsedComposerCursor,
+  extractDraggedImageUrls,
+  isComposerAttachmentDrag,
   replaceTextRange,
 } from "../../composer-logic";
+import { fetchDroppedImageAsFile } from "../../lib/fetchDroppedImage";
 import { deriveComposerSendState, readFileAsDataUrl } from "../ChatView.logic";
 import {
   type ComposerImageAttachment,
@@ -1572,21 +1575,21 @@ export const ChatComposer = memo(
     };
 
     const onComposerDragEnter = (event: React.DragEvent<HTMLDivElement>) => {
-      if (!event.dataTransfer.types.includes("Files")) return;
+      if (!isComposerAttachmentDrag(event.dataTransfer.types)) return;
       event.preventDefault();
       dragDepthRef.current += 1;
       setIsDragOverComposer(true);
     };
 
     const onComposerDragOver = (event: React.DragEvent<HTMLDivElement>) => {
-      if (!event.dataTransfer.types.includes("Files")) return;
+      if (!isComposerAttachmentDrag(event.dataTransfer.types)) return;
       event.preventDefault();
       event.dataTransfer.dropEffect = "copy";
       setIsDragOverComposer(true);
     };
 
     const onComposerDragLeave = (event: React.DragEvent<HTMLDivElement>) => {
-      if (!event.dataTransfer.types.includes("Files")) return;
+      if (!isComposerAttachmentDrag(event.dataTransfer.types)) return;
       event.preventDefault();
       const nextTarget = event.relatedTarget;
       if (nextTarget instanceof Node && event.currentTarget.contains(nextTarget)) return;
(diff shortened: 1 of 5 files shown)


juliusmarminge:

PR #2125, opened 2026-04-17
Preserve provider bindings when stopping sessions

### LINKED ISSUE #2121: [Bug]: Claude losing earlier messages / not compacting correctly
### Before submitting

### Area

apps/desktop

### Steps to reproduce

1. Begin thread with Claude 4.7 on 200K Extra High 
2. Send Claude off to do some things 
3. Claude reaches max context window
4. Enter new prompt 
5. Claude confused because there's no context of previous conversation

### Expected behavior

Expected behavior is that when Claude reaches context limit, automatic compaction takes place and Claude continues with compressed context from earlier conversation. 

### Actual behavior

Claude apparently gets a fresh empty context window, and has no context on earlier conversation. 

### Impact

Blocks work completely

### Version or commit

Latest

### Environment

MacOS

### Logs or stack traces

```shell

```

_No response_

### Workaround

_No response_

HaukeSchnau:
Closes #2121

## Summary
- preserve provider bindings when the idle session reaper stops a session
- keep explicit stop-session behavior destructive so manual stop still detaches the provider thread
- add coverage for preserved-binding recovery and the reaper stop payload

## Why
Idle reaping currently uses the same destructive stop path as an explicit manual stop. For the Codex provider, that drops the resume cursor and forces the next turn onto a fresh Codex thread instead of resuming the previous provider thread.

This change keeps the resume binding for idle cleanup only, so a later send can recover the same provider thread while leaving explicit stop semantics unchanged.

## Verification
- `bun fmt`
- `bun x vitest run apps/server/src/provider/Layers/ProviderService.test.ts apps/server/src/provider/Layers/ProviderSessionReaper.test.ts`
- `bun lint`
- `bun typecheck`
- manual end-to-end validation against a real T3 + Codex flow, verifying that a thread still resumes after the idle reaper stops the provider session

## Notes
- callback state for old approvals or user-input requests is still not resumable; this only preserves the provider resume handle across idle cleanup
- `bun lint` passes with existing repo warnings outside this patch

---

> [!NOTE]
> **Medium Risk**
> Changes provider session lifecycle/persistence semantics (stop no longer unbinds), which can affect routing and recovery paths across providers and restarts; coverage is improved but behavior changes are system-wide.
> 
> **Overview**
> Provider session stopping is changed from a **destructive unbind** to a **soft stop**: `ProviderServiceLive.stopSession` now `upsert`s the thread binding with `status: "stopped"` (and clears `activeTurnId`) instead of deleting it, keeping the persisted provider + resume cursor so a later `sendTurn` can recover/resume the same provider thread.
> 
> `ProviderSessionDirectory` removes the `remove(threadId)` API entirely, and tests are updated/added to verify stopped bindings remain persisted and that routing can resume after a stop. `ProviderCommandReactor` session restart logic is tightened to always prefer the thread-bound provider and to only drop `resumeCursor` on model-change restarts, with test reordering/adjustments around restart-failure and provider-switch rejection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b33fd1f801925ca48ada9f0ac08399366d60cd8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Preserve provider bindings when stopping sessions instead of deleting them
> - `ProviderSessionDirectory` removes the `remove` method; stopped sessions are now persisted with `status: 'stopped'` and `activeTurnId: null` rather than deleted from the directory.
> - `ProviderService.stopSession` upserts the binding on stop, so a subsequent `sendTurn` can resume from the correct `resumeCursor` and provider without re-binding from scratch.
> - `ProviderCommandReactor` no longer restarts a session solely due to a provider change; it always uses the thread's bound provider and preserves `resumeCursor` unless a runtime mode or model-change condition requires a restart.
> - Behavioral Change: Sessions that were previously cleaned up on stop are now retained in persistence; any code path that relied on a missing binding after stop will behave differently.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b33fd1f.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
@@ -750,58 +750,6 @@ describe("ProviderCommandReactor", () => {
     });
   });
 
-  it("rejects a first turn when requested provider conflicts with the thread model", async () => {
-    const harness = await createHarness({
-      threadModelSelection: { provider: "codex", model: "gpt-5-codex" },
-    });
-    const now = new Date().toISOString();
-
-    await Effect.runPromise(
-      harness.engine.dispatch({
-        type: "thread.turn.start",
-        commandId: CommandId.make("cmd-turn-start-provider-first"),
-        threadId: ThreadId.make("thread-1"),
-        message: {
-          messageId: asMessageId("user-message-provider-first"),
-          role: "user",
-          text: "hello claude",
-          attachments: [],
-        },
-        modelSelection: {
-          provider: "claudeAgent",
-          model: "claude-opus-4-6",
-        },
-        interactionMode: DEFAULT_PROVIDER_INTERACTION_MODE,
-        runtimeMode: "approval-required",
-        createdAt: now,
-      }),
-    );
-
-    await waitFor(async () => {
-      const readModel = await Effect.runPromise(harness.engine.getReadModel());
-      const thread = readModel.threads.find((entry) => entry.id === ThreadId.make("thread-1"));
-      return (
-        thread?.activities.some((activity) => activity.kind === "provider.turn.start.failed") ??
-        false
-      );
-    });
-
-    expect(harness.startSession).not.toHaveBeenCalled();
-    expect(harness.sendTurn).not.toHaveBeenCalled();
-
-    const readModel = await Effect.runPromise(harness.engine.getReadModel());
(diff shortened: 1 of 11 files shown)


juliusmarminge:
reading this now i don't remember why we ever remove the provider binding 🤔

lastobelus:
ouch, this caused a lot of pain in the last few days. I have to go apologize to several codex threads.

PR #2132, opened 2026-04-17
fix(server): back off failed git status refreshes

gmackie:
## Summary
- treat nonzero `git fetch` exits during status auto-refresh as real failures instead of cache successes
- back off repeated upstream auto-refresh attempts for 10 minutes after a failure so broken remotes/auth do not hammer `git fetch` on every status poll
- cover the regression with a GitCore test that simulates a failed fetch result and verifies the retry only happens after the cooldown expires

## Root Cause
`fetchRemoteForStatus` called `executeGit(..., { allowNonZeroExit: true }).pipe(Effect.asVoid)`, so a failed `git fetch` still populated the cache as a success. That meant the status poll kept retrying on the normal 15-second success interval instead of using the failure cooldown.

## Test Plan
- bun --cwd apps/server vitest run src/git/Layers/GitCore.test.ts --testNamePattern "backs off upstream auto-refresh after a failed status fetch|refreshes upstream before statusDetails so behind count reflects remote updates"
- bun fmt
- bun lint
- bun typecheck

---

> [!NOTE]
> **Medium Risk**
> Changes how upstream refresh failures are detected and cached, which can affect how quickly branch ahead/behind info updates after transient network issues. Low code churn but impacts a core status path used widely in the server.
> 
> **Overview**
> `GitCore` now treats non-zero exit codes from the background `git fetch` used by `statusDetails` as a real failure (emitting a `GitCommandError`) instead of silently ignoring the result.
> 
> The upstream refresh failure cooldown has been increased from `5s` to **`10m`**, so repeated `statusDetails` calls will back off much longer after a failed refresh to avoid retry storms; a new test using `TestClock` validates the backoff behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6861654962fd05086f0d72a4bef2ceb14eedfce7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix upstream git status refresh to back off for 10 minutes after a failed fetch
> - `fetchRemoteForStatus` in [GitCore.ts](https://github.com/pingdotgg/t3code/pull/2132/files#diff-9e2e5027bebfaa9721be501cd8057c4a36400119ad0ad4797a8d6aca7f3c7865) now fails the effect on non-zero exit codes instead of discarding the result, so the cache TTL logic can observe failures.
> - `STATUS_UPSTREAM_REFRESH_FAILURE_COOLDOWN` is increased from 5 seconds to 10 minutes, so a failed fetch won't be retried immediately.
> - Behavioral Change: previously a failed `git fetch` was treated as success; it now propagates as a `GitCommandError`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6861654.</sup>
> 
>


juliusmarminge:
Closing as superseded by #2686, which added VCS remote-refresh backoff.

PR #2136, opened 2026-04-17
Show symlinks and macOS Finder aliases in folder browser

### LINKED ISSUE #2135: [Feature]: Add-project path autocomplete omits symlinked directories and macOS Finder aliases
### Before submitting

### Area

apps/server

### Problem or use case

 Typing a path in the add-project autocomplete (e.g. ~/) hides two kinds of valid directory targets:
  1. Directory symlinks — common on every platform (e.g. ~/Dropbox -> ~/Library/CloudStorage/Dropbox created by Dropbox, stow-managed dotfiles,    
  /opt/homebrew shortcuts).                                                                                                                        
  2. macOS Finder aliases — binary bookmark files created via Make Alias or Option+Cmd drag.                                                       
                                                                                                                                                   
  Root cause in apps/server/src/workspace/Layers/WorkspaceEntries.ts → browse(): it filters dirents on dirent.isDirectory(), but readdir uses lstat
   semantics so symlinks are excluded, and Finder aliases are plain files that no fs call can follow.     

### Proposed solution

  1. In browse(), also consider isSymbolicLink() dirents and keep those whose stat() target is a directory.
  2. On macOS, also consider isFile() dirents, detect Finder aliases by their book… magic bytes, and batch-resolve via a single osascript call     
  using Finder's original item of ….                                                                                                          
  3. Add two optional flags to FilesystemBrowseEntry: isSymlink (icon hint, set for both) and isAlias (navigation hint, set when fullPath is a     
  resolved target).                                                                                                                           
  4. In the command palette, show a distinct icon (e.g. lucide FolderSymlink) for isSymlink entries. For regular dirs and symlinks, keep the       
  existing "append clicked name" navigation so ~/ style paths are preserved. For aliases, jump directly to the resolved fullPath — appending the
  alias name would point at a non-directory and readdir would fail.                                                                                
                                                                   
  Scope stays inside the interactive browse() path. The recursive workspace indexer in the same file is left alone to avoid cycles.  

### Why this matters

 Common developer shortcuts are silently invisible, forcing users to paste resolved paths or fall back to the native picker — which defeats the   
  typed-path autocomplete's purpose. Electron's native folder dialog already handles both cases transparently, so the current gap is an            
  inconsistency, not a design choice.

### Smallest useful scope

 Just the symlink fix: detect symlinks, stat() them, include if the target is a directory. Single-function change, no contract change. Covers the
  Dropbox and dotfiles cases. Finder alias support + contract additions + icon + alias navigation can ship as a follow-up. 

### Alternatives considered

  - Return resolved real paths for symlinks — rejected; the user typed ~/Dropbox, they should pick ~/Dropbox. Alias asymmetry is unavoidable since
  the alias file itself isn't a directory.                                                                                                         
  - Parse the bookmark binary format in JS — rejected; format varies across macOS versions, osascript is simpler and battle-tested.
  - Add a native module for xattrs/bookmarks — rejected; not worth the dependency for one UI affordance.  

### Risks or tradeoffs

  - One extra stat() per symlink candidate — negligible, interactive only.                                                                         
  - One osascript subprocess per browse call with alias candidates on macOS — ~50–200 ms, short timeout so a hung Finder can't block the UI; on    
  failure entries drop silently.                                                                                                               
  - macOS may prompt once for Finder automation permission the first time an alias is resolved.                                                    
  - Symlinks keep the typed path, aliases jump to the resolved path — the distinct icon mitigates the asymmetry.

### Examples or references

  - Repro (symlinks): on a default macOS + Dropbox install, ls -la ~/Dropbox shows … Dropbox -> /Users/<you>/Library/CloudStorage/Dropbox. Type ~/ 
  in the add-project picker — Dropbox is missing.                                                                                                  
  - Repro (aliases): create a Finder alias (Cmd+L) somewhere under ~/. file reports MacOS Alias file, xxd shows book…mark… magic. Type ~/ in the
  picker — the alias is missing.                                                                                                                   

### Contribution

sak0a:
Fixes #2135

## What Changed

The command-palette folder browser silently dropped symlinks and Finder aliases that point to directories, even though Electron's native picker resolves both. This PR brings the type-to-browse path to parity.

- Symbolic links that resolve to a directory (e.g. `~/Dropbox`) are shown and navigable.
- Modern bookmark-format macOS Finder aliases are detected by magic bytes and resolved to their targets via a single batched `osascript` call (30s timeout to cover the first-run TCC "control Finder" prompt).
- An alias that appears mid-path (e.g. `~/my-alias/`) is resolved before `readdir`, so the listing works.
- Both kinds use a distinct `FolderSymlinkIcon`, get unique palette values so two aliases to the same target don't collide, and navigate to the resolved path directly via `browseToPath`.
- Probe concurrency is bounded, the `Effect.promise` bodies have outer fallbacks, and the `osascript` output is NUL-delimited so filenames containing tabs or newlines parse correctly.

## Why

`readdir({ withFileTypes: true })` uses `lstat` semantics, so symlinks to directories don't report `isDirectory()` and Finder aliases are regular files, not symlinks at all. The picker therefore silently hid common locations like `~/Dropbox` and any Finder alias dragged into the home directory. Electron's native folder dialog already resolves both transparently, so this brings the type-to-browse path to parity with the native picker.

## Smaller-scope alternative

If the alias half feels out of scope, the symlink half alone is a much smaller bug fix (**~+65 / −4 across 5–6 files**, cross-platform, no `osascript`, no TCC prompt, no new helpers). Happy to reduce this PR to symlink-only and leave Finder alias support as a follow-up — just let me know.

## UI Changes

BEFORE:
before

AFTER:
after

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core filesystem browsing behavior and adds macOS-specific alias resolution via `osascript` (including timeouts/permissions), which could introduce latency or edge-case failures despite best-effort fallbacks.
> 
> **Overview**
> The filesystem browse API now returns directory candidates that are *symlinks* and (on macOS) Finder *bookmark aliases*, marking them via new `FilesystemBrowseEntry.isSymlink`/`isAlias` flags and excluding symlinks that point to files or are broken.
> 
> On macOS, alias entries are detected by probing magic bytes (`isMacOSBookmarkAlias`) and then resolved in a single batched `osascript` call; browsing also resolves a trailing-path alias to its target before calling `readdir` so listings work when the user types an alias path.
> 
> The command palette browse UI is updated to show a symlink folder icon, generate stable unique item values for aliases (even when multiple aliases resolve to the same target), navigate aliases via a new `browseToPath` flow, and prefetch the resolved target path instead of appending the alias filename.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b15e373452e6f047176518527d75175873e1e380. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Show symlinks and macOS Finder aliases as folders in the workspace folder browser
> - Extends `WorkspaceEntries.browse` to include symlinked directories and macOS Finder aliases alongside regular directories, annotating entries with `isSymlink` and `isAlias` flags in [`FilesystemBrowseEntry`](https://github.com/pingdotgg/t3code/pull/2136/files#diff-80a6cb4180ec3fb0a1c336a148b6ad7313a759f0abf07941a4faf3019d6bf553).
> - Detects macOS Finder aliases by probing the first 4 bytes for the `book` magic and resolving targets via AppleScript (`osascript`) with a 30s timeout.
> - Updates the command palette to display a symlink icon for symlinks/aliases and navigate to the resolved `fullPath` for aliases rather than appending the visible name.
> - Adds `buildBrowseItemValue` to generate stable, unique item values that disambiguate aliases sharing the same resolved target or names containing colons.
> - Risk: alias resolution invokes a subprocess (`osascript`) per browse call on macOS, which may add latency for directories containing many aliases.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b15e373.</sup>
> 
>

### DIFF
--- a/apps/server/src/workspace/Layers/WorkspaceEntries.test.ts
+++ b/apps/server/src/workspace/Layers/WorkspaceEntries.test.ts
@@ -444,5 +444,23 @@ it.layer(TestLayer)("WorkspaceEntriesLive", (it) => {
         expect(result).toBe(false);
       }),
     );
+
+    it.effect("rejects files larger than the probe cap without reading magic", () =>
+      Effect.gen(function* () {
+        const path = yield* Path.Path;
+        const cwd = yield* makeTempDir({ prefix: "t3code-bookmark-magic-" });
+        const largePath = path.join(cwd, "big-fake-alias.bin");
+        // Starts with the bookmark magic but is larger than the probe cap
+        // (64 KB). Without the size prefilter this would return true; with
+        // it the file is skipped before we open it.
+        const header = Buffer.from("book\x00\x00\x00\x00mark\x00\x00\x00\x00", "binary");
+        const padding = Buffer.alloc(128 * 1024, 0);
+        yield* Effect.promise(() =>
+          fsPromises.writeFile(largePath, Buffer.concat([header, padding])),
+        );
+        const result = yield* Effect.promise(() => isMacOSBookmarkAlias(largePath));
+        expect(result).toBe(false);
+      }),
+    );
   });
 });
--- a/apps/server/src/workspace/Layers/WorkspaceEntries.test.ts
+++ b/apps/server/src/workspace/Layers/WorkspaceEntries.test.ts
@@ -8,7 +8,7 @@ import { ServerConfig } from "../../config.ts";
 import { GitCoreLive } from "../../git/Layers/GitCore.ts";
 import { GitCore } from "../../git/Services/GitCore.ts";
 import { WorkspaceEntries } from "../Services/WorkspaceEntries.ts";
-import { WorkspaceEntriesLive } from "./WorkspaceEntries.ts";
+import { WorkspaceEntriesLive, isMacOSBookmarkAlias } from "./WorkspaceEntries.ts";
 import { WorkspacePathsLive } from "./WorkspacePaths.ts";
 
 const TestLayer = Layer.empty.pipe(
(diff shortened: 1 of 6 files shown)


itsklimov:
cool feature

PR #2204, opened 2026-04-19
reduce startup time by ~47% and memory usage by up to 150-300MB

justsomelegs:
## What Changed

Improves desktop/server startup performance by avoiding full orchestration snapshot hydration on the critical startup path.

- Bootstraps the orchestration engine from a lightweight command read model instead of loading every message/activity/checkpoint into memory.
- Serves initial client state through lightweight shell snapshots.
- Moves targeted backend consumers from `getReadModel()` to narrower projection queries where they only need project/thread/session/checkpoint context.
- Reduces saved-environment/client startup fan-out by avoiding duplicate config fetches and coalescing initial saved-environment sync.
- Optimizes Codex discovery cleanup so the probe exits quickly while preserving account/read failure behavior.
- Fixes regressions found during PR audit:
  - Checkpoint revert now preserves persisted `latestTurnId` for the retained turn.
  - Command validation still sees deleted project/thread tombstones.
  - Provider runtime ingestion no longer hydrates full thread detail for every runtime event.

### Benchmark on real persisted local state, isolated Chromium appdata **_(UPDATED)_**:

```
before -> upstream/main (40b3a8003)
runs:   4518, 5412, 4815, 4476, 4567, 4822, 4564, 4870, 4638, 4502 ms
avg:    4718.4 ms
median: 4602.5 ms

after -> branch (this PR, 492027ef7)
runs:   2515, 2437, 2532, 2384, 2424, 2437, 2427, 2517, 2607, 2397 ms
avg:    2467.7 ms
median: 2437 ms

Delta:
average: -2250.7 ms (~47.7% faster)
median:  -2165.5 ms (~47.1% faster)
```

### Memory Reduction after startup

> A side effect of improving the performance is that T3Code now uses 150 -> ~300MB less memory on startup.

<table>
<tr>
 <td>BEFORE (9df3c6402)
 <td>AFTER (This PR)
<tr>
 <td>
Screenshot 2026-04-19 172306
 <td>
Screenshot 2026-04-19 172231
</table>

## Why

Startup was scaling with persisted user state because the server eagerly loaded the full orchestration read model before the window was ready. For users with many projects, threads, messages, activities, or checkpoints, this created unnecessary SQLite work, object allocation, and memory pressure during startup.

This keeps startup focused on the minimum state needed to accept commands and render navigation, while preserving full thread/detail hydration for the paths that actually need it. The approach avoids adding caches or increasing steady-state memory usage.

No UI changes.	

> [!NOTE]
> ### Reduce server startup time by ~47% and memory by 150-300MB by replacing in-memory read model with `ProjectionSnapshotQuery`
> - Removes `getReadModel()` from `OrchestrationEngineShape`; the engine now maintains an internal `commandReadModel` seeded from a lightweight `ProjectionSnapshotQuery.getCommandReadModel()` snapshot instead of a full hydration on startup.
> - Adds `getCommandReadModel()` and `getSnapshotSequence()` to `ProjectionSnapshotQueryShape`, backed by targeted SQL queries that avoid loading the full projection snapshot.
> - Migrates all consumers (`CheckpointReactor`, `ProviderCommandReactor`, `ProviderRuntimeIngestion`, `ProviderSessionReaper`, `ProjectSetupScriptRunner`, `ws.ts`) from full read-model reads to targeted `ProjectionSnapshotQuery` lookups by thread/project ID.
> - Adds migration 29 (`029_ProjectionThreadDetailOrderingIndexes`) which creates two indexes on `projection_thread_activities` and `projection_thread_messages` to support efficient per-thread detail queries.
> - Fixes `thread.reverted` projection to set `latest_turn_id` to the most recent retained turn instead of clearing it, and fixes snapshot latest-turn queries to exclude deleted threads.
> - On the web client, skips `subscribeLifecycle`/`subscribeConfig` for primary connections that have no handlers, and coalesces saved-environment sync calls to avoid redundant startup work.
> - Risk: `OrchestrationEngineShape.getReadModel` is removed entirely; any external consumer not updated in this PR will fail at compile time.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d818a84.</sup>
> 
>
> 

---

> [!NOTE]
> **High Risk**
> High risk because it refactors core orchestration bootstrapping and multiple reactors to stop using the in-memory `getReadModel()` API, changes snapshot SQL/query semantics, and adds a new migration affecting projection indexes and `latest_turn_id` behavior.
> 
> **Overview**
> Cuts server startup memory/time by **removing in-memory read-model hydration from the orchestration engine**: `OrchestrationEngineShape.getReadModel()` is deleted and the engine now bootstraps/validates commands from a lightweight `ProjectionSnapshotQuery.getCommandReadModel()`.
> 
> Adds new projection APIs (`getCommandReadModel()`, `getSnapshotSequence()`), updates snapshot SQL to derive `latestTurn` via `projection_threads.latest_turn_id` (excluding deleted threads for shell snapshots), and dedupes repository identity resolution by workspace root.
> 
> Refactors runtime consumers/tests (e.g., `CheckpointReactor`, `ProviderCommandReactor`, `ProviderRuntimeIngestion`, CLI tests) to use targeted `ProjectionSnapshotQuery` lookups instead of full snapshots, and fixes `thread.reverted` projection to retain the latest retained turn. Includes migration `029_ProjectionThreadDetailOrderingIndexes` adding indexes to speed per-thread detail ordering queries.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d818a8477d7ed7d56f3bcce7bc755929f44521fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


justsomelegs:
further optimisations from the last couple commits brings the performance to this to:

```
before -> upstream/main (40b3a8003)
runs:   4518, 5412, 4815, 4476, 4567, 4822, 4564, 4870, 4638, 4502 ms
avg:    4718.4 ms
median: 4602.5 ms

after -> branch (this PR, 492027ef7)
runs:   2515, 2437, 2532, 2384, 2424, 2437, 2427, 2517, 2607, 2397 ms
avg:    2467.7 ms
median: 2437 ms

Delta:
average: -2250.7 ms (~47.7% faster)
median:  -2165.5 ms (~47.1% faster)
```

apps/web/src/environments/runtime/service.ts
```diff
@@ -100,8 +100,70 @@ let needsProviderInvalidation = false;
...
+        clearTimeout(timeoutId);
+        resolve(config);
+      },
+      () => {
+        clearTimeout(timeoutId);
+        resolve(null);
+      },
+    );
+  });
+}
```
juliusmarminge, line 178: if we don't get the snapshot within `timeoutMs`, there's no providers to use and the app is unusable? shouldn't that reject or trigger some reconnect w/ backoff or?

apps/server/src/orchestration/Layers/ProjectionPipeline.ts
```diff
@@ -739,9 +739,31 @@ const makeOrchestrationProjectionPipeline = Effect.fn("makeOrchestrationProjecti
...
+              turn.checkpointTurnCount === null ||
+              turn.checkpointTurnCount > event.payload.turnCount
+            ) {
+              continue;
+            }
+            if (turn.checkpointTurnCount > latestCheckpointTurnCount) {
+              latestCheckpointTurnCount = turn.checkpointTurnCount;
+              latestTurnId = turn.turnId;
+            }
+          }
```
juliusmarminge, line 762: this fix a different bug or? doens't seem perf related

apps/web/src/environments/runtime/service.ts
justsomelegs: the `timeoutMS` is only trying to avoid an extra `server.getConfig()` if the subscription snapshot shows up immediately. if it doesn’t, we pass null and `refreshSavedEnvironmentMetadata ` falls back to `client.server.getConfig()`, so providers should still load. if that request fails we hit the existing error path. but i can make this more explicit if needs be?

apps/server/src/orchestration/Layers/ProjectionPipeline.ts
justsomelegs: its not directly perf related its more correctness from the perf changes required by the changes from switching snapshots queries to trust the `projection_threads.latest_turn_id` instead of deriving the latest turn by scanning all the turn rows, so `thread.reverted` has to keep the pointer otherwise after a revert the optimized snapshot path can show `latestTurn: null `/ stale state even though retained turns still exist.

juliusmarminge:
Heya I will properly review your PRs when we get back from Miami. Seems like some good stuff just don't have capacity to review big PRs while we're here

justsomelegs:
yeah of course sounds good, hope you enjoy miami

juliusmarminge:
whoops sorry i might have fucked something up. sent an agent to get this up to date and i think it didn't see that you had committed hehe

justsomelegs:
@juliusmarminge do you need me to force push from my local branch? i imagine what your agent did is basically the same anyhows

juliusmarminge:
> @juliusmarminge do you need me to force push from my local branch? i imagine what your agent did is basically the same anyhows

i've not tested it yet so if u have a tested rebase locally feel free to push it up

justsomelegs:
> > @juliusmarminge do you need me to force push from my local branch? i imagine what your agent did is basically the same anyhows
> 
> i've not tested it yet so if u have a tested rebase locally feel free to push it up

yeah ill get it pushed up :)

justsomelegs:
@juliusmarminge should be good now once CI is done

juliusmarminge (approved):

PR #2352, opened 2026-04-26
fix(git): pass --no-ext-diff to patch-emitting git diff calls

avshyz:
Without `--no-ext-diff`, patches passed to `parsePatchFiles` (`@pierre/diffs`) honor the user's `diff.external` config — anyone with difftastic / delta set globally got "Unsupported diff format. Showing raw patch." in the diff panel.

Adds `--no-ext-diff` to the three call sites whose stdout is parsed:
- `CheckpointStore.diffCheckpoints`
- `GitCore.prepareCommitContext.stagedPatch`
- `GitCore.readRangeContext.diffPatch`

## Repro

```
git config --global diff.external difft
```

Make an edit → diff panel falls back to raw patch.


juliusmarminge:
Closing as superseded by #2553, which added --no-ext-diff to the relevant Git commands.

PR #1652, opened 2026-04-01
fix(web): prevent iOS Safari auto-zoom on input focus

bergholmm:
## What Changed

Use `text-[16px] sm:text-[14px]` on the composer editor, its placeholder, and the sidebar thread rename input. This sets font size to 16px on mobile (< 640px) while preserving the original smaller sizes on desktop (≥ 640px).

## Why

iOS Safari automatically zooms into any input/contenteditable element with `font-size < 16px` when focused. This breaks the viewport in mobile web/PWA — the user has to manually pinch-to-zoom back out after every interaction. 16px is the threshold that prevents this behavior.

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [ ] I included before/after screenshots for any UI changes
- [ ] I included a video for animation/interaction changes

---

> [!NOTE]
> **Low Risk**
> Purely presentational Tailwind class tweaks; no logic, data, or security-sensitive code paths are changed.
> 
> **Overview**
> Prevents iOS Safari’s focus auto-zoom by increasing the base font size to 16px on small screens for the `ComposerPromptEditor` contenteditable + placeholder and the sidebar thread rename input (keeping prior sizes via `sm:` overrides).
> 
> Also standardizes tooltip max-width styling in the composer mention/skill chips by switching from `max-w-[30rem]` to `max-w-120`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8aa0401a07339d0061617833ef2bf3ee7f4edf8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Prevent iOS Safari auto-zoom on input focus in composer and sidebar
> iOS Safari auto-zooms inputs with font sizes below 16px. This fix sets text size to `text-base` (16px) by default with a `sm:` breakpoint override restoring the original smaller size on larger screens.
>
> - [ComposerPromptEditor.tsx](https://github.com/pingdotgg/t3code/pull/1652/files#diff-12fcb05b3e147eba955638d8df57440b8cf7f796941af0989c9bfa57b5b4f4a4): editor and placeholder switch from `text-[14px]` to `text-base sm:text-[14px]`; `break-words` replaced with `wrap-break-word`
[...]

PR #1562, opened 2026-03-30
fix: prevent iOS composer focus zoom

gigq:
## What Changed
  Prevents the chat composer from triggering iOS Safari's automatic page zoom when the editor is focused on mobile.
  The fix does two things:
  - Sets the composer editor text size to 16px on coarse-pointer devices, which avoids the common iOS auto-zoom trigger.
  - Adds a small iOS-specific focus guard that temporarily locks the viewport while the composer is focused, then restores the original viewport
    settings on blur.
  This is scoped only to the composer and includes a small unit test for the viewport helper.
  ## Why
  On iPhone, tapping into the composer caused the page to zoom in slightly when the keyboard opened, which left the UI at the wrong zoom level
  afterward and made the mobile experience feel broken.
  This approach keeps the fix narrow:
  - It only affects the composer, not the whole app.
  - It preserves the normal viewport configuration outside of composer focus.
  - It handles iOS/iPadOS specifically instead of applying a global zoom restriction.
  ## UI Changes
  Behavior change only. The composer no longer causes the page to zoom when focused on iPhone.
  ## Checklist
  - [x] This PR is small and focused
  - [x] I explained what changed and why
  - [ ] I included before/after screenshots for any UI changes
  - [x] I included a video for animation/interaction changes
  
https://github.com/user-attachments/assets/90d9e5ce-a134-4734-a575-9dfa8391dbd9

https://github.com/user-attachments/assets/bca309df-ba21-48f3-b28e-4d68aeef35ea

---

> [!NOTE]
> **Medium Risk**
> Adds iOS-specific viewport meta manipulation on focus/blur and adjusts editor font sizing; mistakes could impact mobile zoom/scroll behavior if the viewport content isn’t restored correctly.
> 
> **Overview**
> Prevents iOS Safari from auto-zooming when the chat composer is focused by **raising the composer font size to 16px on coarse-pointer devices** and wiring new focus/blur/touch handlers into `ComposerPromptEditor`.
> 
> Introduces a new `usePreventIosInputZoom` hook that, on iOS/iPadOS, temporarily locks the page viewport (`maximum-scale=1`, `user-scalable=no`) while the editor is focused and restores the original viewport on blur/unmount, with unit tests covering platform detection and viewport content rewriting.
> 
[...]

Issue #1265, opened 2026-03-21
[Bug]: Focusing the chat input zooms the app in causing weird scroll on mobile


chirino:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/web

### Steps to reproduce

Focusing the chat input.  You will see that a zoom in occurs.

### Expected behavior

the sticky header and footer should stay stick and no scrolling should occur.

### Actual behavior

See https://github.com/user-attachments/assets/fe87b3e0-86db-4d3e-aa48-9a2c4baedc67

### Impact

Cosmetic issue

### Version or commit

main @ 9e29c9d72895022322da52d8e961b38702bad9cc 

### Environment

iOS Chrome

### Logs or stack traces

```shell

```

### Screenshots, recordings, or supporting files

[0320.mp4](https://github.com/user-attachments/assets/b8afe957-3b94-4550-a1a2-3dc3e653d926)

JKamsker:
Bug exists for me aswell (samsung galaxy s24 ultra)

alecramos-sudo:
Note: #1652 was merged today and addressed this for the chat composer and the sidebar thread-rename input specifically (`text-base sm:text-[14px]`), but I think this issue should stay open until two remaining gaps are closed:

1. **Other inputs still trigger the same auto-zoom.** Settings text fields, search bars (model picker, branch picker), and any future input added without the `text-base sm:` prefix re-introduce the bug. The Samsung Galaxy S24 Ultra confirmation by @JKamsker above was on a non-composer input — focusing the composer is fixed by #1652, but other touch points are still vulnerable.

2. **The `sm:` breakpoint (640px) is narrower than several touch devices.** iPhone Pro Max in landscape and iPad in any orientation are wider than 640px, so they get the desktop-sized 14px from `sm:text-[14px]` and Safari auto-zooms again on focus.

I just opened #2434 with a defense-in-depth fix: a single `@media (hover: none) and (pointer: coarse)` rule that floors form controls to `max(16px, 1em)` on touch devices only. It complements #1652 (per-component classes still work) and leaves desktop UIs completely unchanged. Diff is purely additive (+21 lines in `apps/web/src/index.css`, no deletions, no other files touched).

Happy to fold this into a different shape if there's a preferred direction — file a comment on #2434 if so.

PR #2434, opened 2026-05-01
fix(web): defense-in-depth against iOS Safari auto-zoom on input focus

alecramos-sudo:
## What Changed

Adds a global `@media (hover: none) and (pointer: coarse)` rule in `apps/web/src/index.css` that floors all `<input>`, `<textarea>`, `<select>`, and `[contenteditable]` elements to `font-size: max(16px, 1em)` on touch devices.

## Why

#1652 (merged) patched two specific components — the chat composer and the sidebar thread-rename input — by switching to `text-base sm:text-[14px]`. That's a great per-component fix, but it leaves two failure modes open and #1265 remains reproducible:

1. **Other inputs are still vulnerable.** Settings text fields, search bars (model picker, branch picker), filter inputs, and any future input added without the `text-base sm:` prefix silently re-introduce the bug. There's no linter rule to catch it. Confirmed in #1265 by another user on a Samsung Galaxy S24 Ultra after #1652 was merged — same auto-zoom on a non-composer input.

2. **The `sm:` breakpoint is 640px, but iOS still auto-zooms past 640px.** iPhone Pro Max in landscape (932px wide) and iPad in any orientation are wider than `sm:` but are still touch devices that trigger Safari's auto-zoom on focus. With `text-base sm:text-[14px]`, those devices get the desktop-sized 14px and the bug returns.

## Approach

Defense-in-depth at the CSS layer. The media query `(hover: none) and (pointer: coarse)` matches any device the browser self-reports as touch-only — regardless of viewport width — and explicitly excludes mouse-pointer environments. So:

- **iOS Safari (iPhone, iPad), iPadOS, Android phones/tablets** → all form controls forced to ≥16px on focus → Safari never auto-zooms
- **Desktop macOS / Windows / Linux with a mouse** → media query never matches → existing `text-xs` / `text-sm` / `text-[14px]` density classes apply unchanged → no visual change

Using `max(16px, 1em)` instead of a flat `16px` preserves any intentionally larger font (e.g. heading-style inputs, `.dialog-title` editable headers) while still raising the floor.

This complements rather than replaces #1652 — the per-component `text-base sm:text-[14px]` pattern in `ComposerPromptEditor.tsx` and `Sidebar.tsx` keeps working on desktop and stays self-documenting at the component level. The new global rule is a safety net that catches everything else.

## Why not the other approaches

- **`maximum-scale=1` in the viewport meta** (or `usePreventIosInputZoom` from #1562) disables pinch-to-zoom entirely while focused, hurting accessibility.
- **A lint rule on JSX `<input>` elements** would only catch missing classes at write time and wouldn't cover the `sm:` breakpoint gap on iPad/landscape phones.
- **Setting `text-base` everywhere** would change desktop UI density (12-14px is intentional for the sidebar / settings rows).

The CSS-scoped media query is the smallest, lowest-risk solution that closes both gaps.

## Files

- `apps/web/src/index.css` — +21 lines, no deletions, no formatting changes to surrounding code

## Verification

- Diff is purely additive (21 insertions, 0 deletions).
- `bunx prettier --check` on the new block: clean.
- `(hover: none) and (pointer: coarse)` is a [well-supported](https://caniuse.com/css-media-interaction) Media Queries Level 4 feature — Safari 9+, all evergreen browsers.

## Closes / refs

- Refs #1265 (closes the remaining gap after #1652)
- Follow-up to #1652

## Checklist

> The change is presentational/preventative — it only takes effect on focus on touch devices, and the visible result is "the page does **not** zoom" — which is hard to capture as a screenshot. The behavior matches what #1652 already demonstrated for the composer; this PR extends the same outcome to every form control, every viewport size on touch devices.

> [!NOTE]
> ### Prevent iOS Safari auto-zoom on input focus for touch devices
> Adds a CSS media rule in [index.css](https://github.com/pingdotgg/t3code/pull/2434/files#diff-70ad657a634ce55957729ff8d1c93f697cc7e3ca2a5cb5fe6b00c90d7d515fcd) targeting touch devices (`hover: none` and `pointer: coarse`) that sets `font-size: max(16px, 1em)` on all `input`, `textarea`, `select`, and `contenteditable` elements. iOS Safari auto-zooms when an input's font size is below 16px; this rule ensures the minimum is always met without overriding larger sizes.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6f857e7.</sup>
> 
>

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -83,6 +83,27 @@
   animation-duration: 0s !important;
 }
 
+/* iOS Safari auto-zooms into any input/textarea/contenteditable whose
+   computed font-size is below 16px. Per-component fixes (text-base
+   sm:text-sm etc.) catch the obvious cases, but it's easy to miss one,
+   and the `sm:` breakpoint (640px) leaves a gap on iPad and iPhone Pro
+   Max in landscape where the desktop-sized 14px kicks in but the
+   browser still treats the device as touch-zoom-capable.
+
+   This rule scopes a "minimum 16px" floor to touch devices only — so
+   desktop UIs keep their 12-14px density, while every form control on
+   any iOS / iPadOS / Android phone or tablet stays at >=16px and the
+   browser never auto-zooms on focus. `max(16px, 1em)` preserves any
+   intentionally larger font (e.g. heading-style inputs). */
+@media (hover: none) and (pointer: coarse) {
+  input,
+  textarea,
+  select,
+  [contenteditable]:not([contenteditable="false"]) {
+    font-size: max(16px, 1em);
+  }
+}
+
 :root {
   color-scheme: light;
   --radius: 0.625rem;


juliusmarminge:
Closing — the underlying auto-zoom issue was fixed on main via #1652; we're not taking the extra defense-in-depth layer for now. Thanks!

PR #2443, opened 2026-05-01
fix(server): detect git repo from subdirectories so diffs and checkpoints work in monorepos

### LINKED ISSUE #2441: [Bug]: Diffs/checkpoints not shown when workspace cwd is a subdirectory of a git repo
### Before submitting

### Area

apps/web

### Steps to reproduce

1. Have a git repository at /path/to/repo (so /path/to/repo/.git exists).
2. Configure a project/open a workspace whose cwd is a sub-package, e.g. /path/to/repo/apps/server.
3. Start a thread, let the agent edit files.
4. Open the diff / checkpoint panel.

### Expected behavior

Per-turn diffs should appear in the diff panel and checkpoints should be captured, the same way they do when the workspace is opened at the repo root. Running git status from that same cwd in a terminal works fine, so t3code should treat it as a git workspace too.

### Actual behavior

The diff panel stays empty and no checkpoints are recorded for the thread. Git-backed features silently no-op, with no UI hint that the workspace was classified as non-git.

Image

Image

### Impact

### Version or commit

main @ 17b43960 and 0.0.22-nightly.20260501.170 (17b43960e863)

### Environment

macOS 15.7.3 (arm64), Node v22.22.2, Bun 1.3.13

### Logs or stack traces

```shell

```

_No response_

### Workaround

_No response_

imabdulazeez:
## What Changed

- `isGitRepository(cwd)` in `apps/server/src/git/Utils.ts` now walks up the directory tree until it finds a `.git` entry or hits the filesystem root, instead of only checking `cwd/.git`.
- No behavioral change at the repo root — same result for cwds that already had `.git` directly inside them.

## Why

Closes #2441.

When a workspace cwd points at a sub-package of a git repo (e.g. `apps/server` in a monorepo), the previous one-level check returned `false`, so:

- `CheckpointReactor` (`apps/server/src/orchestration/Layers/CheckpointReactor.ts:152`) treated the workspace as non-git and skipped checkpoint capture.
- `ProviderRuntimeIngestion` (`apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts:569`) skipped git ingestion.
- `CheckpointStore.isGitRepository` (`apps/server/src/checkpointing/Layers/CheckpointStore.ts:78`) inherited the same blind spot.

The user-visible symptom is an empty diff panel and missing checkpoints for any thread whose cwd isn't the repo root, even though `git status` works fine from that same cwd. Walking parents matches `git`'s own discovery behavior and fixes the gated features without changing any other code paths.

## UI Changes

Not directly, but the visible effect is that the existing diff panel and checkpoint UI now start populating for sub-package workspaces. Tested the change based on the same repo in the issue #2441

### Checkpoint button

image

________________

### Diff view

image

## Checklist

- [x] I included before/after screenshots for any UI changes (Only after images, before images are highlighted in #2441 

> [!NOTE]
> ### Fix `isGitRepository` to detect git repos from subdirectories for monorepo support
> Previously, `isGitRepository` in [Utils.ts](https://github.com/pingdotgg/t3code/pull/2443/files#diff-1ee616952fa622d2c71d75cb27872f6de2db838eec8e9f9b38b0fea34bf11b7a) only returned `true` if the provided directory itself contained a `.git` folder. It now walks up the directory tree using a do-while loop, checking each ancestor until it finds `.git` or reaches the filesystem root. This fixes diffs and checkpoints when the working directory is a subdirectory within a monorepo.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 14169b8.</sup>
> 
>
>


apps/server/src/git/Utils.ts
```diff
@@ -8,10 +8,24 @@ import { Schema } from "effect";
...
+    if (current === root) {
+      return false;
+    }
+    const parent = dirname(current);
+    if (parent === current) {
+      return false;
+    }
+    current = parent;
+  }
 }
```
gameroman: something like this maybe
```suggestion
export function isGitRepository(current: string): boolean {
  while (current !== (current = dirname(current))) {
    if (existsSync(join(current, ".git"))) return true;
  }
  return false;
}
```

imabdulazeez: Thank you so much, this is so much cleaner. Applied in 352a5043. Tested locally against a sub-package workspace: both checkpoints and the diff view work as expected.

apps/server/src/git/Utils.ts
```diff
@@ -8,10 +8,13 @@ import { Schema } from "effect";
...
+import { dirname, join } from "node:path";
 
-export function isGitRepository(cwd: string): boolean {
-  return existsSync(join(cwd, ".git"));
+export function isGitRepository(current: string): boolean {
+  while (current !== (current = dirname(current))) {
+    if (existsSync(join(current, ".git"))) return true;
+  }
+  return false;
 }
```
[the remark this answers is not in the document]
imabdulazeez: Thanks, check change in [594fc8d](https://github.com/pingdotgg/t3code/pull/2443/commits/594fc8dd4cdbf177a07c6e91730eeae3b959497c) addresses this concern.

imabdulazeez:
> #### Approvability
> **Verdict:** Needs human review
> 
> An unresolved review comment identifies a critical bug: the loop structure skips checking the initial directory for .git, which would break detection when the starting path is the repo root. This logic error needs to be addressed before merging.
> 
> You can customize Macroscope's approvability policy. [Learn more](https://docs.macroscope.com/approvability#example-custom-eligibility-rules).

This has been addressed too. thanks

gameroman (approved):

imabdulazeez:
Hi @juliusmarminge -- Friendly reminder to review if possible. This is making working with monorepos a bit difficult and this should fix it :)

PR #2455, opened 2026-05-02
fix(server): handle proposed_plan fallback after mode switch

pedrokpp:
## What Changed

This fixes proposed-plan rendering when a user starts in Plan mode, switches the same thread back to Build mode, and then asks for another plan.

On the server side, proposed-plan fallback extraction now runs when an assistant message completes with `<proposed_plan>...</proposed_plan>` in its final text, even if the provider did not emit a canonical `turn.proposed.completed` event.

When that happens, T3 Code now:
- extracts the markdown inside `<proposed_plan>`
- upserts a real proposed plan for the thread
- strips the `<proposed_plan>` block out of the visible assistant chat message so the raw tagged text is not rendered alongside the plan card
- still records the assistant completion state needed for the normal plan follow-up flow

As a result, the Build-mode fallback now behaves like a normal proposed plan:
- the plan card appears correctly
- the raw tagged assistant text does not remain visible
- if the user later switches back to Plan mode, they can implement or refine that plan as normal

Regression coverage was added around assistant-message completion so the plan is projected correctly and the assistant message is suppressed visually while preserving the turn bookkeeping required by the follow-up UI.

## Why

There was a mode-transition bug in the Codex flow:

1. Start in Plan mode
2. Generate a plan
3. Switch the thread to Build mode
4. Ask the assistant to propose another plan

In that path, the provider could return a `<proposed_plan>` block as ordinary assistant text instead of emitting the canonical proposed-plan event.

Before this fix, the UI could render the raw tagged markdown as a normal assistant message, or render both the raw tagged message and the proposed plan card at the same time. A follow-up regression also showed that suppressing the assistant text too aggressively could break the normal `Implement` flow when switching back to Plan mode.

That behavior was incorrect. Proposed-plan blocks should project as first-class proposed plans, the raw tagged payload should not remain in the chat transcript when it only exists to back the plan card, and the resulting plan should still behave like a normal actionable plan afterward.

This server-side fallback is the smallest fix because it reuses the existing proposed-plan projection pipeline instead of changing web rendering rules or broader mode behavior.

## UI Changes

Before

After

Repro shown in before/after:
1. Start in Plan mode
2. Switch to Build mode
3. Ask for another plan
4. Observe raw `<proposed_plan>` text in chat before this fix
5. Observe only the proposed plan card after this fix
6. Switch back to Plan mode
7. Confirm the plan can still be implemented normally

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches `ProviderRuntimeIngestion` message finalization/proposed-plan projection, which can affect chat transcript rendering and plan upsert behavior across turn/approval boundaries, but is scoped and covered by new tests.
> 
> **Overview**
> Ensures proposed plans still project when providers return a `<proposed_plan>...</proposed_plan>` block as plain assistant text (e.g., after mode switches), by **extracting the block on assistant completion** and upserting it into `thread.proposedPlans`.
> 
> `finalizeAssistantMessage` now strips `<proposed_plan>` blocks from emitted assistant deltas, still completes the assistant message even if the visible text becomes empty, and returns the raw text so callers can derive fallback plan markdown; this fallback projection is triggered on `item.completed`, `turn.completed`, and approval-boundary finalization.
> 
> Adds regression tests covering plan extraction from tagged assistant text on message completion, turn completion, and buffered text flushed at `request.opened`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bbc954c95d5971e7ca240c170ad8de1807c172e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `proposed_plan` block handling in assistant messages after mode switch
> - Adds detection and extraction of `<proposed_plan>` blocks embedded in assistant message text using two new regex constants and a helper in [ProviderRuntimeIngestion.ts](https://github.com/pingdotgg/t3code/pull/2455/files#diff-22a6c9818b956463b848a73a5b3e787b44a144bd003b1cc25fa2b1b3110cdea7).
> - Strips `<proposed_plan>` content from assistant message deltas so plan markup is never rendered as chat text; instead, a `proposedPlan` entity is projected for the active turn.
> - Handles three completion paths: `assistant.message.completed`, `turn.completed`, and `request.opened` (pause boundary) — each now extracts and finalizes any buffered proposed plan.
> - Buffered assistant text containing a proposed plan is held back from flushing as a delta until a completion event arrives.
> - Behavioral Change: `assistant.complete` is now emitted even when visible text is empty, as long as raw content (e.g. a plan block) was present.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bbc954c.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
@@ -738,6 +738,61 @@ describe("ProviderRuntimeIngestion", () => {
     expect(message?.streaming).toBe(false);
   });
 
+  it("projects a proposed plan from assistant message text tagged with proposed_plan", async () => {
+    const harness = await createHarness();
+    const now = new Date().toISOString();
+
+    harness.emit({
+      type: "content.delta",
+      eventId: asEventId("evt-message-delta-plan-1"),
+      provider: ProviderDriverKind.make("codex"),
+      createdAt: now,
+      threadId: asThreadId("thread-1"),
+      turnId: asTurnId("turn-assistant-plan"),
+      itemId: asItemId("item-assistant-plan"),
+      payload: {
+        streamKind: "assistant_text",
+        delta: "<proposed_plan>\n# Streamed fallback plan\n\n- one\n</proposed_plan>",
+      },
+    });
+    harness.emit({
+      type: "item.completed",
+      eventId: asEventId("evt-message-completed-plan"),
+      provider: ProviderDriverKind.make("codex"),
+      createdAt: now,
+      threadId: asThreadId("thread-1"),
+      turnId: asTurnId("turn-assistant-plan"),
+      itemId: asItemId("item-assistant-plan"),
+      payload: {
+        itemType: "assistant_message",
+        status: "completed",
+      },
+    });
+
+    const thread = await waitForThread(harness.engine, (entry) =>
+      entry.proposedPlans.some(
+        (proposedPlan: ProviderRuntimeTestProposedPlan) =>
+          proposedPlan.id === "plan:thread-1:turn:turn-assistant-plan",
+      ),
+    );
+
+    expect(
+      thread.messages.find(
+        (entry: ProviderRuntimeTestMessage) => entry.id === "assistant:item-assistant-plan",
+      ),
+    ).toMatchObject({
+      text: "",
+      streaming: false,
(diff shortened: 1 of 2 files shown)


juliusmarminge:
First-class proposed-plan ingestion and projection now handle this flow, superseding the text-fallback implementation.

PR #2461, opened 2026-05-02
feat(vcs): JJ

juliusmarminge:
## Summary
- Introduce a provider-neutral VCS layer with a new `VcsDriver`, `VcsProcess`, registry, and Git/JJ driver split.
- Add a pluggable source-control provider layer and refactor GitHub PR handling to fit the new provider abstraction.
- Update server and web consumers to use the new VCS/source-control contracts, including branch/status, checkout, and PR-related flows.
- Refresh contracts, shared helpers, tests, and release workflow wiring to match the new architecture.

## Testing
- Not run locally in this turn.
- Expected validation: `bun fmt`
- Expected validation: `bun lint`
- Expected validation: `bun typecheck`

---

> [!NOTE]
> **Medium Risk**
> Adds a new VCS driver and changes the status contract to require a `kind` field, which can affect validation/serialization and behavior across server + web. Also touches workflow routing and status caching/normalization, so regressions could surface in repo detection and Git-only action flows.
> 
> **Overview**
> Adds first-class Jujutsu (`jj`) support by introducing a new `JjVcsDriver` (repo detection, workspace file listing, remotes, ignore filtering via git fallback, bookmarks/workspaces/current change) and registering it in `VcsDriverRegistry`, with *auto-detection preferring JJ over Git* for colocated repos.
> 
> Extends VCS status contracts/results to include a required `kind` (`git`/`jj`/`unknown`) and refactors status composition (`GitWorkflowService` now merges local+remote parts and returns explicit non-Git local status). Updates `VcsStatusBroadcaster` to normalize CWDs via `FileSystem.realPath` and propagates the new status shape through tests.
> 
> Updates the web UI to be VCS-aware: introduces `vcsPresentation` terminology (branches vs bookmarks/workspaces), adjusts labels/errors, and disables Git-only workflow actions when the active repo kind isn’t Git. CI/release workflows now install `jj` via a new composite `setup-jj` action.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a251619fe093100a116694d90f1204755dfdad0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add Jujutsu (jj) VCS driver with auto-detection and VCS-aware UI terminology
> - Introduces a full `JjVcsDriver` in [JjVcsDriver.ts](https://github.com/pingdotgg/t3code/pull/2461/files#diff-c93c1d85120852491ab09df4669ddc424061e8a90835afe162aaf13a2b1ab0c3) supporting repository detection, file listing, remote/bookmark/workspace listing, command execution, and ignore filtering via a `git check-ignore` fallback.
> - Updates `VcsDriverRegistry` to attempt JJ detection before Git when auto-detecting, and registers JJ as a known driver.
> - Adds a `kind` field (`'git'`, `'jj'`, or `'unknown'`) to `VcsStatusLocalShape` and all status result objects, propagated through `GitWorkflowService`, `GitManager`, shared contracts, and stream event handlers.
> - Replaces hardcoded Git terminology in `BranchToolbarBranchSelector`, `ChatHeader`, `DiffPanel`, and `GitActionsControl` with VCS-aware labels via a new `vcsPresentation.ts` module; Git-specific workflow actions are disabled with a tooltip for non-Git repos.
> - Installs the `jj` CLI in CI via a new reusable [setup-jj](https://github.com/pingdotgg/t3code/pull/2461/files#diff-8e7c0e728830dc75686baab02ea0f5632363774db54ab65987f1e28f32a91ca0) composite action used in both quality and release workflows.
> - Risk: `VcsStatusLocalShape` now requires a `kind` field; any producer or consumer of this contract that does not include `kind` will fail schema validation.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a251619.</sup>
> 
>

### DIFF
--- a/.github/actions/setup-jj/action.yml
+++ b/.github/actions/setup-jj/action.yml
@@ -1,0 +1,49 @@
+name: Setup Jujutsu
+description: Install the jj CLI from the official jj-vcs release artifacts.
+
+inputs:
+  version:
+    description: Jujutsu release tag to install.
+    required: false
+    default: v0.40.0
+
+runs:
+  using: composite
+  steps:
+    - name: Install jj
+      shell: bash
+      env:
+        JJ_VERSION: ${{ inputs.version }}
+      run: |
+        set -euo pipefail
+
+        case "${RUNNER_OS}-${RUNNER_ARCH}" in
+          Linux-X64)
+            asset="jj-${JJ_VERSION}-x86_64-unknown-linux-musl.tar.gz"
+            ;;
+          Linux-ARM64)
+            asset="jj-${JJ_VERSION}-aarch64-unknown-linux-musl.tar.gz"
+            ;;
+          macOS-X64)
+            asset="jj-${JJ_VERSION}-x86_64-apple-darwin.tar.gz"
+            ;;
+          macOS-ARM64)
+            asset="jj-${JJ_VERSION}-aarch64-apple-darwin.tar.gz"
+            ;;
+          *)
+            echo "Unsupported runner for jj install: ${RUNNER_OS}-${RUNNER_ARCH}" >&2
+            exit 1
+            ;;
+        esac
+
+        install_dir="${RUNNER_TEMP}/jj-${JJ_VERSION}"
+        mkdir -p "$install_dir"
+
+        curl -fsSL \
+          "https://github.com/jj-vcs/jj/releases/download/${JJ_VERSION}/${asset}" \
+          -o "${RUNNER_TEMP}/${asset}"
+        tar -xzf "${RUNNER_TEMP}/${asset}" -C "$install_dir"
+
+        chmod +x "${install_dir}/jj"
+        echo "$install_dir" >> "$GITHUB_PATH"
+        "${install_dir}/jj" --version
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,9 @@ jobs:
         with:
           node-version-file: package.json
 
+      - name: Setup Jujutsu
+        uses: ./.github/actions/setup-jj
+
       - name: Cache Bun and Turbo
         uses: actions/cache@v5
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -95,6 +95,9 @@ jobs:
         with:
           node-version-file: package.json
 
+      - name: Setup Jujutsu
+        uses: ./.github/actions/setup-jj
+
       - name: Install dependencies
         run: bun install --frozen-lockfile
 
--- a/apps/server/integration/OrchestrationEngineHarness.integration.ts
+++ b/apps/server/integration/OrchestrationEngineHarness.integration.ts
@@ -331,6 +331,7 @@ export const makeOrchestrationIntegrationHarness = (
           refreshLocalStatus: () =>
             Effect.succeed({
               isRepo: true,
+              kind: "git",
               hasPrimaryRemote: false,
               isDefaultRef: true,
               refName: "main",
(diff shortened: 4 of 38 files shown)


juliusmarminge (commented):
Reviewed this against the pluggable VCS/source-control plan. This is a useful direction, especially the driver registration, shared contract tests, and making the web UI aware that not every repository is Git. I do not think we should treat this as final JJ support yet.

What I would change before this becomes the long-term shape:

- Prefer JJ detection over Git for colocated JJ repositories. `VcsDriverRegistry` currently tries Git first, so `jj --git` repos will resolve as Git unless project config pins JJ. That makes auto-detect wrong for one of the most likely JJ setups.
- Do not represent JJ power only through the generic `VcsDriver` plus disabled Git actions. The generic core is fine for status/diff/process basics, but JJ needs a typed extension surface for changes, bookmarks, workspaces, and evolution operations. Otherwise the app will only ever support the lowest common denominator.
- Revisit the capability names. `supportsWorktrees: true` is misleading unless the generic API actually exposes worktree operations. If this means JJ workspaces, that should be a JJ capability/workflow, not a Git-term capability flag.
- Make `JjVcsDriver` fully Effect-native. It currently uses raw `node:fs`/`node:path`/`node:os` temp directory handling. New driver code should follow the service style from this phase: detailed errors, `FileSystem`/`Path`, scoped acquire/release cleanup, and the shared process/scoping discipline.
- The `git check-ignore --no-index` approach may be acceptable as a Git-compatible ignore fallback, but it should be explicitly modeled as that. It should not become the core JJ ignore semantics by accident.

The big architectural point: this PR should introduce `jj` as a first-class VCS with generic core operations plus JJ-specific extension points, not as a driver that mostly proves it can fit into a Git-shaped surface.

juliusmarminge (commented):
Second pass after the follow-up commits. This fixed several concrete issues from the first review: JJ is now detected before Git, temp directory handling moved to Effect `FileSystem`, and the capability flags are less misleading. That is progress.

I still do not think this should merge as the long-term JJ shape yet.

Required changes / architectural blockers:

- `JjVcsDriver` still only implements the generic `VcsDriver` surface (`execute`, `detectRepository`, `isInsideWorkTree`, `listWorkspaceFiles`, `filterIgnoredPaths`). That proves JJ can be detected, but it does not establish the extension-method model we need. A JJ PR should introduce a typed JJ extension service or extension registry shape now, even if only a small set is implemented. Examples: current change, bookmarks, workspace operations, operation log/evolution primitives. Without that, JJ support remains lowest-common-denominator VCS support with Git UI disabled.
- `packages/contracts/src/vcs.ts` now has generic capabilities like `supportsBookmarks` and `supportsAtomicSnapshot`, but no typed API behind them. These flags are not enough for extensibility. Capabilities should either describe operations exposed on the generic core, or point to typed extension services. Otherwise callers can see that a capability exists but have no safe way to use it.
- `ignoreClassifier: "git-compatible-fallback"` is better than silently pretending this is native, but `JjVcsDriver.filterIgnoredPaths` still shells out to Git and creates a synthetic bare Git dir. If this stays, it needs to be very clearly treated as a compatibility fallback and isolated from the eventual JJ-native extension path. We should not build future JJ behavior on top of Git semantics by default.
- The UI path still mostly treats non-Git VCS as “disable Git actions and swap some labels”. That is acceptable for a detection-only slice, but not for claiming JJ as the first extension-capable VCS driver. The PR should either add the extension point now or explicitly narrow the scope/title to detection/status-only JJ support.

The core requirement from the plan is: generic VCS core plus VCS-specific extension methods. This PR improves the generic core, but it still does not add the extension-method half. I would hold the line here so we do not merge a shape that makes future JJ features feel bolted on.

u-k-g:
anything i can do to help move this further?

juliusmarminge:

PR #2462, opened 2026-05-02
feat(scm): Gitlab

juliusmarminge:
## Summary
- Introduce a provider-neutral VCS layer with a new `VcsDriver`, `VcsProcess`, registry, and typed contracts.
- Add source-control provider abstractions plus GitHub and GitLab implementations, and route GitManager/WebSocket flows through the new layers.
- Refactor server text generation and git orchestration code to use the new module boundaries and shared contracts.
- Update web UI state, parsing, and presentation code for provider-neutral refs and pull-request handling.
- Refresh release workflow and Discord notification scripts to match the new build and notification flow.

## Testing
- Not run
- Expected repo checks: `bun fmt`
- Expected repo checks: `bun lint`
- Expected repo checks: `bun typecheck`

---

> [!NOTE]
> **Medium Risk**
> Adds a new GitLab `glab`-backed provider and modifies core git status/push/upstream logic, which can affect PR checkout, branch tracking, and remote pushes across repositories.
> 
> **Overview**
> Adds first-class **GitLab** support via a new `GitLabCli` + `GitLabSourceControlProvider` (list/get/create/checkout merge requests, repo clone URLs, default branch) and wires it into server startup/`SourceControlProviderRegistry`, marking GitLab discovery as implemented.
> 
> Extends VCS status with `aheadOfDefaultCount` (contracts, server, driver, websocket, and UI) so the UI can enable *Create PR* when a branch is synced with upstream but ahead of the default branch.
> 
> Fixes and hardens git operations: pushing no longer blocks on a dirty worktree, `pushCurrentBranch` pushes to the correct remote branch name (avoids creating `origin/<branch>` on the remote), and PR local checkout now restores upstream tracking even when provider head-repo metadata is missing, including fetching the remote-tracking ref when needed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a812dcb0521a684ddb02f1709e40e186b2df3989. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add GitLab as a source control provider via `glab` CLI
> - Implements `GitLabCli` wrapping the `glab` CLI for MR list/view/create/checkout and repository queries, with structured error normalization via `GitLabCliError`.
> - Adds `GitLabSourceControlProvider` adapting GitLab MR data to the provider-neutral `ChangeRequest` shape and registers it in `SourceControlProviderRegistry` and the server's `GitManagerLayerLive`.
> - Introduces `aheadOfDefaultCount` to `VcsStatusRemoteShape`, `GitStatusDetails`, and all status paths so the UI can distinguish ahead-of-default from ahead-of-upstream.
> - Updates `buildMenuItems` and `resolveQuickAction` to use `aheadOfDefaultCount` for enabling Create PR, and removes the uncommitted-changes guard from the Push action.
> - Fixes `pushCurrentBranch` to strip remote shorthands before constructing the `refs/heads/<branch>` push target, preventing spurious branches like `origin/main` on the remote.
> - Risk: Behavioral change — Push is now allowed even with a dirty working tree; previously it was blocked.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a812dcb.</sup>
> 
>


juliusmarminge (commented):
Reviewed this against the source-control provider plan. The provider-neutral `GitLabSourceControlProvider` direction is right, and this is a useful implementation reference. I would tighten a few things before this becomes the long-term provider pattern.

What I would change:

- `GitLabCli` should not use the old `processRunner`/raw `readFile` path. For new source-control providers, we should add/use a shared Effect-native source-control process service: `effect/unstable/process`, scoped child processes, typed exit codes, stream-first stdout/stderr with explicit collection helpers, and detailed provider errors.
- `createMergeRequest` reads the whole body file and passes it as `--description <body>`. That exposes the full generated MR body in process args and risks command length limits. Prefer a file/stdin/API transport for long body content.
- `normalizeHeadSelector` strips `owner:` and keeps only the branch. The existing flow uses owner-qualified selectors to disambiguate forks. GitLab needs to preserve source project/fork semantics where possible, or the provider contract needs a richer selector shape than a provider-neutral string.
- `SourceControlProviderRegistry` is edited directly to add GitLab. That is okay for a first experiment, but it is not the pluggable shape we want. The next provider should be registered through layers/driver registration rather than requiring central registry edits every time.
- The JSON list decoder silently skips invalid entries. For provider reliability, we should at least surface diagnostics/count skipped entries, and ideally fail if the CLI schema has drifted enough that nothing valid remains.

The larger goal is to make GitLab a `SourceControlProvider`, not a second hardcoded CLI path beside GitHub. This PR is close in naming, but the process layer, selector semantics, and provider registration need to match that architecture.

juliusmarminge (commented):
Second pass after the follow-up commits. The process wrapper work and API-based MR creation addressed real issues from the first review. The PR is better, but I still think it violates the long-term provider/VCS separation in an important way.

Required changes / architectural blockers:

- `SourceControlProviderRegistry` still depends directly on `GitVcsDriver` and `@t3tools/shared/git` URL parsing to resolve the provider. That means source-control provider routing is not actually independent from Git. In a JJ repo, Sapling repo, or any future VCS with remotes, provider detection should not have to go through `GitVcsDriver.readConfigValue("remote.origin.url")` or `git remote -v` output. The registry needs generic remote/provider context from the VCS layer, or a separate source-control remote detection service, not a Git-only implementation.
- The provider registry remains a central hardcoded map (`github`, `gitlab`) plus direct construction. That makes every provider PR edit the registry instead of registering a provider. If the goal is pluggable source-control providers, this should move toward a provider registration/layer model now, before GitLab/Azure both copy the same central pattern.
- `SourceControlProviderShape` uses only `cwd` and strings like `headSelector`. That keeps the provider layer dependent on provider-specific string parsing inside each CLI wrapper. We should introduce a small provider context/selector model: detected provider kind, remote name/url, repository identity, source ref, target ref, and optional provider-specific source repository. That would preserve fork/source-project semantics instead of normalizing everything into a single branch-looking string.
- `GitLabCli.normalizeHeadSelector` still strips `owner:` and keeps only the branch. That can be wrong for forks and is exactly the kind of GitHub-shaped simplification that the generic provider layer should avoid. If GitLab needs source project/source branch separately, the provider contract should express that instead of throwing information away.
- There is duplication with the Azure PR: provider registry detection, verbose remote parsing, CLI error normalization patterns, JSON decode wrapping, and web presentation terminology. We should pull the shared parts into this branch/core instead of letting each provider PR establish a slightly different pattern.

This is a useful GitLab adapter, but it still builds on Git-only provider routing. The long-term architecture needs source-control providers layered above generic VCS remote context, not above `GitVcsDriver`. I would not merge this until that boundary is corrected.

PR #2506, opened 2026-05-04
Preserve thread visited state across partial syncs

hogeheer499-commits:
## What changed
- Preserve `threadLastVisitedAtById` entries when `syncThreads` receives a partial snapshot that does not include a thread.
- Keep the existing pruning behavior for `threadChangedFilesExpandedById`.
- Add a regression test for the partial-snapshot visited-state case.

## Why this should exist
This is a small, focused frontend state bug fix.

A production crash report showed React error 185 with `markThreadVisited` in the stack. On current `main`, `ChatView` calls `markThreadVisited(threadKey, completedAt)`, but `syncThreads` could delete that completed-turn timestamp when a partial/recovery snapshot omitted the active thread. After that, the same completed turn could be marked visited again instead of remaining idempotent.

The regression test proves the exact state transition:
1. mark a thread visited at `completedAt`
2. run `syncThreads([])` as a partial snapshot
3. call `markThreadVisited(thread, completedAt)` again

With the old pruning behavior, the timestamp is lost. With this change, it is preserved and the second mark returns the same state object.

## Scope
- No UI or layout change.
- Does not change WebSocket reconnect behavior.
- Does not change server/provider turn lifecycle behavior.
- Separate from the server projection issue tracked in later work.

## Validation
- `npx --yes bun@1.3.11 run --filter @t3tools/web test src/uiStateStore.test.ts`
- `npx --yes bun@1.3.11 fmt`
- `npx --yes bun@1.3.11 lint` - exit 0, existing warnings only
- `npx --yes bun@1.3.11 typecheck`

---

> [!NOTE]
> **Low Risk**
> Small, localized change to client UI state sync with regression tests; no auth, server, or persistence behavior changes.
> 
> **Overview**
> **`syncThreads` no longer drops `threadLastVisitedAtById` for threads missing from a snapshot.** Visit timestamps are kept across partial or recovery syncs; only **new** snapshot threads still get `seedVisitedAt` when they have no entry. **`threadChangedFilesExpandedById` is still pruned** to threads in the snapshot.
> 
> Tests cover omitted threads keeping visit state and **`markThreadVisited` staying idempotent** after `syncThreads([])`—addressing a path where wiping the timestamp could re-trigger visits and contribute to React error 185 in `ChatView`.
> 
[...]

PR #2513, opened 2026-05-05
Keep WebSocket reconnects alive

hogeheer499-commits:
## What changed
- Change the WebSocket RPC retry schedule from a fixed 7-retry budget to continuous retries.
- Keep the existing exponential backoff and 64s maximum delay.
- Update reconnect state/label logic so an unbounded retry loop shows `Attempt N` instead of `Attempt N/max`.
- Add tests that prove reconnect scheduling continues past the old retry budget.

## Why this should exist
This is a small reliability fix.

T3 Code is often used through LAN, tailnet, SSH-forwarded, or other remote endpoints. A temporary server restart or private-network drop should not leave an already connected client permanently stopped after a fixed retry budget. With the previous `Schedule.recurs(7)` policy, the client could exhaust reconnect attempts and require focus/online/manual retry before trying again.

Continuous retry is still bounded by the existing backoff cap, so repeated failures do not spin aggressively.

Important scope clarification from later incident debugging: this PR is not presented as the root-cause fix for one local `:3777` reproducer where a custom hotpatch proxy healthcheck was restarting the proxy process. That local loop was fixed outside this repository. This PR only addresses the upstream web client's behavior after a real WebSocket/server/network drop has already happened.

## Scope
- WebSocket reconnect behavior only.
- No API, persistence, or provider-runtime changes.
- No layout change. The reconnect surface only changes the attempt label for unbounded retries.
- The timing behavior is covered by deterministic reconnect scheduling tests rather than a visual recording; there is no animation or transition change.
- Separate from the visited-state fix in #2506 and the server projection fix in #2517.

## Validation
- `bun run --filter @t3tools/web test src/rpc/wsTransport.test.ts` - 25/25 passed
- `bun run --filter @t3tools/web test src/rpc/wsConnectionState.test.ts src/components/WebSocketConnectionSurface.logic.test.ts` - 10/10 passed
- `bun run --filter @t3tools/web test` - 96 files, 995 tests passed
- `bun run fmt` - passed
- `bun run lint` - 0 errors, existing warnings only
- `bun run typecheck` - 12/12 packages passed, existing Effect language-service messages only

### DIFF
--- a/apps/web/src/components/WebSocketConnectionSurface.tsx
+++ b/apps/web/src/components/WebSocketConnectionSurface.tsx
@@ -8,7 +8,6 @@ import {
   type WsConnectionStatus,
   type WsConnectionUiState,
   useWsConnectionStatus,
-  WS_RECONNECT_MAX_ATTEMPTS,
 } from "../rpc/wsConnectionState";
 import { stackedThreadToast, toastManager } from "./ui/toast";
 import { getPrimaryEnvironmentConnection } from "../environments/runtime";
@@ -42,11 +41,10 @@ function describeOfflineToast(): string {
 }
 
 function formatReconnectAttemptLabel(status: WsConnectionStatus): string {
-  const reconnectAttempt = Math.max(
-    1,
-    Math.min(status.reconnectAttemptCount, WS_RECONNECT_MAX_ATTEMPTS),
-  );
-  return `Attempt ${reconnectAttempt}/${status.reconnectMaxAttempts}`;
+  const reconnectAttempt = Math.max(1, status.reconnectAttemptCount);
+  return status.reconnectMaxAttempts === null
+    ? `Attempt ${reconnectAttempt}`
+    : `Attempt ${Math.min(reconnectAttempt, status.reconnectMaxAttempts)}/${status.reconnectMaxAttempts}`;
 }
 
 function describeExhaustedToast(): string {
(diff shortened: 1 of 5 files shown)


hogeheer499-commits:
Adding context for the reconnect behavior change flagged by Macroscope.

The intent is not to create an aggressive retry loop. The retry cadence still uses the existing exponential backoff and remains capped at `WS_RECONNECT_MAX_DELAY_MS` (`64_000` ms), so a persistently unavailable backend settles at roughly one reconnect attempt per minute for a single browser session. The transport also closes the previous session before replacing it on explicit reconnect, and the tests cover stale lifecycle events so old sockets do not keep mutating connection state.

Why I think this is worth human review rather than an automatic approval-only change: T3 Code is often used through LAN, tailnet, SSH-forwarded, or desktop-managed remote environments. In those setups, a server restart or short network interruption should recover without requiring the user to focus the tab or manually retry after the previous 8-attempt budget has been exhausted.

Verification relevant to this risk:

- `wsTransport.test.ts` now proves a 9th WebSocket attempt is created after the old retry budget would have stopped.
- `bun run --filter @t3tools/web test` passed: 96 files, 995 tests.
- `bun run typecheck` passed: 12/12 packages.

If maintainers prefer a bounded version, I can adjust this to a larger finite retry window instead of `Schedule.forever`; I chose the capped continuous loop because it matches the expected behavior of a desktop-style app connection.

juliusmarminge:
The new client-runtime connection supervisor already performs continuous reconnects with bounded backoff and wakeup handling, superseding this transport-level retry patch.

PR #2517, opened 2026-05-05
Keep active turns running until session settles

hogeheer499-commits:
## What changed
- Keep a projected turn in `running` while the thread session still reports that same turn as active.
- Complete or error the latest running turn when the session settles instead of completing it from every non-streaming assistant message.
- Reopen a stale completed turn if a later running session update reports that turn as active again.
- Add a regression test that covers both live projection and replay from the event log.

## Why this should exist
This is a focused server projection correctness fix.

A local T3 Code 0.0.22 desktop state showed this inconsistent projection:

```text
projection_thread_sessions.status = running
projection_thread_sessions.active_turn_id = 019df64c-8cf7-79f0-8b0f-7591e0dbbbbc
projection_turns.state for active turn = completed
projection_turns.completed_at = 2026-05-05T04:01:59.270Z
provider_session_runtime.status = running
provider_session_runtime.runtime_payload.activeTurnId = 019df650-a59b-7fa3-870e-b47ea9d1bf61
later assistant messages for the completed turn = yes
```

The event stream showed a running session, then a non-streaming assistant message for the active turn, then more provider activity. The old projection logic treated that non-streaming assistant message as final turn completion. That is too early for providers that emit non-streaming assistant segments during longer work.

The regression test mirrors that sequence. It then deletes the projection rows and bootstraps from the event log again, so the test covers both online projection and replay.

## Scope
- Server projection only.
- No UI change.
- No provider runtime behavior change.
- No reconnect behavior change.
- Separate from the visited-state fix in #2506 and the reconnect retry fix in #2513.
- The diff is larger than the logic change because the regression test needs full orchestration event envelopes.

## Validation
- `npx --yes bun@1.3.11 run --filter t3 test src/orchestration/Layers/ProjectionPipeline.test.ts` - 19 passed
- `npx --yes bun@1.3.11 run --filter t3 test src/orchestration/Layers/ProviderRuntimeIngestion.test.ts` - 38 passed
- `npx --yes bun@1.3.11 run --filter t3 test` - 121 files passed, 961 tests passed, 1 file / 4 tests skipped
- `npx --yes bun@1.3.11 run fmt:check` - passed
- `npx --yes bun@1.3.11 run lint` - 0 errors, existing warnings only
- `npx --yes bun@1.3.11 run --filter t3 typecheck` - exit 0, existing Effect language-service messages only

### DIFF
--- a/apps/server/src/orchestration/Layers/ProjectionPipeline.test.ts
+++ b/apps/server/src/orchestration/Layers/ProjectionPipeline.test.ts
@@ -2165,6 +2165,282 @@ it.effect("restores pending turn-start metadata across projection pipeline resta
   ),
 );
 
+it.layer(Layer.fresh(makeProjectionPipelinePrefixedTestLayer("t3-active-turn-segment-")))(
+  "OrchestrationProjectionPipeline",
+  (it) => {
+    it.effect("keeps the active running turn open across non-streaming assistant segments", () =>
+      Effect.gen(function* () {
+        const projectionPipeline = yield* OrchestrationProjectionPipeline;
+        const eventStore = yield* OrchestrationEventStore;
+        const sql = yield* SqlClient.SqlClient;
+        const appendAndProject = (event: Parameters<typeof eventStore.append>[0]) =>
+          eventStore
+            .append(event)
+            .pipe(Effect.flatMap((savedEvent) => projectionPipeline.projectEvent(savedEvent)));
+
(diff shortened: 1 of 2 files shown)


juliusmarminge:
The current projection pipeline now keeps active turns running until the session settles and includes regression coverage for interim assistant messages. This fix has effectively landed through newer work.

Issue #2379, opened 2026-04-28
[Bug]: Opencode fails to render code blocks correctly


olekgolus11:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

1. Start the app with Opencode as provider, select GPT-5.4 Low Plan
2. Send this prompt: 
```
Can you write me a small block with python bubble sort and an explanation under? Write only this:
1. code block
2. explanation
```

### Expected behavior

The explanation should not be repeated, and should not be included in the code block. The issue happens each time I ask for a code.

### Actual behavior

You will see a code block with the explanation in this code block, then a repeated part of the code, and then the explanation again under the code block

<img width="825" height="799" alt="Image" src="https://github.com/user-attachments/assets/a695a100-2565-4627-9160-75cf606dd050" />

---

<img width="800" height="379" alt="Image" src="https://github.com/user-attachments/assets/7c97b945-6da6-4a54-ad12-ec5843ab9963" />

### Impact

Cosmetic issue

### Version or commit

0.0.21 (188df6da074b)

### Environment

macOS 26.4.1

### Logs or stack traces

```shell

```

### Screenshots, recordings, or supporting files

[Screenshot 2026-04-28 at 12.18.32.png](https://github.com/user-attachments/assets/ae787f9c-d63d-44f6-b872-93b96ff6dc48)
[Screenshot 2026-04-28 at 12.18.20.png](https://github.com/user-attachments/assets/b916f3b7-e285-4f56-8e9b-4b4ae2d26492)

PR #2526, opened 2026-05-05
Fix OpenCode raw text delta assembly

### LINKED ISSUE #2380: [Bug]: OpenCode provider can duplicate the assistant response / replay a large suffix chunk at completion
### Before submitting

### Area

apps/server

## Summary

When using the OpenCode provider in T3 Code, the final assistant response can appear corrupted near completion, often  looking like the final response was duplicated or that a large suffix chunk was replayed.

After deeper investigation, this appears to be caused by T3 Code’s OpenCode delta assembly logic rather than the desktop shell itself.

## Steps to reproduce
   1. Start the desktop app (also reproduced from source on current `main`).
   2. Configure and use the **OpenCode** provider.
   3. Select an OpenCode-backed model (for me this reproduced with **Kimi K2.6**).
   4. Send a prompt that produces a medium-length multi-paragraph response.
   5. Watch the assistant response as it streams and especially when it finishes.
   6. Repeat a few times if needed; it does not happen on every very short response, but it is frequent enough with ordinary
 longer outputs.

   Notes:
   - This does not require a special prompt.
   - Any response long enough to stream across many chunks seems sufficient.
   - I reproduced this in the official desktop app and on current upstream source.

 ## Expected behavior

   The assistant response should stream once and finalize once.

   The final message should exactly match the intended output with:
   - no dropped characters during streaming
   - no duplicated suffix chunk
   - no large corrective replay at completion

## Actual behavior

   The assistant response streams mostly normally, but near the end the final message can become corrupted and look like:
   - the final response was emitted twice, or
   - a previously generated suffix chunk was appended again at completion

   There can also be a brief visual flash right as the response settles, but that seems secondary. The main issue is that
 the final assistant text itself can be wrong.

## Impact

Major degradation / frequent failure.

This makes OpenCode-backed models frustrating to use because the final answer can be visibly corrupted, especially on normal multi-paragraph responses.

## Version or commit
`upstream/main @ dbebc38`
Also reproduced in the official desktop app build.

## Environment

   - OS: Windows 11
   - App: Desktop app
   - Provider: OpenCode
   - Model: Kimi K2.6
   - Bun: 1.3.10 locally
   - Also reproduced in the official desktop app build, not only in a local fork

## Logs or stack traces

   Likely relevant area:
   - `apps/server/src/provider/Layers/OpenCodeAdapter.ts`

   After inspecting failing provider logs, the likely problem is:
   - raw OpenCode `message.part.delta` chunks are already valid
   - T3 Code tries to dedupe overlap between consecutive delta chunks
   - that overlap logic strips legitimate leading characters from normal chunks
   - later `message.part.updated` arrives with the correct full text
   - T3 Code then emits a large corrective suffix, which looks like duplicated final output

   Examples observed from real logs:
   - `"Bonus"` became `"onus"`
   - `" usage"` became `"usage"`
   - `"different"` became `"ifferent"`
   - `"and"` became `"d"`

   This strongly suggests the problem is in OpenCode delta assembly, especially around:
   - `appendOpenCodeAssistantTextDelta(...)`
   - the `message.part.delta` handling branch

## Workaround

   I verified a local fix in my fork by changing the OpenCode delta assembly logic to:
   - append raw `message.part.delta` chunks exactly as received
   - keep `message.part.updated` snapshot merging only as a reconciliation step

   That local change appears to stop the persistent final-response corruption for me.

   So there is a practical code-level workaround in the OpenCode adapter path, but no user-facing workaround in the official
 app besides avoiding OpenCode for affected models.

## Additional note
This appears provider-specific. I have mainly observed it with OpenCode-backed models and not as a general T3 Code
 rendering issue.

Kiri110K:
## What changed

- Treat OpenCode `message.part.delta` text chunks as raw incremental deltas and append them exactly as received.
- Keep `message.part.updated` snapshot handling as the reconciliation path.
- Add regression coverage for coincidental suffix/prefix overlap, where the old dedupe logic could strip valid leading characters from a delta.

## Why this should exist

Fixes #2380.
Likely related to #2379.

OpenCode text deltas are already valid incremental chunks. The previous overlap heuristic tried to dedupe suffix/prefix overlap between the already assembled assistant text and the next delta, but normal text can overlap coincidentally.

For example, if the current text ends in `B` and the next raw delta is `Bonus`, stripping the overlap turns `Bonus` into `onus`. A later `message.part.updated` snapshot then reconciles the missing text by emitting a larger corrective suffix, which can make the assistant response appear duplicated or corrupted near completion.

This change makes raw deltas authoritative and leaves full part snapshots to reconcile state without mutating normal streamed chunks.

## Scope

- OpenCode provider assistant text assembly only.
- No UI change.
- No protocol or persistence change.
- No behavior change for other providers.

## Validation

- `bun fmt`
- `bun lint` - 0 errors, existing warnings only
- `bun typecheck` - passed
- `bun run --cwd apps/server test src/provider/Layers/OpenCodeAdapter.test.ts` - 14 passed

## UI Changes

Not applicable. This is a server-side provider streaming fix.

## Checklist

- [ ] I included before/after screenshots for any UI changes (N/A)

---

> [!NOTE]
> **Medium Risk**
> Changes OpenCode assistant text streaming assembly; incorrect handling could impact user-visible output ordering/duplication, but scope is limited to a single provider and covered by new regression tests.
> 
> **Overview**
> Fixes OpenCode assistant text assembly by **treating `message.part.delta` text as raw incremental chunks** and appending them verbatim, removing the prior suffix/prefix overlap de-duplication heuristic.
> 
> Keeps `message.part.updated` handling as the reconciliation mechanism, and updates/adds tests to cover overlapping-update scenarios and the regression where coincidental prefix overlap (e.g. `B` + `Bonus`) previously caused valid characters to be stripped.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 584ab1c2d389270847151398a3a443e004acbc80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `appendOpenCodeAssistantTextDelta` to append raw deltas without stripping prefix overlaps
> - Removes the `suffixPrefixOverlap` helper and the trimming logic from `appendOpenCodeAssistantTextDelta` in [OpenCodeAdapter.ts](https://github.com/pingdotgg/t3code/pull/2526/files#diff-a4267388b15e5043ae1899beded972cbee9e52533ca3d9d65d438a91a9f7fbe7); deltas are now appended as-is.
> - Behavioral Change: `nextText` is now a direct concatenation of `previousText` and the full delta, and callers receive the original delta string rather than an overlap-trimmed version.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 584ab1c.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/OpenCodeAdapter.test.ts
+++ b/apps/server/src/provider/Layers/OpenCodeAdapter.test.ts
@@ -565,17 +565,98 @@ it.layer(OpenCodeAdapterTestLayer)("OpenCodeAdapterLive", (it) => {
     }),
   );
 
-  it.effect("deduplicates overlapping assistant text deltas after part updates", () =>
+  it.effect("appends raw assistant text deltas and reconciles part update snapshots", () =>
     Effect.sync(() => {
       const firstUpdate = mergeOpenCodeAssistantText(undefined, "Hello");
       const overlapDelta = appendOpenCodeAssistantTextDelta(firstUpdate.latestText, "lo world");
-      const secondUpdate = mergeOpenCodeAssistantText(overlapDelta.nextText, "Hello world!");
+      const secondUpdate = mergeOpenCodeAssistantText(overlapDelta.nextText, "Hellolo world");
 
       assert.deepEqual(
         [firstUpdate.deltaToEmit, overlapDelta.deltaToEmit, secondUpdate.deltaToEmit],
-        ["Hello", " world", "!"],
+        ["Hello", "lo world", ""],
       );
-      assert.equal(secondUpdate.latestText, "Hello world!");
+      assert.equal(secondUpdate.latestText, "Hellolo world");
+    }),
+  );
+
+  it.effect("does not strip coincidental prefix overlap from OpenCode part deltas", () =>
+    Effect.gen(function* () {
+      const adapter = yield* OpenCodeAdapter;
+      const threadId = asThreadId("thread-opencode-raw-delta");
+      const part = {
+        id: "part-raw-delta",
+        sessionID: "http://127.0.0.1:9999/session",
+        messageID: "msg-raw-delta",
+        type: "text",
+        text: "A B",
+        time: { start: 1 },
+      };
+      runtimeMock.state.subscribedEvents = [
+        {
+          type: "message.updated",
+          properties: {
+            sessionID: "http://127.0.0.1:9999/session",
+            info: {
+              id: "msg-raw-delta",
+              role: "assistant",
+            },
+          },
+        },
+        {
+          type: "message.part.updated",
+          properties: {
+            sessionID: "http://127.0.0.1:9999/session",
+            part,
+            time: 1,
+          },
+        },
+        {
+          type: "message.part.delta",
+          properties: {
+            sessionID: "http://127.0.0.1:9999/session",
+            messageID: "msg-raw-delta",
+            partID: "part-raw-delta",
+            field: "text",
+            delta: "Bonus",
+          },
+        },
+        {
+          type: "message.part.updated",
+          properties: {
+            sessionID: "http://127.0.0.1:9999/session",
+            part: {
+              ...part,
+              text: "A BBonus",
+              time: { start: 1, end: 2 },
+            },
+            time: 2,
+          },
+        },
+      ];
+      const eventsFiber = yield* adapter.streamEvents.pipe(
+        Stream.filter((event) => event.threadId === threadId),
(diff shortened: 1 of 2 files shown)


juliusmarminge (approved):
i have no idea why we did this in the first place 🫣

rafaeltab:
image

PR #2586, opened 2026-05-08
Optimize VCS diff loading to be up to 98% faster

justsomelegs:
## What Changed

- Moved checkpoint filesystem operations behind a generic VCS checkpoint capability instead of scripting Git directly in `CheckpointStore`.
- Reworked `VcsProcess.run` and `processRunner` onto a shared `capturedProcess` primitive for one-shot collected process execution.
- Optimized checkpoint diff generation by:
  - removing redundant checkpoint preflight checks on the diff path
  - resolving the VCS driver once per checkpoint operation
  - using explicit commit peeling (`^{commit}`) for checkpoint diff refs
- Added a narrower full-thread diff lookup path and supporting tests.

## Why

Opening diffs and switching turns was spending too much time in process execution overhead and checkpoint diff orchestration.

This branch improves that by:
- reducing per-command process overhead in the hot path
- moving checkpoint behavior behind the VCS abstraction instead of hand-assembling Git commands in the store
- reducing Git revision-resolution cost for checkpoint refs in repos with large checkpoint ref namespaces

## Benchmark Summary

Synthetic workload used for all numbers below:
- `24` changed files
- `690,422` patch bytes
- `26,137` patch lines

Benchmarks were run sequentially to avoid contention between benchmark loops.

### Mean Latency

| Metric | `upstream/main` | Current branch |
| --- | ---: | ---: |
| `checkpointStore.diffCheckpoints` | `1179.81ms` | `72.84ms` |
| `checkpointDiffQuery.getTurnDiff` | `1649.32ms` | `52.05ms` |
| `parseTurnDiffFilesFromUnifiedDiff` | `12.42ms` | `11.97ms` |

### Tail Latency (`p99`)

| Metric | `upstream/main` | Current branch |
| --- | ---: | ---: |
| `checkpointStore.diffCheckpoints` | `2262.90ms` | `355.04ms` |
| `checkpointDiffQuery.getTurnDiff` | `2937.21ms` | `191.04ms` |
| `parseTurnDiffFilesFromUnifiedDiff` | `16.43ms` | `19.63ms` |

### Key Measurements

Compared directly with `upstream/main`:
- `checkpointStore.diffCheckpoints`: `1179.81ms` -> `72.84ms` mean (`16.20x` faster, `93.8%` lower)
- `checkpointDiffQuery.getTurnDiff`: `1649.32ms` -> `52.05ms` mean (`31.69x` faster, `96.8%` lower)
- `checkpointStore.diffCheckpoints`: `2262.90ms` -> `355.04ms` p99
- `checkpointDiffQuery.getTurnDiff`: `2937.21ms` -> `191.04ms` p99

**_Before_**

https://github.com/user-attachments/assets/15613af8-917b-4627-bb97-3a0d11ad48dc

_**AFTER**_

https://github.com/user-attachments/assets/ae1f9435-e267-406a-a6f8-35847c14f21e

## Checklist

> [!NOTE]
> ### Speed up VCS diff loading by removing preflight checkpoint checks and adding a dedicated full-thread diff path
> - `getTurnDiff` in [`CheckpointDiffQuery.ts`](https://github.com/pingdotgg/t3code/pull/2586/files#diff-a7f6357814d1957a562de3ea25f26bd4dc2f18e4ccf790082ae04e2bc484da51) no longer calls `hasCheckpointRef` before diffing; `getFullThreadDiff` now uses a dedicated `getFullThreadDiffContext` SQL query and diffs directly from turn 0 to the stored checkpoint ref instead of delegating to `getTurnDiff`.
> - Replaces the imperative Promise-based `runProcess` function with a new Effect-based [`ProcessRunner`](https://github.com/pingdotgg/t3code/pull/2586/files#diff-5a47a42313dddcd1d898b2fd498aab076e20a9447c1fc581a0eaa174b917068c) service that exposes typed errors (`ProcessSpawnError`, `ProcessTimeoutError`, `ProcessOutputLimitError`, etc.), structured timeout handling, and a separate `spawnCwd` option.
> - Adds checkpoint operations (`captureCheckpoint`, `restoreCheckpoint`, `diffCheckpoints`, `hasCheckpointRef`, `deleteCheckpointRefs`) to `GitVcsDriver`; `CheckpointStore` now delegates all checkpoint logic to the active VCS driver instead of executing git commands directly.
> - Renames the `truncateOutputAtMaxBytes` option to `appendTruncationMarker` across all VCS process interfaces and updates git subprocesses to spawn from the host CWD and use `-C <cwd>` to target the repository.
> - Risk: Breaking interface changes — `VcsProcess.withProcess` is removed, `truncateOutputAtMaxBytes` is gone, and all callers of `runProcess` must migrate to the `ProcessRunner` service.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 427bc9c.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Touches core VCS/process execution and checkpoint diff logic; regressions could break diffing, checkpoint restore/capture, or host environment detection across platforms.
> 
> **Overview**
> **Speeds up checkpoint diff loading** by removing redundant checkpoint-ref preflight checks, adding a dedicated `getFullThreadDiff` path, and using a narrower snapshot query (`getFullThreadDiffContext`) to avoid loading full thread checkpoint context.
> 
> **Moves checkpoint filesystem/Git work behind the VCS abstraction**: `CheckpointStore` now resolves the active driver once per operation and delegates to a new optional `driver.checkpoints` capability implemented for Git (capture/restore/diff/delete), including commit-peeling (`^{commit}`) and safer diff flags.
> 
> **Replaces ad-hoc process execution with an Effect `ProcessRunner` service** and rewires consumers (`VcsProcess`, server environment label resolution, repository identity resolver, terminal manager, VCS discovery) to use the injected runner; this also renames `truncateOutputAtMaxBytes` to `appendTruncationMarker` and updates tests/layers to provide required Node/Process services.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 427bc9ccc6862d76db100035f4daa01aea911534. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/server/src/capturedProcess.ts
juliusmarminge, line 1: any reason this is using `node:child_process` and async isntead of effect's module? I've been slowly ripping out the processRunner module in favor for the effect ones that are auto scoped and shit. everything using `async` is considered legacy in my head 🫠

justsomelegs: main reason is when benchmarked the old `effect/unstable/process` streamed setup was a lot slower / expensive for the exact thing diff loading does, which is basically just "run one command, collect stdout, return".

for `vcsprocess.run` comparison the difference is:

`effect/unstable/process` - `~285.85ms`

new version - `~70.38ms`


I could go back to effect stream if you'd like or just wrap them in a more Effect native style?

IMax153: If you can provide a small, isolated reproduction of the difference in performance as two separate runnable scripts, we can take a look.

justsomelegs: yeah, gist is here: https://gist.github.com/justsomelegs/7ca4f328fcf71d7333c2633ef5d5ca32

my guess is it’s mostly the overhead of the current stream-first process path vs a direct collected-output path.

for the same workload, the old effect path was:

- `238.93ms` vs `66.75ms` on a `690,422` byte payload
- `355.36ms` vs `98.30ms` on `2MB`
- `471.42ms` vs `178.62ms` on `20MB`

this specific `spawn -> stream -> collect` path probably does a lot more setup/teardown then a one-shot collected command I imagine.

justsomelegs: there might just be a easier/more performant way to do it in effect that im not aware of as well

IMax153: With Effect there is always going to be some overhead given the additional work we do in the runtime to enable all the features Effect provides.

That said, @tim-smart did a small test comparing direct consumption of a `node:child_process` stream vs. doing the same with Effect's `ChildProcess` and `Stream` module with a ~6K line diff and the results were nearly identical ([see the gist](https://gist.github.com/tim-smart/49aee31e9fe80719cdb17e58e4be156b)), so I think there may be something else going on here.

I also think there are a few opportunities for improving performance with the Effect version you linked. For example, when you consume the `Stream` via `runFold`, you end up consuming every element of the stream even after `truncated` has been set to `true`.

@tim-smart demonstrates a possible method you could use to collect a partial stream using a `Sink`:
```ts
import { NodeRuntime, NodeServices } from "@effect/platform-node"
import { Console, Effect, Sink, Stream } from "effect"
import { ChildProcess } from "effect/unstable/process"
import { ChildProcessSpawner } from "effect/unstable/process/ChildProcessSpawner"

Effect.gen(function*() {
  const spawner = yield* ChildProcessSpawner
  const handle = yield* spawner.spawn(ChildProcess.make`git diff`)
  yield* handle.stdout.pipe(
    Stream.decodeText,
    Stream.run(collectorSink)
  )
}).pipe(
  Console.withTime("effect"),
  Effect.scoped,
  Effect.provide(NodeServices.layer),
  NodeRuntime.runMain
)

const collectorSink = Sink.reduceWhileArray<{
  text: string
  bytes: number
}, string>(
  () => ({
    text: "",
    bytes: 0
  }),
  (s) => s.bytes < 1000,
  (s, chunks) => {
    for (let i = 0; i < chunks.length; i++) {
      const chunk = chunks[i]
      s.text += chunk
      s.bytes += chunk.length
      if (s.bytes >= 1000) break
    }
    return s
  }
)
```

justsomelegs: you're right, it turned out to be something else.

i dug into it a bit more and the main issue seems to have been on the windows cleanup path around our old `VcsProcess` usage, not `decodeText` / `mkString` by themselves.

two things mattered:

- on the old path we had our own extra finalizer calling `child.kill()` even though the Effect spawner already has scoped cleanup
- on windows that meant we could end up paying for another expensive `taskkill` path unnecessarily

also i tried explicitly waiting for `handle.exitCode` but by itself it wasn't the whole fix but did help.

i've now got `VcsProcess.run` back on the Effect process path in a leaner shape, and on my repro it brought the Effect version back down from the old `~250-300ms` range to `~60ms` mean, which is in the same performance range as the node version.

apps/server/src/processRunner.ts
```diff
@@ -93,51 +95,6 @@ function normalizeBufferError(
...
-    };
-  }
-  return {
-    next: `${target}${chunk.subarray(0, remaining).toString()}`,
-    nextBytes: currentBytes + remaining,
-    truncated: true,
-  };
-}
-
 export async function runProcess(
```
juliusmarminge: can we get rid of this functoin and use the `runCapturedProcess` in all consumers? or atleast remove the async wrapper and make it effectful e2e?

justsomelegs: yeah sorry got busy was in the middle of doing that so that we use the effect e2e process instead of the one i came up with, ill get it done now and remove the async wrappers too.

apps/server/src/capturedProcess.ts
juliusmarminge: @justsomelegs any reason you didn't take Tim's Sink collector?

justsomelegs: @juliusmarminge that exact sink stops consuming once the limit is hit. the one being used just stops storing after `maxOutputBytes` to keep memory bounded but keeps draining so the git command can still exit cleanly without backpressure / timeout weirdness. 

the current one tended to perform better once output went over `maxOutputBytes`, but i can switch to it if you'd prefer as under the limit they were on par if not the same most of the time

juliusmarminge: aren't we loading unnecessary stuff into memory that way?

should we not accumulate the Sink until the limit, then `yield* child.exitCode` to await completion? 

I havne't dug into exactly how it works under the hook but collecting a large output just to discard it seems unnecessary?

justsomelegs: the only thing we retain should be the capped command output.

we could terminate the child process when the limit is reached, but on windows that makes perf worse because cleanup goes through `taskkill`, which is pretty expensive compared to just draining and letting git exit normally.

i could split this to a windows only behaviour path if you wanted though?

juliusmarminge: makes sense. fine as is then i think

apps/server/src/checkpointing/Layers/CheckpointDiffQuery.ts
```diff
@@ -100,70 +124,130 @@ const make = Effect.gen(function* () {
...
+
+      return turnDiff;
+    },
+  );
 
-      if (!fromExists) {
+  const getFullThreadDiff: CheckpointDiffQueryShape["getFullThreadDiff"] = (
+    input: OrchestrationGetFullThreadDiffInput,
+  ) =>
+    Effect.gen(function* () {
```
juliusmarminge: `Effect.fn`

apps/server/src/orchestration/Services/ProjectionSnapshotQuery.ts
```diff
@@ -125,6 +135,15 @@ export interface ProjectionSnapshotQueryShape {
     threadId: ThreadId,
   ) => Effect.Effect<Option.Option<ProjectionThreadCheckpointContext>, ProjectionRepositoryError>;
 
+  /**
+   * Read only the narrow context needed to compute a full-thread diff from
+   * checkpoint 0 to a specific turn count.
+   */
+  readonly getFullThreadDiffContext?: (
```
juliusmarminge: why optional?

apps/server/src/project/Layers/RepositoryIdentityResolver.ts
```diff
@@ -83,50 +89,61 @@ interface RepositoryIdentityResolverOptions {
...
-  try {
-    const topLevelResult = await runProcess("git", ["-C", cwd, "rev-parse", "--show-toplevel"], {
-      allowNonZeroExit: true,
-    });
-    if (topLevelResult.code !== 0) {
-      return cacheKey;
-    }
+const resolveRepositoryIdentityCacheKey = Effect.fn("resolveRepositoryIdentityCacheKey")(function* (
+  cwd: string,
+  runManagedProcess: ManagedProcessRunner,
```
juliusmarminge: nit: DI the ChildProcessSpawner (Effect.provideService) and `runProcess` like normal

apps/server/src/terminal/Layers/Manager.test.ts
```diff
@@ -31,6 +33,20 @@ import {
...
 
+function joinPath(...parts: ReadonlyArray<string>): string {
+  return parts
+    .map((part, index) => {
+      if (index === 0) return part.replace(/[\\/]+$/g, "");
+      return part.replace(/^[\\/]+|[\\/]+$/g, "");
+    })
+    .filter((part) => part.length > 0)
+    .join("/");
+}
```
juliusmarminge: what happened to `path.join`. ?

apps/server/src/vcs/VcsProcess.test.ts
juliusmarminge, line 1: @effect/vitest

apps/server/src/vcs/VcsProcess.ts
```diff
@@ -72,164 +57,99 @@ function commandLabel(command: string, args: ReadonlyArray<string>): string {
...
+      ...(input.spawnCwd !== undefined ? { spawnCwd: input.spawnCwd } : {}),
+      ...(input.stdin !== undefined ? { stdin: input.stdin } : {}),
+      ...(input.env !== undefined ? { env: input.env } : {}),
+      timeoutMs: input.timeoutMs ?? DEFAULT_TIMEOUT_MS,
+      maxOutputBytes: input.maxOutputBytes ?? DEFAULT_MAX_OUTPUT_BYTES,
+      outputMode: "truncate",
+      truncatedMarker: input.appendTruncationMarker ? OUTPUT_TRUNCATED_MARKER : "",
+      timeoutBehavior: "error",
+    }).pipe(
+      Effect.mapError((cause) => {
```
juliusmarminge: effect pattern i sometimes use is adding static methods like

```
class VcsProcessSpawnError extends ... {
  static fromProcessSpawnError = () => ...
}
```

sometimes ends up with cleaner business logic. Ref: https://github.com/pingdotgg/t3code/blob/main/packages/effect-acp/src/errors.ts#L50

apps/server/src/processRunner.test.ts
juliusmarminge, line 1: @effect/vitest and provide a test implemntation of the `ChildProcessSpawner` instead of writing scripts to FS?

PR #2729, opened 2026-05-16
feat(server): replace Claude Agent SDK runtime with a direct stream-json CLI driver (subscription billing)

akarabach:
I know it will never be merged, but, it's alternative solution not to create emdash/cmax for claude. We can spawn claude process!

Ive not reviewed the code (working on internal browser integration and dont want to spend time on this pr)

Screenshot 2026-05-16 at 15 55 15

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -23,7 +23,6 @@
     "test": "vitest run"
   },
   "dependencies": {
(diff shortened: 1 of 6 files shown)


akarabach:
since we still can use t3codes for claude - closing it for now

PR #2981, opened 2026-06-06
fix(web): remember Composer Fast mode across new chats

aravhawk:
## What Changed

Rebased the stale Fast-mode fix onto current `main` (the previous branch was ~2,000 commits behind and conflicted).

- Default Cursor Composer `fastMode` to **Normal** when the user has not explicitly chosen Fast. The injected `false` is treated as an explicit selection so it still goes out through today's `buildExplicitProviderOptionSelectionsFromDescriptors` path. That is required because Cursor ACP reports `fastMode: true` as the provider default, and omitting the option leaves Fast on.
- Apply the same implicit Normal default in trait controls so the Fast/Normal badge matches what will be sent.
- Preserve sticky trait options when a model-only picker update omits `options`, matching the contract `setModelSelection` already uses for drafts. Switching Composer 2 → 2.5 (or opening a new chat) no longer drops the last Fast/Normal choice.

## Why

The original PR still described a real bug: it is **not** implemented on current `main`. A mechanical rebase was not viable because composer state now uses explicit-only dispatch, `planModeEnabled`, and prompt-injection state.

This replay keeps Fast opt-in and remembers the last manual choice across new chats and model changes, without sending other descriptor defaults as overrides.

## UI Changes

N/A — behavior-only change. New Composer chats start in Normal unless the user previously chose Fast, and switching models no longer resets Fast/Normal to the provider default.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * New chats now default to Normal mode when no Fast mode preference is selected.
  * Sending a new message can switch away from a previous Fast mode session when appropriate.
  * Model selections retain existing provider options when switching models without specifying new options.

* **Bug Fixes**
  * Explicit Fast or Normal mode choices are now preserved correctly.
  * Provider capability displays now accurately show Normal when it is the effective default.

### DIFF
--- a/apps/web/src/components/chat/composerProviderState.test.tsx
+++ b/apps/web/src/components/chat/composerProviderState.test.tsx
@@ -5,12 +5,14 @@ import {
   type ProviderOptionSelection,
   type ServerProviderModel,
 } from "@t3tools/contracts";
+import { getProviderOptionDescriptors } from "@t3tools/shared/model";
 import {
   getComposerProviderState,
   renderProviderTraitsMenuContent,
   renderProviderTraitsPicker,
   withImplicitFastModeDefault,
 } from "./composerProviderState";
+import { getProviderModelCapabilities } from "../../providerModels";
 
 // Everything in composerProviderState is now data-driven by the model's
 // optionDescriptors, so these tests use a single synthetic provider/model and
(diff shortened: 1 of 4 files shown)


aravhawk:
Rebased onto current main and resolved the conflicts. Same Fast-mode sticky fix, adapted to the explicit-dispatch path. Ready if you still want this.

PR #2987, opened 2026-06-07
fix(vcs): reflect external git changes in status via .git watcher

denniskasper:
## What Changed

Added a filesystem watcher on the checkout's git directory in `VcsStatusBroadcaster`. When `HEAD`/`packed-refs` change on disk, it pushes a fresh local status (debounced so one operation = one refresh). It resolves the worktree gitdir when `.git` is a file, and is ref-counted alongside the existing remote-status poller (torn down with the last subscriber). Includes an end-to-end test.

## Why

VCS status was only re-read on window `focus`/`visibilitychange` and after app-initiated git actions. So a branch switch or checkout done in a **terminal** (or another git client) wasn't reflected in the UI until the window regained focus. This is the root cause behind the stale-branch-list symptom in #2926 — the watcher resyncs automatically, so no manual refresh button is needed.

Related to #2926.

## UI Changes

No markup change — behavioral. Observable effect: after a terminal `git switch`, the branch toolbar/list updates on its own instead of only after refocusing the window.

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] I included before/after screenshots for any UI changes (N/A — no visual change)
- [x] I included a video for animation/interaction changes (to attach)

> [!NOTE]
> ### Refresh VCS local status when `.git` directory changes on disk
> - Adds a filesystem watcher on the repo's `.git` directory (or linked worktree gitdir) in [`VcsStatusBroadcaster`](https://github.com/pingdotgg/t3code/pull/2987/files#diff-1cb99aea20f9603d7de458a7b72df1f1f1dc7c4a0a3424be90cbbbca9ab6f952) that triggers `refreshLocalStatus` when `HEAD` or `packed-refs` change.
> - Debounces watch events by 150ms to coalesce rapid bursts (e.g. during a git checkout).
> - The watch fiber is started alongside the existing remote polling fiber on first subscriber and interrupted when subscriber count reaches zero.
> - Handles linked worktrees by reading the `.git` file to resolve the actual gitdir path.
> - Behavioral Change: `ActiveRemotePoller` now tracks two fibers; both are interrupted on teardown instead of only the remote polling fiber.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9cc0017.</sup>
> 
>

[vcs-status-live-update-external-git-switch.webm](https://github.com/user-attachments/assets/8d5e1e91-a713-4aad-8aad-bfce1a5ea468)

---

> [!NOTE]
> **Medium Risk**
[...]

Issue #2926, opened 2026-06-03
[Feature]: Add a refresh button to the git branch list


ji233-Sun:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I am describing a concrete problem or use case, not just a vague idea.

### Area

apps/web

### Problem or use case

The git branch list does not reliably reflect the current state of the repository. Sometimes after deleting a branch (e.g. via the terminal, another git client, or as part of cleaning up merged branches), the deleted branch still shows up in the branch list. The only way I have found to get an accurate list again is to fully reload, which is disruptive when I'm in the middle of working on a thread.

### Proposed solution

Add a manual **Refresh** button (or a refresh icon) next to / inside the git branch list that re-queries the repository's branches on demand and updates the displayed list. This gives a quick, non-disruptive way to resync the UI with the real repo state without reloading everything.

### Alternatives considered

- Reloading the app/window — works but is heavy-handed and loses my current context.
- Automatic polling/refresh on focus — would also solve it, but a manual refresh button is simpler, predictable, and a reasonable first step.

### Risks or tradeoffs

Minimal. The main consideration is debouncing rapid clicks and showing a brief loading state while branches are re-fetched, so the button can't trigger overlapping git calls.

### Examples or references

Most git GUI tools (e.g. VS Code's Source Control view, GitKraken, Fork) expose a refresh control for exactly this stale-state scenario.

juliusmarminge:
Closing as fixed by merged PR #2978. Opening the branch selector now refreshes branch refs, resolving the stale branch list described here.

PR #2988, opened 2026-06-07
fix(web): keep the branch toolbar in sync with the real checkout

denniskasper:
## What Changed

Two related branch-toolbar reconciliation fixes:

1. **Refresh VCS status (not just the ref list)** when the branch combobox opens and after branch actions. Previously, opening the picker reloaded the refs but left the trigger label driven by stale status — so the list could update while the label didn't.
2. When git authoritatively reports a **detached HEAD** (on no branch — e.g. the thread's branch was deleted out from under the checkout), stop falling back to the now-dangling thread branch and show "Select ref" instead of a branch name that isn't actually checked out.

Related to #2926.

## Why

The toolbar could display a branch you weren't on: after deleting/switching the underlying branch externally, the label kept showing the old/deleted branch (e.g. `throwaway/repro-error`) even though HEAD was detached. These changes make the toolbar reflect the real checkout.

## UI Changes

**No layout/styling change — behavioral.** In detached HEAD (e.g. the thread's branch was deleted), the trigger label previously showed the stale/deleted branch name; it now shows **"Select ref"**. The combobox also re-syncs its status when opened, so the label can't lag behind the ref list.

## Checklist

- [x] I included before/after screenshots for any UI changes (N/A — no visual/layout change; behavioral, described above)
- [x] I included a video for animation/interaction changes (N/A — covered by the watcher demo in #2987)

### DIFF
--- a/apps/web/src/components/BranchToolbar.logic.test.ts
+++ b/apps/web/src/components/BranchToolbar.logic.test.ts
@@ -82,6 +82,29 @@ describe("resolveBranchToolbarValue", () => {
       }),
     ).toBe("main");
   });
+
+  it("falls back to the thread branch while local status is still loading", () => {
+    expect(
+      resolveBranchToolbarValue({
+        envMode: "local",
+        activeWorktreePath: null,
+        activeThreadBranch: "feature/base",
+        currentGitBranch: null,
+      }),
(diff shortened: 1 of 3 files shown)


juliusmarminge:

PR #2605, opened 2026-05-08
[codex] add configurable automatic git fetch interval

juliusmarminge:
## Summary

Adds a Source Control setting for configuring the background Git fetch interval. The default remains 30 seconds, and setting the interval to `0` disables automatic fetches while preserving explicit Git actions.

## What changed

- Replaced the unshipped `automaticGitFetches` boolean with `automaticGitFetchInterval`, stored as an Effect duration setting with millisecond JSON encoding.
- Added a `Fetch interval` number field to the Source Control settings page, with `0 seconds` shown as disabled.
- Routed VCS status subscriptions through the configured interval so passive remote refresh polling can be slowed down or disabled before it reaches `git fetch`.
- Set `SSH_ASKPASS_REQUIRE=never` on background upstream status fetches so OpenSSH askpass dialogs are not triggered by passive refreshes.
- Kept explicit user-triggered Git operations such as pull, push, and PR branch fetches unchanged.
- Updated settings merge/persistence handling so duration settings are merged atomically and persisted in encoded form.
- Included the setting in restore-defaults behavior and added status broadcaster coverage for the disabled path.

## Why

Background `git fetch` can trigger SSH agent, passphrase, Touch ID, security-key, or Windows GCM prompts for users with secure Git setups. Making the interval configurable lets users reduce that background activity, and `0` lets those prompts happen only on explicit Git actions. The OpenSSH `SSH_ASKPASS_REQUIRE=never` guard also prevents passive refreshes from opening askpass UI when a fetch still runs.

Closes #356
Closes #1190
Closes #1418
Closes #1467
Closes #1965
Closes #2285

Refs #1921

## Validation

- `bun fmt`
- `bun lint` (passes with existing warnings)
- `bun typecheck`
- `bun run test --filter=t3 -- src/vcs/VcsStatusBroadcaster.test.ts`
- `bun run test --filter=t3 -- src/vcs/GitVcsDriverCore.test.ts`
- `bun run test --filter=t3 -- src/serverSettings.test.ts`
- `bun run test --filter=@t3tools/web -- src/rpc/wsTransport.test.ts`
- `bun run test:browser src/components/settings/SettingsPanels.browser.tsx`
- `bun run test:browser src/components/ChatView.browser.tsx`
- `bun run test:browser src/components/KeybindingsToast.browser.tsx`
- `bun run test:browser`

---

> [!NOTE]
> **Medium Risk**
[...]

Issue #1418, opened 2026-03-25
[Bug]: App continue to trigger and close Github gcm auth request while in focus on Windows


IRelaxxx:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

1. Start the app without github logged in
2. Wait with the app in focus and a thread open

### Expected behavior

The window opening to ask the user to auth with github stay open.

### Actual behavior

The window opening to ask the user to auth with github opens and closes again after a few seconds. This happens again and again.

### Impact

Minor bug or occasional failure

### Environment

Windows 11 with app version 0.0.14

### Logs or stack traces

```shell
server.log contains this message a bunch of times.

timestamp=2026-03-25T20:08:41.983Z level=Info fiber=#25622 cause="Error: Git command failed in GitCore.fetchUpstreamRefForStatus: git fetch --quiet --no-tags origin +refs/heads/main:refs/remotes/origin/main (C:\Dev\ai\ai-chat) - git fetch --quiet --no-tags origin +refs/heads/main:refs/remotes/origin/main timed out.
    at onNone (file:///C:/Users/Alexander/AppData/Local/Programs/t3-code-desktop/resources/app.asar/apps/server/dist/index.mjs:17802:31)
    at file:///C:/Users/Alexander/AppData/Local/Programs/t3-code-desktop/resources/app.asar/node_modules/effect/dist/Option.js:206:22
    at Module.<anonymous> (file:///C:/Users/Alexander/AppData/Local/Programs/t3-code-desktop/resources/app.asar/node_modules/effect/dist/Function.js:94:18)
    at Effect.fn (file:///C:/Users/Alexander/AppData/Local/Programs/t3-code-desktop/resources/app.asar/apps/server/dist/index.mjs:20218:9)
    at Effect.fn (definition) (file:///C:/Users/Alexander/AppData/Local/Programs/t3-code-desktop/resources/app.asar/apps/server/dist/index.mjs:19735:29)"
```

### Workaround

Run `git fetch --quiet --no-tags origin +refs/heads/main:refs/remotes/origin/main` manually and auth or catch the window as it pops up.

PR #3133, opened 2026-06-17
[codex] Make background VCS fetch non-interactive

Andrew-Forster:
## What Changed

Background upstream status fetches now run with non-interactive Git auth env vars, not just the SSH askpass guard. The status-refresh fetch already had `SSH_ASKPASS_REQUIRE=never`; this adds `GCM_INTERACTIVE=never`, `GIT_TERMINAL_PROMPT=0`, and empty `GIT_ASKPASS`/`SSH_ASKPASS` for that same passive fetch path.

## Why

This keeps passive VCS polling from opening credential UI. The earlier fetch interval work helps users slow down or disable background fetches, but Windows HTTPS remotes can still go through Git Credential Manager. If GCM doesn't have a cached credential, it can open a GitHub auth window from a background status refresh. That window can then get killed by the short fetch timeout and reopen on the next poll. For background status checks, failing quietly is better than stealing focus. Explicit user actions like pull, push, and manual fetch still use the normal environment.

Refs #1418 and #2605.

## User Workaround

Users can usually fix their local GitHub HTTPS setup by authenticating Git directly, for example with `gh auth setup-git` if they already use GitHub CLI, or by logging into Git Credential Manager outside T3 Code. That still leaves an app-side problem though: passive polling should not be the thing that starts an interactive auth flow. If credentials are missing, the background refresh should fail quietly and let an explicit Git action handle authentication.

## Tradeoff

When credentials are not already available, passive remote status may fail quietly and ahead/behind counts can stay stale until the user authenticates through an explicit Git action or disables/re-enables background fetching. That is intentional here: background polling should not steal focus or spawn short-lived auth windows.

## UI Changes

None.

## Checklist

## Validation

- `pnpm --filter t3 exec vp test run src/vcs/GitVcsDriverCore.test.ts --reporter=dot`
- `pnpm exec vp check`
- `pnpm exec vp run typecheck`

> [!NOTE]
> ### Make background VCS fetch non-interactive by setting additional environment variables
> Adds `GCM_INTERACTIVE=never`, `GIT_ASKPASS=""`, `GIT_TERMINAL_PROMPT=0`, and `SSH_ASKPASS=""` to [`GitVcsDriverCore.STATUS_UPSTREAM_REFRESH_ENV`](https://github.com/pingdotgg/t3code/pull/3133/files#diff-1c3e2b9763c6526006485633fb677c83dc54cc1ea5563a1e51aab707c35854d8) alongside the existing `SSH_ASKPASS_REQUIRE=never`. The test in [`GitVcsDriverCore.test.ts`](https://github.com/pingdotgg/t3code/pull/3133/files#diff-ac30488287dac9aa8d4cb819c508d77d2e47267d4aae35819e6fa24b86312c26) is updated to verify all five variables are set to non-interactive values and to restore them after execution.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c405d30.</sup>
> 
>

---

> [!NOTE]
> **Low Risk**
> Scoped to cached background status fetches only; user-initiated pull/push/fetch paths are unchanged.
> 
> **Overview**
> Background upstream status refreshes now run Git with a fuller **non-interactive** env overlay, not only `SSH_ASKPASS_REQUIRE=never`. The passive `fetchRemoteForStatus` path sets `GCM_INTERACTIVE=never`, `GIT_TERMINAL_PROMPT=0`, and clears `GIT_ASKPASS` / `SSH_ASKPASS` so polling does not open GCM or terminal credential UI (e.g. on Windows HTTPS).
> 
> Explicit user Git actions still use the normal process environment. If credentials are missing, background refresh is expected to fail quietly and ahead/behind may stay stale until the user authenticates via an explicit action.
> 
> The integration test was broadened to assert all of these vars reach the child Git/SSH process when `statusDetails` triggers upstream refresh.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c405d3042bbcc90bd22398a16317adc15a3c7298. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/server/src/vcs/GitVcsDriverCore.test.ts
+++ b/apps/server/src/vcs/GitVcsDriverCore.test.ts
@@ -216,7 +216,7 @@ it.layer(TestLayer)("GitVcsDriver core integration", (it) => {
       }),
     );
 
-    it.effect("disables SSH askpass for background upstream status fetches", () =>
+    it.effect("makes background upstream status fetches non-interactive", () =>
       Effect.gen(function* () {
         const cwd = yield* makeTmpDir();
         const tempDir = yield* makeTmpDir("git-vcs-driver-ssh-env-");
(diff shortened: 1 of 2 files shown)


Andrew-Forster:
This is the window I was seeing from passive status polling. Because the background fetch times out, the auth window opens for a few seconds, closes, then reopens on the next poll.

Screenshot 2026-06-17 124549

PR #3297, opened 2026-06-20
[codex] add browser, terminal, and thread-management MCP tools

JoshuaRileyDev:
## Summary
This PR expands the MCP surface in T3 Code so agents can directly use the built-in browser and terminal surfaces, and adds project/thread management tools for orchestration workflows. It also makes those capabilities explicit in the chat turn instructions so they are surfaced when prompts are generated.

The branch has been squashed into a single commit for easier review.

## What changed
- Added a dedicated browser-opening alias, `browser_open`, that routes to the existing internal preview/browser broker.
- Added `terminal_run` so an MCP client can open the scoped built-in terminal and send a shell command in one call.
- Added orchestration MCP tools for:
  - `projects_list`
  - `threads_list`
  - `threads_create`
  - `threads_archive`
- Extended shared contracts for the new orchestration and terminal tool inputs/results.
- Registered the new toolkits in the MCP HTTP server and covered them with server tests.
- Updated the Codex developer instructions so chat prompts explicitly mention the new browser, terminal, and thread-management tools.

## Behavior details
- `browser_open` reveals the internal browser for the current thread and can optionally navigate to a URL immediately.
- `terminal_run` reuses the scoped terminal session when possible, opens it if needed, and writes the command with a trailing newline.
- `projects_list` returns the visible project shells from the current orchestration snapshot.
- `threads_list` filters threads by `projectId`.
- `threads_create` uses project defaults when a thread-level model selection is not supplied, then reloads the created thread from projection state.
- `threads_archive` is idempotent when the thread is already archived and otherwise dispatches the archive command before reloading the updated thread.
- The developer prompt now names these capabilities explicitly so they are visible in chat-based turn construction, not just as hidden MCP registrations.

## Validation
- `bun fmt`
- `bun lint`
- `bun typecheck`

## Notes
The repository already had the internal preview/browser surface; this PR makes that capability available as a clearer MCP entrypoint for agents and pairs it with terminal and thread-management actions.

### DIFF
--- a/apps/server/src/mcp/McpHttpServer.test.ts
+++ b/apps/server/src/mcp/McpHttpServer.test.ts
@@ -1,15 +1,31 @@
 import { expect, it } from "@effect/vitest";
 import { NodeHttpServer } from "@effect/platform-node";
-import { EnvironmentId, PreviewTabId, ProviderInstanceId, ThreadId } from "@t3tools/contracts";
+import {
+  DEFAULT_PROVIDER_INTERACTION_MODE,
+  DEFAULT_RUNTIME_MODE,
+  EnvironmentId,
+  PreviewTabId,
+  ProjectId,
+  ProviderInstanceId,
+  ThreadId,
+  type OrchestrationProjectShell,
+  type OrchestrationThreadShell,
+  type TerminalSessionSnapshot,
+} from "@t3tools/contracts";
 import * as Effect from "effect/Effect";
 import * as Layer from "effect/Layer";
+import * as Option from "effect/Option";
 import * as Stream from "effect/Stream";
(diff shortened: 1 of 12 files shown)


juliusmarminge:

#2829 ships an equivalent capability natively — its MCP orchestrator toolkit (`apps/server/src/mcp/toolkits/orchestrator/`) already exposes `delegate_task`, `create_threads`, `t3_thread_start/list/read/send/wait/interrupt` and `schedule_task`, so this PR's parallel implementation would land as a competing surface on a base that no longer exists in the same shape.

**If there's a capability here that the V2 toolkit doesn't cover, please rebase onto `main` after #2829 merges and reopen** — I'd rather extend the shipped toolkit than run two.

PR #3500, opened 2026-06-22
Raise RPC WebSocket ping interval to prevent reconnect loop on large workspaces

mincua:
## What Changed

  Raise the Effect RPC client WebSocket ping interval from **5s to 45s**, via the existing `effect` patch (`patches/effect@4.0.0-beta.78.patch`).
  One-line keepalive change; no source or behavior changes beyond the ping cadence.

  ## Why

  Opening a thread on a large workspace drops the app into an endless WebSocket disconnect/reconnect loop, and the thread never becomes usable.

  The Effect RPC client pinger (`makePinger` in `effect`'s `dist/unstable/rpc/RpcClient.js`) pings every 5s and tears down the transport if the peer
  doesn't answer within that window. Opening a thread keeps the server event loop busy for ~10–15s — per-project VCS/git status, repository detection,
  and favicon resolution across every sidebar project (scales with project count and repo size). The server misses the ping, the client declares the
  transport dead and reconnects, the reconnect restarts the same heavy load, the next ping is missed, and it loops.

  Confirmed from logs: each WS session ends `clean-client-closed` with `aliveMs ≈ 13000`, roughly one per ~13–26s; the client side reports
  `ConnectionTransientError reason:"transport"` at ~10s, repeatedly. The backend PID stays stable (not a process restart) and the establishment /
  socket-open timeouts are never reached — only the 5s ping is. Raising it to 45s lets the connection survive a busy-but-alive backend.
  
  
  A more thorough fix would keep the event loop responsive during thread load (cache/dedup the per-project VCS detection and favicon resolution, and
  bound concurrent `git` spawns) and/or expose the ping interval as a `makeProtocolSocket` option upstream in Effect instead of hard-coding it in a
  vendored patch. Happy to take it in either direction.

  ## Checklist

  - [x] I included before/after screenshots for any UI changes — N/A, no UI changes
  - [x] I included a video for animation/interaction changes — N/A

---

> [!NOTE]
> **Medium Risk**
> Changes keepalive behavior for all Effect RPC WebSocket clients monorepo-wide; slower ping failure detection trades off against fewer false disconnects on a loaded backend.
> 
> **Overview**
> Updates the vendored **`effect@4.0.0-beta.78`** pnpm patch so the RPC socket client’s **`makePinger`** loop waits **45 seconds** between pings instead of **5 seconds**, giving the server more time to answer while the event loop is busy (e.g. heavy thread/workspace load) before the client treats the transport as dead and reconnects.
> 
> The patch refresh also carries the repo’s other **`RpcClient`** customizations (request/connection hooks, ping-timeout handling) and **`pnpm-lock.yaml`** is updated so every workspace consumer picks up the new patch hash.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01bf240b756982758f6bd70b847ee0877b66cf3f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Raise RPC WebSocket ping interval from 5s to 45s to prevent reconnect loops on large workspaces
> - Increases the heartbeat ping interval in `RpcClient.makeProtocolSocket` from 5 seconds to 45 seconds, preventing frequent reconnects on large workspaces where ping/pong round-trips may take longer.
> - Adds an optional `onPing` hook via a new `ConnectionHooks` context tag, allowing callers to run an effect before each ping is sent.
> - Exports `ConnectionHooks` and `layerHook` from the `RpcClient` module so consumers can register hooks at the call site.
> - The returned socket object now includes a `reset` function (exposed as `latch.await`) alongside `timeout`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 01bf240.</sup>
> 
>

### DIFF
--- a/patches/effect@4.0.0-beta.78.patch
+++ b/patches/effect@4.0.0-beta.78.patch
@@ -1,5 +1,5 @@
 diff --git a/dist/unstable/ai/McpServer.js b/dist/unstable/ai/McpServer.js
-index 4819ff0..f5e0f60 100644
+index 38eb099d99026120a437701adc4b0fc5b1d0c714..fe29f3368f6d5e52034728ce789c73a318ccc453 100644
 --- a/dist/unstable/ai/McpServer.js
 +++ b/dist/unstable/ai/McpServer.js
 @@ -235,6 +235,26 @@ export const run = /*#__PURE__*/Effect.fnUntraced(function* (options) {
@@ -30,7 +30,7 @@ index 4819ff0..f5e0f60 100644
      clientSessions
    }));
 diff --git a/dist/unstable/rpc/RpcClient.d.ts b/dist/unstable/rpc/RpcClient.d.ts
-index b0f61df..ead663e 100644
+index b0f61df34613dee99a6f6f1b446a282b2319f5ab..ead663e54cee98fd799a3f3f0761f3f00c6e8420 100644
 --- a/dist/unstable/rpc/RpcClient.d.ts
 +++ b/dist/unstable/rpc/RpcClient.d.ts
 @@ -2,6 +2,7 @@ import * as Cause from "../../Cause.ts";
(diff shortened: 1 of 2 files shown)


anschmieg:
I confirmed the 5-second heartbeat failure this PR addresses on v0.0.28, using a macOS desktop client and a remote Linux server over both Tailscale and Cloudflare.

The WebSocket handshake and `server.getConfig` succeeded, but eager thread hydration produced 1.28 MB and 2.27 MB frames, with another subscribed thread having a 69.9 MB cached snapshot. The client sent its first RPC `Ping` while those ordered frames were still draining; the `Pong` did not arrive within five seconds, so the client reconnected and repeated the same hydration cycle. Loopback remained healthy because it drained the initial state quickly.

As a workaround, archiving the three heavy, stopped threads appears to have stopped the reconnect loop for me. I have not tested this branch directly, but raising the interval to 45 seconds matches the observed failure and should provide an effective mitigation. However, I think a more structural solution should be sought for this.

juliusmarminge:
Closing in favor of author-designated replacement #3817, which carries the same heartbeat mitigation with the maintained 60-second value.

PR #3, opened 2026-02-09
Git integration: branch picker + worktrees

juliusmarminge:
Closes #7 (Does not implement multiple threads per worktree)

## Summary

Users can now right-click on "+ New thread" to create threads from specific git branches, with optional worktree creation for isolated agent work.

## Features

- Branch context menu showing local branches sorted by current/default/alphabetical
- Option to create git worktree for each branch before starting a thread
- Graceful handling of non-git repositories with "Initialize git repo" button
- Thread model extended with branch/worktreePath fields for persistence
- Agent sessions use worktree path as cwd when available
- Updated persistence schema to store branch and worktree info

## Test Plan

- Right-click "+ New thread" on a git-tracked project to see branch list
- Select a branch to create thread with that branch in metadata
- Click "worktree" button to create isolated worktree before starting thread
- On non-git project, click to see "Not a git repository" with init option
- Verify agent runs in worktree cwd when applicable
- Verify branch/worktree badge displays in chat header

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

<a href="https://app.devin.ai/review/pingdotgg/codething-mvp/pull/3" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Branch toolbar beneath chat input to choose branch and environment (Local vs Worktree); environment locks after the first message.
  * Branch picker auto-populates from the repository; toolbar hidden for non-git projects.
  * First message in Worktree mode can create a new worktree and associate it with the thread; branch/worktree badges appear in header and sidebar.

* **Tests**
  * Extensive unit and integration tests for branch, worktree, checkout, and end-to-end git flows.

PR #3623, opened 2026-06-30
fix(desktop): resolve WSL backend getting stuck connecting (#3611)

### LINKED ISSUE #5: feat: Github Integration
Main goal: Go from idea to PR blazingly fast

- [x] Core features (commit, push, create/open PR)
- [ ] commit status, mergability checks etc

### Conductor:
+ ✅ One click action for a lot of stuff (these are really just prompt templates exposed as buttons)
  - Commit & Push
  - Create PR
  - Resolve Conflicts
  - Fix Errors
    - Their solution here is subpar at best though.. they just paste the entire ci log to context. Models are veeeery good at using the `gh` cli already so they can very easily just figure it out on their own and selectively get what they need, especially when paired with a SKILL 
- ❌ Their sync is terrible...
- ❌ Everything HAS to be a github repo

### Codex
- ✅ UI is more selective
  - You can choose to only commit for example wihtout pushing  
- ❌ You can't one-click create feature branch if you have local changes on main 

## Others

- Would like to have stacking (Graphite?)

### LINKED ISSUE #3611: [Bug]: Windows desktop 0.0.28 gets stuck on connecting to WSL after enabling WSL support
## Before submitting

## Area

apps/desktop

## Steps to reproduce

1. Install and start T3 Code desktop on Windows.
2. Use T3 Code `0.0.28.0` / file version `0.0.28`.
3. In Settings, enable the new WSL support for the `Ubuntu` WSL distro. The persisted desktop settings after this were:

   ```json
   {"wslBackendEnabled":true,"wslDistro":"Ubuntu","wslOnly":true}
   ```

4. Let the app restart.
5. Restart the app again, or uninstall and reinstall the app.
6. Install PowerShell 7 and Git for Windows, restart T3 Code, then enable WSL support again.

## Expected behavior

The desktop app should connect to the WSL backend, or it should fail with an actionable error and provide a recoverable path back to the Windows backend.

Uninstalling/reinstalling should also not leave the app permanently stuck in a WSL-only startup path without an obvious recovery option.

## Actual behavior

The app remains on the splash state showing `connecting to WSL...`.

Restarting does not fix it. Uninstalling and reinstalling does not fix it because `%USERPROFILE%\.t3\userdata\desktop-settings.json` still contains the WSL-only settings.

Installing PowerShell 7 and Git for Windows did not fix the WSL startup loop. After installing PowerShell 7, `pwsh.exe` is found, so the original `spawn pwsh.exe ENOENT` warning is no longer the active blocker.

In the logs, WSL preflight succeeds, then the backend exits before readiness and the desktop keeps scheduling restarts. The app polls the WSL backend URL, but the child process exits with `code=0` before it becomes ready. No Linux-side process ever listens on the selected WSL backend port.

## Impact

Blocks work completely

## Version or commit

T3 Code desktop `0.0.28.0` / file version `0.0.28`

## Environment

- Windows desktop app
- Windows version: `10.0.26200.8655`
- WSL distro: `Ubuntu`, WSL version `2`, state `Running`
- T3 settings after enabling WSL support:

  ```json
  {"wslBackendEnabled":true,"wslDistro":"Ubuntu","wslOnly":true}
  ```

- Initially, Windows `pwsh.exe` was not available on PATH:

  ```text
  where.exe pwsh
  INFO: Could not find files for the given pattern(s).
  ```

- After installing PowerShell 7, `pwsh.exe` is found:

  ```text
  C:\Users\mstep\AppData\Local\Microsoft\WindowsApps\pwsh.exe
  ```

- Git for Windows is installed and works by absolute path:

  ```shell
  C:\Program Files\Git\cmd\git.exe --version
  git version 2.54.0.windows.1
  ```

- There was already a WSL-side `npx t3` process listening on port `3773`, so the Windows desktop selected backend port `3774`.

## Logs or stack traces

Initial desktop log excerpts, redacted to remove user-specific paths:

```shell
WARN component=desktop-shell-environment probe=powershell-no-profile executable=pwsh.exe
cause: Error: spawn pwsh.exe ENOENT

WARN component=desktop-shell-environment probe=powershell-profile executable=pwsh.exe
cause: Error: spawn pwsh.exe ENOENT

INFO component=desktop-bootstrap selected backend port via sequential scan
port: 3774
startPort: 3773

INFO component=desktop-window connecting splash shown

# WSL preflight in desktop.trace.ndjson appeared to complete successfully:
# - probeDistros success
# - windowsToWslPath success
# - ensureNodePty success
# - getDistroIp success
# Then desktop polled the WSL backend endpoint:
# http://172.x.x.x:3774/.well-known/t3/environment

ERROR component=desktop-backend-instance:primary backend exited unexpectedly; restart scheduled
reason: code=0
delayMs: 500
```

After installing PowerShell 7 and Git for Windows, re-enabling WSL reproduced the same startup loop without the `pwsh.exe ENOENT` blocker.

Fresh WSL-mode behavior:

```shell
# WSL preflight still succeeds:
# - probeDistros success
# - windowsToWslPath success
# - ensureNodePty success
# - getDistroIp success

# Desktop polls:
http://172.19.227.101:3774/.well-known/t3/environment

# server-child.log repeats child starts and clean exits with no stdout/stderr:
backend child process session start ... runId=d98663cc90e4 ... pid=3464 port=3774 cwd=C:\Users\mstep
backend child process session end   ... runId=d98663cc90e4 ... pid=3464 code=0
```

Captured T3 WSL backend child command line from Windows process inspection:

```shell
wsl.exe -d Ubuntu --exec env \
  "PATH=/home/stepanek/.nvm/versions/node/v18.20.8/bin:...:/mnt/c/Program Files/Git/cmd:..." \
  /home/stepanek/.nvm/versions/node/v18.20.8/bin/node \
  /mnt/c/Users/mstep/AppData/Local/Programs/t3code/resources/app.asar.unpacked/apps/server/dist/bin.mjs \
  --bootstrap-fd 0
```

The child exits with `code=0` before readiness. Since there is no server output at all in this path, this looks like either the WSL backend bootstrap payload/FD path is failing before logging, or the server process is treating the bootstrap input as complete/empty and exiting cleanly.

No Linux-side process ever listens on `3774`; `ss -ltnp` in WSL only shows the existing unrelated/local `127.0.0.1:3773` T3 server.

After disabling WSL mode, the Windows backend produces normal server logs, for example:

```shell
runId=f4c7297cd553
Listening on http://127.0.0.1:3774
POST /oauth/token 200
```

## Workaround

Stop the Windows T3 Code process and edit:

```text
%USERPROFILE%\.t3\userdata\desktop-settings.json
```

Change the settings from:

```json
{"wslBackendEnabled":true,"wslDistro":"Ubuntu","wslOnly":true}
```

to:

```json
{"wslBackendEnabled":false,"wslDistro":"Ubuntu","wslOnly":false}
```

Then restart T3 Code. The app boots again using the Windows backend on `127.0.0.1:3774`.

## Confirmed root cause from source/runtime investigation

The WSL preflight is selecting Node `v18.20.8` from nvm default, even though the server package declares `engines.node = "^22.16 || ^23.11 || >=24.10"`.

In `apps/server/src/bin.ts`, the packaged server entrypoint is gated by:

```ts
if (import.meta.main) {
  Command.run(cli, { version: packageJson.version })...
}
```

Runtime check on this machine:

```text
/home/stepanek/.nvm/versions/node/v18.20.8/bin/node -> import.meta.main is undefined
/home/stepanek/.nvm/versions/node/v22.22.1/bin/node -> import.meta.main is true
```

So with Node 18, the bundled WSL backend module loads and exits `0` before the CLI/server starts. That matches the desktop symptom: no WSL server ever listens on the selected port, desktop keeps polling `/.well-known/t3/environment`, and the splash remains stuck at "connecting to WSL...".

Why preflight misses it: `apps/desktop/src/wsl/DesktopWslEnvironment.ts` passes the server engine range into the WSL node resolver, but after nvm default leaves an out-of-range `node` on PATH, the node-pty probe can still succeed on Node 18. The code then returns `{ ok: true, nodePath, resolvedPath }` on `probe.exitCode === 0` before enforcing that the selected `nodePath` satisfies `serverPackageJson.engines.node`.

Local workaround that makes the Windows-launched WSL probe select a compatible runtime:

```bash
source ~/.nvm/nvm.sh
nvm alias default 22.22.1
nvm use 22.22.1
```

After that, a Windows-launched WSL probe resolves:

```text
nodePath:/home/stepanek/.nvm/versions/node/v22.22.1/bin/node
nodeVersion:22.22.1
importMetaMain:true
```

Suggested product fix: make WSL preflight fail/fallback when the resolved `nodePath` does not satisfy the server package engine range, even if the node-pty probe succeeds. Also log the selected WSL `nodePath` and `node -v` in the backend launch/preflight logs so this is visible without process inspection.

Jgratton24:
Fixes #3611

## What Changed

Five related fixes for the WSL backend getting permanently stuck on "Connecting to WSL...":

1. **Hold the WSL bootstrap stdin open** instead of closing it right after writing the bootstrap envelope (`DesktopBackendManager.ts`), fixing a race where the envelope could be lost if stdin closed before the WSL-relayed reader finished reading it. Same mechanism as #3613.
2. **Validate the WSL Node version against the server's required engine range** even when a Node binary does resolve, not just when none resolves at all (`DesktopWslEnvironment.ts`), so an incompatible `nvm` default (e.g. Node 18) is caught at preflight instead of the server silently no-op'ing, and fail closed if the probe doesn't report a version at all when a range is required. Same fix as #3621.
3. **Cap consecutive post-preflight backend exits and fall back to Windows** instead of restarting forever (`DesktopBackendManager.ts`), so any backend that keeps exiting before becoming ready surfaces an actionable error and recovers automatically — including across reinstall, where a persisted `wslOnly: true` previously left the app stuck with no way out. Uses a dedicated counter scoped strictly to post-spawn exits, separate from the general restart-backoff counter, so ordinary pre-spawn preflight retries (e.g. WSL cold-start) can't trip it early, and resets on a fresh start so the cap firing once doesn't permanently remove the retry budget.
4. **Detach the bootstrap reader once the envelope is parsed**, closing the readline interface and releasing the fd on every terminal path instead of only on interruption (`apps/server/src/bootstrap.ts`).
5. **Listen for the bootstrap reader's errors on the readline interface, not the raw stream** (`apps/server/src/bootstrap.ts`), closing a gap where an error on the stream *before* any line ever arrived was still an uncaught exception even with fix #4 in place — Node's `readline.Interface` re-emits stream errors onto itself, and nothing was listening there.

## Why

#3611 has two independent root causes that produce the identical "stuck connecting" symptom, plus a structural gap that turns either one into a permanent trap:

- The stdin race (#1 above), already identified and fixed by #3613.
- A WSL Node-engine-version preflight gap (#2 above), already identified and fixed by #3621 — this is the reporter's own diagnosed cause on their machine.
- No cap on post-preflight restart loops (#3 above) — this is why restart, and even reinstall, didn't help the reporter; nothing ever flipped `wslOnly` back to `false`.

Fixes #4 and #5 exist because of what we found testing #1 against a real WSL2 box rather than just unit tests: holding stdin open forever (#1 / #3613's approach) only works if something on the server side actually stops reading from that now-permanently-open stream once it has what it needs, and never crashes on a stream-level error at any point in its lifecycle. The existing bootstrap reader did neither. We observed this live once: the WSL backend crash-looped 5/5 times with an uncaught `EAGAIN` before ever logging "Listening on", with #1 applied but not yet #4. We were not able to independently reproduce that exact EAGAIN trigger in isolation afterward (a standalone script holding stdin open against real `wsl.exe` for several minutes did not reproduce it), so we're not claiming it's a guaranteed, deterministic failure — treat it as a real observed failure, not a proven-deterministic one.

What we *have* deterministically reproduced, independent of that one observation, is a structurally related bug in the same reader: verified directly against the production WSL Node version (v24.14.0), an error on the bootstrap stream before any line is ever read is an uncaught exception with only a stream-level error listener (fix #4 alone doesn't help here, since `cleanup()` is never reached — the process crashes before our own handler ever runs), and is handled cleanly once the listener is registered on the readline interface instead (fix #5). Combined, #4 and #5 mean the bootstrap reader no longer crashes the process on a stream error at any point in its lifecycle, before or after the envelope is parsed. #3613 alone leaves both exposed.

We bundled all five here because #4 and #5 only make sense in the context of #1, and #3 is a generic safety net for the failure class #1's and #2's bugs both fall into. We're aware #1 and #2 overlap with #3613 and #3621 respectively and are happy to drop either piece from this PR if those merge first — #3, #4, and #5 are the parts that aren't covered anywhere else yet.

## UI Changes

No new UI, but fix #3 means the existing "WSL backend couldn't start, falling back to Windows" dialog (already shown in #3621's after-screenshot) now also fires for failures that happen after a successful preflight, not just preflight failures.

## Validation

Validated against the real running app on Windows + WSL2, and against the actual production WSL Node version directly, not just unit tests:
- Forced the original stuck-connecting bug on unpatched `main`.
- Forced the EAGAIN crash live once with fix #1 applied alone (no fix #4) — confirmed 5/5 crash-loop. This specific trigger was not independently reproduced in isolation afterward; see the "Why" section for what that does and doesn't establish.
- Independently and deterministically verified (against production WSL Node v24.14.0, outside the full app) that an error on the bootstrap stream before any line arrives is an uncaught exception without fix #5, and is handled cleanly with it.
- Confirmed WSL-only mode reaches ready in ~25s with zero crashes once all five fixes are applied, and parallel (Windows + WSL) mode has no regression.
- New unit tests added across `DesktopBackendManager.test.ts`, `DesktopWslEnvironment.test.ts`, and `apps/server/src/bootstrap.test.ts`, including regression tests for the never-ready-counter isolation/reset (#3) and the fail-closed missing-version case (#2); existing suites pass with no regressions; typecheck and lint clean. (We could not find a reliable way to add an automated in-process unit test for fix #5's exact scenario — Node's own `fs.ReadStream` internals don't honor module-level mocking of `node:fs` from the test process — so that one rests on the direct production-Node verification above rather than an automated test.)

## Checklist

- [ ] This PR is small and focused — it bundles five fixes; see "Why" for why they're grouped
- [ ] I included before/after screenshots for any UI changes — no new UI, see "UI Changes"

---

> [!NOTE]
> **Medium Risk**
> Changes desktop child-process bootstrap I/O, WSL preflight gating, and backend restart/fallback behavior—high user impact on Windows+WSL but scoped with fd3/stdin branching and new regression tests.
> 
> **Overview**
> Addresses **#3611** by fixing several ways the WSL backend could stay on "Connecting" forever.
> 
> For **stdin-delivered** bootstrap (`bootstrapDelivery: 'stdin'`), the desktop now **keeps stdin open** after writing the JSON line (`Stream.never`) so early EOF cannot race the child's readline reader and drop the envelope. The server **`readBootstrapEnvelope`** now **tears down readline and the fd stream on every exit path** and listens for errors on the **readline interface** instead of the raw stream, which matters once stdin stays open.
> 
> **`DesktopBackendManager`** adds a **`neverReadyAttempt`** counter (separate from preflight/restart backoff): after **five consecutive post-spawn exits before HTTP readiness** on the stdin path, it surfaces a fatal failure via **`onPreflightFailed`** (Windows fallback dialog) and stops looping; **fd3** Windows-native backends are **not** capped. Counters reset on readiness and on a fresh start after parking.
> 
> **WSL preflight** (`ensureNodePtyImpl`) now emits **`nodeVersion:`** from the existing probe and **enforces `engines.node`** when configured—fatal with **`retryLimit: 1`** for a known mismatch, fail-closed if the version is missing when a range is required.
> 
> Extensive **unit tests** cover stdin vs fd3 streams, never-ready escalation, counter isolation, and engine-range preflight.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80f8711e13459bd6671d82c923de7c21196a58d0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix WSL backend getting stuck in connecting state by capping never-ready exits and closing bootstrap stdin
> - For `bootstrapDelivery:'stdin'`, stdin is now held open indefinitely (via `Stream.never`) instead of closing after writing the bootstrap line, preventing the backend from seeing premature EOF.
> - Adds a `neverReadyAttempt` counter in [DesktopBackendManager.ts](https://github.com/pingdotgg/t3code/pull/3623/files#diff-45848e60bcec64dd7239fc34379d23407c47d0ade3f351e89a4efc7f6b42757f): after `MAX_PREFLIGHT_FAILURE_ATTEMPTS` consecutive exits before readiness, a fatal failure is surfaced and the instance either restarts or stops.
> - The bootstrap reader in [bootstrap.ts](https://github.com/pingdotgg/t3code/pull/3623/files#diff-134419e5ebf3f75491ebaa844716de9728271ae9e3b8ceb6f33964851d52a147) now closes its fd immediately after parsing the envelope or on error, preventing lingering readers and late error events.
> - WSL preflight in [DesktopWslEnvironment.ts](https://github.com/pingdotgg/t3code/pull/3623/files#diff-eb443cdb82c5c1542b51d3b5bd796cd760f44161c7cd5f8ead2affd539c22d1d) now validates the resolved Node.js version against a configured engine range, returning a fatal failure if the version is missing or out of range.
> - Risk: WSL instances using `bootstrapDelivery:'stdin'` that previously restarted indefinitely will now stop after hitting the never-ready cap.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 80f8711.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/backend/DesktopBackendManager.test.ts
+++ b/apps/desktop/src/backend/DesktopBackendManager.test.ts
@@ -6,6 +6,7 @@ import { assert, describe, it } from "@effect/vitest";
 import * as Deferred from "effect/Deferred";
 import * as Duration from "effect/Duration";
 import * as Effect from "effect/Effect";
+import * as Fiber from "effect/Fiber";
 import * as FileSystem from "effect/FileSystem";
 import * as Layer from "effect/Layer";
 import * as Option from "effect/Option";
@@ -218,6 +219,101 @@ describe("DesktopBackendManager", () => {
         assert.deepEqual(yield* decodeBootstrap(bootstrapJson), configWithObservability);
       }),
     ),
+  );
+
+  it.effect(
+    "holds stdin open forever for bootstrapDelivery:'stdin' (#3611 fix 1) while fd3 still terminates normally",
+    () =>
+      Effect.gen(function* () {
+        // --- stdin-delivered (WSL) path: the bootstrap stream must never
+        // complete in normal operation, eliminating the readline 'line' vs
+        // 'close' race in apps/server/src/bootstrap.ts. ---
+        let stdinStream: ChildProcess.CommandInput | undefined;
+        const stdinSpawned = yield* Deferred.make<void>();
+        const stdinSpawnerLayer = Layer.succeed(
+          ChildProcessSpawner.ChildProcessSpawner,
+          ChildProcessSpawner.make((command) =>
+            Effect.gen(function* () {
+              if (command._tag === "StandardCommand") {
+                stdinStream = command.options.stdin;
+              }
+              yield* Deferred.succeed(stdinSpawned, void 0);
+              return makeProcess({ exitCode: Effect.never });
+            }),
+          ),
+        );
+
+        yield* Effect.scoped(
+          Effect.gen(function* () {
+            const instance = yield* makeTestInstance({
+              spawnerLayer: stdinSpawnerLayer,
+              config: {
+                ...baseConfig,
+                bootstrapDelivery: "stdin",
+                args: ["/server/bin.mjs", "--bootstrap-fd", "0"],
+              },
+              httpClientLayer: httpClientLayer(() => Effect.never),
+            });
+
+            yield* instance.start;
+            yield* Deferred.await(stdinSpawned);
+            assert.isDefined(stdinStream);
+            if (typeof stdinStream === "string") {
+              throw new Error("Expected stdin to be a Stream, not a CommandInput string.");
+            }
+
+            const drained = yield* Stream.runDrain(stdinStream).pipe(
+              Effect.timeoutOption(Duration.seconds(30)),
+              Effect.forkScoped,
+            );
+            yield* TestClock.adjust(Duration.seconds(30));
+            const result = yield* Fiber.join(drained);
+            assert.isTrue(Option.isNone(result), "stdin stream completed but must never end");
+          }),
+        );
+
+        // --- fd3-delivered (Windows-native) path: unaffected regression
+        // check -- the bootstrap stream must still terminate normally. ---
+        let fd3Stream: Stream.Stream<Uint8Array> | undefined;
+        const fd3Spawned = yield* Deferred.make<void>();
+        const fd3SpawnerLayer = Layer.succeed(
+          ChildProcessSpawner.ChildProcessSpawner,
+          ChildProcessSpawner.make((command) =>
+            Effect.gen(function* () {
+              if (command._tag === "StandardCommand") {
+                const fd3 = command.options.additionalFds?.fd3;
+                if (fd3?.type === "input") fd3Stream = fd3.stream;
+              }
+              yield* Deferred.succeed(fd3Spawned, void 0);
+              return makeProcess({ exitCode: Effect.never });
+            }),
+          ),
+        );
+
+        yield* Effect.scoped(
+          Effect.gen(function* () {
+            const instance = yield* makeTestInstance({
+              spawnerLayer: fd3SpawnerLayer,
+              httpClientLayer: httpClientLayer(() => Effect.never),
+            });
+
+            yield* instance.start;
+            yield* Deferred.await(fd3Spawned);
+            assert.isDefined(fd3Stream);
+
+            const drained = yield* Stream.runDrain(fd3Stream!).pipe(
+              Effect.timeoutOption(Duration.seconds(30)),
+              Effect.forkScoped,
+            );
+            yield* TestClock.adjust(Duration.seconds(30));
+            const result = yield* Fiber.join(drained);
+            assert.isTrue(
+              Option.isSome(result),
+              "fd3 stream should still terminate normally (no regression)",
+            );
+          }),
+        );
+      }).pipe(Effect.provide(TestClock.layer())),
   );
 
   it.effect("retries HTTP readiness before reporting the backend ready", () =>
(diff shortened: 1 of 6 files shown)


Jgratton24:
Closing this in favor of #5998, with each piece accounted for:

- **Fix 2 (validate WSL Node against the engine range): superseded** — #3621 merged on July 18 with the same fix, which is also what closed the anchor issue #3611. Current main gates on `satisfiesSemverRange` at preflight.
- **Fix 3 (cap consecutive never-ready exits + surface with Windows fallback): ported to #5998** (commit 02343e8), rebuilt on top of the counter machinery that PR adds for readiness timeouts — same design as here: a dedicated counter scoped strictly to post-spawn exits, reset on ready/stop/fresh start, surfacing through the same path as exhausted preflight failures. #5998 also bounds the two sibling failure modes this PR didn't cover (readiness timeouts on a live-but-unreachable backend, and the persisted `wslOnly` trap via the Windows-primary stop).
- **Fixes 1, 4, 5 (hold bootstrap stdin open + server-side bootstrap reader hardening): not carried over.** Fix 1 is #3613's change, which is still open — the reader-lifecycle findings from fixes 4/5 belong in that PR's discussion, since they only matter once stdin is held open. Worth noting for whoever picks #3613 back up: current main's `bootstrap.ts` still listens for errors on the raw stream rather than the readline interface, so the uncaught-exception gap described here may still apply.

This branch is also six weeks behind main (`CONFLICTING`) with a blocking review finding on the exit-counter implementation, so a fresh port was cleaner than a rebase.

PR #3696, opened 2026-07-05
fix(orchestrator): Claude plan-mode plan proposals + real provider error causes

shivamhwp:
## What Changed

Five focused bug fixes for the Claude adapter / run execution in the v2 orchestrator (this PR's branch):

1. **Surface real causes in Claude provider failures** — `ClaudeAgentSdkQueryRunnerError.message` was a fixed string (`"Claude Agent SDK query failed."`), so every provider-error card showed that and nothing else, and the run-path logs printed `cause: [Object]`. The error message now appends the failing SDK method and the underlying defect message, and the three warn/error logs on the run path render `Cause.pretty` output.
2. **Let Claude plan mode propose approvable plans** — plan-mode sessions were launched with `installPermissionCallback: false`, so the SDK never exposed `ExitPlanMode` and plan mode could only ever end in prose: no plan artifact, no "Plan Ready" state, no approve/implement flow (`orchestration_v2_projection_plans` stayed empty). The callback is now installed in plan mode, and `canUseTool` intercepts `ExitPlanMode`: it emits the proposed plan as v2 node/plan/turn-item artifacts (`status: "active"`, which flips `hasActionableProposedPlan`) and denies the tool call with guidance to end the turn, so the session stays in plan mode until the user approves from the app. `emitsProposedPlan` capability flipped to `true`, regression test added for the plan-mode query policy.

3. **Drop poison provider events instead of killing ingestion** — one event failing `ingestNormalized` (e.g. a late async-subagent notification referencing a finalized run) tore down the whole `Stream.tap` pipeline, leaving the session's event routing dead and the next run wedged at `starting` until a server restart. Failed non-terminal events are now logged and dropped; root `turn.terminal` failures still escalate to the existing fallback finalization. Regression test included.
4. **Resolve async subagent completions after turn finalization** — the SDK's `system/task_notification` for an async Task usually arrives after the parent turn finalized (`activeTurn === null`), so it was silently dropped: the subagent projection stayed `running` forever and `RunExecutionService` kept waiting on a child that could never terminalize. Turn contexts with still-running tasks are now retained (keyed by native task id) and late notifications resolve against the owning context, emitting the subagent/node/turn-item updates under the original run.
5. **Skip a stale resume pin after a dirty-shutdown query failure** — resuming pins `resumeSessionAt` to the recorded conversation head; after a dirty shutdown that head may never have been flushed to the native transcript, so the first query failed and only an unexplained manual retry recovered. A pinned query that dies before producing any SDK message now marks the native thread so the next open resumes unpinned (one-shot).

The `ClaudeAgentSdkQueryRunnerError` message is derived from structural fields only (per the Effect service conventions); the underlying SDK error reaches the provider-failure card via `makeProviderFailure`'s existing redaction/bounding pipeline instead.

Known follow-up (not in this PR): a stop/cancel affordance for runs stuck in `queued/preparing/starting`.

## Why

Found while testing this PR end-to-end with the Claude provider:

- Every failure — missing cwd, SDK spawn failure, broken session resume — rendered as the same opaque card (`transport_error · "Claude Agent SDK query failed." · code: null`), and the logs gave nothing more. `makeProviderFailure` already extracts `Error#message`; the information was being discarded at the error-class layer.
- Plan mode on `claudeAgent` was a dead-end: the model itself reports it has no `ExitPlanMode` tool, writes its plan as chat prose, and the plan interaction mode can never hand off to the build flow. With this fix the full loop works: plan request → plan card + "Plan Ready" pill → Implement → build run executes the plan.

Both verified live against a scratch project (screenshots below); full server suite passes (1199 tests).

## UI Changes

No component changes — both fixes are server-side, but they change what the existing UI displays.

**Provider error card** (same failure class — a project whose workspace folder was deleted). Before: the bare one-liner with `code: null`. After: the card names the failing SDK method and the real cause.

| Before | After |
| --- | --- |
| before: bare provider error card | after: provider error card with method and real cause |

**Plan mode.** Before: the plan exists only as chat prose — no plan card, no Plan Ready, nothing to approve — and the model states it has no `ExitPlanMode` tool. After: a proposed-plan card is emitted, the thread flips to "Plan Ready", and Implement kicks off the build run that executes the plan (card + pill mid-flow state in the collapsed section below).

| Before | After |
| --- | --- |
| before: plan as prose, no ExitPlanMode tool | after: plan implemented via the plan flow with changed files |

<details>
<summary>Extra: the proposed-plan card + "Plan Ready" pill + Implement button state (mid-flow)</summary>

plan card with Plan Ready pill and Implement button

</details>

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core orchestration behavior (event ingestion, Claude session resume, plan/subagent lifecycle) in the v2 run path; mistakes could wedge runs or mis-route subagent/plan state, though failures are mostly isolated with logging and tests.
> 
> **Overview**
> Hardens the **Claude v2 adapter** and **run execution** path with five targeted fixes.
> 
> **Plan mode** now installs the SDK permission callback in plan mode and handles `ExitPlanMode` by emitting proposed-plan node/plan/turn-item artifacts (then denying the tool so review stays in-app). Capability `emitsProposedPlan` is set to true.
> 
> **Provider failures** unwrap `ClaudeAgentSdkQueryRunnerError` to the underlying `Error` for `makeProviderFailure`, and run-path logs use `Cause.pretty` instead of opaque cause objects.
> 
> **Async subagents** retain turn context for running tasks after parent finalization so late `task_notification` messages still terminalize subagents on the original run.
> 
> **Resume** skips a stale `resumeSessionAt` pin when a pinned query fails before any SDK message (one-shot per native thread).
> 
> **Run ingestion** logs and drops non-terminal events that fail normalization instead of killing the stream; root `turn.terminal` failures still escalate. Regression tests cover plan-mode query policy and poison-event handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eeb5f9554f115768454469690ebbcd2915d2cc37. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix Claude plan-mode plan proposals and provider error cause unwrapping
> - Sets `emitsProposedPlan` to `true` in `ClaudeProviderCapabilitiesV2` and installs the permission callback in plan mode so the adapter can intercept `ExitPlanMode` tool calls, emit a `proposed_plan` artifact, and keep the session in plan mode.
> - Introduces `claudeProviderFailureCause` to unwrap the underlying `Error` from `ClaudeAgentSdkQueryRunnerError`, so provider failure logs now report the real SDK error cause rather than the wrapper.
> - Retains `ActiveClaudeTurnContext` for async tasks after a turn finalizes so `system.task_notification` messages received late are still resolved correctly.
> - Drops non-terminal provider events that fail ingestion normalization in `RunExecutionService` (logging a warning) while still aborting on root `turn.terminal` failures.
> - If a query opened with `resumeSessionAt` fails before any SDK message arrives, the next open for that native thread suppresses the resume pin and logs a warning.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized eeb5f95.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration-v2/Adapters/ClaudeAdapterV2.test.ts
+++ b/apps/server/src/orchestration-v2/Adapters/ClaudeAdapterV2.test.ts
@@ -209,6 +209,21 @@ describe("ClaudeAdapterV2 runtime query policy", () => {
       installPermissionCallback: false,
     });
   });
+
+  it("installs the permission callback in plan mode so ExitPlanMode can propose plans", () => {
+    const queryPolicy = claudeRuntimeQueryPolicyForRuntimePolicy(
+      ProviderAdapterV2RuntimePolicy.make({
+        runtimeMode: "full-access",
+        interactionMode: "plan",
+        cwd: "/workspace",
+      }),
+    );
+
+    assert.deepEqual(queryPolicy, {
+      permissionMode: "plan",
+      installPermissionCallback: true,
+    });
+  });
 });
 
 describe("ClaudeAdapterV2 native protocol logging", () => {
--- a/apps/server/src/orchestration-v2/Adapters/ClaudeAdapterV2.ts
+++ b/apps/server/src/orchestration-v2/Adapters/ClaudeAdapterV2.ts
@@ -276,16 +276,23 @@ export class ClaudeAgentSdkQueryRunnerError extends Schema.TaggedErrorClass<Clau
   },
 ) {
   override get message(): string {
-    const detail =
-      this.cause instanceof Error
-        ? this.cause.message
-        : typeof this.cause === "string"
-          ? this.cause
-          : null;
-    return detail === null || detail.trim().length === 0
-      ? `Claude Agent SDK query failed (${this.method}).`
-      : `Claude Agent SDK query failed (${this.method}): ${detail}`;
+    return `Claude Agent SDK query failed (${this.method}).`;
   }
+}
+
+/**
+ * Provider-failure payloads render the failure's own message, so the runner
+ * wrapper would mask the actionable SDK error (missing binary, resume
+ * rejection, spawn failure). Surface the underlying error to the existing
+ * redaction/bounding pipeline in makeProviderFailure; the wrapper stays the
+ * failure for anything that is not a real Error.
+ */
+const isClaudeAgentSdkQueryRunnerError = Schema.is(ClaudeAgentSdkQueryRunnerError);
+
+function claudeProviderFailureCause(failure: unknown): unknown {
+  return isClaudeAgentSdkQueryRunnerError(failure) && failure.cause instanceof Error
+    ? failure.cause
+    : failure;
 }
 
 export interface ClaudeAgentSdkQueryRunnerShape {
@@ -2744,7 +2751,10 @@ export function makeClaudeAdapterV2(
               ? {}
               : {
                   failure: makeProviderFailure({
-                    cause: cause === undefined ? undefined : Cause.squash(cause),
+                    cause:
+                      cause === undefined
+                        ? undefined
+                        : claudeProviderFailureCause(Cause.squash(cause)),
                     class: "transport_error",
                   }),
                 }),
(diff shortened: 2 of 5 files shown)


juliusmarminge:
Closing because this five-fix bundle is now too stale and conflicted to rebase safely. The target orchestration code has moved substantially, so resolving the conflicts would be a behavioral reimplementation rather than a mechanical update, and the combined scope makes regression attribution difficult. Any still-reproducible Claude plan/error/notification issues should be reopened as focused PRs against the current adapter with current regression coverage.

Issue #3709, opened 2026-07-05
WSL backend hangs at "Connecting to WSL…" (silent exit) when login-shell node is < 22.16


arturskruze:
## Summary

On Windows+WSL, the desktop backend hangs forever at **"Connecting to WSL…"** when the WSL login shell resolves a `node` that is present but **below the engine floor** (`< 22.16`). The server process starts, loads, and exits `0` **silently** without binding — so the desktop just respawns it in a loop with no diagnostic.

Two independent issues combine to make this a silent, hard-to-diagnose failure.

## Environment

- Windows 11 + WSL2 (Ubuntu), backend launched via `wsl.exe -- node … --bootstrap-fd 0`
- `nvm` default node was **v22.12.0**; a satisfying **v22.21.1** was also installed but not the default
- Server `engines.node`: `^22.16 || ^23.11 || >=24.10`

## Root cause 1 — `import.meta.main` guard fails silently on Node < 22.16

`apps/server/src/bin.ts` gates all startup on `import.meta.main`:

```ts
if (import.meta.main) {
  Command.run(cli, { version: packageJson.version }).pipe(/* … */ NodeRuntime.runMain);
}
```

`import.meta.main` was added in Node **22.16.0**. On **22.12.0** it is `undefined`, so the guard is falsy: the module loads its entire import graph and the process exits `0` with **no output and no error**. Verified:

```
$ node -v            # v22.12.0
$ node -e 'x' --input-type=module ...   # import.meta.main === undefined
# server run under 22.12.0 -> loads, exits 0, never creates a socket (strace: zero bind/listen)
# server run under 22.21.1 -> binds and serves /.well-known/t3/environment correctly
```

Because it exits `0` cleanly, the desktop's restart loop keeps respawning it and the UI shows "Connecting to WSL…" indefinitely. Suggestion: detect an unsupported/`undefined import.meta.main` Node and **fail loudly** (clear stderr message + non-zero exit) so the desktop can surface it, e.g. a robust entrypoint check plus an explicit engine assertion:

```ts
import { pathToFileURL } from "node:url";
const isEntrypoint =
  import.meta.main ??
  (process.argv[1] !== undefined && import.meta.url === pathToFileURL(process.argv[1]).href);
if (isEntrypoint) { /* … */ }
```

## Root cause 2 — `ensure_remote_node_path` won't upgrade away from an unsatisfying node

[...]

PR #3710, opened 2026-07-05
fix(ssh): select a satisfying node when the login-shell node fails the engine range

arturskruze:
## What

`ensure_remote_node_path` (in `REMOTE_NODE_ENV_SCRIPT`) only fell back to version managers (`mise`/`fnm`/`nodenv`/`nvm`, and the nvm version scan) when there was **no `node` on `PATH` at all** (`! command -v node`). When a login shell already exposes a `node` that **fails the engine range** — e.g. an `nvm default` below the minimum, or a system `/usr/bin/node` — every fallback was skipped and the unsatisfying node was exec'd anyway (`… || true`).

This gates the fallbacks on a combined **"node present *and* satisfies engine"** check (`remote_node_ready`), so a satisfying node that's already installed via a version manager actually gets selected. The nvm version scan also stops at the first satisfying version instead of relying on glob ordering.

## Why

On Windows+WSL this made the desktop backend launch under an unsupported node: e.g. the nvm `default` was `22.12.0` while a satisfying `22.21.1` was installed but not selected. Because the server bundle gates startup on `import.meta.main` (added in Node **22.16**), it then loaded its module graph and **exited `0` silently** without binding — and the desktop respawned it forever, showing **"Connecting to WSL…"** with no diagnostic.

Full trace and repro in #3709. (That issue also notes a second, separate hardening idea: have `bin.ts` fail *loudly* on an unsupported node instead of exiting silently — kept out of this PR to stay focused.)

## Scope / risk

- Single function; **behavior-preserving** for the "satisfying node already on PATH" and "no node at all" cases. Only the "unsatisfying node present" case changes.
- No new deps, no interface changes. Existing `packages/ssh/src/tunnel.test.ts` assertions (`ensure_remote_node_path()`, `if ! ensure_remote_node_path; then`, `remote_node_satisfies_engine()`) are unaffected.

---

> [!NOTE]
> **Low Risk**
> Single embedded shell function in SSH remote bootstrap; behavior only changes when an unsatisfying node is already on PATH, with no API or dependency changes.
> 
> **Overview**
> **Remote SSH Node selection** now treats “node on PATH” and “node satisfies the engine range” as one gate via a new `remote_node_ready` helper in the embedded `REMOTE_NODE_ENV_SCRIPT`.
> 
> Previously, `ensure_remote_node_path` only ran **mise/fnm/nodenv/nvm** fallbacks when `node` was missing entirely. If the login shell already had a `node` that **failed** the semver check (e.g. nvm `default` below minimum while a newer install existed), fallbacks were skipped and the bad binary was used—leading to silent exit and stuck “Connecting to WSL…” on Windows+WSL.
> 
> Fallback activation is now **`! remote_node_ready`** at each step. The nvm version-directory scan **stops at the first PATH prefix** where `remote_node_ready` succeeds instead of depending on glob order. Behavior is unchanged when no node exists or when the current `node` already satisfies the range.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27be28c55e67be5f8cfe076a5bc469246113618b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix SSH tunnel node selection when the login-shell node fails the engine range
> - Adds a `remote_node_ready` helper in [tunnel.ts](https://github.com/pingdotgg/t3code/pull/3710/files#diff-7338be65675ab0699d0c8c6426e54f3421747885231f9bbb0ace4ec129cb2072) that checks both node presence and engine compatibility in one place.
> - `ensure_remote_node_path` now continues activating version managers (mise, fnm, nodenv, nvm) even when a node binary exists but doesn't satisfy the engine requirement, instead of short-circuiting on presence alone.
> - The NVM versions loop now breaks as soon as a satisfying version is found rather than iterating all candidates.
> - Behavioral Change: SSH sessions that previously used an incompatible node binary will now attempt to resolve a compatible one via available version managers.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 27be28c. 1 file reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/packages/ssh/src/tunnel.ts
+++ b/packages/ssh/src/tunnel.ts
@@ -336,8 +336,12 @@ remote_node_satisfies_engine() {
 NODE
 }
 
+remote_node_ready() {
+  command -v node >/dev/null 2>&1 && remote_node_satisfies_engine >/dev/null 2>&1
+}
+
 ensure_remote_node_path() {
-  if command -v node >/dev/null 2>&1 && remote_node_satisfies_engine >/dev/null 2>&1; then
+  if remote_node_ready; then
     return 0
   fi
 
@@ -363,7 +367,7 @@ ensure_remote_node_path() {
 
   prepend_path_if_dir "$HOME/.local/share/mise/shims"
   prepend_path_if_dir "$HOME/.mise/shims"
-  if ! command -v node >/dev/null 2>&1 && command -v mise >/dev/null 2>&1; then
+  if ! remote_node_ready && command -v mise >/dev/null 2>&1; then
     eval "$(mise activate sh)" >/dev/null 2>&1 || true
   fi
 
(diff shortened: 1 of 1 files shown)


shivamhwp:

Please add a regression that executes the generated shell bootstrap with an incompatible Node already on PATH and a compatible NVM version available. Cover retaining an already-compatible Node and skipping a later incompatible NVM directory too. The PR changes fallback selection without adding behavioral coverage; string assertions cannot protect that selection.

The entrypoint part of #3709 has changed on main: `bin.ts` now uses `isEntrypoint`, including a fallback when `import.meta.main` is unavailable. Update the description so the remaining version-selection fix is distinct from the historical silent-exit explanation.

PR #3725, opened 2026-07-06
Plugin system (1/8): thread ownership

ccdwyer:
## What Changed

Adds `projection_threads.owner` (default `'user'`, migration 033) so that agent threads created by plugins are attributable and can be filtered out of user-facing surfaces. User-facing snapshots / lists / relay / the live shell stream filter to `user`-owned threads; the command read model deliberately sees **every** thread (the decider must, to validate and replay). Nothing plugin-specific ships in this part — it stands on its own.

Extending t3code today means forking it. This is **part 1 of 8** of a small, dependency-ordered stack that adds a **runtime plugin system**: install third-party features (server + web/Electron) at runtime behind curated, capability-scoped host APIs — no fork required. Thread ownership is the foundation: once plugins can spawn agent threads, those threads need to be attributable so they don't pollute the user's thread list. This piece is useful and mergeable on its own even if none of the later parts land.

**A deliberate goal here is to take pressure off the core.** Right now, anyone who wants to customize t3code has to fork it or try to land their feature upstream — which is exactly the stream of unwanted-feature PRs this repo asks people not to open. A plugin system gives that energy somewhere else to go: contributors ship their customizations as installable plugins instead of PRs, so maintainers field fewer feature requests they have to decline and the core stays lean. The intent is to make this the *last* large feature you have to review — after it, most "can you add X" turns into "ship X as a plugin."

## The full project — which parts deliver what

The repo asks for small, focused PRs, so the plugin system is submitted as 8 dependency-ordered PRs. **Merge bottom-up** (each later part's diff is cumulative against `main` until its parent merges — GitHub can't base a fork PR on a fork branch; parts 2–8 are drafts blocked on their parent):

| Part | PR | Delivers | Builds on |
|---|---|---|---|
| 1 | **this** | thread ownership *(standalone)* | — |
| 2 | #3726 | plugin host lifecycle + `@t3tools/plugin-sdk` | 1 |
| 3 | #3727 | WS RPC transport + plugin-aware auth scopes | 2 |
| 4 | #3728 | mechanical capability façades (db/secrets/env/projections/textgen/sourcecontrol/terminals/http) | 3 |
| 5 | #3729 | agents + vcs capabilities | 4 |
| 6 | #3730 | web host + extension points (routes/sidebar/settings/palette) | 5 |
| 7 | #3731 | marketplace + install lifecycle + settings UI + fixture | 6 |
| 8 | #3732 | filesystem/httpClient capabilities + hardening | 7 |

[...]

PR #3726, opened 2026-07-06
Plugin system (2/8): host foundation + plugin SDK

ccdwyer:
## What Changed
The server plugin host: a per-plugin Effect `Scope` lifecycle (verify → import → register → migrate → recover → mount → ready), `PluginMigrator` (plugin tables gated to a `p_<id>_*` prefix at migration time via a `sqlite_master` diff, core migration 034), `PluginModuleLoader` + an ESM resolve hook sharing the host's `effect`/SDK singletons, a single-writer `PluginLockfileStore`, and the `@t3tools/plugin-sdk` package.

## Why
**Part 2 of 8** of a runtime plugin system (see #3725 for the full overview + which parts deliver what). Builds on #3725. This is the foundation every capability and the marketplace sit on: it owns each plugin's lifecycle, isolation scope, and on-disk state.

## Review — four-model tribunal (GPT-5.5 · Grok · GLM-5.2 · Fable)
- **`start` bypassed the single-flight lock (MUST):** an enable/install RPC racing startup could double-load a plugin and orphan a runtime scope → routed through the per-plugin activation lock.
- **`Effect.exit` doesn't capture *external* interrupts in effect beta.78 (MUST):** on real shutdown / a canceled RPC, `Scope.close` + the activating-marker clear never ran → wrapped the ladder in `uninterruptibleMask` (confirmed against the vendored runtime; verified by runtime probes since the integration test hangs in vitest under forked activation).
- **SDK `writeFileAtomic` couldn't replace an existing file (MUST)** → overwrite-permitted rename mode.
- SHOULDs: non-atomic stale-lock reclaim → atomic `rename`-claim + owner-token re-verify; singleton-resolution latching before/despite failure; non-crash-safe preserve-data; no host-shutdown scope-close; manifest-version validation.

## UI Changes
N/A — server only.

## Checklist
- [x] Screenshots for UI changes *(no UI)*
- [x] Video for animation/interaction *(N/A)*

https://claude.ai/code/session_011H86UHL1RPuWBX3LBTUPsk


ccdwyer:
Superseded by #3993, which consolidates this 8-PR stack into a single PR (all 9 commits preserved and individually reviewable). Closing to keep the review surface in one place; the `wb/*` branch is retained.

PR #3735, opened 2026-07-06
fix(opencode): persist and reuse session cursor

### LINKED ISSUE #3604: [Bug]: OpenCode provider loses thread context on follow-up — t3code starts a new OpenCode session instead of resuming (no durable session binding)
### Before submitting
- [x] I searched existing issues and did not find a duplicate. Related but distinct: #2579 (connection drops) and #3601 (unpaginated history fetch / slow "pull"). Different root cause — see "Relationship to #3601".

### Area
apps/server

### Summary
On the OpenCode provider, t3code does not durably bind a thread to its OpenCode `ses_…` id, and the adapter cannot resume. When the in-memory binding is lost — most plausibly when `ProviderSessionReaper` reaps the session after ~30 min idle (e.g. after a long sub-agent turn), or on an app/server restart — the next follow-up in the **same visible thread** is sent to a **brand-new, empty OpenCode session**. t3code keeps rendering the prior conversation from its own projection DB, so the user still sees the history, but the model has no context and behaves as if a new session started.

### Steps to reproduce
1. Start a thread on the OpenCode provider; have a real exchange (ideally one that spawns sub-agents / runs a while, so the session then goes idle).
2. Let it sit idle past the reaper threshold (~30 min), or restart the app, between turns.
3. Send a follow-up message in the same thread.
4. The agent answers with no awareness of the earlier conversation that is still visible in the chat.

### Expected behavior
A follow-up in an existing thread continues the same OpenCode session with full prior context — or, if the session genuinely can't be resumed, fails with a clear explicit error rather than silently starting an empty session under the same visible thread.

### Actual behavior
The follow-up runs in a new, empty OpenCode session. No hang and no error — the model just has no context. The visible chat (t3code's own projection) and the OpenCode session the prompt is sent to are desynced.

#### Root cause (source + data evidence)

**1. The adapter always creates a new session and ignores the resume cursor** — `apps/server/src/provider/Layers/OpenCodeAdapter.ts`:

```ts
// startSession always creates; input.resumeCursor is ignored
1073: const openCodeSession = yield* runOpenCodeSdk("session.create", () =>
1074:   client.session.create({ title: `T3 Code ${input.threadId}`, ... }))
1133: openCodeSessionId: started.openCodeSession.id,   // kept in memory only
// follow-up turns target the in-memory id:
1241: context.client.session.promptAsync({ sessionID: context.openCodeSessionId, ... })
```

The returned `ProviderSession` carries no `resumeCursor`, and `startSession` never reuses one even though `ProviderService` passes a persisted cursor on recovery (`ProviderService.ts` ~400-408).

**2. Nothing durable is persisted to re-attach.** In `~/.t3/userdata/state.sqlite`:
- `provider_session_runtime.resume_cursor_json` is **NULL for every OpenCode thread**.
- `projection_thread_sessions.provider_session_id` / `provider_thread_id` are **empty** for OpenCode threads (even currently-running ones).

So there is no stored link from a t3code thread to its OpenCode `ses_…`.

**3. The reaper stops idle sessions after 30 min** — `ProviderSessionReaper.ts` (`DEFAULT_INACTIVITY_THRESHOLD_MS = 30 * 60 * 1000`); `last_seen_at` is refreshed on turn send/start, not on runtime events, so a long sub-agent turn can finish and then be reaped before the next user message.

**Direct proof from OpenCode's DB (`~/.local/share/opencode/opencode.db`).** t3code titles each session `T3 Code <threadId>`, so a correctly-resumed thread would have exactly one such session. Instead, multiple threads have **two** independent top-level (`parent_id IS NULL`) sessions for the same thread id, created ~45–90 min apart:

| T3 thread | OpenCode session | created (UTC) | msgs |
|---|---|---|---|
| a72d0ec7… | ses_0ec30151… | 14:37 | 18 |
| a72d0ec7… | ses_0ec07e26… | 15:20 | 57 |
| 5d69a12b… | ses_0ec90f05… | 12:51 | 25 |
| 5d69a12b… | ses_0ec421e3… | 14:17 | 34 |
| 279f293f… | ses_0ec8dfba… | 12:54 | 32 |
| 279f293f… | ses_0ec432a4… | 14:16 | 111 |

Each pair is one visible t3code thread split across two separate OpenCode sessions; the second does not contain the first's history (OpenCode loads context per `session_id`), so a turn that lands in the new session is contextless. (Sub-agent/Task sessions are excluded — they carry different titles like `… (@explorer subagent)`.)

**OpenCode is not at fault:** given the same `session_id`, OpenCode loads prior history for a prompt (`packages/opencode/src/session/prompt.ts` ~1092-1279; messages selected by `session_id` in `message-v2.ts`). The CLI/TUI keeps using the same session id directly, which is why it does not lose context on the same DB.

### Suggested fix
1. Make `OpenCodeAdapter` resumable: persist the OpenCode `ses_…` (e.g. as the `resumeCursor` already supported by `provider_session_runtime.resume_cursor_json`), and on `startSession` with a cursor, **reuse** that session id (validate via `session.get`) instead of `session.create`.
2. If a thread has prior state but no resumable session, surface an explicit recovery error instead of silently creating a new empty session under the same visible thread (`ProviderCommandReactor`).
3. Harden the reaper: refresh `last_seen_at` on runtime events / turn completion, and/or don't reap providers that can't currently resume.

### Impact
Major degradation — silent context loss mid-thread; the model answers confidently without the prior conversation, which can lead to wrong actions. Intermittent (tied to idle/reaper/restart), so easy to hit and confusing to diagnose.

### Version or commit
T3 Code Nightly (desktop). OpenCode server 1.17.11; @opencode-ai/sdk 1.15.13.

### Environment
macOS (Apple Silicon); provider = opencode (t3code-spawned `opencode serve`).

### Logs or stack traces
Corroborating: `provider.session.reaped { provider: 'opencode', reason: 'inactivity_threshold', idleDurationMs: ~30+ min }` in `~/.t3/userdata/logs/server-child.log`, plus the two-sessions-per-thread split above. There is no error at the moment of context loss (a fresh session is created normally), matching "no hang, just contextless."

### Relationship to #3601
#3601 (unpaginated `readThread` -> slow/hung "pull") is a **separate** potential performance concern and does **not** cause this context loss (it is not on the send path, and the symptom here is contextless execution, not a hang). Filing this separately as the real root cause.

davidvanderklay:
## What Changed

OpenCode sessions now persist a small resume cursor containing the upstream OpenCode session id.

When T3 Code recovers a provider session, the OpenCode adapter uses `session.get` with the persisted id instead of always creating a fresh upstream OpenCode session.

The race cleanup path now only aborts OpenCode sessions created by the losing local start attempt, so an adopted durable session is not aborted.

## Why

Fixes #3604.

Previously, after provider reaping or a server restart, follow-up messages could lose OpenCode conversation context because T3 Code created a new upstream OpenCode session.

This keeps the change adapter-local and avoids broader lifecycle, cwd-forking, permission reapplication, or reaper changes.

## Manual Smoke Test

Verified locally with OpenCode:

1. Asked the thread to remember `t3code-opencode-resume-smoke-3604`.
2. Confirmed the phrase was recalled.
3. Restarted the T3 Code dev server.
4. Reopened the same thread.
5. Confirmed the phrase was still recalled after restart.

## Validation

- `corepack pnpm exec vp check`
- `corepack pnpm exec vp run typecheck`
- `corepack pnpm exec vp run --filter t3 test -- apps/server/src/provider/Layers/OpenCodeAdapter.test.ts`

## UI Changes

None.

---

> [!NOTE]
> **Medium Risk**
> Touches provider session lifecycle (resume vs create, conditional remote abort, and in-memory state preservation) where mistakes could drop conversation context or leave stray upstream sessions; behavior is well covered by new adapter tests.
> 
> **Overview**
> OpenCode `startSession` now exposes a **`resumeCursor`** (`schemaVersion: 1`, upstream `sessionId`) on the returned session and honors an incoming cursor by calling **`session.get`** instead of always **`session.create`**. Invalid or malformed cursors are ignored and behave like a fresh start.
> 
> Restarting a thread with the **same** upstream session id no longer **`session.abort`**s that remote session on teardown; in-memory adapter state (pending permissions/questions, turns, active turn, etc.) is **carried over** when the cursor matches the existing binding. Concurrent `startSession` races only abort **newly created** remote sessions, not adopted resumed ones; failed **`session.get`** for a valid cursor surfaces as a process error instead of silently creating an empty session.
> 
> Tests extend the OpenCode runtime double with `session.get`, permission replies, and delays, and add coverage for cursor return, resume, restart preservation, follow-up prompts, failure paths, and race cleanup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2df4ec5c4215737a7bd61da84ebe28723df7a2e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Persist and reuse session cursor in `OpenCodeAdapter.startSession`
> - `startSession` now returns a `resumeCursor` (`{schemaVersion: 1, sessionId}`) for new sessions and accepts one on subsequent calls to resume the same upstream session via `client.session.get` instead of creating a new one.
> - When restarting with a matching cursor, in-memory state (pending permissions, message/part caches, turns, active fields) is carried forward into the new context without aborting the remote session.
> - A new `closeOpenCodeContext` effect supports closing local resources with `abortRemote: false`, used when handing off a session to a resumed context.
> - On a concurrency race, only newly created losing sessions are aborted; adopted (resumed) sessions are preserved.
> - Behavioral Change: `stopOpenCodeContext` now delegates to `closeOpenCodeContext` with `abortRemote: true`, and malformed or version-mismatched cursors are treated as absent rather than erroring.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2df4ec5.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/OpenCodeAdapter.test.ts
+++ b/apps/server/src/provider/Layers/OpenCodeAdapter.test.ts
@@ -54,6 +54,11 @@ const runtimeMock = {
   state: {
     startCalls: [] as string[],
     sessionCreateUrls: [] as string[],
+    sessionCreateDelay: null as Promise<void> | null,
+    sessionGetCalls: [] as string[],
+    sessionGetError: null as Error | null,
+    sessionGetDataById: new Map<string, { id: string }>(),
+    sessionGetDelay: null as Promise<void> | null,
     authHeaders: [] as Array<string | null>,
     abortCalls: [] as string[],
     closeCalls: [] as string[],
@@ -67,6 +72,11 @@ const runtimeMock = {
   reset() {
     this.state.startCalls.length = 0;
     this.state.sessionCreateUrls.length = 0;
+    this.state.sessionCreateDelay = null;
+    this.state.sessionGetCalls.length = 0;
+    this.state.sessionGetError = null;
+    this.state.sessionGetDataById = new Map();
+    this.state.sessionGetDelay = null;
     this.state.authHeaders.length = 0;
     this.state.abortCalls.length = 0;
     this.state.closeCalls.length = 0;
@@ -124,10 +134,25 @@ const OpenCodeRuntimeTestDouble: OpenCodeRuntimeShape = {
       session: {
         create: async () => {
           runtimeMock.state.sessionCreateUrls.push(baseUrl);
+          if (runtimeMock.state.sessionCreateDelay) {
+            await runtimeMock.state.sessionCreateDelay;
+          }
           runtimeMock.state.authHeaders.push(
             serverPassword ? `Basic ${btoa(`opencode:${serverPassword}`)}` : null,
           );
           return { data: { id: `${baseUrl}/session` } };
+        },
+        get: async ({ sessionID }: { sessionID: string }) => {
+          runtimeMock.state.sessionGetCalls.push(sessionID);
+          if (runtimeMock.state.sessionGetDelay) {
+            await runtimeMock.state.sessionGetDelay;
+          }
+          if (runtimeMock.state.sessionGetError) {
+            throw runtimeMock.state.sessionGetError;
+          }
+          return {
+            data: runtimeMock.state.sessionGetDataById.get(sessionID) ?? { id: sessionID },
+          };
         },
         abort: async ({ sessionID }: { sessionID: string }) => {
           runtimeMock.state.abortCalls.push(sessionID);
@@ -227,6 +252,11 @@ beforeEach(() => {
 const advanceTestClock = (ms: number) =>
   TestClock.adjust(`${ms} millis`).pipe(Effect.andThen(Effect.yieldNow));
 
+const waitUntil = (predicate: () => boolean): Effect.Effect<void> =>
+  Effect.suspend(() =>
+    predicate() ? Effect.void : Effect.sleep("1 millis").pipe(Effect.andThen(waitUntil(predicate))),
+  );
+
 it.layer(OpenCodeAdapterTestLayer)("OpenCodeAdapterLive", (it) => {
   it.effect("reuses a configured OpenCode server URL instead of spawning a local server", () =>
     Effect.gen(function* () {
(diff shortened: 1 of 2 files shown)


davidvanderklay:
Addressed in the latest commit. `startSession` now preserves the upstream OpenCode session when replacing a local in-memory context with a matching `resumeCursor`, so the adapter no longer aborts the same session it is about to re-adopt via `session.get`.

Explicit stop/stopAll cleanup still aborts remote sessions; this change is limited to the orchestrated same-session restart path flagged by Bugbot.

davidvanderklay:
Addressed the latest Macroscope feedback in the latest commit. Same-session re-adoption now preserves the prior OpenCode context volatile in-memory state, including pending permissions/questions, part/text tracking, completed assistant parts, turn snapshots, and active turn metadata.

That keeps outstanding permission replies and streamed message bookkeeping available after local context replacement while still limiting remote-session preservation to the matching resumeCursor path.

davidvanderklay:
Noting that this overlaps with #3617.

The intent of this PR is the narrower fix for #3604: persist/reuse the OpenCode session id and avoid silently replacing a valid persisted cursor with a fresh empty session if resume fails. I intentionally left cwd forking, stale-session fallback, and permission reapplication out of scope to keep the behavior change smaller.

Happy to close this if maintainers prefer #3617’s broader approach or a different approach altogether. Otherwise I can keep this PR focused as the minimal alternative.

PR #3752, opened 2026-07-06
[orchestrator-v2] fix(orchestrator): Surface Claude background wake turns as continuation runs

mwolson:
## Summary

Background tasks started inside a Claude turn (`run_in_background` Bash)
currently vanish from t3code when they settle: the CLI's post-settle wake
turn is silently dropped, so the displayed thread diverges from the real
native transcript, and the 30 minute idle release kills the CLI (and the
pending task) before it can even wake. This PR ingests wake turns as
first-class continuation runs and keeps the provider session pinned while
background work is pending.

Stacked on #3750 (`fix/claude-session-resume`); one commit on
`fix/claude-wake-turns` above it.

## Problem and Fix

| Problem and Why it Happened | Fix |
| --- | --- |
| When a background task settles, the Claude CLI re-invokes itself and streams a whole new turn (task notification, assistant response, result) over the still-open SDK query. `ClaudeAdapterV2.handleSdkMessage` early-returns when `activeTurn` is null, so every message of the wake turn is dropped. | Null-`activeTurn` messages now buffer per native thread when they are wake evidence (assistant/user/result, or a task notification for a tracked pending task, with or without a summary). A pending-task notification or result offers one continuation request (deduped per wake) to a new `ProviderContinuationRequests` queue; a worker (`ProviderContinuationService`) dispatches an internal `message.dispatch` (`createdBy: "agent"`, `creationSource: "provider"`, `queue_after_active`), so the wake flows through the normal run pipeline, starting immediately on an idle thread and queueing behind any active run. The continuation turn sends no prompt to the CLI and instead drains the buffered wake messages into itself (results replayed last; a result finalizes the turn). |
| `ProviderSessionManager` idle-releases the provider session after 30 minutes. Closing the SDK query kills the CLI child, and any pending background task dies with it (observed: a task notification flushed to the transcript 13ms after `query.close`, with no response). | The adapter exposes an optional `hasPendingBackgroundWork` runtime capability (pending `local_bash` tasks, non-empty wake buffer, or outstanding continuation request). `releaseIfStillIdle` defers the `idle_timeout` release while it reports true, re-arming the timer, with cumulative deferral capped by a new `maxIdlePinMs` option (default 4h) so a hung task cannot pin a session forever. Other release reasons are unchanged. |

Design notes:

- The idle-release recheck is guarded atomically: `hasPendingBackgroundWork`
  yields to the adapter between the idle check and the release, so
  `releaseEntry` now takes `onlyIfIdleGeneration` and revalidates
[...]

PR #3756, opened 2026-07-07
[orchestrator-v2] fix(orchestrator): Prevent Claude session release from hanging on idle CLI reads

mwolson:
Stacked on #3752 (which stacks on #3750); the last commit is this PR's delta.

## Summary

Releasing an idle Claude provider session could deadlock silently after
removing the in-memory entry: the CLI child process stayed alive
indefinitely, no released event was persisted, the thread projection's
provider session stayed `"ready"`, and nothing was logged. The stale CLI
keeps holding the native session file, which is the likely original
mechanism behind `Session ID ... is already in use` (#3750 works around that
symptom by forcing `--resume`; this PR fixes the leak itself).

Discovered while live-verifying #3750/#3752 against a real 30-minute idle
release: the release fired (the entry was deleted), then wedged before
`query.close`, leaving the CLI orphaned.

## Problem and Fix

| --------------------------- | --- |
| Session scope finalizers run LIFO, so the message-pump fiber (forked into the session scope by `openQuery`) is interrupted before the `closeSession` finalizer that calls `query.close`. Interruption runs Effect's `Channel.fromAsyncIterable` finalizer `Effect.promise(() => iter.return())`, and `iter` is the SDK `Query`'s raw `sdkMessages` async generator (that is what `Query[Symbol.asyncIterator]()` returns). `return()` on a generator suspended at an internal `await` (reading the next message from an idle CLI) queues behind the in-flight read forever, so the scope close never completes and `query.close` is never reached. | `claudeQueryMessages` hands the stream an iterable whose `[Symbol.asyncIterator]()` returns the `Query` itself. The SDK's `Query.return()` runs `cleanup()` first, which closes the transport, terminates the CLI child, and unblocks the pending read, so interruption completes and the remaining finalizers run. `Query.next()` already delegates to `sdkMessages.next()`, so streaming behavior is unchanged. |

## Defensive Fixes

| --------------------------- | --- |
| Any wedged session-scope finalizer parks `releaseEntry` forever with no released events and no log line, leaving the projection stuck on `"ready"`. | `releaseEntry` forks the scope close (`Effect.forkDetach({ startImmediately: true })`) and joins with a 30s timeout. On timeout it logs `provider-session-scope-close-timeout`, attaches a detached observer that logs late completion or failure, and persists the released events regardless. `startImmediately` keeps well-behaved synchronous finalizer stacks completing on the same tick, and close failures still propagate after event writes exactly as before. |

## Known limitations

- The 30s timeout is a diagnosable backstop, not a substitute for cleanup:
  if a close still wedges (a future adapter regression), the projection
  shows "stopped" while the provider process may linger; the
  `provider-session-scope-close-timeout` warning is the operational signal.
  Close failures that finish after the timeout are logged by the detached
  observer rather than surfaced as `ProviderSessionReleaseError`.
- Pre-existing race, slightly widened only in the wedged case: the release
  step revokes MCP credentials thread-wide, so a replacement provider
  session for the same thread opened during the close window could have its
  fresh credentials revoked when the old release finishes. The window
  already existed (entry removal precedes close and MCP cleanup); per-entry
  credential ownership is a candidate follow-up. With the root-cause fix in
  place the Claude close path no longer wedges, so hitting the widened
  window requires a new bug.

## Validation

- `vp check`: 0 errors; `vp run typecheck`: exit 0 (full monorepo)
- `ClaudeAdapterV2.test.ts`: 20/20, including a new regression test that
  interrupts the message stream while a read is in flight; with the wrapper
  reverted to the raw generator the test deadlocks and times out at 60s
  (verified during development)
- `ProviderSessionManager.test.ts`: 19/19, including a new test where a
  hanging session-scope finalizer no longer blocks release: released events
  persist, and the wedged finalizer is confirmed via the adapter close count
  staying 0
- `ClaudeReplayFixtures.integration.test.ts`: 4 pass / 1 skipped;
  `OrchestratorReplayFixtures.integration.test.ts -t "claude"`: 16 pass /
  53 skipped

---

> [!NOTE]
> **High Risk**
> Touches provider session lifecycle, Claude adapter turn/stream handling, and automatic continuation runs—areas where races or wedged closes could leave stale CLI processes or mis-ordered wake vs user turns.
> 
> **Overview**
> Fixes **idle Claude provider session release** deadlocking when the SDK message stream is interrupted mid-read, and adds **orchestration for Claude background-task wake turns** so buffered CLI output is ingested without orphaning processes or losing work.
> 
> **Session close:** `claudeQueryMessages` streams via the `Query` object (not the raw `sdkMessages` generator) so `Query.return()` runs transport cleanup before an in-flight read. **Provider session manager** time-boxes scope close (30s), persists released events on timeout, and defers idle release while `hasPendingBackgroundWork` is true (capped by `maxIdlePinMs`), with generation guards against stale idle decisions.
> 
> **Wake turns:** The Claude adapter buffers SDK messages when no turn is active, tracks background `local_bash` tasks, emits one `ProviderContinuationRequest` per wake, and drains the buffer on provider-authored continuation turns without re-offering the CLI. A new **ProviderContinuationService** worker dispatches internal `message.dispatch` runs (`queue_after_active`). Native session **resume** now considers persisted `providerTurnOrdinal > 1` after idle release, not only in-memory `openedNativeThreads`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84245db6d897f9b0ed51b262200ae434a71a6209. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix Claude session release hanging on idle CLI reads by introducing background wake turn buffering
> - Fixes a deadlock where session scope close could hang indefinitely if the SDK message stream was interrupted mid-read; scope close is now time-boxed with a timeout and late completion is logged.
> - Introduces `claudeQueryMessages` in [ClaudeAdapterV2.ts](https://github.com/pingdotgg/t3code/pull/3756/files#diff-688f35750a1a36e66ae76758c856df3d4f0115f93627eb17b390b3abb76afef4) to iterate the Claude Query runtime (not the raw generator), ensuring `Query.return()` runs transport cleanup before the in-flight read.
> - Adds background wake turn detection in the Claude adapter: wake output is buffered outside active turns, a single `ProviderContinuationRequest` is issued per wake event, and a provider-authored continuation turn drains buffered messages without re-prompting the CLI.
> - Adds a new [ProviderContinuationService.ts](https://github.com/pingdotgg/t3code/pull/3756/files#diff-ab3081ad4b5b3bed0f3fa277f207e2f8ad4bf8a312a47106667f5dadcd3fb9a1) worker that consumes continuation requests and dispatches internal messages via `ThreadManagementService` to trigger continuation runs.
> - Idle session release in [ProviderSessionManager.ts](https://github.com/pingdotgg/t3code/pull/3756/files#diff-74a311dd376e2e0f313b4c0ee48dab44fa71495e6c482d09f5bcebb0ac110bee) is deferred while `hasPendingBackgroundWork` is true, capped by `maxIdlePinMs` to prevent indefinite pinning; idle generation guards prevent races with newly busy sessions.
> - Risk: scope close timeout means adapter cleanup may complete after session release is recorded; late-completing finalizers are logged but not retried.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 84245db.</sup>
> 
>
>

### DIFF
--- a/apps/server/src/orchestration-v2/Adapters/ClaudeAdapterV2.test.ts
+++ b/apps/server/src/orchestration-v2/Adapters/ClaudeAdapterV2.test.ts
@@ -1,4 +1,8 @@
-import type { SDKMessage, SDKUserMessage } from "@anthropic-ai/claude-agent-sdk";
+import type {
+  Query as ClaudeQuery,
+  SDKMessage,
+  SDKUserMessage,
+} from "@anthropic-ai/claude-agent-sdk";
 import * as NodeServices from "@effect/platform-node/NodeServices";
 import {
   ChatAttachmentId,
@@ -21,11 +25,13 @@ import {
 import { assert, describe, it } from "@effect/vitest";
 import * as DateTime from "effect/DateTime";
 import * as Effect from "effect/Effect";
+import * as Exit from "effect/Exit";
 import * as FileSystem from "effect/FileSystem";
 import * as Layer from "effect/Layer";
 import * as Path from "effect/Path";
 import * as Queue from "effect/Queue";
 import * as Schema from "effect/Schema";
+import * as Scope from "effect/Scope";
 import * as Stream from "effect/Stream";
 
 import { attachmentRelativePath } from "../../attachmentStore.ts";
@@ -44,6 +50,7 @@ import {
   CLAUDE_READ_ONLY_ALLOWED_TOOLS,
   ClaudeProviderCapabilitiesV2,
   claudeMcpQueryOverrides,
+  claudeQueryMessages,
   claudeRuntimeQueryPolicyForRuntimePolicy,
   loggedClaudeQueryOptions,
   makeClaudeAdapterV2,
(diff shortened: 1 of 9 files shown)


mwolson:
Superseded by #3860, which consolidates #3750, #3752, and #3756 into a single PR (shared continuation plumbing commit + one Claude specialty commit, mirroring #3578). Content is byte-identical to this PR's tip.

PR #3794, opened 2026-07-08
Improve iPad mobile navigation and offline caching

juliusmarminge:
## Summary

- Replace Expo Router with typed native-stack navigation and adaptive iPad split-view workspace behavior.
- Refine mobile headers, toolbar/search controls, inspector routing, and empty-connection actions.
- Persist the last server configuration and unfiltered Git refs on mobile and web so offline task drafts can still select models and branches.

## Why

The mobile navigation stack needed a native, predictable structure for iPad and iPhone layouts. Separately, cached project snapshots were insufficient for offline pending tasks because models and Git branches were only available from live server requests.

## Impact

The mobile app has native-stack navigation and iPad-aware workspace surfaces. Offline users retain the last known provider/model and branch state; live server data replaces cached values after reconnecting, and environment removal clears the cache.

## Validation

- `vp check`
- `vp run typecheck`
- `vp run lint:mobile`
- `vp test packages/client-runtime/src/state/server.test.ts packages/client-runtime/src/state/vcs.test.ts`


juliusmarminge:
Superseded by #3795, which is rebased on current main and contains only the offline-cache change.

Issue #2365, opened 2026-04-27
[Feature]: Switch between Claude Code-compatible providers within a thread


leynier:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I am describing a concrete problem or use case, not just a vague idea.

### Area

apps/server

### Problem or use case

Claude Code can work with alternative Claude-compatible endpoints by setting environment variables such as `ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_API_KEY`, and the `ANTHROPIC_DEFAULT_*_MODEL` aliases.

Some coding-model subscriptions expose Claude-compatible endpoints, for example Kimi, MiniMax, and Z.ai. These can be useful as Claude Code backends, but today T3 Code does not provide a first-class way to select one of these compatible providers for a Claude thread.

The main use case is not only starting a new thread with a different endpoint. I want to continue the same T3 Code conversation and switch between official Claude Code and a Claude-compatible provider while preserving the thread history as much as possible.

### Proposed solution

Add support for Claude Code-compatible provider profiles under the existing Claude provider.

A profile would provide the environment needed to start Claude Code against a compatible endpoint:

- base URL
- auth token or API key
- model mapping for Claude aliases:
  - `haiku`
  - `sonnet`
  - `opus`

Compatible provider models should appear as Claude model options rather than as separate agents. For example:

- `Kimi · Kimi K2.6`
- `MiniMax · MiniMax M2.7`
- `Z.ai · GLM 5.1`
- `Z.ai · GLM 5 Turbo`

When the selected model belongs to a compatible provider, T3 Code should start Claude Code with the matching environment variables.

When the user switches between official Claude and a compatible provider inside an existing thread, T3 Code should preserve the Claude resume/session cursor whenever possible. Since changing providers changes environment variables, the Claude SDK session may need to be restarted internally, but this should not create a new user-visible conversation.

### Why this matters

Users with multiple coding-model subscriptions often want to choose the best backend for the current task without losing the thread context.

Example workflows:

- start a thread with official Claude Code
- switch to GLM, Kimi, or MiniMax for a specific turn
- switch back to official Claude Code if the compatible provider has a limitation
- continue using the same T3 Code conversation instead of manually copying context into a new thread

Without in-thread switching, compatible Claude endpoints are much less useful because the user has to split work across multiple conversations.

### Smallest useful scope

[...]

PR #3799, opened 2026-07-08
feat: allow switching provider/model mid-conversation

### LINKED ISSUE #3797: [Feature]: Switch provider/model mid-conversation via transcript handoff
### Before submitting

### Area

apps/server

### Problem or use case

A thread is pinned to the provider/model it started on. Once a conversation begins on Codex you cannot continue it on Claude (or vice-versa) — the picker locks and the server rejects the switch.

### Proposed solution

Allow switching provider/model mid-conversation **without** trying to resume the previous provider's session.

In #2365 the decision was to *not* enable cross-provider continuation because the old provider's session could not be resumed (API response mismatches when switching back to Anthropic). This approach avoids that entirely: on a switch it starts a **fresh** session on the target provider and replays a rendered transcript as the first turn's prelude — prior user/assistant messages **plus a structured tool/command trail** (commands, file edits, MCP/skill calls, exit codes, and output). So there is no cross-provider session continuation to break; the new provider just receives context as text. An inline "Switched from X to Y" notice marks the boundary in the timeline.

(Related to #1911, but transcript-replay into a fresh session instead of session compaction/continuation.)

### Why this matters

Different models are better at different steps (e.g. plan on one, implement on another). Today that requires starting a new thread and losing the working context.

### Smallest useful scope

Same-driver model changes already work; the missing piece is cross-driver handoff. The minimal version is: on a cross-provider switch, start a fresh target session and prepend the prior conversation (messages only) as text. The tool/command trail is an enhancement on top.

### Alternatives considered

Resuming the old provider's session across providers — which is exactly what #2365 found doesn't work. Transcript replay sidesteps it.

### Risks or tradeoffs

This is a large change (~1.1k lines) and overlaps the in-flight orchestrator-v2 work, so it may not be wanted as-is. Opening as an issue first per CONTRIBUTING to gauge interest before any review — happy to close or split. The new provider sees a rendered transcript, not native tool-call state, so cross-switch turn rollback/checkpointing doesn't span the boundary. Adds no DB migrations.

Brechard:
Closes #3797.

> **Heads up on scope:** this is a large change (~1.1k lines) and overlaps the in-flight orchestrator-v2 work. I'm opening it for visibility per the linked issue and CONTRIBUTING's "issues first" — happy to close, split, or defer if it's not wanted as-is.

## What changed

Allows switching provider/model mid-conversation, **without resuming the old provider's session** (which #2365 found doesn't work across providers).

- On a cross-provider switch, `ensureSessionForThread` starts a **fresh** session on the target provider (dropping the old resume cursor) and persists the new `modelSelection`, instead of rejecting the switch.
- `providerHandoffTranscript` renders the prior conversation — user/assistant messages **plus a structured tool/command trail** (commands, file edits, MCP/skill calls, exit codes, output) — and prepends it to the first turn on the new provider, so it inherits the actual work, not just prose.
- An inline "Switched from X to Y" activity is rendered as a timeline notice above the triggering message.
- The picker is unlocked on started threads; the in-thread block is kept only for providers that require a new thread for a model change.

## Why this differs from prior attempts

- #2365 (closed not-planned): cross-provider *continuation* was disabled because the old session couldn't be resumed. This never resumes it — it replays a transcript into a fresh session.
- #1911: similar goal (handoff summary), but session compaction/continuation; this uses transcript replay instead.

## Notes

- Adds **no DB migrations**.
- Same-driver model switches are unchanged; this only adds the cross-driver handoff path.
- Verified end-to-end: a cross-provider switch carries context, including command output and non-zero exit codes the previous assistant never narrated.
- Tradeoff: the new provider sees a rendered transcript, not native tool-call state, so per-turn rollback/checkpointing doesn't span the switch boundary.

## Before / after

pr3799-provider-AFTER-switched-notice

---

> [!NOTE]
> **High Risk**
> Changes core turn-start/session binding and provider input shaping in orchestration; incorrect handoff detection or transcript handling could mis-route turns or leak stale resume state, though behavior is heavily covered by new tests.
> 
> **Overview**
> **Mid-conversation provider/model switches** are allowed when the target instance uses a different driver or continuation key. The server no longer rejects those turns; it starts a **fresh provider session** (no old `resumeCursor`), persists the new `modelSelection`, and prepends a **`[Conversation handoff]`** transcript to the first turn on the new provider via new `renderProviderHandoffPrelude` (messages plus deduped tool/command/MCP trail, size-capped).
> 
> `ProviderCommandReactor` detects handoffs, skips in-session model-change rules for them, appends **`thread.model-changed`** activities after a successful (re)start, and resolves “current” instance from the bound session when needed. **`ProviderService`** tolerates transient binding mismatches during handoff and dedupes overlapping live sessions per thread.
> 
> The **web UI** unlocks the model picker on started threads, only blocks same-driver/same-continuation switches for `requiresNewThreadForModelChange` providers, and shows **inline “Switched from X to Y”** timeline notices from those activities.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 853c8d5e306ad1cd7489197ca025599ae0eb2448. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Allow switching provider/model mid-conversation with transcript handoff
> - When a user switches to a provider with a different driver or continuation group, the orchestrator starts a fresh session on the new provider instead of rejecting the change, prepending a `[Conversation handoff]` transcript prelude to the next turn so the new provider has prior context.
> - A `thread.model-changed` activity is appended to the thread on handoff; the UI renders this as a pill-styled notice row in the conversation timeline via a new `NoticeTimelineRow` component.
> - `listSessions` in `ProviderService` no longer dies on provider/instance binding mismatches; it returns at most one session per thread, preferring the authoritative binding match.
> - The `getStartedThreadModelChangeBlockReason` logic is relaxed so cross-instance switches are allowed unless both instances share the same driver and continuation group.
> - Resume cursors are cleared when rebinding to a different provider instance to prevent stale cursor reuse across drivers.
> - Risk: the handoff prelude is capped at 80,000 chars and will silently omit older history when truncated, which may cause the new provider to miss earlier context.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 853c8d5.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
@@ -1461,7 +1461,7 @@ describe("ProviderCommandReactor", () => {
     expect(thread?.session?.runtimeMode).toBe("full-access");
   });
 
-  it("rejects provider changes after a thread is already bound to a session provider", async () => {
+  it("hands off a thread to a different provider driver mid-conversation", async () => {
     const harness = await createHarness();
     const now = "2026-01-01T00:00:00.000Z";
 
@@ -1487,6 +1487,18 @@ describe("ProviderCommandReactor", () => {
 
     await Effect.runPromise(
       harness.engine.dispatch({
+        type: "thread.message.assistant.delta",
+        commandId: CommandId.make("cmd-assistant-provider-switch-1"),
+        threadId: ThreadId.make("thread-1"),
+        messageId: asMessageId("assistant-message-provider-switch-1"),
+        delta: "codex reply before handoff",
+        turnId: asTurnId("turn-1"),
+        createdAt: now,
+      }),
+    );
+
+    await Effect.runPromise(
+      harness.engine.dispatch({
         type: "thread.turn.start",
         commandId: CommandId.make("cmd-turn-start-provider-switch-2"),
         threadId: ThreadId.make("thread-1"),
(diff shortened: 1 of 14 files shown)


juliusmarminge:

sscarduzio:
The single most wanted feature of the whole project 😭

Issue #3666, opened 2026-07-03
[Bug]: Grok provider ACP startup times out on Linux when Grok emits unsolicited skills-reload JSON-RPC responses


thanostourik:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

1. Install and log in to Grok CLI.
2. Configure T3 Code Grok provider to use grok.
3. Start T3 Code provider discovery.
4. T3 reports: Unavailable - Grok CLI is installed but ACP startup timed out after 15000ms.

### Expected behavior

T3 Code should tolerate unrelated or unsolicited JSON-RPC responses/notifications while waiting for the response to its own request IDs, and complete Grok ACP startup.

### Actual behavior

On Linux, T3 Code fails Grok provider startup with:

Unavailable - Grok CLI is installed but ACP startup timed out after 15000ms.

The Grok CLI itself works. Running grok agent stdio directly can initialize, authenticate, and create a session successfully outside T3 Code.

During ACP startup, Grok emits unsolicited JSON-RPC response messages like:

{"jsonrpc":"2.0","id":"skills-reload","result":{"result":{"reloaded":0}}}

These messages appear before or around the normal authenticate/session startup flow. T3 Code appears to hang or time out when these messages are present.

### Impact

Blocks work completely

### Version or commit

T3 Code: AppImage 0.0.28 | Grok CLI: 0.2.82 stable

### Environment

Linux Mint 22.3

### Logs or stack traces

```shell

```

### Workaround

I created a local wrapper that proxies grok agent stdio and filters only lines containing:

"id":"skills-reload"

With that wrapper configured as the Grok binary, Grok becomes usable inside T3 Code.

Why I think this is the issue:
The wrapper does not change auth, PATH, cwd, or Grok startup. It only removes the skills-reload JSON-RPC response lines from Grok stdout. After that, T3 Code can use Grok normally.

danielcor:
I asked Grok to analyze this issue and it basically says that the WSL T3 server needs to run in a WSL Linux directory not a Windows path.

> So Grok is fine on Linux paths, but when T3 probes with /mnt/c/..., 9/10 times it will sit near or over the 15s limit. That matches the error.
> 
> T3 launches the probe with cwd: process.cwd() of its server, not your project folder — so a Windows-mounted home is enough to break Grok model discovery even if Grok works in the terminal.
> 
[...]

Issue #3580, opened 2026-06-27
[orchestrator-v2] Grok v2 steer rows vanish and runs wedge on Working after reply


mwolson:
## Summary

On orchestrator v2 with Grok, four user-visible reliability problems showed up in manual testing against the `t3code/codex-turn-mapping` stack (#2829):

1. **Steered user text vanishes** on the v2 timeline between server events.
2. **Runs stay on Working** after Grok already streamed a full reply; the next message is auto-routed as a steer instead of starting a new turn.
3. **Tool turns stop early** after an assistant preamble — the run completes before tools execute.
4. **Stop → continue wedges** after interrupting a mid-turn Grok run — the next prompt hits `task_already_running` in the zombie ACP child.

## Area

- `apps/web` (steer row visibility)
- `apps/server` (Grok ACP v2 root-turn settlement and post-interrupt process restart)

### Steer visibility

1. Start T3 Code on orchestrator v2 with Grok.
2. Send a message that triggers a long-running Grok turn.
3. While the run is in flight, steer with a short follow-up (for example `continue`).
4. Watch the timeline as `message.updated` and `turn-item.updated` events arrive.

### Hung Working / mis-steer

1. On orchestrator v2 with Grok, send a message that produces a full assistant reply.
2. Wait for the reply to finish streaming in the UI.
3. Observe whether the run indicator clears (Working → idle).
4. Send a new question intended as a fresh turn (for example `what's strategy A?`).
5. Check whether the second message is treated as steer (`restart_active`) instead of `turn_start`.

### Tool turn stops after preamble

1. On orchestrator v2 with Grok, send a message that should invoke tools (for example a file search or command).
2. Watch for an assistant preamble (for example "Running the Codex review now.").
3. Observe whether the run completes before tool execution turn items appear.

### Stop → continue zombie

1. On orchestrator v2 with Grok, start a tool turn and press Stop mid-flight.
2. Send `continue` (or any follow-up intended as a new turn).
3. Check the provider log for `task_already_running` on the recovery prompt.

## Expected behavior

- Steered user text stays visible until the committed `user_message` turn item lands on the timeline.
- After Grok finishes replying, the run completes and the next user message starts a normal turn.
- Tool turns keep running through tool execution and final assistant output before completing.
- After Stop, the next message restarts the Grok ACP child and completes normally.

## Actual behavior

| Symptom | What happens |
| --- | --- |
[...]

PR #3850, opened 2026-07-10
fix(acp): drop Grok skills-reload responses that kill the protocol

connorhet:
## Summary

Grok threads in T3 can show **no assistant output** (user messages only) on Linux when the ACP protocol loop dies mid-session.

### Root cause

Grok CLI emits unsolicited JSON-RPC **responses** like:

```json
{"jsonrpc":"2.0","id":"skills-reload","result":{"result":{"reloaded":0}}}
```

Effect's `RpcMessage.RequestId` coerces wire ids with `BigInt(id)`. Non-numeric ids throw:

```
SyntaxError: Cannot convert skills-reload to a BigInt
```

That exception tears down the ACP client protocol loop. After that, turns never complete in the UI — which looks like a silent / dead Grok chat even though auth and SuperGrok are fine.

Reproduced against Grok CLI 0.2.93 + T3 Code 0.0.28. Matches [#3666](https://github.com/pingdotgg/t3code/issues/3666).

### Fix

1. **`packages/effect-acp`**: Before forwarding orphan `Exit`/`Chunk` messages to `RpcClient`, drop those whose request ids cannot be coerced to `BigInt`. Pending extension requests still match by string id as before.
2. **`AcpSessionRuntime`**: Only interrupt a prompt RPC fiber if it is still running (`pollUnsafe() === undefined`), avoiding unnecessary re-interrupt after a clean join.

### Test plan

- [x] `vp test packages/effect-acp apps/server/src/provider/acp/XAiAcpExtension.test.ts apps/server/src/provider/acp/AcpJsonRpcConnection.test.ts` (55 tests pass)
- [x] New test: protocol survives multiple `skills-reload` responses and still completes a subsequent extension request
- [ ] Manual: T3 + Grok on Linux — new thread, send `hi`, assistant reply sticks

### Notes

- This is separate from orchestrator-v2 settlement work in #3580 / #3578 (reply shows then wedges). This patch targets the **no assistant output at all** path caused by protocol death.
- Upstream CONTRIBUTING says contributions are not actively accepted; opening this because the bug is reproducible and the fix is isolated + tested.

---

> [!NOTE]
> **Medium Risk**
> Touches core ACP wire routing and prompt fiber lifecycle; behavior change is narrow and covered by tests, but incorrect filtering could drop legitimate RpcClient traffic.
> 
> **Overview**
> Fixes **silent Grok threads** (no assistant output) when the ACP protocol loop dies after Grok emits unsolicited JSON-RPC responses with id `"skills-reload"`. Effect's RpcClient coerces ids with `BigInt()`, which throws on that string and tears down the session.
> 
> In **`effect-acp`**, adds `isRpcClientSafeRequestId` and **drops** orphan `Exit`/`Chunk` messages whose ids are not BigInt-coercible before they reach RpcClient; pending extension requests still correlate by string id. A protocol test asserts survival after multiple `skills-reload` frames and a normal extension round-trip.
> 
> In **`AcpSessionRuntime`**, prompt cleanup **only interrupts** the prompt RPC fiber when it is still running (`pollUnsafe() === undefined`), so a completed join is not re-interrupted while races (e.g. xAI `prompt_complete` fallback) can still interrupt losers to release the permit.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e807573e00087a560078f85ad7dead5a0d31c9dd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Drop unsolicited Grok `skills-reload` responses with non-numeric IDs to prevent protocol termination
> - Grok emits unsolicited JSON-RPC responses with non-numeric IDs (e.g. `skills-reload`), which cause `RpcClient` to fail on BigInt coercion and kill the protocol loop.
> - Adds `isRpcClientSafeRequestId` in [protocol.ts](https://github.com/pingdotgg/t3code/pull/3850/files#diff-c41432f9e7fbaf0a7a40c325bf318ea40df32aed6b771c8b6fa5c935e37006d2) to detect non-numeric or empty string IDs, and drops unsolicited Exit/Chunk messages with unsafe IDs instead of forwarding them to the client queue.
> - Also fixes [AcpSessionRuntime.ts](https://github.com/pingdotgg/t3code/pull/3850/files#diff-947405407dc15b91fd18625aab3132939b58753dce5a014fe17c590f2abfa5d6) to only interrupt the prompt RPC fiber after join if it is still running, avoiding redundant interrupts.
> - Behavioral Change: unsolicited messages with safe (numeric) IDs continue to be forwarded as before; only non-numeric unsolicited IDs are dropped.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e807573.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/acp/AcpSessionRuntime.ts
+++ b/apps/server/src/provider/acp/AcpSessionRuntime.ts
@@ -733,7 +733,13 @@ export const make = (
               ),
               Effect.ensuring(
                 Effect.gen(function* () {
-                  yield* Fiber.interrupt(promptRpcFiber).pipe(Effect.ignore);
+                  // Only interrupt a still-running RPC fiber. After a successful
+                  // join, re-interrupting is unnecessary; when the parent race
+                  // (xAI prompt_complete fallback) abandons a loser mid-flight
+                  // we still need to interrupt so the permit is released.
+                  if (promptRpcFiber.pollUnsafe() === undefined) {
+                    yield* Fiber.interrupt(promptRpcFiber).pipe(Effect.ignore);
+                  }
                   yield* Ref.set(activePromptFiberRef, Option.none());
                 }),
               ),
--- a/packages/effect-acp/src/protocol.test.ts
+++ b/packages/effect-acp/src/protocol.test.ts
@@ -3,6 +3,7 @@ import * as AcpError from "./errors.ts";
 import * as Effect from "effect/Effect";
 import * as Deferred from "effect/Deferred";
 import * as Fiber from "effect/Fiber";
+import * as Option from "effect/Option";
 import * as Queue from "effect/Queue";
 import * as Schema from "effect/Schema";
 import * as Stream from "effect/Stream";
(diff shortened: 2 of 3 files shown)


connorhet:
Closing for now — I opened this before verifying the fix against a local T3 + Grok run. Will reopen after confirming the UI path works.

PR #3864, opened 2026-07-10
fix(config): preserve production state paths

PixPMusic:
## Summary

- keep production server state in the established `userdata` directory
- preserve the existing Electron user-data identities so upgrades retain history and settings
- continue isolating development state under `dev` and align SSH runtime discovery

## Open question

#2829 deliberately introduced `userdata-v2` in `3b864044c`. Reverting that rename preserves installations that only have the original `userdata` state, but it can orphan state already written by v2 preview builds. This PR is held as a draft while upstream decides between reverting before release or retaining v2 with an explicit migration/import policy.

As v2 was never released, I'm leaning towards migrating v1 userdata to v2 in place with an explicit backup policy. This would break downgrades potentially--and I suppose that's exactly why the rename was introduced.

## Validation

- `vp check`
- `vp run typecheck`
- focused server, desktop, and SSH path tests (73 tests)

Addresses review feedback on #2829.


juliusmarminge:
Closing after evaluation because this is superseded by #4400 and is unsafe as a standalone migration. The path-only switch from `userdata-v2` to `userdata` can strand existing preview state; #4400 covers the same convergence with explicit legacy import/migration handling. The focused tests passed, but this branch is stale/conflicting and there is no safe reason to land it independently.

PR #3963, opened 2026-07-14
Cache editor discovery results

CoolCoderSuper:
## What Changed

I use windows and my PATH, is extremely long and as a result the editor discovery scanner would run for up to 15 seconds, which caused frequent timeouts.

I resolved it by using Effect.forEach to do the scan in parallel rather than sequentially.

## Why

Timeouts when connecting to instances, and notifications of tasks taking longer than expected.

## Checklist

- [ ] This PR is small and focused
- [ ] I explained what changed and why


CoolCoderSuper:
No longer an issue

PR #3997, opened 2026-07-15
feat: add persistent OpenAI Realtime voice panel

Winds-AI:
## Summary

- replace the Grok voice layer with OpenAI Realtime over WebRTC, with `gpt-realtime-2.1-mini` as the default and `gpt-realtime-2.1` as the stronger selectable model
- mint short-lived OpenAI Realtime client secrets on the T3 server so the long-lived API key never enters the renderer
- use WebRTC-native audio, interruption, playback truncation, and echo handling instead of the custom raw-PCM WebSocket pipeline
- add documented Realtime controls for voice, 0.25x-1.5x speech speed, input-language hint, reasoning effort, semantic-VAD eagerness, and microphone noise profile
- keep one global voice session active across task navigation and application switches
- seed only the latest completed assistant message, with paginated access to earlier messages from the origin task
- retain safe composer read/replacement tools and add a Pi-inspired atomic exact-edit tool; voice can draft but never send prompts
- retain Parallel Search and Extract as deterministic custom web tools whose results must return before OpenAI continues
- migrate voice settings and trace storage away from the xAI schema, remove old xAI voice history, and best-effort delete the old stored xAI voice key
- preserve the separate T3 Code Voice macOS identity and add the correct OpenAI microphone permission text

## Realtime behavior

- the server creates the ephemeral session with automatic responses disabled, then the renderer installs the researched prompt, tools, and user settings before listening begins
- the prompt uses labeled role, tone, reasoning, verbosity, tool, unclear-audio, task-context, and composer-editing sections from OpenAI's Realtime prompting guidance
- semantic VAD supports patient, balanced, and quick turn-taking while WebRTC handles interruptions and truncates unplayed model audio
- tool calls clear queued output as a guard, and stale tool completions cannot create a response after a user interruption
- `search_web` waits for Parallel Search; `extract_web_pages` is available for selected URLs when ranked excerpts are insufficient

## Security and migration

- OpenAI and Parallel long-lived keys remain in the selected T3 server secret store
- only the short-lived OpenAI client secret reaches the renderer
- renderer microphone permission remains audio-only and limited to the trusted T3 renderer origin
- composer exact edits require unique matches and are applied atomically with an expected-current-text check
- task context is treated as untrusted conversation data, not as instructions
- the unrelated T3 Grok harness/provider remains available; this PR removes only the xAI voice integration

## Validation

- `vp check` passes with 9 pre-existing React warnings in unrelated files
- `vp run typecheck` passes
- full server suite: 1,403 passed, 7 skipped
- full web suite: 1,292 passed
- desktop artifact tests: 27 passed
- production desktop build, arm64 DMG, and ZIP completed successfully
- installed `/Applications/T3 Code Voice.app` is ad-hoc signed and passes strict deep signature verification
- Computer Use verified the OpenAI key UI, Parallel tools, model/voice/speed/language/reasoning/VAD/noise settings, global-session copy, and empty migrated history
- a live OpenAI audio/tool session remains to be tested after an OpenAI API key is entered in Voice settings

## References

- https://developers.openai.com/api/docs/models/gpt-realtime-2.1-mini
- https://developers.openai.com/api/docs/models/gpt-realtime-2.1
- https://developers.openai.com/api/docs/guides/realtime-webrtc
- https://developers.openai.com/api/docs/guides/realtime-conversations
- https://developers.openai.com/api/docs/guides/realtime-models-prompting
- https://docs.parallel.ai/api-reference/search/search
- https://docs.parallel.ai/api-reference/extract/extract
- https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/src/core/tools/edit.ts


t3dotgg:
> [!NOTE]

Closing this PR after an automated pass over open pull requests. Duplicates the maintainer-owned voice implementation in #5213.

PR #4015, opened 2026-07-15
fix(claude): isolate capability probe from user MCP servers

### LINKED ISSUE #3909: Provider capability probe spawns a full `claude` session that loads the user's MCP servers, every ~5 minutes, from the server's cwd ($HOME) — amplified a third-party MCP-server bug into a ~46 GiB host freeze
## Environment

- t3 **0.0.29-nightly.20260705.729** (`apps/server`, npx-installed, launched via t3's own ssh-launch mechanism — the t3-generated `~/.t3/ssh-launch/<id>/run-t3.sh` — on a Linux dev machine: veLinux 2 / Debian-based, kernel 5.15.152, 32 vCPU / 62.6 GiB RAM, no swap)
- Node via mise shims (engine range `^22.16 || ^23.11 || >=24.10`)
- Server runs detached 24/7; `run-t3.sh` performs no `cd`, so `process.cwd()` = the login home directory

## Behavior observed

The provider status/refresh loop (`checkClaudeProviderStatus` → `runClaudeCommand` / `makeClaudeEnvironment`, `probeCodexAppServerProvider`) runs every **301–305 s**, continuously, whether or not any client is connected. Per cycle it spawns, among others, a Claude **capability/auth probe**:

```
claude --output-format stream-json --verbose --input-format stream-json \
  --setting-sources=user,project,local --permission-mode default --no-session-persistence
```

Two properties of this probe are problematic:

1. **`--setting-sources=user,...` loads the user's global `~/.claude.json`, including `mcpServers`** — so every probe boots every user-configured stdio MCP server (in our case serena, a Python language-server frontend), just to check provider capability. That's ≈285 cold-starts per configured MCP server per day as a side effect of health checking. (The `claude --version` probe and the `codex app-server` probe do NOT load MCP servers — only the capability probe does; source-verified in the t3 dist bundle.)
2. **cwd is inherited from the t3 server = `$HOME`** — MCP servers that key off the working directory (serena's `--project-from-cwd`) treat the entire home directory as their project.

## Impact on our host

serena v1.5.3 has a bug (reported upstream separately: oraios/serena#1683 — a pre-stdio project auto-generation scan: a recursive symlink-following `os.scandir` walk with per-entry gitignore matching, **minutes-long on large trees — ~200 s measured on our `$HOME`**) that leaves serena **blind to its client's exit until the scan completes**. Our home directory is itself a git repository, so serena adopted the entire `$HOME` as its project root; at a ~300 s probe cadence, scanners accumulated once per-scan duration exceeded the probe interval under the growing load. Result: **≈84% of capability probes were followed by a surviving ~100%-CPU serena orphan** (≈76 probes between the MCP-config change at 00:43 and the 07:09 snapshot; 66 surviving orphans, 64 of them born within 2 s of a probe).

Forensic correlation (atop process accounting, 10 s resolution): **64 of 66** surviving orphans were born within **2 seconds** of a t3 Claude probe span (6 same-second, 38 at +1 s, 20 at +2 s). Over ~21 unattended hours the fleet reached **167 orphans (peak observation) consuming ~46 GiB RSS combined**; the host (62.6 GiB usable RAM, no swap) froze and required a reboot.

While the runaway accumulation stems from serena's pre-stdio blindness, t3's probe design is what turned it into an outage on this host — an attribution inferred from the 2 s birth correlation and the source-verified spawn chain above — and it is heavyweight even when MCP servers are well-behaved.

## Suggested directions

- Probe with MCP loading disabled: `--strict-mcp-config` with an empty `--mcp-config`, or restrict `--setting-sources` so user-level `mcpServers` are not booted for a health check.
- Run probes from a dedicated benign cwd (e.g. an empty directory under `~/.t3`), not the server's inherited cwd.
- Consider caching capability results and/or pausing probes while no client is connected (the server probed all weekend with, as far as we can tell, no client attached — inferred from the client machine's power logs and persisted session state, not from t3's own logs).
- Consider `PR_SET_PDEATHSIG`/job-object semantics for any child the probe spawns, so probe children do not outlive the probe (subject to pdeathsig's direct-parent semantics and setup race).

## Evidence available

- `~/.t3/userdata/logs/server.trace.ndjson*` span counts: 92 Claude checks between 23:21:30 and 07:00:47 (≈7.7 h), i.e. every 301–305 s, all `exit: Success`
- atop raw logs correlating probe spans to orphan birth times
- The probe argv recovered from the dist bundle and matched against live process records
- Public-source confirmation at tag `v0.0.29-nightly.20260705.729`: capability probe uses `settingSources: ["user","project","local"]` and `persistSession: false` (`apps/server/src/provider/Layers/ClaudeProvider.ts:599-602`, spawned via `@anthropic-ai/claude-agent-sdk` `query()` with streaming input — matching the recovered stream-json argv); ClaudeProvider passes no `cwd` option, so the SDK child inherits the server's cwd, while the codex probe passes `cwd: process.cwd()` explicitly (`CodexProvider.ts:504`); the refresh interval is `Duration.minutes(5)` with a 5-minute probe cache TTL (`ClaudeDriver.ts:60-61`), driven by an unconditional `Effect.forever` loop with no connected-client gate (`makeManagedServerProvider.ts:141-146`); the generated `run-t3.sh` is pure exec with no `cd` (`packages/ssh/src/tunnel.ts:413ff`)

---

Related: the serena-side report of the same incident is [oraios/serena#1683](https://github.com/oraios/serena/issues/1683).

jbbottoms:
## What Changed

- Claude capability probe (`probeClaudeCapabilities`) now passes empty `mcpServers` with `strictMcpConfig: true`, so the ~5 minute refresh loop does not boot user/project/local MCP servers.
- Probe env sets `ENABLE_CLAUDEAI_MCP_SERVERS=false` to block claude.ai connectors on the same path.
- Probe runs from `~/.t3/claude-capability-probe` instead of inheriting the server process cwd (often `$HOME`).
- Added unit coverage for the probe option builder.

## Why

Fixes #3909. The capability probe loaded `settingSources: ["user","project","local"]` without MCP isolation. Every refresh spawned configured stdio MCP servers from the server cwd. A wedged MCP (e.g. serena scanning `$HOME`) accumulated orphans and froze the host (~46 GiB RSS in the report).

Real chat sessions are unchanged -- only the lightweight init probe is isolated.

## Validation

- `pnpm exec vp test run src/provider/Layers/ClaudeCapabilitiesProbe.test.ts` -> 2/2 passed

## Notes

- CONTRIBUTING: small reliability fix; no feature scope.
- Authorship: Jaret Bottoms <25535214+jbbottoms@users.noreply.github.com>

> [!NOTE]
> ### Isolate `probeClaudeCapabilities` from user MCP server configuration
> The capability probe was previously susceptible to user-configured MCP servers interfering with results. `buildClaudeCapabilitiesProbeQueryOptions` now constructs probe options with `mcpServers={}`, `strictMcpConfig=true`, and `ENABLE_CLAUDEAI_MCP_SERVERS=false`, while retaining user/project/local setting sources. Tests verify that probe invocations match these strict no-MCP expectations.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 33f8b50.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude capability detection reliability by isolating probes from user and project MCP configurations.
  * Prevented capability checks from persisting sessions or invoking unintended tools.
  * Ensured probe data uses a stable working directory and avoids repeated side effects.

* **Tests**
  * Added coverage for Claude probe configuration, environment handling, and working-directory resolution.

---

> [!NOTE]
> **Medium Risk**
> Changes only the periodic Claude init probe, but it alters how the Agent SDK is invoked (MCP isolation); misconfiguration could break provider verification without affecting normal sessions.
> 
> **Overview**
> Fixes capability health checks repeatedly spawning user/project/local MCP servers (and claude.ai connectors), which could wedge the host during the ~5 minute refresh loop.
> 
> Introduces **`buildClaudeCapabilitiesProbeQueryOptions`** so `probeClaudeCapabilities` uses a single, explicit SDK options shape: **`mcpServers: {}`** with **`strictMcpConfig: true`**, **`allowedTools: []`**, unchanged **`settingSources`** (`user`, `project`, `local`) for slash-command discovery, and probe env **`ENABLE_CLAUDEAI_MCP_SERVERS=false`** even when the caller env has it enabled. Real chat sessions are untouched—only this lightweight init probe.
> 
> Adds **`ClaudeCapabilitiesProbe.test.ts`** for the builder assertions and a fake-Claude integration test that verifies strict no-MCP CLI flags while account/slash-command data still resolves.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33f8b5090821a6235e940e916663fc766c6ca108. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/server/src/provider/Layers/ClaudeCapabilitiesProbe.test.ts
+++ b/apps/server/src/provider/Layers/ClaudeCapabilitiesProbe.test.ts
@@ -1,0 +1,138 @@
+import { ClaudeSettings } from "@t3tools/contracts";
+import * as NodeServices from "@effect/platform-node/NodeServices";
+import { assert, it } from "@effect/vitest";
+import * as Effect from "effect/Effect";
+import * as FileSystem from "effect/FileSystem";
+import * as Path from "effect/Path";
+import * as Schema from "effect/Schema";
+
+import {
+  buildClaudeCapabilitiesProbeQueryOptions,
+  CLAUDE_CAPABILITIES_PROBE_SETTING_SOURCES,
+  probeClaudeCapabilities,
+} from "./ClaudeProvider.ts";
+
+const decodeClaudeSettings = Schema.decodeSync(ClaudeSettings);
+
+it("isolates Claude capability probes without dropping workspace setting sources", () => {
+  const abortController = new AbortController();
+  const options = buildClaudeCapabilitiesProbeQueryOptions({
+    executablePath: "/usr/bin/claude",
+    abortController,
+    environment: {
+      HOME: "/home/user",
+      ENABLE_CLAUDEAI_MCP_SERVERS: "true",
+    },
+    cwd: "/workspace/project",
+  });
+
+  assert.deepEqual(options.mcpServers, {});
+  assert.equal(options.strictMcpConfig, true);
+  assert.equal(options.cwd, "/workspace/project");
+  assert.deepEqual(options.settingSources, [...CLAUDE_CAPABILITIES_PROBE_SETTING_SOURCES]);
+  assert.deepEqual(options.allowedTools, []);
+  assert.equal(options.persistSession, false);
+  assert.equal(options.pathToClaudeCodeExecutable, "/usr/bin/claude");
+  assert.equal(options.abortController, abortController);
+  assert.equal(options.env?.HOME, "/home/user");
+  assert.equal(options.env?.ENABLE_CLAUDEAI_MCP_SERVERS, "false");
+});
+
+it.layer(NodeServices.layer)("Claude capability probe SDK boundary", (it) => {
+  it.effect("serializes strict no-MCP options and still resolves account capabilities", () =>
+    Effect.gen(function* () {
+      const fs = yield* FileSystem.FileSystem;
+      const path = yield* Path.Path;
+      const tempDir = yield* fs.makeTempDirectoryScoped({ prefix: "t3-claude-probe-sdk-" });
+      const executablePath = path.join(tempDir, "fake-claude.mjs");
+      const invocationPath = path.join(tempDir, "invocation.json");
+      const workspaceCwd = path.join(tempDir, "workspace");
+      yield* fs.makeDirectory(workspaceCwd, { recursive: true });
+
+      yield* fs.writeFileString(
+        executablePath,
+        [
+          "#!/usr/bin/env node",
+          'import { existsSync, readFileSync, writeFileSync } from "node:fs";',
+          'import { createInterface } from "node:readline";',
+          "const args = process.argv.slice(2);",
+          'const mcpConfigIndex = args.indexOf("--mcp-config");',
+          "const rawMcpConfig = mcpConfigIndex >= 0 ? args[mcpConfigIndex + 1] : undefined;",
+          "let mcpConfig;",
+          "if (rawMcpConfig) {",
+          '  const contents = existsSync(rawMcpConfig) ? readFileSync(rawMcpConfig, "utf8") : rawMcpConfig;',
+          "  try { mcpConfig = JSON.parse(contents); } catch { mcpConfig = contents; }",
+          "}",
+          "writeFileSync(process.env.T3_PROBE_INVOCATION_PATH, JSON.stringify({",
+          "  args,",
+          "  cwd: process.cwd(),",
+          "  connectorEnv: process.env.ENABLE_CLAUDEAI_MCP_SERVERS,",
+          "  mcpConfig,",
+          "}));",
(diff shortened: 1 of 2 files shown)


jbbottoms:
@coderabbitai review

juliusmarminge:
> Probe runs from ~/.t3/claude-capability-probe instead of inheriting the server process cwd (often $HOME).

doesn't this mean we miss skills/commands?

PR #4031, opened 2026-07-16
fix(server): Codex missed skills defined in the current repo

D3OXY:
Skills committed to a repository never showed up in the composer's `$` picker when using Codex — only globally installed ones did. If you added a skill under your repo and reached for it in the composer, it simply wasn't there.

**This is Codex-only.** Claude already works: its provider probe scans the filesystem for skills using the instance's own `cwd`, so `<cwd>/.claude/skills` is picked up as `project` scope. Codex instead gets its skills from the app-server's `skills/list`, which the provider snapshot calls without workspace context and caches per instance — so repo-local skills are invisible to it by construction.

## What changed

`ProviderDriver` gains an optional `listSkills(cwd)`. Codex implements it by querying `skills/list` against the active workspace on demand, exposed through a new `server.listProviderSkills` RPC and a `useProviderSkills` hook that the composer reads instead of the snapshot.

The hook is provider-generic, but Codex is the only driver that implements it. Drivers that don't keep serving their snapshot, so Claude's picker returns exactly the same list it does today.

Discovery is best-effort. The query times out after ten seconds and falls back to the snapshot, and an empty result is treated as "couldn't answer" rather than "no skills" — a slow or unreachable workspace can never leave the picker empty. The snapshot fallback drops repo-scoped entries, since a cached snapshot may describe a different workspace than the one you're in.

## Testing

`ProviderRegistry` covers the driver-implements and driver-doesn't-implement paths; `server.test.ts` covers the RPC end to end. Format, lint, typecheck, and the server, contracts, client-runtime, and web suites pass against `a20923ce4`.

One unrelated pre-existing failure on this machine: `apps/web/src/terminal/ghostty/runtimeAbi.test.ts` fails to load `ghostty-vt.wasm?inline` through vite's import analysis. It fails identically on a clean checkout of `main`, and this branch touches no ghostty files.

## Notes

Rebuilt on current `main` — the previous revision had gone stale and conflicted. Same approach as before, with the composer's empty-result fallback and a stable memo identity for the skill list fixed along the way.

### DIFF
--- a/apps/server/src/auth/RpcAuthorization.ts
+++ b/apps/server/src/auth/RpcAuthorization.ts
@@ -31,6 +31,7 @@ export const RPC_REQUIRED_SCOPES = {
   [ORCHESTRATION_WS_METHODS.subscribeThread]: AuthOrchestrationReadScope,
   [WS_METHODS.serverProbe]: AuthOrchestrationReadScope,
   [WS_METHODS.serverGetConfig]: AuthOrchestrationReadScope,
+  [WS_METHODS.serverListProviderSkills]: AuthOrchestrationReadScope,
   [WS_METHODS.serverRefreshProviders]: AuthOrchestrationOperateScope,
   [WS_METHODS.serverUpdateProvider]: AuthOrchestrationOperateScope,
   [WS_METHODS.serverUpdateServer]: AuthOrchestrationOperateScope,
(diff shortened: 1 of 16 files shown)


Mnigos (changes requested):
Image

Mentioning skills itself works fine, but when message is sent, the mentioned skill is not formatted properly.

Also why not add this also for claude code?

D3OXY:
Superseded by #8528.

This PR fixes Codex by moving discovery from process.cwd() to a configured cwd. #8528 fixes the picker path itself: T3 resolves the active project or thread worktree on the server and asks the selected provider to discover skills there. It covers Codex, Claude, Cursor, Grok, and OpenCode across web and mobile.

#8528 is stacked on #8336.

Issue #4057, opened 2026-07-16
Support non-image file attachments (PDF, logs, CSV) in the chat composer


andreidrSB:
## Problem

The composer only accepts images (`image/*` gate in `ChatComposer.tsx` / `Normalizer.ts`). Dropping a PDF, log file, or CSV shows "Please attach image files only." When driving an agent remotely (desktop app or web paired to `t3 serve` on another machine), there is no way to hand the agent a file — you have to scp it to the server and type the path by hand.

Related: #2803 (mobile web attach) is adjacent but different surface.

## Proposed design (implemented, PR ready)

Images keep their exact current flow (native image content blocks per provider). Non-image files reuse the existing upload path — base64 data URL → server writes to `<stateDir>/attachments/` — and are delivered **by reference**: at the single `sendTurn` choke point in `ProviderCommandReactor`, a line like

```
[Attached file: /abs/state/attachments/<id>.pdf (report.pdf, application/pdf, 2.1 MB). Read it from disk when needed.]
```

is appended to the turn input, and the attachment list passed to the provider is filtered to images only. The agent reads the file with its own tools, so this works identically for Claude/Codex/Grok with zero adapter changes, and works for remote environments by construction (the file lands on the machine the agent runs on).

Details handled:
- New `ChatFileAttachment` contract variant (additive union member — old persisted rows decode unchanged; old clients hit the usual version-skew surface on threads containing file attachments, same class as #4035).
- Disk extensions are allowlist-gated (`SAFE_FILE_EXTENSIONS`), `.bin` fallback — keeps `.html`-class extensions off the same-origin asset route.
- Attachment GC keep-set in `ProjectionPipeline` updated (it currently skips non-image attachments, which would prune the files).
- Non-image asset requests get `Content-Disposition: attachment` (no inline render).
- Composer: same drag/drop + paste, non-images render as a name+size chip; timeline renders a download chip. Limits unchanged (10 MB, 8/message).
- Known follow-up: shipped mobile renders file attachments as image tiles by id; a small mobile render fix can follow.

Happy to open the PR if this direction works for you — it's ~16 files, images' behavior untouched, with tests across contracts/server/web.

juliusmarminge:
Closing as a duplicate of #2126, the canonical non-image attachment request. The active broad implementation is PR #4058, with narrower text-file work in PR #3927; validate remote upload/reference behavior before resolving #2126.

PR #4058, opened 2026-07-16
feat(chat): support non-image file attachments in the composer

andreidrSB:
Implements #4057.

## What

The composer currently rejects anything that isn't an image. This PR lets users drag/paste **any file** (PDF, logs, CSV, …) into the composer and have the agent read it — including on remote environments, where there was previously no way to hand the agent a file at all.

**Images are untouched**: same native image content blocks per provider, same preview tiles, same expand dialog. Limits unchanged (10 MB per attachment, 8 per message).

## How

Non-image files reuse the existing upload path (base64 data URL → server writes to `<stateDir>/attachments/`), and are delivered **by reference** at the single `sendTurn` choke point in `ProviderCommandReactor`:

```
[Attached file: /abs/.../attachments/<id>.csv (vendor-report.csv, text/csv, 40 B). Read it from disk when needed.]
```

is appended to the turn input, and the attachment list passed to the provider is filtered to images only. The agent reads the file with its own tools — **zero changes to the Claude/Codex/Grok adapters**, and remote environments work by construction (the file lands on the machine the agent runs on).

## Details

- New `ChatFileAttachment` contract variant — additive union member; existing persisted image rows and localStorage drafts decode unchanged (persisted draft `type` is optional, missing ⇒ image, no storage-version bump).
- **GC fix**: `collectThreadAttachmentRelativePaths` skipped non-image attachments while the pruning pass deletes anything not in the keep-set — without this, file attachments would be deleted from disk on the next projection pass.
- **Reactor filter is required, not optional**: Codex/Grok adapters wrap *any* attachment in an image block today; only the Claude adapter skips non-images.
- Disk extensions are allowlist-gated (`SAFE_FILE_EXTENSIONS`, `.bin` fallback) so `resolveAttachmentPathById`'s fixed-extension scan keeps working and `.html`-class extensions never land on the same-origin asset route. Non-image asset requests get `Content-Disposition: attachment` (download, never inline render).
- Prompt line sanitizes the file name (control chars/brackets stripped); the on-disk name is `<server-generated-id>.<allowlisted-ext>`, so the original filename never touches a path.
- Timeline renders file attachments as name+size chips that download via the signed asset URL; `historyBootstrap` summarizes them for provider switches.

## Screenshots

Composer with a dropped CSV staged as a chip (images keep their tiles):

composer chip

Sent message in the timeline with the file chip:

timeline chip

## Verified

- `vp check` 0 errors (the new reactor test uses `it.effect` — no new manual-runner debt), `vp run typecheck` clean repo-wide, all package test suites green: server 1419, web 1313, contracts 183, mobile 505, relay 187.
- +19 server tests / +12 contracts+web tests: Normalizer ingestion (new file), GC keep/prune for file attachments, reactor images-only filter + exact injected prompt line, AssetAccess download flag, extension inference, draft persistence compat, timeline chip rendering.
- Live E2E on a dev instance: dropped a CSV in the browser → file persisted server-side with correct id/extension → Codex's own session rollout log shows the turn input containing the exact `[Attached file: …]` line with the resolved path.

## Known version-skew / follow-ups (called out per #4035 precedent)

- Old clients fail to decode a thread snapshot that contains a `type:"file"` attachment (usual wire-evolution surface, handled by the version-mismatch banner).
- Shipped mobile renders file attachments as broken image tiles (it renders every attachment by id as an image). Small mobile render fix is a natural follow-up; kept out to keep this PR single-purpose.
- Possible v2: native document blocks per provider SDK (e.g. Claude PDF blocks) instead of path injection for specific types.

---

> [!NOTE]
> **Medium Risk**
> Touches attachment ingestion, same-origin asset serving, and the provider turn pipeline (security-sensitive), though mitigated by extension allowlists, download headers, and image-only provider payloads.
> 
> **Overview**
> Adds **`ChatFileAttachment`** alongside images so the composer can stage PDFs, logs, CSVs, and other files—not only pictures—while keeping the same size/count limits.
> 
> **Server:** uploads persist with allowlisted extensions (`inferFileExtension`, `.bin` fallback); **`ProviderCommandReactor`** appends `[Attached file: …]` paths to turn input and sends **only images** to providers; projection GC now **keeps file attachments** referenced in messages. Non-image assets get **`Content-Disposition: attachment`** on `/api/assets` to avoid inline rendering of risky types.
> 
> **Web:** drag/paste accepts any file type; timeline shows **file chips** (name + size) vs image tiles; drafts persist `type: "file"` with legacy image default when `type` is missing.
> 
> **Contracts:** additive union on `ChatAttachment` / upload payloads; tests cover decode and limits.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f54a4d4901dbc724226b9032e22546f59d76d8a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Support non-image file attachments in the chat composer and message timeline
> - Extends the composer to accept any file type via paste or drag-and-drop; file attachments render as chips with a file icon and formatted size, while images retain the existing thumbnail grid.
> - Adds `ComposerFileAttachment` and `ComposerAttachment` union types to the draft store so file and image attachments coexist in the same array, with backward-compatible hydration for legacy drafts.
> - Introduces `ChatFileAttachment` and `UploadChatFileAttachment` contract schemas, extending the `ChatAttachment` and `UploadChatAttachment` unions.
> - On the server, file attachments are persisted to disk via the normalizer, appended to the provider prompt as text references (path, name, MIME, size) rather than sent as content blocks, and included in attachment GC during thread revert.
> - Non-image attachments resolved via `AssetAccess` are marked with `download: true`, causing the HTTP asset route to set a `Content-Disposition: attachment` header.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0f54a4d.</sup>
> 
>

### DIFF
--- a/apps/server/src/assets/AssetAccess.test.ts
+++ b/apps/server/src/assets/AssetAccess.test.ts
@@ -198,6 +198,32 @@ describe("AssetAccess", () => {
       expect(yield* resolveAsset(token, "ignored.png")).toEqual({
         kind: "file",
         path: attachmentPath,
+        download: false,
+      });
+    }).pipe(Effect.provide(testLayer)),
+  );
+
+  it.effect("marks non-image attachment capabilities for download", () =>
+    Effect.gen(function* () {
+      const config = yield* ServerConfig.ServerConfig;
+      const fileSystem = yield* FileSystem.FileSystem;
+      const path = yield* Path.Path;
+      const attachmentId = "thread-1-00000000-0000-4000-8000-000000000002";
+      const attachmentPath = path.join(config.attachmentsDir, `${attachmentId}.pdf`);
+      yield* fileSystem.makeDirectory(config.attachmentsDir, { recursive: true });
+      yield* fileSystem.writeFile(attachmentPath, new Uint8Array([1, 2, 3]));
+
+      const result = yield* issueAssetUrl({
+        resource: { _tag: "attachment", attachmentId },
+      });
+      const suffix = result.relativeUrl.slice(`${ASSET_ROUTE_PREFIX}/`.length);
+      const separatorIndex = suffix.indexOf("/");
+      const token = suffix.slice(0, separatorIndex);
+
+      expect(yield* resolveAsset(token, "ignored.pdf")).toEqual({
+        kind: "file",
+        path: attachmentPath,
+        download: true,
       });
     }).pipe(Effect.provide(testLayer)),
   );
--- a/apps/server/src/assets/AssetAccess.ts
+++ b/apps/server/src/assets/AssetAccess.ts
@@ -36,6 +36,7 @@ import {
 } from "../auth/utils.ts";
 import * as ServerSecretStore from "../auth/ServerSecretStore.ts";
 import { resolveAttachmentPathById } from "../attachmentStore.ts";
+import { SAFE_IMAGE_FILE_EXTENSIONS } from "../imageMime.ts";
 import * as ServerConfig from "../config.ts";
 import * as ProjectFaviconResolver from "../project/ProjectFaviconResolver.ts";
 import * as WorkspacePaths from "../workspace/WorkspacePaths.ts";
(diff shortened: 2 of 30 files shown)


juliusmarminge:
i wonder if we should just provide the path where the agent can find the file? 

claude code has a limit on 5MB per image which we already don't honor.

a workflow i frequently use in Codex is i just attach a video and let the agent figure it out, and that video can sometimes be quite large if it's a long screen recording hehe.

jakeleventhal:
The demonstrated screenshots should mirror the UI from https://github.com/pingdotgg/t3code/pull/3927 where it uses the native file type icons

juliusmarminge:

Issue #2765, opened 2026-05-19
[Bug]: Codex Update Always Failing


DommyMM:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

Open the app, click update when it detects a newer codex version

### Expected behavior

Like Claude Code, it just runs it and updates and tells me I'm gucci

### Actual behavior

Failed to run update command npm: NotFound: ChildProcess.spawn (npm install -g @openai/codex@latest)
<img width="623" height="422" alt="Image" src="https://github.com/user-attachments/assets/6c2ecc90-221e-4cf4-a273-ffff7f0505b6" />
Seems to always happen on both my thinkpad and desktop (both W11), but me manually typing the command works

### Impact

Major degradation or frequent failure

### Environment

Windows11

### Logs or stack traces

```shell

```

### Screenshots, recordings, or supporting files

[image.png](https://github.com/user-attachments/assets/7c833379-92a6-40db-89bb-ee50439727e2)

### Workaround

Manually typing the npm install ig?

Haplois:
I have the same on Windows 11 issue too, manually running the update commands on a console also works for me. 

This looks like b/c of missing environment variables (probably PATH) during process creation. Happy to do a quick PR if needed.

niallobrien:
Same issue on OSX for me

michft:
That for me is the CLI versions need updating. in MacOs I can Update the Codex App but unless I do a 'brew update;  brew upgrade codex' I continue to get that error. Updating the codex app does not seem to effect the codex CLI

jordanbailey00:
Make sure you don't have a stale copy of codex on your machine. T3 might be calling the stale copy instead of the correct updated copy

Destreyf:
I investigated this against the current provider-maintenance path. Modern Codex releases can avoid the package-manager spawn entirely: `codex update` was added upstream in openai/codex#19933 and first appeared in the stable npm line in 0.128.0.

I have a focused change ready that:

- uses structured `codex` + `["update"]` execution for valid installed versions >= 0.128.0;
- retains the existing npm/pnpm/Bun/Vite+/Homebrew detection below that boundary or when the version is missing/malformed;
- keeps the selected capability shared between the version advisory and maintenance runner; and
- passes the configured provider-instance environment, including effective `CODEX_HOME`, to the native update process.

[...]

PR #4065, opened 2026-07-16
Use the native Codex update command for supported versions

### LINKED ISSUE #4066: [Feature]: Use native Codex updates for supported CLI versions
### Before submitting

### Area

apps/server

### Problem or use case

T3 Code currently chooses the Codex update command by inspecting the resolved installation path and then invoking npm, pnpm, Bun, Vite+, or Homebrew. Modern Codex releases can perform this installation-method detection themselves through the native `codex update` command.

This is distinct from #2765, which reports Windows-specific child-process/PATH failures when spawning the existing package-manager command. The capability-selection change here is cross-platform and version-dependent.

### Proposed solution

When the installed Codex version is valid semver and >= 0.128.0, use structured executable/argument fields for `codex` with `["update"]`.

For older versions, or when the installed version is missing or malformed, retain the existing installation-path-based package-manager selection. Keep the resolved capability shared between the displayed version advisory and the maintenance runner, and preserve the configured provider-instance environment.

### Why this matters

This delegates installation detection to Codex on versions that support it, reducing duplicated package-manager assumptions in T3 Code while preserving compatibility for older installations. It also makes update behavior more consistent across installation methods and platforms.

### Smallest useful scope

Apply the native update path only to the built-in Codex driver at the conservative stable boundary of 0.128.0. Do not change Claude, OpenCode, Cursor, Grok, or custom provider drivers.

### Alternatives considered

- Continue extending T3 Code installation-path heuristics for every package manager.
- Use the first prerelease containing the command, 0.126.0-alpha.9.

The stable 0.128.0 boundary is more conservative, and retaining the existing path below it avoids optimistic native command execution.

### Risks or tradeoffs

- Prereleases below 0.128.0 that technically support `codex update` continue using the legacy path.
- Missing or non-semver versions cannot opt into the native command.
- Capability selection must occur after the provider status probe without creating a second independent command-selection path.

### Examples or references

- Native command implementation: openai/codex@b985768dc11446c60e092af510b75e5e8863e514 / openai/codex#19933
- First stable npm release containing the command: 0.128.0
- Related but distinct Windows spawn issue: #2765

### Contribution

Destreyf:
## What Changed

- Use the configured Codex executable with structured `["update"]` arguments when the installed version is valid semver and >= 0.128.0 and the installation is npm-, Bun-, Homebrew-, or standalone-managed. Default installations continue to execute `codex update`, while custom names and absolute paths update the same executable that was probed.
- Retain the existing npm, pnpm, Bun, Vite+, and Homebrew path detection for older, missing, or malformed versions. pnpm and Vite+ retain their legacy commands at every version because upstream native detection currently classifies their JS wrappers as npm-managed.
- Store the resolved maintenance capability with each provider snapshot so `versionAdvisory.updateCommand`, `canUpdate`, and runner execution stay aligned.
- Preserve package-manager lock grouping for native updates and reserve the Codex-native lock for standalone or otherwise unclassified installations.
- Preserve the configured provider-instance environment and effective home for native updates. Standalone executables override auth-overlay shadow homes with the shared installation home derived from the already-resolved path.
- Add strict semver validation and focused coverage for the stable boundary, prereleases, legacy managers, manual-only paths, configured executables, install context, and update concurrency.

## Why

Codex added its native `codex update` command in openai/codex@b985768dc11446c60e092af510b75e5e8863e514. It first appeared in `0.126.0-alpha.9`, but `0.128.0` was the first stable npm release containing it, so this uses `0.128.0` as the conservative capability boundary.

Supported modern Codex installations can therefore perform their own installation-method detection instead of relying on T3 Code to invoke a package manager. Older or indeterminate installations, plus pnpm and Vite+ installations not currently supported by the native detector, continue using the existing behavior. Other provider drivers are unchanged.

Fixes #4066.

Related: #2765 covers Windows-specific child-process/PATH failures when spawning package-manager update commands. This PR may avoid that path for supported modern Codex installations, but it does not replace the Windows-specific fix.

## Testing

- Focused provider-maintenance suite — 98 passed
- Provider-instance, adapter-registry, and Codex-home integration suite — 11 passed
- `vp check` — passed with 9 existing unrelated web lint warnings
- `vp run typecheck` — passed

## Checklist

- [x] I included a video for animation/interaction changes (no animation or interaction changes)

> [!NOTE]
> ### Use native `codex update` command for Codex versions >= 0.128.0
> - Introduces version-gated maintenance capability selection in [`codexMaintenance.ts`](https://github.com/pingdotgg/t3code/pull/4065/files#diff-64d9f6240b5d9ff01e2cf445d89add515ba5406fa8895b1523fb60bb981dc0a1): versions >= 0.128.0 use the native `codex update` command; older versions retain legacy package-manager (pnpm/vp) updates.
> - Adds `makeCodexMaintenanceEnvironment` to set `CODEX_HOME` based on whether the executable is from a shared standalone install or an instance-specific path.
> - Changes `ServerProviderShape.maintenanceCapabilities` from a synchronous property to `getMaintenanceCapabilities`, an Effect, allowing capabilities to update when the probed version changes.
> - Extends `ProviderMaintenanceCommandAction` with an optional `env` field; `ProviderMaintenanceRunner` now passes this env (e.g. `CODEX_HOME`) to spawned update processes.
> - Fixes `parseSemver` in both [`shared/src/semver.ts`](https://github.com/pingdotgg/t3code/pull/4065/files#diff-509b00c0da909c9a18e64a8659c9e208b0a7b423de4e59f76e73abd7215f4c15) and [`providerMaintenance.ts`](https://github.com/pingdotgg/t3code/pull/4065/files#diff-e29fae218a488cafd27faaeea37f0b7c1dccd03ba4a994975736ddf89e3bde14) to strip build metadata before parsing numeric segments.
> - Behavioral Change: consumers of `ServerProviderShape` must now call `getMaintenanceCapabilities` as an Effect instead of reading a direct property.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 90b261b.</sup>
> 
>
> 

---

> [!NOTE]
> **Medium Risk**
> Changes provider self-update execution (spawn env, version-gated commands) and makes maintenance capabilities async per snapshot; mistakes could run wrong update commands or misreport update availability for Codex.
> 
> **Overview**
> Codex provider updates now prefer the probed executable’s native **`codex update`** flow when the installed version is valid semver and **≥ 0.128.0**; older, missing, or malformed versions keep the existing package-manager path detection (with **pnpm** and **vp** always on legacy commands).
> 
> **Maintenance capabilities are tied to the live snapshot:** `ServerProviderShape` exposes **`getMaintenanceCapabilities`** as an Effect instead of a fixed field, and `makeManagedServerProvider` can **`resolveMaintenanceCapabilities`** after each probe so version advisories, `canUpdate`, and the update runner use the same command. Codex wires this through new **`codexMaintenance`** helpers (version gate, optional update **`env`** including **`CODEX_HOME`** for standalone installs) and **`resolveProviderMaintenanceCapabilitiesWithPathsEffect`** for resolved binary paths.
> 
> The maintenance runner forwards optional **`env`** into child-process spawn; shared semver gains **`isValidSemverVersion`** and build-metadata stripping in **`parseSemver`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e0ae1d7dda54f942c158592e1225084bede9dc79. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/server/src/provider/codexMaintenance.ts
```diff
@@ -0,0 +1,25 @@
...
+  readonly installedVersion: string | null | undefined;
+  readonly legacyCapabilities: ProviderMaintenanceCapabilities;
+  readonly environment?: NodeJS.ProcessEnv;
+}): ProviderMaintenanceCapabilities {
+  return selectVersionGatedProviderMaintenanceCapabilities({
+    currentVersion: input.installedVersion,
+    minimumVersion: CODEX_NATIVE_UPDATE_MINIMUM_VERSION,
+    legacyCapabilities: input.legacyCapabilities,
+    nativeUpdate: {
+      executable: "codex",
```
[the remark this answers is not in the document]
Destreyf: Fixed in 744624e19. The native selector now receives `effectiveConfig.binaryPath`, which is the same configured executable used by the Codex status probe. Default installations still execute `codex` with `["update"]`; custom command names and absolute paths execute that exact binary with the same provider-instance environment. The explicit-path test now asserts `/custom/tools/codex` with `["update"]`.

apps/server/src/provider/codexMaintenance.ts
```diff
@@ -0,0 +1,26 @@
...
+export function selectCodexProviderMaintenanceCapabilities(input: {
+  readonly installedVersion: string | null | undefined;
+  readonly legacyCapabilities: ProviderMaintenanceCapabilities;
+  readonly executable: string;
+  readonly environment?: NodeJS.ProcessEnv;
+}): ProviderMaintenanceCapabilities {
+  return selectVersionGatedProviderMaintenanceCapabilities({
+    currentVersion: input.installedVersion,
+    minimumVersion: CODEX_NATIVE_UPDATE_MINIMUM_VERSION,
+    legacyCapabilities: input.legacyCapabilities,
```
NicL9923, line 57: **High — preserve pnpm/Vite+ update paths above the version gate.**

This selector switches every valid Codex >= 0.128.0 to `codex update`, including installations that `legacyCapabilities` identified as pnpm or Vite+. The upstream native updater only maps Npm, Bun, Brew, Standalone, and Other; the npm JS wrapper marks every non-Bun package launch as `CODEX_MANAGED_BY_NPM`. For pnpm/Vite+ this can therefore invoke `npm install -g` (or fail) while leaving the configured pnpm/vp binary unchanged.

Please retain the legacy capability when `legacyCapabilities.update.executable` is `pnpm` or `vp`, and add above-boundary tests for both.

Upstream references: [update-action mapping](https://github.com/openai/codex/blob/b985768dc11446c60e092af510b75e5e8863e514/codex-rs/tui/src/update_action.rs#L23-L33), [wrapper environment selection](https://github.com/openai/codex/blob/b985768dc11446c60e092af510b75e5e8863e514/codex-cli/bin/codex.js#L168-L175).

apps/server/src/provider/Drivers/CodexDriver.ts
```diff
@@ -144,10 +145,24 @@ export const CodexDriver: ProviderDriver<CodexSettings, CodexDriverEnv> = {
...
+      const legacyMaintenanceCapabilities = yield* resolveProviderMaintenanceCapabilitiesEffect(
+        UPDATE,
+        {
+          binaryPath: effectiveConfig.binaryPath,
+          env: processEnv,
+        },
+      );
+      const maintenanceEnvironment = {
+        ...processEnv,
+        ...(effectiveConfig.homePath ? { CODEX_HOME: effectiveConfig.homePath } : {}),
```
NicL9923: **High — do not use the auth-overlay `CODEX_HOME` for standalone update detection.**

`effectiveConfig.homePath` becomes `shadowHomePath` in auth-overlay mode, and this passes that value to `codex update`. Upstream standalone detection only succeeds when the current executable lives under `${CODEX_HOME}/packages/standalone/releases`. If Codex is installed under the shared/default home while this instance uses a shadow or custom auth home, detection becomes `Other` and the update command aborts.

Please derive the maintenance `CODEX_HOME` from the installation location, or omit the override unless the executable is installed beneath that home. A shadow-home standalone test would cover the regression.

Upstream reference: [standalone install-context detection](https://github.com/openai/codex/blob/b985768dc11446c60e092af510b75e5e8863e514/codex-rs/install-context/src/lib.rs#L128-L140).

apps/server/src/provider/codexMaintenance.ts
```diff
@@ -0,0 +1,26 @@
...
+    nativeUpdate: {
+      executable: input.executable,
+      args: ["update"],
+      lockKey: "codex-native",
```
NicL9923: **Medium — retain package-manager lock grouping.**

Hard-coding `codex-native` lets a native Codex update run concurrently with another provider update holding `npm-global`, `bun-global`, or `homebrew`, even though upstream `codex update` immediately delegates to one of those same shared package managers. The existing coordinator relies on `lockKey` to serialize those mutations.

Please reuse `legacyCapabilities.update?.lockKey` when the underlying manager is known, with `codex-native` reserved for standalone or otherwise independent installs. Concurrency coverage with another npm-managed provider would pin this down.

NicL9923 (commented):
Three install-context and update-coordination issues below. The capability/snapshot state plumbing and validation otherwise look solid.

apps/server/src/provider/codexMaintenance.ts
Destreyf: Fixed in dba59abc1. Qualifying Codex versions now retain the legacy capability when its update executable is `pnpm` or `vp`, so neither path is misclassified as npm by the upstream wrapper. Added above-boundary regression cases for both managers, including their original args and lock keys.

apps/server/src/provider/Drivers/CodexDriver.ts
Destreyf: Fixed in dba59abc1. The maintenance resolver now exposes the resolved and real command paths from its existing lookup. `CODEX_HOME` is overridden with the shared install home only when that real executable is under `<sharedHome>/packages/standalone/releases/`; auth-overlay shadow homes are no longer passed unconditionally. Added a shadow-home standalone regression test and a non-standalone preservation test.

apps/server/src/provider/codexMaintenance.ts
Destreyf: Fixed in dba59abc1. Native Codex actions now reuse `legacyCapabilities.update.lockKey` for known npm/Bun/Homebrew installations and reserve `codex-native` for manual/standalone contexts. The runner concurrency test now specifically holds `npm-global` with `codex update` and verifies an npm-managed OpenCode update remains queued.

Dropje97:
Just added a second, independent reproduction of the underlying bug in #5629 — same standalone-installer layout, but with npm's global prefix set to `~/.local`, the `npm install -g` fallback collides with the installer's own symlink and exits 1 (`EEXIST`) instead of silently installing a shadow copy. So on that configuration the in-app update isn't just cosmetically stuck, it fails outright with no path forward from the UI.

Mentioning it here mostly as a data point that the `nativeUpdate` path this PR adds fixes more than the "still needs an update" loop.

@Destreyf — this is currently conflicting against `main` and has been idle since late July. Any appetite to rebase it? Happy to test a rebased branch against a standalone install on Linux if that helps.

Destreyf:
@Dropje97 should be rebased appropriately now.

Dropje97:
Tested `e0ae1d7` on Linux against a standalone install.

```
~/.local/bin/codex -> ~/.codex/packages/standalone/releases/0.151.0-x86_64-unknown-linux-musl/bin/codex
codex-cli 0.151.0, binaryPath left at the default "codex", CODEX_HOME unset
```

**Before this PR** the advisory resolves to the npm command, which exits 1 on this machine — my npm global prefix is `~/.local`, the same bin dir the standalone installer owns, so npm's bin-link step hits `EEXIST` (this is #5629). Straight out of the provider cache on my running instance:

```json
"status": "behind_latest", "currentVersion": "0.151.0", "latestVersion": "0.152.0",
"updateCommand": "npm install -g --allow-scripts=@openai/codex @openai/codex@latest"
```

**With this PR**, driving the real resolution code against my actual filesystem and the running server's PATH:

```
resolvedCommandPath = ~/.local/bin/codex
realCommandPath     = ~/.codex/packages/standalone/releases/0.151.0-.../bin/codex
legacy              = npm install -g --allow-scripts=@openai/codex @openai/codex@latest
selected            = codex ["update"], lockKey "npm-global", CODEX_HOME=~/.codex
advisory            = updateCommand "codex update", canUpdate true
```

The version gate lets 0.151.0 through, the standalone `CODEX_HOME` override fires, and the UI would show the native command instead of the broken npm one.

Test runs on the rebased branch: `apps/server/src/provider` 740 passed / 8 skipped / 0 failed, `packages/shared/src/semver.test.ts` 12 passed. Rebase looks clean from here — merge-base is exactly current `main`, no drift.

Two minor observations, neither a blocker:

1. This standalone install inherits `lockKey: "npm-global"` rather than `codex-native`, because the legacy layer classifies a bare `binaryPath` as npm. It over-serializes, so it is safe, and it looks deliberate and test-covered — just noting it reads differently from the literal wording of the earlier review point about lock grouping.
2. A standalone install located outside the configured home would not get the `CODEX_HOME` override, and would presumably be classified as "Other" by upstream's own detection. Not a regression, but uncovered. I am less confident about this one — I could not exercise that layout here.

**Update — the native command was run end to end, and it works.**

I ran it exactly as the runner would spawn it:

```
$ CODEX_HOME=~/.codex codex update
Updating Codex via `sh -c 'curl -fsSL https://chatgpt.com/codex/install.sh | CODEX_NON_INTERACTIVE=1 sh'`...
==> Updating Codex CLI from 0.151.0 to 0.152.0
==> Detected platform: Linux (x64)
==> Installing standalone package to ~/.codex/packages/standalone/releases/0.152.0-x86_64-unknown-linux-musl
==> ~/.local/bin is already on PATH
Codex CLI 0.152.0 installed successfully.
exit 0
```

Post-conditions checked:

- `codex --version` → `codex-cli 0.152.0` (was 0.151.0)
- `~/.codex/packages/standalone/current` correctly repointed to `releases/0.152.0-...`, symlink chain via `~/.local/bin/codex` intact
- **no npm copy created** — `~/.local/lib/node_modules/` still contains no `@openai/codex`

That last point is the whole difference. Upstream's own install-context detection correctly identified the standalone layout and updated in place, where the legacy npm path either aborts with `EEXIST` (my prefix) or silently creates a second, unused install (the #5629 reporter's prefix).

So on this configuration the PR takes the provider from "update is impossible from the UI" to "update succeeds", verified end to end.

juliusmarminge:
Superseded by #9325 (merged), which gates provider updates on installer ownership of the resolved binary and uses `codex update` for standalone installs — covering #5629 / #4066.

PR #4075, opened 2026-07-17
fix(server): stabilize non-repository Git diagnostics

EricTsai83:
## What Changed

- Use `LC_ALL=C` only for the two Git commands whose non-repository control flow parses an English diagnostic.
- Add an integration test that verifies general Git subprocesses continue to preserve a caller-provided locale.

## Why

T3 Code identifies a non-repository result by matching Git's English diagnostic text. Git localizes that diagnostic through gettext, while the subprocess previously inherited the host locale. On systems with Git translations installed, a normal non-repository status could therefore be misclassified as a `GitCommandError`.

Using the C locale at these two call sites makes the parsed diagnostic deterministic without changing locale behavior for unrelated Git operations. `LC_ALL` is scoped to the commands instead of using only `LC_MESSAGES` so the fix also works when the host already defines `LC_ALL`, which takes precedence over individual locale categories.

## Validation

- `vp check`
- `vp run typecheck`
- `vp test apps/server/src/vcs/GitVcsDriverCore.test.ts`

## Checklist

- [x] No UI changes

### DIFF
--- a/apps/server/src/vcs/GitVcsDriverCore.test.ts
+++ b/apps/server/src/vcs/GitVcsDriverCore.test.ts
@@ -78,6 +78,22 @@ const initRepoWithCommit = (
   });
 
 it.layer(TestLayer)("GitVcsDriver core integration", (it) => {
+  describe("process environment", () => {
+    it.effect("preserves the caller locale for general Git subprocesses", () =>
+      Effect.gen(function* () {
+        const cwd = yield* makeTmpDir();
+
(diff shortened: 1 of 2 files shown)


EricTsai83:
Superseded by #4077, which uses the replacement `fix/git-non-repository-locale` branch and centralizes stable diagnostic execution across all three affected command paths. Closing this draft so review stays on the complete replacement.

PR #4129, opened 2026-07-18
Refresh turn checkpoints at completion

akshaylingamaneni:
## Issue

Changed-files checkpoints can be captured from a `turn.diff.updated` event while the provider turn is still running. Edits made after that capture are absent from the turn's checkpoint and can first appear in the next turn's diff.

## Root cause

`ProviderRuntimeIngestion` emits a placeholder checkpoint for the diff update, and `CheckpointReactor` replaces it with a real Git checkpoint immediately. When the final `turn.completed` event arrives, the reactor exits as soon as it finds that real checkpoint, leaving the early filesystem state attached to the turn.

## Fix

Successful or failed turn completion now refreshes an older checkpoint for the same turn while preserving its turn count and Git ref. A real or placeholder checkpoint with the same or newer completion timestamp remains unchanged, keeping repeated, delayed, and stale completion events idempotent. Cancelled and interrupted completions preserve an existing real checkpoint so its projected metadata stays synchronized with the Git ref.

The regression covers an early diff checkpoint, additional filesystem edits before completion, final checkpoint refresh, stable turn count, duplicate completion handling, cancelled or interrupted completion events, and an older completion arriving before a newer placeholder is captured.

## Validation

- `pnpm exec vp test apps/server/src/orchestration/Layers/CheckpointReactor.test.ts` — 17 tests
- `pnpm exec vp run typecheck`
- `pnpm exec vp check --no-lint apps/server/src/orchestration/Layers/CheckpointReactor.ts apps/server/src/orchestration/Layers/CheckpointReactor.test.ts`
- `git diff --check`
- `pnpm exec vp check` currently reports existing formatting drift in `apps/web/src/index.css` on current `main`; scoped lint also reports the existing manual Effect runtime call later in `CheckpointReactor.test.ts`.

## Checklist

- [x] Screenshots for UI changes *(N/A — server only)*
- [x] Video for animation/interaction changes *(N/A — server only)*

> [!NOTE]
> ### Refresh turn checkpoints on completion while preserving newer or non-missing checkpoints
> - The `turn.completed` handler in [CheckpointReactor.ts](https://github.com/pingdotgg/t3code/pull/4129/files#diff-79dc41615f79575ae1404bd102df915797fb24ba34989557f80e229d305521fb) now skips capture if an existing checkpoint has a `completedAt` timestamp equal to or newer than the incoming event, preventing redundant overwrites of early ready checkpoints.
> - If a non-missing checkpoint already exists for the turn and the incoming completion resolves to `missing` (e.g. cancelled or interrupted), the existing checkpoint is preserved.
> - When capturing, the handler reuses the existing checkpoint's `checkpointTurnCount` rather than always incrementing from the max.
> - Tests cover: refreshing an early diff checkpoint on completion, preserving a newer placeholder when an older completion arrives, and preserving a ready checkpoint when completion is cancelled or interrupted.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 250f3e3.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/CheckpointReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/CheckpointReactor.test.ts
@@ -420,6 +420,44 @@ describe("CheckpointReactor", () => {
     };
   }
 
+  async function createEarlyCheckpointHarness() {
+    const harness = await createHarness({ seedFilesystemCheckpoints: false });
+    const threadId = ThreadId.make("thread-1");
+    const turnId = asTurnId("turn-1");
+    const checkpointRef = checkpointRefForThreadTurn(threadId, 1);
+    const capturedAt = "2026-01-01T00:01:00.000Z";
+
+    harness.provider.emit({
+      type: "turn.started",
+      eventId: EventId.make("evt-turn-started-early-checkpoint"),
+      provider: ProviderDriverKind.make("codex"),
+      createdAt: "2026-01-01T00:00:00.000Z",
+      threadId,
+      turnId,
+    });
+    await waitForGitRefExists(harness.cwd, checkpointRefForThreadTurn(threadId, 0));
+
+    NodeFS.writeFileSync(NodePath.join(harness.cwd, "README.md"), "v2\n", "utf8");
+    await Effect.runPromise(
+      harness.engine.dispatch({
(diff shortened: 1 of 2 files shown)


juliusmarminge:

PR #4136, opened 2026-07-18
fix(shared): fix deepMerge data-corruption bugs and add tests

kridaydave:
Fixes two deepMerge bugs: it destroyed Date/Map/class-instance values (prototype loss) and silently discarded the entire base on a primitive patch.

- Only recurse into plain objects now; Date/Map/instances are overwritten whole.
- Throw on a primitive patch instead of dropping the base (null still allowed).
- Added regression tests in Struct.test.ts and Struct.edgeCases.test.ts.

All gates green (typecheck, check, test).

---

> [!NOTE]
> **Medium Risk**
> Shared utility behavior change: callers that passed a non-null primitive or array as the top-level patch will now throw instead of getting a silent wrong result; merge semantics for non-plain nested values also change.
> 
> **Overview**
> **`deepMerge` in `Struct.ts` now recurses only when both sides are plain objects** (`Object.prototype`), so `Date`, `Map`, class instances, arrays, and functions are replaced wholesale instead of being walked like nested records (which broke `Date` prototypes and similar values).
> 
> **Invalid top-level patches throw** instead of silently returning a primitive and dropping the base object; `null` at the top level is still allowed.
> 
> New **`Struct.test.ts`** and **`Struct.edgeCases.test.ts`** lock in normal merge behavior and document remaining quirks (array overwrite, `null` not skipped, shallow copy of nested refs).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3405a012d347f52cc7cf79dc7d9489c16c6bee3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix data-corruption bugs in `deepMerge` and add test coverage
> - Introduces `isPlainObject` in [Struct.ts](https://github.com/pingdotgg/t3code/pull/4136/files#diff-0192f560ae74187af1e3e9c733f21dc7a33bb96714dd7ed43c5e222e6ebc9b4b) to restrict recursion to plain objects; dates, maps, and class instances are now overwritten rather than recursively merged.
> - `deepMerge` now throws when the top-level patch is a non-null primitive (e.g. number, array, function) instead of silently returning it.
> - Adds primary tests in [Struct.test.ts](https://github.com/pingdotgg/t3code/pull/4136/files#diff-0c862f13fde5986e3cd427b33c593060dc2d950c9cc0ba10768510682341a625) and edge-case tests in [Struct.edgeCases.test.ts](https://github.com/pingdotgg/t3code/pull/4136/files#diff-8d79be20639d3dbd313b251f98b5958ef955267fc22f47eb7c9ef5335dcf855e) documenting both corrected and remaining lossy behaviors.
> - Behavioral Change: callers passing a primitive or array as the top-level patch will now receive a thrown error instead of a silent no-op overwrite.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3405a01.</sup>
> 
>
>


t3dotgg:
> [!NOTE]
>

Closing because this change does not address a demonstrated settings failure. Current callers merge schema-validated settings objects, not class instances or primitive patches. The new shared throw changes behavior for inputs outside that settings path. A failing case through a supported settings path would justify a focused fix.

PR #4143, opened 2026-07-19
Repair legacy shell summary migrations

salemkode:
## What changed

- Make migration 25 tolerate databases where `pending_approval_count` is not present yet.
- Add migration 34 to re-run the idempotent shell-summary schema, backfill, and cleanup migrations.
- Leave migration ID 33 retired because `ProjectionThreadsSettled` shipped in a nightly before it was reverted.
- Add regression coverage for databases where legacy migrations already occupy IDs 23 and 24.

## Why

Older nightly databases can record IDs 23 and 24 as `ProjectionThreadsAssociatedWorktreeRef` and `ProjectionThreadsArchivedAt`. The current migration registry reuses those IDs for the projection thread shell-summary migrations, so the migrator skips them and then migration 25 fails with `no such column: pending_approval_count`.

That backend failure causes the desktop process to continuously restart its child server without ever creating the main window. I reproduced this on macOS 27.0 arm64 with the current nightly.

## Impact

Affected existing installations self-repair on their next startup without resetting user data. Fresh databases remain unchanged apart from recording migration 34.

## Validation

- Focused migration regression tests pass.
- `vp check` passes with zero errors (only existing unrelated warnings).
- `vp run typecheck` passes for all 15 packages.
- `vp run --filter t3 build` succeeds.
- A copy of the affected 126 MB database successfully ran migrations 25-32 and 34 with the rebuilt server.
- The desktop app then opened on macOS 27.0 arm64 and rendered the existing projects and threads from that migrated database.

> [!NOTE]
> ### Repair legacy shell summary migrations by adding migration 34
> - Adds [034_RepairProjectionThreadShellSummary.ts](https://github.com/pingdotgg/t3code/pull/4143/files#diff-0887ae3cf70ec8db63b74f703414dc75ed1a94f7b3096dd52dfde87dc3ce5906) (migration id 34) that re-executes the logic from migrations 23, 24, and 25 under a new id, repairing `projection_threads` summary fields for databases that missed those migrations.
> - Registers migration 34 in [Migrations.ts](https://github.com/pingdotgg/t3code/pull/4143/files#diff-87557f6881a7a3dd24fa7e31f895b009cc6ede9a3571779088a10c5093110707), skipping id 33 which was shipped and reverted.
> - Guards [025_CleanupInvalidProjectionPendingApprovals.ts](https://github.com/pingdotgg/t3code/pull/4143/files#diff-553b6c1031b5ca14f35533ec7a6fe621673a9449cce6a7cdbc50f4cf6add67e6) with a column existence check so it skips the `pending_approval_count` update if the column is absent.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a659fb1.</sup>
> 
>

### DIFF
--- a/apps/server/src/persistence/Migrations.ts
+++ b/apps/server/src/persistence/Migrations.ts
@@ -45,6 +45,7 @@ import Migration0029 from "./Migrations/029_ProjectionThreadDetailOrderingIndexe
 import Migration0030 from "./Migrations/030_ProjectionThreadShellArchiveIndexes.ts";
 import Migration0031 from "./Migrations/031_AuthAuthorizationScopes.ts";
 import Migration0032 from "./Migrations/032_AuthPairingProofKeyThumbprint.ts";
+import Migration0034 from "./Migrations/034_RepairProjectionThreadShellSummary.ts";
 
 /**
  * Migration loader with all migrations defined inline.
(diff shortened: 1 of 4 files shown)


juliusmarminge:

PR #4039, opened 2026-07-16
Add scoped SSH preview port forwarding

juliusmarminge:
## Summary

- add an Effect-based environment port router with scoped route ownership
- acquire ref-counted, loopback-only SSH port forwards through the desktop bridge
- route preview automation, discovered ports, and manual navigation consistently
- keep T3 Connect forwarding explicitly unsupported for now

## Verification

- `vp check`
- `vp run typecheck`
- 48 focused SSH, IPC, client-runtime, and preview tests
- React Doctor changed-file scan (no actionable diagnostics)
- the orchestration-timeout test file from the all-repo parallel run passes 39/39 standalone

---

> [!NOTE]
> **Medium Risk**
> Cross-cutting changes to SSH tunnel lifecycle and preview navigation; lease or scope bugs could leak forwards or break remote previews, though auth and relay paths are largely unchanged.
> 
> **Overview**
> Adds **scoped SSH preview port forwarding** so localhost/dev-server URLs on SSH environments resolve through ref-counted loopback tunnels instead of only rewriting hosts for direct private-network access.
> 
> The stack wires **desktop IPC** (`acquireSshPortForward` / `releaseSshPortForward`) through `SshEnvironmentManager` (shared forwards per remote port, TCP readiness, cleanup on disconnect), exposes **`forwardPort`** on `SshEnvironmentGateway` (desktop bridge with acquire/release; mobile blocked), and introduces **`EnvironmentPortRouter`** in client-runtime for direct vs private-network vs **`ssh-forward`** resolution (relay/T3 Connect ports still rejected).
> 
> **Web preview** replaces synchronous `browserTargetResolver` URL mapping with async **route acquisition** (`commit`/`release` per tab, SSH scopes held until tab close). Preview automation, manual navigation, discovered ports, and session close all use this lifecycle. Contracts add **`ssh-forward`** to `PreviewUrlResolution`; **`NetService.hasListenerOnHost`** supports forward readiness probes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5f5c40fdc3f82a828a75f5852664ffa85ae9b35. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add scoped SSH preview port forwarding with reference-counted lease management
[...]

PR #4145, opened 2026-07-19
fix: harden SSH tunnel lifecycle

### LINKED ISSUE #4144: SSH reconnects can accumulate stale local-forward processes
## Summary

T3 Code 0.0.28 can accumulate stale SSH local-forward processes when a saved SSH environment repeatedly reconnects across sleep, network transitions, or temporarily slow readiness responses.

In a redacted production observation, one desktop instance owned 11 forwards for the same remote backend: 10 still answered and one was dead. The backend's worker-local readiness response was approximately 1 ms, while the same response through existing forwards intermittently took 2.6–5 seconds. The backend remained healthy and externally owned throughout.

## Likely lifecycle race

`packages/ssh/src/tunnel.ts` currently combines:

- a 1-second per-request readiness probe;
- a 2-second readiness window when deciding whether to reuse an existing tunnel;
- replacement when that short check fails;
- asynchronous child finalization and per-scope cleanup.

This makes a temporarily slow but valid tunnel look stale and allows repeated callers or connection scopes to replace it before the prior child has been fully cleaned up. An externally managed backend is also passed through the generic remote-stop finalizer path, even though the remote script currently avoids killing that server kind.

## Expected behavior

- One authoritative tunnel generation per connection key.
- Concurrent ensure/reconnect requests converge on that generation.
- Replaced children are always reaped on replacement and manager shutdown.
- Readiness tolerates ordinary multi-second latency without treating one slow response as stale.
- Disconnecting an externally owned backend never requests remote shutdown.
- Repeated reconnect cycles keep both process and listener counts constant.

## Proposed coverage

I have a small patch prepared that:

- raises the per-request readiness timeout to 5 seconds;
- gives reuse checks a 15-second bounded readiness window;
- explicitly skips remote stop for `serverKind: external`;
- adds a 20-way concurrent ensure test proving one child is created;
- adds a 100-cycle reconnect stress test proving the active tunnel count never exceeds one and every child is reaped.

The focused SSH package suite passes (27 tests), package typecheck passes, and `vp check` passes with only the repository's existing unrelated warnings.

No credentials, prompts, thread titles, repository content, hostnames, or Tailnet identifiers are included in this report.

NotReliable:
## Summary

- tolerate ordinary multi-second latency when probing and reusing an SSH forward
- never request remote shutdown for an externally owned backend
- cover concurrent ensure calls and 100 reconnect cycles with constant child counts

Fixes #4144.

## Why

The previous one-second request timeout and two-second reuse window could classify a temporarily slow forward as stale. Repeated callers would then replace the forward, increasing the chance that old child processes outlived the connection that created them.

This keeps the existing pending-map serialization, but gives a reusable tunnel a realistic bounded readiness window and makes external-backend ownership explicit in finalization.

## Verification

- `pnpm --filter @t3tools/ssh test` — 27 passed
- `pnpm exec vp run --filter @t3tools/ssh typecheck` — passed
- `pnpm exec vp check` — passed (10 pre-existing unrelated React warnings)

The repository-wide `vp run typecheck` planner was also attempted, but this filtered checkout did not have the unrelated Marketing workspace's `astro` executable linked. The affected SSH package typecheck above is clean.

> [!NOTE]
> ### Harden SSH tunnel lifecycle to prevent duplicate tunnels and skip remote stop for external backends
> - `ensureEnvironment` now waits up to 15s (was 2s) for an existing tunnel's HTTP endpoint before deciding not to reuse it, preventing spurious duplicate tunnel creation under concurrency.
> - The tunnel finalizer skips calling `stopRemoteServer` when `remoteServerKind` is `'external'`, only terminating the local tunnel process.
> - `SSH_READY_PROBE_TIMEOUT_MS` is increased from 1,000ms to 5,000ms; `SSH_EXISTING_TUNNEL_READY_TIMEOUT_MS` (15,000ms) is introduced as a distinct constant.
> - New tests in [tunnel.test.ts](https://github.com/pingdotgg/t3code/pull/4145/files#diff-76d3724ec8959625b119aff8272a5d733f73551706a487e303ddbd13e10a06ad) validate single-tunnel concurrency and stable tunnel counts across 100 reconnect cycles.
> - Behavioral Change: external backends no longer have their remote server stopped when a tunnel is closed.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 39db8ec. (Automatic summaries will resume when PR exits draft mode or review begins).</sup>
> 
>
>

### DIFF
--- a/packages/ssh/src/tunnel.test.ts
+++ b/packages/ssh/src/tunnel.test.ts
@@ -1,6 +1,8 @@
 import { assert, describe, it } from "@effect/vitest";
+import * as NodeHttpClient from "@effect/platform-node/NodeHttpClient";
 import * as NodeServices from "@effect/platform-node/NodeServices";
 import * as NetService from "@t3tools/shared/Net";
+import * as Deferred from "effect/Deferred";
 import * as Duration from "effect/Duration";
 import * as Effect from "effect/Effect";
 import * as Fiber from "effect/Fiber";
@@ -76,6 +78,25 @@ const testHttpClient = HttpClient.make((request) =>
 );
 
 const hangingHttpClient = HttpClient.make(() => Effect.never);
+
+const DELAYED_HTTP_FORWARD_SCRIPT = `
+const http = require("node:http");
+const port = Number(process.argv[1]);
+const responseDelayMs = Number(process.argv[2]);
+const server = http.createServer((_request, response) => {
+  setTimeout(() => {
+    response.writeHead(200, { "content-type": "text/plain" });
+    response.end("ready");
+  }, responseDelayMs);
+});
+const shutdown = () => {
+  server.closeAllConnections?.();
+  server.close(() => process.exit(0));
+};
+process.on("SIGINT", shutdown);
+process.on("SIGTERM", shutdown);
+server.listen(port, "127.0.0.1", () => process.stdout.write("listening\\n"));
+`;
 
 const testNetService = NetService.NetService.of({
   canListenOnHost: () => Effect.succeed(true),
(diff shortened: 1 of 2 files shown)


t3dotgg:
> [!NOTE]

The same SSH tunnel lifecycle code is under active review in the open [#4039](https://github.com/pingdotgg/t3code/pull/4039) and its open follow-up [#7639](https://github.com/pingdotgg/t3code/pull/7639). We are keeping that work together instead of maintaining a separate implementation here.

PR #4221, opened 2026-07-21
fix(server): bound editor discovery latency

### LINKED ISSUE #4210: [Bug]: Provider page stays on "Checking provider status" because server.getConfig blocks on editor discovery
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Start T3 Code Alpha 0.0.28 on Windows with the backend ready.
2. Use a normal Windows PATH containing multiple provider CLI and editor locations. My PATH had 53 entries.
3. Open the Providers page.
4. Observe that every provider remains on "Checking provider status" indefinitely, even after waiting for hours.
5. Inspect `server.trace.ndjson` and the provider cache files under `%USERPROFILE%\.t3\caches`.

### Expected behavior

Provider statuses should render promptly and independently. Optional external-editor detection should not block `server.getConfig`. Failed or slow editor checks should be time-limited, cached, and returned as partial results without preventing provider cards from resolving.

### Actual behavior

`server.getConfig` calls external editor discovery. In 0.0.28, approximately 20 possible editor commands are checked sequentially across the inherited PATH. On this machine the request took between roughly 15 and 99 seconds. The UI interrupted/cancelled the request after approximately 15 seconds, so the provider cards stayed in their loading state even though provider cache files were eventually updated.

This was isolated by locally changing the editor checks to run concurrently with a 3-second timeout per editor. After that change, the same `server.getConfig` RPC completed successfully in 3.158 seconds and the provider UI received completed statuses.

The completed provider cache states after the workaround were: Claude ready/authenticated (7 models), Codex ready/authenticated (6), OpenCode ready/authenticated (461), Cursor authenticated with a model-discovery warning (36), and Grok error due to the separately reported ACP issue.

### Impact

### Version or commit

T3 Code Alpha 0.0.28 (product version 0.0.28.0)

### Environment

Windows 11 24H2 build 26100.8875, AMD64; T3 desktop 0.0.28 x64; Node.js 24.15.0; PowerShell 7.5.8; inherited PATH with 53 entries.

### Logs or stack traces

```shell
Observed request path:
server.getConfig -> externalLauncher.resolveAvailableEditors() -> sequential editor command checks

Before local patch: approximately 15 to 99 seconds and frequently interrupted by the UI.
After concurrent checks with 3-second per-editor timeout: 3158.2986 ms, Exit Success.

Provider caches continued updating under:
%USERPROFILE%\.t3\caches

Relevant trace:
%USERPROFILE%\.t3\userdata\logs\server.trace.ndjson
```

_No response_

### Workaround

I locally patched the unpacked 0.0.28 server bundle so editor checks run concurrently and each editor has a 3-second timeout. This made `server.getConfig` complete in about 3.16 seconds. This is not a suitable end-user workaround because an application update can overwrite the patched bundle.

StiensWout:
## Summary

- run editor availability probes concurrently instead of serializing every command lookup
- cap each editor probe at three seconds and keep successful partial results when another probe stalls
- preserve the configured editor order and add deterministic virtual-clock regression coverage

## Root cause

`server.getConfig` waits for external editor discovery before returning its configuration snapshot. Editor discovery checked every configured editor sequentially, and every missing Windows command could traverse the full inherited `PATH` across all `PATHEXT` variants. Slow filesystem lookups therefore accumulated across the entire editor catalog and blocked provider statuses from reaching the client.

## Impact

Optional editor detection can no longer block configuration indefinitely. Fast probes remain visible even when another editor lookup times out, allowing the Providers page and other configuration consumers to resolve within a bounded interval.

## Validation

- `pnpm exec vp test run apps/server/src/process/externalLauncher.test.ts` (5 tests passed)
- `pnpm exec vp lint --report-unused-disable-directives apps/server/src/process/externalLauncher.ts apps/server/src/process/externalLauncher.test.ts`
- `pnpm exec vp fmt --check apps/server/src/process/externalLauncher.ts apps/server/src/process/externalLauncher.test.ts`
- isolated authenticated web verification: the Providers page rendered completed provider states; on Windows the editor discovery span completed in about 1.62 seconds while individual missing-command probes took about 1.5 seconds

Closes #4210

---

> [!NOTE]
> **Medium Risk**
> Changes timing and completeness of editor lists (timeouts may hide slow-but-valid editors) and affects config paths that wait on discovery; behavior is bounded and covered by new tests.
> 
> **Overview**
> **Editor discovery** in `ExternalLauncher` no longer runs serial PATH probes that could block `getConfig` indefinitely on slow Windows lookups.
> 
> Probes for each configured editor now run **in parallel**, with a **3-second cap per editor** (`EDITOR_DISCOVERY_TIMEOUT`). Editors that finish in time are included; timed-out or hanging probes are dropped while faster ones still appear in the list. **Command alias resolution** for launch uses `Effect.raceAll` so the first responsive alias wins instead of checking aliases one-by-one.
> 
> `makeWithOptions` exposes injectable `editorCommandAvailability` and `editorDiscoveryTimeout` for tests; production still defaults to `isCommandAvailable`. New tests cover Zed fallback alias racing and virtual-clock behavior that partial results survive when some probes never complete.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ce82e3f6dd06c62d38bfebdff6156e4814cf950. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Bound editor discovery latency by probing aliases concurrently with a 3-second timeout
> - Editor discovery in [`externalLauncher.ts`](https://github.com/pingdotgg/t3code/pull/4221/files#diff-157ca66b5181820e0a628ef443d244577cda2329dd1d88fe5f2aff441da05e56) now probes all editors concurrently instead of sequentially, applying a per-editor timeout (default 3 seconds) and returning partial results for editors that resolved in time.
> - For multi-alias editors (e.g. `zed`/`zeditor`), the selected command is now the first alias whose availability check resolves, rather than using a fixed order.
> - A new `makeWithOptions` factory allows injecting a custom `editorCommandAvailability` function and `editorDiscoveryTimeout`, used by tests to control probe timing.
> - Behavioral Change: editors whose availability probes never complete (e.g. hung filesystem checks) are now silently dropped from results after the timeout rather than blocking discovery indefinitely.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0ce82e3.</sup>
> 
>

### DIFF
--- a/apps/server/src/process/externalLauncher.test.ts
+++ b/apps/server/src/process/externalLauncher.test.ts
@@ -186,7 +186,7 @@ it.effect("bounds slow editor probes concurrently and keeps partial results", ()
     yield* TestClock.adjust(discoveryTimeout);
 
     const editors = yield* Fiber.join(discoveryFiber);
-    assert.deepEqual(editors, ["cursor", "vscode"]);
+    assert.deepEqual(editors, ["cursor", "vscode", "zed"]);
   }).pipe(
     Effect.scoped,
     Effect.provide(
@@ -200,7 +200,9 @@ it.effect("bounds slow editor probes concurrently and keeps partial results", ()
             if (command === "cursor") {
               return Effect.sleep(Duration.seconds(1)).pipe(Effect.as(true));
             }
-            return command === "code" ? Effect.succeed(true) : Effect.never;
+            return command === "code" || command === "zeditor"
+              ? Effect.succeed(true)
+              : Effect.never;
           },
         }),
       ),
--- a/apps/server/src/process/externalLauncher.test.ts
+++ b/apps/server/src/process/externalLauncher.test.ts
@@ -150,6 +150,31 @@ it.effect("launches an installed editor with platform-safe arguments", () =>
   }).pipe(Effect.scoped, Effect.provide(NodeServices.layer)),
 );
 
+it.effect("launches an editor through a responsive fallback alias", () => {
+  let spawned: ChildProcess.StandardCommand | undefined;
+
+  return Effect.gen(function* () {
+    const launcher = yield* ExternalLauncher.ExternalLauncher;
+    yield* launcher.launchEditor({ editor: "zed", cwd: "/workspace" });
+
+    assert.ok(spawned);
+    assert.equal(spawned.command, "zeditor");
+    assert.deepEqual(spawned.args, ["/workspace"]);
+  }).pipe(
+    Effect.provide(
+      testLayer({
+        platform: "linux",
+        env: { PATH: "/bin" },
+        editorCommandAvailability: (command) =>
+          command === "zeditor" ? Effect.succeed(true) : Effect.never,
+        onSpawn: (command) => {
+          spawned = command;
+        },
+      }),
+    ),
+  );
+});
+
 it.effect("discovers editors through the service API", () =>
   Effect.gen(function* () {
     const fileSystem = yield* FileSystem.FileSystem;
--- a/apps/server/src/process/externalLauncher.test.ts
+++ b/apps/server/src/process/externalLauncher.test.ts
@@ -1,12 +1,15 @@
 import * as NodeServices from "@effect/platform-node/NodeServices";
 import { assert, it } from "@effect/vitest";
 import * as ConfigProvider from "effect/ConfigProvider";
+import * as Duration from "effect/Duration";
 import * as Effect from "effect/Effect";
 import * as FileSystem from "effect/FileSystem";
+import * as Fiber from "effect/Fiber";
 import * as Layer from "effect/Layer";
 import * as Path from "effect/Path";
 import * as Sink from "effect/Sink";
 import * as Stream from "effect/Stream";
+import * as TestClock from "effect/testing/TestClock";
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process";
 
 import { HostProcessPlatform } from "@t3tools/shared/hostProcess";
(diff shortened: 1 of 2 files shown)


lmmontoya-ai:
Windows cross-machine validation for `0ce82e3f6` against merge-base `c0bb23734`:

**Method**
- Called the real `ExternalLauncher.resolveAvailableEditors()` service under a live Effect clock.
- 5 serial measurements per revision, same host/environment, after dependencies and transforms were warmed.
- Windows build `26200.8875`, AMD64, Windows PowerShell `5.1.26100.8875`.
- The effective `PATH` has 72 entries; the installed Cursor / VS Code CLI directories occur relatively late.
- This measures the editor-discovery service used by `server.getConfig`, not the complete Electron/renderer connection timeline.

| Revision | n | min | median | max | Result each run |
|---|---:|---:|---:|---:|---|
| base `c0bb23734` | 5 | 9,172.3 ms | 9,360.2 ms | 9,960.6 ms | `cursor`, `vscode`, `vscode-insiders`, `file-manager` |
| PR `0ce82e3f6` | 5 | 3,004.9 ms | 3,005.6 ms | 3,017.3 ms | `file-manager` only |

That is a **67.9% median latency reduction**, and it cleanly prevents the roughly 10-second serialized lookup seen on this affected Windows host.

There is an important completeness tradeoff here: Cursor, VS Code, and VS Code Insiders are installed and consistently detected by the base revision, but all three were dropped in every PR run because their real PATH probes did not complete within the 3-second deadline. This matches the risk noted in the PR description, but demonstrates it on a non-synthetic affected machine.

The focused PR test also passes locally: `apps/server/src/process/externalLauncher.test.ts` — **6/6 passed**.

My read: the concurrent/bounded direction is validated, but the 3-second per-editor cap may need a companion improvement (for example case-insensitive PATH deduplication, prioritizing known editor CLI directories, caching/background refresh, or keeping the hard deadline specifically at the initial config boundary) so startup speed does not hide valid installed editors.

PR #4243, opened 2026-07-21
Sidebar v2: server-backed settled lifecycle (settledAt)

t3dotgg:
## What

**Stacked on #4026** (base branch: `t3code/new-sidebar-client-only`) — do not merge before it.

Replaces the client-only "settle rides archive" model with a real server-backed settled lifecycle, ported from the #3960 spike:

- **Server**: `thread.settle` / `thread.unsettle` commands, `thread.settled` / `thread.unsettled` events, `settled_override` + `settled_at` columns on `projection_threads` (migration 033). The decider auto-un-settles on real activity: a user message, a session coming alive, or a new approval/user-input request emits `thread.unsettled(reason: "activity")` ahead of the triggering event.
- **Clients**: settle/unsettle dispatch the new commands; `effectiveSettled` keys on `settledOverride`/`settledAt`. `unsettle(reason: "user")` pins a thread active, suppressing auto-settle until real activity.

## Why (what this fixes structurally)

Settled threads now **stay in the live shell stream** — settled ≠ archived. That one property deletes the whole workaround layer the client-only model needed:

- No more `settledHolds` optimistic-hold machinery (web + mobile)
- No more archived-snapshot merging into the v2 lists
- Opening a settled row is plain navigation — **history is readable without un-settling** (the client-only model had to unarchive-on-open)
- Un-settle works uniformly on every settled row (the auto-settled silent no-op case is gone)
- Archive regains its original meaning (hidden from lists), fully distinct from settle

## Safety

- Same beta gating as #4026: all v2 UI behind `sidebarV2Enabled` / `threadListV2Enabled`, default off.
- Migration 033 is additive (two nullable columns, idempotent `PRAGMA`-guarded `ALTER TABLE`), no backfill, no data rewritten.
- New commands/events are additive to the protocol; old clients simply never send them. Contract fields decode with null defaults, so old snapshots parse.
- Threads archived during the client-only beta stay archived (hidden) — they are **not** auto-converted to settled. Users can unarchive from Settings → Archived as before.

## Review focus

- `apps/server/src/orchestration/decider.ts` — settle/unsettle command handling + the three activity-un-settle injection points (message send, session set, activity append). Idempotency-by-re-emission is documented inline.
- `apps/server/src/orchestration/projector.ts` + `ProjectionPipeline.ts` — `reason: "user"` → override `"active"` (pin), `reason: "activity"` → override `null` (neutral).
[...]

PR #4245, opened 2026-07-22
feat(orchestrator): queue follow-up messages during active turns, with explicit steer

t3dotgg:
## Summary

Messages sent while a turn is running are now **queued server-side by default** and drained automatically when the turn completes naturally. Each queued message renders as a chip above the composer with two actions:

- **Steer** — dispatch it immediately, injecting into the active turn (Codex uses the protocol-native `turn/steer` with an `expectedTurnId` precondition, falling back to `turn/start`; other providers steer through their existing sendTurn-while-running paths).
- **Remove** — drop it from the queue.

This replaces the previous behavior where a mid-turn send always steered immediately.

image

## How it works

- **Contracts**: new client commands `thread.queue.steer` / `thread.queue.remove`, internal `thread.queue.drain`, events `thread.message-queued` / `thread.queued-message-removed`, and `queuedMessages` on the thread schema (decoding default keeps old snapshots valid).
- **Decider**: `thread.turn.start` on a busy thread (session `starting`/`running`) emits `thread.message-queued` instead of starting a turn; bootstrap starts are exempt. Steer dispatches unconditionally (degrades to a normal turn start if the thread went idle). Drain is invariant-guarded: it refuses when the thread is busy again, so a user message racing the drain wins and the queue retries on the next completion.
- **Ingestion**: auto-drain fires only on natural `turn.completed` — never after an interrupt or failure — and is dispatched after assistant finalization so the drained message sequences after the assistant text it follows up on.
- **Persistence**: `projection_queued_messages` table (migration 033) so queues survive server restarts.
- **Web**: `QueuedMessageChips` above the composer; the composer's local-dispatch acknowledgment treats a projected queued message as server acknowledgment so it doesn't stick in "Sending".
- Queued messages are not part of the thread timeline; they enter it via the normal `thread.message-sent` event when dispatched. Stale `sourceProposedPlan` references are dropped rather than failing the send. Queue capped at 50 per thread.

## Known gaps (follow-ups)

- **Mobile has no queued-message UI yet.** Mobile's client-side outbox already holds sends while a thread is busy (`resolveThreadOutboxDeliveryAction` waits on `threadBusy`), so mobile sends generally won't hit the server queue — but once its outbox dispatches into a turn that just started, the message queues server-side invisibly. Mobile chips + steer/remove need a follow-up.
- No keyboard shortcut for "send immediately as steer" from the composer; steering requires the chip button.

## Validation

- `vp run typecheck` clean across the monorepo
- `vp check` clean on all 36 changed files
- Server orchestration + persistence: 214 tests passing, including 8 new decider queue tests (`decider.queue.test.ts`) covering queue-on-busy, steer, remove, drain, drain-races, and stale plan references
- Web + client-runtime + contracts: 1825 tests passing (new coverage for queued-send acknowledgment)
- Full `apps/server` suite passes except `ProviderRegistry.test.ts > re-probes when settings change the codex binaryPath`, which fails only under the full-suite run and passes in isolation and on main — pre-existing flake, unrelated to this diff

---

> [!NOTE]
> **High Risk**
> Changes core turn-start semantics, event sourcing projections, and provider steering paths across server, persistence, and clients; incorrect queue/drain or pending-turn handling could drop messages, double-dispatch, or leave threads stuck.
> 
> **Overview**
> **Follow-ups sent while a turn is starting or running are queued server-side** instead of immediately starting another turn or steering. Idle threads still get an immediate `thread.message-sent` + `thread.turn-start-requested`; busy threads emit `thread.message-queued` (50-message cap enforced in the decider).
> 
> New commands **`thread.queue.steer`**, **`thread.queue.remove`**, and internal **`thread.queue.drain`** dispatch or drop queued items. After a **natural** turn completion, runtime ingestion auto-dispatches drain so the FIFO head becomes the next turn. **`pendingTurnStart`** on the thread read model blocks duplicate drains and follow-up sends in the gap before the provider adopts a turn; session-set projection clears stale pending rows on settled statuses (including `ready` after mid-turn steer).
> 
> **Persistence**: migration 034 adds `projection_queued_messages`, a dedicated projector (ordered before thread-messages on replay so revert pruning keeps queued attachments), and snapshot/query hydration of `queuedMessages` + `pendingTurnStart`. Revert and user-removal paths update attachment pruning to include queued rows.
> 
> **Web**: `QueuedMessageChips` above the composer (steer/remove); composer acknowledgment correlates sends by `messageId` when timeline or queue projects the message. **Codex** `sendTurn` tries protocol `turn/steer` with `expectedTurnId` before falling back to `turn/start`. Contracts, client-runtime reducer/commands, and WS detail events extended for the new queue events.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6be48eb238cf310a60cc9571240601e774c7d381. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Queue follow-up messages during active turns and surface them as steerable chips in the chat UI
> - When a `thread.turn.start` command arrives while a turn is active or pending, the decider now emits `thread.message-queued` instead of starting a turn immediately, capped at 50 queued messages per thread.
> - After a turn completes successfully, the ingestion layer auto-dispatches a `thread.queue.drain` to send the oldest queued message, enabling sequential follow-up delivery.
> - Three new commands are introduced: `thread.queue.steer` (dispatch a specific queued message into the running turn via a protocol-native steer), `thread.queue.remove` (drop a queued message), and `thread.queue.drain` (internal drain after turn completion).
> - The chat UI renders queued messages as chips above the composer with "Steer" and "Remove" actions via a new `QueuedMessageChips` component.
> - A new `projection_queued_messages` DB table (migration 034) persists queued messages server-side; the projector and snapshot query propagate `queuedMessages` and `pendingTurnStart` through the read model.
> - Local dispatch acknowledgment in the web client is now correlated by a specific `messageId` appearing in either the timeline or the queue, replacing the global tail heuristic.
> - Risk: the steer path in `CodexSessionRuntime` falls back to `turn/start` only on a specific `receive-response` rejection; other errors propagate without fallback.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6be48eb.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added queued follow-up messages for active threads.
  * Messages sent while a turn is running are queued and processed in order when the turn completes.
  * Added controls to steer a queued message immediately or remove it.
  * Queued messages now appear in thread details and the chat interface.
  * Added support for provider-native steering, with fallback behavior when unavailable.
* **Bug Fixes**
  * Improved local message acknowledgement and optimistic timeline cleanup for queued messages.
  * Preserved queued messages consistently across thread updates and synchronization.


t3dotgg:
Addressed all review feedback (rebased onto latest main; migration renumbered 033 → 034 after #4243 took slot 033):

**Bugbot — queue cap desync (high):** Fixed by moving the 50-message cap into the decider: a `thread.turn.start` that would exceed the cap is rejected with an invariant error, so a `thread.message-queued` event past the cap is never emitted and the in-memory read model, SQL projection, and event replay can no longer diverge. The projector's local `slice()` clamp is removed. New decider test covers the overflow rejection.

**CodeRabbit — unscoped `turn/steer` fallback (major):** The fallback to `turn/start` now catches only `CodexAppServerRequestError` (the request-level precondition rejection); transport and protocol failures propagate instead of racing a duplicate provider turn.

**CodeRabbit — blob preview leak (minor):** Optimistic sends acknowledged only by the queue now revoke their blob preview URLs instead of handing them off (queued chips render text only, so the handoff would never be promoted).

**CodeRabbit — `getCommandReadModel` updatedAt (minor):** Queued-message timestamps now advance `updatedAt`, matching `getSnapshot`.

**CodeRabbit — reducer test coverage (trivial):** Added threadReducer lifecycle tests for queue, replace-by-messageId, and remove.

An independent gpt-5.6-sol (Codex CLI) review of the full branch diff found two additional issues, both fixed:

- **P1 — steer during session startup:** `thread.queue.steer` while the session was `starting` would fall through to `turn/start` in the Codex runtime (no active turn id yet) and race the pending startup with a second provider turn — after the message had already been dequeued. The decider now rejects steer while `starting`; the message stays queued and drains or steers normally once running. Covered by a new decider test.
- **P2 — queued attachments deleted by revert pruning:** checkpoint-revert attachment pruning computed retained paths only from timeline messages, so reverting a thread deleted image files still referenced by queued messages. Queued-message attachments are now included in the retained set.

Validation: monorepo typecheck clean, `vp check` clean on all 37 changed files, 1192 tests across server orchestration/persistence, client-runtime, and web components passing.

t3dotgg:
Rebased onto latest main and addressed the second round of review findings:

**Bugbot / Macroscope — composer acknowledges unrelated queue changes (medium):** Rewrote the acknowledgment to be correlated instead of heuristic. `LocalDispatchSnapshot` now records the dispatched `messageId` (all three send paths pass it), and during a running turn the composer clears "Sending" only when that exact message appears in the projected timeline or queue. Another client queueing a message, or a different chip being steered/removed, no longer clears this client's send state. Dispatches without a known messageId (plan-implementation flow) keep the legacy latest-user-message heuristic. New logic tests cover both the match and the unrelated-change cases.

**Macroscope — same-timestamp queue mis-ordering after restart (medium):** Queue drain order is now SQLite `rowid` (insertion) order instead of `queued_at, message_id`, so same-millisecond enqueues can't swap after a restart. The upsert is now `INSERT OR REPLACE` so a re-queued messageId moves to the queue tail, matching the in-memory projector's replace semantics. Migration 034's index drops the now-unused `queued_at` column (034 hasn't shipped, so edited in place).

**Bugbot — deleted thread keeps in-memory queue (low):** The in-memory projector now clears `queuedMessages` on `thread.deleted`, mirroring the SQL projection.

**Macroscope — bootstrap projector ordering deletes queued attachments (high):** The `queuedMessages` projector now bootstraps before `threadMessages`, so the revert handler's retained-attachment set sees queued messages during replay.

**Bugbot — steer blocked during mid-turn reconnect (medium):** The steer guard now rejects only the pending-start shape (`starting` with no `activeTurnId`). A session reporting `starting` while still tracking an active turn (e.g. Codex `session/connecting` mid-turn) is steerable again.

Also updated `ProviderCommandReactor` tests whose follow-up sends now queue by default: the harness gained a `settleSession` helper that returns the projected session to `ready` (preserving provider identity) before the second send, mirroring what turn completion does in production.

Validation: monorepo typecheck clean, `vp check` clean on all 39 changed files, and the full affected suites pass — 3649 tests across apps/server, apps/web, client-runtime, and contracts.

t3dotgg:
Round 3 findings addressed:

**Bugbot — idle send bypasses pending queue / Macroscope — drain race window (medium):** This was the real design gap in the round. Fixed with an event-driven signal instead of timestamps: the thread read model now carries `pendingTurnStart`, set by `thread.turn-start-requested` and cleared when the session adopts the turn (`running` with `activeTurnId`) or reaches a terminal status (a failed start sets session `error`, which clears it). The decider's queue-by-default check, the drain guard, and the steer guard all consult it, so a send racing in behind a just-dispatched turn start (or drain) queues behind existing chips instead of opening a second concurrent turn. It hydrates from the existing SQL pending-turn-start rows on restart. I deliberately did **not** use the message-timestamp heuristic the settle guard uses — message `createdAt` is client-supplied, and a skewed client clock would wrongly queue sends on an idle thread where nothing ever drains them (the settle guard keeps its heuristic, renamed `hasRecentUnadoptedUserMessage`, where the failure mode is only a delayed settle). New decider test covers queue-behind-pending-start and drain-rejected-while-pending.

**Bugbot + Macroscope — connecting-phase sends never acknowledge (medium):** The projected `expectedMessageId` check now runs in every composer phase, not just `running` — the server also queues sends while a session is `starting`, which the client renders as `connecting`.

**Macroscope — queued-message removal leaks attachment files (medium):** User removals now record a prune of the thread's attachment set (retained timeline + remaining queue) so the removed message's files are deleted. Dispatch removals intentionally keep everything — the same attachments re-enter the timeline via the paired `thread.message-sent`.

**Macroscope — steer fallback on decode failures duplicates input (in the drain-race comment):** The Codex `turn/steer` fallback now matches only `operation: "receive-response"` request errors — the server actively rejected the steer, so it was not applied. Decode failures on an accepted steer's response (and transport/protocol errors) propagate instead of re-sending the same input via `turn/start`.

Validation: monorepo typecheck clean, `vp check` clean on the branch's changed files, and the affected suites pass — 3,649 tests across apps/server, apps/web, client-runtime, and contracts (single failure is the known pre-existing `ProviderRegistry` binary-probe flake, which passes in isolation and on main).

t3dotgg:
Round 4 addressed — all three were real consequences of round 3's `pendingTurnStart`, and the first was a genuine high-severity catch:

**Bugbot — stuck pending blocks drain (high):** Confirmed. A mid-turn steer emits `thread.turn-start-requested` while the session is already `running`, re-arming `pendingTurnStart` — and since the session never passes through a fresh adoption transition, only the old terminal-status clearing (error/stopped/interrupted) would ever release it. Natural completion (`ready`) left it stuck, so every subsequent auto-drain was rejected and swallowed: queued follow-ups would sit forever. Both the in-memory projector and the SQL projection now clear the pending start on **every settled status including `ready`**. That's safe for genuinely-pending starts because ingestion never projects those as `ready` — it maps ready-with-pending-start to `starting`. New decider test walks the full lifecycle: queue two → steer one mid-turn → complete → verify the flag released → drain dispatches the second.

**Bugbot — steer races pending adoption (high):** The guard now rejects steer whenever a turn start awaits adoption and the session tracks no active turn — including `running` with a null `activeTurnId` (provider waiting shape), which the previous status-based condition let through.

**Bugbot — detail snapshot drops pending start (medium):** `getThreadDetailById` now hydrates `pendingTurnStart` from the SQL pending row, so thread-detail refreshes agree with the command read model instead of silently defaulting to null.

Validation: typecheck clean, `vp check` clean on the diff (2 remaining warnings are pre-existing in `CommandPalette.tsx` from main), 3,650 tests passing across server/web/client-runtime/contracts — sole failure remains the known `ProviderRegistry` probe flake (passes in isolation and fails identically on clean main).

xptea:
Need

juliusmarminge:

PR #4251, opened 2026-07-22
Show working status for new threads

t3dotgg:
Shows “Working” while a new thread waits for its first provider turn, with safe cleanup for failed or ended starts.

Works in both sidebar versions. Validated with focused tests, typechecks, lint, and an isolated browser run.


t3dotgg:
Closing as superseded by current pending-start projection behavior and #4101. Main retains starting while a turn is pending, and Sidebar v2 already renders starting as Working.

PR #4408, opened 2026-07-23
Persist desktop preferences across builds and restarts

danieliser:
## What Changed

- Standardized packaged Alpha, Nightly, and stable builds on the canonical `t3code` Electron profile while keeping unpackaged development builds isolated in `t3code-dev`.
- Added durable desktop-backed storage for renderer UI state and sticky composer preferences instead of relying only on origin-scoped `localStorage`.
- Hydrated client settings, UI state, and sticky model/provider choices before first render using guarded three-way reconciliation so startup reads cannot overwrite newer user edits.
- Serialized writes, surfaced persistence failures, and added bounded flush barriers for direct window close, app quit, relaunch, signals, and update installation.
- Migrated valid legacy Alpha/Nightly browser state and pre-v3 composer preference shapes without promoting malformed or partial documents.
- Kept thread drafts, images, and thread-local composer content out of the durable preference document.

## Why

Desktop preferences were split across Electron profile directories, renderer-origin storage, and server-side settings. Build identity and origin changes could therefore make labels, manual project ordering, explicit unread state, and default model selections appear to reset even though other durable state still existed.

This gives all packaged channels one stable source of truth, preserves the desired Alpha/Nightly sharing behavior, and prevents unpackaged development sessions from modifying that packaged profile.

## Data Safety

- Canonical data wins; legacy profile migration runs only when the canonical profile is absent.
- Malformed durable or browser documents fail closed and do not overwrite valid state with defaults.
- Startup reconciliation preserves edits made while hydration is in flight.
- Writes use serialized queues and atomic replacement.
- Close and shutdown paths wait for renderer acknowledgement with a bounded timeout so Quit cannot hang indefinitely.
- Legacy cleanup happens only after the durable write succeeds.

## Verification

- 111 focused web persistence tests passed.
- 40 focused desktop persistence tests passed.
- Web and desktop TypeScript checks passed.
- Targeted lint, formatting, and `git diff --check` passed.
- The combined patched-nightly integration passed 192 focused persistence tests plus web and desktop typechecks.
- Independent review findings covering migration validation, write ordering, close handling, and updater shutdown were reproduced where valid and fixed with focused regressions.

## Checklist

- [x] Packaged channels share one canonical profile
- [x] Unpackaged development remains isolated
- [x] Legacy preferences migrate without accepting corrupt documents
- [x] Direct close, quit, relaunch, signals, and updater install flush pending state
- [x] No user-facing layout or animation changes are included


t3dotgg:
> [!NOTE]

This branch combines desktop identity, settings, onboarding, window state, and update persistence in one 41-file change. Those stores have different lifetimes and failure cases, so one review cannot safely define all of their restart and migration behavior.

PR #4453, opened 2026-07-24
feat(web): prompt stash — cmd+S saves the composer to a per-provider queue

t3dotgg:
## What

Copies Claude Code's "cmd+S to stash a prompt" flow into the web composer. Pressing **⌘S / Ctrl+S** snapshots the in-progress prompt — text **and image attachments** plus the model selection — into a client-only stash queue, then clears the composer with an Undo toast. The queue is stored in localStorage and scoped per connection method (`ProviderInstanceId`, so Claude Code / Codex / Cursor / custom instances each get their own queue).

Design + interactive mocks: https://puxe56fxtfjc.postplan.dev (this ships mocks **D + B**).

## UX

- **⌘S with content** → stash: composer clears, a ghost of the prompt flies into a bookmark badge on the composer's shoulder (one-shot 550ms animation, `prefers-reduced-motion` respected), success toast with **Undo**.
- **⌘S on an empty composer** (or clicking the badge) → opens the stash picker above the composer: arrows to navigate, **Enter** to restore, **⌘⌫** or the hover ✕ to delete, **Esc** to close.
- **Restore** rehydrates everything: prompt (appended if the composer already has text), images rebuilt into live `File`s through the existing persisted-attachment codec, and the saved model selection (only if that provider instance is still enabled). Restoring removes the entry — it's a queue, not a library.
- Picker shows a hint when other connection methods have stashed prompts.

## Implementation

- `composer.stash` command in the keybinding contract + default `mod+s` (`!terminalFocus`) — rebindable via the existing registry. The handler runs capture-phase on `window` and always `preventDefault()`s so the browser save dialog never appears, even in states that can't stash (approvals, plan questions, palette open).
- New `promptStashStore` (`t3code:prompt-stash:v1`): zustand `persist` + Effect-Schema validation + debounced storage + `beforeunload` flush, cloned from the composer draft store's conventions. 20 entries per queue (oldest evicted with a toast note), ~2MB per-entry attachment budget with dropped images surfaced by name, quota-safe writes that degrade to in-memory instead of throwing in the debounce timer.
- Terminal/element contexts are intentionally not stashed — they reference live sessions and would dangle; inline placeholders are stripped from the stashed prompt.

## Testing

- 8 new unit tests for the store (scoping incl. `__none__` bucket, LIFO order, eviction at cap, take-once semantics, attachment budget partitioning).
- Full monorepo typecheck clean; web unit suite 1513 passed; shared/contracts suites pass; lint clean on touched files.
- Not visually verified — automated browsers can't see authenticated views here, so the badge/fly/picker need an eyeball pass.

---

> [!NOTE]
> **Medium Risk**
> Rebinds mod+s in the composer and persists large base64 payloads in shared localStorage; failure modes are mostly toasts and no-op, but quota or non-durable storage can leave stash/composer state surprising after reload.
> 
> **Overview**
> Adds a **prompt stash** to the chat composer: **⌘S / Ctrl+S** (`composer.stash`, `!terminalFocus`) always prevents the browser save dialog, then either snapshots the current draft or toggles the stash picker when the composer is empty.
> 
> **Stash behavior:** Text (inline terminal placeholders stripped), model selection, and images go into a **per–connection-method** queue in `promptStashStore` (localStorage, 20 entries, immediate persist with quota / durability handling). The composer clears **only** prompt + images via `clearComposerPromptAndImages`, leaving terminal/element contexts and review comments. Images are compressed for storage (`compressImageForStash`); restore rehydrates files, dedupes, respects attachment limits, and can reapply model selection when the provider is still available.
> 
> **UI:** `ComposerStashBadge` + `ComposerStashMenu` (keyboard navigation, delete, cross-scope hints) with a short count pulse animation in CSS.
> 
> **Tests:** Unit coverage for the stash store and image compression pipeline.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69fcdddfe26efed03d803276e14713ff6e8dc10f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add cmd+S prompt stash to composer with per-provider queue and stash menu
> - Pressing `mod+s` in the composer saves the current prompt and images to a provider-scoped stash queue (up to a bounded cap), clears the prompt and images, and briefly pulses a new `ComposerStashBadge` showing the queue count.
> - New [`promptStashStore`](https://github.com/pingdotgg/t3code/pull/4453/files#diff-c7ec119c2fc8d8344e657f320c2e7a369aa9f37467ce6e034ddb1ff9fc809bf8) persists stash entries to `localStorage` with an in-memory fallback, enforces per-queue eviction, and handles async image finalization via `compressImageForStash`.
> - New [`ComposerStashMenu`](https://github.com/pingdotgg/t3code/pull/4453/files#diff-2fcc66d7913594ea9e35be547ee395de158ea142ad58496a4f8089f6ca66ebff) lists stashed entries with keyboard navigation (arrows, Enter to restore, `mod+Backspace` to delete) and shows attachment previews, image save status, and a footer notice when other provider queues have entries.
> - Restoring an entry merges prompt text and adds images up to the attachment capacity limit; warnings are shown for dropped, unreadable, or over-limit images.
> - Risk: stash persistence relies on `localStorage`; if storage is unavailable or non-durable, toasts warn the user but the in-memory state still works for the session.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 69fcddd.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/chat/ChatComposer.tsx
+++ b/apps/web/src/components/chat/ChatComposer.tsx
@@ -51,13 +51,28 @@ import {
   type ComposerImageAttachment,
   type DraftId,
   type PersistedComposerImageAttachment,
+  hydrateImagesFromPersisted,
   useComposerDraftStore,
   useComposerThreadDraft,
   useEffectiveComposerModelState,
 } from "../../composerDraftStore";
 import {
+  EMPTY_PROMPT_STASH_QUEUE,
+  flushPromptStashStorage,
+  partitionStashAttachments,
+  promptStashScopeKey,
+  usePromptStashStore,
+  type PromptStashEntry,
+} from "../../promptStashStore";
+import { ComposerStashBadge, type StashFlyGhost } from "./ComposerStashBadge";
+import { ComposerStashMenu } from "./ComposerStashMenu";
+import { isCommandPaletteOpen } from "../../commandPaletteBus";
+import { getTerminalFocusOwner } from "../../lib/terminalFocus";
+import { resolveShortcutCommand } from "../../keybindings";
+import {
   type TerminalContextDraft,
   type TerminalContextSelection,
+  INLINE_TERMINAL_CONTEXT_PLACEHOLDER,
   insertInlineTerminalContextPlaceholder,
   removeInlineTerminalContextPlaceholder,
 } from "../../lib/terminalContext";
@@ -723,6 +738,8 @@ export const ChatComposer = memo(function ChatComposer(props: ChatComposerProps)
   const clearComposerDraftPersistedAttachments = useComposerDraftStore(
     (store) => store.clearPersistedAttachments,
   );
+  const clearComposerDraftContent = useComposerDraftStore((store) => store.clearComposerContent);
+  const setComposerDraftModelSelection = useComposerDraftStore((store) => store.setModelSelection);
   const syncComposerDraftPersistedAttachments = useComposerDraftStore(
     (store) => store.syncPersistedAttachments,
   );
@@ -961,6 +978,8 @@ export const ChatComposer = memo(function ChatComposer(props: ChatComposerProps)
   const [isComposerModelPickerOpen, setIsComposerModelPickerOpen] = useState(false);
   const [isComposerFocused, setIsComposerFocused] = useState(false);
   const [composerMenuAnchor, setComposerMenuAnchor] = useState<HTMLDivElement | null>(null);
+  const [isStashMenuOpen, setIsStashMenuOpen] = useState(false);
+  const [stashGhost, setStashGhost] = useState<StashFlyGhost | null>(null);
   const isMobileViewport = useMediaQuery("max-sm");
   const isComposerCollapsedMobile =
     isMobileViewport && !forceExpandedOnMobile && !isComposerFocused;
@@ -980,6 +999,8 @@ export const ChatComposer = memo(function ChatComposer(props: ChatComposerProps)
   const mobileComposerExpandReleaseFrameRef = useRef<number | null>(null);
   const mobileComposerExpandInFlightRef = useRef(false);
   const dragDepthRef = useRef(0);
+  const stashGhostKeyRef = useRef(0);
+  const stashGhostTimeoutRef = useRef<number | null>(null);
 
   // ------------------------------------------------------------------
   // Derived: composer send state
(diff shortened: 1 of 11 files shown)


t3dotgg:
Addressed all 7 review comments (rebased onto latest `main`, so the branch was force-pushed).

**Cursor Bugbot**

1. **Async stash wipes newer composer input** — fixed. The composer is now cleared *synchronously* at keypress, before any image-encoding `await`, so edits made during encoding can't be clobbered. Added an in-flight guard so a repeat ⌘S during encoding can't stash the same snapshot twice.
2. **Restore drops excess images permanently** — fixed. Attachments past the composer's limit are now reported by name in a warning toast instead of being silently sliced away.
3. **Unreadable images blamed on size limit** — fixed. `unreadableImageNames` is now a separate field on the entry, so the UI says "could not be read" vs "exceeded the stash size limit" vs "composer is at its N-image limit" as appropriate.
4. **Stash wipes unrestorable contexts** — fixed. Added `clearComposerPromptAndImages` to the draft store; stash now clears only what it can round-trip, leaving terminal/element contexts, preview annotations, and review comments intact.

**Macroscope**

5. **`__proto__` prototype pollution** (`promptStashStore.ts:128`) — guarded with `Object.hasOwn`. Note the premise is narrower than described: `ProviderInstanceId` is validated against `/^[a-zA-Z][a-zA-Z0-9_-]*$/`, so no real provider id can be `__proto__`. The guard still matters because persisted queue keys decode as plain `Schema.String`, so a corrupted/hand-edited localStorage payload can carry one. Regression test covers that rehydration path.
6. **`__none__` scope collision** (`promptStashStore.ts:44`) — fixed as suggested: provider keys are now namespaced `provider:<id>`, so they can never equal the unscoped sentinel. (Same caveat as above — a provider named `__none__` isn't constructible today, but the prefix makes it structural rather than incidental.)
7. **Enter on the delete button restores instead of deleting** — fixed; the capture-phase handler now bails when the event originates from a focused control inside the row.

**Also in this push** (from local testing, not review comments): stashed images are re-encoded (WebP with JPEG fallback, 2048px, quality ladder to 0.68) instead of being dropped when over budget — only the stashed copy is compressed, so sending without stashing still uploads the original. The save toast and fly-in animation were removed in favor of a subtle badge pulse.

Verification: web typecheck clean, 1535 unit tests passing (17 for the stash store + compression), lint clean.

t3dotgg:
Second round addressed in `01bb40b3a`.

**Failed image stash still clears composer** (High) — fixed, and worth explaining since it's in direct tension with the previous round's fix. Clearing *after* the async encode loses edits typed during encoding; clearing *before* loses the prompt if the tab closes mid-encode. Neither ordering is safe on its own, so the write is now split: the text-only entry is persisted and flushed **first**, the composer is cleared **second**, and images are attached via a new `finalizeEntryImages` action as they finish encoding. Entries carry a `pendingImageCount` in the meantime (shown as "saving N images…" in the menu), so a crash mid-encode leaves the prompt recoverable rather than gone.

**Stash lock never clears on error** (Medium) — fixed; `stashInFlightRef` is now cleared in a `finally`.

**Queue eviction toast never shown** (Medium) — fixed; the evicted entry returned by `stashEntry` is now surfaced in a warning toast. (The general save toast stays removed per feedback in-thread; this one only fires on actual data loss.)

**Size failures marked unreadable** (Medium) — fixed at the source: `compressImageForStash` now returns a tagged `{ ok: false, reason: "too-large" | "unreadable" }` instead of `null`, so the caller no longer has to guess. Budget failures and decode failures are tracked and reported separately.

**Fallback downscale never shrinks below source** (`stashImageCompression.ts:191`, Medium) — good catch, this was a real no-op. The passes multiplied a fixed `MAX_DIMENSION`, but `encodeWithinBudget` clamps to the bitmap, so for any source under 2048px all three passes resolved to the source size. They now scale off `min(MAX_DIMENSION, longest edge)`. Added a regression test with an 800px source asserting a genuine sub-source downscale is attempted.

**`localStorage` property access can throw `SecurityError`** (`promptStashStore.ts:128`, High) — fixed as suggested with a guarded `resolveBaseStorage()` helper, so a blocked-storage or sandboxed-iframe context falls back to in-memory instead of crashing at module import.

**Restore drops excess images permanently** — this one is a re-post of a comment already resolved in `cad63658b`; `restoreStashEntry` reports overflow by name in a warning toast rather than discarding silently.

Verification: web typecheck clean, 1536 unit tests passing (18 across the stash store + compressor), lint clean.

t3dotgg:
Third round addressed in `60c290477`.

**Clear ignores failed stash persist** (Medium) — real bug, fixed. `createQuotaSafeStorage` swallowed quota errors, so a rejected write still cleared the composer and a reload lost both copies. `flushPromptStashStorage()` now returns whether the write actually landed; on failure the entry is rolled back, the composer is left untouched, and an error toast explains why.

**Restore drops in-flight images** (High) — fixed. `finalizeEntryImages` now returns `false` when its target entry is gone instead of silently no-oping, and the caller warns that the images could not attach. Restore is also blocked (click and Enter) while an entry still has a `pendingImageCount`, so the window is much harder to hit in the first place.

**Encode throw leaves pending forever** (Low) + **Exceptions from `encodeWithinBudget` escape** (`stashImageCompression.ts:208`, Medium) — same root cause, fixed together. The per-pass encode is now wrapped, so a throw from canvas allocation or `convertToBlob` returns `{ ok: false, reason: "unreadable" }` rather than escaping the loop, skipping finalization, and stranding the entry as permanently "saving…". The `finally` that closes the bitmap is retained.

Four other comments in this batch (`Failed image stash still clears composer`, `Size failures marked unreadable`, `Stash lock never clears on error`, `Queue eviction toast never shown`) are re-posts against the pre-fix commit — all were resolved in `01bb40b3a`. They show with `line: null` in the API, which is how Bugbot surfaces comments whose anchor no longer exists in the current diff.

New tests: `finalizeEntryImages` attaching vs. reporting a lost entry. Web typecheck clean, 1538 unit tests passing, lint clean.

t3dotgg:
Pushed `5c1955d01` — six more fixes, from an independent audit rather than review comments (the same classes of bug the bots have been finding, so worth getting ahead of).

**High — reload mid-encode created permanently unrestorable entries.** `pendingImageCount` is written before encoding, but the loop that clears it does not survive a reload. Orphaned entries showed "saving…" forever and refused to restore. The count is now settled during rehydration, with the lost images recorded so the prompt itself stays restorable.

**High — cross-tab writes deleted each other's entries.** Every mutation persisted the tab's entire `queuesByScopeKey` from its own hydrated snapshot, so two tabs stashing would clobber one another. Mutations now re-read from storage and union by entry id (preferring whichever copy has images attached), newest-first, capped at the queue limit.

**High — a rejected second-phase write was ignored.** The images flush discarded its boolean, so if the text entry fit but the image payloads exceeded quota, disk kept the phase-one entry with a pending count and the user saw nothing. Now warned.

**Medium — the in-flight guard rejected unrelated prompts.** It was a single boolean held for the whole encode, but the composer is cleared before the first `await` — so typing a genuinely new prompt (or switching threads) and pressing ⌘S was silently dropped. Now keyed by target+prompt+image-ids.

**Medium — restore capacity used id-only dedup.** The draft store also dedupes by `mimeType+sizeBytes+name` (`composerDraftStore.ts:628`), so a metadata duplicate consumed a capacity slot the store then refused to fill, pushing a genuinely unique image into overflow and deleting it with the entry. Restore now filters on the same key.

**Low — image-only restore appended blank lines and moved the caret.** Restoring a stash with no text into a non-empty composer added `\n\n` and yanked focus to the end. Both are now conditional on the prompt actually changing.

New tests: cross-tab merge and pending-count recovery. Web typecheck clean, 1540 unit tests passing, lint clean.

t3dotgg:
Pushed `f0abec9cf`. Most of this round's comments were bugs in the cross-tab merge I added last round, so rather than patch it (tombstones, flush-before-read) I replaced the approach.

**The merge was unsound.** An id-union of disk and memory can't distinguish "another tab created this entry" from "this tab deleted it" — so deletes resurrect, and no amount of tombstoning makes that clean. Every mutation now does a **synchronous read-modify-write against storage**: read live disk state, apply the change, write back. Writes are immediate rather than debounced; stashing is a deliberate keystroke, not a per-character autosave, so there was nothing to coalesce and the debounce window was pure race surface. This resolves at once:

- **Merge resurrects deleted stash entries** (High) — structurally impossible now; disk is the source of truth.
- **Mutations skip cross-tab merge** (High) — `takeEntry` and `finalizeEntryImages` go through the same path as `stashEntry`.
- **Merge overwrites finalized images** (High) — no merge, no preference rule to get wrong.
- **Failed stash rollback wipes peers** (High) — rollback is re-reading disk, not replaying a stale snapshot.
- **Read-merge-write is not atomic** (`promptStashStore.ts:238`) and **stashEntry resurrects deleted entries** (`:306`) — same root cause.
- **Merge keeps orphaned pending** (Medium) — stale counts are settled on *every* read, not just hydration.

Also dropped the zustand persist middleware (writes are direct now) and added a `storage` event listener so a tab picks up another tab's changes live.

**Orphaned stash unrestorable** (High, + the Macroscope variant) — the `pendingImageCount` guard is gone. Orphans are settled on read, so a nonzero count now genuinely means "encoding right now", and restoring then is already handled by the existing "images did not attach" warning. Blocking restore was the wrong trade: it trapped recoverable text.

**Delete ignores persist failure** (Medium) — `deleteStashEntry` and `restoreStashEntry` now surface a rejected write instead of discarding the result.

**Non-durable fallback reported as durable** (`promptStashStore.ts:117`) — durability is now tracked separately from write success, so the in-memory fallback no longer claims a stash will survive a reload.

**Compression skips smaller scales** (Medium) — an encode throw is often caused by the target size itself, so the loop now continues to the smaller fallback passes instead of returning immediately; the failure reason reflects whether the *last* attempt threw.

New tests cover the cross-tab guarantees directly: another tab's entries survive a stash, a deleted entry is not resurrected, a take doesn't clobber a peer's entries, and a stale pending count is settled on read. Web typecheck clean, 1542 unit tests passing, lint clean.

t3dotgg:
Pushed `69fcdddfe` — removed the cross-tab handling entirely.

Stepping back: no other persisted store in this app does cross-tab merging or storage-event syncing. `composerDraftStore` (which persists live drafts and the same base64 images to the same origin), `rightPanelStore`, `uiStateStore`, `terminalUiStateStore`, `diffPanelStore` — all plain last-write-wins. The stash had accumulated bespoke cross-tab machinery through successive review rounds, and that machinery caused more real bugs than the edge case it guarded (two tabs stashing prompts simultaneously, which the rest of the app tolerates everywhere). Cross-tab consistency is explicitly out of scope for this store now, matching the app's conventions.

What remains is the simple shape: mutate in-memory state, persist it synchronously, hydrate once at startup. Kept from the review rounds, because they fix real single-tab bugs: immediate (non-debounced) writes so the composer only clears when the write actually landed, durability reporting (including the non-durable in-memory fallback), rollback of a rejected stash write, and settling stale `pendingImageCount` on hydration.

Net for the store rounds: `promptStashStore.ts` is now ~140 lines smaller than at its peak. Future cross-tab review findings on this store are working-as-intended per app convention.

Web typecheck clean, 1539 unit tests passing, lint clean.

PR #4485, opened 2026-07-24
fix: clean up the remote server update reconnect flow

t3dotgg:
Clicking **Update server** on a relay-connected server showed a spinner, then a disconnected banner with a raw transport error, then hung until the page was refreshed.

Three separate causes, fixed independently — the last two improve every reconnect, not just updates.

### 1. The update RPC never acknowledged on the systemd path

`selfUpdate.ts` ran `systemctl --user restart` synchronously inside the RPC handler, so the process died before the acknowledgement flushed. Every *successful* update reached the client as an interrupt, which `ServerUpdateAction` released quietly — no success toast, and nothing told the connection layer a restart was coming. (The respawn path already deferred its exit by 2s, so it usually won that race.)

The restart is now deferred the same way. The trade-off: a rejected restart can no longer be reported through the already-acknowledged RPC, so rollback and the in-flight reset move onto the detached fiber and log instead. Rollback behavior itself is unchanged and still covered by tests.

### 2. Backoff left the client asleep when the server returned

Reconnect backoff escalates 1s → 2s → 4s → 8s → 16s, and only resets after 30s of *stable* connection — so a restart could even inherit backoff debt from earlier flakiness. The client was frequently asleep in a 16s delay exactly when the server came back. Refreshing reset the failure count, which is why refreshing "fixed" it.

A restart-expected overlay (`serverRestartStore.ts`) is armed **before dispatch** and retries on a flat 2s cadence while active. It's deliberately *not* a new supervisor phase — the transport state machine stays transport-only and this is UI-side intent that expires on its own after 90s. Two safeguards worth noting:

- It only pokes `retryNow` while `retryAt` is set (supervisor sleeping in backoff); waking a live attempt would cancel and restart it.
- Arming before dispatch, rather than treating an interrupt as proof of acceptance, avoids showing "restarting" for a request the server never accepted. An explicit RPC failure withdraws it, since that proves the server stayed up.

### 3. Reconnect threw away the cached DPoP token

Two consecutive websocket-ticket failures evicted the cached token, forcing a full relay bootstrap + token exchange during reconnect. Eviction now only happens for failures that suggest a *stale endpoint*; plain unreachability (`network`/`timeout`) — the exact shape of a restart — keeps the token, so reconnect is a single request.

### Also: raw transport errors no longer reach the banner

`Failed to fetch remote environment endpoint https://prod-…t3coderelay.com/api/auth/websocket-ticket (HttpClientError: …)` was flowing verbatim from `rpc/http.ts` into the composer. Presentation now maps each failure reason to a stable summary and carries `reason` for callers; the raw message stays on the supervisor state for connection diagnostics. Authored blocked-failure messages (auth/permission/config) pass through unchanged since they're already actionable.

## Result

Click Update → button stays "Updating…" → one calm "bb-1 is restarting" banner → reconnects the moment the server is back, typically 10–20s, no refresh. The error UI still appears if the server genuinely never returns.

## Testing

- Full suite green: 4,218 passing, 0 failures. Typecheck and lint clean across all 15 projects.
- New coverage: deferred boot-service restart + rollback-after-deferred-failure, transport-message sanitization, blocked-message passthrough, and cached-token retention under pure unreachability.
- Not verified by running the flow end-to-end against a live relay server — the behavior above is reasoned from the code and covered by unit tests. Worth a manual update against a real remote before merging.

---

> [!NOTE]
> **Medium Risk**
> Touches self-update restart timing, in-flight guards, and reconnect/auth token eviction—behavior changes during server restarts and remote updates, though heavily test-covered.
> 
> **Overview**
> Fixes the remote **Update server** flow so clients get a successful RPC before the process dies, then reconnect calmly instead of showing raw transport errors and sleeping in backoff.
> 
> **Server:** Boot-service self-update now **defers** `systemctl restart` (like respawn) so the update RPC can acknowledge first. Failed or defective deferred restarts roll back the unit, log, and **release the in-flight guard** so another update can be tried; success no longer surfaces as an RPC interrupt.
> 
> **Client:** Adds a per-environment **restart-expectation window** (`serverRestartStore`) plus app-level **`useServerRestartReconnect`** (mounted in `AppRoot`) that pokes fast retries while the supervisor is in backoff (`retryAt` set), without waking in-flight attempts. **`ServerUpdateAction`** arms/re-arms the window around dispatch and success/interrupt, clears it on real failures (but not on “already in progress”), and releases the pending button when restart resolves without a new version. **Chat** and **Connections** show a neutral “restarting” state and hide the version-mismatch update action during an expected restart.
> 
> **Reconnect/auth:** Cached DPoP tokens are **not** evicted on pure `network`/`timeout` unreachability—only on reasons that imply a stale endpoint. Connection presentation adds **`reason`** and **`retryAt`**, and maps transport failures to **stable user-facing summaries** (URLs stay in diagnostics). Mobile/tests pick up the expanded connection shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83faf56b65cf7973045312a0d1d79993c0275143. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Clean up the server self-update reconnect flow with expected-restart tracking and proactive reconnection
> - Adds [serverRestartStore.ts](https://github.com/pingdotgg/t3code/pull/4485/files#diff-88749cff4b4dbc4467fa04322a43c2daeabf439d43b5ca3cf96d233fcdf2e354) — a zustand store tracking per-environment restart expectation windows (90s), distinguishing expected restarts from unrelated outages
> - Adds [useServerRestartReconnect.ts](https://github.com/pingdotgg/t3code/pull/4485/files#diff-f74312c7d30f72824ad41ad774ceba2686612d4619496943864ac72209b3b1d7) — a global hook that pokes reconnection on a fast-then-easing cadence while supervisors are in backoff, and clears the window once the environment reconnects
> - Mounts `ServerRestartReconnectHost` app-wide in [AppRoot.tsx](https://github.com/pingdotgg/t3code/pull/4485/files#diff-21926ebb90f26239608d00b3283a87aac57ccce9e94d529a524ce198d179567b) to drive reconnection during expected restarts
> - `ServerUpdateAction` arms the restart window before dispatch, re-arms on RPC success or interrupt, and only clears on real failures (not `already running` rejections)
> - [ChatView.tsx](https://github.com/pingdotgg/t3code/pull/4485/files#diff-4b49e092ccd43be0f0de24abe85ba522e09f04288a5d84253b0263e1a389400e) shows a spinner banner during expected restarts and suppresses the version-mismatch update prompt to avoid double-triggering
> - Server-side [selfUpdate.ts](https://github.com/pingdotgg/t3code/pull/4485/files#diff-ec2e07779fe6544111c93dbc73b0783e8d6f7a09caa236f543bd500d08c4617b) defers the `systemctl restart` asynchronously, rolls back the systemd unit on failure, and releases the in-flight guard to allow retries
> - Authorization layer retains cached tokens on pure network/timeout transients, avoiding unnecessary re-bootstrap during restarts
> - Connection presentation now exposes `reason` and `retryAt` fields to consumers
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 83faf56.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/features/review/ReviewSheet.tsx
+++ b/apps/mobile/src/features/review/ReviewSheet.tsx
@@ -771,7 +771,9 @@ export function ReviewSheet(props: ReviewSheetProps) {
                 environment.presentation?.connection ?? {
                   phase: "available",
                   error: null,
+                  reason: null,
                   traceId: null,
+                  retryAt: null,
                 }
               }
               resourceName="review"
--- a/apps/mobile/src/features/terminal/ThreadTerminalRouteScreen.tsx
+++ b/apps/mobile/src/features/terminal/ThreadTerminalRouteScreen.tsx
@@ -1209,7 +1209,9 @@ export function ThreadTerminalRouteScreen(props: ThreadTerminalRouteScreenProps)
               environment.presentation?.connection ?? {
                 phase: "available",
                 error: null,
+                reason: null,
                 traceId: null,
+                retryAt: null,
               }
             }
             resourceName="terminal"
--- a/apps/mobile/src/state/workspaceModel.test.ts
+++ b/apps/mobile/src/state/workspaceModel.test.ts
@@ -40,7 +40,9 @@ function environment(
     connection: {
       phase,
       error: phase === "error" ? "Connection failed." : null,
+      reason: phase === "error" ? "authentication" : null,
       traceId: phase === "error" ? "trace-1" : null,
+      retryAt: null,
     },
     serverConfig: null,
   };
--- a/apps/server/src/cloud/selfUpdate.test.ts
+++ b/apps/server/src/cloud/selfUpdate.test.ts
@@ -26,6 +26,30 @@ import * as SelfUpdate from "./selfUpdate.ts";
 
 const NODE_PATH = "/usr/local/bin/node";
 
+/**
+ * The deferred restart runs on a detached fiber, so its rollback performs
+ * real filesystem I/O that advancing the TestClock does not await. Poll on
+ * real time until the expected content lands.
+ */
+const eventuallyFileString = Effect.fn("test.eventuallyFileString")(function* (
+  filePath: string,
+  expected: string,
+) {
+  const fs = yield* FileSystem.FileSystem;
+  let last = "";
+  for (let iteration = 0; iteration < 1_000; iteration += 1) {
+    last = yield* fs.readFileString(filePath);
+    if (last === expected) {
+      return last;
+    }
+    // Yielding (rather than sleeping) keeps this on the real event loop:
+    // TestClock would only advance virtual time and never let the
+    // detached fiber's filesystem writes land.
+    yield* Effect.yieldNow;
+  }
+  return yield* Effect.die(new Error(`Expected file contents were not observed at ${filePath}.`));
+});
+
 interface RecordedCommand {
   readonly command: string;
   readonly args: ReadonlyArray<string>;
@@ -490,7 +514,7 @@ it.layer(NodeServices.layer)("ServerSelfUpdate.update", (it) => {
     }).pipe(Effect.provide(TestClock.layer())),
   );
 
-  it.effect("rewrites the systemd unit and restarts the boot service", () =>
+  it.effect("rewrites the systemd unit and defers the boot-service restart", () =>
     Effect.gen(function* () {
       const context = yield* makeContext({ bootService: true });
       const result = yield* context.service.update({ targetVersion: "0.0.29" });
@@ -504,12 +528,15 @@ it.layer(NodeServices.layer)("ServerSelfUpdate.update", (it) => {
         context.path.join(context.home, ".config", "systemd", "user", "t3code.service"),
       );
       assert.include(unit, `ExecStart=${NODE_PATH} ${pinnedEntry} serve`);
+      // The restart is deferred so the RPC acknowledgement flushes before
+      // systemd stops this process.
       assert.deepEqual(
         context.commands.map((entry) => entry.command),
-        ["npm", NODE_PATH, "systemctl", "systemctl"],
+        ["npm", NODE_PATH, "systemctl"],
       );
       assert.deepEqual(context.commands[2]?.args, ["--user", "daemon-reload"]);
 
+      yield* TestClock.adjust(Duration.seconds(10));
       assert.deepEqual(context.commands[3], {
         command: "systemctl",
         args: ["--user", "restart", "t3code.service"],
@@ -520,7 +547,7 @@ it.layer(NodeServices.layer)("ServerSelfUpdate.update", (it) => {
     }).pipe(Effect.provide(TestClock.layer())),
   );
 
-  it.effect("restores the previous unit and permits a retry when systemd restart fails", () =>
+  it.effect("restores the previous unit and permits a retry when the deferred restart fails", () =>
     Effect.gen(function* () {
       let failRestart = true;
       const context = yield* makeContext({
@@ -542,9 +569,12 @@ it.layer(NodeServices.layer)("ServerSelfUpdate.update", (it) => {
       );
       const previousUnit = yield* context.fs.readFileString(unitPath);
 
-      const first = yield* context.service.update({ targetVersion: "0.0.29" }).pipe(Effect.flip);
-      assert.include(first.reason, "Restarting the systemd boot service failed");
-      assert.equal(yield* context.fs.readFileString(unitPath), previousUnit);
+      // The RPC acknowledges before the restart runs, so a rejected
+      // restart can only roll back and log — never fail the request.
+      const first = yield* context.service.update({ targetVersion: "0.0.29" });
+      assert.deepEqual(first, { targetVersion: "0.0.29", method: "boot-service" });
+      yield* TestClock.adjust(Duration.seconds(10));
+      yield* eventuallyFileString(unitPath, previousUnit);
       assert.deepEqual(
         context.commands.slice(-2).map((entry) => entry.args),
         [
--- a/apps/server/src/cloud/selfUpdate.test.ts
+++ b/apps/server/src/cloud/selfUpdate.test.ts
@@ -62,6 +62,8 @@ const makeRecordingRunnerLayer = (
     readonly stdoutFor?:
       | ((command: string, args: ReadonlyArray<string>) => string | undefined)
       | undefined;
+    /** Raises a defect (not a typed failure) to exercise unexpected-error paths. */
+    readonly dieWhen?: ((command: string, args: ReadonlyArray<string>) => boolean) | undefined;
   },
 ) =>
   Layer.succeed(
@@ -70,6 +72,9 @@ const makeRecordingRunnerLayer = (
       run: (input) =>
         Effect.sync(() => {
           commands.push({ command: input.command, args: input.args });
+          if (options?.dieWhen?.(input.command, input.args) === true) {
+            throw new Error(`${input.command} died unexpectedly`);
+          }
           const failed = options?.failWhen?.(input.command, input.args) === true;
           const versionFromPath =
             input.command === NODE_PATH && input.args[1] === "--version"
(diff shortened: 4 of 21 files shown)


t3dotgg:
Thanks — all three findings were real and are fixed in 7aecc5d47.

**1. Restart window armed only at click time** (Macroscope + Bugbot both flagged this)

Correct, and the component's own comment about installs consuming the request window is what made it likely in practice. The window is now re-armed from the acknowledgement *and* from the interrupt path — mirroring what `keepPendingForRestart` already did for the button expiry. The pre-dispatch arming stays, since it's what covers the boot-service ack losing the race against its own disconnect.

**2. Deferred restart defects leave `inFlight` set**

Correct — `Effect.catch` only sees typed failures, so a defect skipped the reset and left a live process rejecting every further update. The guard is now released from `Effect.onExit`, outside the typed handler.

One deliberate asymmetry: on **success** the guard is intentionally left set, because systemd is about to stop the process and a second update in that window would race the handoff. Only failure/defect/interrupt releases it.

I verified the new regression test actually catches this — with the `onExit` block removed it fails with exactly `ServerSelfUpdateError: A server update is already in progress.`, the stranded-guard symptom described.

**3. In-progress error clears restart overlay**

Correct, and the sharpest of the three: that failure is the one case where a restart genuinely *is* still pending. Rather than match on the message text, the reason moved into contracts as `SERVER_SELF_UPDATE_ALREADY_RUNNING_REASON` with an `isAlreadyRunning` helper, so server and client agree structurally and a future wording change can't silently break the check.

Full suite green (4,891 passing), typecheck and lint clean. Each fix has a regression test.

Still worth a manual update against a live remote before merge — the end-to-end flow isn't something the unit tests can cover.

t3dotgg:
An independent gpt-5.6-sol review round found a defect the bots and I both missed, and it was the important one: **the restart overlay never actually engaged.** Fixed in 02abc2cb3.

### The overlay was inert

`ChatView` cleared the expectation on any render where the connection was `connected`. But the window is armed *while the old server is still connected* — that's the entire point, since the acknowledgement races its own disconnect. So arming the window triggered a render that destroyed it within milliseconds, long before the server went away. Both re-arm points had the same problem, and the server's deliberate 2s `RESTART_DELAY` gave React ample time to clear it again.

Net effect: every code path in this PR was in place, and the user still got the outage banner and escalating backoff. The feature was decorative.

The window now tracks whether the restart's disconnect was actually **observed**. A connected environment only ends the window once that's happened; before it, "connected" just means the restart hasn't started yet. Re-arming preserves an already-seen disconnect, so a late-acknowledging slow install doesn't reset that.

Worth noting this is exactly the class of bug the unit tests couldn't catch — `ServerUpdateAction`'s tests mock the restart store, so they verified the arming calls while the real interaction lived in a different component.

### Follow-on: action stuck disabled after a rejected restart

Same review, also confirmed. Since a rejected restart can no longer report through the already-acknowledged RPC, the update action could sit disabled for the full 12-minute safety window with the mismatch still valid and the server answering again. A *settled* restart window (disconnect seen, window since closed) now releases it for an immediate retry.

### On testability

The restarting/settled decisions are pure exported predicates rather than inline effect logic, because the component test harness stubs `useEffect` to a no-op and cannot exercise effect-driven behavior. I'd rather have the decision covered than a test that renders and asserts nothing. Both predicates are tested, and I verified both tests fail against the old logic — reintroducing the `connected`-clears-immediately behavior fails `does not claim a still-connected environment is restarting`.

Full suite green (4,901 passing), typecheck and lint clean.

The earlier caveat stands and now matters more: this needs one manual update against a live remote before merge. Three review rounds found three real defects in the client-side coordination, and the end-to-end path is the part no test here covers.

t3dotgg:
All three fixed in 3c9f1db60 — but not individually, because they shared a root cause worth naming.

### Root cause: the window had two jobs

I'd given the restart window a second responsibility — releasing the update action — while the window itself deleted the evidence that job depended on. That's why three separate symptoms appeared:

- **Settled never releases failed restart** (High): a restart that fails *without* dropping the connection never sets `sawDisconnect`, so nothing releases the action.
- **Reconnect clears before settle**: clearing on reconnect drops `sawDisconnect` before the action can read it.
- **Install blip clears restart window**: any transient non-connected phase poisons the flag, and the subsequent reconnect then clears the window before the real restart.

Patching each would have preserved the coupling and invited a fourth. So `sawDisconnect` is gone.

### The window is now write-once-and-expire

It's armed while the old server is still connected, so **nothing observed on the connection can reliably end it early** — that was the false premise in my previous fix. Instead, presentation gates on the environment actually being unavailable, which makes a window outliving its restart inert rather than wrong. That removes the blip and clear-before-settle problems structurally; there's no flag left to poison.

### Releasing the action is its own signal

Independent of the window, covering both shapes a rejected restart can take:

1. The environment went away and came back → something restarted but the skew is still here, so it didn't carry the new version.
2. The window lapsed with the connection never dropping → no restart ever happened (systemd refused and rolled back).

**#2 is the High-severity case** — previously unreachable, since nothing set the flag.

### Verification

`isExpectedRestartResolved` is a pure exported predicate (the component harness stubs `useEffect`, so effect logic isn't testable in place). I removed the lapsed-window branch and confirmed it fails exactly `resolves when the window lapses without the connection ever dropping`, then restored.

Full suite green (4,927 passing), typecheck and lint clean.

---

Tally so far: **4 review rounds, 10 confirmed defects, all in the client-side coordination this PR adds.** Three of those rounds found something the previous round missed, and one found that the feature was entirely inert. I'd treat the server-side changes (deferred restart, DPoP eviction, error sanitization) as solid — they're small and well covered — but the restart-overlay coordination has earned skepticism, and it still has not been run once against a live remote. Please do that before merging rather than trusting the green checks.

t3dotgg:
Both fixed in 85b8a3845, plus a correction to my own previous round.

**Stale `wentAwayRef`** (Macroscope) — confirmed, and the suggested fix was right. The ref lives for the component's lifetime, so an outage from before the click satisfied the resolve predicate on the next render, re-enabled the button mid-update, and allowed a second dispatch. Both refs now reset when an attempt is armed, so only outages during that attempt count.

**Unrelated outage shown as a restart** (Bugbot) — also confirmed, and this one is my regression. Making the window write-once-and-expire last round traded one bug for another: any disconnect inside the 90s was dressed up as an intentional restart.

The window now ends once the environment has **gone away and come back** — the restart it was armed for has been seen through. That's the narrow version of what my first attempt got wrong by ending it on any healthy connection; the difference is requiring the disconnect first. The observation feeds only the banner. The update action's release path stays independent, because coupling those is exactly what produced the earlier family of bugs where each consumer destroyed the other's evidence.

### Two things I got wrong that reviews didn't catch

**I committed a test that verified nothing.** The regression test in the previous commit passed identically with and without its fix — the component harness stubs `useEffect`, so the effect under test never ran. Removed rather than left as false assurance. Worth flagging because the "19 passing" number in my earlier comments was carrying it.

**I started implementing a backoff change that was wrong, and reverted it.** Reviewing my own plan I noticed item 4 — resetting `failureCount` on post-connection drops — was never implemented. On writing it, `keeps escalating backoff when a newly opened session flaps` failed, and that test is right: a flapping session *should* escalate or the client hammers a sick server. The 2s retry loop already covers the restart case without weakening that protection, so the supervisor is deliberately unchanged in this PR. Flagging it as a scoped-out item rather than a silent omission.

Full suite green (4,928 passing), typecheck and lint clean, supervisor diff empty.

---

Five review rounds, twelve confirmed defects, every one in the client-side coordination this PR adds. Twice now a fix of mine has introduced the next finding. That pattern is the actual signal here: **the server-side changes are small and well covered, but this overlay logic should not be merged on the strength of green CI.** It needs one real Update click against a live remote, watching the banner, the button, and the reconnect timing. I can't do that from here.

t3dotgg:
Preferred: https://github.com/pingdotgg/t3code/pull/4903

PR #4508, opened 2026-07-25
Make ClaudeTextGeneration tests hermetic on Windows

### LINKED ISSUE #4507: [Bug]: ClaudeTextGeneration tests are not hermetic on Windows and invoke the real Claude CLI
### Area

apps/server

### Steps to reproduce

1. On Windows, with a real `claude` CLI installed and authenticated.
2. Check out unmodified `main` at `41a430a8`.
3. Run `vp test run apps/server/src/textGeneration/ClaudeTextGeneration.test.ts`.

### Expected behavior

All 5 tests pass against the stub `claude` script the fixture writes into a temp `bin` directory. No network calls, and the run finishes in a couple of seconds.

### Actual behavior

All 5 tests fail. The stub is never invoked — the real installed Claude CLI runs instead, so assertions compare fixture strings against live model output. The run takes ~134s because every case makes a real API request.

```
FAIL  apps/server/src/textGeneration/ClaudeTextGeneration.test.ts
      > falls back when Claude thread title normalization becomes whitespace-only

Expected: "New thread"
Received: "Name Coding Thread Conversation"

 Test Files  1 failed (1)
      Tests  5 failed (5)
   Duration  134.28s
```

Depending on auth state the failures surface either as live model prose replacing the expected fixture (above), or as live `429` / `Not logged in` errors.

### Cause

Two POSIX-only assumptions in the fixture:

1. **PATH separator.** `ClaudeTextGeneration.test.ts:99` builds the search path with a literal `:`:

   ```ts
   process.env.PATH = `${binDir}:${previousPath ?? ""}`;
   ```

   Windows uses `;`, so `binDir` never becomes its own PATH entry. `path.delimiter` would cover both.

2. **Extension-less executable.** `:27` writes the stub as `bin/claude` with no extension, and `:66` sets `0o755` via `fs.chmod`, which is a no-op on Windows. Windows resolves executables through `PATHEXT`, so an extension-less file is skipped even when its directory is on PATH. A `claude.cmd` shim on Windows would fix this half.

Both need fixing — either alone still leaves the real CLI being picked up.

### Side effect

`:292` creates `.claude-work-test/` under `process.cwd()` and leaves it behind after the run, so the repo is dirty afterwards.

### Impact

Scoped to Windows contributors running this one file, but worth flagging because it fails open rather than closed: instead of erroring on a missing stub, it silently calls the real API and consumes live quota.

### Version or commit

main @ 41a430a8

### Environment

Windows 11, pnpm workspace, Claude Code 2.1.220 installed and authenticated.

mihneaptu:
## What Changed

The `ClaudeTextGeneration` test fixture now works on Windows. The fake `claude` binary is a Node script with a platform-appropriate entry point — `claude.cmd` on Windows, a `#!/bin/sh` shim elsewhere — and PATH is built with the platform's separator instead of a hardcoded `:`.

One file, test-only. No production code changes.

## Why

Fixes #4507.

On Windows all 5 tests in this file fail on unmodified `main`. The fixture never takes effect, so the real installed Claude CLI runs instead: assertions compare fixture strings against live model output, and the run makes real billed API calls.

Two POSIX-only assumptions caused it:

1. PATH was joined with a literal `:`. Windows uses `;`, so the fixture's `bin` directory never became its own PATH entry.
2. The stub was written as extension-less `bin/claude` containing a `#!/bin/sh` script. Windows resolves executables through `PATHEXT`, so an extension-less file is skipped — and even with a correct extension a shebang script is not executable there.

Fixing only the separator is not enough, since the shebang stub still would not run. So the stub logic moved into Node (`claude-stub.mjs`) with a thin platform-specific launcher in front of it. The `T3_FAKE_CLAUDE_*` env contract and every exit code are unchanged.

No production change was needed: `resolveSpawnCommand` already resolves through `PATHEXT` and spawns `.cmd` with `shell: true`.

## Validation

Windows 11, this file only:

| | result |
| --- | --- |
| before | 5 failed, 134.28s (live API calls) |
| after | 5 passed, 4.09s (stub invoked) |

The 134s to 4s drop is the signal that no network calls happen anymore.

`vp fmt`, `vp lint --report-unused-disable-directives`, `git diff --check`, and `pnpm --filter t3 typecheck` are all clean for this scope.

I do not have a POSIX machine, so the shell shim is verified behaviourally under `sh` rather than by running the suite: happy path, args-must-contain pass and fail, stdin-must-contain pass and fail, custom exit code, and config-dir mismatch all return the same exit codes as the old script (0, 0, 2, 4, 0, 7, 5). The full suite on Linux and macOS is unverified by me and worth confirming in CI.

One note so it is not a surprise: on Windows the run now emits `DEP0190` (passing args with `shell: true`). That comes from the existing `resolveSpawnCommand` path and is not introduced here — it simply was not reachable from this test before, because the stub never resolved.

## Not included

`ClaudeTextGeneration.test.ts` builds `claudeConfigDir` from `process.cwd()`, which leaves a `.claude-work-test/` directory behind in the repo after the run. It is recorded in #4507 and left out here to keep this focused.

## Checklist

- [ ] I included before/after screenshots for any UI changes — n/a, no UI changes

---

> [!NOTE]
> **Low Risk**
> Changes are confined to one test file; no production code paths are modified.
> 
> **Overview**
> **Makes `ClaudeTextGeneration` tests use a portable fake Claude CLI** so Windows runs hit the stub instead of the real CLI on PATH (which caused failures, live API calls, and slow runs).
> 
> The test fixture’s fake `claude` binary is no longer a single `#!/bin/sh` script. Stub logic lives in **`claude-stub.mjs`** (same `T3_FAKE_CLAUDE_*` env checks and exit codes). **Windows** gets **`claude.cmd`** that forwards to Node; **Unix** keeps a small executable shim that `exec`s the same stub.
> 
> **`withFakeClaudeEnv`** prepends the fake `bin` to **`PATH`** using **`;` on Windows** and **`:` elsewhere**, so the fixture directory is actually discoverable on Windows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68d1bb1501b937c125540cf4fc6a5f4a45708583. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Make `ClaudeTextGeneration` tests hermetic on Windows
> - Replaces the POSIX shell-based fake `claude` binary with a Node stub (`claude-stub.mjs`) that replicates the same validation logic using `process` APIs.
> - On Windows, a `claude.cmd` batch file invokes the Node stub; on non-Windows, a shell launcher script is written and marked executable.
> - Fixes PATH assembly in `withFakeClaudeEnv` to use `;` as the delimiter on Windows instead of `:`.
> - Platform detection uses `isHostWindows` from `@t3tools/shared/hostProcess`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 68d1bb1.</sup>
> 
>

### DIFF
--- a/apps/server/src/textGeneration/ClaudeTextGeneration.test.ts
+++ b/apps/server/src/textGeneration/ClaudeTextGeneration.test.ts
@@ -1,13 +1,13 @@
-import { isHostWindows } from "@t3tools/shared/hostProcess";
-import { ClaudeSettings, ProviderInstanceId } from "@t3tools/contracts";
 import * as NodeServices from "@effect/platform-node/NodeServices";
 import { it } from "@effect/vitest";
+import { ClaudeSettings, ProviderInstanceId } from "@t3tools/contracts";
+import { isHostWindows } from "@t3tools/shared/hostProcess";
+import { createModelSelection } from "@t3tools/shared/model";
 import * as Effect from "effect/Effect";
 import * as FileSystem from "effect/FileSystem";
 import * as Layer from "effect/Layer";
 import * as Path from "effect/Path";
 import * as Schema from "effect/Schema";
-import { createModelSelection } from "@t3tools/shared/model";
 import { expect } from "vite-plus/test";
 
 import * as ServerConfig from "../config.ts";
@@ -77,7 +77,7 @@ function makeFakeClaudeBinary(dir: string) {
         "}",
         "",
         'process.stdout.write(process.env.T3_FAKE_CLAUDE_OUTPUT ?? "");',
-        "process.exit(Number(process.env.T3_FAKE_CLAUDE_EXIT_CODE ?? 0));",
+        "process.exitCode = Number(process.env.T3_FAKE_CLAUDE_EXIT_CODE ?? 0);",
         "",
       ].join("\n"),
     );
--- a/apps/server/src/textGeneration/ClaudeTextGeneration.test.ts
+++ b/apps/server/src/textGeneration/ClaudeTextGeneration.test.ts
@@ -1,3 +1,4 @@
+import { isHostWindows } from "@t3tools/shared/hostProcess";
 import { ClaudeSettings, ProviderInstanceId } from "@t3tools/contracts";
 import * as NodeServices from "@effect/platform-node/NodeServices";
 import { it } from "@effect/vitest";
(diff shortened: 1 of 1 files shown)


CDVolvik (commented):
5/5 still green on Linux, so the POSIX path is unaffected by the rewrite. Moving the stub into Node and keeping a thin `.cmd` entry point is the right shape, and the `PATHEXT` note in the comment is the part most people miss.

One thing to change while it is being touched. The stub ends with:

```js
process.stdout.write(process.env.T3_FAKE_CLAUDE_OUTPUT ?? "");
process.exit(Number(process.env.T3_FAKE_CLAUDE_EXIT_CODE ?? 0));
```

`process.stdout.write` to a pipe is asynchronous, and `process.exit` does not flush what is still queued. The shell version did not have this problem because `printf` blocks until the write completes, so this is a behaviour the port introduces rather than inherits.

Measured with the same write-then-exit shape, payload generated inside the child so nothing large goes through the environment:

```
payload=    1024   process.exit():     1024   |   process.exitCode:     1024
payload=   65536   process.exit():    65536   |   process.exitCode:    65536
payload=  262144   process.exit():   182720  TRUNCATED   |   process.exitCode:   262144
payload= 1048576   process.exit():   146176  TRUNCATED   |   process.exitCode:  1048576
payload= 4194304   process.exit():   146176  TRUNCATED   |   process.exitCode:  4194304
```

It is latent today rather than broken: `T3_FAKE_CLAUDE_OUTPUT` arrives through the environment, so current outputs sit under the threshold. But the cutoff lands between 64KB and 256KB, and it depends on pipe buffer state rather than on anything in the test, so the first person who needs a larger fixture gets a partial read with no error and a passing-then-failing test. That is an awkward thing to debug in a file whose stated purpose is hermeticity.

`process.exitCode = Number(...)` instead of `process.exit(...)` fixes it and still yields the right exit status, since the stub already drains stdin to EOF and has nothing else keeping the loop alive. The `fail()` helper has the same shape, though its messages are short enough that it will not bite in practice.

Minor: the new import sits above the `@t3tools/contracts` line rather than in sorted position, which may or may not matter depending on whether the import ordering rule is enforced here.

Issue #4589, opened 2026-07-26
[Bug]: An environment's threads silently stop updating until the client is restarted


colonelpanic8:
> **Update — root cause identified.** This is not specific to settling. A client's subscriptions to an environment can die permanently after a transport blip while the connection stays healthy, leaving that environment's threads frozen until the client restarts. "Settle does nothing" was just the first reachable symptom. See the [confirmed server-side evidence](https://github.com/pingdotgg/t3code/issues/4589#issuecomment-5086343764), the [mechanism in `subscribeDynamic`](https://github.com/pingdotgg/t3code/issues/4589#issuecomment-5086367448), and the [dead-subscription vs dead-connection asymmetry](https://github.com/pingdotgg/t3code/issues/4589#issuecomment-5086395043). The original report follows unchanged.

---

## Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

## Area

apps/web

## Steps to reproduce

Observed multiple times, no deterministic repro yet. The pattern:

1. Desktop app, thread that the server already considers settled.
2. The thread shows in the unsettled/active group anyway.
3. Click Settle on it.
4. Nothing happens — it stays showing as unsettled.

## Expected behavior

A settled thread displays as settled. Failing that, clicking Settle converges the view — or says why it can't.

## Actual behavior

The thread displays as unsettled while the server considers it settled, and **clicking Settle does nothing at all** — no state change, no toast, no error, no feedback of any kind. There is no way to move the thread into the settled group from the UI. Repeated clicks are equally silent.

**Restarting the desktop client clears it.**

## Technical detail

The command round-trip is itself a state sync: `threadUpsertOrRemove` (`apps/server/src/ws.ts:725-750`) re-reads the *current* projected thread row and emits a `thread-upserted` carrying it after any thread event — including the idempotent settle re-emit below. So a settle sent against a client that disagrees with the server should return the authoritative row and converge the client.

Two candidate explanations, in order of how well they fit "nothing visibly happens":

[...]

PR #4602, opened 2026-07-27
fix(client-runtime): recover subscriptions that lose their transport

colonelpanic8:
## What Changed

A durable subscription that fails with a transport-level `RpcClientError` now asks the supervisor to re-establish the session, instead of draining and waiting for a session that may never come.

Requests are coalesced per session via a `WeakSet` keyed on the session object, so a burst of subscription failures produces exactly one reconnect request.

## Why

`subscribeDynamic` handles a transport failure by logging and draining, deferring to "the next session". Nothing guarantees one arrives. `supervisor.session` is written only when a connection lease is established (`connection/supervisor.ts:532`) or cleared on teardown (`:243`), and the only liveness probe runs on an `application-active` wakeup — a desktop window that stays visible never health-checks its connection.

So a stream that dies while the session stays usable — a half-open socket (NAT idle timeout, sleep/resume, VPN flap, where no close is ever delivered), or a failure confined to that one stream — leaves the subscription dead for the lifetime of the session.

Nothing surfaces it. Unary RPCs on that same session keep working, so the app looks connected and commands still round-trip while the environment's state silently stops updating.

This was diagnosed from a case where client and server were on different machines, so both sides could be inspected independently (#4589). On one connection at one moment:

- ten `thread.settle` commands were accepted server-side, with receipts
- the pre-existing shell subscription was dead — a new thread created during the incident never appeared in the sidebar, though the server had it running
- a thread-detail subscription created *after* the failure worked fine

That asymmetry is the signature: the drain is per subscription stream, so `subscribeToSession()` runs fresh for anything started afterward while already-drained streams stay dead. A connection-level failure could not produce it — it would take out the new subscription too. Other clients against the same server were unaffected throughout.

The user-visible result is an environment frozen in place: threads stuck showing Working long after they finished, new threads missing from the sidebar, and lifecycle actions succeeding as no-ops against a stale view. Only restarting the client recovers it, because a restart is the one thing that makes a new session.

Note that the branch immediately below this one already retries within the same session when `retryExpectedFailureAfter` is set (`state/shell.ts` passes `"250 millis"`). Only the transport branch had no path back.

## Why coalesce

An environment carries many durable subscriptions — the shell, the thread list, every open thread — and a dead transport tends to take all of them at the same moment. Each `retryNow` is a signal the supervisor acts on, and a signal delivered mid-establishment aborts that attempt (`supervisor.ts:362-384`). N unfiltered requests would abort N reconnects and stall recovery exactly when it is needed. The regression test for this fails with `expected 2 to be 1` if the guard is removed.

## Checklist

- [ ] I included before/after screenshots for any UI changes — no UI change

---

One existing test changed its expectation. `"keeps durable subscriptions alive across a transport failure and new session"` asserted `retryCount === 0`, and passed only because the test manually supplies the replacement session that production has no way to produce. It now expects one request.

Validation (scoped per `AGENTS.md`, which reserves repo-wide `vp check` / `vp run typecheck` for CI):

```
vitest run packages/client-runtime/src/    # 473 passed
tsgo --noEmit                              # packages/client-runtime — clean
vp lint --report-unused-disable-directives # no findings in changed files
```

Not addressed here, both worth their own change: unary `request` (`rpc/client.ts:107`) has no timeout, so a call into a hung session never settles — which can latch an in-flight guard in the UI; and there is no periodic liveness probe, only the `application-active` one.

---

> [!NOTE]
> **Medium Risk**
> Changes connection recovery for all durable environment subscriptions; incorrect coalescing or stale-session handling could over- or under-reconnect, but scope is limited to transport failures and is heavily tested.
> 
> **Overview**
> Durable RPC subscriptions that fail with a transport-level `RpcClientError` no longer only log and drain while hoping the supervisor swaps sessions. **`subscribeDynamic`** now calls **`requestSessionRecovery`**, which triggers **`supervisor.retryNow`** when the failing session is still the active one, so reconnect can happen even when the socket looks fine to unary calls.
> 
> Recovery is **coalesced per session** with `WeakSet` guards so many subscriptions dying together produce **one** reconnect signal; stale sessions and in-flight duplicates are ignored. Domain failures still do not reconnect.
> 
> Tests cover single and multi-subscription reconnect, stale-session filtering, resubscribe interrupting recovery, and the updated expectation that transport failure implies one `retryNow`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5836f14b7815f0277d04ebfa8142162647d9195. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `subscribeDynamic` to recover subscriptions that lose their transport
> - When an `RpcClient` subscription fails with a transport error, `subscribeDynamic` now calls `supervisor.retryNow` before draining to await the next session.
> - A new `requestSessionRecovery` effect coalesces recovery signals per session: only one `retryNow` is issued per broken session, and signals from already-replaced sessions are ignored.
> - New tests in [client.test.ts](https://github.com/pingdotgg/t3code/pull/4602/files#diff-cef86a2ddd8a877a388fe59d61004f3479209d6e8ea5e9b00a8fd025f43ac9ed) cover deduplication, stale-session filtering, and in-flight resubscription interleaving.
> - Behavioral Change: subscriptions that previously silently drained on transport loss now actively trigger supervisor reconnection.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b5836f1.</sup>
> 
>
>

### DIFF
--- a/packages/client-runtime/src/rpc/client.test.ts
+++ b/packages/client-runtime/src/rpc/client.test.ts
@@ -5,6 +5,7 @@ import {
 } from "@t3tools/contracts";
 import { describe, expect, it } from "@effect/vitest";
 import * as Cause from "effect/Cause";
+import * as Deferred from "effect/Deferred";
 import * as Effect from "effect/Effect";
 import * as Exit from "effect/Exit";
 import * as Fiber from "effect/Fiber";
@@ -25,7 +26,13 @@ import {
 import * as EnvironmentSupervisor from "../connection/supervisor.ts";
 import * as RpcSession from "../rpc/session.ts";
 import type { WsRpcProtocolClient } from "../rpc/protocol.ts";
-import { EnvironmentRpcRequestObserver, request, runStream, subscribe } from "./client.ts";
+import {
+  EnvironmentRpcRequestObserver,
+  request,
+  runStream,
+  subscribe,
+  subscribeDynamic,
+} from "./client.ts";
 
 const TARGET = new PrimaryConnectionTarget({
   environmentId: EnvironmentId.make("environment-1"),
@@ -53,7 +60,9 @@ function session(client: WsRpcProtocolClient): RpcSession.RpcSession {
   };
 }
 
-const makeHarness = Effect.fn("TestEnvironmentRpc.makeHarness")(function* () {
+const makeHarness = Effect.fn("TestEnvironmentRpc.makeHarness")(function* (options?: {
+  readonly retryNow?: Effect.Effect<void>;
+}) {
   const state = yield* SubscriptionRef.make<SupervisorConnectionState>(AVAILABLE_CONNECTION_STATE);
   const activeSession = yield* SubscriptionRef.make<Option.Option<RpcSession.RpcSession>>(
     Option.none(),
@@ -67,7 +76,7 @@ const makeHarness = Effect.fn("TestEnvironmentRpc.makeHarness")(function* () {
     prepared,
     connect: Effect.void,
     disconnect: Effect.void,
-    retryNow: Ref.update(retryCount, (count) => count + 1),
+    retryNow: options?.retryNow ?? Ref.update(retryCount, (count) => count + 1),
   } satisfies EnvironmentSupervisor.EnvironmentSupervisor["Service"]);
   return {
     activeSession,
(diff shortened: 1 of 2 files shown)


juliusmarminge:

**Reason:** Jul conflicted subscription-recovery; reconnect/subscription work already exists (#1972 and later).
**Related:** #1972

PR #4603, opened 2026-07-27
fix(shared): stop reporting "not a worktree" when .git cannot be read

NoahHendrickson:
Follow-up to #4555. Found while reviewing that commit's behavior in a linked worktree.

## The problem

`resolveGitWorktreePath` swallows every filesystem failure:

```ts
const info = yield* fileSystem.stat(gitPath).pipe(Effect.option);
// ...
const contents = yield* fileSystem.readFileString(gitPath).pipe(Effect.orElseSucceed(() => ""));
```

`Effect.option` turns *any* `stat` error into `None`, so the walk continues upward. `orElseSucceed(() => "")` turns any read error into an empty string, which parses as "no gitdir pointer". Both paths answer `undefined` — *not a worktree*.

That answer isn't harmless in this module specifically. `resolveWorktreeT3Home`'s own docstring makes the argument:

> Deliberately does not require the directory to exist yet: falling back because it is missing **would send callers at the shared home**.

Right — and that's exactly what the error paths do. `dev-runner.ts` falls through to an ambient `T3CODE_HOME`, which `cli/config.ts` treats as an explicit base dir, which selects the `userdata` leaf. So an `EACCES` on a `.git` file, a transient `EBUSY`, or a stalled network mount quietly points a worktree dev server at `~/.t3/userdata` — the installed app's live database. The one outcome the module exists to prevent, reached through the one path that isn't guarded.

## The change

Distinguish absence from failure. Only a `NotFound` reason means "no `.git` here, keep walking":

```ts
const isAbsent = (error: PlatformError.PlatformError): boolean => error.reason._tag === "NotFound";
```

Everything else propagates, so the error channel becomes `PlatformError` instead of `never`. Once `stat` has established the file exists, a read failure is a fault rather than an answer, so `readFileString` propagates too.

`error.reason._tag === "NotFound"` matches existing precedent in the codebase — `DesktopSavedEnvironments.ts:250` and `relayClient.ts:150` do the same thing.

**The trade is deliberate and worth stating:** a dev server now refuses to start on an unreadable `.git` rather than starting against the wrong database. Failing loudly is recoverable; silently opening production state isn't. Happy to make it a warning-and-fall-back instead if you'd rather, but that reintroduces the hazard.

Only one production call site (`dev-runner.ts:514`), and the widened error channel typechecks through it without changes.

## Tests

The fixture named `"unreadable-git-file"` wrote a perfectly readable file containing `not a gitdir pointer` — that's the *unparseable* path. The genuinely unreadable case had no coverage at all. Renamed it, and added a real `chmod 000` case.

**Both new tests fail against the current implementation and pass against this one:**

```
 × resolveGitWorktreePath > fails instead of reporting `not a worktree` when .git cannot be read
 × resolveWorktreeT3Home > propagates an unreadable .git rather than falling back
   Tests  2 failed | 13 passed (15)
```

They're skipped where the mode proves nothing — running as root, and Windows, which has no equivalent — following the `it.live.skipIf` precedent in `orchestrationEngine.integration.test.ts`.

Also covered, all previously untested:

- **relative `gitdir:` pointers** — git ≥ 2.48 with `worktree.useRelativePaths`; the `path.normalize` call exists for this and nothing exercised it
- **Windows `gitdir:` pointers** — the sole reason `replaceAll("\\", "/")` exists
- **the `segments.length >= 3` boundary** — `gitdir: worktrees/x`
- **a repository nested inside a worktree** — the "nesting one repository inside another does not make the outer one this root" comment was documented but unguarded
- **`resolveWorktreeT3Home`'s negative case** — it had one test, and not the safety-relevant branch

Renamed the `"bare"` fixture to `"no-repo"`: it creates no `.git` at all, which read oddly next to `"bare-repo-worktree"`, which genuinely is the bare case.

## Verification

| Check | Result |
| --- | --- |
| `packages/shared` tests | 39 files, 302 tests, pass |
| `scripts` tests | 16 files, 152 tests, pass |
| `typecheck` (all 15 packages) | exit 0 |
| `lint` | exit 0, 0 errors |
| `fmt --check` | clean |

## Not addressed

`pointsAtLinkedWorktree` does `.replaceAll("\\", "/")` before `normalize`, which mangles a POSIX filename containing a literal backslash. Real but vanishingly unlikely, and fixing it properly means branching on platform — left alone rather than bundled into a correctness fix.

---

> [!NOTE]
> **Medium Risk**
> Changes dev data-directory resolution so I/O errors block startup rather than silently using shared/live userdata; behavior change at the dev-runner call site but intentionally safer.
> 
> **Overview**
> **`resolveGitWorktreePath` / `resolveWorktreeT3Home` no longer turn `.git` I/O failures into “not a worktree”.** Only `NotFound` on `stat` continues the parent walk; permission errors, busy mounts, etc. fail the effect. After `stat` proves a `.git` file exists, `readFileString` errors propagate instead of being treated as an empty/unparseable pointer. Both exports now surface `PlatformError` on the error channel.
> 
> Tests split **unparseable** vs **unreadable** `.git` fixtures, add `chmod 000` coverage (skipped on root/Windows), and exercise relative/Windows `gitdir:` pointers, shallow `worktrees/x` paths, nested repos inside worktrees, plus `resolveWorktreeT3Home` fallbacks and error propagation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9bb41383e37901601dac9182eb794bec1e826373. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `resolveGitWorktreePath` to propagate errors when `.git` exists but cannot be read
> - `resolveGitWorktreePath` previously swallowed read errors on `.git`, which could silently report a directory as "not a worktree". It now propagates any non-`NotFound` `PlatformError` instead of returning `undefined`.
> - `resolveWorktreeT3Home` inherits this behavior and will also fail with a `PlatformError` when `.git` is unreadable.
> - A new `isAbsent` helper in [devHome.ts](https://github.com/pingdotgg/t3code/pull/4603/files#diff-77ea98aaec3bbe7740817fc180fa324901af50b98efee5617284c8deaaf9a40b) distinguishes `NotFound` errors (treated as absence) from other failures (propagated).
> - Tests in [devHome.test.ts](https://github.com/pingdotgg/t3code/pull/4603/files#diff-62ba2dbb7ca5ee76582382442eafd51423683f10c3f3c30d13453cae844ee0fa) are extended to cover relative/Windows gitdir pointers, shallow gitdir rejection, nested repo boundaries, and unreadable `.git` propagation; unreadable-file tests are skipped when the platform cannot enforce file permissions.
> - Behavioral Change: callers of `resolveGitWorktreePath` and `resolveWorktreeT3Home` must now handle `PlatformError` failures where they previously never failed.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9bb4138.</sup>
> 
>

### DIFF
--- a/packages/shared/src/devHome.test.ts
+++ b/packages/shared/src/devHome.test.ts
@@ -8,15 +8,27 @@ import * as Effect from "effect/Effect";
 
 import { resolveGitWorktreePath, resolveWorktreeT3Home } from "./devHome.ts";
 
+/**
+ * Denying read on the `.git` file only proves anything for a user the mode
+ * applies to. Root ignores it, and Windows has no equivalent, so the
+ * unreadable case is skipped rather than silently passing there.
+ */
+const canDenyReads = typeof process.getuid === "function" && process.getuid() !== 0;
+
 const makeRepo = (
   kind:
     | "worktree"
     | "checkout"
-    | "bare"
+    | "no-repo"
     | "submodule"
+    | "unparseable-git-file"
     | "unreadable-git-file"
     | "bare-repo-worktree"
-    | "custom-common-dir-worktree",
+    | "custom-common-dir-worktree"
+    | "relative-worktree"
+    | "windows-worktree"
+    | "shallow-gitdir"
+    | "nested-checkout",
 ) =>
   Effect.acquireRelease(
     Effect.sync(() => {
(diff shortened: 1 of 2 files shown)


NoahHendrickson:
Closing this — opened prematurely, not something we're pursuing upstream. Apologies for the noise.

PR #4616, opened 2026-07-27
feat(web): add a reset-all button to keybindings settings

### LINKED ISSUE #4576: [Feature]: reset keybinds to default
### Before submitting

### Area

apps/web

### Problem or use case

Image

some old version of t3 code left behind conflicting keybinds. there is no way to reset keybinds to default from the GUI. 

### Proposed solution

A reset button next to each one would be nice, maybe along with a reset all button somewhere. If the reset button next to each keybind looks too cluttered, then maybe include it in the popup which appears when I click the keybind in the settings

### Why this matters

Currently this breaks without a nice UX way to resolve it. I dont want to manually resolve keybind conflicts or throw my agent at some json to keep working keybinds

### Smallest useful scope

add a `reset all` button

### Alternatives considered

Actual migrations for the keybinds in the config file. If something has not been changed from the default by a user, then it should change automatically if the upstream default changes. 

### Risks or tradeoffs

migrations are more work to implement, ideally both a reset button and migrations would be there

### Examples or references

look at how firefox about:config does this

Image

makes messing with the config a whole lot less scary

### Contribution

m-de-graaff:
Closes #4576

## Problem

Keybindings settings can reset a single row to its default, but there is no way to get the whole set back. A user carrying customizations from an older version can end up with conflicting bindings and no recourse short of hand-editing `keybindings.json` — which is exactly what someone in that state does not want to do, and which the issue calls out directly.

## Change

- **`server.resetKeybindings`** RPC → `Keybindings.resetKeybindingRulesToDefaults`, which rewrites the config with `DEFAULT_KEYBINDINGS`, refreshes the cache, and emits the change event like the existing upsert/remove paths.
- **Reset button** in the Keybindings settings header, beside Add keybinding and Open keybindings.json. It confirms first, since it discards every customization at once.

Two deliberate behaviors, both covered by tests:

- **Project script bindings (`script.*`) are kept.** They have no default to restore, so dropping them would delete shortcuts the reset cannot give back. The confirm dialog says so.
- **A config that fails to parse is replaced outright** instead of erroring. A malformed `keybindings.json` is the state this button exists to escape, so refusing to overwrite it would defeat the feature. `upsertKeybindingRule` keeps its existing refuse-to-overwrite behavior.

## Tests

`apps/server/src/keybindings.test.ts`:

- reset restores the defaults, drops customizations, and keeps the `script.*` rule;
- reset replaces a malformed config with the defaults.

`vp test run src/keybindings.test.ts` — 23 passed. (`fails when config directory is not writable` fails on this Windows host both with and without this change: `fs.chmod` does not restrict the directory there. Verified against a clean tree.)

Typecheck clean for `packages/contracts`, `packages/client-runtime`, `apps/server`, `apps/web`; lint and format checked on the changed files.

## Verified in the app

Ran the web stack against an isolated `--home-dir`, planted a customized `keybindings.json` (remapped `sidebar.toggle` and `terminal.toggle`, plus a `script.run-tests.run` binding), and clicked the button:

- confirm text: "Reset all keybindings to their defaults? / Every customization is discarded. Project script bindings are kept."
- on disk afterwards: the non-script rules are byte-identical to `DEFAULT_KEYBINDINGS`, the remapped shortcuts are gone, and `script.run-tests.run` survives;
- the table updates live, 42 bindings (41 defaults + the script rule).

## Not included

The issue also floats migrations for defaults that change upstream. That is a separate mechanism and is not part of this change; the startup sync still backfills newly added defaults as it did before.

---

> [!NOTE]
> **Medium Risk**
> Bulk-writes user keybinding config and overwrites malformed files on reset; behavior is intentional but can surprise users who relied on a broken on-disk file.
> 
> **Overview**
> Adds **reset all keybindings to defaults** end-to-end: a new `server.resetKeybindings` RPC (orchestration operate scope) wired through contracts, client-runtime, and the Keybindings settings header.
> 
> Server-side **`resetKeybindingRulesToDefaults`** rewrites `keybindings.json` with `DEFAULT_KEYBINDINGS`, keeps `script.*` rules, updates cache, and broadcasts changes like upsert/remove. **Malformed configs are replaced** on reset (unlike upsert, which still refuses to overwrite). **Unreadable configs fail** without writing. If script rules exceed the cap after reset, **defaults are never dropped**—oldest script entries are trimmed and a warning is logged.
> 
> The settings UI adds a **Reset all to defaults** control with confirmation (notes that script bindings stay), guards against environment changes while the dialog is open, and surfaces errors via toasts.
> 
> **Tests** cover happy path, malformed replacement, read failure, and max-count script truncation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e832d8a8a4eeea3c4f6ab127f8c02eac13063e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add a reset-all button to keybindings settings
> - Adds a "Reset All" toolbar button in [KeybindingsSettings.tsx](https://github.com/pingdotgg/t3code/pull/4616/files#diff-12068495a79a13f54503fa3061aafb91f05537621626906b290ed059e660d4d4) that prompts for confirmation before resetting keybindings to defaults.
> - Implements `resetKeybindingRulesToDefaults` in [keybindings.ts](https://github.com/pingdotgg/t3code/pull/4616/files#diff-8ceb6d58b5dd5f4fe3318b65f1602637866a7883986552e17df439ac3f7da19e): writes defaults plus any existing script bindings, updates the in-memory cache, and broadcasts a change event.
> - Exposes a new `server.resetKeybindings` WebSocket RPC in [ws.ts](https://github.com/pingdotgg/t3code/pull/4616/files#diff-7f2100e8ffa23a58d9bf425c5b3ceafc39239911ae33bc569291a00ecb8e9125), gated by orchestration operate scope.
> - Script bindings are preserved up to `MAX_KEYBINDINGS_COUNT`; excess script rules are dropped (trailing rules survive) with a warning logged.
> - Risk: malformed keybinding configs are silently overwritten on reset; configs that cannot be read (e.g. a directory at the path) return a failure instead.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1e832d8.</sup>
> 
>
>

### DIFF
--- a/apps/server/src/keybindings.test.ts
+++ b/apps/server/src/keybindings.test.ts
@@ -1,4 +1,9 @@
-import { KeybindingCommand, KeybindingRule, KeybindingsConfig } from "@t3tools/contracts";
+import {
+  KeybindingCommand,
+  KeybindingRule,
+  KeybindingsConfig,
+  MAX_KEYBINDINGS_COUNT,
+} from "@t3tools/contracts";
 import * as NodeServices from "@effect/platform-node/NodeServices";
 import { assert, it } from "@effect/vitest";
 import { assertFailure } from "@effect/vitest/utils";
@@ -496,6 +501,54 @@ it.layer(NodeServices.layer)("keybindings", (it) => {
         persistedView,
         Keybindings.DEFAULT_KEYBINDINGS.map(({ key, command }) => ({ key, command })),
       );
+    }).pipe(Effect.provide(makeKeybindingsLayer())),
+  );
+
+  it.effect("refuses to reset a keybindings config it cannot read", () =>
+    Effect.gen(function* () {
+      const fs = yield* FileSystem.FileSystem;
+      const { keybindingsConfigPath } = yield* ServerConfig.ServerConfig;
+      // A directory at the config path fails the read the same way a transient
+      // filesystem error would, without overwriting rules that are still there.
+      yield* fs.makeDirectory(keybindingsConfigPath, { recursive: true });
+
+      const result = yield* Effect.gen(function* () {
+        const keybindings = yield* Keybindings.Keybindings;
+        return yield* keybindings.resetKeybindingRulesToDefaults;
+      }).pipe(toDetailResult);
+
+      assertFailure(result, "failed to read keybindings config");
+      assert.isTrue((yield* fs.stat(keybindingsConfigPath)).type === "Directory");
+    }).pipe(Effect.provide(makeKeybindingsLayer())),
+  );
+
+  it.effect("keeps every default when preserved script bindings exceed the max count", () =>
+    Effect.gen(function* () {
+      const { keybindingsConfigPath } = yield* ServerConfig.ServerConfig;
+      const scriptBudget = MAX_KEYBINDINGS_COUNT - Keybindings.DEFAULT_KEYBINDINGS.length;
+      const scriptRules = Array.from(
+        { length: scriptBudget + 5 },
+        (_, index): KeybindingRule => ({
+          key: "mod+r",
+          command: `script.s${index}.run`,
+        }),
+      );
+      yield* writeKeybindingsConfig(keybindingsConfigPath, scriptRules);
+
+      yield* Effect.gen(function* () {
+        const keybindings = yield* Keybindings.Keybindings;
+        return yield* keybindings.resetKeybindingRulesToDefaults;
+      });
+
+      const persisted = yield* readKeybindingsConfig(keybindingsConfigPath);
+      const persistedDefaults = persisted
+        .filter((rule) => !String(rule.command).startsWith("script."))
+        .map(({ key, command }) => ({ key, command }));
+      assert.deepEqual(
+        persistedDefaults,
+        Keybindings.DEFAULT_KEYBINDINGS.map(({ key, command }) => ({ key, command })),
+      );
+      assert.equal(persisted.length, MAX_KEYBINDINGS_COUNT);
     }).pipe(Effect.provide(makeKeybindingsLayer())),
   );
 
--- a/apps/server/src/keybindings.test.ts
+++ b/apps/server/src/keybindings.test.ts
@@ -549,6 +549,13 @@ it.layer(NodeServices.layer)("keybindings", (it) => {
         Keybindings.DEFAULT_KEYBINDINGS.map(({ key, command }) => ({ key, command })),
       );
       assert.equal(persisted.length, MAX_KEYBINDINGS_COUNT);
+      // Later rules win, so the surviving scripts must be the trailing ones.
+      assert.deepEqual(
+        persisted
+          .filter((rule) => String(rule.command).startsWith("script."))
+          .map((rule) => rule.command),
+        scriptRules.slice(5).map((rule) => rule.command),
+      );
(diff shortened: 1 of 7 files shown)


m-de-graaff:
All three findings were correct. Fixed in 90f7248d4.

**`Effect.result` swallowed filesystem errors** (flagged by both bots) — the worst of the three. `loadWritableCustomKeybindingsConfig` fails for `fs.exists`, `fs.readFileString`, *and* parse errors, and I collapsed all of them to `[]`. A transient read error would have written defaults over a config whose `script.*` bindings were still on disk — the exact data loss the preserve logic exists to prevent.

The reset now goes through `loadRuntimeCustomKeybindingsConfig`, which already draws that line correctly: an unparseable config yields an empty rule set (with issues), while a read failure propagates. So a malformed config is still replaced — that is the point of the button — but an unreadable one aborts the reset instead of overwriting it.

**The max-count cap dropped defaults, not scripts** — `slice(-MAX)` keeps the tail, so with enough script bindings the reset would fail to restore the defaults it exists to restore. Now the preserved scripts are truncated to `MAX_KEYBINDINGS_COUNT - DEFAULT_KEYBINDINGS.length` and the drop is logged.

**Stale environment id across the confirm dialog** — the handler captured `primaryEnvironment.environmentId` at render and awaited a native modal before sending. It now re-checks the current primary environment after the dialog resolves and bails if it changed.

Two tests added:

- `refuses to reset a keybindings config it cannot read` — a directory at the config path makes the read fail the way a transient filesystem error would; the reset fails and the path is untouched.
- `keeps every default when preserved script bindings exceed the max count` — 220 script rules in, all 41 defaults still present and the file capped at 256.

`vp test run src/keybindings.test.ts` — 25 passed. (`fails when config directory is not writable` still fails on this Windows host with and without the change; `fs.chmod` does not restrict directories there.) Typecheck, lint, and format clean for `apps/server` and `apps/web`.

m-de-graaff:
Both follow-up findings were correct. Fixed in 327f00c1c.

**Script truncation dropped the wrong end.** I capped with `slice(0, budget)`, keeping the earliest scripts — but later rules have higher precedence everywhere else in this file (`mergeWithDefaultKeybindings` and `syncDefaultKeybindingsOnStartup` both keep the tail, and say so). So the cap could discard the script bindings actually in effect and keep the ones they shadow. Now drops from the front.

The existing cap test passed either way, which is why it did not catch this. It now asserts the surviving scripts are exactly the trailing ones, so the direction is pinned.

**Silent abort after confirm.** The environment guard I added in the previous round returned without feedback, so a user who confirmed a reset that then did not run had no way to tell. It now toasts "Keybindings were not reset — the active environment changed while the confirmation was open."

`vp test run src/keybindings.test.ts` — 25 passed, same single pre-existing Windows failure (`fails when config directory is not writable`; `fs.chmod` does not restrict directories on this host, and it fails identically on a clean tree). Typecheck, lint, and format clean for `apps/server` and `apps/web`.

juliusmarminge:

**Reason:** Jul conflicted keybinding reset-all; settings pages were unified later (#9281).
**Related:** #9281

Issue #231, opened 2026-03-07
feat: add Steer and Queue follow-up modes alongside Chat/Plan


Fanzzzd:
Right now T3 Code appears to only model `default` and `plan` interaction modes. I’d like to request support for `Steer` and `Queue` style follow-up behavior while a run is already active.

### Why

When the agent is already working, users often want two different behaviors:

- `Steer`: inject immediate guidance into the active run
- `Queue`: save the follow-up so it runs after the current work settles

This is a meaningful UX distinction and avoids forcing users into interrupt-or-wait behavior only.

### Settings ask

It would be even better if this can be configured in Settings, for example:

- default follow-up behavior while a run is active
- whether the composer shows `Steer` / `Queue` explicitly
- an optional shortcut for sending the non-default behavior once

### Suggested acceptance criteria

- The composer exposes both follow-up modes when a run is active.
- The selected mode is visible and predictable.
- Users can configure the default active-run follow-up behavior in Settings.
- Queued follow-ups survive reconnects/reloads.
- Steering a run does not silently discard already queued follow-ups.

### Relevant code context

- `packages/contracts/src/orchestration.ts`
- `apps/web/src/components/ChatView.tsx`
- `apps/server/src/orchestration/Layers/ProviderCommandReactor.ts`

### Reference implementation ideas

- CodexMonitor already exposes configurable follow-up behavior with `Queue` vs `Steer` while a run is active.

cafeai:
+1 for queue support please...
i typically queue a significant amount of things up when i know i'll be away from the computer

tiagonrodrigues:
+1! Is this WIP?

felixsebastian:
"/btw" would also be sickkk

cafeai:
I ended up making my own fork here: https://github.com/cafeai/cafe-code

Not trying to compete with T3 Code at all — I just wanted something more minimal for my own use, and wanted to be able to fix the bugs I was hitting quickly.

Really appreciate all the work the T3 team has put into the project so far.

<img width="1529" height="967" alt="Image" src="https://github.com/user-attachments/assets/ea9ef651-5f43-4522-aca8-6f172cca5396" />

davidvornholt:
Any plans for implementing a queue?

ralfboltshauser:
not being able to queue is currently preventing me from moving to t3 code from codex.

also one big thing to me for queuing is how it behaves with remote hosts: 
[...]

PR #4621, opened 2026-07-27
fix(orchestrator): surface provider turn interrupt failures in the thread

### LINKED ISSUE #4524: [Bug]: Provider turn interrupt failures are silently swallowed — Stop generation looks successful while the turn keeps running
### Before submitting

### Area

`apps/server` (with a smaller follow-on in `apps/web`)

### Summary

When the provider call behind **Stop generation** fails, the failure is only written to a log line. The orchestration command is still receipted as `accepted`, no thread activity is appended, and the UI shows nothing. From the user's perspective the Stop button is simply dead while the turn keeps running.

The codebase already has the right mechanism for this — it is just not wired to this branch. In `processTurnInterruptRequested`, the "no active provider session" case calls `appendProviderFailureActivity({ kind: "provider.turn.interrupt.failed", ... })`. The case where the provider request itself fails does not.

### Steps to reproduce

1. Start a long-running turn (long tool call, browser automation, etc.).
2. Put the provider in a state where `turn/interrupt` is rejected, so `providerService.interruptTurn` fails with a `ProviderAdapterRequestError`. Any provider-side rejection reaches the same path — see "How I hit it" below for the concrete trigger I observed.
3. Click **Stop generation** (repeatedly, if you like).
4. Watch the thread UI, the activity feed, and `~/.t3/userdata/logs/server-child.log`.

### Expected behavior

A failed interrupt is surfaced to the user — a `provider.turn.interrupt.failed` activity in the thread, or a thread error — consistent with how the "no active provider session" branch already behaves.

### Actual behavior

- The command is receipted `accepted` with no error.
- `thread.turn-interrupt-requested` is persisted normally.
- The provider call fails.
- The only trace is a `logWarning`; the thread gets no error activity and the UI stays silent.
- The turn continues running.

### Code path (upstream/main @ 41a430a88)

- `apps/server/src/orchestration/Layers/ProviderCommandReactor.ts:907` — the interrupt is issued by session only (`interruptTurn({ threadId })`), so the runtime's internal turn id decides the target.
- `apps/server/src/orchestration/Layers/ProviderCommandReactor.ts:1072-1081` — `processDomainEventSafely` catches the cause and terminates it in `Effect.logWarning("provider command reactor failed to process event", …)`. Nothing is emitted to the thread.
- `apps/web/src/components/ChatView.tsx:4830` — `onInterrupt` only calls `setThreadError` when the *dispatch* fails. A dispatch that succeeds and a provider call that later fails is indistinguishable from success in the UI.

Contrast with `ProviderCommandReactor.ts:896-904`, which does the right thing for the missing-session case.

### Impact

Major degradation or frequent failure — when it happens the user has no working way to stop a running agent and no indication why.

### Version or commit

`upstream/main @ 41a430a88` (code paths above read directly from that ref). Observed at runtime on a downstream fork that is based on it.

### Environment

Linux Mint 22.3, kernel 7.0.0-28-generic, Node 22.19.0, desktop AppImage build, Codex provider (`gpt-5.6-sol`), Codex app-server.

### Logs or stack traces

```shell
# ~/.t3/userdata/logs/server-child.log — repeated 16x in 17s, one per click
{
  eventType: 'thread.turn-interrupt-requested',
  cause: 'ProviderAdapterRequestError: Provider adapter request failed (codex) for turn/interrupt: expected active turn id 019f98f9-… but found 019f98f7-…
      at mapCodexRuntimeError (…/apps/server/dist/bin.mjs:66909:9)'
}
```

Meanwhile, in `state.sqlite`:

```shell
-- 16 events, 16 distinct command ids, every receipt accepted, zero receipt errors
SELECT r.status, count(*), sum(r.error IS NOT NULL)
  FROM orchestration_events e
  JOIN orchestration_command_receipts r ON r.command_id = e.command_id
 WHERE e.event_type = 'thread.turn-interrupt-requested';
-- accepted|16|0
```

No `provider.turn.interrupt.failed` activity was appended to the thread in that window.

### How I hit it, and why it may matter for #231

Full disclosure on the trigger: the specific mismatch above came from a `turn/steer` implementation in my own fork, **not** from upstream — upstream/main has no `turn/steer` call. So the *root cause* of my mismatch is mine to fix, and I am not asking you to fix that part.

The reason I am filing here anyway is that the swallow-the-failure path is entirely upstream code and will hide *any* provider interrupt rejection, whatever the cause.

Since #231 (`feat: add Steer and Queue follow-up modes`) is in progress, one concrete note that may save some time. In my fork the sequence was:

1. `turn/steer` is called on the running turn with `expectedTurnId: <current>`.
2. The response carries a **different** turn id than the turn the provider is actually still executing.
3. That returned id is written straight into the runtime's `activeTurnId`.
4. Every later interrupt reads that value and is rejected with `expected active turn id X but found Y`.

I confirmed step 2/3 independently: the id T3 sent decodes as a UUIDv7 minted at `11:11:31.027Z`, 72 ms after the steer message — while the provider's own event stream stayed on the original turn id for the entire turn (2078 occurrences of the original id, 0 of the new one).

If the upstream steer implementation also adopts the steer response's turn id as the active one, it can land in the same state. Validating that response — or at least making the interrupt failure visible — would cover it.

### Workaround

Send a plain text instruction ("stop working now, take no further actions"). That is cooperative, not a real interrupt, so an in-flight tool call can still finish first. In my incident the turn only ended that way.

m-de-graaff:
Fixes #4524

## Problem

When the provider rejects `turn/interrupt`, the failure goes nowhere the user can see. `processTurnInterruptRequested` issues the call bare:

```ts
yield* providerService.interruptTurn({ threadId: event.payload.threadId });
```

so the cause propagates up to `processDomainEventSafely`, which terminates it in a `logWarning`. The orchestration command is still receipted `accepted`, `thread.turn-interrupt-requested` is persisted normally, and no activity is appended. **Stop generation looks like it worked while the turn keeps running**, and the only trace is a line in `server-child.log`. The reporter clicked it 16 times in 17 seconds — 16 accepted receipts, 0 receipt errors, 0 activities.

## Change

The mechanism already exists and is already used one branch up in the same function: the "no active provider session" case appends a `provider.turn.interrupt.failed` activity. The sibling `processApprovalResponseRequested` likewise wraps its provider call in `Effect.catchCause` → `appendProviderFailureActivity`. The interrupt call was simply not wired to it.

This catches the cause and appends the same activity kind, so a rejected interrupt now shows up in the thread exactly like the missing-session case:

- `formatFailureDetail` is reused, so a `ProviderAdapterRequestError` surfaces its `detail` (`expected active turn id X but found Y`) instead of a pretty-printed cause. That is the same helper the turn-start failure path uses.
- Interrupt-only causes stay silent, matching `handleTurnStartFailure`.

No client change is needed: the thread already renders `tone: "error"` activities, which is the behavior the issue asks for.

## Tests

`ProviderCommandReactor.test.ts` — new case makes `interruptTurn` fail with a `ProviderAdapterRequestError` and asserts a `provider.turn.interrupt.failed` activity lands on the thread carrying the provider's own detail and the requested turn id.

I confirmed it is not vacuous: with the source change stashed the new test fails, and it passes with the change. `vp test run src/orchestration/Layers/ProviderCommandReactor.test.ts` — 35 passed.

The `interruptTurn` mock is now declared as `vi.fn<ProviderServiceShape["interruptTurn"]>` so a failing effect is assignable, matching how `respondToRequest` is already declared in that harness. The test is written with `effectIt.effect` rather than `Effect.runPromise`, so it adds no new entries against the `no-manual-effect-runtime-in-tests` baseline.

Typecheck, lint, and format clean for `apps/server`.

## Out of scope

The issue also notes that its own trigger came from a downstream `turn/steer` implementation adopting the steer response'"'"'s turn id as the active one, and flags that #231 could land in the same state. That is a separate concern; this PR only makes any such rejection visible.

---

> [!NOTE]
> **Low Risk**
> Localized orchestration error-handling around interrupt; no auth or data-model changes, and behavior aligns with existing approval-response failure surfacing.
> 
> **Overview**
> When **Stop generation** triggers `turn/interrupt`, a provider rejection used to disappear into server logs while the command still looked accepted. This change wires `processTurnInterruptRequested` so `interruptTurn` failures are caught and surfaced as a **`provider.turn.interrupt.failed`** thread activity (same path as the no-session case and approval responses), using **`formatFailureDetail`** so adapter errors show their `detail` text.
> 
> Interrupt-only causes are still swallowed, matching turn-start failure handling. A reactor test mocks a failing `interruptTurn` and asserts the error activity appears with the expected tone, turn id, and payload.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4be0c97eb539fc88ad5f9911df31472a820a9a57. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Surface provider turn interrupt failures as activities in the thread
> - In `ProviderCommandReactor.processTurnInterrupt`, wraps `providerService.interruptTurn` with `Effect.catchCause` to handle failures gracefully.
> - On failure, appends a `provider.turn.interrupt.failed` activity with formatted detail via `appendProviderFailureActivity`; interrupt-only causes are silently ignored.
> - Adds a test verifying that a `ProviderAdapterRequestError` from `interruptTurn` produces the expected failure activity with correct tone, turnId, and payload detail.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4be0c97.</sup>
> 
>
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
@@ -231,7 +231,7 @@ describe("ProviderCommandReactor", () => {
         turnId: asTurnId("turn-1"),
       }),
     );
-    const interruptTurn = vi.fn((_: unknown) => Effect.void);
+    const interruptTurn = vi.fn<ProviderServiceShape["interruptTurn"]>(() => Effect.void);
     const respondToRequest = vi.fn<ProviderServiceShape["respondToRequest"]>(() => Effect.void);
     const respondToUserInput = vi.fn<ProviderServiceShape["respondToUserInput"]>(() => Effect.void);
     const stopSession = vi.fn((input: unknown) =>
@@ -1757,6 +1757,74 @@ describe("ProviderCommandReactor", () => {
       threadId: "thread-1",
     });
   });
+
+  effectIt.effect(
+    "appends an interrupt failure activity when the provider rejects the interrupt",
+    () =>
+      Effect.gen(function* () {
+        const harness = yield* Effect.promise(() => createHarness());
+        const now = "2026-01-01T00:00:00.000Z";
+        harness.interruptTurn.mockImplementation(() =>
+          Effect.fail(
+            new ProviderAdapterRequestError({
+              provider: "codex",
+              method: "turn/interrupt",
+              detail: "expected active turn id turn-2 but found turn-1",
+            }),
+          ),
+        );
+
+        yield* harness.engine.dispatch({
+          type: "thread.session.set",
+          commandId: CommandId.make("cmd-session-set-interrupt-failure"),
+          threadId: ThreadId.make("thread-1"),
+          session: {
+            threadId: ThreadId.make("thread-1"),
+            status: "running",
+            providerName: "codex",
+            runtimeMode: "approval-required",
+            activeTurnId: asTurnId("turn-1"),
+            lastError: null,
+            updatedAt: now,
+          },
+          createdAt: now,
+        });
+
+        yield* harness.engine.dispatch({
+          type: "thread.turn.interrupt",
+          commandId: CommandId.make("cmd-turn-interrupt-failure"),
+          threadId: ThreadId.make("thread-1"),
+          turnId: asTurnId("turn-1"),
+          createdAt: now,
+        });
+
+        yield* Effect.promise(() =>
+          waitFor(async () => {
+            const readModel = await harness.readModel();
+            const thread = readModel.threads.find(
+              (entry) => entry.id === ThreadId.make("thread-1"),
+            );
+            return (
+              thread?.activities.some(
+                (activity) => activity.kind === "provider.turn.interrupt.failed",
+              ) ?? false
+            );
+          }),
+        );
+
+        const readModel = yield* Effect.promise(() => harness.readModel());
+        const thread = readModel.threads.find((entry) => entry.id === ThreadId.make("thread-1"));
+        expect(
+          thread?.activities.find((activity) => activity.kind === "provider.turn.interrupt.failed"),
+        ).toMatchObject({
+          tone: "error",
+          turnId: "turn-1",
+          payload: {
+            detail: "expected active turn id turn-2 but found turn-1",
+          },
+        });
+      }),
+  );
 
   it("starts a fresh session when only projected session state exists", async () => {
     const harness = await createHarness();
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts
@@ -904,7 +904,22 @@ const make = Effect.gen(function* () {
     }
 
     // Orchestration turn ids are not provider turn ids, so interrupt by session.
-    yield* providerService.interruptTurn({ threadId: event.payload.threadId });
+    yield* providerService.interruptTurn({ threadId: event.payload.threadId }).pipe(
+      Effect.catchCause((cause) =>
+        Cause.hasInterruptsOnly(cause)
+          ? Effect.void
+          : // Without this the rejection only reaches a log line, so Stop
+            // generation looks like it worked while the turn keeps running.
+            appendProviderFailureActivity({
+              threadId: event.payload.threadId,
+              kind: "provider.turn.interrupt.failed",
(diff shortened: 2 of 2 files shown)


juliusmarminge:

PR #4646, opened 2026-07-27
feat(mobile): Surface waiting background work in the thread feed and list

mwolson:
## Carried from another PR

- `feat(mobile): Surface waiting background work in the thread feed and list` — this PR.
- `feat(orchestrator): Surface waiting background work` — carried for #4378, which
  is not merged yet. Review it there, not here. It is the parent this branch is
  stacked on, and it supplies the `pendingBackgroundTasks` shell field and the
  shared `derivePendingBackgroundWork` helper that the mobile change consumes.

The Files tab therefore shows ~10.5k lines, of which this PR owns 495 across 8
files, all under `apps/mobile`. To review only this PR:
`git diff 7c6e48e43...HEAD`.

## Summary

Desktop has shown a post-settlement "Waiting on background task" indicator since
#4378; iOS showed nothing. This adds the mobile half on two surfaces: a row in
the thread feed, and a status pill on the classic thread list.

#4378 changes no `apps/mobile` file at all, so this gap was not going to close on
its own.

## Problem and Fix

| --- | --- |
| The conversation-level Waiting row existed only on web. `ChatView.tsx` derived `pendingBackgroundTasks` and the web timeline appended a `waiting-background` row; nothing under `apps/mobile` called `derivePendingBackgroundWork`, so the feed simply dropped the working row at settlement. | `deriveThreadPendingBackgroundWork` in `apps/mobile/src/lib/threadActivity.ts` derives the roster from the thread projection, and `deriveThreadFeedPresentation` appends a `waiting-background` entry when no active work remains. |
| The conversation view prefers `deriveThreadRuntime(projection)`, which has no roster parking, so `activeWorkStartedAt` went null at settlement and the row it gated vanished. | The row is driven by the roster rather than by run timing, using the same shared helper the server uses for the shell roster. |
| The mobile thread list showed no status either. `resolveThreadListV2Status` handles Waiting, but the v2 list sits behind the `threadListV2Enabled` preference and is off, so the list that actually renders is the classic one, whose `resolveThreadStatus` had no Waiting branch. Same shape as the classic web sidebar gap #4378 fixed. | `resolveThreadStatus` gained a static muted Waiting pill, ranked below approvals, input, active work and failures, and above Plan Ready. |

## Defensive Fixes

| --- | --- |
| Mobile restores projections from disk and `shouldPersistThread` persists only settled ones, so a cached projection is exactly the shape that reproduces a Waiting row for work that already finished. Web cannot hit this. | The roster is gated on a live thread detail. `synchronizing` is refused alongside `cached`, because a subscription passes through it before any fresh projection arrives, so its data can still be the cached one. The cost is a brief gap on reconnect, which is preferable to asserting work that is over. |
| The shared helper's `runs` argument is optional, and omitting it leaves the rolled-back-run exclusion set empty. The web call site passes it; an earlier revision of this change did not, so iOS would have shown Waiting for turn items owned by a `rolled_back` run where desktop shows nothing. | `runs: projection.runs` is passed, matching `ChatView.tsx` argument for argument. A focused test pins it: deleting that one argument fails exactly that test. |
| `formatPendingBackgroundWorkLabel` interpolates the raw description, which for a `subagent` item is the entire child prompt, wrapping into a wall of text on a phone. | The label is clamped to two lines. |

## Validation

- Focused: `apps/mobile/src/lib/threadActivity.test.ts` and
  `apps/mobile/src/features/threads/threadPresentation.test.ts`, 25 tests.
- Negative check, run independently: removing `runs: projection.runs` fails
  exactly the rolled-back test and nothing else.
- `vp check` 0 errors; `vp run typecheck` exit 0.
- The derivation lives in a pure helper rather than inline in the composer hook
  specifically so the policy is unit tested; `apps/mobile` has no
  hook-rendering test infrastructure.
- Verified on an iPhone 17 Pro Simulator against an isolated server: the row
  appears under a settled reply while a delegated subagent is still running,
  clears once the child finishes, and the list pill matches the row.

## Known limitations

- The list pill reads the server-derived shell roster while the feed derives
  from the live projection, so the two can disagree until the next shell update.
  The server SQL shell path omits `runs` by design, so the list can over-report
  after a rollback.
- The two mobile surfaces can disagree in one case: a failed run holding a
  nonempty roster reads Error on the list and Waiting in the feed, because the
  feed has no Error row. Mobile also ranks Error above Waiting while the classic
  web sidebar has no Error pill at all.
- Status pills in the classic mobile list are not part of the row's
  `accessibilityLabel`, so VoiceOver does not announce Waiting. Pre-existing for
  every pill on that surface.
- Mobile gates the working row on `activeWorkStartedAt` while web gates on a
  broader `isWorking`, so between a send tap and the new run reaching the
  projection, iOS can show Waiting where desktop shows Working. Pre-existing
  mobile signal gap, now more visible.
- The roster derivation is duplicated across web, client-runtime and now mobile.
  A shared `derivePendingBackgroundWorkForProjection(projection)` wrapper would
  collapse all three, but it touches web and client-runtime and is better done
  once #4378 lands.

> [!NOTE]
> ### Surface waiting background work in thread feeds and sidebars for web and mobile
> - Adds a `pendingBackgroundTasks` field to `OrchestrationV2ProviderThread` and `OrchestrationV2ThreadShell` contracts, with a new `OrchestrationV2PendingBackgroundTask` schema type; historical JSON without the field decodes to `[]`.
> - Adds `derivePendingBackgroundWork` and `formatPendingBackgroundWorkLabel` to the shared package to compute a normalized, deduplicated pending task list from settled runs, provider thread rosters, and non-terminal turn items, excluding rolled-back runs.
> - Web sidebar and message timeline render a non-pulsing 'Waiting' pill/row when background tasks are present but no active work is running; mobile thread feed appends an animated `WaitingBackgroundTimelineRow` with pulsing dots.
> - Claude adapter (`ClaudeAdapterV2`) tracks background task rosters per native thread, emits `provider_thread.updated` with `pendingBackgroundTasks`, and gates wake/continuation logic until native frames are present.
> - ACP adapter (`AcpAdapterV2`) extends subagent tracking with `pending` status, projects carryover terminals exactly once, and pins `hasPendingBackgroundWork` until unprojected terminals are cleared.
> - `RunExecutionService` seeds inherited background turn items from prior interrupted/failed/cancelled runs, routes their updates through `routeProviderEvent`, and uses a thread-scoped `hasPendingBackgroundWorkForThread` probe instead of session-wide to gate stream release.
> - `EventSink` gains `writeIfProviderThreadOwner`, an ownership-guarded atomic write that rejects commits when the provider thread has changed hands.
> - `ProviderRuntimeRecoveryService` now cancels stale background-capable turn items and clears `pendingBackgroundTasks` on all provider threads at recovery time.
> - Risk: the new per-thread ownership gate (`writeIfProviderThreadOwner`) silently drops post-terminal roster writes when ownership is lost; callers receive `committed=false` rather than an error.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7daf129.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/features/threads/ThreadDetailScreen.tsx
+++ b/apps/mobile/src/features/threads/ThreadDetailScreen.tsx
@@ -1,6 +1,7 @@
 import { type EnvironmentConnectionPhase } from "@t3tools/client-runtime/connection";
 import type { EnvironmentThreadShell } from "@t3tools/client-runtime/state/shell";
 import type { EnvironmentThreadStatus } from "@t3tools/client-runtime/state/threads";
+import type { PendingBackgroundWorkTask } from "@t3tools/shared/orchestrationV2PendingBackgroundWork";
 import { useKeyboardChatComposerInset, useKeyboardScrollToEnd } from "@legendapp/list/keyboard";
 import type { LegendListRef } from "@legendapp/list/react-native";
 import type {
@@ -52,6 +53,7 @@ export interface ThreadDetailScreenProps {
   readonly environmentLabel: string | null;
   readonly selectedThreadFeed: ReadonlyArray<ThreadFeedEntry>;
   readonly activeWorkStartedAt: string | null;
+  readonly pendingBackgroundTasks: ReadonlyArray<PendingBackgroundWorkTask>;
   readonly activePendingApproval: PendingApproval | null;
   readonly respondingApprovalId: RuntimeRequestId | null;
   readonly activePendingUserInput: PendingUserInput | null;
@@ -363,6 +365,7 @@ export const ThreadDetailScreen = memo(function ThreadDetailScreen(props: Thread
             threadTitle={props.selectedThread.title}
             latestRun={props.selectedThread.latestRun}
             activeWorkStartedAt={props.activeWorkStartedAt}
+            pendingBackgroundTasks={props.pendingBackgroundTasks}
             listRef={listRef}
             freeze={freeze}
             anchorMessageId={anchorMessageId}
--- a/apps/mobile/src/features/threads/ThreadFeed.tsx
+++ b/apps/mobile/src/features/threads/ThreadFeed.tsx
@@ -12,6 +12,7 @@ import {
 } from "@t3tools/contracts";
 import { CHAT_LIST_ANCHOR_OFFSET, resolveChatListAnchoredEndSpace } from "@t3tools/shared/chatList";
 import { formatElapsed } from "@t3tools/shared/orchestrationTiming";
+import type { PendingBackgroundWorkTask } from "@t3tools/shared/orchestrationV2PendingBackgroundWork";
 import { SymbolView } from "../../components/AppSymbol";
 import { HeaderHeightContext } from "@react-navigation/elements";
 import { useNavigation } from "@react-navigation/native";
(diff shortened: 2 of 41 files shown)


mwolson:
Folding this into #4378, which is where the Waiting feature lives on `t3code/codex-turn-mapping`. The mobile surfaces are part of the same body of work rather than a separate concern, and stacking them in their own PR only made the carried diff harder to read.

The commit is now the second commit on `feat/orchestrator-v2-background-waiting`, unchanged.

PR #4684, opened 2026-07-27
feat(terminal): close thread terminals when a thread settles

SunkenInTime:
Settling a thread means the work is done, but any dev server it started keeps holding its port and CPU. This closes a settled thread's terminals, behind a new server setting **`closeTerminalsOnThreadSettle`, defaulting to off**.

## Why two mechanisms

Only one of the two settle paths is an event, which shapes the whole design:

- **Explicit `✓ Settle`** dispatches `thread.settle`, so `ws.ts` closes terminals right after the command is accepted. A settle the decider rejects (active session, pending approval, queued turn start) fails before that point, so blocked work never loses its terminals.

- **Auto-settle** (inactivity window, merged/closed PR) emits *nothing*. `effectiveSettled` is a clock-derived predicate, and its inputs — `sidebarAutoSettleAfterDays` (a per-client setting) and change-request state (never persisted server-side; it lives only in `VcsStatusBroadcaster`'s per-cwd memory while a client holds a subscription) — are both invisible to the server. The client is therefore the only possible observer of that transition.

So the client reports it via a new `terminal.close-settled` RPC. **The report is advisory, never authoritative**: the server re-derives `canSettle` against the thread shell and ignores any report for a thread still holding live work. It also skips archived threads and threads explicitly pinned `active`. Idempotent, so several clients reporting the same thread is harmless.

## Shared predicate instead of a fourth copy

The settle rules were already duplicated across the server projector, the SQL projection pipeline, and the client reducer. Rather than hand-write a fourth, `canSettle` and `hasQueuedTurnStart` move to `@t3tools/shared` — the one package both sides already import. `client-runtime` re-exports them, so **no import site changes**. `effectiveSettled` itself stays client-side, since it genuinely depends on data the server lacks.

## Behavior notes

- Closes terminal *sessions*, matching what archive already does. No `deleteHistory`, so **scrollback survives** and the terminal can be reopened.
- **Unsettling never restarts anything.**
- **Snooze is deliberately excluded** — it means "not now", not "done", and a snoozed thread stays `active` in the data model.
- Version skew handled by a new `settledTerminalCleanup` capability flag, so new clients against old servers send nothing rather than erroring.

## Testing

- 3 new server tests: fires on a report, **ignores a report for a thread with a running session**, ignores when the setting is off. Plus 2 for the explicit path and 2 contract tests.
- Typecheck clean across all packages; `apps/web` 1595 passed; `@t3tools/contracts` 203 passed; lint/fmt clean.
- Pre-existing failures on this machine (systemd/POSIX-permission/fd tests on Windows) were confirmed unrelated by re-running them on a clean tree.

## Not done

- **Mobile does not report auto-settles.** Its `threadListV2` truncates the settled list to `settledLimit` and returns only `items` + `hiddenSettledCount`, so reporting off it would silently cover just the rendered rows. Correct wiring needs a return-shape change plus test updates. The server already accepts reports from any client.
- **Not yet exercised end-to-end** — the decision logic is unit-tested, but no real dev server has been killed by a real settle yet.

> [!NOTE]
> ### Close thread terminals when a thread settles
> - Adds a `closeTerminalsOnThreadSettle` server setting (default `false`) that triggers terminal closure when a thread settles, configurable via a new toggle in the General settings panel.
> - On explicit settle via `thread.settle` dispatch, the server closes all thread terminals if the setting is enabled.
> - Adds a `terminal.close-settled` RPC and a corresponding client-side effect in [SidebarV2.tsx](https://github.com/pingdotgg/t3code/pull/4684/files#diff-87e093a89032045fe7c876d3d324b60251f65a65d07e96e2a0c8eb92cb759d6c) that reports auto-settled threads to the server; the server validates settle eligibility using shared `canSettle` guards before closing.
> - Extracts `canSettle`, `hasQueuedTurnStart`, and `QUEUED_TURN_START_GRACE_MS` into a new shared module at [packages/shared/src/threadSettled.ts](https://github.com/pingdotgg/t3code/pull/4684/files#diff-3cc3433b47d82bcf9a99aa14f21f5960c11ec30e13854e3716884d768af01360) for consistent evaluation on both client and server.
> - Behavioral Change: terminals that were previously left open after thread settlement will now be closed automatically when the setting is enabled.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9795ea3.</sup>
> 
>

### DIFF
--- a/apps/server/src/environment/ServerEnvironment.ts
+++ b/apps/server/src/environment/ServerEnvironment.ts
@@ -142,6 +142,7 @@ export const make = Effect.gen(function* () {
       connectionProbe: true,
       threadSettlement: true,
       threadSnooze: true,
+      settledTerminalCleanup: true,
       ...(serverSelfUpdate === null ? {} : { serverSelfUpdate }),
     },
   };
--- a/apps/server/src/server.test.ts
+++ b/apps/server/src/server.test.ts
@@ -6442,6 +6442,212 @@ it.layer(NodeServices.layer)("server router seam", (it) => {
     }).pipe(Effect.provide(NodeHttpServer.layerTest)),
   );
 
+  it.effect("closes thread terminals after settle when the setting is enabled", () =>
+    Effect.gen(function* () {
+      const threadId = ThreadId.make("thread-settle-closes-terminals");
+      const effects: string[] = [];
+
+      yield* buildAppUnderTest({
+        layers: {
+          serverSettings: {
+            getSettings: Effect.succeed({
+              ...DEFAULT_SERVER_SETTINGS,
+              closeTerminalsOnThreadSettle: true,
+            }),
+          },
+          terminalManager: {
+            close: (input) =>
+              Effect.sync(() => {
+                effects.push(`terminal.close:${input.threadId}`);
+              }),
+          },
+          orchestrationEngine: {
+            dispatch: (command) =>
+              Effect.sync(() => {
+                effects.push(`dispatch:${command.type}`);
+                return { sequence: 1 };
+              }),
+          },
+        },
+      });
+
+      const wsUrl = yield* getWsServerUrl("/ws");
+      const dispatchResult = yield* Effect.scoped(
+        withWsRpcClient(wsUrl, (client) =>
+          client[ORCHESTRATION_WS_METHODS.dispatchCommand]({
+            type: "thread.settle",
+            commandId: CommandId.make("cmd-thread-settle"),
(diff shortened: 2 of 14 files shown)


t3dotgg:
> [!NOTE]

Settling a thread is reversible and does not mean its terminal process has no useful state. Automatically closing terminals can kill shells, servers, or unsaved interactive work, so the cleanup benefit does not justify the data-loss risk.

PR #4759, opened 2026-07-28
fix(orchestrator): Close OpenCode tool items when a turn ends under them

mwolson:
## Summary

Interrupting an OpenCode turn left whatever tool was mid-flight projected as
`running` permanently. In the UI that is a spinner that never stops, on a run
the user deliberately stopped.

Stop itself was never broken. The run projects as `interrupted` and sequential
recoveries map to their own completed runs. Only the tool item leaked.

## Problem and Fix

| Problem and Why it Happened                                                                                                                                                                                                                                                              | Fix                                                                                                                                                                           |
| ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `finalizeTurn` flushed only `text` and `reasoning` parts and never tool parts, so a tool still running when `session.abort` landed kept its last observed `running` state. OpenCode reports no final state for an aborted tool, so nothing else ever closed it and the row spun forever. | Sweep non-terminal tool parts in `finalizeTurn` and emit them with the turn's terminal status, via an optional `terminal` override threaded through `emitToolPart`.           |
| A subagent caught the same way would have collapsed to `failed`, since that was the only non-success status reachable without a mapping.                                                                                                                                                 | `emitSubagent` takes the same override. `OrchestrationV2Subagent` and `OrchestrationV2TurnItem` carry the same status literals, so an interrupted subagent stays interrupted. |

## Validation

Typecheck, lint, unit tests, and a live interrupt scenario against a real
`opencode serve`, all on this branch rather than on the branch where the fix was
originally written.

- `vpr typecheck` across all 15 packages: 0 errors
- `vp check`: 0 errors
- `vp test src/orchestration-v2/Adapters/OpenCodeAdapterV2.test.ts`: 10 passed
- Live, direct Stop plus two sequential recoveries on `opencode/glm-5.2`: pass

The live run captures the projection at several points. Counting
`command_execution` items by status:

- Mid-turn: 2 completed, 2 pending
- At the interrupt boundary: 2 completed, **2 running**
- After the interrupt: 2 completed, **2 interrupted**
- Final: 2 completed, 2 interrupted

The boundary capture shows the tools genuinely in flight; the next one shows
them closed. Before this change they stayed `running` in every subsequent
capture, which is the reported symptom.

Manual re-test scenarios, in the published guide at
<https://nam7nt0rbtm6.postplan.dev>:

- [OpenCode · interrupt, scenario 1 (Stop while a command is running)](https://nam7nt0rbtm6.postplan.dev#opencode-interrupt-1)
- [OpenCode · interrupt, scenario 2 (send after Stop lands on a new run)](https://nam7nt0rbtm6.postplan.dev#opencode-interrupt-2)
- [OpenCode · interrupt, scenario 3 (second send lands on its own run)](https://nam7nt0rbtm6.postplan.dev#opencode-interrupt-3)

Scenario 1 is the one this fix changes. Scenarios 2 and 3 passed before and
after, and are included because they are what would catch an over-broad fix that
closed the interrupted run's items by tearing down the provider thread.

## Replay coverage

`turn_interrupt_mid_tool` gains an OpenCode provider, so the fix is driven
through the real adapter rather than asserted on a pure helper. The transcript
reproduces the ordering from a live run rather than an idealised one: the tool
goes running, the abort goes out, `session.error` and idle arrive, and only
afterwards does OpenCode's cleanup emit its own tool part update. That late
update lands on an already-finalized turn and is ignored, so the item's
`interrupted` status has to come from the sweep.

Confirmed the fixture catches the regression: with the sweep commented out it
fails with `expected 'running' to equal 'interrupted'`, and nothing else breaks.

## Notes for review

The fix was found during an audit of OpenCode background-work and interrupt
coverage that also produced an OpenCode 2.0 driver. That driver is a separate,
much larger change and is deliberately not in this PR; this one is independently
shippable and fixes a bug users can hit today.

OpenCode 1.x has no background work that can outlive the root turn, so
`hasPendingBackgroundWork` and `hasPendingBackgroundWorkForThread` remaining
unimplemented on this adapter is correct rather than an oversight.

## Current CI Note

The branch-local validation above passes. GitHub Check, Release Smoke, Mobile
Native Static Analysis, Cursor Bugbot, CodeRabbit, and both Macroscope code
checks pass. Test currently fails in five unchanged queued-turn replay
assertions from the newer `t3code/codex-turn-mapping` merge result; this PR does
not touch queued-turn behavior.

<!-- Macroscope's pull request summary starts here -->
<!-- Macroscope will only edit the content between these invisible markers, and the markers themselves will not be visible in the GitHub rendered markdown. -->
<!-- If you delete either of the start / end markers from your PR's description, Macroscope will append its summary at the bottom of the description. -->
> [!NOTE]
> ### Fix OpenCode tool items left in running state when a turn ends
> - Adds `terminalToolStatus` in [OpenCodeAdapterV2.ts](https://github.com/pingdotgg/t3code/pull/4759/files#diff-047e153882de41edc9dabe56bdf9951ed145e3197c6ca27e83f08b9b3eb13250) to map a terminal turn status (`completed`, `interrupted`, `cancelled`, `failed`) to matching node/item statuses for tools that never reported their own terminal state.
> - `finalizeTurn` now sweeps any tool parts still in `pending` or `running` state, calling `emitToolPart` with the turn's terminal status so they close with a matching status and a non-null `completedAt`.
> - `emitSubagent` and `emitToolPart` accept an optional terminal status parameter; when present, they use `terminalToolStatus` output instead of the observed `toolStatus`/`toolCompletedAt`.
> - Adds a `turn_interrupt_mid_tool` fixture with an OpenCode transcript and assertion function to cover the interrupted-mid-tool scenario end-to-end.
> - Behavioral Change: subagents and tool items swept at turn finalization now carry the precise terminal status of the turn instead of remaining `running` indefinitely.
>
> <!-- Macroscope's review summary starts here -->
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a33de3b.</sup>
> <!-- Macroscope's review summary ends here -->
>
> <!-- macroscope-ui-refresh -->
<!-- Macroscope's pull request summary ends here -->

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Changes turn-finalization and tool/subagent projection in the OpenCode adapter—a user-visible orchestration path—but scope is narrow and covered by new unit and replay tests.
> 
> **Overview**
> Fixes **OpenCode** runs where **Stop** left mid-flight tools projected as **`running`** forever (stuck UI spinners) even though the run correctly showed **`interrupted`**.
> 
> **`finalizeTurn`** now sweeps tool parts still **`pending`** or **`running`** when the turn ends (e.g. after **`session.abort`**) and re-emits them via **`emitToolPart`** / **`emitSubagent`** with an optional terminal override. New **`terminalToolStatus`** maps the turn’s terminal outcome (`interrupted`, `cancelled`, `failed`, `completed`) onto matching item/node statuses; subagents use the same literals instead of collapsing non-success to **`failed`**.
> 
> Adds unit tests for **`terminalToolStatus`** and an OpenCode **`turn_interrupt_mid_tool`** replay fixture (including late provider cleanup events on an already-finalized turn).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a33de3b47cb5a78b05535d65b2ff176222066d5f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->


mwolson:
Absorbed into #4786 cleanup-drain. That PR now carries the OpenCode 1.x interrupt-tool commits plus the cleanup drain on current t3code/codex-turn-mapping. Review the OpenCode 1.x work there.

PR #4781, opened 2026-07-28
fix(web): keep send reachable while a turn is running on mobile

### LINKED ISSUE #4775: [Bug]: No way to send a follow-up mid-turn on mobile viewports — send button is replaced by stop and Enter-to-send is disabled below `sm`
### Before submitting

### Area

apps/web

### Steps to reproduce

No phone required — the trigger is viewport width, not touch.

1. Open a thread in the web UI and narrow the window below the `sm` breakpoint (< 640px), or open it on a phone.
2. Send a prompt so the turn starts running.
3. Type a follow-up in the composer while the turn is still running.
4. Try to send it.

### Expected behavior

The follow-up can be sent, the same as on a desktop viewport where pressing <kbd>Enter</kbd> mid-run submits it.

### Actual behavior

There is no way to submit. The composer's primary action is replaced by the stop button while the turn runs, and Enter-to-send is disabled on mobile viewports, so the message can only be sent after the turn ends.

### Root cause

The two guards overlap, leaving no send path below `sm`:

- `apps/web/src/components/chat/ComposerPrimaryActions.tsx` — `if (isRunning)` returns *only* the stop button, replacing the submit button entirely.
- `apps/web/src/composer-logic.ts` — `shouldSubmitComposerOnEnter` returns `!isMobileViewport && !shiftKey`, so Enter never submits below `sm`.

Sending mid-run is otherwise fully supported: `onSend` in `ChatView.tsx` has no `isRunning` guard, and `isSendBusy` is `activeLocalDispatch !== null` rather than the turn state. Desktop reaches that path via Enter; mobile has no path to it at all. So this is a missing affordance on one viewport, not a missing capability.

### Notes

- This is narrower than #231 (Steer/Queue follow-up modes) — it asks for no new follow-up semantics, only for the existing desktop behavior to be reachable by touch. If #231 already covers the mobile affordance, feel free to close this.
- #1462 mentioned the same send/stop swap in passing while asking for queue-vs-steer configuration.
- Unrelated to #4713 / #4524, which are about interrupt itself being a no-op server-side.

I have a small patch for this (one prop, rendering send next to stop while running when the composer has content; desktop unchanged) and am happy to open a PR if you want it.

AMohamedAakhil:
Fixes #4775

## What Changed

On viewports below `sm`, there is no way to send a follow-up while a turn is running.

Two guards overlap and between them leave no send path:

- `ComposerPrimaryActions.tsx` returns *only* the stop button whenever `isRunning` is set, replacing the submit button.
- `shouldSubmitComposerOnEnter` in `composer-logic.ts` returns `!isMobileViewport && !shiftKey`, so Enter never submits below `sm`.

This adds a `showSendWhileRunning` prop, wired to `isMobileViewport`, that renders the send button next to stop while a turn is running and the composer has sendable content. Desktop rendering is unchanged.

To render both buttons without duplicating markup, the running branch moves below the send button, which becomes a `sendButton` const. That reorders it past the plan follow-up branch, which is inert: `showPlanFollowUpPrompt` requires `latestTurnSettled`, so it cannot be set while a turn is running.

Non-test diff is +29 / -17 across two files.

## Why

Sending mid-run is already supported. `onSend` in `ChatView.tsx` has no `isRunning` guard, and `isSendBusy` is `activeLocalDispatch !== null` rather than the turn state. Desktop users reach that path by pressing Enter. Mobile has no path to it at all.

So this is not a new follow-up mode, and it takes no position on queue vs steer semantics (#231). It only makes an action the app already supports reachable by touch.

The collapsed mobile composer hides how sharp this is. It passes `isRunning={false}` (ChatComposer.tsx:2748, :3032), so it shows a send arrow while a turn runs. But tapping it to type expands the composer, and the expanded footer is where send is swapped for stop. The affordance is only visible while you cannot type into it.

## UI Changes

Captured in a real browser at 390x844 against a dev server, with a turn running and text typed in the composer.

Before, stop is the only action:

before

After, send sits next to stop:

after

Enumerating the composer's visible buttons in the same states gives `["Stop generation"]` before and `["Stop generation", "Send message"]` after.

No visual change on desktop: `showSendWhileRunning` defaults to `false`, and the only caller passes `isMobileViewport`.

## Tests

`ComposerPrimaryActions.test.ts` became `.test.tsx` so the component can be rendered with `renderToStaticMarkup`, matching `ComposerBannerStack.test.tsx`. Three cases added:

- running with content, Enter-to-send available: stop only
- running with content, Enter-to-send unavailable: stop and send
- running with an empty composer: stop only

`vp test run apps/web/src/components/chat/ComposerPrimaryActions.test.tsx` passes 11/11. `vp lint` and `vp fmt --check` are clean, and `vp run --filter @t3tools/web typecheck` exits 0.

## Checklist

- [ ] I included a video for animation/interaction changes (no motion or interaction changes)

---

> [!NOTE]
> **Low Risk**
> Focused composer UI wiring with no backend or auth changes; desktop defaults preserve prior behavior.
> 
> **Overview**
> On mobile, **Send** can appear next to **Stop** while a turn is running when the composer has sendable content, so follow-ups are reachable without Enter-to-send.
> 
> `ComposerPrimaryActions` gains **`showSendWhileRunning`** (default off). When a turn is running it no longer returns only the stop control; it can render stop plus the existing submit button when that flag is set and there is content. Stop sizing shifts slightly when both buttons show. **`ChatComposer`** passes **`showSendWhileRunning={isMobileViewport}`** on the expanded mobile footer; desktop behavior is unchanged.
> 
> Tests cover stop-only on desktop-style running state, stop+send when Enter-to-send is unavailable, and stop-only with an empty composer.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da4fad351fe35fd47d87860c6adb6a1f931d3f64. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Show Send button alongside Stop button during a running turn on mobile viewports
> - Adds a `showSendWhileRunning` prop to `ComposerPrimaryActions` that, when true, renders the Send button next to the Stop button while a turn is running and there is sendable content.
> - `ChatComposer` passes `showSendWhileRunning={isMobileViewport}` so the behavior is only active on mobile, where Enter-to-send is unavailable.
> - The Stop button size increases slightly when Send is also shown to maintain visual balance.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized da4fad3.</sup>
> 
>


AMohamedAakhil:
@juliusmarminge can you please take a look at this 👀

AMohamedAakhil:
Some extra context that may help triage this.

**This is a recent regression, not long-standing behavior.** `shouldSubmitComposerOnEnter` gained its `isMobileViewport` guard in 6a1762967 (#3930, merged 18 Jul). That change is right on its own, since Enter should insert a newline on a software keyboard. But `ComposerPrimaryActions` already replaced send with stop whenever `isRunning`, so removing the Enter path removed the last remaining way to send mid-turn below `sm`. Before #3930, a mobile user could still interject by pressing Enter.

**It lands hardest on T3 Connect.** #4775 has a report reproducing it over a managed relay tunnel from an iPhone on 5G. With Connect now GA you can reach an agent from your phone on any network, but you cannot interject once a turn starts, which removes most of the value of remote control for anything longer than a quick question. Their workaround is a useful confirmation of the cause: rotating to landscape crosses the `sm` breakpoint, 430px to 932px on a 15 Pro Max, and Enter-to-send comes back. So the trigger is the breakpoint, not anything touch specific.

**Status:** checks are green, non-test diff is +29/-17 across two files, and desktop rendering is unchanged since `showSendWhileRunning` defaults to false and the only caller passes `isMobileViewport`.

No rush if this collides with the Steer and Queue work in #231. Happy to close it if that already covers the mobile affordance.

PR #4821, opened 2026-07-29
fix(server): stop replayed history from re-arming mouse reporting

### LINKED ISSUE #4776: [Bug]: Terminal occasionally displays raw ANSI escape sequences when opened
### Before submitting

### Area

apps/desktop

### Steps to reproduce

1. Open the project.
2. Open the integrated terminal.
3. Repeat closing and reopening the terminal several times.
4. Occasionally, the terminal starts with garbled output containing visible ANSI escape sequences.

### Expected behavior

The terminal should interpret ANSI escape sequences correctly and display the prompt and startup output with normal formatting.

### Actual behavior

The terminal sometimes prints ANSI escape/control sequences as plain text, for example sequences beginning with characters such as:

^[4;...
^[38;...
^[15;...

This causes the startup output and prompt to appear corrupted. Exiting the active shell via ctrl+x or ctrl+c doenst work. After terminating and reopening the terminal, it usually works normally again.  This happened after running bun run dev on a nextjs project.

https://github.com/user-attachments/assets/738131bf-90be-4e19-8e08-ec13e8a7a585

### Impact

### Version or commit

0.0.30-nightly.20260728.931

### Environment

macOS 26.3

### Logs or stack traces

```shell

```

_No response_

### Workaround

terminating the shell completly

Sy-D:
## What Changed

`snapshot()` now clears the mouse tracking modes at the end of a replayed terminal history — but only when the session has no running child process.

## Why

Reported in #4776: reopening the terminal occasionally fills it with sequences like `^[[<35;65;13M`.

Those are SGR mouse reports (mode 1006) — something a terminal sends *to* a program, never the reverse. Seeing them as text means reporting is armed while nothing is left to consume it, so the tty echoes each pointer move into the shell. That also explains why Ctrl-C does not help and why only killing the shell does: it is terminal state, not a stuck process.

Mouse reporting is state, not text, but the history replays it as text. A TUI killed before it could restore the mode leaves the enabling sequence in the retained history, and every reopen — including switching threads and coming back — replays it into a fresh xterm.

The gate matters: a live TUI (vim, htop, lazygit) legitimately depends on the replay restoring its mouse support after a thread switch. Clearing unconditionally would have broken that silently. With a child process running the history is passed through untouched.

## Notes For Review

`sanitizeTerminalHistoryChunk` already defends against this class of replay hazard — it strips DSR, CPR, DA and OSC colour *queries* so a replay cannot make the terminal answer into the shell. Mode setting is the same hazard in a different shape, but it cannot be stripped on the way in, because at record time the mode may still be wanted. The decision belongs at replay time, which is why this sits in `snapshot()`.

Only the retained history is touched; live output still reaches the client verbatim.

## Validation

The trigger is intermittent and I could not make it fire on demand, so here is what is measured rather than assumed:

1. **The mechanism reproduces exactly.** Driving a real `xterm.js` instance through the same steps `writeTerminalBuffer` performs — write `ESC c`, then a history containing `ESC[?1002h ESC[?1006h` — a fresh terminal emits nothing on pointer events before the replay and `ESC[<0;14;3M`, `ESC[<32;25;4M` after it. Same shape as the reported screenshot. With the reset appended, it emits nothing again.
2. **Private modes really do survive into the persisted history.** A live session log contains `ESC[?2004h` (bracketed paste), confirming DECSET sequences reach the replayed buffer.
3. **The reporter's dev server is not the source.** `next dev` (Next.js 16.2.9, Turbopack) captured under a real PTY emits no private mode sequences at all, so the stale enable comes from something else in the long-lived session — consistent with a hard-killed TUI.

Also run:

- `vp test run apps/server/src/terminal/Manager.test.ts` — 53 passed, including three new cases (clears when idle, passes through with a running child, untouched when no mouse mode was ever set)
- `vp run --filter t3 test` — 1673 passed, 7 skipped
- `vp run --filter t3 typecheck` — no errors

Closes #4776.

## Checklist

- [x] No UI change, so no screenshots (terminal byte stream only)

---

> [!NOTE]
> **Medium Risk**
> Changes terminal snapshot replay semantics for all reconnects; incorrect subprocess detection could strip mouse modes from an active TUI, though attach now refreshes that flag first.
> 
> **Overview**
> Fixes **#4776**, where reconnecting to a terminal could flood the shell with literal mouse report sequences (`^[[<…M`) after a TUI exited without restoring terminal modes.
> 
> **Replay-time sanitization:** `snapshot()` now runs history through `replaySafeTerminalHistory` before clients receive it. If the retained transcript contains a DECSET that enabled mouse **tracking** (modes 9, 1000–1003), the server appends matching `DECRST` sequences for tracking and related encoding modes. Encoding-only enables (e.g. `1006` alone) are left unchanged. When `hasRunningSubprocess` is true, history is passed through so live TUIs still get mouse support after a thread switch.
> 
> **Attach freshness:** `attachStream` calls `pollSubprocessActivity()` immediately before the initial snapshot so the idle-vs-child gate does not rely on a stale poller interval.
> 
> Unit tests cover idle reset, active-child passthrough, combined/8-bit CSI, X10 mode, and no-op histories.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit add5fa2be2571cb7dd1f304223bd445f0b4cb92d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Stop replayed terminal history from re-arming mouse reporting on reconnect
> - Adds `replaySafeTerminalHistory` in [Manager.ts](https://github.com/pingdotgg/t3code/pull/4821/files#diff-9b98fe5056129db24d311ed1bd8f66a72127cf90fa7529da47e0725f2813532e) that appends DECSET reset sequences for mouse tracking modes (9, 1000–1003) and encoding modes (1006, 1015, 1016) when replaying history that enabled mouse reporting.
> - The reset is only appended when no subprocess is running; if a subprocess is active the history is left unchanged to avoid disrupting live sessions.
> - Refreshes the subprocess activity state via `pollSubprocessActivity` immediately before emitting the initial snapshot so the replay decision reflects current state.
> - Behavioral Change: clients reconnecting to an idle terminal that had mouse reporting enabled will now receive reset sequences, clearing mouse mode from their terminal.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized add5fa2.</sup>
> 
>
>

### DIFF
--- a/apps/server/src/terminal/Manager.test.ts
+++ b/apps/server/src/terminal/Manager.test.ts
@@ -1702,3 +1702,27 @@ it.layer(
     }).pipe(Effect.provide(TestClock.layer())),
   );
 });
+
+const MOUSE_ENABLED_HISTORY = "$ lazygit\r\n\u001b[?1002h\u001b[?1006h";
+
+it("clears mouse reporting from a replayed history once the shell is idle", () => {
+  const replayed = TerminalManager.replaySafeTerminalHistory(MOUSE_ENABLED_HISTORY, false);
+
+  expect(replayed.startsWith(MOUSE_ENABLED_HISTORY)).toBe(true);
+  expect(replayed.slice(MOUSE_ENABLED_HISTORY.length)).toBe(
+    "\u001b[?1000l\u001b[?1002l\u001b[?1003l\u001b[?1006l\u001b[?1015l",
+  );
+});
+
+it("leaves mouse reporting alone while a child process can still consume it", () => {
+  assert.equal(
+    TerminalManager.replaySafeTerminalHistory(MOUSE_ENABLED_HISTORY, true),
+    MOUSE_ENABLED_HISTORY,
+  );
+});
+
+it("does not touch a history that never enabled mouse reporting", () => {
+  const history = "$ echo hello\r\nhello\r\n\u001b[?2004h";
+
+  assert.equal(TerminalManager.replaySafeTerminalHistory(history, false), history);
+});
--- a/apps/server/src/terminal/Manager.test.ts
+++ b/apps/server/src/terminal/Manager.test.ts
@@ -1710,7 +1710,7 @@ it("clears mouse reporting from a replayed history once the shell is idle", () =
 
   expect(replayed.startsWith(MOUSE_ENABLED_HISTORY)).toBe(true);
   expect(replayed.slice(MOUSE_ENABLED_HISTORY.length)).toBe(
-    "\u001b[?1000l\u001b[?1002l\u001b[?1003l\u001b[?1006l\u001b[?1015l",
+    "\u001b[?1000l\u001b[?1001l\u001b[?1002l\u001b[?1003l\u001b[?1006l\u001b[?1015l\u001b[?1016l",
   );
 });
 
@@ -1719,6 +1719,34 @@ it("leaves mouse reporting alone while a child process can still consume it", ()
     TerminalManager.replaySafeTerminalHistory(MOUSE_ENABLED_HISTORY, true),
     MOUSE_ENABLED_HISTORY,
   );
+});
+
+it("detects mouse reporting enabled through a combined DECSET", () => {
+  const history = "$ lazygit\r\n\u001b[?1002;1006h";
+
+  expect(TerminalManager.replaySafeTerminalHistory(history, false)).not.toBe(history);
+});
+
+it("detects mouse reporting enabled through an 8-bit CSI introducer", () => {
+  const history = "$ lazygit\r\n\u009b?1003h";
+
+  expect(TerminalManager.replaySafeTerminalHistory(history, false)).not.toBe(history);
+});
+
+it("ignores an encoding mode that cannot make the terminal report on its own", () => {
+  const history = "$ echo hi\r\n\u001b[?1006h";
+
+  expect(TerminalManager.replaySafeTerminalHistory(history, false)).toBe(history);
+});
+
+it("resets every mouse mode it recognises as an enable", () => {
+  const reset = TerminalManager.replaySafeTerminalHistory("\u001b[?1002h", false).slice(
+    "\u001b[?1002h".length,
+  );
+
+  for (const mode of [1000, 1001, 1002, 1003, 1006, 1015, 1016]) {
+    expect(reset).toContain(`\u001b[?${mode}l`);
+  }
 });
 
 it("does not touch a history that never enabled mouse reporting", () => {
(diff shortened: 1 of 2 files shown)


Sy-D:
All three findings were valid and are fixed in b4ce3f5b7.

**Combined and 8-bit DECSET** — correct, the `includes` check was too naive. `replaySafeTerminalHistory` now parses DECSET properly, handling semicolon-separated parameters and both the 7-bit `ESC [` and 8-bit `\u009b` introducer.

**Missing reset codes** — correct, and chasing it surfaced a second problem: detection and reset were sharing one list, which conflated two different things. They are now separate, because only a *tracking* mode makes the terminal emit reports while an encoding merely shapes them:

- tracking (detected and reset): 1000, 1001, 1002, 1003
- encoding (reset only): 1006, 1015, 1016

So a history that enables just `?1006h` is now correctly left alone — it cannot make the terminal report on its own — while the reset still clears everything, 1001 and 1016 included.

**Stale `hasRunningSubprocess`** — the strongest of the three, since it undercut the very gate the fix depends on. The flag is now refreshed at attach time by running the existing subprocess poll before the snapshot is built, so the decision no longer uses state up to a poll interval old. The remaining window is the microseconds between that refresh and the snapshot rather than up to a second.

Four new test cases cover the combined DECSET, the 8-bit introducer, the encoding-only history, and the completeness of the reset list.

`vp run --filter t3 test` — 1677 passed, 7 skipped. Typecheck and lint clean.

Sy-D:
Correct, and fixed in 7d116374a — verified rather than taken on trust.

Driving a real `xterm.js` instance and dispatching pointer events after replaying each history:

| replayed history | terminal emits |
| --- | --- |
| `ESC[?9h` | nothing |
| **`ESC[?9h ESC[?1006h`** | **`ESC[<0;14;3M`** |
| `ESC[?1000h` | nothing |
| `ESC[?1000h ESC[?1006h` | `ESC[<0;14;3M` |

So mode 9 does arm reporting, exactly as you said. A tracking mode on its own emitted nothing in this harness, but that is not the realistic case — a TUI sets a tracking mode and an encoding together, and then X10 behaves like the rest.

`MOUSE_TRACKING_MODES` now starts at `9`. Since `MOUSE_REPORTING_RESET` is derived from that list, `ESC[?9l` is emitted automatically, and the test asserting the reset covers every recognised mode picks it up without being restated.

`vp run --filter t3 test` — 1678 passed, 7 skipped. Typecheck and lint clean.

Sy-D:
Closed as fixed in https://github.com/pingdotgg/t3code/pull/5319

PR #4826, opened 2026-07-29
fix(web): stop the status label stacking on the row's settle actions

### LINKED ISSUE #4817: [Bug]: Settled label and schedule icon appear on a working inbox item
### Before submitting

### Area

apps/desktop

### Steps to reproduce

The issue is intermittent and I have not found a fully deterministic reproduction yet.

### Expected behavior

A working inbox item should continue to show only its Working state and switch on hover to its settle option.

### Actual behavior

When hovering over a working inbox item, the Settled label and schedule icon are also displayed.

Depending on the item, both states appear at the same time or the Working state is incorrectly replaced by the Settled state on hover. 

Image

### Impact

### Version or commit

0.0.30-nightly.20260728.931

### Environment

macOS 26.3

### Logs or stack traces

```shell

```

_No response_

### Workaround

_No response_

Sy-D:
## What Changed

Two things in the v2 sidebar row's status slot:

1. The status label no longer stays visible when focus promotes the settle/snooze actions into the same slot.
2. The label aligns to the row's baseline, with its leading icon optically centred on the text.

### The overlap (#4817)

The label and the hover actions share one slot and are meant to be mutually exclusive. The hover branch was symmetric — the label gets `opacity-0`, the actions get `opacity-100`. The focus branch was not: `group-focus-within` pulled the label out of flow with `absolute right-0` but never faded it, so the actions came up to full opacity underneath a label that was still fully opaque and pinned right.

That explains why the report reads as intermittent and hover-driven when it is neither. Focus is sticky: clicking Settle or Snooze leaves focus on the button, and dismissing the snooze popover returns focus to its trigger, at which point `snoozeMenuOpen` stops forcing the label transparent while focus keeps the actions promoted. The overlap then outlives the pointer. Keyboard tabbing reaches it too. In the reporter's screenshot the affected row is the *active* row, not a hovered one — which is what pointed at focus rather than hover.

Because the correctness property is a *pairing* — every variant that promotes the actions must fade the label — both class strings now live in one helper, and the tests assert that pairing generically instead of matching literal strings. Removing the new `opacity-0` fails them.

### The alignment

An `inline-flex` derives its baseline from its first flex item, and an SVG has none, so the leading 16px icon was setting the label's baseline and lifting "Working" above the project title. The text run now aligns on the baseline and the icon opts out with `self-center`, so it rides the centre of the text's line box. No pixel nudges.

## Validation

Measured in a running client on a seeded working row, toggling the classes live to get a real before/after:

**Alignment** — baselines in page coordinates:

| | project label | "Working" | duration | icon centre | label box centre |
| --- | --- | --- | --- | --- | --- |
| before | 243 | 241 | — | 237 | 237 |
| after | 243 | 243 | 243 | 239 | 239 |

Off by 2px before, exactly aligned after; the icon stays centred on the text in both.

**The overlap** — computed styles with focus inside the slot:

| | label opacity | actions opacity | both visible |
| --- | --- | --- | --- |
| without the fix | 1 (absolute, right-0) | 1 | **yes** |
| with the fix | 0 | 1 | no |

Idle is unchanged: label 1, actions 0.

- `vp run --filter @t3tools/web test` — 1644 passed

## Notes For Review

The actions' `focus-within:` is restated as `group-focus-within/v2-status-slot:` so both halves key off the identical condition. Behaviourally equivalent here — the action buttons are the only focusable elements in the slot — but it makes the pairing checkable.

The slim settled/snoozed row has a similar `inline-flex items-center` badge. It never renders the working state, so it is untouched.

Closes #4817.

## Checklist

- [x] Before/after screenshots — the issue carries the "before"; an "after" still needs attaching, so this stays draft until then

Before
image

After with alignment fix

https://github.com/user-attachments/assets/e777d10a-c475-4961-b5b0-6fd985171339

> [!NOTE]
> > > [!NOTE]
> ### Fix status label overlapping settle actions in sidebar rows
> - Introduces `resolveSidebarV2StatusSlotClassNames` in [Sidebar.logic.ts](https://github.com/pingdotgg/t3code/pull/4826/files#diff-529ae8997a33774d7ec3514d7abdb655942eaa993f71e6ec6728c892285d251a) to compute paired Tailwind class names for the status label and action buttons in a shared slot.
> - The status label fades out and moves out of flow on group hover/focus or when the snooze popover is open; actions become visible and in-flow under the same conditions.
> - Updates [SidebarV2.tsx](https://github.com/pingdotgg/t3code/pull/4826/files#diff-87e093a89032045fe7c876d3d324b60251f65a65d07e96e2a0c8eb92cb759d6c) to use the new utility, replacing the previous inline class logic that caused the label and actions to stack.
> - Also fixes a 2px vertical drift by aligning the status container to `items-baseline` with `self-center` on leading icons.
> >
> 
> >
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d9c2058.</sup>
> 
>

---

> [!NOTE]
> **Low Risk**
> Scoped UI/CSS in sidebar v2 rows with regression tests on the class pairing; no auth, data, or API changes.
> 
> **Overview**
> Fixes **v2 sidebar card rows** where the status text (e.g. "Working") could sit on top of **Settle/Snooze** after focus stayed on those controls—the focus path promoted actions but did not fade the label, unlike hover.
> 
> **`resolveSidebarV2StatusSlotClassNames`** in `Sidebar.logic.ts` now returns paired Tailwind classes for the label and action halves of the shared status slot (hover, `group-focus-within/v2-status-slot`, and open snooze popover). `SidebarV2.tsx` uses that helper instead of inline `cn(...)` on each span. Action visibility keys off the same `group-focus-within/v2-status-slot` condition as the label hide.
> 
> Also adjusts status **vertical alignment**: `items-baseline` on the label row and `self-center` on leading icons so the working icon no longer pulls the text ~2px above the project title.
> 
> Unit tests assert the pairing invariant (every action-reveal variant must apply matching `opacity-0` on the label).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9c205898ef6533b629e985d127c0f2f708b1a93. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/web/src/components/Sidebar.logic.test.ts
+++ b/apps/web/src/components/Sidebar.logic.test.ts
@@ -17,6 +17,7 @@ import {
   resolveSidebarStageBadgeLabel,
   resolveThreadRowClassName,
   resolveSidebarV2Status,
+  resolveSidebarV2StatusSlotClassNames,
   resolveThreadStatusPill,
   resolveWorkingStartedAt,
   formatWorkingDurationLabel,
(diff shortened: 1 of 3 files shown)


Sy-D:
fixed by https://github.com/pingdotgg/t3code/pull/4574, merged in https://github.com/pingdotgg/t3code/commit/78eb3ecae981a6fce108cd3ecc44a6832b4cd3a0. dismissed settle/snooze popovers no longer leave mouse-focused controls overlapping the status label. closing as completed.

Issue #4818, opened 2026-07-29
[Bug]: SQLError Failed to execute OrchestrationCommandReceiptRepository.upsert.query


Sy-D:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

Couldnt replicate for now.

### Expected behavior

No error.

### Actual behavior

<img width="816" height="81" alt="Image" src="https://github.com/user-attachments/assets/7cab8d74-23eb-481a-85b6-6dcebf07fc7b" />

### Impact

Minor bug or occasional failure

### Version or commit

0.0.30-nightly.20260728.931

### Environment

macOS 26.3

### Logs or stack traces

```shell

```

Sy-D:
I dug into this against `1b4830ff0`. Short version: the message cannot show its own cause, so the screenshot alone can never identify the root failure.

### Why the message is empty of information

`apps/server/src/persistence/Errors.ts:41`

```ts
override get message(): string {
  return this.detail === undefined
    ? `SQL error in ${this.operation}`
    : `SQL error in ${this.operation}: ${this.detail}`;
}
```

`apps/server/src/persistence/Errors.ts:76`

```ts
export function toPersistenceSqlError(operation: string) {
  return (cause: unknown): PersistenceSqlError =>
    new PersistenceSqlError({
      operation,
      detail: `Failed to execute ${operation}`,
      cause,
    });
}
```

`detail` restates `operation`, so the rendered message is the operation name twice. The real failure *is* captured in `cause`, it just never reaches the message.

Worth noting: `Errors.test.ts` already contains a test named **"keeps SQL operation context without a tautological detail"**, which constructs the error without a detail and asserts the message is `SQL error in AuthSessionRepository.list:query`. The intent is documented and tested; `toPersistenceSqlError` is the path that does not honour it, and it is the path this repository uses.

### Which call site produced it

Two callers of `OrchestrationCommandReceiptRepository.upsert`:

- `OrchestrationEngine.ts:281` — the rejected-command path, which ends in `.pipe(Effect.catch(() => Effect.void))`. It cannot surface anything.
- `OrchestrationEngine.ts:190` — the accepted-command path, running inside `withTransaction` alongside `eventStore.append` and the projection pipeline.

So this came from the accepted path, mid-transaction.

### What I deliberately cannot conclude

[...]

PR #4837, opened 2026-07-29
fix(server): surface the cause behind a persistence SQL error

Sy-D:
## What Changed

`toPersistenceSqlError` no longer restates the operation as the detail. It carries a payload-safe description of the actual cause instead.

## Why

Reported in #4818. The screenshot there reads:

> SQL error in OrchestrationCommandReceiptRepository.upsert:query: Failed to execute OrchestrationCommandReceiptRepository.upsert:query

The operation name twice, and nothing else. `PersistenceSqlError.message` composes `SQL error in ${operation}: ${detail}`, and the mapper set `detail` to `Failed to execute ${operation}`. The real failure was captured in `cause` and never rendered.

`Errors.test.ts` already asserts this intent, in a test named **"keeps SQL operation context without a tautological detail"** — it builds the error without a detail and expects `SQL error in AuthSessionRepository.list:query`. `toPersistenceSqlError` was the path that did not honour it.

## Notes For Review

**This does not fix the SQL failure in #4818.** It makes it identifiable. The upsert that failed there runs inside `withTransaction` in `OrchestrationEngine.ts:190`, so `SQLITE_BUSY` is plausible — and so is a `Request` schema encode failure, which this mapper would have labelled a "SQL error" despite not being one. Those want opposite fixes and today's message cannot tell them apart, which is presumably why the issue is filed as not reproducible.

**On the payload rule.** The neighbouring test "maps schema errors without copying rejected payloads into diagnostics" establishes that rejected values must not reach diagnostics, so the cause is not copied blindly:

- a schema failure contributes `summarizeSchemaIssue`, which emits issue tags only,
- a driver failure contributes its own message, which names the SQLite condition and at most the constraint it violated, never a bound value,
- anything else leaves the detail unset, which `message` already renders cleanly.

A new test asserts a rejected value cannot appear in the message.

**Blast radius.** This mapper shapes every persistence error message that goes through it. The diff is small, the effect is broad — that is the point, but it is worth a deliberate look.

**Follow-up, deliberately not in this PR.** Newer call sites (`AuthPairingLinks.ts:112`) route a schema cause to `PersistenceDecodeError` rather than labelling it a SQL error, which is the better shape. The comment above `toPersistenceDecodeError` notes these call sites are "being revamped separately", so I left the routing alone and only fixed the message.

## Validation

- `vp run --filter t3 test` — 1690 passed, 7 skipped. No existing test depended on the old tautological detail.
- `vp run --filter t3 typecheck` — no errors

New cases in `Errors.test.ts`:

| cause | resulting message |
| --- | --- |
| `Error("SQLITE_BUSY: database is locked")` | `SQL error in …upsert:query: SQLITE_BUSY: database is locked` |
| schema error | issue tags only; asserts the rejected value is absent |
| a value with no message | `SQL error in …list:query` (unchanged) |

## Checklist

- [x] No UI change, so no screenshots

---

> [!NOTE]
> **Medium Risk**
> The mapper shapes every persistence SQL error message site-wide; behavior changes are intentional but broad, with privacy-sensitive formatting rules that must stay correct.
> 
> **Overview**
> Fixes tautological persistence SQL error messages (operation repeated twice with no real failure hint) by changing how **`toPersistenceSqlError`** sets **`detail`**.
> 
> Instead of always using `Failed to execute <operation>`, it now derives **`detail`** from the cause when it can do so safely: **SQLite** failures get a normalized `SQLITE(<errcode>) <errstr>` string by walking up to four levels of the cause chain (so wrapped driver errors still work), without copying the driver’s unbounded constraint text into **`message`**. **Schema** causes routed through this mapper get issue tags via **`summarizeSchemaIssue`**, not rejected values. Unrecognized causes leave **`detail`** unset so **`message`** stays `SQL error in <operation>`.
> 
> New tests in **`Errors.test.ts`** lock in normalized SQLite details, nested causes, omission of column/table prose from messages, and schema paths that must not leak sentinel payload strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b41332e5771f9f6d88351aa2d2411a044bd01d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Surface SQLite condition and schema issue details in `toPersistenceSqlError`
> - `toPersistenceSqlError` in [Errors.ts](https://github.com/pingdotgg/t3code/pull/4837/files#diff-a49007546447945afb00362617c4fe342b8a0113ab70f400117cb1fe2c48ed61) now derives `detail` from the error cause instead of always setting `'Failed to execute <operation>'`.
> - A new `sqliteCondition` helper walks up to four levels of the cause chain to find `errcode`/`errstr` properties and returns a normalized string like `'SQLITE(1555) constraint failed'`.
> - A new `describeSqlCause` helper categorizes the cause: SQLite errors get the normalized condition string; Schema errors get a summary via issue tag; uncategorizable causes result in `detail` being omitted entirely.
> - Behavioral Change: `detail` on persistence SQL errors changes from the fixed string `'Failed to execute <operation>'` to a normalized SQLite condition, a schema issue tag, or `undefined`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7b41332.</sup>
> 
>
>


Sy-D:
Both points taken, fixed in 9ff106f2d — and the second suggestion turned out to be the better design, so I took that rather than dropping the SQL side entirely.

Limiting `detail` to the schema summary alone would have left a real SQLite failure with an operation-only message, which is the state #4818 is stuck in. Your alternative — a bounded normalized code — keeps the diagnosis without the unbounded text.

SQLite names its condition through a fixed result-code table, and `node:sqlite` exposes it as `errcode` / `errstr`. Both are bounded and carry no query data, unlike the message. Checked against a real failure:

```
message : UNIQUE constraint failed: t.a          <- driver prose, names the column
errcode : 1555
errstr  : constraint failed                       <- fixed table, value-free
```

So `describeSqlCause` now contributes:

- schema failure → `summarizeSchemaIssue`, issue tags only, unchanged
- driver failure → `SQLITE(<errcode>) <errstr>`, read through the wrapping error since the mapper receives the `SqlError`
- anything else → nothing, leaving the operation-only message

Verified end to end against a genuine `node:sqlite` constraint violation on a table holding a sentinel value:

```
message : SQL error in OrchestrationCommandReceiptRepository.upsert:query: SQLITE(1555) constraint failed
detail  : SQLITE(1555) constraint failed
sentinel present in message? false
cause preserved? true
```

On the test comment: the assertion on spliced driver text is gone. The suite now asserts structural attributes — the exact `detail`, the preserved `cause`, that the driver's column name cannot reach the message, and that an uncategorizable cause leaves `detail` unset.

`vp run --filter t3 test` — 1692 passed, 7 skipped. Typecheck and lint clean.

Sy-D:
Checked this against the real construction, and the conclusion does not hold — though the premise is a fair read of the code.

`NodeSqliteClient` does build the error as described, with only a `reason`:

```ts
catch: (cause) => new SqlError({ reason: classifySqliteError(cause, { message: "...", operation: "prepare" }) })
```

But Effect's `SqlError` exposes a `cause` derived from that `reason`, and the classified reason keeps the driver error as *its* cause. So the chain `SqlError → reason → driver error` is walkable, which is what `sqliteCondition` does.

Built that exact shape and ran it through the mapper:

```
classifySqliteError exportiert? function
hat .cause?               true
hat .reason?              true
reason.cause vorhanden?   true
detail : SQLITE(1555) constraint failed
```

Across the failure kinds an upsert can realistically hit:

| failure | detail | driver prose leaked? |
| --- | --- | --- |
| UNIQUE violation | `SQLITE(1555) constraint failed` | no |
| NOT NULL violation | `SQLITE(1299) constraint failed` | no |
| syntax error | `SQLITE(1) SQL logic error` | no |
| missing table | `SQLITE(1) SQL logic error` | no |
| missing column | `SQLITE(1) SQL logic error` | no |

So a `#4818`-style upsert failure does get a categorized detail rather than an operation-only message, which is the outcome the review was worried about.

The walk is bounded at four levels of `cause`. If a future client wraps more deeply than that, the detail would fall back to unset — that is the intended failure mode, not silence about a value.

t3dotgg:
> [!NOTE]

Merged replacement [#9536](https://github.com/pingdotgg/t3code/pull/9536). It retains the original fix, adds the missing Bun SQLite error codes, and tests the real SQL client. The squash commit credits @Sy-D and preserves the original co-author trailer.

Verified database-lock and schema failures through the receipt repository in Node and Bun. CI passed and review findings are resolved. This improves diagnosis. It does not fix the underlying failure in [issue #4818](https://github.com/pingdotgg/t3code/issues/4818).

PR #4838, opened 2026-07-29
fix(claude): classify stop diagnostics as interrupted

D3OXY:
## Summary

- correlate Claude SDK interrupt requests with the active provider turn
- treat a diagnostic-only `error_during_execution` result as interrupted only when it follows an explicit interrupt for that turn
- keep unrelated Claude execution diagnostics classified as failures

Claude can return this after stopping during tool use:

```text
[ede_diagnostic] result_type=user last_content_type=n/a stop_reason=tool_use
```

Without explicit correlation, the adapter classified it as a failed turn and surfaced the internal diagnostic as a runtime error.

## Verification

- `vp test run apps/server/src/provider/Layers/ClaudeAdapter.test.ts` (63 passed)
- `vp fmt --check apps/server/src/provider/Layers/ClaudeAdapter.ts apps/server/src/provider/Layers/ClaudeAdapter.test.ts`
- `vp lint apps/server/src/provider/Layers/ClaudeAdapter.ts apps/server/src/provider/Layers/ClaudeAdapter.test.ts`

> [!NOTE]
> ### Classify diagnostic-only Claude stop results as interrupted when an explicit interrupt was requested
> - Adds `interruptRequestedTurnId` to `ClaudeSessionContext` to track which turn had an explicit interrupt requested, set in `interruptTurn` and cleared after result handling.
> - Introduces `isClaudeExecutionDiagnosticOnlyResult` to detect `error_during_execution` results with a single `[ede_diagnostic]` error string.
> - `isInterruptedResult` and `turnStatusFromResult` now accept an `explicitInterruptRequested` flag; diagnostic-only results are classified as `interrupted` when this flag is true.
> - If the interrupt call fails, the stored turn ID is cleared to avoid misclassifying future results.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 57fd6f8.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
+++ b/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
@@ -3959,4 +3959,61 @@ describe("ClaudeAdapterLive", () => {
       Effect.provide(harness.layer),
     );
   });
+
+  it.effect(
+    "treats a diagnostic-only Claude result after an explicit interrupt as interrupted",
+    () => {
+      const harness = makeHarness();
+      return Effect.gen(function* () {
+        const adapter = yield* ClaudeAdapter;
+
+        const runtimeEventsFiber = yield* Stream.take(adapter.streamEvents, 6).pipe(
+          Stream.runCollect,
+          Effect.forkChild,
+        );
+
+        const session = yield* adapter.startSession({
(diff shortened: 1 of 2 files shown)


juliusmarminge:

PR #4853, opened 2026-07-29
fix(web): remember the rendered-markdown choice across threads

### LINKED ISSUE #4645: [Bug]: Rendered-markdown view for a file in the right sidebar resets to source when you switch threads
### Before submitting

### Area

apps/web

### Steps to reproduce

1. Open a thread and open a `.md` file in the right-hand file preview panel.
2. Click the eye toggle ("Show rendered markdown"). The panel switches to the
   rendered view.
3. Switch to a different thread.
4. Switch back to the original thread and open the same `.md` file.

### Expected behavior

The panel is still showing rendered markdown. The view mode should be sticky —
once I've said I want to read markdown rendered, that's my preference until I
toggle it back.

### Actual behavior

The panel is back to raw markdown source, and I have to click the toggle again
every time I return to the thread.

### Impact

### Version or commit

main @ 5719e8ac

### Environment

macOS 26.5.2, desktop app

### Logs or stack traces

```shell
The toggle is backed by component-local state in `FilePreviewPanel`:

  apps/web/src/components/files/FilePreviewPanel.tsx:675
    const [markdownView, setMarkdownView] = useState<{
      path: string | null;
      revealRequestId: number | null;
    }>({ path: null, revealRequestId: null });

Nothing persists it, and it's invalidated two ways:

- `renderMarkdown` requires `markdownView.path === relativePath`
  (FilePreviewPanel.tsx:681), so any change of the open file drops back to
  source.
- The panel is mounted with `key={environmentId:workspaceRoot}`
  (ChatView.tsx:5588), so switching to a thread in another project remounts it
  and resets the state entirely.

The adjacent explorer-open toggle in the same component already persists via
`FILE_EXPLORER_STORAGE_KEY` in localStorage (FilePreviewPanel.tsx:701-711),
which looks like the precedent to follow here.
```

_No response_

### Workaround

Click the toggle again each time.

Sy-D:
## What Changed

The rendered-markdown choice in the file preview panel is now a persisted preference instead of panel state tied to one file path.

## Why

Reported in #4645. The state was:

```ts
const [markdownView, setMarkdownView] = useState<{ path: string | null; revealRequestId: number | null }>(...)
const renderMarkdown = isMarkdown && markdownView.path === relativePath && ...
```

It said "rendered is on for *this file*", so leaving the thread tore the panel down and returning showed source again. The reporter describes it as a preference — "once I've said I want to read markdown rendered, that's my preference until I toggle it back" — so it now persists the way this same panel already persists its explorer state (`FILE_EXPLORER_STORAGE_KEY` / `initialExplorerOpen`), rather than introducing a new mechanism or touching the settings schema.

A line reveal still wins over the preference, because a line only exists in the source. It no longer clears the preference on the way past, so toggling once returns you to rendered.

## Notes For Review

**A limitation worth knowing, which this PR does not change.** `rightPanelStore` keeps `revealLine` / `revealRequestId` as durable surface state and rehydrates them. So if the last thing you did with a file surface was jump to a line, returning to it still shows source: the stored reveal is indistinguishable, at mount, from a fresh one, and treating it as already handled would break the case the reveal exists for — clicking a line link while the panel is closed.

That is not a regression. Before this change, returning showed source in *every* case; now it shows source only in that one. Making it complete means letting the store consume a reveal once it has been applied, which changes shared surface semantics and its persistence, so it felt like a separate change rather than something to fold in here.

## Validation

Driven in a running client, on a markdown file whose own content links back to itself so the reveal path can be exercised:

| step | expected | observed |
| --- | --- | --- |
| Toggle rendered | preference stored | `t3code.renderMarkdown` = `true`, view rendered |
| Click a `README.md:9` link while rendered | source, line revealed | toggle flipped to source; preference still `true` |
| Toggle back | rendered | rendered |
| Leave the thread and return, no pending reveal | rendered | rendered — verified by the reporter |

The last row is the reported scenario. I could not reproduce it myself: the file explorer in my seeded fixture workspace lists no files, so I had no way to open a markdown file without going through a line link. It was checked by the reporter on a real project instead, and I am flagging that rather than implying I measured it.

- `vp run --filter @t3tools/web test` — 1663 passed
- `vp lint` / `vp fmt --check` on the changed file — clean

Closes #4645.

## Checklist

- [x] No visual change beyond which of the two existing views is shown

---

> [!NOTE]
> **Low Risk**
> UI preference persistence in the file preview panel only; no auth, data, or API changes.
> 
> **Overview**
> Fixes #4645 by persisting the file preview **rendered markdown** toggle in `localStorage` (`t3code.renderMarkdown`) via `useLocalStorage`, matching how explorer open/close is already stored—not per-file panel state that reset when switching threads.
> 
> **Line reveals** still force **source** view (the line only exists there). A new `handledReveal` state pairs path with `revealRequestId` so a reveal handled on one file does not block reveals on another; toggling rendered again records the current reveal as handled without clearing the stored preference.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3238bcb4f5763aa4568d9290f02b7b6711ae62ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Persist rendered-markdown toggle choice across threads in `FilePreviewPanel`
> - Replaces the local `markdownView` state with a `useLocalStorage`-backed preference keyed at `t3code.renderMarkdown`, so the toggle choice survives navigation between threads and files.
> - When a line reveal is active for a file and hasn't been marked as handled for that path/requestId, the panel forces source view regardless of the stored preference.
> - Toggling to rendered markdown records the current file path and requestId as handled, preventing the reveal logic from overriding the preference again for the same request.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3238bcb.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/files/FilePreviewPanel.tsx
+++ b/apps/web/src/components/files/FilePreviewPanel.tsx
@@ -685,14 +685,20 @@ export default function FilePreviewPanel({
   // Reading markdown rendered is a preference, not a property of one file. Keeping
   // it on the panel meant a thread switch dropped it and forced source back.
   const [renderMarkdownPreferred, setRenderMarkdownPreferred] = useState(initialRenderMarkdown);
-  const [revealHandledRequestId, setRevealHandledRequestId] = useState<number | null>(null);
+  // Paired with the path on purpose: each file surface counts its reveals from
+  // one, so a bare id would let a dismissed reveal on one file swallow the first
+  // reveal on the next.
+  const [handledReveal, setHandledReveal] = useState<{ path: string; requestId: number } | null>(
+    null,
+  );
   const breadcrumbRef = useRef<HTMLDivElement>(null);
   const isMarkdown = relativePath ? isMarkdownPreviewFile(relativePath) : false;
   // A reveal still wins over the preference: the line only exists in the source.
   const renderMarkdown =
     isMarkdown &&
     renderMarkdownPreferred &&
-    (revealLine === null || revealHandledRequestId === revealRequestId);
+    (revealLine === null ||
+      (handledReveal?.path === relativePath && handledReveal.requestId === revealRequestId));
   const canOpenInBrowser =
     relativePath !== null && isPreviewSupportedInRuntime() && isBrowserPreviewFile(relativePath);
   const absolutePath = relativePath ? resolvePathLinkTarget(relativePath, cwd) : null;
@@ -800,7 +806,11 @@ export default function FilePreviewPanel({
                     pressed={renderMarkdown}
                     onPressedChange={(pressed) => {
                       setRenderMarkdownPreferred(pressed);
-                      setRevealHandledRequestId(pressed ? revealRequestId : null);
+                      setHandledReveal(
+                        pressed && relativePath !== null
+                          ? { path: relativePath, requestId: revealRequestId }
+                          : null,
+                      );
                       try {
                         setLocalStorageItem(RENDER_MARKDOWN_STORAGE_KEY, pressed, Schema.Boolean);
                       } catch (error) {
--- a/apps/web/src/components/files/FilePreviewPanel.tsx
+++ b/apps/web/src/components/files/FilePreviewPanel.tsx
@@ -79,6 +79,7 @@ interface FilePreviewPanelProps {
 }
 
 const FILE_EXPLORER_STORAGE_KEY = "t3code.fileExplorerOpen";
+const RENDER_MARKDOWN_STORAGE_KEY = "t3code.renderMarkdown";
 const FILE_SAVE_DEBOUNCE_MS = 500;
 const FILE_LINK_REVEAL_ATTRIBUTE = "data-file-link-reveal";
 const FILE_LINK_REVEAL_UNSAFE_CSS = `
@@ -645,6 +646,15 @@ function initialExplorerOpen(): boolean {
   }
 }
 
+function initialRenderMarkdown(): boolean {
+  try {
+    return getLocalStorageItem(RENDER_MARKDOWN_STORAGE_KEY, Schema.Boolean) ?? false;
+  } catch (error) {
+    console.error(error);
+    return false;
+  }
+}
+
 export default function FilePreviewPanel({
   environmentId,
   cwd,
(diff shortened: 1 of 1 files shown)


Sy-D:
Right on both counts, and it was a regression I introduced — the state I replaced paired the path with the request id, and I dropped the path. Fixed in a533f3a8d.

Confirmed the precondition rather than taking it on faith. `rightPanelStore` derives the id per surface:

```ts
const surfaceId = `file:${relativePath}`;
const surface = fileSurface(relativePath, normalizeRevealLine(line), (existing?.revealRequestId ?? 0) + 1);
```

so every file starts counting at one. Reproduced with two markdown files that link to each other by line, and read the persisted store afterwards:

```
README.md: revealLine=9 requestId=1
NOTES.md:  revealLine=5 requestId=1
```

Both at 1, exactly the collision described. Walked through it in a running client:

| step | expected | observed |
| --- | --- | --- |
| Click `README.md:9` | source, line revealed | source |
| Toggle back to rendered | rendered, reveal dismissed | rendered |
| Click `NOTES.md:5` from there | source — different file, its own first reveal | source |

Without the pairing the last row would have stayed rendered, since the stored id 1 matched.

`handledReveal` now carries `{ path, requestId }` and the comparison checks both.

`vp run --filter @t3tools/web test` — 1663 passed. Typecheck and lint clean.

apps/web/src/components/files/FilePreviewPanel.tsx
```diff
@@ -672,16 +682,23 @@ export default function FilePreviewPanel({
   const isImage = relativePath !== null && isWorkspaceImagePreviewPath(relativePath);
   const file = useProjectFileQuery(environmentId, cwd, relativePath, !isImage);
   const [explorerOpen, setExplorerOpen] = useState(initialExplorerOpen);
-  const [markdownView, setMarkdownView] = useState<{
-    path: string | null;
-    revealRequestId: number | null;
-  }>({ path: null, revealRequestId: null });
+  // Reading markdown rendered is a preference, not a property of one file. Keeping
+  // it on the panel meant a thread switch dropped it and forced source back.
+  const [renderMarkdownPreferred, setRenderMarkdownPreferred] = useState(initialRenderMarkdown);
```
juliusmarminge: do we not have a `useLocalStorage` hook yet??

Sy-D:
oh yea we do and its used by `editorPreferences`, `providerUpdateDismissal` and a couple of others. Switched to it in 3238bcb4f.

I had followed the neighbouring pattern in this file instead: `FILE_EXPLORER_STORAGE_KEY` with `initialExplorerOpen()` and a manual `setLocalStorageItem` in the toggle, a few lines above where this landed. Matching the file seemed safer than reaching past it, but the hook is more sleek than using multiple try catch..

PS: the explorer state right above still uses the hand-rolled version. Happy to convert it too, but it is unrelated to this fix so I left it alone rather than widening the diff.

apps/web/src/components/files/FilePreviewPanel.tsx
Sy-D: We do see comment below @juliusmarminge fixed it and why this approach has been chosen previously.

PR #4884, opened 2026-07-29
fix(web): stop the transcript scrolling itself on every streamed chunk

AsimNet:
The chat transcript follows the live edge while an agent streams, which fires a scroll once per chunk. Over a long session that continuous movement is tiring to read, and there is no way to turn it off.

## How it's fixed

Two per-device client settings in **Settings → Appearance**:

- **`chatAutoScroll`** gates live-follow, defaulting to **off**. The transcript holds still while output streams; the send-time anchor still positions the new turn once, so you are not left looking at old messages; and the scroll-to-end pill becomes the way back to the edge.
- **`reduceMotion`** forces the reduced-motion path on regardless of the OS preference — scrolls become instant instead of animated, and the looping indicator animations stop (status pulse, skeleton shimmer, working indicator, ultrathink gradients).

Two details worth review attention:

The pill's visibility guard needed the setting too. It treats "left the live edge" as the list settling rather than the reader navigating — which is only true while something is actually following. Without that, turning follow off hid the pill exactly when it became the only way back.

Spinners are deliberately left spinning under reduce-motion. A frozen spinner reads as a hung app, which is worse than the motion it removes.

## Verification

`tsgo --noEmit`, `vp test run --project unit` (1666 tests), `vp lint`, and `vp build` all pass on this branch against current `main`. `MessagesTimeline.test.tsx` gains a case asserting the list only re-pins to the end when the setting is on.

## Screenshots

The new rows in Settings → Appearance:

Appearance settings

This change is behavioural rather than visual — the difference is motion during streaming, which a still image can't show. Happy to record a short clip if that's wanted before review.

---

Written by Claude Opus 4.5 in Claude Code.

---

> [!NOTE]
> **Medium Risk**
> Default-off live-follow changes behavior for all users during streaming, and scroll-mode/pill logic in ChatView is easy to regress; changes are UI-only with no auth or data impact.
> 
> **Overview**
> **Behavior change:** streaming transcripts no longer auto-follow the live edge by default. A new **`chatAutoScroll`** client setting (default **off**) gates per-chunk live-follow in `ChatView` and LegendList **`maintainScrollAtEnd`** in `MessagesTimeline`. Send-time turn anchoring still runs once; the scroll-to-end pill is shown when the reader is behind the edge, with settling guards updated so the pill is not suppressed when follow is off.
> 
> Adds **`reduceMotion`** (default off) plus a shared **`prefersReducedMotion`** helper and **`data-reduce-motion`** on `<html>` so scroll jumps (anchor, pill, minimap), draft-hero / mobile composer transitions, and looping status animations respect the setting or OS **`prefers-reduced-motion`** (spinners stay animated).
> 
> **Settings → Appearance** exposes both toggles, restore-defaults, and user docs. Chat timeline tests assert list re-pin only when auto-scroll is enabled. Minor RTL-oriented class swaps (`ps`/`ms`/`start`, **`force-ltr`** on tool output).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c376c94a79a2ae9db95933993bad6991a89fe54a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Stop transcript auto-scrolling on every streamed chunk with a new "Follow agent output" setting
> - Adds `chatAutoScroll` and `reduceMotion` boolean client settings (both defaulting to `false`) in [`settings.ts`](https://github.com/pingdotgg/t3code/pull/4884/files#diff-82d7525e9c2a6a61cd5d416ceaeba40992140373eb7ac00fa3a91c2edd00bc3c); when `chatAutoScroll` is off the message list no longer re-pins to the end during streaming.
> - Introduces a settle window (`LIVE_FOLLOW_SETTLE_MS`) in [`ChatView.tsx`](https://github.com/pingdotgg/t3code/pull/4884/files#diff-90b4b4dee89df040fe8cb45b1788c1d5876bd5092229a414007b0359c2142f4f) to suppress the scroll-to-end pill from flashing during thread switches and post-send transitions.
> - Adds a `prefersReducedMotion` utility in [`reducedMotion.ts`](https://github.com/pingdotgg/t3code/pull/4884/files#diff-bddb06905dd8bf21efd91f023404a94a184e596c77f490872f52a1abd3814607) that checks both the new setting and the OS `prefers-reduced-motion` media query; mirrors the result to `html[data-reduce-motion]` for CSS.
> - CSS in [`index.css`](https://github.com/pingdotgg/t3code/pull/4884/files#diff-70ad657a634ce55957729ff8d1c93f697cc7e3ca2a5cb5fe6b00c90d7d515fcd) disables continuous animations (pulses, shimmers, view transitions) when reduced motion is active.
> - Exposes both settings as toggles in the Appearance settings panel in [`SettingsPanels.tsx`](https://github.com/pingdotgg/t3code/pull/4884/files#diff-9dbbb4f86928d777550dead352b372341ccfb3f589782aa6d99b98f792c89d18).
> - Behavioral Change: existing users default to `chatAutoScroll: false`, meaning auto-follow of streaming output is **off by default** for all users.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c376c94.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/ChatView.tsx
+++ b/apps/web/src/components/ChatView.tsx
@@ -166,7 +166,8 @@ import {
 import { newDraftId, newMessageId, newThreadId } from "~/lib/utils";
 import { getProviderModelCapabilities, resolveSelectableProvider } from "../providerModels";
 import { NO_PROVIDER_MODEL_SELECTION } from "../providerInstances";
-import { useClientSettings, useEnvironmentSettings } from "../hooks/useSettings";
+import { getClientSettings, useClientSettings, useEnvironmentSettings } from "../hooks/useSettings";
+import { prefersReducedMotion } from "../reducedMotion";
 import { useNowMinute } from "../hooks/useNowMinute";
 import { useNewThreadHandler } from "../hooks/useHandleNewThread";
 import { resolveAppModelSelectionForInstance } from "../modelSelection";
@@ -307,6 +308,12 @@ import {
 } from "../versionSkew";
 import { useAssetUrls } from "../assets/assetUrls";
 
+// How long after live-follow takes the scroll position we still treat a
+// not-at-end report as that scroll settling rather than the reader moving. Only
+// consulted when follow-output is off; while it is on, live-follow owns the
+// position for as long as it holds it.
+const LIVE_FOLLOW_SETTLE_MS = 500;
+
 const IMAGE_ONLY_BOOTSTRAP_PROMPT =
   "[User attached one or more images without additional text. Respond using the conversation context and the attached image(s).]";
 const EMPTY_ACTIVITIES: OrchestrationThreadActivity[] = [];
@@ -333,9 +340,7 @@ function useDraftHeroLayoutTransition(isDraftHeroState: boolean) {
     const transitionGroup = transitionGroupRef.current;
     const nextComposerRect = composerAnchorRef.current?.getBoundingClientRect() ?? null;
     const stateChanged = previousStateRef.current !== isDraftHeroState;
-    const prefersReducedMotion =
-      typeof window !== "undefined" &&
-      window.matchMedia?.("(prefers-reduced-motion: reduce)").matches;
+    const reducedMotion = prefersReducedMotion(getClientSettings().reduceMotion);
     const mobileComposerTransitionActive =
       typeof document !== "undefined" &&
       document.documentElement.dataset.mobileComposerRouteTransition === "true";
@@ -346,7 +351,7 @@ function useDraftHeroLayoutTransition(isDraftHeroState: boolean) {
     const previousComposerRect = previousComposerRectRef.current;
     if (
       stateChanged &&
-      !prefersReducedMotion &&
+      !reducedMotion &&
       !mobileComposerTransitionActive &&
       transitionGroup &&
       previousComposerRect &&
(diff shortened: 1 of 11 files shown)


AsimNet:
Thanks — all four findings were real. Fixed and force-pushed as one clean commit.

**`packages/contracts/tmp-rexec.ts` — dropped.** Correct call: it was a scratch remote-exec script that a `git add -A` swept in, unrelated to this change, outside `src` so the package never typechecked it, and it hardcoded a private tailnet IP. It is gone from the branch; the follow-on notes about the buffer never being trimmed, `process.exit()` truncating piped stdout, and `join(" ")` losing argument boundaries were all fair, and moot now that the file is out.

**Undefined `--safe-area-inset-start/end`** — real bug, and my fault for how I split this work. Those logical variables and their `[dir="rtl"]` overrides belong to a separate RTL change; carving this PR out brought the *use* without the *definition*, so the four declarations were invalid and the header lost its horizontal padding on mobile web. Reverted to the original `env(safe-area-inset-left/right)` form. Good catch — a browser silently dropping an invalid `calc()` is exactly the kind of thing that survives a typecheck.

**Scroll pill flashing on thread switch** — also real, and the fix needed more than the gate I had. Leaving the live edge without a user gesture means a programmatic scroll is still settling rather than the reader moving away. With follow on, live-follow owns the position for as long as it holds it. With it off, only the window just after a thread switch or a send is settling, because LegendList reports not-at-end while `initialScrollAtEnd` lands. So suppression is now keyed on *both*: the existing generation match plus, when follow is off, a short settle window stamped whenever live-follow takes the position. Suppressing indefinitely in that mode would hide the pill exactly when it is the only way back; not suppressing at all would flash it on every switch.

**Bouncing provider-update indicator** — right, and it undercut the whole point of the setting. `ProviderInstanceCard` uses an arbitrary `[animation:bounce_...]` class, which the `animate-*` selectors never matched, and its `motion-reduce:` variant only answers to the OS preference. Both the media-query and `data-reduce-motion` paths now cover it. Verified in the built stylesheet rather than the source, since the minifier escapes the colon in that selector.

`tsgo --noEmit`, 1666 unit tests, `vp lint` and `vp build` all pass on the updated branch.

juliusmarminge:

**Reason:** Jul conflicted transcript auto-scroll; streaming list virtualization landed later.
**Related:** #9688

PR #4887, opened 2026-07-29
feat(release): publish a Linux .deb alongside the AppImage

NaveDanan:
## Problem

Linux is the only platform where T3 Code ships a single artifact. The AppImage works, but installing it leaves users without a desktop entry, an icon, dependency resolution, or any package manager integration unless they wire all of it up by hand. `.deb` covers the large majority of Linux desktop users and costs one extra packaging pass over an app we have already compiled.

## What changed

**Made `--target deb` actually work.** `scripts/build-desktop-artifact.ts` already forwarded `--target` straight through to electron-builder, so `deb` looked supported. It was not: fpm aborts unless it can resolve a project homepage and a `Name <email>` maintainer, and the generated config set neither, because AppImage never needed them.

```
⨯ Please specify project homepage, see https://www.electron.build/configuration#metadata
  Please specify author 'email' in the application package.json
  It is required to set Linux .deb package maintainer.
```

Two fields fix it — `homepage` on the staged `package.json` and `linux.maintainer`, which becomes both the deb `Maintainer` and `Vendor`.

**Release workflow builds and publishes it.** A platform-gated step reruns packaging with `--skip-build`, reusing the app the AppImage step already compiled, then `.deb` is added to the collect and publish globs. This deliberately avoids a second Linux matrix leg, which would have duplicated a ~25 minute compile and collided on the `desktop-linux-x64` and `resource-monitor-linux-x64` artifact names.

**Guarded the updater manifest.** `FpmTarget` sets `isWriteUpdateInfo: true`, exactly like `AppImageTarget`, so the deb build overwrites `latest-linux.yml`. `DesktopUpdates.ts` refuses to install any Linux update unless the app is running from an AppImage, so an unguarded version of this change would publish a manifest pointing at a `.deb` the app rejects and silently break auto-update for every existing AppImage user. The workflow stashes the manifest and restores it after packaging.

Also adds `vp run dist:desktop:linux:deb` for local parity with `dist:desktop:linux`, and documents the artifact plus the update caveat.

## Verification

Built end-to-end on Ubuntu 24.04 at v0.0.31:

```
Package: t3code          Architecture: amd64
Version: 0.0.31          Maintainer: pingdotgg <support@t3.codes>
Homepage: https://t3.codes
Depends: libgtk-3-0, libnotify4, libnss3, libxss1, libxtst6,
         xdg-utils, libatspi2.0-0, libuuid1, libsecret-1-0
```

- All 9 hicolor icon sizes present; packaged icon is pixel-identical to `assets/prod/black-universal-1024.png`.
- Valid desktop entry with `StartupWMClass=t3code`; postinst registers `/usr/bin/t3code` and handles `chrome-sandbox` permissions.
- Every declared dependency resolves on 24.04 (`libgtk-3-0` and `libatspi2.0-0` via the `t64` packages that `Provide` them).
- New regression test fails if `linux.maintainer` is removed, so the build break cannot silently return.
- `vp test run scripts/build-desktop-artifact.test.ts` (31 passed), `scripts` typecheck, lint, and `fmt:check` all clean.
- Workflow shell logic simulated in isolation under `set -euo pipefail`, including the case where no manifest exists.

## Notes for reviewers

**The maintainer address needs your call.** The repo has no packaging contact anywhere, so `support@t3.codes` is a placeholder derived from the site. It gets baked permanently into every published package — please replace it if there is a real address.

**In-app updates are intentionally not enabled for deb.** The `.deb` ships both `app-update.yml` and electron-builder's `package-type` marker, so `electron-updater` would select `DebUpdater`, but T3's own Linux gate blocks it first and deb users will see "Automatic updates on Linux require running the AppImage build." Turning that on means accepting a `pkexec dpkg -i` prompt per update and pointing the feed at the deb, which is a separate decision I did not make here. Most distro-packaged apps leave updates to the package manager.

**Scope.** `.snap` was built and verified alongside this but is deliberately excluded: it needs a Snap Store account, credentials, and manual review for classic confinement, which is a product decision rather than a packaging one. Happy to open it separately if wanted.

Per `CONTRIBUTING.md` I recognise this may not be something you want right now — close it freely if so.

Generated with claude-opus-5 running in opencode.

---

> [!NOTE]
> **Medium Risk**
> Release workflow changes affect Linux updater metadata and published assets; incorrect manifest handling could break AppImage auto-update, though the backup/restore step mitigates that. Maintainer email is baked into every deb package.
> 
> **Overview**
> Adds a **Linux `.deb`** to release artifacts so Debian/Ubuntu users can install via the package manager, while **in-app auto-update stays AppImage-only**.
> 
> The desktop packager now supplies **homepage** and **`linux.maintainer`** (fpm requirements for `deb`) on the staged `package.json` and Linux build config, plus a **`dist:desktop:linux:deb`** script and a regression test so deb builds cannot break silently.
> 
> The **release workflow** runs a second Linux packaging pass with `--skip-build` after the AppImage step, **backs up and restores** `*-linux*.yml` updater manifests (deb packaging would overwrite them with `.deb` metadata the app refuses to install), and includes `.deb` in collect/publish globs. Docs note the deb vs AppImage update behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a37813e526d223c617a8cd1cf225b5691a603076. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Publish a Linux .deb package alongside the AppImage in release builds
> - Adds a new build step in [release.yml](.github/workflows/release.yml) that runs `dist:desktop:artifact` with `--target deb` after the AppImage is built, stashing and restoring the AppImage update manifest so it isn't overwritten by the `.deb` build.
> - Adds `maintainer` and `homepage` metadata to the Linux build config in [build-desktop-artifact.ts](scripts/build-desktop-artifact.ts), which are required by electron-builder's fpm-backed targets like `.deb`.
> - Adds a `dist:desktop:linux:deb` npm script in `package.json` for local builds.
> - Behavioral Change: in-app updates are disabled when running from the `.deb` package; only the AppImage supports update delivery.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a37813e.</sup>
> 
>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -573,6 +573,37 @@ jobs:
 
           vp run dist:desktop:artifact "${args[@]}"
 
+      - name: Build Linux deb package
+        if: matrix.platform == 'linux'
+        shell: bash
+        run: |
+          set -euo pipefail
+          shopt -s nullglob
+
+          # electron-builder writes update metadata for every fpm target, so the
+          # deb build would overwrite the AppImage updater manifest. Linux
+          # auto-update only supports the AppImage, so stash the manifest and put
+          # it back once the deb is packaged.
+          manifest_backup="$RUNNER_TEMP/linux-update-manifest"
+          mkdir -p "$manifest_backup"
+          for manifest in release/*-linux*.yml; do
+            cp "$manifest" "$manifest_backup/"
+          done
+
+          # --skip-build reuses the app compiled by the previous step; this only
+          # reruns packaging.
+          vp run dist:desktop:artifact \
+            --platform linux \
+            --target deb \
+            --arch "${{ matrix.arch }}" \
+            --build-version "${{ needs.preflight.outputs.version }}" \
+            --skip-build \
+            --verbose
+
+          for manifest in "$manifest_backup"/*.yml; do
+            cp "$manifest" release/
+          done
+
       - name: Collect release assets
         shell: bash
         run: |
@@ -584,6 +615,7 @@ jobs:
             "release/*.dmg" \
             "release/*.zip" \
             "release/*.AppImage" \
+            "release/*.deb" \
             "release/*.exe" \
             "release/*.blockmap" \
             "release/*.yml"; do
@@ -804,6 +836,7 @@ jobs:
             release-assets/*.dmg
             release-assets/*.zip
             release-assets/*.AppImage
+            release-assets/*.deb
             release-assets/*.exe
             release-assets/*.blockmap
             release-assets/*.yml
@@ -824,6 +857,7 @@ jobs:
             release-assets/*.dmg
             release-assets/*.zip
             release-assets/*.AppImage
+            release-assets/*.deb
             release-assets/*.exe
             release-assets/*.blockmap
             release-assets/*.yml
--- a/docs/operations/release.md
+++ b/docs/operations/release.md
@@ -11,10 +11,10 @@ This document covers the unified release workflow for stable and nightly desktop
   - manual `workflow_dispatch` for either channel
 - Runs quality gates first: lint, typecheck, test.
 - Reads the shared production T3 Connect relay URL and Clerk client configuration before packaging clients.
-- Builds four artifacts in parallel for both channels:
+- Builds four platform bundles in parallel for both channels:
   - macOS `arm64` DMG
   - macOS `x64` DMG
-  - Linux `x64` AppImage
+  - Linux `x64` AppImage, plus a `.deb` repackaged from the same build
   - Windows `x64` NSIS installer
 - Publishes one GitHub Release with all produced files.
   - Stable tags with a suffix after `X.Y.Z` (for example `1.2.3-alpha.1`) are published as GitHub prereleases.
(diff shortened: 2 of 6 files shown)


masterflitzer:
you can just use gear lever to manage appimages, it has desktop integration & supports auto update: <https://github.com/mijorus/gearlever>

NaveDanan:
@masterflitzer Thanks — Gear Lever is definitely a valid option for people who want to manage AppImages.

I'm not making any claim about Gear Lever's maintainer or the quality of the project. My concern is that, because of what it does, its Flatpak package currently requires fairly broad permissions, including read/write access to the host filesystem, network access, /tmp, and access to the Flatpak interface. That's also why Flathub marks it as "potentially unsafe."
image
More importantly, though, I don't think users should need to install a separate AppImage manager just to get normal desktop/package integration for T3 Code.

The goal of this PR is to make T3 Code itself integrate natively with the Debian/Ubuntu ecosystem: install/uninstall through the system package manager, dependency resolution, desktop entry/icon integration, and a familiar installation path for users who may not even know what an AppImage manager is.

I think that's particularly worthwhile for a project like T3 Code. Making the Linux experience closer to the Windows/macOS experience lowers the barrier for non-developers as well, and gives the project a better foundation for future distro/repository distribution.

masterflitzer:
@NaveDanan i get your point, but imo time & additional maintenance burden is better invested in more/better cross distro solutions, e.g. something specific for debian/ubuntu won't be useful for someone using fedora etc.

another option that comes to mind is to extend homebrew support of t3code from macos only to linux brew as well, that way everybody on linux benefits from a simpler update experience and i guess it'll be much easier to maintain given t3code already supports brew (assuming one is willing to use brew of course, but gh & glab clis for instance which are supported by t3code are also primarily distributed on linux brew, i'd imagine for similar reasons)

all this is obviously for the mainteners to decide, i'm just trying to give helpful input as a fan of this project

t3dotgg:
> [!NOTE]

Closing this PR after an automated pass over open pull requests. Theo already rejected adding more Linux package formats to release CI. External package mirrors exist.

PR #4896, opened 2026-07-30
fix(desktop): resolve claude executable on windows

### LINKED ISSUE #4846: [Bug] Windows desktop resolver ignores User PATH, so Claude in %USERPROFILE%\.local\bin is not found
## Summary

On Windows, T3 Code can report Claude as missing even when the native installer has installed a working executable at `%USERPROFILE%\.local\bin\claude.exe` and that directory is present in the user's persistent `Path` setting.

The desktop environment resolver attempts to repair `PATH`, but its PowerShell probe reads the inherited process environment instead of the User/Machine environment registry values. If the Electron process was started with a stale `PATH`, the repair retains that stale value and misses `%USERPROFILE%\.local\bin`.

## Reproduction

1. Install Claude Code with the native Windows installer. It creates `%USERPROFILE%\.local\bin\claude.exe`.
2. Ensure `%USERPROFILE%\.local\bin` is in the **User** `Path` setting (for example, inspect it with `[Environment]::GetEnvironmentVariable('Path', 'User')`).
3. Start T3 Code from an environment whose inherited `PATH` predates that setting or otherwise omits the directory.
4. Select/use the local Claude provider.

Actual: T3 Code cannot find `claude`; an integrated PowerShell also fails with `Get-Command claude`.

Expected: T3 Code should discover the executable from the current User/Machine PATH or explicitly include the standard Claude native installation directory.

## Evidence

On the affected machine:

```powershell
& "$HOME\.local\bin\claude.exe" --version
# 2.1.220 (Claude Code)

[Environment]::GetEnvironmentVariable('Path', 'User')
# ...;C:\Users\<user>\.local\bin

$env:PATH -split ';' | Select-String '\\.local\\bin'
# no result
```

The application process tree is T3 Code -> Codex app-server -> integrated PowerShell, so the backend and shells inherit the unresolved desktop environment.

## Root cause in source

`apps/desktop/src/shell/DesktopShellEnvironment.ts` runs PowerShell with a command generated by `captureWindowsEnvironmentCommand`. That command uses:

```powershell
[Environment]::GetEnvironmentVariable('PATH')
```

The one-argument overload reads the *current process environment*, not the persistent user or machine variables. Therefore both the `-NoProfile` and profile probes repeat the stale inherited PATH. `installWindowsEnvironment` merges those values with a fixed list of Node-related directories; that list includes Bun and Scoop but not `%USERPROFILE%\.local\bin`.

## Suggested fix

Read and merge the persistent values explicitly, then merge them with the current process PATH. For example:

```powershell
[Environment]::GetEnvironmentVariable('Path', 'User')
[Environment]::GetEnvironmentVariable('Path', 'Machine')
```

Alternatively/additionally, include `%USERPROFILE%\.local\bin` in `knownWindowsCliDirs`, which is the standard location used by the Claude Code native Windows installer.

A regression test should start with a stale process PATH, provide a User PATH containing `C:\Users\test\.local\bin`, and assert the installed PATH contains it.

shubhu121:
Fixes pingdotgg/t3code#4846

On Windows, T3 Code would report Claude as missing if it was installed in the user's persistent PATH but the process inherited a stale PATH.

This fixes the desktop environment resolver to explicitly read and merge the `Machine`, `User`, and `Process` environment variables for `PATH` on Windows, and adds `%USERPROFILE%\.local\bin` to the known CLI directories fallback.

---

> [!NOTE]
> **Low Risk**
> Windows-only desktop startup PATH probing; improves CLI discovery without touching auth, data, or network paths.
> 
> **Overview**
> Fixes false “Claude not found” on Windows when Claude lives on the **User** or **Machine** PATH but the Electron process inherited a stale **Process** PATH.
> 
> **`captureWindowsEnvironmentCommand`** now treats `PATH` specially: the PowerShell probe reads **Machine**, **User**, and **Process** values via `[Environment]::GetEnvironmentVariable`, concatenates non-empty segments, and emits that merged string between the existing markers. Other env vars still use the single-scope lookup unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c5a2c95f6e9c946f50d50f330dc8b55a7fa04d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix Claude executable resolution on Windows by merging Machine, User, and Process PATH scopes
> The `captureWindowsEnvironmentCommand` util in [DesktopShellEnvironment.ts](https://github.com/pingdotgg/t3code/pull/4896/files#diff-ed26aa44c93ddc9fedab8bc7cf7aa64c7b0b8fc0e22a3960d5c02c30c940dabd) previously captured only the process-scoped PATH, which could miss entries added by installers to the Machine or User scopes. PATH is now resolved by fetching all three scopes separately, filtering out null/empty entries, and joining them with `;`. Behavioral Change: the captured PATH value will now be broader than the process-scoped value alone, which may surface executables not previously visible.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0c5a2c9.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/shell/DesktopShellEnvironment.ts
+++ b/apps/desktop/src/shell/DesktopShellEnvironment.ts
@@ -242,6 +242,17 @@ const captureWindowsEnvironmentCommand = (names: ReadonlyArray<string>) =>
   [
     "$ErrorActionPreference = 'Stop'",
     ...names.flatMap((name) => {
+      if (name.toUpperCase() === "PATH") {
+        return [
+          `Write-Output '${startMarker(name)}'`,
+          `$machineValue = [Environment]::GetEnvironmentVariable('${name}', 'Machine')`,
+          `$userValue = [Environment]::GetEnvironmentVariable('${name}', 'User')`,
+          `$processValue = [Environment]::GetEnvironmentVariable('${name}')`,
+          `$merged = ($machineValue, $userValue, $processValue | Where-Object { $null -ne $_ -and $_.Length -gt 0 }) -join ';'`,
+          "if ($null -ne $merged -and $merged.Length -gt 0) { Write-Output $merged }",
+          `Write-Output '${endMarker(name)}'`,
+        ];
+      }
       return [
         `Write-Output '${startMarker(name)}'`,
         `$value = [Environment]::GetEnvironmentVariable('${name}')`,
--- a/apps/desktop/src/shell/DesktopShellEnvironment.ts
+++ b/apps/desktop/src/shell/DesktopShellEnvironment.ts
@@ -248,7 +248,7 @@ const captureWindowsEnvironmentCommand = (names: ReadonlyArray<string>) =>
           `$machineValue = [Environment]::GetEnvironmentVariable('${name}', 'Machine')`,
           `$userValue = [Environment]::GetEnvironmentVariable('${name}', 'User')`,
           `$processValue = [Environment]::GetEnvironmentVariable('${name}')`,
-          `$merged = ($machineValue, $userValue, $processValue | Where-Object { $null -ne $_ -and $_.Length -gt 0 }) -join ';'`,
+          `$merged = ($processValue, $userValue, $machineValue | Where-Object { $null -ne $_ -and $_.Length -gt 0 }) -join ';'`,
           "if ($null -ne $merged -and $merged.Length -gt 0) { Write-Output $merged }",
           `Write-Output '${endMarker(name)}'`,
         ];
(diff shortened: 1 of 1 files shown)


CDVolvik (commented):
Windows check against today's `main` (`6ae9662d8`).

Half of this PR is already on main. `knownWindowsCliDirs` already includes `%USERPROFILE%\.local\bin` (and this box has that directory). The remaining unique change is the PATH capture: Machine + User + Process, joined.

That join is the better shape for the desktop profile probe. `[Environment]::GetEnvironmentVariable('PATH')` with no target is Process scope, so after `loadProfile: true` it still sees fnm/profile prepends. #6356 reads only Machine+User for *both* probes, so `profile.PATH` and `noProfile.PATH` become the same registry string. If these two land independently, keep this merge and drop #6356's PATH rewrite (the Codex dir addition in #6356 can stay).

This PR only edits `DesktopShellEnvironment.ts`. `packages/shared/src/shell.ts` `buildWindowsEnvironmentCaptureCommand` — the path `fixPath()` uses on the server — is unchanged, so a refresh after install still misses User-PATH writes on the shared side.

CONFLICTING against current main; needs a rebase before the merge is real. No tests in the diff, so I did not get a revert-fail.

juliusmarminge:
Closing as superseded by #5074, which has merged and covers this change on the current architecture. Thanks for the contribution.

PR #4199, opened 2026-07-20
fix(server): keep provider sessions alive during background agent work

ChamaruAmasara:
Closes #4198

## What Changed

- Add a targeted `touchLastSeen` to the provider session runtime repository and session directory: a lightweight update that bumps only `last_seen_at`, and only for non-`stopped` rows (so a reaped session is never resurrected).
- In `ProviderService`, refresh a session's `last_seen_at` from runtime activity: `processRuntimeEvent` now calls `touchLastSeen` when the adapter emits runtime events, throttled per thread (default 60s) so an event burst is at most one lightweight write per window. The touch is best-effort — a failed write never breaks event processing.
- Tests: persistence/directory tests for `touchLastSeen` (bumps a live row; leaves a stopped row untouched); a `ProviderService` test proving a runtime event triggers a touch and that a rapid burst is throttled.

## Why

`ProviderSessionReaper` stops any provider session after 30 min of `last_seen_at` inactivity. Background agent work — dynamic workflows and subagents — keeps running after the foreground turn settles, but at that point the adapter session has gone `ready` with no active turn, and nothing refreshes `last_seen_at`. So a long-running background workflow with no foreground turn is torn down mid-run, and the in-flight background orchestration is lost.

Runtime events are the reliable signal for background activity: while a workflow runs, the adapter keeps emitting events (e.g. `task.progress` from subagents) even with no active foreground turn. Refreshing `last_seen_at` from these events keeps the session warm exactly when the reaper would otherwise reap it. Throttling keeps DB churn negligible (one write per thread per window instead of one per event).

## Validation

- `vp test run apps/server/src/provider/Layers/ProviderService.test.ts` — includes a new test that a runtime event touches `lastSeenAt` and that a burst is throttled (verified it fails without the fix).
- `vp test run apps/server/src/provider/Layers/ProviderSessionDirectory.test.ts` — `touchLastSeen` bumps a live binding; leaves a `stopped` binding untouched.
- All affected provider test files pass (83 tests).
- End-to-end: with a live server and a running background workflow, `last_seen_at` kept advancing from `task.progress` activity and the reaper did not reap the session past its idle threshold.
- Typecheck: 0 errors. Lint: clean.

## UI Changes

None.

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] No UI changes

---

> [!NOTE]
> **Medium Risk**
[...]

PR #4994, opened 2026-07-30
fix(server): idle reaper no longer kills in-flight background agent work

### LINKED ISSUE #4198: [Bug]: Idle session reaper kills in-flight background agent work (dynamic workflows / subagents)
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Start a long-running background agent task in a thread — a dynamic workflow or a task that fans out subagents that keep working after the foreground turn settles.
2. Let the foreground turn complete so the thread has no active user turn, while the background work keeps running.
3. Leave the thread otherwise idle (no new `sendTurn`) for longer than the reaper's inactivity threshold (default 30 min; swept every 5 min).
4. Observe that `ProviderSessionReaper` calls `stopSession` for the thread and tears down the provider session while the background agents are still in flight.

### Expected behavior

A session with live background agent work should not be reaped for inactivity. The reaper already skips sessions with an active foreground turn; the same protection should extend to in-flight background work.

### Actual behavior

`ProviderSessionReaper` (`apps/server/src/provider/Layers/ProviderSessionReaper.ts`) stops any provider session after 30 min of `last_seen_at` inactivity. Its only busy-guard is the projection snapshot's `activeTurnId`. When the foreground turn completes, the adapter session goes `ready` with no active turn, and nothing refreshes `last_seen_at` from ongoing background activity. So a long-running background workflow with no foreground turn is torn down mid-run. The next interaction resumes the conversation from a transcript cursor as a fresh process, but the in-flight background orchestration (subagent work, queued steps) is gone, leaving the run half-completed.

### Impact

### Version or commit

`main` at `5d34f9ff2`

### Environment

T3 Code server, running background dynamic workflows / subagents.

### Logs or stack traces

No exception is thrown. The reaper logs `provider.session.reaped` with `reason: "inactivity_threshold"` for a thread whose adapter session is still running background work.

A deterministic regression test is included in the linked fix.

### Workaround

Send a message in the thread periodically (every <30 min) to refresh `last_seen_at` and keep the session alive while the background work runs.

### Suggested fix

Refresh `last_seen_at` from runtime activity: bump the binding's timestamp when the adapter emits runtime events (e.g. `task.progress` from background subagents), throttled per thread so an event burst is at most one lightweight write per window. The adapter keeps emitting runtime events during background work even with no active foreground turn, so this keeps the session warm exactly when the reaper would otherwise reap it.

Zeus-Deus:
Closes #4198. Builds on #4199 by @ChamaruAmasara (credited as co-author) — this incorporates its `last_seen_at` refresh together with its review feedback, and adds the hard busy guard and orphan-settling discussed in the issue comments.

## What Changed

**Liveness refresh from runtime activity** (from #4199, plus review fixes)
- `touchLastSeen` through the runtime repository and session directory: a targeted `UPDATE` of `last_seen_at` that skips `stopped` rows (a reaped session is never resurrected) and reports whether a live row was actually updated.
- `ProviderService.processRuntimeEvent` refreshes `last_seen_at` on runtime events, throttled per thread (default 60s). The throttle arms only after a confirmed live-row write, so a failed or no-op touch retries on the next event instead of leaving the timestamp stale. The touch is best-effort (`Effect.catch`, so fiber interrupts propagate) and bounded by a timeout so a stalled write cannot block the per-instance event pump. Because every runtime event funnels through here, ordinary and synthetic follow-up turns now refresh liveness too — reporters observed `last_seen_at` frozen at initial startup despite later foreground turns.
- The throttle map is evicted on `stopSession`/`runStopAll` (the unresolved memory-leak review on #4199), and entries are only created for threads with a live binding row.

**Hard busy guard: outstanding background tasks**
- `ProviderService` tracks started-but-not-terminal background tasks in memory from `task.started`/`task.progress`/`task.completed` events (`task.progress` also registers unknown tasks, covering subscriptions that attach mid-task). In-memory is deliberate: in-flight background work never survives a server restart, so a persisted entry could only pin a session that is already gone.
- New `hasOutstandingBackgroundTasks(threadId)`; the reaper checks it immediately before `stopSession` and skips busy threads (`provider.session.reaper.skipped-outstanding-tasks`). The existing `activeTurnId` guard is preserved unchanged. A quiet-but-outstanding task blocks reaping past the inactivity threshold.
- Staleness backstop: after an extended period with no task events at all (default 90 min, 3× the reap threshold), tracked tasks are presumed dead, settled, and logged (`provider.session.tasks.presumed-abandoned`) — a stranded entry (e.g. an instance rebuilt by a settings edit, which emits no `session.exited`) cannot exempt a thread from reaping forever.

**Never silently orphan a task**
- On `stopSession`, `runStopAll`, and `session.exited`, whatever is still outstanding is settled by publishing a synthetic terminal `task.completed(status: "stopped")` with an explanatory summary, so downstream ingestion records a terminal activity instead of leaving a spinner alive forever. `session.exited` settles only tasks recorded against the same provider instance (fail-closed), so a stale exit cannot clobber a newer session's tasks.

## Why

The reaper stops any provider session after 30 min of `last_seen_at` inactivity, and its only busy signal was the projection's `activeTurnId`. Background work keeps running after the foreground turn settles (`activeTurnId` → null), and nothing refreshed `last_seen_at` from runtime activity, so long-running background workflows were torn down mid-run — killing the provider child process and everything it owned — and the loss only surfaced on the next user message. Multiple independent reports on the issue confirmed both halves: stale timestamps despite later turns, and reaps landing ~60–90s after background tasks started.

Two complementary mechanisms are needed: liveness refresh handles the common case cheaply, and the busy guard protects tasks that are temporarily quiet; settling makes any remaining race visible instead of silent.

## Validation

- `vp test run` on all 8 touched test files: **209 tests pass** (ProviderService 37, ProviderSessionReaper 7, ProviderSessionDirectory 8, plus adapter/orchestration doubles).
- Regression tests match the criteria from the issue discussion: a stale binding with no active foreground turn but an outstanding task is not stopped by a sweep, and becomes reapable once the task reaches a terminal state (verified to fail with the guard removed); touch bumps a live row and never resurrects a stopped one; a failed touch does not arm the throttle; two-thread isolation; fail-closed instance matching on `session.exited`; the staleness backstop settles exactly once.
- Typecheck (`tsgo --noEmit`): 0 errors. Lint: clean. `vp fmt` applied.

## UI Changes

None directly; threads whose session is stopped while background tasks are outstanding now show those tasks as stopped ("Session stopped before the task finished") instead of a spinner that never resolves.

## Checklist

- [x] This PR is focused on one concern
- [x] No UI changes

---

> [!NOTE]
> **Medium Risk**
> Changes session lifecycle, reaper behavior, and runtime event ordering in `ProviderService`; mistakes could leave sessions pinned or reap active work, but the design is heavily tested and fails closed on instance matching.
> 
> **Overview**
> Fixes sessions being stopped mid-run when background workflows (dynamic tasks, subagents, shells) keep working after the foreground turn ends. The idle reaper only looked at `activeTurnId` and stale `last_seen_at`, so it could kill the provider child while work was still running.
> 
> **Liveness from runtime activity** — Adds `touchLastSeen` on the runtime repository and session directory (updates only live rows, never stopped ones). `ProviderService` bumps `last_seen_at` on runtime events with per-thread throttling (default 60s), arms the throttle only after a successful write, and bounds the DB call with a timeout so the event pump does not stall.
> 
> **Outstanding background tasks** — In-memory tracking from `task.started` / `task.progress` / `task.completed`, plus `hasOutstandingBackgroundTasks`. The session reaper skips stop when that returns true. After prolonged silence (default 90 min), tasks are presumed abandoned, settled, and the thread becomes reapable again.
> 
> **Synthetic task settlement** — On stop, exit, or `runStopAll`, still-running tasks get a published `task.completed` with `status: stopped` and a user-facing summary so the activity feed does not spin forever. `session.exited` settles only tasks owned by the exiting provider instance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e578131f84f2def55bdd5224488483df695b4713. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix idle reaper to skip sessions with in-flight background agent tasks
> - Adds `hasOutstandingBackgroundTasks(threadId)` to `ProviderServiceShape` and implements it in [`ProviderService.ts`](https://github.com/pingdotgg/t3code/pull/4994/files#diff-b9b9384b78856ff56cff5dd0ccb106054fb5d2d1b7500fcd30d8938ebfec5775), tracking in-memory background tasks per thread with staleness detection.
> - Updates [`ProviderSessionReaper.ts`](https://github.com/pingdotgg/t3code/pull/4994/files#diff-06512dd4ebe9bbf2d3c8222739f659850d9baa547215a0aae698850d1cc39196) to call `hasOutstandingBackgroundTasks` before stopping a stale session, skipping reap if tasks are still running.
> - Adds `touchLastSeen(threadId)` to `ProviderSessionDirectory` and the underlying `ProviderSessionRuntimeRepository`, updating `last_seen_at` only for non-stopped rows; `ProviderService` calls this on runtime activity, throttled per thread.
> - Emits synthetic `task.completed` events to settle outstanding tasks when a session stops, exits, or tasks go stale.
> - Behavioral Change: sessions with active background tasks are no longer reaped by the idle sweeper even when they appear inactive.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized e578131. 6 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/apps/server/src/orchestration/Layers/CheckpointReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/CheckpointReactor.test.ts
@@ -123,6 +123,7 @@ function createProviderServiceHarness(
         },
       }),
     rollbackConversation,
+    hasOutstandingBackgroundTasks: () => Effect.succeed(false),
     get streamEvents() {
       return Stream.fromPubSub(runtimeEventPubSub);
     },
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
@@ -338,6 +338,7 @@ describe("ProviderCommandReactor", () => {
         });
       },
       rollbackConversation: () => unsupported(),
+      hasOutstandingBackgroundTasks: () => Effect.succeed(false),
       get streamEvents() {
         return Stream.fromPubSub(runtimeEventPubSub);
       },
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
@@ -121,6 +121,7 @@ function createProviderServiceHarness() {
       });
     },
     rollbackConversation: () => unsupported(),
+    hasOutstandingBackgroundTasks: () => Effect.succeed(false),
     get streamEvents() {
       return Stream.fromPubSub(runtimeEventPubSub);
     },
--- a/apps/server/src/persistence/ProviderSessionRuntime.ts
+++ b/apps/server/src/persistence/ProviderSessionRuntime.ts
@@ -58,6 +58,12 @@ export type GetProviderSessionRuntimeInput = typeof GetProviderSessionRuntimeInp
 export const DeleteProviderSessionRuntimeInput = Schema.Struct({ threadId: ThreadId });
 export type DeleteProviderSessionRuntimeInput = typeof DeleteProviderSessionRuntimeInput.Type;
 
+export const TouchLastSeenInput = Schema.Struct({
+  threadId: ThreadId,
+  lastSeenAt: IsoDateTime,
+});
+export type TouchLastSeenInput = typeof TouchLastSeenInput.Type;
+
 /**
  * ProviderSessionRuntimeRepository - Service tag for provider runtime persistence.
  */
@@ -92,6 +98,21 @@ export class ProviderSessionRuntimeRepository extends Context.Service<
       ReadonlyArray<ProviderSessionRuntime>,
       ProviderSessionRuntimeRepositoryError
     >;
+
+    /**
+     * Bump only `last_seen_at` for an existing, non-stopped row.
+     *
+     * A targeted update used to keep a session's inactivity clock fresh from
+     * background runtime activity (e.g. a running dynamic workflow) without
+     * rewriting the full runtime payload. Rows in `stopped` status are left
+     * untouched so a reaped session is never resurrected.
+     *
+     * Returns whether a live row was actually updated, so callers can tell a
+     * real refresh from a no-op against an absent or stopped row.
+     */
+    readonly touchLastSeen: (
+      input: TouchLastSeenInput,
+    ) => Effect.Effect<boolean, ProviderSessionRuntimeRepositoryError>;
 
     /**
      * Delete provider runtime state by canonical thread id.
@@ -226,6 +247,22 @@ export const make = Effect.gen(function* () {
       `,
   });
 
+  // `RETURNING` reports which rows the update actually matched: an absent or
+  // already-stopped row yields none, and the caller needs that distinction to
+  // avoid treating a no-op as a successful refresh.
+  const touchLastSeenRow = SqlSchema.findAll({
+    Request: TouchLastSeenInput,
+    Result: Schema.Struct({ threadId: Schema.String }),
+    execute: ({ threadId, lastSeenAt }) =>
+      sql`
+        UPDATE provider_session_runtime
+        SET last_seen_at = ${lastSeenAt}
+        WHERE thread_id = ${threadId}
+          AND status != 'stopped'
+        RETURNING thread_id AS "threadId"
+      `,
+  });
+
   const deleteRuntimeByThreadId = SqlSchema.void({
     Request: DeleteRuntimeRequestSchema,
     execute: ({ threadId }) =>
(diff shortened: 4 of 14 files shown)


juliusmarminge:

PR #5000, opened 2026-07-30
fix(mobile): move the thread sync indicator into the header

carTloyal123:
## What changed

The "Syncing threads..." indicator flashed in and out above the thread list. There were two independent causes, and this fixes both.

### 1. It lived inside the scroll view

`WorkspaceConnectionStatus` was rendered as `ListHeaderComponent` — literally row 0 — in both `HomeScreen` and `ThreadNavigationSidebar`. Every sync-state flip mounted or unmounted a row, which flashed the indicator *and* reflowed every row beneath it.

It now renders as a header button instead. The Android floating overlay that duplicated it is gone too.

### 2. The signal itself strobes

Moving the UI alone would not have stopped the blinking. `hasSynchronizingShell` re-enters `"synchronizing"` on every websocket resubscribe and retries expected failures on a 250ms cycle, so even a healthy connection produces a stream of sub-second sync blips.

The raw state is now reduced to a tone (`offline` / `error` / `disconnected` / `connecting` / `syncing` / `idle`) and **settled** before it renders:

- Transient tones must persist **400ms** before appearing — longer than the retry cycle, so a reconnect that resolves on its own never reaches the header at all.
- Once shown, they stay **900ms**, so a sync that resolves immediately after crossing the threshold doesn't appear and vanish within a frame.
- Real faults (offline, connection error, no ready environment) **bypass both delays** — delaying those would mean hiding a problem worth seeing.

The settling decision is a pure function (`planWorkspaceSyncToneSettlement`) so the timing behaviour is unit-tested rather than implicit in a hook.

## Before / after

Captured on iOS Simulator (iPhone 17 Pro, iOS 26.5) against a local server with six seeded threads. Each pair is the same screen: connected and idle, then with the environment disconnected.

| | Idle (connected) | Disconnected / syncing |
|---|---|---|
| **Before** |  |  |
| **After** |  |  |

**Before:** the status becomes row 0 of the list. "Tidy up mobile header spacing" is pushed from y≈307 to y≈422 — every row below it shifts down, then snaps back when the state clears.

**After:** the status is an icon in the header capsule next to `•••`. The list does not move at all — the first row stays at y≈307 in both states.

Screen recordings of the same disconnect/reconnect cycle:

- [before-flash.mp4](https://raw.githubusercontent.com/carTloyal123/t3code/media/sync-indicator/media/before-flash.mp4) — row appearing and disappearing, list reflowing
- [after-stable.mp4](https://raw.githubusercontent.com/carTloyal123/t3code/media/sync-indicator/media/after-stable.mp4) — header icon changes, list stays put

(Media lives on a [separate branch](https://github.com/carTloyal123/t3code/tree/media/sync-indicator) so it stays out of this diff.)

## Details worth noting

- **The button keeps its slot when idle** — it renders an empty box of the same size — so neighbouring header controls never shift as sync state changes. Trading a little dead space for zero layout movement seemed right for the bug being fixed.
- Wired into all three headers: the Android home header, the iOS native header items, and the split-view sidebar.
- iOS native bar-button items can't host a spinner, so they fall back to an SF Symbol; the React headers show an `ActivityIndicator`.
- `optionsVersion` had to be extended — native header factories are stabilized, so a tone captured inside one can't change the options signature on its own and the icon would never update.
- Tapping still opens environment settings, unchanged.
- **The full-page empty state deliberately keeps its inline pill.** With no threads on screen it's the only thing explaining why the page is blank, and there's nothing below it to reflow.

## Verification

- `tsc --noEmit` clean
- `vp lint` clean on changed files
- Full mobile suite green (587 tests, 13 new covering tone derivation and settling)
- Manually verified on iOS Simulator as above

Branched off current `main`. Split from #4999 so the two unrelated fixes stay reviewable separately.

### DIFF
--- a/apps/mobile/src/features/home/HomeHeader.tsx
+++ b/apps/mobile/src/features/home/HomeHeader.tsx
@@ -25,6 +25,13 @@ import {
   PROJECT_SORT_OPTIONS,
   THREAD_SORT_OPTIONS,
 } from "./home-list-options";
+import { useSettledWorkspaceSyncTone } from "./use-settled-workspace-sync-tone";
+import {
+  WorkspaceSyncStatusButton,
+  workspaceSyncToneSymbol,
+} from "./WorkspaceSyncStatusButton";
+import { workspaceConnectionStatusLabel } from "./workspace-connection-status";
+import type { WorkspaceState } from "../../state/workspaceModel";
 
 export type HomeHeaderEnvironment = HomeListFilterMenuEnvironment;
 
@@ -43,6 +50,9 @@ export function HomeHeader(props: {
   readonly onThreadSortOrderChange: (sortOrder: SidebarThreadSortOrder) => void;
   readonly onOpenSettings: () => void;
   readonly onStartNewTask: () => void;
+  /** Workspace connection/sync state — surfaced as the header sync button. */
+  readonly catalogState: WorkspaceState;
+  readonly onOpenEnvironments: () => void;
 }) {
   if (Platform.OS === "android") {
     return <AndroidHomeHeader {...props} />;
@@ -212,6 +222,14 @@ function AndroidHomeHeader(props: HomeHeaderProps) {
               </View>
             </View>
 
+            {/* Sync/connection state — moved out of the thread list's scroll
+                view, where mounting and unmounting a list header row made it
+                flash and reflowed every row below it. */}
+            <WorkspaceSyncStatusButton
+              state={props.catalogState}
+              onPress={props.onOpenEnvironments}
+            />
+
             <ControlPillMenu
               actions={menuActions}
               isAnchoredToRight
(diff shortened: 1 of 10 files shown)


carTloyal123:
Superseded by #5023, which fixes the same bug with a better approach.

This PR moved the status into a header icon button. That removed the flash, but traded away the full status text — you lost "Reconnecting to <environment>" in favour of an ambiguous glyph, and a static sync icon does not read as "data is moving".

#5023 keeps the original status bar and its wording, and instead pins it above the list so it can no longer mount/unmount and reflow the rows. At rest it reports `Synced N environments at <time>` rather than disappearing, so the bar is a standing answer instead of something you only notice mid-sync.

Closing rather than force-pushing, since it is a different approach rather than an iteration on this one.

PR #5012, opened 2026-07-30
fix(server): pass the explicit user to loginctl enable-linger (fixes WSL2 service install)

### LINKED ISSUE #4894: [Bug]: `t3 service install` fails on WSL2 — `loginctl enable-linger` is called with no username and cannot resolve a session
### Before submitting

### Area

apps/server

### Steps to reproduce

On Ubuntu under WSL2:

1. `npm install -g t3@latest`
2. `t3 service install`

Install fails at the final activation step and rolls back.

The rollback itself is clean — no dangling unit is left in `~/.config/systemd/user/`, and `t3 service status` correctly reports `not installed`. That part works exactly as the code comments describe.

### Expected behavior

The systemd user unit is installed, enabled and started, as it is on non-WSL Linux.

Lingering is not actually the obstacle here: on this machine `Linger=yes` was already set for the user (`loginctl show-user $(id -un)` → `Linger=yes`), so the step is a no-op that fails purely on argument resolution.

### Actual behavior

```
BootServiceCommandError: Background setup failed while enabling lingering for this user (exit code 1).
```

**Root cause.** `install` runs `loginctl enable-linger` with **no username argument**:

https://github.com/pingdotgg/t3code/blob/main/apps/server/src/cloud/bootService.ts#L346

```ts
yield* runStep("enabling lingering for this user", "loginctl", ["enable-linger"]);
```

The comment above it explains the reasoning:

> No username argument: loginctl defaults to the calling user, which is always right, while `$USER` can be stale (su without -l) or unset.

That assumption does not hold under WSL2. `XDG_SESSION_ID` is unset in a normal WSL2 shell, so logind cannot resolve a session for the caller and the no-arg form fails, while the explicit form succeeds:

```
$ loginctl enable-linger $(id -un)
$ echo $?
0

$ loginctl enable-linger
Could not enable linger: No such device or address
$ echo $?
1
```

Note this is not a "systemd is missing" case — `systemctl --user is-system-running` returns `running`, and `daemon-reload` / `enable` / `start` all succeed. Only the no-arg `enable-linger` fails.

### Suggested fix

Pass the resolved username explicitly rather than relying on the caller-default:

```ts
yield* runStep("enabling lingering for this user", "loginctl", ["enable-linger", os.userInfo().username]);
```

`os.userInfo().username` comes from the passwd database for the effective uid, so it avoids the stale/unset `$USER` problem the current comment is guarding against while also working when there is no logind session.

Alternatively, treat a failing `enable-linger` as non-fatal when lingering is already enabled (check `loginctl show-user <user>` for `Linger=yes` first), since in that case the step has nothing to do.

### Impact

(`t3 service install` and `t3 service update` are unusable on WSL2 without a workaround. Everything else works well — see Workaround.)

### Version or commit

t3@0.0.31

### Environment

Ubuntu on WSL2 (kernel 6.18.33.2-microsoft-standard-WSL2), Windows host, node v22.22.1, npm 10.9.4, systemd user manager running, `Linger=yes` already set.

### Logs or stack traces

```shell
$ t3 service install
[20:27:02.062] ERROR (#5): BootServiceCommandError: Background setup failed while enabling lingering for this user (exit code 1).
    at .../t3/dist/bin.mjs:17006:71
    at cloud.boot_service.run_step (.../t3/dist/bin.mjs:17062:11)
    at cloud.boot_service.install (.../t3/dist/bin.mjs:17065:17)
    at cli.service.reconcile (.../t3/dist/bin.mjs:19594:24)
    at cli.service.run (.../t3/dist/bin.mjs:19593:196)

$ t3 service status
T3 Code service
  Status: not installed
  Next: Run `t3 service install`.
```

### Workaround

Shim only the one call, for the duration of the install:

```bash
mkdir -p /tmp/t3-linger-shim
REAL=$(command -v loginctl)
cat > /tmp/t3-linger-shim/loginctl <<EOF
#!/usr/bin/env bash
if [ "\$1" = "enable-linger" ] && [ \$# -eq 1 ]; then
  exec "$REAL" enable-linger "\$(id -un)"
fi
exec "$REAL" "\$@"
EOF
chmod +x /tmp/t3-linger-shim/loginctl

PATH="/tmp/t3-linger-shim:$PATH" t3 service install
```

With that, the install completes normally and the service is healthy on WSL2:

```
$ t3 service status
T3 Code service
  Status: installed · t3@0.0.31
  Unit: /home/<user>/.config/systemd/user/t3code.service

$ systemctl --user is-active t3code
active
$ ss -ltnH 'sport = :3773'
LISTEN 0 511 127.0.0.1:3773 0.0.0.0:*
```

Cold start was ~3s with the data dir on ext4, and Claude Code / Codex / Grok providers were all detected correctly afterwards.

Happy to open a PR with the one-line change if that is welcome — I saw the note in the README that small fixes may be considered.

arhxam:
## What Changed

`t3 service install` now runs `loginctl enable-linger <user>` with the user resolved from the effective uid (`os.userInfo()`, exposed as a new `HostProcessUserName` reference), instead of the bare `loginctl enable-linger`.

Files: `apps/server/src/cloud/bootService.ts`, `packages/shared/src/hostProcess.ts`, plus tests.

## Why

Closes #4894.

The bare `loginctl enable-linger` resolves the caller through their **logind session**. Under WSL2 a normal shell has no `XDG_SESSION_ID`, so logind cannot resolve a session and the step exits non-zero — install then rolls back and the service is never set up, even when lingering is already enabled for that user.

The existing comment justified the no-arg form by wanting to avoid a stale-or-unset `$USER`. Resolving the username from the effective uid keeps that intent (it is never `$USER`) while dropping the session dependency, so `loginctl enable-linger <user>` succeeds on WSL2 **and** on ordinary systemd Linux. This is the standard way to enable-linger for a specific user and does not depend on an active session.

Verified: `bootService.test.ts` passes, including a new test asserting the resolved user is forwarded verbatim; `apps/server` typecheck is clean.

## Checklist

- [x] No UI changes

---

> [!NOTE]
> **Low Risk**
> Narrow change to Linux boot-service install linger invocation and a shared host-process username ref; no auth, data, or API surface changes.
> 
> **Overview**
> **Fixes `t3 service install` failing on WSL2** when the linger step runs `loginctl enable-linger` with no arguments. That form needs a logind session; WSL2 shells often lack `XDG_SESSION_ID`, so the step fails and install rolls back.
> 
> Install now runs **`loginctl enable-linger <user>`**, with the username from a new **`HostProcessUserName`** reference in `@t3tools/shared/hostProcess` (default `os.userInfo().username` from the effective uid, not `$USER`). **`bootService`** reads that ref and passes it into the linger command.
> 
> Tests expect the explicit user in the install command list and add a case that the resolved name is forwarded verbatim (e.g. `caesar`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3e132e70799b187089ba5b707fbc95a3566d048. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `loginctl enable-linger` to pass an explicit username for WSL2 service install
> - Adds `HostProcessUserName` to [hostProcess.ts](https://github.com/pingdotgg/t3code/pull/5012/files#diff-bfd8aa15bef02a4e3e39ffafc1b2868e593c03d10852d3d3490c5456149a099b), a `Context.Reference<string>` that resolves the current user's account name from the system passwd database (via `os.userInfo().username`) rather than the `$USER` env var.
> - Updates `bootService.ts` to pass the resolved username explicitly to `loginctl enable-linger <user>`, fixing a failure in WSL2 where the no-argument form fails without an active logind session.
> - Behavioral Change: `loginctl enable-linger` is now always called with an explicit username argument instead of relying on the session-derived default.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f3e132e.</sup>
> 
>

### DIFF
--- a/apps/server/src/cloud/bootService.test.ts
+++ b/apps/server/src/cloud/bootService.test.ts
@@ -12,6 +12,7 @@ import {
   HostProcessArguments,
   HostProcessExecutablePath,
   HostProcessPlatform,
+  HostProcessUserName,
 } from "@t3tools/shared/hostProcess";
 
 import { reconcileService } from "../cli/service.ts";
@@ -60,10 +61,11 @@ const makeHost = (entry: string): BootService.BootServiceHost => ({
   cliEntryPath: entry,
 });
 
-const provideHostRefs = (home: string, platform: NodeJS.Platform = "linux") =>
+const provideHostRefs = (home: string, platform: NodeJS.Platform = "linux", userName = "t3user") =>
   Effect.provide(
     Layer.mergeAll(
       Layer.succeed(HostProcessPlatform, platform),
+      Layer.succeed(HostProcessUserName, userName),
       ConfigProvider.layer(ConfigProvider.fromEnv({ env: { HOME: home } })),
     ),
   );
@@ -222,7 +224,10 @@ it.layer(NodeServices.layer)("BootService", (it) => {
           // restart (not enable --now) so repairing a stale unit replaces a
           // running process instead of leaving the old one until reboot.
           "systemctl --user restart t3code.service",
-          "loginctl enable-linger",
+          // The user is passed explicitly (resolved from the effective uid),
+          // because the bare form resolves the caller through a logind session
+          // that does not exist under WSL2.
+          "loginctl enable-linger t3user",
         ],
       );
 
@@ -243,6 +248,32 @@ it.layer(NodeServices.layer)("BootService", (it) => {
       assert.isFalse(statusAfter.installed);
       const removedAgain = yield* service.uninstall;
       assert.isFalse(removedAgain);
+    }),
+  );
+
+  it.effect("enables linger for the resolved user so it works without a logind session", () =>
+    Effect.gen(function* () {
+      const { dirs } = yield* makeTestContext();
+      const commands: Array<RecordedCommand> = [];
+      const service = yield* BootService.make({
+        baseDir: dirs.baseDir,
+        logsDir: dirs.logsDir,
+        cliVersion: "0.0.27",
+        host: makeHost(dirs.stableEntry),
+      }).pipe(
+        Effect.provide(makeRecordingRunnerLayer(commands)),
+        provideHostRefs(dirs.home, "linux", "caesar"),
+      );
+
+      yield* service.install;
+
+      // The username must be forwarded verbatim — the bare `loginctl
+      // enable-linger` needs a logind session (absent under WSL2), so the
+      // explicit form is what makes install succeed there.
+      assert.deepEqual(
+        commands.filter(({ command }) => command === "loginctl"),
+        [{ command: "loginctl", args: ["enable-linger", "caesar"] }],
+      );
     }),
   );
 
--- a/apps/server/src/cloud/bootService.ts
+++ b/apps/server/src/cloud/bootService.ts
@@ -13,6 +13,7 @@ import {
   HostProcessArguments,
   HostProcessExecutablePath,
   HostProcessPlatform,
+  HostProcessUserName,
 } from "@t3tools/shared/hostProcess";
 
 import * as ProcessRunner from "../processRunner.ts";
@@ -197,6 +198,7 @@ export const make = Effect.fn("cloud.boot_service.make")(function* (input: {
     cliEntryPath: hostArguments[1] ?? "",
   };
   const platform = yield* HostProcessPlatform;
+  const userName = yield* HostProcessUserName;
   const homeDir = yield* Config.string("HOME").pipe(Config.withDefault(""));
   const fs = yield* FileSystem.FileSystem;
   const path = yield* Path.Path;
@@ -340,10 +342,13 @@ export const make = Effect.fn("cloud.boot_service.make")(function* (input: {
         BOOT_SERVICE_UNIT_FILE,
       ]);
       // Linger keeps the user manager (and this service) running without an
-      // open session — the whole point on a box reached over SSH. No
-      // username argument: loginctl defaults to the calling user, which is
-      // always right, while $USER can be stale (su without -l) or unset.
-      yield* runStep("enabling lingering for this user", "loginctl", ["enable-linger"]);
+      // open session — the whole point on a box reached over SSH. Pass the
+      // user explicitly (resolved from the effective uid, so it is never the
+      // stale-or-unset $USER): the bare `loginctl enable-linger` resolves the
+      // caller through their logind *session*, which does not exist under
+      // WSL2 (no XDG_SESSION_ID), so the no-arg form fails there while the
+      // explicit user works everywhere.
+      yield* runStep("enabling lingering for this user", "loginctl", ["enable-linger", userName]);
     }).pipe(Effect.tapError(() => rollbackFailedInstall(previousUnit)));
 
     return plan;
--- a/packages/shared/src/hostProcess.ts
+++ b/packages/shared/src/hostProcess.ts
@@ -51,4 +51,16 @@ export const HostProcessArguments = Context.Reference<ReadonlyArray<string>>(
   },
 );
 
+/**
+ * The calling user's account name, resolved from the effective uid rather than
+ * `$USER`. `$USER` can be stale (`su` without `-l`) or unset, while
+ * `os.userInfo()` reads the passwd database by uid and is always the real user.
+ */
+export const HostProcessUserName = Context.Reference<string>(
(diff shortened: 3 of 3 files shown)


martinfrancois:
Confirming this on a second platform, and I think the polkit detail is worth recording since it
is the reason the fix works.

openSUSE MicroOS 20260802, `t3@0.0.31`, headless (no display, no login session). `t3 service
install` fails at `enable-linger` with `Access denied ... requires interactive authentication`.

logind picks its polkit action from the target uid: `set-self-linger` when it is the caller's
own, `set-user-linger` otherwise. Those are governed differently, and only the first is usable
with no authentication agent to prompt anyone:

```
$ pkcheck --action-id org.freedesktop.login1.set-self-linger --process $$; echo $?
0

$ pkcheck --action-id org.freedesktop.login1.set-user-linger --process $$; echo $?
polkit.result=auth_admin_keep
2

$ loginctl enable-linger "$(id -un)"; echo $?   # named user, my own
0

$ loginctl enable-linger; echo $?            # bare, as t3 calls it today
Could not enable linger: Access denied as the requested operation requires interactive
authentication. However, interactive authentication has not been enabled by the calling program.
1
```

Same host, same user, same moment. So naming the user is not only about resolving a session, it
selects the action that needs no password, which is why this fixes headless hosts as well as
WSL2.

Two limits on that: I did not build and run #5012 itself, only the call it makes, and lingering
was already on for this user so the successful call was a no-op write. The polkit check runs
before the write, so I would not expect a first-time write to differ.

**Workaround until this lands**, for anyone else who finds this issue. First confirm lingering
is already on, because otherwise you would be skipping a step that genuinely needs to run and
the service will not start at boot:

```sh
loginctl show-user "$(id -un)" --property=Linger    # must print Linger=yes
```

If it does, shim out just that one call for the duration of the install:

```sh
mkdir -p /tmp/linger-shim
printf '#!/bin/sh\n[ "$1" = enable-linger ] && exit 0\nexec /usr/bin/loginctl "$@"\n' \
  > /tmp/linger-shim/loginctl
chmod +x /tmp/linger-shim/loginctl

PATH=/tmp/linger-shim:$PATH t3 service install

rm -rf /tmp/linger-shim
```

The unit you end up with is identical to a normal install.

**AI Disclosure:** Prepared with the assistance of Claude Code (Claude Opus 5). Every command
and output above was run and observed on my own machine. Reviewed by me before posting.

CDVolvik:
Ran this on Linux (Docker `node:24`) rather than a Windows host, since `bootService.test.ts` is Linux-only by design.

- `main`: 7/7 pass
- this branch: 17/17 pass

So no regression, and the 10 added tests all pass.

Two limits worth stating rather than implying a stronger result:

I could not run the new tests against `main`'s source to show they fail without the fix. They depend on `HostProcessUserName`, which this PR adds to `packages/shared/src/hostProcess.ts`, so reverting the source makes the test file fail to load instead of failing an assertion.

Docker Linux is not WSL2, so the actual `loginctl enable-linger` behavior under WSL2 is still untested from here.

Separately, in case anyone tries to check this from Windows and thinks the PR is at fault: `bootService.test.ts` already fails 4/7 on a Windows host on `main`, because it asserts a systemd `ExecStart` built from the host path (`ExecStart=/usr/local/bin/node C:\Users\...`). Same non-hermetic class as #4507.

CDVolvik (commented):
Re-checked on today's main (`6ae9662d8`). `HostProcessUserName` is still absent there, so this is not already shipped.

The WSL2 claim is right: `loginctl enable-linger` with no user goes through the caller's logind session (`XDG_SESSION_ID`). WSL2 usually has none. Passing the uid-resolved name (`os.userInfo().username`, not `$USER`) is the form that works without a session.

Applied the three PR files onto current main and ran `bootService.test.ts`: 15 passed, 2 failed. The new case (`enable-linger caesar`) is in the 15. The two failures are `pinnedRuntime.ts:176` (`validate(stagingPaths)` not iterable) on the npx-pin tests — that is this PR sitting on a stale `bootService` against a newer `pinnedRuntime`, i.e. the CONFLICTING base, not the linger args.

No live `loginctl` on this Windows host. I am not claiming an end-to-end WSL2 service install.

juliusmarminge:

**Reason:** Equivalent linger fix is on main: enable-linger already passes the uid.
**Related:** main bootService.ts enable-linger uid

PR #5013, opened 2026-07-30
fix(server): put ~/.local/bin on the boot service PATH so providers resolve

### LINKED ISSUE #4913: [Bug]: systemd user unit from `t3 service install` sets no PATH, so providers in ~/.local/bin are invisible to the service
### Before submitting

Not a duplicate, but same family as #4697 ("No installed editors found") and #4846 (Windows desktop resolver ignores User PATH). Those two plus this one all reduce to: **the server resolves external commands against whatever PATH it inherited, and that PATH is wrong whenever it wasn't launched from an interactive login shell.**

### Area

apps/server

### Steps to reproduce

On a headless Ubuntu box, with providers installed the ordinary way so their launchers land in `~/.local/bin`:

1. `command -v claude codex` → `/home/caesar/.local/bin/claude`, `/home/caesar/.local/bin/codex`
2. `npm i -g t3` (npm prefix is `~/.local/npm-global`, no sudo needed)
3. `t3 service install`
4. `systemctl --user start t3code.service`
5. Open the web UI and try to start a thread.

### Expected behavior

The service resolves `claude` / `codex` exactly as `t3 serve` does when run by hand from a login shell.

### Actual behavior

Providers are not found. The server itself is healthy — it listens, pairs, authenticates, and the web client works — but there are no usable providers, so no thread can run.

The cause is that the generated unit sets no `PATH`, so the service inherits the systemd user manager's minimal environment:

- **Service PATH** (what `t3` gets): `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin`
- **Login shell PATH** (what a human gets): `/home/caesar/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin`

`~/.local/bin` is prepended by Ubuntu's default `~/.profile`, which only runs for **login** shells. A systemd user unit never sources it. So the one directory where user-scoped CLI tools normally live is exactly the one the service cannot see.

This also makes the failure confusing to diagnose: running `t3 serve` manually over SSH works if you have a login shell, so the provider list looks fine when you test by hand and empty when the service runs.

### Impact

The headless Linux path is the "self-hosted server" use case, and on a fresh install it silently produces a server with zero providers.

### Version or commit

`t3@0.0.31`

### Environment

Ubuntu 26.04 LTS, kernel 7.0.0-28-generic, x86-64 · node `v22.22.1` at `/usr/bin/node` · npm prefix `/home/caesar/.local/npm-global` · systemd **user** service with `Linger=yes` · no desktop session, headless, reached over Tailscale.

### Logs or stack traces

```shell
# unit generated by `t3 service install` (verbatim) - note the absence of any PATH
[Unit]
Description=T3 Code server
StartLimitIntervalSec=300
StartLimitBurst=5

[Service]
Type=simple
WorkingDirectory=%h
Environment=T3CODE_HOME=/home/caesar/.t3
Environment=T3_BOOT_SERVICE_UNIT=t3code.service
ExecStart=/usr/bin/node /home/caesar/.local/npm-global/bin/t3 serve
Restart=always
RestartSec=5
StandardOutput=append:/home/caesar/.t3/userdata/logs/boot-service.log
StandardError=append:/home/caesar/.t3/userdata/logs/boot-service.log

[Install]
WantedBy=default.target

# what the running service actually has, before the workaround
$ tr '\0' '\n' < /proc/$(systemctl --user show t3code.service -p MainPID --value)/environ | grep ^PATH=
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin

# where the providers actually are
$ bash -lc 'command -v claude codex'
/home/caesar/.local/bin/claude
/home/caesar/.local/bin/codex
```

### Workaround

A systemd drop-in at `~/.config/systemd/user/t3code.service.d/override.conf`, which also survives `t3 service update` (editing the generated unit does not):

```ini
[Service]
Environment=PATH=/home/caesar/.local/bin:/home/caesar/.local/npm-global/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

After `systemctl --user daemon-reload && systemctl --user restart t3code.service`, providers resolve and threads run normally. Note that `systemctl --user enable --now` does **not** restart an already-active unit, so a plain `enable --now` after editing the drop-in silently keeps the old environment.

### Suggested fix

Have `t3 service install` emit an explicit `Environment=PATH=` in the unit, captured from the installing user's login PATH (`bash -lc 'echo $PATH'`) rather than from the environment the installer happens to run in. At minimum, prepend `~/.local/bin` and the npm global bin directory.

Doing it at install time also fixes #4697-style reports on Linux, and is the same underlying defect as #4846 on Windows.

### Related papercut (secondary, happy to split out)

`t3 service install` exposes no `--host` / `--port`, and the unit it writes binds to loopback. On a headless box that means the freshly installed service is unreachable from anywhere until you discover that `T3CODE_HOST` / `T3CODE_PORT` work as environment variables. I verified both do, via the same drop-in:

```ini
Environment=T3CODE_HOST=100.x.x.x
Environment=T3CODE_PORT=3773
```

Either exposing flags on `service install` or documenting those variables in the headless setup path would save the guesswork.

arhxam:
## What Changed

The generated systemd user unit (`renderBootServiceUnit`) now sets an explicit `Environment=PATH=` that prepends `~/.local/bin` and `~/bin` (via the `%h` specifier) to the standard system directories.

`apps/server/src/cloud/bootService.ts`, plus the byte-for-byte render test.

## Why

Closes #4913.

The unit set no `PATH`, so `t3code.service` inherited the systemd user manager's minimal environment, which does not include `~/.local/bin` — exactly where user-scoped provider CLIs (`claude`, `codex`) land on a headless install. Ubuntu's `~/.profile` prepends `~/.local/bin`, but that only runs for **login** shells and a systemd user unit never sources it. The result: a server that listens, pairs and authenticates but exposes **zero providers**, so no thread can run. It is also confusing to diagnose, because `t3 serve` run by hand over an SSH login shell finds every provider while the service cannot.

`%h` expands to the home dir (the same specifier the unit already uses for `WorkingDirectory`), and directories absent on a given distro are simply skipped during lookup. `renderBootServiceUnit` stays pure and deterministic, so `status` still flags an older unit as out-of-date and offers a repair that rewrites it with the PATH.

Verified: `bootService.test.ts` passes (render test updated byte-for-byte); `apps/server` typecheck is clean.

## Checklist

- [x] No UI changes

---

> [!NOTE]
> **Low Risk**
> Scoped to generated systemd unit environment and test expectations; no auth, data, or runtime logic changes beyond PATH for CLI discovery.
> 
> **Overview**
> The generated **systemd user unit** for `t3code.service` now sets an explicit `Environment=PATH=` that prepends `%h/.local/bin` and `%h/bin` ahead of the usual system directories, matching what login shells get from `~/.profile` on typical Ubuntu headless installs.
> 
> **Why it matters:** without this, `t3 serve` under the user manager inherited a minimal PATH and could not find user-scoped provider CLIs (e.g. `claude`, `codex`) in `~/.local/bin`, so the service listened and paired but exposed **zero providers**—while the same command over SSH looked fine.
> 
> `renderBootServiceUnit` stays deterministic, so **status** still compares the installed unit byte-for-byte; units written before this change are reported **out of date** and repair/install rewrites them with the new PATH. Tests assert the full rendered unit and PATH prefix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bba10621ae5f5dd2131fd7314e3ffd8f92830196. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add `~/.local/bin` to PATH in generated systemd boot service unit
> The `BootService.renderBootServiceUnit` function now injects an `Environment=PATH=` line into the `[Service]` section, prepending `%h/.local/bin` and `%h/bin` before standard system directories. This ensures user-installed binaries (e.g. provider CLIs) are resolvable when the boot service starts. Behavioral Change: the boot service process now runs with an explicit PATH instead of inheriting the default systemd user environment.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized bba1062. 1 file reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/apps/server/src/cloud/bootService.test.ts
+++ b/apps/server/src/cloud/bootService.test.ts
@@ -108,6 +108,7 @@ it("renders a systemd unit with absolute paths and append-mode logging", () => {
       "[Service]",
       "Type=simple",
       "WorkingDirectory=%h",
+      "Environment=PATH=%h/.local/bin:%h/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin",
       "Environment=T3CODE_HOME=/home/theo/.t3",
       "Environment=T3_BOOT_SERVICE_UNIT=t3code.service",
       "ExecStart=/usr/local/bin/node /home/theo/.t3/runtime/versions/0.0.27/node_modules/t3/dist/bin.mjs serve",
@@ -137,6 +138,8 @@ it("quotes systemd values containing spaces and escapes percent specifiers", () 
   });
   assert.include(unit, 'ExecStart="/home/me/my tools/node" "/home/me/T3 Data/bin.mjs" serve');
   assert.include(unit, 'Environment=T3CODE_HOME="/home/me/T3 Data"');
+  // ~/.local/bin is prepended so user-scoped provider CLIs resolve.
+  assert.include(unit, "Environment=PATH=%h/.local/bin:%h/bin:");
   // append: paths take the rest of the line literally (spaces are fine,
   // quoting is not), but % still goes through specifier expansion.
   assert.include(unit, "StandardOutput=append:/home/me/100%%logs/boot.log");
--- a/apps/server/src/cloud/bootService.ts
+++ b/apps/server/src/cloud/bootService.ts
@@ -100,6 +100,15 @@ export function renderBootServiceUnit(plan: BootServicePlan): string {
     "[Service]",
     "Type=simple",
     "WorkingDirectory=%h",
+    // A systemd user unit inherits the user manager's minimal PATH, which does
+    // NOT include ~/.local/bin — the directory Ubuntu's ~/.profile prepends and
+    // where user-scoped provider CLIs (claude, codex) land on a headless
+    // install. Without this, `t3 serve` finds every provider by hand from a
+    // login shell but the service starts with zero providers. Prepend the
+    // user-local bins to the standard system PATH; %h expands to the home dir
+    // (like WorkingDirectory), and absent entries (e.g. /snap/bin on non-Debian
+    // distros) are simply skipped during lookup.
+    "Environment=PATH=%h/.local/bin:%h/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin",
     `Environment=T3CODE_HOME=${quoteSystemdValue(plan.baseDir)}`,
     `Environment=${BOOT_SERVICE_UNIT_ENV}=${BOOT_SERVICE_UNIT_FILE}`,
     `ExecStart=${quoteSystemdValue(plan.nodePath)} ${quoteSystemdValue(plan.t3EntryPath)} serve`,


shivamhwp:

The fixed PATH replaces the systemd user manager's PATH, so provider directories supplied through environment.d or import-environment can disappear. For example, a tool installed only in `/opt/company/bin` is retained by the current login-shell PATH merge but is lost with this unit when no shell profile adds it back. Preserve inherited entries when adding user-local directories.

The server already calls `fixPath()` before startup and merges the login shell PATH with the inherited PATH. The missing directory in `/proc/<pid>/environ` alone does not establish the provider lookup failure. Please pin down the failed hydration or lookup path before choosing the fallback, and rebase onto the current service launcher implementation.

PR #5022, opened 2026-07-30
feat(mobile): add a refresh-all button to the environments screen

carTloyal123:
## What changed

Adds an `arrow.clockwise` button to the Environments settings header, next to the existing add-environment button, that retries every registered environment at once.

## Why it earns its place

This is **not a placebo refresh**. `onReconnectEnvironment` bottoms out in `environmentCatalog.retryNow`, which short-circuits the reconnect backoff — so it is the one way to escape a long backoff delay without waiting it out. That is exactly the moment a user is staring at a stale list wanting to do *something* about it.

It also pairs with the workspace status indicator, which already navigates to this screen: the status reports that something is wrong, and this is the remedy sitting where the user lands.

## Details

- Disabled when no environments are registered, so it can never be a dead tap.
- Wired into both header styles — `NativeHeaderToolbar` on iOS, `AndroidScreenHeader` actions on Android.
- Retries *every* environment, including already-connected ones. Harmless (it just re-establishes) and matches the plain reading of "refresh all". Easy to narrow to only disconnected/backing-off environments if reviewers prefer.

## Verification

- `tsc --noEmit` clean
- Mobile suite green

No screenshots: this adds one standard header icon button in the existing toolbar and changes nothing else visually.

Kept separate from the status-bar fix so the two stay independently reviewable.

---

> [!NOTE]
> **Low Risk**
> UI-only addition reusing existing reconnect plumbing; no new auth or data paths, with harmless duplicate retries for already-connected environments.
> 
> **Overview**
> Adds a **refresh all** header action on the Environments settings screen so users can trigger reconnect for every registered environment in one tap.
> 
> `handleRefreshAll` loops `connectedEnvironments` and calls `onReconnectEnvironment` for each id (backed by `retryNow`, so it can skip long reconnect backoff). The `arrow.clockwise` control is **disabled** when there are no environments. The same action is wired on **Android** (`AndroidScreenHeader` actions) and **iOS** (`NativeHeaderToolbar.Button`), with ordering comments so both platforms show refresh beside the existing add button.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89e4504b2c93ef0192a90834e09b87f69d628272. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add a refresh-all button to the environments settings screen
[...]

PR #5023, opened 2026-07-30
fix(mobile): pin the workspace status above the thread list

carTloyal123:
## What changed

The "Syncing threads..." status flashed in and out above the thread list, and shoved every row down and back each time it did.

## Why it happens

It was the list's `ListHeaderComponent` — literally row 0 — in both `HomeScreen` and `ThreadNavigationSidebar`. Because `hasSynchronizingShell` re-enters `"synchronizing"` on every websocket resubscribe and retries expected failures every 250ms, a healthy connection still produces a stream of sub-second flips, and **each one mounted or unmounted a row**.

## Before / after

iPhone 17 Pro, iOS 26.5, against a live workspace with two connected environments. Each column is the same screen: connected and idle, then with an environment dropped.

| | Idle | Reconnecting |
|---|---|---|
| **Before** |  |  |
| **After** |  |  |

**Before:** idle shows nothing at all, so the first row sits at the top. The moment the connection drops, the status becomes row 0 and every thread shifts down ~115pt — then snaps back when it clears. That shove-and-snap, repeating on each sub-second sync flip, *is* the flash.

**After:** the bar is always there, so the first row sits at the same offset in both states. Nothing moves; only the bar's own contents change.

## The fix

The status is now pinned *outside* the scroll view and always mounted. Three things make that work:

**It never goes away.** At rest it reports what synced and when — `Synced 2 environments at 1:20 PM` — instead of going blank, so a quiet bar still answers "am I up to date?" rather than leaving the user to infer it from an absence. This needed `readyEnvironmentCount` on `WorkspaceState` (a boolean couldn't carry the count) and reuses the snapshot timestamp the model already tracked.

**Busy states show a spinner**, not a static glyph — it reads as "data is moving" far more clearly. Static symbols are now only for steady states: synced, offline, error.

**The busy/quiet transition is settled before rendering.** It must persist 400ms to appear — longer than the 250ms retry cycle, so self-resolving blips never surface — and holds 900ms once shown. Real faults bypass both delays, since delaying those would hide a problem worth seeing. The decision is a pure function (`planWorkspaceStatusSettlement`) so the timing is unit-tested rather than implicit in a hook.

## Notes

- On iOS the bar needs an explicit `HeaderHeightContext` offset: the list is what normally sits under the transparent nav bar, so an in-flow sibling starts at y=0 and overlaps the system status bar. Android's header is already an in-flow sibling and needs none.
- The full-page empty state deliberately keeps its inline status — with no threads on screen it is the only thing explaining why the page is blank, and there are no rows below it to reflow.
- The Android floating overlay that duplicated this status is gone.

## Verification

- `tsc --noEmit` clean
- `vp lint` clean on changed files
- Mobile suite green (592 tests, 11 new covering the resting label and the settling logic)
- Manually exercised on iOS Simulator as above

(Media lives on a [separate branch](https://github.com/carTloyal123/t3code/tree/media/sync-status-bar) so it stays out of this diff.)

Split from #5022 so the unrelated feature stays independently reviewable.

### DIFF
--- a/apps/mobile/src/features/home/HomeScreen.tsx
+++ b/apps/mobile/src/features/home/HomeScreen.tsx
@@ -18,11 +18,13 @@ import type {
 } from "@t3tools/contracts";
 import { useAtomSet, useAtomValue } from "@effect/atom-react";
 import { AsyncResult } from "effect/unstable/reactivity";
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { useCallback, useContext, useEffect, useMemo, useRef, useState } from "react";
 import { ActivityIndicator, FlatList, Platform, Pressable, View } from "react-native";
 import type { SwipeableMethods } from "react-native-gesture-handler/ReanimatedSwipeable";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
 import { useThemeColor } from "../../lib/useThemeColor";
+
+import { HeaderHeightContext } from "@react-navigation/elements";
 
 import { AppText as Text } from "../../components/AppText";
 import { EmptyState } from "../../components/EmptyState";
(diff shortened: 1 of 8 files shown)


carTloyal123:
Slop/too big of a feature for now

PR #5086, opened 2026-07-31
fix(server): regenerate thread titles with the thread's own provider

StiensWout:
## Status

🚧 Draft — investigation done, implementation to follow.

## Investigation result

The reported symptom ("Regenerate title only works on Codex threads") is real in
effect, but the cause is **not** a per-thread provider gate. There is no branch
on the thread's provider anywhere in the flow. What actually happens:

`regenerateThreadTitle` (`apps/server/src/orchestration/Layers/ProviderCommandReactor.ts:835`)
always generates with the **global** `serverSettings.textGenerationModelSelection`
(same file, ~`:862`), which defaults to instance `codex` / model `gpt-5.6-luna`
(`packages/contracts/src/settings.ts:496`, `packages/contracts/src/model.ts:147`).
The thread's own `modelSelection` is never read.

Supporting evidence that the provider itself is not the gate:

- All five drivers implement `generateThreadTitle` identically, including the
  `previousTitle` regeneration prompt (Claude / Codex / Cursor / Grok / OpenCode).
- The UI gate is the per-**environment** capability `threadTitleRegeneration`
  (`apps/server/src/environment/ServerEnvironment.ts:145`), not a per-thread
  provider check (`apps/web/src/components/SidebarV2.tsx:2118`).
- Running the Claude structured-output path with the exact flags the server uses
  (`claude -p --output-format json --json-schema … --dangerously-skip-permissions`)
  returns a valid `{"title": …}`, so the non-Codex text-generation path is healthy.

So: a Claude/Grok/Cursor thread gets its title written by Codex, and if Codex is
not installed or authenticated on the server, regeneration fails on *every*
thread — which reads as "it only works for Codex".

## Other defects found in the same flow

1. **Stranded "Regenerating…" state.** Every `{ _tag: "Superseded" }` return skips
   `dispatchThreadTitleRegenerationCompletion`, so the `titleRegeneration` marker
   stays armed forever. Most reachable via the post-generation recheck
   (`latestThread.title !== previousTitle`): regenerate on a young thread while the
   first-turn auto-title is still landing and the marker never clears.
2. **No client-side timeout.** `isThreadTitleRegenerationPending` /
   `threadTitleRegenerationRemainingMs` exist on the #4814 branch but never landed
   in `main`. A stranded marker leaves the row dimmed and the menu item disabled
   with no retry; only a server restart clears it via
   `clearInterruptedThreadTitleRegenerations`.
3. **Serial worker.** `threadTitleRegenerationWorker` is a single-consumer
   `DrainableWorker`, so bulk "Regenerate titles (N)" runs one CLI spawn at a time
   and one slow provider call blocks every other thread's regeneration.
4. **Silent no-op.** When the model returns the previous title verbatim, the flow
   completes with no title and no user feedback.

## Planned fix

- Generate with the thread's own `modelSelection`, falling back to the global
  text-generation selection only when that instance is unavailable.
- Make every terminal path dispatch a completion so the marker always clears.
- Land the pending-timeout helpers so a stuck row self-heals and stays retryable.
- Surface generation failures instead of only logging them.


StiensWout:
Closing — not a bug.

The missing "Regenerate title" action was server version skew, not a provider gate. The `t3code` environment was served by `t3 serve` on v0.0.31-nightly.20260729.944, which predates #4810 and therefore does not advertise the `threadTitleRegeneration` capability. Per `packages/contracts/src/environment.ts:52` that key is `optionalKey` and clients hide the action when it is absent, so SidebarV2 correctly omitted the menu item. First build that has it is 0.0.32-nightly.20260730.956.

The three latent defects noted in the description above (generation always uses the global `textGenerationModelSelection` rather than the thread's provider; `Superseded` returns skip the completion dispatch and strand the `titleRegeneration` marker; the #4814 pending-timeout helpers never landed in main) are real but unrelated to the reported symptom. Leaving them for a separate PR.

Issue #4533, opened 2026-07-25
[Feature]: Allow selection of a default terminal


lmerynda:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I am describing a concrete problem or use case, not just a vague idea.

### Area

apps/server

### Problem or use case

I'm a windows user and I cannot change terminal that t3 code runs in as well as the sidebar terminal. It makes all the commands that LLM execute run in powershell instead of linux-styled terminal like git bash. I don't want to maintain windows scripts in my project. If t3 code would allow changing default terminal, then people can avoid the windows mess while running t3 code on windows

### Proposed solution

In ideal world I would just be using WSL, but t3 code doesn't connect to it properly and I know the full support for it is coming. However, changing default terminal would be a simple but helpful feature. Vscode has it nicely done where I can select which terminal I spawn and also it allows to configure in settings the default one

### Why this matters

People don't need to maintain powershell scripts in their projects while running t3 code on windows

### Smallest useful scope

A single setting which would allow selection of a default terminal in which t3 code is spawned

### Alternatives considered

Tried running T3 code in wsl, but it doesn't seem to connect properly, the app just hangs and I needed to remove ~/.t3 to start the configuration all over again

### Examples or references

Vscode functionality around managing default terminal and spawning new terminals is a good sample of what solves these issues

### Contribution

- [x] I would be open to helping implement this.

PR #5122, opened 2026-07-31
perf(desktop): unblock Windows startup and shrink the installer

simon-curtis:
## Problem

On Windows, the installed desktop app could spend nearly a minute with no window while synchronous command discovery and provider initialization blocked backend readiness. Upgrade installation was similarly opaque and slow because oversized packages contained thousands of unpacked files that the previous uninstaller and antivirus scanned individually.

Windows users also could not choose the executable used for new T3 terminal sessions.

## How this fixes it

- Resolves Windows shell commands with a cached, bounded native lookup instead of thousands of sequential filesystem probes.
- Preserves PATH/PATHEXT precedence, handles directories that permit direct probing but deny listing, and invalidates stale cached hits safely.
- Installs a cheap Windows PATH baseline synchronously, then hydrates PowerShell profile PATH/FNM state asynchronously without delaying backend readiness.
- Refreshes live provider environment objects and environment-sensitive caches after hydration without rebuilding adapters or interrupting active turns.
- Serializes provider reconciliation and makes cancellation cleanup transactional so scopes cannot leak or remain publicly visible after closure.
- Defers provider maintenance path discovery, shares concurrent work, retries interrupted discovery, and invalidates successful results only when environment hydration requires it.
- Omits the optional Linux/WSL runtime and Linux-only staged dependencies from Windows packages unless a Linux node-pty prebuild is supplied.
- Shows explicit removing, installing, and finishing phases in the Windows installer.
- Adds an environment-owned default terminal executable setting for new T3 terminal sessions, including remote clients. This does not change shells chosen internally by provider CLIs.

## Windows performance results

The original comparison used installed production builds on the same Windows host. Startup was measured from process spawn to the first completed HTTP 200 and to the first visible Electron window.

| Build | Backend ready | Window |
|---|---:|---:|
| Previous Nightly | 46.0 s | 46.6 s to window creation |
| New release, cold | 4.352 s | 5.866 s visible |
| New release, warm 1 | 4.492 s | 5.994 s visible |
| New release, warm 2 | 4.234 s | 5.724 s visible |
| New release median | **4.352 s** | **5.866 s** visible |

Performance test head (`e4a322213`) was rebuilt as a production Windows package and rerun from its staged executable with isolated `APPDATA`, isolated `T3CODE_HOME`, and a dedicated port. The staged package was used to avoid overwriting the developer's installed app.

| Latest-head rerun | Backend ready | Window visible |
|---|---:|---:|
| Run 1 | 4.160 s | 5.600 s |
| Run 2 | 11.290 s | 18.250 s |
| Run 3 | 4.282 s | 5.814 s |
| Median | **4.282 s** | **5.814 s** |

The latest-head median is 1.6% faster to HTTP readiness and 0.9% faster to a visible window than the original new-release median. Run 2 is retained as a host-level outlier rather than discarded; temporary unpacked builds showed intermittent Windows scanning/contention, while the other two runs and the median remained in the original range. Individual Windows command-resolution spans remained bounded rather than reproducing the previous synchronous startup stall.

All measurements require HTTP 200 rather than treating an open TCP socket as ready. Each run was fully stopped before the next launch and used isolated state rather than the live user database.

## Packaging and installation results

| Metric | Before | After | Change |
|---|---:|---:|---:|
| NSIS installer | 324.19 MB | 134.55 MB | **58.5% smaller** |
| Unpacked runtime payload | 890 MB | 29.07 MB | **96.7% smaller** |
| Unpacked files | 3,532 | 200 | **94.3% fewer** |
| Observed upgrade | ~100 s removing the previous build alone | ~21 s complete upgrade | substantially faster |

Upgrade timings are observational rather than a controlled benchmark: the old measurement removed the original 2,494-file Nightly, while the final measurement replaced an already-compacted build.

## User-visible behavior

- General Settings exposes the default terminal executable, searchable and resettable.
- The setting belongs to the connected environment, so web, desktop, and mobile-created terminals use the same value. Mobile does not currently expose the editing control.
- Windows installer banners show removal, installation, and completion phases.
- User documentation describes scope, quoting, fallback behavior, and remote environments.

## Verification

- Rebased onto upstream `main` at `2a04db134c2d88f06e5b8d61a8410cb51ea07430`.
- **211 focused tests passed across 12 files** under Node 24.13.1; the final Claude environment fix then passed 90/90 core tests plus its portable text-generation regression.
- Post-review provider registry regressions passed **10/10 focused tests across 2 files**, including secret-materialization failure and interrupted reconciliation coverage.
- Affected server, web, desktop, contracts, shared, and scripts typechecks passed.
- Targeted lint, formatting, and `git diff --check` passed.
- The custom NSIS include compiled with warnings-as-errors using Electron Builder's pinned NSIS 3.12; the latest head also completed a fresh production Windows package build and startup rerun.
- Focused coverage includes command cache/PATH/PATHEXT behavior, inaccessible directory fallback, runtime PATH shell resolution, settings subscription ordering, asynchronous environment hydration, provider environment override precedence, maintenance cancellation/invalidation, concurrent and interrupted provider reconciliation, settings UI/search/reset behavior, WSL package inclusion/omission, and installer phase/error ordering.

## Checklist

- [x] This PR is focused on the Windows experience
- [x] The branch is rebased onto current upstream main
- [x] Focused tests and affected typechecks pass
- [ ] I included before/after screenshots for UI changes
- [ ] I included a short video for timing/interaction changes

Model: GPT-5 / GPT-5.4  
Harness: OpenAI Codex desktop app with dev-loop specialist review

---

> [!NOTE]
> **Medium Risk**
> Touches Windows PATH timing at startup, provider registry reconciliation, and live environment refresh for all drivers; incorrect hydration or refresh ordering could leave tools undiscovered until profile load completes or briefly stale maintenance metadata.
> 
> **Overview**
> This PR targets **Windows startup and upgrades** by stopping synchronous PowerShell profile work from blocking server readiness, tightening command/PATH resolution, and refreshing provider state in place after the profile finishes loading.
> 
> **Startup and environment:** The server applies a **baseline Windows PATH** immediately and **forks PowerShell profile hydration** (`PATH`, FNM vars) in the background. The desktop app no longer spawns PowerShell to build `PATH` at launch. Shared **`resolveCommandPath`** gains caching, PATH/PATHEXT precedence, concurrent probing, and safer invalidation when directories are unreadable.
> 
> **Providers:** Instance env is built from **`makeProviderInstanceEnvironmentSource`** with **`refreshEnvironment`** so host PATH changes merge without recreating adapters. Maintenance capabilities are **lazy/cached** and refresh after hydration. After Windows profile hydration completes, the registry **refreshes snapshots in place** (no reconcile/replace). Reconcile is **serialized and interrupt-safe** with staged entries during rebuilds. Settings watcher **re-reads settings** and can **restore redacted secrets** when secret-store reads fail.
> 
> **Terminals and settings:** New **`defaultTerminalShell`** setting (General settings, contracts, docs) drives spawn behavior; Windows resolves configured names through PATH, treats the value as an executable only (no embedded args), and subscribes to live settings updates.
> 
> **Installer / docs:** NSIS hooks show explicit **“Removing the previous version…”** / install phase text. README links **terminal shell** user docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b990f526e9bb5121e67a4c7c74b526a09955fc93. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Improve Windows environment hydration, command-path resolution, and terminal shell configuration
> - On Windows, server startup now applies a baseline PATH immediately and hydrates PowerShell profile environment variables (PATH, FNM) asynchronously in the background via `hostEnvironmentHydrationLayer` in [os-jank.ts](https://github.com/pingdotgg/t3code/pull/5122/files#diff-01387cb8f0521a1e152ae377c15805f328384e404ccf93da7a82c02d30cf694c), unblocking startup.
> - Adds a `defaultTerminalShell` setting to `ServerSettings` and the General settings panel; on Windows the configured shell name is resolved against PATH/PATHEXT before spawning.
> - Windows command-path resolution in [shell.ts](https://github.com/pingdotgg/t3code/pull/5122/files#diff-89a91f85db41a9a0912bc86c336bf9429751f3797ee50a089248c68cde8489ce) now uses a per-process cache keyed on platform/command/PATH/PATHEXT, probes PATH directories concurrently (batched 32, concurrency 16), and handles directory-listing failures and case-insensitive semantics.
> - Provider drivers (Claude, Codex, Cursor, Grok, OpenCode) gain a `refreshEnvironment` effect that re-syncs inherited host env values and invalidates maintenance capability caches without replacing the instance.
> - `ProviderInstanceRegistry.reconcile` is now serialized, interruption-safe, and supports a `{ force: true }` option; previous snapshots remain visible during rebuild.
> - Windows Desktop builds without a WSL runtime no longer unpack the full asar payload; NSIS installer now displays phase text ('Removing the previous version…' / 'Installing the new version…') during one-click install.
> - Risk: FNM_DIR and FNM_MULTISHELL_PATH are no longer populated by `installWindowsEnvironment` on Desktop; PATH enrichment from the PowerShell profile happens asynchronously and may arrive after the first terminal is opened.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b990f52.</sup>
> 
>

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ Full docs live in [docs/](./docs). There's no docs site yet.
 - [Install and first run](./docs/user/install.md)
 - [Permission modes](./docs/user/permission-modes.md)
 - [Keyboard shortcuts](./docs/user/keybindings.md)
+- [Choose the default terminal shell](./docs/user/terminal-shell.md)
 - [Remote access from a phone or another machine](./docs/user/remote-access.md)
 - [Keeping app and server in sync](./docs/user/updating.md)
 - [Source control integrations](./docs/user/source-control.md)
--- a/apps/desktop/resources/installer.nsh
+++ b/apps/desktop/resources/installer.nsh
@@ -1,0 +1,54 @@
+!include "getProcessInfo.nsh"
+
+Var pid
+
+!macro setInstallerPhaseText TEXT
+  !ifndef BUILD_UNINSTALLER
+    ${IfNot} ${Silent}
+      FindWindow $0 "#32770" "" $hwndparent
+      FindWindow $0 "#32770" "" $hwndparent $0
+      GetDlgItem $0 $0 1000
+      SendMessage $0 ${WM_SETTEXT} 0 "STR:${TEXT}"
+    ${EndIf}
+  !endif
+!macroend
+
+# electron-builder calls this immediately before checking for and removing the
+# installed version. Keep its standard app-running check, then make the long
+# silent-uninstaller phase explicit in the one-click installer banner.
+!macro customCheckAppRunning
+  !insertmacro IS_POWERSHELL_AVAILABLE
+  !insertmacro _CHECK_APP_RUNNING
+  !insertmacro setInstallerPhaseText "Removing the previous version..."
+!macroend
+
+# electron-builder delegates these hooks the old uninstaller result handling.
+# Preserve its standard failure behavior before advancing the phase text.
+!macro handleOldUninstallerResult NEXT_TEXT
+  IfErrors 0 +3
+  DetailPrint `Uninstall was not successful. Not able to launch uninstaller!`
+  Return
+
+  ${If} $R0 != 0
+    MessageBox MB_OK|MB_ICONEXCLAMATION "$(uninstallFailed): $R0"
+    DetailPrint `Uninstall was not successful. Uninstaller error code: $R0.`
+    SetErrorLevel 2
+    Quit
+  ${EndIf}
+
+  !insertmacro setInstallerPhaseText "${NEXT_TEXT}"
+!macroend
+
+!macro customUnInstallCheck
+  # An all-users installation can also remove a per-user installation next.
+  # Keep showing the removal phase until both checks have completed.
+  ${If} $installMode == "all"
+    !insertmacro handleOldUninstallerResult "Removing the previous version..."
+  ${Else}
+    !insertmacro handleOldUninstallerResult "Installing the new version..."
+  ${EndIf}
+!macroend
+
+!macro customUnInstallCheckCurrentUser
+  !insertmacro handleOldUninstallerResult "Installing the new version..."
+!macroend
--- a/apps/desktop/src/shell/DesktopShellEnvironment.test.ts
+++ b/apps/desktop/src/shell/DesktopShellEnvironment.test.ts
@@ -197,7 +197,7 @@ describe("DesktopShellEnvironment", () => {
     }),
   );
 
-  it.effect("loads PowerShell profile environment on Windows", () =>
+  it.effect("hydrates the Windows PATH without starting a shell", () =>
     Effect.gen(function* () {
       const env: NodeJS.ProcessEnv = {
         PATH: "C:\\Windows\\System32",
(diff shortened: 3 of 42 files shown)


simon-curtis:
Fixes #4533

PR #5132, opened 2026-07-31
fix(ssh): surface a failed remote t3 install instead of a silent 0-byte server.log

### LINKED ISSUE #5129: CLI exits 0 with no output when node-pty fails to build (missing C toolchain); breaks desktop SSH launch with empty server.log
**Environment:** Ubuntu 26.04 LTS minimal server (Hetzner), Node 24.18.1 via mise, npm 11.x. Reproduced with `t3@0.0.31` and `t3@0.1.0-alpha.1`.

**Repro:** On a machine without a C toolchain (no `build-essential`):

```
$ npx -y t3@0.0.31 --version
$ echo $?
0
```

Prints **nothing**, exits 0. Same for `--help`, `connect`, `serve`.

**Root cause:** `node-pty` fails its `node-gyp` build (no gcc/make). npm swallows the failure — the npx cache dir ends up essentially empty, `node_modules/.bin/t3` never materializes, and `npx` still exits 0 having run nothing. A plain `npm install t3` in an empty dir shows the real `gyp ERR!` output.

**Knock-on effect:** the desktop app's SSH-launch flow fails with `SshCommandError: Remote T3 server did not become ready on 127.0.0.1:3773` and a **0-byte** `~/.t3/ssh-launch/<key>/server.log` — nothing points at the actual cause. Installing `build-essential` fixes everything instantly.

**Suggestions:**
- preflight check with a clear "install a C toolchain" message
- surface the npm install stderr in the ssh-launch `server.log`
- or ship prebuilds for node-pty on common platforms

TorbenWetter:
Fixes #5129

## What happens
On a Linux host without a C toolchain, `npx --yes t3@… <anything>` exits 0 having executed nothing: npm extracts the package before running its dependencies' native builds, so a failed `node-pty` compile leaves the npx cache with no `t3` executable. In the desktop SSH-launch flow the generated `run-t3.sh` `exec`s that npx call, so the launch "succeeds" instantly, `server.log` stays 0 bytes, and the desktop only ever reports `Remote T3 server did not become ready on 127.0.0.1:3773`. (The cloud/boot path already guards this npm behavior in `pinnedRuntime.ts`; only the SSH path used bare npx.)

## The fix (packages/ssh/src/tunnel.ts)
- `run-t3.sh` now resolves the CLI up front (`npx --yes --package '<spec>' -- sh -c 'command -v t3'`) so an install failure is reported at launch time, with npm's own stderr flowing into `server.log`, and a clear message naming the likely cause (missing C toolchain / node-pty).
- When the readiness probe times out with an *empty* log, the error now says so explicitly instead of tailing nothing.

## Validation
- `pnpm test` in `packages/ssh`: 25/25 pass (assertions extended for the new script content); `pnpm typecheck`, root `pnpm lint`, `vp fmt --check` all clean.
- Generated scripts pass `sh -n`; runner exercised against a stub npx reproducing the broken shape (exit 0, no output → now exit 1 with the new message) and the healthy shape (unchanged `exec` path).
- Real-registry check on npm 11 / node 24: the preflight resolves and prints the bin path.
- Caveat: original repro box was fixed by installing `build-essential` before this patch existed, so the patched script was validated against a simulated broken host, not the live one.

https://claude.ai/code/session_018SnL8Kiba6YBbxNnunPiHg

---

> [!NOTE]
> **Low Risk**
> Changes are limited to generated shell scripts and test assertions for the SSH tunnel path; no auth or data-handling logic.
> 
> **Overview**
> Fixes a case where **SSH remote launch** looked healthy but the server never started: on hosts without a C toolchain, `npx --yes t3@…` can exit **0** with no `t3` binary after a failed **node-pty** build, leaving an empty `server.log` and only a generic readiness timeout.
> 
> The generated **`run-t3.sh`** now runs **`require_installed_t3_cli`** before `exec` via `npx`/`npm exec` (resolves `command -v t3` after install) and fails with a message that points at missing build tools. The **launch script** only tails the log when it is non-empty; otherwise it states that nothing was written to the log path.
> 
> Tests in **`tunnel.test.ts`** assert the new script fragments and empty-log messaging.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d037e31f38f5dd4e5abae6474fb3d660075f750b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Surface failed remote t3 install errors instead of a silent 0-byte server.log
> - Adds a `require_installed_t3_cli` helper to the remote runner script ([tunnel.ts](https://github.com/pingdotgg/t3code/pull/5132/files#diff-7338be65675ab0699d0c8c6426e54f3421747885231f9bbb0ace4ec129cb2072)) that pre-resolves the `t3` executable after `npx`/`npm exec` installs it, aborting with a descriptive stderr message if no executable is found (likely due to native dependency build failure).
> - Updates the launch script failure path to check if the log file is non-empty before tailing it; prints `'It wrote nothing to %s, so it exited before producing any output.'` when the log is empty.
> - Adds test coverage in [tunnel.test.ts](https://github.com/pingdotgg/t3code/pull/5132/files#diff-76d3724ec8959625b119aff8272a5d733f73551706a487e303ddbd13e10a06ad) for both the preflight CLI check and the conditional log-tailing behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d037e31.</sup>
> 
>
>

### DIFF
--- a/packages/ssh/src/tunnel.test.ts
+++ b/packages/ssh/src/tunnel.test.ts
@@ -108,6 +108,9 @@ describe("ssh tunnel scripts", () => {
     assert.include(script, "exec npx --yes 't3@latest' \"$@\"");
     assert.include(script, "exec npm exec --yes 't3@latest' -- \"$@\"");
     assert.include(script, "could not install 't3@latest'");
+    assert.include(script, "require_installed_t3_cli npx --yes --package 't3@latest'");
+    assert.include(script, "require_installed_t3_cli npm exec --yes --package 't3@latest'");
+    assert.include(script, "npm produced no t3 executable");
     assert.include(script, 'prepend_path_if_dir "$HOME/.local/bin"');
     assert.include(script, `T3_NODE_ENGINE_RANGE='${TEST_NODE_ENGINE_RANGE}'`);
     assert.include(script, "remote_node_satisfies_engine()");
@@ -140,6 +143,10 @@ describe("ssh tunnel scripts", () => {
 
     assert.include(script, "exec npx --yes 't3@nightly; touch /tmp/t3-owned' \"$@\"");
     assert.include(script, "exec npm exec --yes 't3@nightly; touch /tmp/t3-owned' -- \"$@\"");
+    assert.include(
+      script,
+      "require_installed_t3_cli npx --yes --package 't3@nightly; touch /tmp/t3-owned'",
+    );
     assert.notInclude(script, "exec npx --yes t3@nightly; touch /tmp/t3-owned");
   });
 
@@ -185,6 +192,8 @@ describe("ssh tunnel scripts", () => {
     assert.notInclude(buildRemoteLaunchScript(), "server-home");
     assert.include(buildRemoteLaunchScript(), "Remote T3 server did not become ready");
     assert.include(buildRemoteLaunchScript(), 'wait_ready "60000"');
+    assert.include(buildRemoteLaunchScript(), 'if [ -s "$LOG_FILE" ]; then');
+    assert.include(buildRemoteLaunchScript(), "It wrote nothing to %s");
     assert.include(buildRemoteLaunchScript({ packageSpec: "t3@nightly" }), "t3@nightly");
     assert.include(
       buildRemotePairingScript(target),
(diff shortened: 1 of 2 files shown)


TorbenWetter:
Good catch — fixed in 71ae1bff1: the quoted spec is now passed as a printf `%s` argument instead of being embedded in the single-quoted format string, so it stays one safe word on the error path as well. `pnpm test` (packages/ssh) and typecheck still green.

PR #5153, opened 2026-07-31
feat(web): add favorite thread markers

HabsaTheDog:
## What changed

- Add an **Add to favorites / Remove from favorites** action to thread context menus in both sidebar versions.
- Persist favorites locally by scoped thread key, including sanitization during hydration.
- Show a filled amber star immediately before the existing timestamp or status slot.
- Keep the marker visible while the row's timestamp/status yields to hover and focus actions.

## Why

There was no lightweight way to visually mark an important thread for later. This keeps the feature local to existing UI state and avoids server, database, or protocol changes.

## Fix

The initial marker was nested inside the timestamp/status element, which intentionally fades out on row hover and focus. A context-menu interaction leaves the row in that state, so the favorite toggled correctly while its star remained hidden. The marker is now a sibling immediately before that transient slot in Sidebar V1, Sidebar V2 cards, and Sidebar V2 slim rows.

## User impact

Right-clicking a thread now toggles a persistent favorite marker that stays visible, including immediately after the context-menu action. The marker survives app restarts and uses the existing amber palette in light and dark themes.

## Validation

- Targeted formatting and lint for both sidebar components
- `vp test run apps/web/src/uiStateStore.test.ts` (16 tests passed)
- Original web unit suite validation (201 files, 1,766 tests passed)

Generated with Codex (GPT-5.6) in T3 Code via the Codex harness.

### DIFF
--- a/apps/web/src/components/Sidebar.tsx
+++ b/apps/web/src/components/Sidebar.tsx
@@ -9,6 +9,7 @@ import {
   LoaderIcon,
   SearchIcon,
   SquarePenIcon,
+  StarIcon,
   TerminalIcon,
   TriangleAlertIcon,
 } from "lucide-react";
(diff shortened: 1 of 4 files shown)


t3dotgg:
> [!NOTE]

Closing this PR after an automated pass over open pull requests. Abandoned favorite-thread draft is replaced by shipped thread pinning in #5312.

PR #5206, opened 2026-08-01
fix(composer): detect slash-command trigger mid-prompt, not just at line start

matdac12:
## What changed

`detectComposerTrigger` only recognized `/` when it was the very first character of the current line (`linePrefix.startsWith("/")`), while `$skill` and `@path` use whitespace-delimited token detection that works anywhere in the text. Typing `/command` in the middle of a prompt therefore never opened the command/skill menu — only starting the message with it did.

This adds a `/` branch to the same token scanner the `$` trigger uses, in both copies of the logic:

- `apps/web/src/composer-logic.ts` (web)
- `packages/shared/src/composerTrigger.ts` (mobile — keeps its `slash-model` special case, so a mid-prompt `/model` opens the model picker just like at line start)

## Why

Slash commands behave inconsistently with the other two composer triggers: `fix the tests $my-skill` autocompletes, while `fix the tests /review` silently does nothing.

## Behavior preserved

- Line-start detection is untouched (the existing line-prefix branch still runs first).
- `/model <args>` on mobile still works via the existing line-prefix branch.
- `"/model spark"` still yields no trigger (existing test unchanged).
- Tokens merely containing slashes (`src/components`, `https://…`) do not trigger, since the token must start with `/`.

## Tests

Added unit tests for mid-prompt `/command`, bare `/`, mid-prompt `/model` (shared), and the non-trigger cases above, in `composer-logic.test.ts` and `composerTrigger.test.ts` (the latter previously had no `detectComposerTrigger` coverage). Full unit suites of `@t3tools/web` (1769 tests) and `@t3tools/shared` (323 tests) pass.

### DIFF
--- a/apps/web/src/composer-logic.test.ts
+++ b/apps/web/src/composer-logic.test.ts
@@ -94,6 +94,40 @@ describe("detectComposerTrigger", () => {
     });
   });
 
+  it("detects slash command trigger in the middle of existing text", () => {
+    const text = "Fix the tests /rev";
+    const trigger = detectComposerTrigger(text, text.length);
+
+    expect(trigger).toEqual({
+      kind: "slash-command",
+      query: "rev",
+      rangeStart: "Fix the tests ".length,
+      rangeEnd: text.length,
+    });
+  });
+
+  it("detects a bare slash mid-text as an empty slash command query", () => {
+    const text = "Fix the tests /";
+    const trigger = detectComposerTrigger(text, text.length);
+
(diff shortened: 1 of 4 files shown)


t3dotgg:
> [!NOTE]

Opening slash-command suggestions after whitespace makes ordinary prose with slash-prefixed tokens interactive and duplicates trigger logic across clients. Keeping slash commands at line start is clearer and does not need this extra rule.

Issue #277, opened 2026-03-07
Per-project MCP


kimkhoi2202:
I have a lot of projects that I'm working on in parallel, which is the point of me using Codex and T3 Code. I want to be able to configure different MCP for different projects. For example, I can use Supabase MCP connected to Supabase Account #1 for project #1 and Supabase MCP connected to Supabase Account #2 for project #2 and so on.

If this is a feature, I don't care what state T3 Code is at (as long as it's stable), I'm going to pay $200/month and cancel my other subscriptions because of how much time it will save me from disconnecting and connecting new MCP every time I switch project. This is the main blocking point stopping me from working on multiple projects at the same time and utilize the full power of Codex.

PR #5227, opened 2026-08-02
feat(settings): list the MCP servers Claude Code loads

jash90:
## What Changed

Adds **Settings → MCP**: a read-only listing of the MCP servers Claude Code loads, on every connected computer, so you can see them without opening a terminal or a config file.

- `apps/server/src/mcpServers/` reads Claude Code's own config — user scope and workspace `local` scope from `.claude.json`, plus approved `.mcp.json` project servers. Reading the files directly avoids `claude mcp list`, which health-checks every server over the network and is far too slow for a settings page. **Nothing is ever written back.**
- `GET /api/mcp-servers` on the environment HTTP API, with the matching client-runtime accessor. Per-environment, so local, relay, and tunnel all work the same way.
- The page groups by computer and then by provider instance, with transport and scope badges, the redacted command line or address, and a click-to-copy config path.

Two things the read is deliberate about:

- **`complete` is load-bearing.** It goes false whenever a config exists but cannot be parsed, and whenever no workspace `cwd` was supplied — the `local` and `project` scopes are unreadable then, so the list is knowingly partial. A later caller that replaces the CLI's own resolution with `--strict-mcp-config` must refuse to act on an incomplete list, or it will silently drop working servers. An unreadable config is reported on its own channel rather than as a property of rows, because it usually yields no rows at all and the empty state would otherwise claim there are no servers.
- **Nothing secret crosses the wire.** `env` is never serialised. Arguments and URLs go through a shape allowlist *plus* a credential-flag check — neither alone is enough: shape catches `postgres://u:p@host/db` and `Authorization: Bearer …` but would happily print a bare `hunter2`, while a flag blocklist catches `--token secret` but misses everything positional. Server names are stripped of control characters and length-bounded, since a U+202E override reorders how a row reads on screen.

Only Claude is covered. Codex needs a `codex mcp list --json` subprocess; Cursor, Grok, and OpenCode run over ACP, which can *add* a server to a session but never reports the ones the agent loads from its own config, so listing them would be misleading.

## Why

This is a rebased, re-scoped port of #4634, which @juliusmarminge closed on 2026-07-30:

> Several of these are real gaps we still want fixed; the base just moved out from under them. Once #2829 merges, please rebase onto `main`, port the change to the V2 equivalent, and reopen (or open a fresh PR).

#2829 is still open and currently conflicted, and the gap is still open too — #277. T3 Code injects itself into every session as an MCP server but gives users no view of the ones their own harness loads. So this targets today's `main`, and is packaged to survive the V2 migration rather than to dodge it:

- Everything substantive lives in modules V2 keeps: `apps/server/src/mcpServers/`, `packages/contracts/`, `packages/client-runtime/`, `apps/web/`.
- **This PR does not touch `ClaudeAdapter.ts` or `CodexAdapter.ts` at all** — the files V2 replaces. It is pure discovery plus a settings page; nothing in the session launch path changes.

Happy to re-cut against V2 the moment #2829 lands. Turning servers off is deliberately not in this PR — that is where the adapter changes live, and it is worth its own review.

## Per-surface decisions

- **Entry points.** Sidebar nav and the Settings search catalog. The section is anchored via `searchableSetting`, so a search hit scrolls to it. The command palette has no settings navigation today, so nothing to add there.
- **Providers.** Claude only, for the reason above. Codex is a follow-up; ACP harnesses are not listable.
- **Clients.** Web and desktop. Mobile has no Providers screen either, so MCP is out of scope there.
- **Connection modes.** One collapsible group per environment, each issuing its own authenticated HTTP request; a response from a replaced connection is dropped rather than written over live data.
- **Reverse states.** "Refresh all" re-reads from disk, so a config edited outside T3 Code shows up.
- **Docs.** `docs/user/mcp.md` linked from `docs/README.md`, a `docs/internals/providers.md` section, and two glossary entries — `docs/` had no mention of MCP before this.

## Verification

`vp test run` on the touched files (83 passing across 14 files), `vp lint` and `vp fmt` on the changed scope, and `typecheck` on `@t3tools/contracts`, `@t3tools/client-runtime`, `t3`, and `@t3tools/web`. `routeTree.gen.ts` is generator output, not hand-edited.

Exercised against a real 25-server config in the desktop app, plus adversarial configs: `__proto__`/`constructor` as server names (no prototype pollution — `Object.entries` and a `Map`, never assignment), `command` as an array or number, null entries, unknown transports, 5000 levels of nesting, and a 12 MB config against the 8 MB read guard. All degrade rather than throw; the empty-state fix above came out of that pass.

## Checklist

- [ ] I included a video for animation/interaction changes — n/a, no motion

---

Claude Opus 5 (1M context), via Claude Code.

---

> [!NOTE]
> **Medium Risk**
> New authenticated remote endpoint surfaces filesystem-derived MCP config metadata; redaction and incomplete-read handling must stay correct so secrets are not leaked and partial lists are not mistaken for empty configs. Session launch is unchanged.
> 
> **Overview**
> Adds **Settings → MCP** and a read-only **`GET /api/mcp-servers`** path so each connected environment can report which MCP servers Claude Code would load from its own config—without writing harness files or changing session launch.
> 
> **Server:** New `mcpServers` modules parse `.claude.json` (user + workspace `local` scopes) and approved `.mcp.json` project entries, mirroring Claude’s scope precedence and `CLAUDE_CONFIG_DIR` / symlink behavior. Reads return a **`complete`** flag and per-file **`unreadable`** signals when parsing fails or workspace scopes were skipped. Inventory rows redact stdio args and remote URLs, sanitize display names, and aggregate across enabled Claude provider instances.
> 
> **Contracts & client:** `McpServerInventory` types and `client-runtime` `fetchEnvironmentMcpInventory` call the authenticated environment HTTP API.
> 
> **Web:** `/settings/mcp` lists servers per environment with search, refresh, grouping, copyable config paths, and warnings when configs could not be fully read. Docs and nav/search entries are updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1598fd564ccd0c773a7b7f2afead98a1c032ef98. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add MCP servers settings page listing servers loaded by Claude Code
> - Adds a new `/settings/mcp` route and `McpServersSettings` component that lists MCP servers across all connected environments, with search and refresh support.
> - Introduces a `GET /api/mcp-servers` endpoint backed by `discoverGlobalMcpInventory`, which reads user, project, and local scope configs from Claude's `.claude.json` and `.mcp.json` files.
> - Servers are grouped by harness/provider instance, showing transport, scope, status, and a redacted detail string; config paths are copyable and shortened for display.
> - Unreadable or malformed config files are surfaced as warnings rather than silently dropped.
> - Risk: credential redaction in stdio command args relies on heuristic flag-name matching, so edge cases may leak or over-redact values.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1598fd5.</sup>
> 
>


t3dotgg:
> [!NOTE]

This reimplements Claude Code's MCP configuration rules across 1,819 lines to show a read-only list in Settings. Claude can change those rules independently, which leaves T3 Code maintaining precedence, redaction, and incomplete-state behavior across the full stack.

PR #5257, opened 2026-08-03
fix(codex): inherit environment for personal skills

anirudhsama:
## What changed

- Always extend the host environment when spawning Codex app-server sessions.
- Preserve explicit provider environment overrides while inheriting host variables such as `HOME` and `USERPROFILE`.
- Add a spawn-boundary regression test for app-server sessions with provider environment overrides.

## Why

Codex discovers personal skills from `$HOME/.agents/skills`. T3 previously set `extendEnv: false` whenever a provider environment object was present, so a partial environment could remove the home-directory variables Codex needs and personal skills would not load.

This is a T3-side workaround for openai/codex#28505.

## Validation

- `vp test run apps/server/src/provider/Layers/CodexSessionRuntime.test.ts` (21 tests)
- `vp fmt --check` on both changed files
- `vp lint --report-unused-disable-directives` on both changed files

> [!NOTE]
> ### Fix `makeCodexSessionRuntime` to always inherit host environment when spawning Codex
> Previously, `extendEnv` was set to `true` only when no environment override was provided, causing the Codex child process to run in an isolated environment when overrides were present. [CodexSessionRuntime.ts](https://github.com/pingdotgg/t3code/pull/5257/files#diff-e29bf409c8e737a2d4e655f7e9e7def9c78c008d98a0a23575baa5e714c4e12c) now unconditionally sets `extendEnv=true` so host environment variables are always inherited alongside any provided overrides. Behavioral Change: Codex child processes now always inherit the full host environment, even when an explicit `env` override is passed.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5f1b3c0.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/CodexSessionRuntime.test.ts
+++ b/apps/server/src/provider/Layers/CodexSessionRuntime.test.ts
@@ -1,10 +1,13 @@
 import * as NodeAssert from "node:assert/strict";
 
+import * as NodeServices from "@effect/platform-node/NodeServices";
 import { it } from "@effect/vitest";
 import * as Effect from "effect/Effect";
+import * as PlatformError from "effect/PlatformError";
 import * as Schema from "effect/Schema";
 import { describe } from "vite-plus/test";
 import { DEFAULT_MODEL, ThreadId } from "@t3tools/contracts";
+import { ChildProcessSpawner } from "effect/unstable/process";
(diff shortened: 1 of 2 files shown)


anirudhsama:
Closing for now so the workaround can be tested on the fork first. I’ll reopen or raise a fresh upstream PR after validation.

PR #5288, opened 2026-08-03
fix(mobile): keep pending input card below the navigation header

KyleKincer:
## What changed

`PendingApprovalCard` / `PendingUserInputCard` render inside the bottom-anchored composer overlay (`KeyboardStickyView`) in `ThreadDetailScreen` with no height cap. When an agent asks several questions (each with option pills plus a custom-answer input), the card grows upward past the top of the screen and renders underneath the transparent navigation header — the card's "User input needed / Fill in the pending answers" heading collides with the header title and back button.

This caps the pending-card stack at the space between the navigation header and the composer (`window height − header height − composer chrome − 12`), and wraps the cards in a `ScrollView` so overflowing questions scroll inside the card area instead of growing under the header. Header height is read from `HeaderHeightContext` with the same `insets.top + 44` fallback `ThreadFeed` already uses. No visual change for short cards — the cap only engages when the card would otherwise underlap the header.

## Why

With a two-question user-input request on an iPhone, the card covers the entire screen and overlaps the header, making the header unreadable and the top of the card unreachable behind it.

## Before / After

Captured on an iPhone 17 Pro simulator (iOS 26.2), dev client paired to a local backend, with a Claude thread raising a two-question `AskUserQuestion`.

| Before (main) | After (this PR) | After — scrolled |
| --- | --- | --- |
| before | after | after scrolled |

Before: the card's heading renders on top of the nav header title and back chevron. After: the card stops below the header and scrolls internally; the Submit answers button stays reachable.

Verified with `pnpm typecheck` (mobile), `pnpm lint`, `pnpm fmt:check`, and live on the simulator as shown above.

### DIFF
--- a/apps/mobile/src/features/threads/ThreadDetailScreen.tsx
+++ b/apps/mobile/src/features/threads/ThreadDetailScreen.tsx
@@ -2,6 +2,7 @@ import { type EnvironmentConnectionPhase } from "@t3tools/client-runtime/connect
 import type { EnvironmentThreadStatus } from "@t3tools/client-runtime/state/threads";
 import { useKeyboardChatComposerInset, useKeyboardScrollToEnd } from "@legendapp/list/keyboard";
 import type { LegendListRef } from "@legendapp/list/react-native";
+import { HeaderHeightContext } from "@react-navigation/elements";
 import type {
   ApprovalRequestId,
   EnvironmentId,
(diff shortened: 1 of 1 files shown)


carlosricojr:
Both bot findings here look real to me, so rather than open a competing PR I've sent a patch to this branch: KyleKincer/t3code#1.

It subtracts the keyboard height (the cards ride `KeyboardStickyView`, so a near-max card goes back under the header once a custom-answer input is focused), replaces the `Math.max(120, …)` floor with the composer's own height so a short viewport collapses the cards instead of clipping the composer, and bounds the overlay container rather than the card stack so the composer's measured height — not `COMPOSER_*_CHROME` — decides the split. The clamp moves to `lib/layout.ts` with tests for those cases.

Yours to take or leave, @KyleKincer.

t3dotgg:
> [!NOTE]

Closing as covered by the newer questionnaire layout. Pending input now owns the composer slot and reserves keyboard space, rather than stacking above the composer with a fixed minimum height. Keep the keyboard-open and short-viewport cases from the follow-up review against that current layout. Related work: [#5986](https://github.com/pingdotgg/t3code/pull/5986).

Issue #728, opened 2026-03-09
macOS: apps launched from integrated terminal can't request TCC permissions (mic, camera)


mikaelweiss:
## Problem

When running an app that needs microphone or camera access from T3 Code's integrated terminal, macOS never shows the permission prompt — it silently fails. The same app works fine when launched from a standalone terminal (e.g. Ghostty).

This affects both dev and production builds.

## Cause

macOS TCC requires two things to show privacy permission prompts for child processes:

1. A valid code signature on the parent app bundle
2. Usage description keys in the app's `Info.plist` (e.g. `NSMicrophoneUsageDescription`)

### Dev builds

The dev launcher (`apps/desktop/scripts/electron-launcher.mjs`) patches the Electron app bundle's `Info.plist` (display name, bundle ID, icon) but doesn't re-sign the bundle afterward. This invalidates the existing code signature, so TCC silently denies all permission prompts for child processes.

### Production builds

The mac build config in `scripts/build-desktop-artifact.ts` has two issues:

1. **No usage description keys** — `NSMicrophoneUsageDescription`, `NSCameraUsageDescription`, etc. are not included in the mac `extendInfo`, so macOS doesn't know the app needs those permissions.
2. **Unsigned builds have no signature at all** — `CSC_IDENTITY_AUTO_DISCOVERY` is set to `"false"` for all unsigned builds, which disables signing entirely. Without even an ad-hoc signature, TCC can't attribute permissions to the app.

### Dev (~11 lines, 1 file)

Ad-hoc re-sign the app bundle after patching plists in `electron-launcher.mjs`:

```js
spawnSync("codesign", ["--force", "--deep", "--sign", "-", appBundlePath]);
```

I've tested this locally and it resolves the issue for dev.

### Prod

1. Add `NSMicrophoneUsageDescription` and `NSCameraUsageDescription` to the mac build config via `extendInfo`.
2. For unsigned mac builds, use ad-hoc signing (`identity: "-"`) instead of disabling signing entirely.

I haven't been able to test the prod fix locally (no signing setup), but the approach matches [Apple's TCC documentation](https://developer.apple.com/documentation/bundleresources/information_property_list/protected_resources).

## Reproduction

1. Open a project in T3 Code (desktop, macOS)
2. In the integrated terminal, run any app that requests microphone or camera access
3. No permission prompt appears; the app fails silently
4. Run the same command in an external terminal — it works

## PR
[...]

PR #5321, opened 2026-08-04
fix(desktop): allow microphone access

emilwagman:
my app needs microphone access when i run it from t3 code's terminal. macos treats t3 code as the responsible app, but the signed desktop build doesn't include the audio input entitlement, so the permission prompt never appears.

this adds that entitlement to the existing mac signing config.

tested: `vp test run scripts/build-desktop-artifact.test.ts`

> [!NOTE]
> ### Add microphone entitlement and `entitlementsInherit` to macOS builds
> - Adds the `com.apple.security.device.audio-input` entitlement to the generated plist in `renderMacPasskeyEntitlements` to enable microphone access.
> - Sets `entitlementsInherit` in `createBuildConfig` to the provided entitlements path when `macPasskeySigning` is enabled.
> - Updates tests in [scripts/build-desktop-artifact.test.ts](https://github.com/pingdotgg/t3code/pull/5321/files#diff-a199d37af35c35dd3b48cf71436cb06b23acccd6857a9fe74db6d48cb73c7663) to verify both additions.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 88c1d42.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Changes macOS code-signing entitlements, including microphone access and helper-process inheritance. Incorrect entitlements can break signing or broaden device access.
> 
> **Overview**
> Signed macOS desktop builds now include the `com.apple.security.device.audio-input` entitlement so the OS can show a microphone permission prompt when T3 Code is the responsible app (e.g. tools launched from its terminal).
> 
> The same entitlements file is also passed as `entitlementsInherit` in electron-builder’s mac config, so helper processes inherit that signing profile. Tests cover both the generated plist and the inherit path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88c1d4236778027b1bcb17800978af958d8c363b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


alidarvishi14:
Independently confirming the `cursor[bot]` finding above, since it is what "Needs human review" is currently hanging on — and adding why this PR matters beyond its original motivation.

**The finding holds.** `entitlementsInherit` appears **zero** times anywhere in this repo. `createBuildConfig` sets only `entitlements: macPasskeySigning.entitlementsPath` (`scripts/build-desktop-artifact.ts:2080`), with no inherit counterpart, so the entitlement this PR adds lands on the main bundle and is not inherited by the Electron helper binaries.

**That splits into two cases, and this PR fully fixes one of them:**

1. **A process launched from T3's integrated terminal** (the author's stated case). macOS attributes the TCC request to the responsible app — the main bundle — so the entitlement on the main plist is exactly what is needed. This PR is correct and sufficient here, and also addresses #728 and #7242.

2. **`getUserMedia` / `MediaRecorder` called from the renderer.** The renderer runs in a helper process, which is signed separately and does not inherit the main bundle's entitlements without `entitlementsInherit`. This path would stay blocked.

**Case 2 is no longer hypothetical.** Every in-flight voice PR records from the web renderer — #5213 and #6625 both add `apps/web/src/lib/voiceTranscription.ts` + `useVoiceTranscription.ts` using `MediaRecorder`, and #6206 and #3997 take microphone input on desktop too. So whichever of those eventually lands will need `entitlementsInherit` regardless, and would rediscover this same gap.

Given that, it seems worth either landing this as-is for case 1 and tracking the helper-process entitlement separately, or adding `entitlementsInherit` here so both paths are covered at once. Either way this is a 3-line change that is currently the only thing standing between signed builds and the OS even being willing to *show* a microphone prompt — worth not letting it go stale.

emilwagman:
Thanks for catching this. I’ve added `entitlementsInherit` alongside `entitlements` so Electron helper/renderer processes receive the microphone entitlement too. I also added a regression assertion, and the desktop artifact tests pass (`30 passed`). Pushed in commit `12ea55e63`.

PR #5339, opened 2026-08-04
feat(web): open a project's terminal from the chat header

Brechard:
## What Changed

Adds a **Terminal** control to the chat header, next to Open-in, that launches the project directory in an installed terminal.

- Supported: Terminal, iTerm, Warp and Ghostty on macOS via their app bundles, plus Ghostty, WezTerm, kitty, Alacritty and Windows Terminal via their CLIs. Only installed ones appear.
- The preferred terminal is remembered exactly like the preferred editor (`t3code:last-terminal` in localStorage, mirroring `usePreferredEditor`). The chevron menu only renders when more than one terminal is installed, so the common case is a single button.
- `EDITORS` gains a `working-directory` launch style. Its target is resolved to a directory first: a `:line:column` suffix is stripped, and a file steps up to its containing folder, so the existing "open this file" call sites work unchanged.
- The `commands === null` file-manager special case becomes a `kind` discriminator (`file-manager` | `terminal`), so terminals reuse the same detection and launch path instead of adding a parallel one.
- Terminals are excluded from the Open-in picker's fallback, so a terminal can never become the editor the Open button uses.

`macAppName` is set only for bundles that actually open a folder argument in that directory. The rest would silently start in the user's home, so they appear only when their CLI is on PATH.

## Why

Opening a project in an editor is one click, but opening a shell in that same directory means leaving the app and `cd`-ing by hand.

It is a separate control rather than a row in the Open-in picker because a terminal opens a shell *in* the project rather than a file in an app, and choosing one should not move the editor the Open button uses.

## UI Changes

Before — header with no terminal affordance:

before

After — Terminal control between Add action and Open:

after

The screenshots are from a machine with only Terminal.app installed, which is why no chevron is shown; with several terminals installed the control gains a chevron menu listing them.

## Verification

- `vp run --filter @t3tools/contracts --filter t3 --filter @t3tools/web typecheck` clean.
- New tests in `externalLauncher.test.ts` (working-directory launch with `:line:column` stripping, and the CLI-less macOS app-bundle path) and `packages/contracts/src/editor.test.ts` (terminal classification, launch style, ordering after editors).
- Checked end to end in a dev build: clicking Terminal spawned a shell whose `cwd` was the project directory, confirmed with `lsof`.

## Checklist

- [ ] I included a video for animation/interaction changes — n/a, no motion

Generated with Claude Sonnet 5 in Claude Code.

---

> [!NOTE]
> **Medium Risk**
> Touches process spawning and path resolution on the server; mistakes could open the wrong directory or mis-detect macOS apps, but scope is limited to external launch helpers with new tests.
> 
> **Overview**
> Adds a **Terminal** control in the chat header (alongside Open-in) that opens the project directory in an installed terminal, with a separate preferred-terminal preference (`t3code:last-terminal`) so it does not change the Open-in editor default.
> 
> **Contracts & launcher:** `EDITORS` gains terminal entries and a **`working-directory`** launch style with optional `cwdArgs` / `macAppName`. `ExternalLauncher` resolves the cwd from file paths (strip `:line:column`, use parent dir), detects macOS `.app` bundles only on darwin, and launches GUI-only terminals via `open -a`. File manager is distinguished with `kind: "file-manager"` instead of `commands === null` alone.
> 
> **Web:** `OpenTerminalPicker` filters installed terminals from `availableEditors`; `usePreferredEditor` / `resolveAndPersistPreferredEditor` skip terminal IDs so Open-in never defaults to a terminal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cf0826276aa0bf8a792aef33b9e683ca759d118c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add terminal launcher to the chat header for opening a project directory
> - Adds an `OpenTerminalPicker` component to [`ChatHeader.tsx`](https://github.com/pingdotgg/t3code/pull/5339/files#diff-124f699b87245f05f11a8bf7ef3d0b17dbe34657c443a51e23125ce90e0fcf0c) that lets users open a terminal in the current project's working directory, with preference persistence and multi-terminal selection.
> - Defines terminal editor entries (Apple Terminal, iTerm, Ghostty, Warp, WezTerm, kitty, Alacritty, Windows Terminal) in [`editor.ts`](https://github.com/pingdotgg/t3code/pull/5339/files#diff-952d09e8dcc92b201e44f8f0b5edc71af952e9af9479b5853c41ce6e93d636f7) using a new `working-directory` launch style and optional `cwdArgs` templates.
> - Extends [`externalLauncher.ts`](https://github.com/pingdotgg/t3code/pull/5339/files#diff-157ca66b5181820e0a628ef443d244577cda2329dd1d88fe5f2aff441da05e56) to resolve the working directory from a file path (stripping `:line:col` suffixes), support GUI-only macOS app bundles via `open -a`, and skip macOS bundle fallback on non-darwin platforms.
> - Separates terminal preference storage from editor preference in [`editorPreferences.ts`](https://github.com/pingdotgg/t3code/pull/5339/files#diff-b95642bd2ec02008455dc34d764b6b380dc3cb63bf4eb8fc249e0a105220efdd) using a new `usePreferredTerminal` hook so terminals are never auto-selected as the default editor.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cf08262.</sup>
> 
>

### DIFF
--- a/apps/server/src/process/externalLauncher.test.ts
+++ b/apps/server/src/process/externalLauncher.test.ts
@@ -252,6 +252,12 @@ it.effect("opens a CLI-less macOS terminal through its app bundle", () =>
     yield* fileSystem.writeFileString(path.join(binDir, "open"), "#!/bin/sh\n");
     yield* fileSystem.chmod(path.join(binDir, "open"), 0o755);
     const workspace = yield* fileSystem.makeTempDirectoryScoped({ prefix: "t3-workspace-" });
+    // The bundle lives under a temp HOME so the test does not depend on the
+    // machine it runs on having Terminal.app.
+    const home = yield* fileSystem.makeTempDirectoryScoped({ prefix: "t3-home-" });
+    yield* fileSystem.makeDirectory(path.join(home, "Applications", "Terminal.app"), {
+      recursive: true,
+    });
 
     let spawned: ChildProcess.StandardCommand | undefined;
     yield* Effect.gen(function* () {
@@ -261,7 +267,7 @@ it.effect("opens a CLI-less macOS terminal through its app bundle", () =>
       Effect.provide(
         testLayer({
           platform: "darwin",
-          env: { PATH: binDir },
+          env: { PATH: binDir, HOME: home },
           onSpawn: (command) => {
             spawned = command;
           },
--- a/apps/server/src/process/externalLauncher.test.ts
+++ b/apps/server/src/process/externalLauncher.test.ts
@@ -155,6 +155,71 @@ it.effect("discovers editors through the service API", () =>
   }).pipe(Effect.scoped, Effect.provide(NodeServices.layer)),
 );
 
+it.effect("opens a terminal at the directory holding the launch target", () =>
+  Effect.gen(function* () {
+    const fileSystem = yield* FileSystem.FileSystem;
+    const path = yield* Path.Path;
+    const binDir = yield* fileSystem.makeTempDirectoryScoped({ prefix: "t3-terminals-" });
+    yield* fileSystem.writeFileString(path.join(binDir, "ghostty"), "#!/bin/sh\n");
+    yield* fileSystem.chmod(path.join(binDir, "ghostty"), 0o755);
+    const workspace = yield* fileSystem.makeTempDirectoryScoped({ prefix: "t3-workspace-" });
+    const file = path.join(workspace, "index.ts");
+    yield* fileSystem.writeFileString(file, "");
+
+    let spawned: ChildProcess.StandardCommand | undefined;
+    yield* Effect.gen(function* () {
+      const launcher = yield* ExternalLauncher.ExternalLauncher;
+      // A file with a position, as the "open this config file" flows send.
+      yield* launcher.launchEditor({ editor: "ghostty", cwd: `${file}:12:4` });
+    }).pipe(
+      Effect.provide(
+        testLayer({
+          platform: "linux",
+          env: { PATH: binDir },
+          onSpawn: (command) => {
+            spawned = command;
+          },
+        }),
+      ),
+    );
+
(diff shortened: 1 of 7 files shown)


Brechard:
Thanks — both findings were real, fixed in e44db99.

- **`externalLauncher.ts:405`** — guarded the CLI branch's app-bundle fallback with `platform === "darwin"`, matching the no-CLI branch below it. Off darwin this now falls through to the plain command-not-found error instead of an `open -a …` launch that cannot work there.
- **`resolveWorkingDirectory`** — the target is now stat'd as-is before any position parsing, so a real directory whose name ends in `:<number>` is left alone. The suffix is only stripped when the target itself is not a directory.

Added a regression test for each.

Brechard:
Good catch, fixed in cf08262. The CLI-less branch now calls `isMacAppAvailable` first and fails with `ExternalLauncherUnsupportedEditorError` when the bundle is gone, rather than detaching `open` and reporting success.

Worth noting it also caught a latent test problem: the existing app-bundle test forced `platform: "darwin"` against the real filesystem, so gating on availability would have tied it to the host actually having `Terminal.app` — green on macOS, red on a Linux runner. It now stages the bundle under a temp `HOME`, so it is host-independent.

I did not add a negative test for the missing-bundle path: `isMacAppAvailable` also probes the real `/Applications`, so any such test would pass or fail depending on what the runner happens to have installed. Happy to add one if you'd prefer the probe roots injected instead.

juliusmarminge:

**Reason:** Conflicted open-project-terminal-from-header; terminal/header UX moved on.

PR #5344, opened 2026-08-04
perf(client): batch large thread sync updates

danvernon:
## Summary

- Batch live thread stream items in the shared client runtime, applying ordered items locally and publishing the final state once per batch.
- Bound batches to 64 items or 16 ms so large-thread sync updates do not trigger web/mobile derivation work for every streamed event.
- Preserve sequence deduplication, completion markers, deletion handling, and cache persistence behavior.

## Why

Large threads can produce many message and activity updates. Previously each event updated the shared `SubscriptionRef`, causing the web and mobile clients to repeatedly rebuild large thread views during synchronization.

## Validation

- `./node_modules/.bin/vp test run packages/client-runtime/src/state/threads-sync.test.ts`
- `./node_modules/.bin/vp test run packages/client-runtime/src/state/threadReducer.test.ts packages/client-runtime/src/state/threads-atoms.test.ts`
- Targeted client-runtime typecheck, lint, formatting, and diff checks

The PR intentionally focuses on client-side publication cost; server snapshot/replay and per-event reducer optimizations remain separate follow-up work.

> [!NOTE]
> ### Batch large thread sync updates to reduce state publications in `EnvironmentThreadState`
> - Introduces `reduceThreadStreamItems`, a pure reducer that processes a batch of `OrchestrationThreadStreamItem` events and produces a single updated state and a persistable snapshot.
> - Stream processing in `EnvironmentThreadState.make` now uses `Stream.groupedWithin` to group items into batches (default size 64, window 16ms) before applying them, so a burst of events results in a single state publication.
> - When a batch contains a settled state followed by non-persistable events, the settled snapshot is enqueued for persistence before the turn starts.
> - Callers can override the batch size via a new `eventBatchSize` option on `EnvironmentThreadStateOptions`.
> - Behavioral Change: state updates are now deferred up to 16ms during a burst; downstream subscribers will see fewer, coarser-grained updates.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 783fd02.</sup>
> 
>


t3dotgg:
> [!NOTE]
>

Closing in favor of [#8309](https://github.com/pingdotgg/t3code/pull/8309), where review of client event batching continues. That version covers thread and shell updates, keeps pagination state, and rejects buffered events from replaced subscriptions. It already includes the settled-mid-batch snapshot case from this proposal. The batching fix is not yet merged.

PR #5345, opened 2026-08-04
perf(telemetry): reduce idle resource monitoring CPU

danvernon:
## Summary

- Reduce idle resource-monitor sampling from 1 second to 5 seconds when diagnostics are not being viewed.
- Refresh expensive memory, disk, and command details only for tracked processes, preserving live diagnostics command refreshes.
- Share desktop-bootstrap polling and settings relative-time ticking across consumers.
- Document the telemetry behavior and update focused tests.

## Why

The always-on telemetry path repeatedly scanned and refreshed process data, while mounted UI consumers created duplicate timers. Those redundant periodic tasks consumed CPU while the app was idle.

## Validation

- `cargo fmt --check --manifest-path native/resource-monitor/Cargo.toml`
- `cargo test --locked --manifest-path native/resource-monitor/Cargo.toml` (15 passed)
- `./node_modules/.bin/vp test run apps/server/src/resourceTelemetry/NativeTelemetryClient.test.ts` (10 passed)
- Targeted web typecheck, lint, formatting, and diff checks

This PR is separate from the large-thread synchronization batching PR.

> [!NOTE]
> ### Reduce idle CPU usage in resource monitoring and UI polling
> - In the native resource monitor, periodic background samples now refresh CPU only for all processes, then fetch memory/disk/cmd details only for tracked PIDs; command lines are re-read only when unset. Live and on-demand samples continue to refresh everything unconditionally.
> - `resolveNativeSampleIntervalMs` now returns a 5,000ms interval when there are no live diagnostic subscribers instead of always using 1,000ms.
> - `useRelativeTimeTick` and `useDesktopLocalBootstraps` are rewritten with `useSyncExternalStore` so multiple consumers share a single timer/polling loop that stops when there are no subscribers.
> - Behavioral Change: background telemetry sampling interval increases from 1,000ms to 5,000ms when no live subscribers are present.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a94c104.</sup>
> 
>


juliusmarminge:

**Reason:** Idle resource-monitor CPU; overlapping native diagnostics already landed in #2679.
**Related:** #2679

PR #5390, opened 2026-08-05
Insert file mentions when non-image files are dropped on the composer

0xjohnnydev:
## Problem

Dropping a file onto the composer routes **every** dropped file through the image-attachment path (`onComposerDrop` → `addComposerImages`). Anything that isn't an image is rejected with:

> Unsupported file type for '<name>'. Please attach image files only.

So there's no way to drag a file from Finder/Explorer to reference it by path — you can only do it by typing `@` or dragging from the in-app file tree.

## Change

`onComposerDrop` now partitions the dropped files:

- **Images** → attach inline, exactly as before.
- **Other files** → become composer file mentions, using the dropped `File`'s on-disk path resolved via Electron `webUtils.getPathForFile`, exposed as a new optional `DesktopBridge.getPathForFile`.
  - Paths inside the workspace cwd are made **workspace-relative** so they match typed / file-tree mentions; paths outside stay absolute.
- Files whose path **can't** be resolved (browser builds, in-memory `File`s) fall back to the existing image path, so the "unsupported file type" error still surfaces and **non-desktop behavior is unchanged**.

## Files

- `packages/contracts/src/ipc.ts` — add optional `DesktopBridge.getPathForFile` (typed `unknown` since contracts is built without the DOM lib).
- `apps/desktop/src/preload.ts` — implement it via `webUtils.getPathForFile` (synchronous, in-process, no IPC hop).
- `apps/web/src/components/chat/composerFileDrop.ts` — `partitionDroppedComposerFiles`, `toComposerMentionPath`, `buildDroppedFileMentions` (pure helpers) + `composerFileDrop.test.ts`.
- `apps/web/src/components/chat/ChatComposer.tsx` — wire helpers into `onComposerDrop`; `insertComposerTextAtEnd` moved above the drop handler so it's defined before use.

## Notes / open questions

- Absolute-vs-relative: in-workspace drops become relative to match existing mentions; out-of-workspace drops keep the absolute path so the reference stays unambiguous. Happy to change this if you'd rather always keep absolute (or reject out-of-workspace drops).
- cwd prefix match is case-sensitive, so a drop differing only in case from the cwd (possible on case-insensitive volumes) keeps its absolute path rather than guessing.
- **Remote environments (reviewer finding, needs a maintainer call):** `getPathForFile` resolves paths on the *local desktop host*, while mentions are resolved against the *environment* workspace. On a local environment (the common case) these agree. On SSH/WSL environments the inserted mention would be a local path the backend can't read (before this PR such drops errored loudly). Options: gate mention insertion on the environment being local, or accept the dangling reference. Happy to implement whichever you prefer.
- Dropping a **folder** now inserts a folder mention instead of erroring (folders have empty MIME type, so they take the mention path).
- Draft: opened for testing/review before it's ready to merge.

> [!NOTE]
> ### Insert file-mention tokens when non-image files are dropped on the composer
> - Adds a `getPathForFile(file)` method to `window.desktopBridge` (via [preload.ts](https://github.com/pingdotgg/t3code/pull/5390/files#diff-eae0078d2bf4ac4beed2884f918b2816a97752917390f09980ba2cfad8c42cac) and the `DesktopBridge` interface) that synchronously returns the absolute backing path for a DOM `File`.
> - Introduces [composerFileDrop.ts](https://github.com/pingdotgg/t3code/pull/5390/files#diff-2cab160942e8a7d0823e6492111ae960dbe4f6207f615c31b589e27dd7a11cd6) with helpers to partition dropped files into images vs. non-images, normalize paths relative to cwd, and serialize them as composer file-mention tokens.
> - Updates the `onComposerDrop` handler in `ChatComposer` to resolve paths via `desktopBridge`, insert mention tokens for resolved non-image files, and fall back to image attachment for unresolved files. An error toast is shown if the composer is busy when mentions need to be inserted.
> - Behavioral Change: focus is deferred when mentions are inserted to avoid overwriting the insertion snapshot.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 254a259.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Composer drop behavior changes on desktop (mentions vs image errors), and resolved paths are host-local which may not match SSH/WSL workspace paths.
> 
> **Overview**
> **OS file drops on the chat composer** no longer send every file through the image-attachment path. Dropped files are split into images (still attached inline) and everything else, which on desktop can become **file mention tokens** in the prompt.
> 
> A new optional **`DesktopBridge.getPathForFile`** (contracts + Electron preload via `webUtils.getPathForFile`) resolves the on-disk path for each non-image `File`. **`composerFileDrop`** helpers relativize in-workspace paths to match typed/file-tree mentions, serialize links with `serializeComposerFileLink`, and are covered by unit tests. **`ChatComposer.onComposerDrop`** inserts those mentions (with error toast if the composer is busy), avoids eager focus after a successful mention insert, and routes path-unresolved files back through **`addComposerImages`** so browser builds and in-memory files keep the prior “unsupported file type” behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 254a259580c82ec365f7faa48e0717de7f113b10. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/desktop/src/preload.ts
+++ b/apps/desktop/src/preload.ts
@@ -5,7 +5,7 @@ import type {
   DesktopPreviewTabState,
 } from "@t3tools/contracts";
 import { exposeClerkBridge } from "@clerk/electron/preload";
-import { contextBridge, ipcRenderer } from "electron";
+import { contextBridge, ipcRenderer, webUtils } from "electron";
 
 import * as IpcChannels from "./ipc/channels.ts";
 
@@ -105,6 +105,9 @@ contextBridge.exposeInMainWorld("desktopBridge", {
       ...(position === undefined ? {} : { position }),
     }),
   openExternal: (url: string) => ipcRenderer.invoke(IpcChannels.OPEN_EXTERNAL_CHANNEL, url),
+  // Synchronous and in-process: webUtils reads the path off the drag payload,
+  // no IPC hop. Returns "" for files without a backing path.
+  getPathForFile: (file) => webUtils.getPathForFile(file as File),
   onMenuAction: (listener) => {
     const wrappedListener = (_event: Electron.IpcRendererEvent, action: unknown) => {
       if (typeof action !== "string") return;
--- a/apps/web/src/components/chat/ChatComposer.tsx
+++ b/apps/web/src/components/chat/ChatComposer.tsx
@@ -2474,11 +2474,15 @@ export const ChatComposer = memo(function ChatComposer(props: ChatComposerProps)
         unresolvedFiles.push(file);
       }
     }
+    let insertedMentions = false;
     if (mentionPaths.length > 0) {
-      const inserted = insertComposerTextAtEnd(buildDroppedFileMentions(mentionPaths, gitCwd), {
-        ensureLeadingBoundary: true,
-      });
-      if (!inserted) {
+      // Trailing space: the mention token grammar requires trailing
+      // whitespace, same as every other mention insert site.
+      insertedMentions = insertComposerTextAtEnd(
+        `${buildDroppedFileMentions(mentionPaths, gitCwd)} `,
+        { ensureLeadingBoundary: true },
+      );
+      if (!insertedMentions) {
         toastManager.add({
           type: "error",
           title: "Unable to add to chat",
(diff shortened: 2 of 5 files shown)


t3dotgg:
> [!NOTE]

[#6203](https://github.com/pingdotgg/t3code/pull/6203) is the stronger local file-mention implementation because it covers drop and paste, shows a compact chip, limits paths to the local desktop environment, and warns for remote targets. This branch covers drop only and leaves absolute-path behavior unresolved.

PR #5453, opened 2026-08-05
fix: clear pending user-input requests when the provider session is gone

NeilTheFisher:
## What Changed

Treat the respond-failure detail `"No active provider session is bound to this thread."` as a terminal/stale pending-request failure in the matchers shared by the web, mobile, server decider, and projection pipeline. This matches how `"Unknown pending user-input request"` failures are already handled.

## Why

When the server restarts while a `user-input.requested` (or approval) popup is open, the provider session dies and the thread session becomes `stopped`. Submitting a response makes the server append a `provider.user-input.respond.failed` activity whose detail is `"No active provider session is bound to this thread."`.

The client's `derivePendingUserInputs` only dismisses a request when it sees a `user-input.resolved` activity or a failure whose detail matches the stale-pending-request pattern. That detail did not match, so the popup stayed open forever — every submit just appended another failure, and the box could never be closed. This is the same "callback state does not survive restarts" case the existing stale matcher was built for.

A user hit this after restarting the server while the question popup was in place: the box was stuck open and the message could not be submitted.

## Checklist

- [x] Included tests for the new detail string in web `session-logic.test.ts`, server `decider.settled.test.ts`, and `ProjectionPipeline.test.ts`

Model + harness: DeepSeek V4 Flash via opencode.

---

> [!NOTE]
> **Low Risk**
> Small, consistent string-matching changes across client and server derivation logic; behavior matches existing stale-failure handling with new tests.
> 
> **Overview**
> After a **server restart**, responding to an open approval or user-input prompt can fail with `"No active provider session is bound to this thread."` That detail was not treated as terminal, so clients kept showing a pending request that could never be satisfied.
> 
> This PR adds that string to the existing **stale pending-request failure** matchers in **web**, **mobile**, **server decider**, and **projection pipeline**, so `provider.*.respond.failed` activities clear pending approvals/user-input the same way as unknown/stale request errors. **Thread settle** logic and projected `pendingUserInputCount` / approval counts stay aligned.
> 
> Tests cover web `derivePendingUserInputs`, decider settle with the new detail, and projection pipeline pending user-input clearing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c5615a78bbe791a57d26f7179cc33cdca4981ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


NeilTheFisher:
Fixes #5454 — pending user-input popup can never be dismissed after the provider session dies.

PR #5536, opened 2026-08-06
fix(web): preserve closed plan panel state

Zeko369:
## Problem

Reopening a thread could auto-open its Plan or Tasks panel even when the user had previously closed it. Existing steps loaded during thread restoration were treated like newly arrived plan updates, overriding the thread's persisted right-panel state.

This made returning to plan-bearing threads disruptive, especially while a turn was still running and continuing to publish task status updates.

## Root cause

The auto-open effect only compared the active plan's turn with the latest turn. On navigation or hydration, the restored plan still belonged to that latest turn, so the effect could not distinguish an existing snapshot from a new update.

## Fix

- Track plan snapshots by their source activity ID.
- Establish a restoration baseline after thread detail finishes loading.
- Preserve an open right panel or a previously closed Plan surface when entering a thread.
- Seed the existing per-thread dismissal tracker when restoring a closed Plan surface, so later updates from the same running turn do not reopen it.
- Keep first visits eligible for automatic opening when the setting is enabled.
- Keep carried-over tasks and manually dismissed turns closed.
- Only consume the "Implement in a new thread" open intent once the target thread exists.

The same logic covers both Plan and Tasks because they share the plan right-panel surface.

## Testing

- `pnpm exec vp test run apps/web/src/components/ChatView.logic.test.ts apps/web/src/session-logic.test.ts apps/web/src/rightPanelStore.test.ts` — 142 tests passed
- Targeted `vp lint` for the five changed files
- Targeted `vp fmt --check` for the five changed files

Browser QA was not run. The behavior is covered by focused state-transition tests.

Reviewed with Claude Opus via `claude -p`.

Implemented with GPT-5.6 Codex via T3 Code.

> [!NOTE]
> ### Preserve closed plan panel state across thread activation and restoration
> - Introduces `resolvePlanSidebarAutoOpen`, `resolvePlanSidebarRestoration`, and `resolvePlanSidebarTurnKey` utilities in [ChatView.logic.ts](https://github.com/pingdotgg/t3code/pull/5536/files#diff-464e876afd6be3800453dd8cc94d805066953bed62ffd825bbd6fdfd86f4d114) to centralize plan sidebar open/close decisions.
> - The plan sidebar now tracks a per-thread activation baseline so it only auto-opens for plan updates that arrive after the thread is activated, not for pre-existing plans.
> - A previously closed panel on the `plan` surface is preserved on thread restoration; dismissal is seeded for that turn so the panel does not re-open.
> - `ActivePlanState` in [session-logic.ts](https://github.com/pingdotgg/t3code/pull/5536/files#diff-9f1f9f19555f14c3c47d4ce52e1b7dcec7e968e54f400820085376797a7223d6) gains an `updateId` field, used to detect whether a plan update is new since the last observed baseline.
> - Behavioral Change: the sidebar now returns `'wait'` during thread-detail loading, `'preserve'` when restoring prior state, and `'none'` in cases such as dismissed turns or stale plans, replacing the previous unconditional open behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 87ead15.</sup>
> 
>


Zeko369:
If I delete tests it's an XS 😅

PR #5551, opened 2026-08-06
feat(web): remove Build/Plan toggle from the composer

t3dotgg:
The Build/Plan toggle in the composer input box is gone. Plan mode is now a legacy feature: Settings → Beta has a "Restore plan mode (legacy)" switch that brings the toggle back for anyone still relying on the old workflow.

Reality check for those upset:
image

With the flag off (the default), the composer no longer shows the Build/Plan control, the compact "…" menu drops its Chat/Plan radio group, the `/plan` and `/default` slash commands disappear from the command menu, and Shift+Tab no longer flips modes. The effective interaction mode is also forced to "build" while the flag is off — even for threads that had plan mode persisted — so nobody can be trapped in plan mode with the toggle hidden. The next send persists the forced mode back to the thread.

Implementation: a new `planModeEnabled` client setting (default false) in the contracts schema, gating plus mode clamping in `ChatComposer`/`ChatView`, and a Beta settings row wired into settings search.

---
Change made by Claude Fable 5 via Claude Code.

---

> [!NOTE]
> **Low Risk**
> Client-side UI and interaction-mode gating only; default-off behavior may surprise users with existing plan threads until they send a message or enable the beta flag.
> 
> **Overview**
> **Plan mode is now legacy and hidden by default.** A new client setting `planModeEnabled` (default `false`) in the settings contract controls whether Build/Plan is available at all.
> 
> With the flag **off**, `ChatView` forces the effective interaction mode to **default** even when a thread still has plan mode stored, so users are not stuck without UI to exit; the next send is expected to persist default. The composer hides the Build/Plan toggle, drops `/plan` and `/default` from the slash menu, disables Shift+Tab mode switching, and treats standalone `/plan`/`/default` as normal message text.
> 
> With the flag **on**, behavior matches the prior plan-mode workflow (toggle, shortcuts, and slash commands). **Settings → Beta** adds **Restore plan mode (legacy)** wired into settings search; desktop settings tests include the new default field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a7f0fc2e70ec0c6bf8d72d5af0f5db807aa91412. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Remove Build/Plan toggle from the composer behind a `planModeEnabled` beta flag
> - Adds `planModeEnabled` (default `false`) to `ClientSettingsSchema` in [settings.ts](https://github.com/pingdotgg/t3code/pull/5551/files#diff-82d7525e9c2a6a61cd5d416ceaeba40992140373eb7ac00fa3a91c2edd00bc3c); the flag can be toggled via a new switch in the Beta settings panel.
> - When `planModeEnabled` is false, the interaction mode is forced to `DEFAULT_INTERACTION_MODE` in `ChatView`, and the `/plan`/`/default` slash commands are treated as plain text.
> - Hides the Build/Plan toggle UI, its `Shift+Tab` shortcut, and the `/plan`/`/default` command palette entries in `ChatComposer` when the flag is off.
> - Adds a 'Restore plan mode (legacy)' entry to settings search pointing to the Beta section.
> - Behavioral Change: Plan mode is now off by default for all users; existing composer interaction-mode state is ignored unless the beta flag is re-enabled.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a7f0fc2.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/settings/DesktopClientSettings.test.ts
+++ b/apps/desktop/src/settings/DesktopClientSettings.test.ts
@@ -30,6 +30,7 @@ const clientSettings: ClientSettings = {
   fontSizeTerminal: 12,
   fontSmoothing: true,
   glassOpacity: 80,
+  planModeEnabled: false,
   providerModelPreferences: {},
   sidebarAutoSettleAfterDays: 3,
   sidebarProjectGroupingMode: "repository_path",
--- a/apps/web/src/components/ChatView.tsx
+++ b/apps/web/src/components/ChatView.tsx
@@ -1467,8 +1467,13 @@ function ChatViewContent(props: ChatViewProps) {
     ? (localServerError ?? activeServerThread?.session?.lastError ?? null)
     : localDraftError;
   const runtimeMode = composerRuntimeMode ?? activeThread?.runtimeMode ?? DEFAULT_RUNTIME_MODE;
-  const interactionMode =
-    composerInteractionMode ?? activeThread?.interactionMode ?? DEFAULT_INTERACTION_MODE;
+  // Plan mode is legacy (Settings → Beta). With the flag off the effective
+  // mode is forced to "default" — even for threads with a stored plan mode —
+  // so nobody is trapped in plan mode while its toggle is hidden. The next
+  // send persists "default" back to the thread.
+  const interactionMode = settings.planModeEnabled
+    ? (composerInteractionMode ?? activeThread?.interactionMode ?? DEFAULT_INTERACTION_MODE)
+    : DEFAULT_INTERACTION_MODE;
   const isLocalDraftThread = !isServerThread && localDraftThread !== undefined;
   const canCheckoutPullRequestIntoThread = isLocalDraftThread;
   const activeThreadId = activeThread?.id ?? null;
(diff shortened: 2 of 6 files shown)


benkingcode:
This is not a legacy or "old workflow"? Plan mode is a first class and supported feature by the major harnesses.

doasfrancisco:
i expect t3code to have the same options as the harness 

it took me time to switch because t3code didn't posses of claude code's options

keep in mind the more you drift away from the harness, the more it feels incomplete

ideally t3code should have a compatibility table like bun's have

ItriIbouanane:
L

benkingcode:
Also the logic of the chart used in this PR doesn't make sense? You're saying about 3% of turns use Plan mode so that means it should be removed? Why are you looking at turns? Obviously it's a small amount because when a session starts in Plan mode most of the turns come _after_ it's approved and switched to Build mode?

OakerO:
I was already surprised when Theo said in a video that he never uses Plan mode. Fine, different people work differently. However, letting that bias dictate which features are removed for everyone seems like arrogance. Instead of the misleading “percentage of turns” graph, why not show how many users used Plan mode per T3 session? Or was that graph chosen to support your bias? I don't mind if I have to enable it in settings but calling it legacy has major implications that you plan (pun intended) to remove it later and that you will not test it anymore. However, I don't understand why remove it at all. Was it bothering someone?

PR #5561, opened 2026-08-07
fix: prevent reconnect loops during server stalls

gfsaaser24:
Server event-loop stalls could make one delayed pong kill the RPC connection, then immediately reconnect into the same expensive bootstrap work. Users saw repeated disconnects and `Syncing...` even when the underlying network was healthy.

This tolerates short pong delays, uses a safer reconnect ladder, and avoids repeated bootstrap work by caching command and editor discovery and resuming shell state from an authoritative cursor. Failed HTTP snapshot refreshes still request a complete socket snapshot, trace attributes are bounded to preserve useful diagnostics, and sends attempted while disconnected now explain why they were not sent.

Focused tests cover pinger tolerance, wake recovery, shell synchronization, discovery caching, and trace truncation.

🤖 Final review fixes and PR cleanup by GPT-5.6 Codex using the Codex harness in T3 Code.

---

> [!NOTE]
> **Medium Risk**
> Changes core connection supervisor, RPC protocol (patched dependency), and shell subscription resume semantics; mis-tuned wake/backoff or resume cursors could cause extra sync work or delayed reconnects.
> 
> **Overview**
> Hardens **client–server connection behavior** against event-loop stalls, aggressive reconnect loops, and expensive session bootstrap.
> 
> **Transport & supervisor:** Patches Effect RPC **ping/pong** to allow two missed 5s windows before closing (~15s silence). Raises the first reconnect delay from **1s → 3s** and, when a **foreground wake probe** fails, **skips the first backoff sleep** once (then normal backoff resumes). Supervisor and session tests are updated for the new timing and wake behavior.
> 
> **Connect cost:** Adds a **30s PATH command-resolution cache** in shared shell code and **60s memoized editor discovery** on the server so repeat connects avoid huge filesystem scans. **Trace file sink** now clamps oversized span attributes (including `db.query.text`) without mutating live spans.
> 
> **Shell sync:** On the same live RPC session, **foreground resubscribe** resumes from the in-memory `snapshotSequence` instead of always forcing a full HTTP refresh; failed HTTP refresh falls back to a full socket snapshot. New/rewritten shell-sync tests cover these paths.
> 
> **Web:** Sending while the environment is unavailable shows a **warning toast** (draft preserved) instead of silently attaching an annotation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c44dad4598b2a28e1501b74b7ed6e02d2034dfe7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Harden connection resilience by fixing idle disconnects with backoff, pinger tolerance, and shell cursor reuse
> - Changes the `EnvironmentSupervisor` retry ladder first rung from 1s to 3s; a failed foreground wake probe now triggers an immediate reconnect instead of waiting for the first backoff sleep
> - Patches the Effect WebSocket pinger to tolerate two missed pong windows before closing, up from zero tolerance
> - Foreground resubscriptions on the same live session resume from the in-memory shell cursor; new sessions fall back to a full socket snapshot when the HTTP load fails
> - Memoizes editor discovery in `ExternalLauncher` (60s TTL) and command path resolution in `shell.resolveCommandPathForPlatform` (30s TTL) to reduce redundant filesystem scans
> - Behavioral Change: attempting to send a chat message when the environment is unavailable now shows a warning toast instead of silently attaching to draft
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c44dad4.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/connection/environment-cache-store.ts
+++ b/apps/mobile/src/connection/environment-cache-store.ts
@@ -17,7 +17,10 @@ import * as Schema from "effect/Schema";
 import * as MobileDatabase from "../persistence/mobile-database";
 
 const SHELL_SNAPSHOT_CACHE_SCHEMA_VERSION = 1;
-const THREAD_SNAPSHOT_CACHE_SCHEMA_VERSION = 2;
+// v3 adds windowed (paginated) snapshots carrying `page` metadata; the bump
+// makes pre-pagination clients discard the record instead of decoding a
+// partial thread as complete (rollback safety).
+const THREAD_SNAPSHOT_CACHE_SCHEMA_VERSION = 3;
 const SERVER_CONFIG_CACHE_SCHEMA_VERSION = 1;
 const VCS_REFS_CACHE_SCHEMA_VERSION = 1;
 
--- a/apps/mobile/src/features/threads/ThreadDetailScreen.tsx
+++ b/apps/mobile/src/features/threads/ThreadDetailScreen.tsx
@@ -61,6 +61,8 @@ export interface ThreadDetailScreenProps {
   readonly connectionStateLabel: EnvironmentConnectionPhase;
   /** Message sync status for the selected thread (drives the composer status pill). */
   readonly threadSyncStatus?: EnvironmentThreadStatus;
+  /** Non-null when older turns exist beyond the loaded window. */
+  readonly loadEarlier?: { readonly loading: boolean; readonly onLoadEarlier: () => void } | null;
   readonly activeThreadBusy: boolean;
   readonly environmentId: EnvironmentId;
   readonly projectWorkspaceRoot: string | null;
@@ -371,6 +373,7 @@ export const ThreadDetailScreen = memo(function ThreadDetailScreen(props: Thread
             usesAutomaticContentInsets={props.usesAutomaticContentInsets}
             onHeaderMaterialVisibilityChange={props.onHeaderMaterialVisibilityChange}
             skills={selectedProviderSkills}
+            loadEarlier={props.loadEarlier ?? null}
           />
         </View>
       ) : (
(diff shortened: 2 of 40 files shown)


t3dotgg:
Update: rather than a separate PR, I pushed the follow-up fixes directly onto this branch as `6eb16ea5d` (maintainer edits are enabled). #5564 is closed. The commit addresses the remaining review findings: monotonic retry ladder (`[3s, 4s, 8s, 16s]`), recursive trace-attribute truncation with a unit test, the resolution cache moved to a `Context.Reference` per the conventions check, and a documented negative-caching window. @gfsaaser24 feel free to amend or push back on any of it.

gfsaaser24:
@t3dotgg Reviewed `6eb16ea5d` line by line — no pushback, taking all three:

- **Monotonic ladder**: correct call. The `[3s, 2s, …]` wart was an artifact of "raise the floor, don't restructure the ladder" on our side — yours is what it should have been from the start.
- **Recursive truncation**: this caught a real gap. Nested `error.stack`s were bypassing the top-level clamp and quietly eating the retention win we claimed in the PR body. The reference-preserving recursion (returning the same object when nothing clamps) + the immutability test is the right shape for a hot serialization path.
- **`Context.Reference` cache**: the better idiom — matches `SpawnExecutableResolution` one screen up in the same file, and it resolves the global-mutable-state thread without the heavier `Cache.makeWith` layer plumbing.

Regression run against this branch tip, on Windows:

- client-runtime: **579/579** (includes the wake-probe skip, warm-resume, and your updated ladder tests)
- shared: green vs. baseline — the only failures are the 5 pre-existing Windows-environment ones that reproduce identically on clean `main` (`ENAMETOOLONG` semantics + POSIX path/extension assumptions in `relayClient`); your two new truncation tests pass
- typecheck clean across `shared` / `client-runtime` / `web` / `t3`

For pressure testing: the underlying fixes have been soaking on our fork's hosted deployment since yesterday, on the exact workload that motivated this PR — a relay session against a 24-core laptop drowning in agent threads, operated by an impatient user with a phone. Disconnect-loop symptoms gone; steering works mid-run.

Happy to pick up the deferred items (steer-button visual pass, mobile parity) here or in a follow-up, and to file the pinger tolerance upstream on effect.

*Sent from my Fable 5*

PR #5564, opened 2026-08-07
fix: connection-resilience RFC follow-ups (#5561 + review fixes)

t3dotgg:
## What this is

#5561 rebased onto latest main, plus one commit addressing the review pushback that was still standing. The first four commits are @gfsaaser24's work, unchanged. Review just the last commit here; review the rest in #5561.

## Problems fixed on top

- **Non-monotonic retry ladder.** #5561 raised the first backoff rung to 3s but left the second at 2s, so a second consecutive failure retried *faster* than the first. The ladder is now `[3s, 4s, 8s, 16s]`.
- **Nested trace attributes escaped the clamp** (unresolved Macroscope finding). `normalizeJsonValue` preserves nested structures, so an `error.stack` or any deep string was serialized in full past the 500-char limit. Truncation now recurses into arrays and plain objects; new unit test covers depth, the `db.query.text` special case, and input immutability.
- **Module-global cache violated Effect service conventions** (failing check on #5561). The command-resolution cache is now a `Context.Reference` (`CommandResolutionCache`), the same idiom as `HostProcessPlatform` in the same package: default behavior is unchanged (one process-wide map), but tests and embedders can provide an isolated instance through the environment.
- Documented the 30s negative-caching window for "not found" PATH lookups so it's a stated tradeoff rather than a surprise.

## Notes

- If #5561 merges first, this rebases to just the one follow-up commit. If this merges, #5561 can close.
- Still open and deliberately untouched here: the steer-Send button's visual treatment (wants a human eyeball), mobile steering parity, and upstreaming the pinger tolerance to effect.

---
Analysis and changes by Claude Code (Fable 5) with Theo.

---

> [!NOTE]
> **Medium Risk**
> Touches reconnect timing, shell resume cursors, and WebSocket resume gaps—behavior changes on wake and stale clients; caching can briefly hide newly installed PATH binaries (documented 30s window).
> 
> **Overview**
> Follow-up on connection-resilience work: faster recovery when users return to the app, less redundant I/O on connect/resume, and targeted fixes from review.
> 
> **Connection supervisor** raises the first retry delay to **3s** and fixes the ladder to **`[3s, 4s, 8s, 16s]`** so later failures never retry faster than earlier ones. Foreground **wake liveness** failures (or a **3s** probe timeout, down from 15s) trigger an **immediate reconnect** without sleeping the first backoff rung; only that one attempt skips the ladder.
> 
> **Shell sync** on the client **resumes WebSocket subscriptions from the in-memory snapshot cursor** instead of reloading the full shell over HTTP when a warm cache exists; foreground wake **resubscribes at the current sequence**. The server widens **`SHELL_RESUME_MAX_GAP`** from 1,000 to **5,000** so more resumes stay on coalesced replay rather than a full snapshot.
> 
> **Bootstrap cost**: PATH command resolution is **memoized for 30s** via a **`CommandResolutionCache`** `Context.Reference` (testable/isolated); server **editor discovery** is cached for **60s** with a new test. Trace file output **truncates nested strings** (e.g. error stacks) and **`db.query.text`** without mutating live span attributes.
> 
> **Chat UI**: while the agent is **running**, the composer can show **Send** alongside Stop; disconnected environments get an explicit **“not connected”** toast instead of silently attaching annotations. **Effect RPC** patch allows **three missed pings** before treating the connection as dead.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 59e3945661c75ea5b367ada851180ec8df9c9760. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Improve connection resilience with faster reconnects, warm shell resume, and send-while-running
> - WebSocket pinger now tolerates up to two missed pongs before timing out (previously one); connection probe timeout reduced from 15s to 3s; first backoff rung increased from 1s to 3s
> - A failed foreground liveness probe triggers an immediate reconnect attempt, skipping the first backoff rung
> - Shell resubscriptions resume from the in-memory cursor without reloading the HTTP snapshot; `SHELL_RESUME_MAX_GAP` increased from 1,000 to 5,000 events
> - The chat composer now shows a Send button alongside Stop when the agent is running; sending while the environment is unavailable shows a warning toast instead of silently failing
> - `ExternalLauncher.resolveAvailableEditors` and `resolveCommandPathForPlatform` results are now cached (60s and 30s respectively) to reduce repeated filesystem scans
> - Trace span attributes are now truncated before emission, with `db.query.text` clamped more aggressively, to reduce trace sink payload size
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 59e3945. 8 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/apps/server/src/process/externalLauncher.test.ts
+++ b/apps/server/src/process/externalLauncher.test.ts
@@ -7,6 +7,7 @@ import * as Layer from "effect/Layer";
 import * as Path from "effect/Path";
 import * as Sink from "effect/Sink";
 import * as Stream from "effect/Stream";
+import * as TestClock from "effect/testing/TestClock";
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process";
 
 import { HostProcessPlatform } from "@t3tools/shared/hostProcess";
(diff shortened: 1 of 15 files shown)


t3dotgg:
Superseded: the follow-up commit was pushed directly onto #5561's branch instead.

PR #5620, opened 2026-08-07
fix(vcs): surface git's stderr on GitCommandError

IAmJSD:
## Problem

When a git command exits non-zero, `GitCommandError` records `stdoutLength` and `stderrLength` — how much git wrote, never what it wrote. The text is captured by the process runner and then dropped.

That leaves the reason nowhere to be found:

- not in the error (only the lengths),
- not in the RPC response the client receives,
- and not in the server log — `SourceControlRepositoryService` returns clone failures to the caller rather than logging them, so `boot-service.log` stays empty for the entire failure.

The result is that an ordinary, self-explanatory git failure becomes opaque.

## What it looks like today

Cloning onto a headless server whose host has no key registered with the remote. The RPC response, in full:

```json
{"_tag":"SourceControlRepositoryError","provider":"unknown","operation":"cloneRepository",
 "detail":"The source control operation could not be completed.",
 "cause":{"name":"GitCommandError",
          "message":"Git command failed in SourceControlRepositoryService.cloneRepository (/root): Git command exited with a non-zero status."}}
```

Git had already said exactly what was wrong. Recovering it required patching `dist/bin.mjs` on the server to `console.error` the discarded text:

```
[infrawrench-patch] git failed: {"cwd":"/root",
 "args":["clone","git@github.com:owner/repo.git","projects"],"exitCode":128,
 "stderr":"Cloning into 'projects'...\ngit@github.com: Permission denied (publickey).\r\nfatal: Could not read from remote repository.\n"}
```

One line, and the problem is obvious. Getting to it took an afternoon.

## Change

Add an optional `stderrTail` to `GitCommandError` and populate it at both non-zero-exit sites in `GitVcsDriverCore`, then include it in the error's `message` so it reaches anywhere the error is already rendered.

Two details that seemed worth getting right rather than dumping the buffer:

- **Truncated from the end**, not the start. Git puts the reason on its last lines, behind a potentially long transfer log — keeping the head would reliably keep the useless half.
- **Credential-bearing URLs redacted** (`//user:token@host` → `//user:***@host`). Git echoes back the remote it was handed, which may embed a token, and this value now travels into logs and RPC responses.

The 2000-character cap is a judgement call; happy to change it, or to gate the whole field behind a debug flag if you'd rather it not be on by default.

## Notes

- `stderrTail` is `Schema.optional`, so existing clients and any serialized errors are unaffected.
- I have not run the full suite — the monorepo needs a `vp`/pnpm setup I did not want to guess at. The change is confined to one new helper plus two field additions, and I have parse-checked both files. Happy to fix whatever CI says.
- Related, but deliberately not in this PR: `SourceControlRepositoryService` never logs clone failures, and the web client renders a generic string in place of the `detail` the server sends. Either would be a good follow-up; this one is the smallest change that makes the failure knowable at all.

---

> [!NOTE]
> **Medium Risk**
> Changes how git failures are classified and what clients see on RPC errors (auth/remote semantics), though the design intentionally avoids leaking stderr or credentials.
> 
> **Overview**
> Non-zero git exits in `GitVcsDriverCore` are now **classified** from stderr (`authentication`, `not-found`, or `command-failed`) and surfaced on `GitCommandError` via an optional **`failureKind`**, with **`detail`** replaced by fixed, non-secret copy instead of a generic exit message or raw git output.
> 
> Classification runs only when failures are turned into errors (not when `allowNonZeroExit` returns raw output). Those paths set **`LC_ALL=C`** so English heuristics stay reliable; `allowNonZeroExit` keeps the caller’s locale.
> 
> Failed commands log **bounded debug metadata** (kinds and stdout/stderr lengths), not command output. Tests add a stubbed failing-git layer and cover auth vs local “permission denied”, not-found vs “access rights”, locale behavior, and that tokens never appear in `detail`, `message`, or `cause`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc881848a2410be17ed954a4160c74a702ca42d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Surface git stderr in `GitCommandError` with failure classification
> - Adds `classifyGitFailure` in [`GitVcsDriverCore.ts`](https://github.com/pingdotgg/t3code/pull/5620/files#diff-1c3e2b9763c6526006485633fb677c83dc54cc1ea5563a1e51aab707c35854d8) which inspects lowercased stderr to categorize failures as `'authentication'`, `'not-found'`, or `'command-failed'`.
> - Populates a new optional `failureKind` field on `GitCommandError` (in [`git.ts`](https://github.com/pingdotgg/t3code/pull/5620/files#diff-2282301d92a7a9ddc7e8d88b986c9628fdf70497dc6a09fe80eb8d46d6a769e7)) and sets a stable, classification-derived `detail` string instead of a generic non-zero exit message.
> - Forces `LC_ALL=C` when git output will be classified to ensure consistent stderr text for heuristic matching; preserves caller locale when `allowNonZeroExit` is true.
> - Emits a structured debug log for failed commands (operation, exit code, stderr/stdout lengths, `failureKind`) without logging raw output to avoid leaking secrets.
> - Behavioral Change: `GitCommandError.detail` now reflects the failure class (e.g. credential advice) rather than a raw non-zero exit message when classification applies.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dc88184.</sup>
> 
>

### DIFF
--- a/apps/server/src/vcs/GitVcsDriverCore.test.ts
+++ b/apps/server/src/vcs/GitVcsDriverCore.test.ts
@@ -770,19 +770,73 @@ for (const failure of localPermissionFailures) {
   });
 }
 
+it.effect("prefers not-found over the access-rights footer git pairs with it", () => {
+  // GitHub/GitLab SSH for a missing repo writes both lines. Matching
+  // "access rights" first would call that authentication.
+  const failing = makeFailingGitLayer(
+    `ERROR: Repository not found.\n` +
+      `fatal: Could not read from remote repository.\n` +
+      `Please make sure you have the correct access rights\n` +
+      `and the repository exists.\n`,
+  );
+
+  return Effect.scoped(
+    Effect.gen(function* () {
+      const driver = yield* GitVcsDriver.GitVcsDriver;
+      const cwd = yield* makeTmpDir();
+      const error = yield* driver
+        .execute({
+          operation: "GitVcsDriver.test.notFoundWithAccessRights",
+          cwd,
+          args: ["clone", "git@github.com:owner/missing.git", "projects"],
+        })
+        .pipe(Effect.flip);
+
+      assert.equal(failing.spawns.length, 1);
+      assert.instanceOf(error, GitCommandError);
+      assert.equal(error.failureKind, "not-found");
+      assert.notInclude(error.detail.toLowerCase(), "authentication");
+      assert.notInclude(error.detail.toLowerCase(), "credentials");
+    }),
+  ).pipe(Effect.provide(failing.layer));
+});
+
 it.layer(TestLayer)("GitVcsDriver core integration", (it) => {
   describe("process environment", () => {
-    it.effect("preserves the caller locale for general Git subprocesses", () =>
+    it.effect("preserves the caller locale when non-zero exits are returned raw", () =>
       Effect.gen(function* () {
         const cwd = yield* makeTmpDir();
+        const driver = yield* GitVcsDriver.GitVcsDriver;
 
-        const locale = yield* git(
+        // allowNonZeroExit skips classification, so the caller's locale rides
+        // through with the raw stdout/stderr instead of being forced to C.
+        const result = yield* driver.execute({
+          operation: "GitVcsDriver.test.printLocale",
           cwd,
-          ["-c", 'alias.print-locale=!printf "%s" "$LC_ALL"', "print-locale"],
-          { LC_ALL: "zh_CN.UTF-8" },
-        );
+          args: ["-c", 'alias.print-locale=!printf "%s" "$LC_ALL"', "print-locale"],
+          env: { LC_ALL: "zh_CN.UTF-8" },
+          allowNonZeroExit: true,
+          timeoutMs: 10_000,
+        });
 
-        assert.equal(locale, "zh_CN.UTF-8");
+        assert.equal(result.stdout.trim(), "zh_CN.UTF-8");
+      }),
+    );
+
+    it.effect("forces a stable locale when a failure will be classified", () =>
+      Effect.gen(function* () {
+        const cwd = yield* makeTmpDir();
+        const driver = yield* GitVcsDriver.GitVcsDriver;
+
+        const result = yield* driver.execute({
+          operation: "GitVcsDriver.test.printLocaleClassified",
+          cwd,
+          args: ["-c", 'alias.print-locale=!printf "%s" "$LC_ALL"', "print-locale"],
+          env: { LC_ALL: "zh_CN.UTF-8" },
+          timeoutMs: 10_000,
+        });
+
+        assert.equal(result.stdout.trim(), "C");
       }),
     );
   });
--- a/apps/server/src/vcs/GitVcsDriverCore.test.ts
+++ b/apps/server/src/vcs/GitVcsDriverCore.test.ts
@@ -42,6 +42,21 @@ const makeNonRepositoryHandle = () =>
     getOutputFd: () => Stream.empty,
   });
 
+const makeFailingHandle = (stderr: string, exitCode = 128) =>
+  ChildProcessSpawner.makeHandle({
+    pid: ChildProcessSpawner.ProcessId(1),
+    exitCode: Effect.succeed(ChildProcessSpawner.ExitCode(exitCode)),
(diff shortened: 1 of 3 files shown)


IAmJSD:
Thanks — all three are fair, and the second one is the important correction. I've dropped the original approach entirely rather than patch around it.

`stderrTail` and the `message` change are gone; the error's public shape is now identical to `main`. Both non-zero-exit sites instead call `Effect.logDebug` with the command context and its stdout/stderr, so the diagnostic is reachable when someone goes looking without becoming part of the serialized error.

Specifically on each point:

1. **Raw output as a direct attribute** — agreed, and I hadn't found #3253. Reintroducing it also regressed the `GitVcsDriverCore.test.ts` case asserting a secret passed as a git argument never reaches `error.message`, which I should have caught before opening this.
2. **`message` from structural attributes only** — reverted; the getter is untouched.
3. **Redaction insufficient** — correct, and I think it argues against sanitizing into a structured field at all rather than for a better regex. Single-token remotes, query-string tokens, and hook output all defeat it, and "redacted" on a field that isn't reliably redacted is worse than not having the field.

The remaining change is the smallest thing that makes the failure knowable: with the error keeping only lengths and clone failures returned to the caller rather than logged, git's reason currently exists nowhere. Debug level felt like the right default given the content can carry credentials — happy to move it behind an explicit diagnostics flag instead if you'd prefer it off by default.

IAmJSD:
> Thanks — all three are fair, and the second one is the important correction. I've dropped the original approach entirely rather than patch around it.
> 
> `stderrTail` and the `message` change are gone; the error's public shape is now identical to `main`. Both non-zero-exit sites instead call `Effect.logDebug` with the command context and its stdout/stderr, so the diagnostic is reachable when someone goes looking without becoming part of the serialized error.
> 
> Specifically on each point:
> 
>     1. **Raw output as a direct attribute** — agreed, and I hadn't found [[codex] Enrich Git VCS driver errors #3253](https://github.com/pingdotgg/t3code/pull/3253). Reintroducing it also regressed the `GitVcsDriverCore.test.ts` case asserting a secret passed as a git argument never reaches `error.message`, which I should have caught before opening this.
> 
>     2. **`message` from structural attributes only** — reverted; the getter is untouched.
> 
>     3. **Redaction insufficient** — correct, and I think it argues against sanitizing into a structured field at all rather than for a better regex. Single-token remotes, query-string tokens, and hook output all defeat it, and "redacted" on a field that isn't reliably redacted is worse than not having the field.
> 
> 
> The remaining change is the smallest thing that makes the failure knowable: with the error keeping only lengths and clone failures returned to the caller rather than logged, git's reason currently exists nowhere. Debug level felt like the right default given the content can carry credentials — happy to move it behind an explicit diagnostics flag instead if you'd prefer it off by default.

^ this was claude lol

IAmJSD:
Fixed — the log payload is now a normalized category plus lengths, and the exact output moved to `cause`.

The annotation reuses the vocabulary `VcsProcess.classifyNonZeroExit` already established (`authentication` / `not-found` / `command-failed`), so a log line says what *kind* of failure it was without carrying the text that says so. You were right that the previous revision just relocated the problem: `stdout`/`stderr` are capped only by `maxOutputBytes`, which is megabytes at some call sites, so "bounded" was never true of that payload.

`GitCommandError`'s direct attributes are untouched, so the existing assertions still hold — a secret passed as a git argument reaches neither `error.message` nor an `stderr` property. `isMissingGitCwdError` is also unaffected: it guards on `cause instanceof PlatformError`, which a string fails exactly as the previous `undefined` did.

Net effect is now three lines of behaviour: a debug log with a category and counts at each non-zero-exit site, and the output on `cause` where it can be inspected without crossing a boundary as a direct attribute.

t3dotgg:
> [!NOTE]
>

Closing in favor of [#8645](https://github.com/pingdotgg/t3code/pull/8645). It gives specific failure reasons for Git commands and excludes hook output from classification while keeping raw stderr private. This branch now uses broader categories, despite the original title. Review continues on the retained PR, with the classification-only locale behavior and its tests recorded there.

PR #5700, opened 2026-08-08
fix(usage): keep the richest copy of a deduped usage record

xkelxmc:
## What changed

Usage de-duplication (within-file and across the scan) now keeps the **richest** copy of a dedupe key instead of the first one encountered, replacing a poorer copy's contribution exactly (tokens, cost, cache savings, record counts). Identical copies still keep the first, so the content-block-repeat case this rule was written for behaves as before.

- `usageScanCache.ts` — `dedupeWithinFile` keeps the copy with the most tokens, in its original position; `USAGE_SCAN_CACHE_VERSION` bumped to 2 so entries holding first-seen copies are re-parsed rather than reused.
- `usageAggregation.ts` — `#seen` grows from a `Set` of keys to a `Map` of contributions, so a richer copy arriving later can retract the poorer copy from its bucket and take its place.

## Why it should exist

Copies sharing a `message.id:requestId` key are not always identical. Claude Code writes intermediate snapshots of a streamed message, and the earlier copies carry a partial `output_tokens` — the copy with the most tokens is the message as billed. First-wins keeps whichever copy the filesystem walk met first, which is frequently a partial one.

On our transcripts this is not a rounding error: comparing both strategies over the same one-week window, first-wins lost roughly **half of all output tokens** (input, cache-read and cache-write totals were identical), which understates Claude cost by a few percent — output is the most expensive channel.

## Notes

- `duplicatesDropped` keeps its meaning: the superseded poorer copy counts as the dropped one.
- Sessions are deliberately not retracted; a superseded copy is the same message from the same session.
- Without the cache version bump the fix appears to do nothing on warm installs — the scan cache stores records already deduped within their file.

Tests cover both arrival orders (partial→finished and finished→partial), the identical-copies case, and the cost retraction.

---

> [!NOTE]
> **Medium Risk**
> Changes how usage and cost are aggregated for billing dashboards; scope is limited to usage parsing/cache with targeted tests and a forced cache refresh.
> 
> **Overview**
> **Usage de-duplication now keeps the highest-token copy per dedupe key** instead of the first seen, so Claude Code streaming snapshots (same key, growing `output_tokens`) bill against the finished message rather than a partial write.
> 
> `dedupeWithinFile` in `usageScanCache.ts` compares `totalTokens` and replaces a held row when a later duplicate is richer; identical duplicates still keep the first. **`USAGE_SCAN_CACHE_VERSION` is bumped to 2** so v1 cache entries (first-seen partials) are discarded and files are re-parsed.
> 
> Tests cover within-file and aggregator paths for partial→finished and finished→partial order, identical copies, and cost retraction when a partial is superseded.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 052e18e9202b3110aa1b2469bdb24cf7af8a35f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix usage deduplication to keep the richest copy of a duplicate record
> - Rewrites `dedupeWithinFile` in [usageScanCache.ts](https://github.com/pingdotgg/t3code/pull/5700/files#diff-ca3f710ebaf0cd708cedab20134ea19de9a81237223250d0402fa68c37498806) to track the position of the first occurrence of each dedupe key, replacing it only if a later record has more total tokens.
> - Updates aggregation logic in [usageAggregation.ts](https://github.com/pingdotgg/t3code/pull/5700/files#diff-fcaad7aef8d1259b6597c6bb7b20ca5418bd61cc8ca843778583e3de2616bae0) to apply the same richest-copy preference when folding duplicates across buckets.
> - Increments `USAGE_SCAN_CACHE_VERSION` from 1 to 2, invalidating existing cache entries so they are re-parsed with the new deduplication semantics.
> - Behavioral Change: previously the first-seen duplicate was always kept; now a later record with more tokens replaces it at its original position.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 052e18e.</sup>
> 
>


caezium:
Could duplicate usage be merged by per-field maxima instead of selecting the record with the largest overall token total? Streaming snapshots can advance input, output, cache-read, 5m cache-write, and 1h cache-write independently, so a single “richest” record can discard a maximum from another snapshot. A crossing-fields regression test would capture this.

PR #5711, opened 2026-08-08
fix(server): cap background-work deferral and report session-teardown losses

gfsaaser24:
## What Changed

Completes #5677. The background-liveness skip gets a wedge cap (`backgroundWorkMaxIdleMs`, default 4h) so a task that never reaches a terminal state cannot pin its session and child process forever. The binding's `lastSeenAt` now refreshes on provider-initiated turn/task lifecycle events, not just user `sendTurn`. `session.exited` flowing through the event pump marks the persisted binding `stopped` — guarded against stale exits from a replaced session, including replacements on a different provider instance. Claude session teardown drains still-live task ids with terminal `task.completed(stopped)` events, and a stream death with live tasks emits a `runtime.error` naming the lost agents plus a structured log.

## Why

#5677 correctly stopped the reaper from killing sessions with live background work, but the edges around it are still lossy or unbounded. The deferral is unconditional, so the failure mode just flips: a wedged task now pins its session forever instead of being killed too early. `lastSeenAt` still only moves on user-initiated turns, so a thread doing autonomous work reads as idle from the user's last keystroke — background liveness defers the reap, but the idle clock itself stays dishonest for anything else keyed to it. Adapter-internal exits (stream end, session replace, `stopAll`) never pass through `stopSession` and leave ghost `running` rows the sweep re-visits forever. And when teardown does kill live agents, the user finds out on their next message, potentially 40+ minutes later.

This was diagnosed from a production day with 13 session stops across three threads, each landing 30.1–39.7 minutes after the last user message while background agents were mid-work — one session was killed while running the test suite that verifies this fix. With the heartbeat in place, the wedge cap means "no heartbeat at all for the cap window", which is the definition of wedged.

The sweep reads `ThreadBackgroundLivenessService` directly rather than the projected shell — the shell's `backgroundLiveness` is computed from the same service at snapshot-build time, so it is the same signal minus the snapshot staleness. #5677's test is kept and adapted to the capped deferral: its months-stale scenario now correctly reaps, and that boundary is pinned by its own wedge-cap test. The stale-exit guard's cross-instance test fails without the guard (verified by reverting and re-running). `ProviderSessionReaper` + `ProviderService` + `ClaudeAdapter`: 110/110 on this base; `ProviderRuntimeIngestion`: 45/45; typecheck clean.

Bigger than the contributing guide prefers — it stays one PR because the four pieces enforce one invariant (background work is neither silently killed nor allowed to pin a session, and its loss is visible), and the heartbeat is what makes the cap meaningful. Happy to split it if you'd rather take it piecewise.

## Checklist

- [ ] This PR is small and focused (focused: one subsystem, one invariant — but not small, see note above)
- [ ] ~~I included before/after screenshots for any UI changes~~ (no UI changes)
- [ ] ~~I included a video for animation/interaction changes~~ (n/a)

---
Change authored by Claude Fable 5 via Claude Code.

---

> [!NOTE]
> **Medium Risk**
> Changes core provider session teardown, persisted binding sync, and reaper heuristics; mistakes could leave ghost running sessions or reap live autonomous work, though extensive guards and tests target the production failure modes.
> 
> **Overview**
> Tightens **background-work session lifecycle** so autonomous provider work is neither silently killed nor allowed to pin processes forever, and teardown losses are visible.
> 
> **Claude adapter:** On SDK stream exit (clean or failed) while `liveTaskIds` remain, the adapter now logs `claude.session.stream-ended-with-live-tasks`, emits a **`runtime.error`** naming the lost agents, and drains each live task with **`task.completed` (`stopped`)**, including a second drain after stream-fiber interrupt to close races with in-flight handlers.
> 
> **Provider event pump:** Adds **`syncBindingOnRuntimeEvent`** so **`session.exited`** marks the persisted binding **`stopped`** (clearing ghost `running` rows from adapter-internal exits), with guards for stale exits (newer replacement session on the same instance, or binding owned by a different provider instance). Turn/task lifecycle events **refresh `lastSeenAt`** (60s throttle) so provider-initiated work does not look idle since the user’s last `sendTurn`.
> 
> **Session reaper:** Defers reaping idle sessions while **`ThreadBackgroundLivenessService`** reports live work, but only up to **`backgroundWorkMaxIdleMs`** (default 4h, floored above the inactivity threshold); beyond that, wedged “live” work is reaped anyway. Sweep uses the liveness service directly instead of the projected shell.
> 
> Tests and **`docs/internals/providers.md`** document the three rules (teardown reports loss, binding follows session, reaper respects background work with a cap).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5aeeb1b52986c4aa5737bda0a2cb2b3d43c8bfd4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Cap background-work deferral in session reaper and report live-task losses on session teardown
> - The Claude adapter's `handleStreamExit` now detects background tasks still tracked as live at stream termination, emits a `runtime.error` describing the loss, and emits `task.completed` (status `stopped`) for each live task. A structured warning is also logged.
> - The `ProviderSessionReaper` defers reaping idle sessions while background work is live, up to a configurable `backgroundWorkMaxIdleMs` cap (default 4h), after which the session is reaped regardless.
> - The `ProviderService` event pump now syncs persisted bindings: marks a binding `stopped` on `session.exited` (guarded against stale cross-instance or replaced-session exits), and refreshes `lastSeenAt` on turn/task lifecycle events with per-thread 60s throttling.
> - Tests and [providers.md](https://github.com/pingdotgg/t3code/pull/5711/files#diff-3d909027abc31a863f0c0e5ecef5aa50ac285620ad827db723eb0a4ac46d3f85) documentation are added for all three behaviors.
> - Behavioral Change: sessions with live background work that exceed the idle wedge cap are now reaped unconditionally, even if liveness still reports active.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 5aeeb1b. 4 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
+++ b/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
@@ -1828,6 +1828,153 @@ describe("ClaudeAdapterLive", () => {
     );
   });
 
+  it.effect("reports lost background agents when the stream ends cleanly", () => {
+    const harness = makeHarness();
+    return Effect.gen(function* () {
+      const adapter = yield* ClaudeAdapter;
+      const runtimeEvents: Array<ProviderRuntimeEvent> = [];
+      const runtimeEventsFiber = yield* Stream.runForEach(adapter.streamEvents, (event) =>
+        Effect.sync(() => {
+          runtimeEvents.push(event);
+        }),
+      ).pipe(Effect.forkChild);
+
+      yield* adapter.startSession({
+        threadId: THREAD_ID,
+        provider: ProviderDriverKind.make("claudeAgent"),
+        runtimeMode: "full-access",
+      });
+      yield* adapter.sendTurn({
+        threadId: THREAD_ID,
+        input: "spawn agents",
+        attachments: [],
+      });
+
+      harness.query.emit({
+        type: "system",
+        subtype: "task_started",
+        task_id: "task-lost-a",
+        description: "Pass 5: fix-wave regression",
+        task_type: "local_agent",
+        tool_use_id: "toolu_lost_a",
+        uuid: "task-lost-a-uuid",
+        session_id: "sdk-session",
+      } as unknown as SDKMessage);
+      harness.query.emit({
+        type: "system",
+        subtype: "task_started",
+        task_id: "task-lost-b",
+        description: "Pass 5: blind feature sweep",
+        task_type: "local_agent",
+        tool_use_id: "toolu_lost_b",
+        uuid: "task-lost-b-uuid",
+        session_id: "sdk-session",
+      } as unknown as SDKMessage);
+
+      // The claude child process exiting ends the SDK iterable cleanly.
+      // This is the incident shape: background agents still live, no error.
+      harness.query.finish();
+
+      yield* Effect.yieldNow;
+      yield* Effect.yieldNow;
+      yield* Effect.yieldNow;
+      runtimeEventsFiber.interruptUnsafe();
+
+      const runtimeError = runtimeEvents.find((event) => event.type === "runtime.error");
+      assert.equal(runtimeError?.type, "runtime.error");
+      if (runtimeError?.type === "runtime.error") {
+        assert.match(runtimeError.payload.message, /2 background agents were still running/);
+        assert.match(runtimeError.payload.message, /Pass 5: fix-wave regression/);
+        assert.match(runtimeError.payload.message, /Pass 5: blind feature sweep/);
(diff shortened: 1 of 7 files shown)


juliusmarminge:

PR #5751, opened 2026-08-08
fix(ssh): preserve external runtime ownership

### LINKED ISSUE #5749: SSH reconnect can replace an external T3 service with a competing managed runtime
## Problem

On a persistent SSH host that already runs the supported T3 systemd service, reconnecting from an updated desktop client can start a second T3 server and a second T3 Connect tunnel against the same base directory.

The failure is deterministic when launcher state says `managed` while `server-runtime.json` belongs to the launcher-managed process, or when a recorded external service is briefly unavailable during its own update. Requests and WebSockets can then land on different servers, producing intermittent reconnects and messages that appear to disappear.

## Expected behavior

The supported external service remains the sole server/tunnel owner. The SSH launcher adopts its advertised port, records external ownership, and never replaces a temporarily restarting external service.

## Proposed fix

The tested source change:

- distinguishes the launcher's own managed runtime record from a real external runtime;
- makes runtime-state cleanup conditional on PID/start-time ownership;
- preserves external ownership across transient unavailability; and
- adds regression coverage for adoption, reconnect, restart, and stale cleanup.

A focused PR follows.

mikeyperes:
Closes #5749.

## What changed

- Distinguish the SSH launcher's own managed runtime record from an externally owned T3 service.
- Adopt a healthy external service without first stopping it.
- Keep external ownership during a transient service restart instead of launching a competing server.
- Clear `server-runtime.json` only when the shutting-down process still owns that exact PID/start-time record.
- Document the persistent-host ownership contract.

## Why

Two T3 servers and two relay tunnels sharing one base directory make reconnect behavior nondeterministic: HTTP requests and WebSockets can reach different processes. This keeps the supported persistent service as the single owner.

## Verification

- `packages/ssh/src/tunnel.test.ts`
- `apps/server/src/serverRuntimeState.test.ts`
- Focused combined reliability suite: 82/82 passing
- Server typecheck passing
- Production server bundle build passing

---

> [!NOTE]
> **Medium Risk**
> Changes SSH launch and persisted runtime discovery on shutdown, including atomic rename/restore races; wrong behavior could leave no server or duplicate servers on one data dir.
> 
> **Overview**
> Keeps a single T3 server owner when SSH reconnects to a host that already runs a persistent service, and stops shutdown from wiping a replacement process’s `server-runtime.json`.
> 
> **SSH remote launch** now treats `server-runtime.json` as the source of truth for an **externally owned** runtime: it reconnects without stopping that process, marks SSH state as `external`, and only tears down a prior **managed** SSH-launched PID when the default runtime port changed. If external ownership is recorded but the server is not ready, launch **fails** instead of starting a competing managed server. Competing managed state converges onto the external owner (including PID-reuse edge cases).
> 
> **Server shutdown** persists runtime state on startup and, on release, calls **`clearPersistedServerRuntimeStateIfOwned`** with that snapshot so only the matching `pid` + `startedAt` record is removed; mismatches are restored via rename/link (with a no-hard-link fallback). Docs note that SSH will not take ownership while an external service is active.
> 
> Tests cover ownership clearing, rename races, link fallback, and end-to-end SSH launch script behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19b6923bf00fdc07d92175a7f16b2ea6ca3e6449. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Preserve external SSH runtime ownership on server shutdown and reconnect
> - Replaces the unconditional `clearPersistedServerRuntimeState` call with `clearPersistedServerRuntimeStateIfOwned`, which atomically renames the state file and only deletes it if the current process's `pid` and `startedAt` match the captured file.
> - If the state is owned by another process, the original file is restored via hard link or exclusive-write fallback to avoid clobbering a replacement written during a race.
> - The SSH launch script in [`tunnel.ts`](https://github.com/pingdotgg/t3code/pull/5751/files#diff-7338be65675ab0699d0c8c6426e54f3421747885231f9bbb0ace4ec129cb2072) now detects externally owned runtimes, reuses them without replacement, avoids killing unrelated processes on port switch, and exits fast with an explanatory message if the external server is not ready.
> - Behavioral Change: servers that did not previously own their runtime state will no longer clear it on shutdown; the SSH launcher will refuse to start a competing server when an external owner is detected.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 19b6923.</sup>
> 
>

### DIFF
--- a/apps/server/src/server.ts
+++ b/apps/server/src/server.ts
@@ -104,7 +104,7 @@ import * as ResourceTelemetry from "./resourceTelemetry/ResourceTelemetry.ts";
 import * as UsageService from "./usage/UsageService.ts";
 import { OrchestrationLayerLive } from "./orchestration/runtimeLayer.ts";
 import {
-  clearPersistedServerRuntimeState,
+  clearPersistedServerRuntimeStateIfOwned,
   makePersistedServerRuntimeState,
   persistServerRuntimeState,
 } from "./serverRuntimeState.ts";
@@ -496,13 +496,19 @@ export const makeServerLayer = Layer.unwrap(
               Effect.logWarning("Failed to persist server runtime state", { cause }),
             ),
           );
+          return state;
         }),
-        () =>
-          clearPersistedServerRuntimeState(config.serverRuntimeStatePath).pipe(
-            Effect.catchCause((cause) =>
-              Effect.logWarning("Failed to clear server runtime state", { cause }),
-            ),
-          ),
+        (state) =>
+          state === undefined
+            ? Effect.void
+            : clearPersistedServerRuntimeStateIfOwned({
+                path: config.serverRuntimeStatePath,
+                state,
+              }).pipe(
+                Effect.catchCause((cause) =>
+                  Effect.logWarning("Failed to clear server runtime state", { cause }),
+                ),
+              ),
       ),
     );
     const tailscaleServeLayer = config.tailscaleServeEnabled
--- a/apps/server/src/serverRuntimeState.test.ts
+++ b/apps/server/src/serverRuntimeState.test.ts
@@ -214,6 +214,44 @@ describe("serverRuntimeState", () => {
     }).pipe(Effect.provide(NodeServices.layer)),
   );
 
+  it.effect("restores replacement state when the filesystem rejects hard links", () =>
+    Effect.gen(function* () {
+      const fileSystem = yield* FileSystem.FileSystem;
+      const path = yield* Path.Path;
+      const root = yield* fileSystem.makeTempDirectoryScoped({
+        prefix: "t3-server-runtime-state-link-fallback-test-",
+      });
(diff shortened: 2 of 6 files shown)


mikeyperes:
Canonical combined source is now published at https://github.com/mikeyperes/hws-t3code/commit/e383737a30218a77de69474c918f43a0b0fc2b76, with the canonical pickup record at https://github.com/mikeyperes/hws-agent-operations/blob/main/T3-AUTOMATIC-RECOVERY-HANDOFF.md.

The branch carries the SSH owner-convergence and update-scope fixes and has source-level regression coverage. It is not a supported release artifact; no live deployment has occurred and server 236 remains red (public relay HTTP 502).

PR #4825, opened 2026-07-29
fix(desktop): increase backend readiness timeout to 3 minutes

carlospedreira:
## Summary

Increases `DEFAULT_BACKEND_READINESS_TIMEOUT` in `apps/desktop/src/backend/DesktopBackendManager.ts` from 1 minute to 3 minutes.

Fixes #4535

## Why

When the packaged server bundle is loaded through `/mnt/c`, WSL backend startup can take well over the current 60-second readiness window (~117 seconds measured on my system — see [this comment](https://github.com/pingdotgg/t3code/issues/4535#issuecomment-5107209213)). The backend eventually becomes healthy, but the desktop app has already stopped probing and stays stuck on "Connecting to WSL…" forever.

This only raises the **maximum** wait: readiness polling still resolves as soon as the endpoint responds, so fast backends connect immediately as before. Precedent: #1979 previously raised this timeout from 10s to 30s, and other remote backend paths already use three-minute startup/readiness windows.

Note this is a pragmatic mitigation — it does not address the underlying `/mnt/c` filesystem performance problem for the WSL bundle.

## Verification

- Reproduced the timeout locally on WSL and confirmed T3 Code connects successfully to the WSL backend with this change (backend became ready at ~117s).
- `pnpm typecheck` (apps/desktop) passes.
- `pnpm test src/backend/DesktopBackendManager.test.ts` — 12 tests pass.
- Lint and format checks pass on the changed file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Single constant change to startup polling duration with no auth, data, or behavioral changes beyond tolerating slower backend boot.
> 
> **Overview**
> Raises **`DEFAULT_BACKEND_READINESS_TIMEOUT`** in `DesktopBackendManager` from **1 minute to 3 minutes** so the desktop app keeps polling `/.well-known/t3/environment` long enough for slow WSL backends (e.g. server bundle on `/mnt/c`) that can take well over 60s to become ready.
> 
> Polling still ends as soon as the endpoint responds, so fast local backends connect immediately; only the maximum wait is longer.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae3d2f1fc70435abab48494ad3077cf17ebcf3fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
[...]

PR #5769, opened 2026-08-09
fix(desktop): Cache Runtime locally on WSL Filesystem, dramatically improving launch times 

ikifar2012:
## What Changed

- packages a Linux-only WSL server runtime beside the Windows desktop app
- verifies the archive SHA-256 before installing it under `~/.t3/wsl-runtime`
- reuses warm installs, repairs corrupted entries, and safely prunes old or invalid caches
- preserves active and concurrently installing runtimes
- removes the legacy Windows extraction tree after the distro-local runtime passes preflight
- keeps the mounted `server.asar` tree as a fallback
- shares one packaged WSL test harness across the seven cache and fallback cases
- normalizes Windows bsdtar CRLF member listings before payload validation
- invalidates legacy completion markers before best-effort recursive cleanup

## Why

Launching the backend from WSL's Linux filesystem avoids repeated `/mnt/c` reads. The cache now has a desktop-owned directory and one verified archive identity, while the build omits Windows/Darwin native payloads and pnpm install metadata that WSL cannot use.

## Verification

- exact head: 180 focused and 819 broader desktop/scripts tests
- synthetic current main: 180 focused and 819 broader desktop/scripts tests
- desktop and scripts typechecks
- lint, formatting, diff checks, and release smoke
- executed POSIX-shell coverage for install locking, archive refusal, corruption recovery, and pruning

https://github.com/user-attachments/assets/9ae39ab7-c013-4510-b376-a9dcf0a5be62

---

> [!NOTE]
> **Medium Risk**
> Reworks packaged Windows WSL backend startup, cache identity, and fallback ordering; mis-handled transient vs fatal preflight outcomes could affect launch reliability despite extensive tests.
> 
> **Overview**
> Packaged Windows builds now ship **`wsl-runtime.tar.gz`** plus a **SHA-256 sidecar** (when a Linux node-pty prebuild is bundled). WSL resolution **prefers** installing that archive into **`~/.t3/wsl-runtime/sha256-<digest>`** via new **`DesktopWslEnvironment`** APIs (`prepareRuntime`, `invalidateRuntime`, `pruneRuntimes`) and generated install/prune shell scripts (locking, digest checks, readiness markers, in-use detection).
> 
> **`runWslPreflight`** no longer always materializes **`server.asar`** up front: it stages and probes the cached runtime first, runs **legacy Windows extraction cleanup** on success, and **falls back** to the mounted `/mnt` server tree when staging fails or a fatal staged **node-pty** probe fails (invalidating the bad cache when the fallback works). **`DesktopBackendManager`** records **`wslRuntimeId`** on the start config and **prunes older caches** after HTTP readiness.
> 
> The **desktop artifact builder** stages the Linux-only archive (excludes Windows/Darwin natives and pnpm metadata), validates archive/hash/members in packaged payloads, and docs describe the new WSL install behavior. Tests cover configuration fallbacks, manager pruning, server-tree cleanup, script generation, and executed POSIX install races.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cee8a4a10878503a846bb4bdd2f25c2057d6e8c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Cache WSL runtime on WSL filesystem to improve desktop launch times
> - Ships a compressed Linux-only WSL runtime archive (`wsl-runtime.tar.gz` + `.sha256` sidecar) as a Windows extraResource; [build-desktop-artifact.ts](https://github.com/pingdotgg/t3code/pull/5769/files#diff-384c39593c6da8887b4f03c00f763af0e3b689525d3555a9fe03fd05c808471d) stages, hashes, and validates it during the Windows build
> - `runWslPreflight` in [DesktopBackendConfiguration.ts](https://github.com/pingdotgg/t3code/pull/5769/files#diff-8a9cd9d84f556b82ed8678110af5a99dea2cc1cfe1e6c6c46dc3a2e3bcb1f94c) now stages the packaged archive into a distro-local cache via `DesktopWslEnvironment.prepareRuntime`, probes it, and falls back to the mounted server tree only when staging fails fatally
> - [DesktopWslEnvironment.ts](https://github.com/pingdotgg/t3code/pull/5769/files#diff-eb443cdb82c5c1542b51d3b5bd796cd760f44161c7cd5f8ead2affd539c22d1d) gains `prepareRuntime`, `pruneRuntimes`, and `invalidateRuntime`; the install script uses POSIX `flock`, SHA-256 digest verification, atomic promotion, and in-use safety; the prune script retains the current and one previous cache
> - After backend readiness, `makeBackendInstance` in [DesktopBackendManager.ts](https://github.com/pingdotgg/t3code/pull/5769/files#diff-45848e60bcec64dd7239fc34379d23407c47d0ade3f351e89a4efc7f6b42757f) calls `pruneRuntimes` to remove older caches; `cleanupLegacy` in [DesktopWslServerTree.ts](https://github.com/pingdotgg/t3code/pull/5769/files#diff-0a2e8afaef8f3ef0d11e204ba9df65ad04378970dbe26e7bfee17313b585647a) removes the old Windows extraction cache once the staged runtime is healthy
> - Risk: `ensureNodePty` now expects a `linuxAppRoot` path instead of converting a Windows path internally; any caller still passing `windowsRepoRoot` will break. WSL launches that previously always used the mounted tree now depend on a staged cache and may fall back if the archive is missing or corrupt
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dc98770.</sup>
> 
>

## request provenance

- requested by: @UtkarshUsername


moinmir:
One idea: key the cache on a hash of the archive instead of `<version>-<arch>`. A release that does not change the server would then reuse the existing directory in `~/.t3/runtime`.

carlospedreira:
This is actually a better solution than the one that I'm proposing on #4825 because it's actually fixing the issue rather than just increasing the wait time. Happy to close my PR and support this one instead.

apps/desktop/src/backend/DesktopBackendConfiguration.ts
```diff
@@ -288,27 +289,28 @@ const runWslPreflight = Effect.fn("desktop.backendConfiguration.wslPreflight")(f
...
-      reason: `missing server entry at ${input.windowsEntryPath}`,
+      reason: `missing WSL runtime source at ${input.windowsSourcePath}`,
       fatal: true,
     } as const;
   }
 
-  const linuxEntry = yield* wslEnv.windowsToWslPath(runningDistro, input.windowsEntryPath);
-  if (Option.isNone(linuxEntry)) {
+  const runtime = yield* wslEnv.prepareRuntime(runningDistro, input.runtimeSource);
+  if (!runtime.ok) {
```
moinmir: A staging failure here removes WSL mode entirely: a non-ok result becomes a fatal preflight failure, and after `MAX_PREFLIGHT_FAILURE_ATTEMPTS` the manager shows the failure dialog and persists Windows mode. So a full disk, a corrupt archive, or an unusual distro ejects users who launched fine (just slowly) before this PR.

Could the archive be an optimization instead of a requirement? On failure, log a warning and fall back to the mounted `app.asar.unpacked` root that older packages already use. Worst case is today's launch time. That would also let `WslRuntimeSource` and the `windowsSourcePath` juggling collapse, which shrinks the diff.

apps/desktop/src/wsl/DesktopWslEnvironment.ts
```diff
@@ -216,6 +253,96 @@ const runWslShell = (
...
+    "trap 'exit 1' HUP INT TERM",
+    'while ! mv -T "$runtime_lock_candidate" "$runtime_lock" 2>/dev/null; do',
+    "  if runtime_is_ready; then",
+    `    printf 'runtimeRoot:%s\\n' "$runtime_root"`,
+    "    exit 0",
+    "  fi",
+    '  runtime_lock_pid=$(cat "$runtime_lock/pid" 2>/dev/null || true)',
+    '  case "$runtime_lock_pid" in',
+    "    ''|*[!0-9]*) runtime_lock_owner_alive=false ;;",
+    '    *) if kill -0 "$runtime_lock_pid" 2>/dev/null; then runtime_lock_owner_alive=true; else runtime_lock_owner_alive=false; fi ;;',
```
moinmir: PIDs do not survive the distro restarting. If an install dies mid-extract or `wsl --shutdown` hits, the lock dir stays on disk, and after restart `kill -0` can find an unrelated live process. Then every launch busy-waits to the full 2 minute timeout, up to 12 retries.

`flock` on a lock file has no stale state; the kernel releases it when the fd closes or the VM goes away:

```sh
exec 9>"$runtime_parent/.${safeRuntimeId}.install.lock"
flock -x 9
```

That also removes the pid file, the stale-lock dance, and the sleep loop.

apps/desktop/src/wsl/DesktopWslEnvironment.ts
```diff
@@ -216,6 +253,96 @@ const runWslShell = (
...
+    '  if [ "$runtime_lock_owner_alive" = false ]; then',
+    `    runtime_lock_stale=$(mktemp -d "$runtime_parent/.${safeRuntimeId}.stale-lock.XXXXXX")`,
+    '    rmdir "$runtime_lock_stale"',
+    '    if mv -T "$runtime_lock" "$runtime_lock_stale" 2>/dev/null; then',
+    '      rm -rf "$runtime_lock_stale"',
+    "    fi",
+    "  fi",
+    "  sleep 0.1",
+    "done",
+    'cleanup_runtime_lock() { rm -rf "$runtime_lock_candidate" "$runtime_lock"; }',
```
moinmir: `mv -T` already consumed `$runtime_lock_candidate` by the time this trap can run, so that path is always gone here.

```suggestion
    'cleanup_runtime_lock() { rm -rf "$runtime_lock"; }',
```

docs/user/install.md
```diff
@@ -41,6 +41,17 @@ Arch Linux:
...
 ```
 
+### Windows Subsystem for Linux
+
+When the desktop app runs a WSL backend, it installs the matching server runtime into
+`~/.t3/runtime` inside the selected distro. The first launch after installing or updating T3 Code
+may take a little longer while that runtime is extracted. Later launches reuse the Linux-local
+copy so startup does not depend on reading application files through `/mnt/c`.
+
+Each T3 Code version uses its own runtime directory. Removing an old version's directory is safe
```
moinmir: Each of these directories holds a full `apps/server/dist` plus `node_modules`, and most users will never find this paragraph. Could the app prune old runtimes itself after a successful launch? Keeping the current version plus one previous still covers rollback, and this section gets shorter.

apps/desktop/src/wsl/DesktopWslEnvironment.test.ts
```diff
@@ -10,6 +10,7 @@ import * as TestClock from "effect/testing/TestClock";
 import { ChildProcessSpawner } from "effect/unstable/process";
 
 import {
+  buildWslRuntimeInstallScript,
   buildWslNodeEnvPreamble,
```
moinmir, line 16: Ordering nit, the rest of this list is alphabetical.

```suggestion
  buildWslNodeEnvPreamble,
  buildWslRuntimeInstallScript,
```

apps/desktop/src/wsl/DesktopWslEnvironment.test.ts
```diff
@@ -125,6 +128,59 @@ describe("buildWslNodeEnvPreamble", () => {
   });
 });
 
+describe("WSL runtime cache", () => {
+  it("sanitizes cache ids before interpolating them into Linux paths", () => {
+    expect(sanitizeWslRuntimeId("1.2.3/x64; touch /tmp/nope")).toBe("1.2.3_x64__touch__tmp_nope");
+  });
+
+  it("installs through a temporary directory and only reuses valid completed caches", () => {
```
moinmir, line 138: These assertions pin the script's literal text, so a reformat breaks them, and none of them would catch a real locking or promotion bug. The script is plain POSIX sh and this package's tests run on Linux in CI. Could the test execute it instead? Point `HOME` at a temp dir, hand it a small fixture tar.gz, and assert the outcomes: marker exists, directory promoted, two concurrent runs extract once, and a failed `tar` leaves no half-built cache.

moinmir (commented):
I was working on a similar PR and saw this created. The solution looks correct. I ran an AI review comparing this PR to my solution and found some issues that I'm posting here. I don't understand the code 100% but these seem reasonable. Feel free to disregard if they are not useful.

ikifar2012:
Asked my agent to implement some of the changes suggested by @moinmir, all seems to work well, waiting on other pending checks on github but I guess I can get to those tomorrow, if the PR agents have any more suggestions... looks like there are a bunch of other WSL PR's as well so idk who's will get merged, after this I gotta see if there's anything I can do about install speed because it takes forever

ReyJ94:
https://github.com/pingdotgg/t3code/pull/5042 strange , i also have a similar solution here, and nothing is merged, despite theo pointing people at this issue and specifically at my pr. I'm using a sidecar just to be able to use t3code lol. I hope one of these solutions merges soon

ikifar2012:
> #5042 strange , i also have a similar solution here, and nothing is merged, despite theo pointing people at this issue and specifically at my pr. I'm using a sidecar just to be able to use t3code lol. I hope one of these solutions merges soon

@ReyJ94 I wish I had seen your PR before starting mine, I searched and couldn't find anything so I started working on my own, I just hope one of ours gets merged, T3 code is basically unusable on Windows right now (yes I have found other bugs too with the Windows backend too and I'm prob gonna submit PR's for those too) and I understand not many devs use Windows but its what I'm stuck on now unfortunately, I feel like Theo needs a dedicated person for addressing/merging the Windows bugs directly, although I feel that its def not a priority for him right now

martinoturrina:
I had this issue, and due to corporate file scanning the startup time was over 2 minutes with the mounted WSL disk.

I had to do a mount copy of t3 to avoid going over the 60s timeout and be stuck at "Connecting to WSL".
This fix would be highly appreciated

SunkenInTime:
> **Current status, 2026-08-20.** This measurement used the older 5232e36 head. The PR now points to f2ce5ce. Current-head testing found three merge blockers and measured a cold WSL post-update launch at 7.373 s median. Treat the measurements below as historical. See the [current review](https://github.com/pingdotgg/t3code/pull/5769#issuecomment-5361550452) and [updated report](https://codex-file-host.shawnadedeji.workers.dev/files/pr-reviews-5769/5b3af22a16d5-pr5769-review-analysis.html?v=9#speed-chart).
Windows package/update measurement on exact head `5232e36a75b0c7f6632d8a496e343582d4d3d76c` versus exact base `ddaa6afef55e0da53950c7ed45f7d91b717475a5`:

- installer: **145.39 MiB → 215.04 MiB** (**+47.9%**, +69.65 MiB)
- the new `wsl-runtime.tar.gz` is **70.05 MiB** with 26,677 tar entries
- electron-builder blockmap comparison finds **78,275,251 bytes** in the head installer whose checksums are not reusable from the base (34.71% of the head installer), so a differential update from this base would need roughly 78.3 MB of new payload before protocol overhead
- isolated clean NSIS install: **96.90s base vs 96.76s head** on the same Windows host (effectively unchanged)

I could not obtain the compensating WSL result on this host because `wsl.exe --list --verbose` itself did not return within 30 seconds, so I am not treating that as evidence against the runtime-cache design. But for a stack intended to make Windows updates/installations faster, this is a substantial and directly measured updater/download regression. Could you add real cold/warm WSL startup numbers and either reduce the archive/update cost or explicitly justify this tradeoff before merge?

I also verified the packaged Windows app itself end-to-end: clean install succeeded and the installed desktop UI became usable; runtime-ready receipt was 18.26s and renderer-visible time was 19.30s.

ReyJ94:
> Windows package/update measurement on exact head `5232e36a75b0c7f6632d8a496e343582d4d3d76c` versus exact base `ddaa6afef55e0da53950c7ed45f7d91b717475a5`:
> 
> - installer: **145.39 MiB → 215.04 MiB** (**+47.9%**, +69.65 MiB)
> - the new `wsl-runtime.tar.gz` is **70.05 MiB** with 26,677 tar entries
> - electron-builder blockmap comparison finds **78,275,251 bytes** in the head installer whose checksums are not reusable from the base (34.71% of the head installer), so a differential update from this base would need roughly 78.3 MB of new payload before protocol overhead
> - isolated clean NSIS install: **96.90s base vs 96.76s head** on the same Windows host (effectively unchanged)
> 
> I could not obtain the compensating WSL result on this host because `wsl.exe --list --verbose` itself did not return within 30 seconds, so I am not treating that as evidence against the runtime-cache design. But for a stack intended to make Windows updates/installations faster, this is a substantial and directly measured updater/download regression. Could you add real cold/warm WSL startup numbers and either reduce the archive/update cost or explicitly justify this tradeoff before merge?
> 
> I also verified the packaged Windows app itself end-to-end: clean install succeeded and the installed desktop UI became usable; runtime-ready receipt was 18.26s and renderer-visible time was 19.30s.

Can you check mine, in my pr mentioned in my comment above ? Is it faster ?

SunkenInTime:
> **Current status, 2026-08-20.** This measurement and its ready-to-merge verdict applied to the older 5232e36 head. That verdict no longer applies. The PR now points to f2ce5ce, where current-head testing reproduced three merge blockers. The latest cold WSL post-update result is 7.373 s median, followed by 2.542 s and 2.538 s warm. See the [current review](https://github.com/pingdotgg/t3code/pull/5769#issuecomment-5361550452) and [updated report](https://codex-file-host.shawnadedeji.workers.dev/files/pr-reviews-5769/5b3af22a16d5-pr5769-review-analysis.html?v=9#speed-chart).
Follow-up to my earlier WSL measurement request: I installed/repaired WSL on the Windows 11 test host and re-ran the current head (`5232e36a75b0`) with a release-equivalent Linux x64 `node-pty` sidecar supplied through `--wsl-prebuild` (the same input shape used by release CI).

Environment: WSL 2.7.11.0, Ubuntu 26.04 LTS, WSL kernel 6.18.33.2, Node 24.19.0.

Measured results:

- Exact release build: **424.1 s**; installer **225,517,572 bytes / 215.070 MiB**; WSL archive **73,481,621 bytes / 70.078 MiB**. The archive contains the Linux x64 `pty.node` plus its validation marker.
- Isolated NSIS install: **92.490 s**, **14,692 files**, **614,753,598 bytes**.
- Empty WSL runtime cache: extraction/prepare **3.136 s**; backend receipt **23.483 s** after Electron launch; main renderer **23.608 s** after launch.
- Cached runtime, warm WSL VM: prepare **0.189 s**; backend receipt **8.715 s**; renderer **8.760 s**.
- Cached runtime after `wsl --shutdown`: prepare **0.189 s**; backend receipt **10.440 s**; renderer **11.523 s**.
- Same installed build with the runtime archive temporarily disabled, forcing the old `/mnt/c/.../app.asar.unpacked` path: cold-VM backend receipt **46.500 s**, renderer **47.395 s**; warm-VM backend receipt **41.028 s**, renderer **41.241 s**.

The most direct cold-VM control is therefore **36.060 s / 77.55% faster to backend readiness** and **35.872 s / 75.69% faster to the main renderer** with the cached ext4 runtime. The real Linux backend ran from `/home/t3review/.t3/runtime/...`; the packaged Electron renderer/window came up successfully in every archive-backed run. The warm mounted control remained stable with one backend and no restart/error signal.

One packaging footgun remains worth noting: a local Windows build made without `--wsl-prebuild` still emits an installer, but enabling WSL then restart-loops with the clear “packaged Linux node-pty binary was not included” error. Release CI supplies the sidecar, so this does not block the release-equivalent path I tested; a future fail-fast for local release packaging may still be useful.

**Verdict update: ready to merge at the tested head.** The previous request for real WSL cold/warm evidence is satisfied, and the measured startup win comfortably justifies the ~70 MiB payload tradeoff.

ReyJ94:
Sol here — I investigated this for #5042 rather than answering as Rey.

On the same host and using the same current #5042 archive, I ran each PR's generated staging script directly:

| path | cold stage | warm stage |
| --- | ---: | ---: |
| #5042 | 8.51s | 0.03s |
| #5769 | 6.57s | <0.01s |

So #5042 is **not faster on the cold staging step** in this comparison. Its roughly 1.94s cost is the full SHA-256 pass over the archive before extraction. Warm reuse is effectively negligible for both. I did not restart a live desktop instance, so this is a staging-script comparison, not a claim about end-to-end backend-ready time.

The current #5042 archive is 75,606,172 bytes (about 72.10 MiB) with 26,585 entries. A Linux-hosted packaging probe produced a 224,661,775-byte NSIS compressed payload, but final installer assembly could not complete without Wine, so I am not presenting that as an exact installer measurement or as an apples-to-apples size win over this PR.

I also applied the packaging issue identified here to #5042 in `0411cf9f`: Windows artifact construction now fails if `--wsl-prebuild` is absent, rather than emitting an installer with a broken WSL backend. The content-addressed runtime race raised on #5042 is fixed separately in `29f1958e`.

— GPT-5.6 Sol (coding agent, posting through @ReyJ94's account)

CDVolvik (commented):
This is the WSL-timeout fix [#6321](https://github.com/pingdotgg/t3code/pull/6321) described and did not ship. Runtime goes to `~/.t3/runtime` inside the distro so later launches do not read the server tree through `/mnt/c`.

`DesktopWslEnvironment.test.ts` on the PR files: 32 passed (install/prune script parsing included). `build-desktop-artifact.test.ts` did not load here (`Cannot find package '@electron/asar'`) — that is this mixed checkout, not a verdict on the new archive-hash tests.

Hash is SHA-256, normalized, rejected if not 64 hex. Install timeout 2 min, prune 30s. First launch after an update will be slower; later launches should not depend on the Windows mount.

MERGEABLE. I did not time a packaged WSL start. Related to #5042 / #6321 — only one of these should own the "copy the runtime onto the Linux filesystem" job.

ReyJ94:
i closed also my PR at https://github.com/pingdotgg/t3code/pull/5042 . The slowdown of wsl2 start has been fixed now on nightly, there should be no reason to keep this up anymore

ikifar2012:
Should be good to go, much faster than the original, here's the updated speed test with the latest changes

https://github.com/user-attachments/assets/e7d59b46-86c7-4741-abe4-aa89f133aefa

SunkenInTime:
## Review result: changes requested

I tested commit `f2ce5ce` on Ubuntu WSL 2 using the POSIX scripts generated by this PR. Formatting, lint, and typechecking pass, and the new WSL tests pass. Three runtime-cache failures still block merge.

### Reproduced blockers

1. **The archive identity is trusted instead of verified.** The desktop code parses the SHA-256 sidecar and uses it as the runtime ID, but it never hashes the archive and compares the bytes. I paired a sidecar containing `aaaa...aaaa` with an archive whose actual SHA-256 was `2ddfa9b013dd5c3b7a7310238c757fc33e68e6ddb46766f3bd5bb27f112e565f`. The generated installer exited 0 and cached it under the claimed identity. See [DesktopBackendConfiguration.ts](https://github.com/pingdotgg/t3code/blob/f2ce5cea9d9d799a9d85bda5c5c4a0292cff174a/apps/desktop/src/backend/DesktopBackendConfiguration.ts#L526).

2. **A broken native payload remains cache-ready.** I installed a runtime, deleted `pty.node` and the source archive, then ran the generated installer again. It exited 0 and reused the cache because readiness checks only the marker, server entry, and `node-pty/package.json`. The later native probe fails without invalidating the cache, retrying the install, or using the mounted-tree fallback. See [DesktopWslEnvironment.ts](https://github.com/pingdotgg/t3code/blob/f2ce5cea9d9d799a9d85bda5c5c4a0292cff174a/apps/desktop/src/wsl/DesktopWslEnvironment.ts#L262).

3. **Pruning can delete another running version's cache.** With active versions A and B plus newer cache C, pruning from A deleted B. Pruning from B then deleted A. This happened in 30 of 30 runs. A prune lock only serializes the damage. Each running backend needs a lifetime lease that every pruner checks. See [DesktopWslEnvironment.ts](https://github.com/pingdotgg/t3code/blob/f2ce5cea9d9d799a9d85bda5c5c4a0292cff174a/apps/desktop/src/wsl/DesktopWslEnvironment.ts#L311) and [DesktopBackendManager.ts](https://github.com/pingdotgg/t3code/blob/f2ce5cea9d9d799a9d85bda5c5c4a0292cff174a/apps/desktop/src/backend/DesktopBackendManager.ts#L948).

### Cold and warm WSL measurements

I ran 10 update sequences with a 45.170 MB compressed runtime, the real Linux `node-pty` probe, WSL IP discovery, the real server, and `/.well-known/t3/environment` as the readiness stop condition. Each first connection followed `wsl --shutdown` and installed a new runtime ID. The second and third connections reused the live VM and runtime.

- First-ever installation, cold WSL: 7.517 s
- First after update, cold WSL: 7.373 s median, 7.178 to 7.669 s, 10 runs
- Second connection, warm WSL: 2.542 s median, 2.519 to 2.733 s, 10 runs
- Third connection, warm WSL: 2.538 s median, 2.518 to 2.556 s, 10 runs
- WSL boot-only control: 4.008 s median, 3.832 to 4.285 s, 10 runs

The second connection is 4.831 s faster than the cold post-update connection, a 65.5% reduction. WSL startup accounts for about 83.0% of that gap. These totals start immediately before the Windows-side `wsl.exe` invocation and stop at HTTP readiness. Shutdown and teardown are excluded.

The full review, fix plan, chart, and reproduction evidence are in the [Blueprint report](https://codex-file-host.shawnadedeji.workers.dev/files/pr-reviews-5769/5b3af22a16d5-pr5769-review-analysis.html?v=9#speed-chart). The [33 readiness rows](https://codex-file-host.shawnadedeji.workers.dev/files/pr-reviews-5769/cold-results.csv) and [10 boot controls](https://codex-file-host.shawnadedeji.workers.dev/files/pr-reviews-5769/wsl-boot-control.csv) are also available as CSV.

ikifar2012:
Hey @SunkenInTime I ran my agent through all your comments, tested it myself and it appears to have fixed the issues you commented on, could you take another look?

ikifar2012:
Can I trigger T3 Code Bot?

ikifar2012:
Ok I think we are all good

apps/desktop/src/backend/DesktopBackendConfiguration.ts
ikifar2012: Done — staging is an optimization now, not a requirement. When `prepareRuntime` fails, `runWslPreflight` logs a warning and continues to the mounted `app.asar.unpacked` tree, so the worst case is the pre-PR launch time. A staged tree that cannot load node-pty also retries against the mount (and invalidates the cache) instead of ending the WSL attempt; only a mounted-tree failure is fatal, and a transient mounted failure stays retryable rather than inheriting the staged verdict.

apps/desktop/src/wsl/DesktopWslEnvironment.ts
ikifar2012: Done — the lock is exactly that:

```sh
exec 9> "$runtime_lock"
flock -x 9
```

The pid file, the stale-lock dance, and the sleep loop are all gone, so a killed install or `wsl --shutdown` leaves nothing to recover from. An executed test now runs two installs concurrently through a real shell and asserts the archive is extracted exactly once and both agree on the cache root.

docs/user/install.md
ikifar2012: Done — `DesktopBackendManager` calls `pruneRuntimes` once the backend reports ready. It keeps the current runtime and the newest previous ready cache, skips any cache a live backend is still running from, and sweeps orphaned install scratch older than two hours. The docs paragraph is now one sentence saying cleanup is automatic.

apps/desktop/src/wsl/DesktopWslEnvironment.test.ts
ikifar2012: Fixed in 1814e23fc — the four `build*` entries are alphabetical with the rest of the list now.

apps/desktop/src/wsl/DesktopWslEnvironment.test.ts
ikifar2012: Done — there is an executed suite now (`WSL runtime install script (executed)`) that runs the generated script through a real POSIX shell against a fixture tar.gz with `HOME` pointed at a temp dir, skipping only where the shell lacks flock/sha256sum/tar (Git Bash), which is neither Linux CI nor a Windows box with WSL.

It covers warm reuse without the archive present, reinstall after the server entry is truncated, failing out to the mounted tree when the cache is broken and unfixable, two concurrent installs extracting once and agreeing on the promoted directory, a mid-stream extraction failure leaving no half-built cache and no scratch, and refusal on both a digest mismatch and a content-id mismatch. The generated-text assertions that remain pin ordering the executed tests cannot observe (that a warm launch exits before the hash, that the digest gate runs before anything mutates the cache).

scripts/build-desktop-artifact.ts
```diff
@@ -2176,6 +2194,42 @@ const stageWslNodePtyPrebuild = Effect.fn("stageWslNodePtyPrebuild")(function* (
...
+const stageWslRuntimeArchive = Effect.fn("stageWslRuntimeArchive")(function* (input: {
+  readonly sourceDir: string;
+  readonly archivePath: string;
+  readonly hashPath: string;
+}) {
+  const fs = yield* FileSystem.FileSystem;
+  const path = yield* Path.Path;
+  yield* fs.makeDirectory(path.dirname(input.archivePath), { recursive: true });
+  yield* runCommand(
+    ChildProcess.make("tar", buildWslRuntimeArchiveArgs(input.archivePath), {
```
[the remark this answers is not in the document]
ikifar2012: Fixed — `stageWslRuntimeArchive` now passes `wslRuntimeArchiveTarTarget(path.relative(sourceDir, archivePath))` with `cwd: sourceDir`, so the `-f` target is relative and colon-free no matter what the stage root looks like.

The regression covers the real staging call rather than `buildWslRuntimeArchiveArgs()` alone: it builds a stage where the archive is at an absolute path in a sibling directory, stands in for tar by resolving the `-f` target against the spawned cwd exactly as tar would, and asserts the target is relative, contains no colon, and still resolves to `archivePath`.

docs/user/install.md
```diff
@@ -43,6 +43,14 @@ Stable:
 yay -S t3code-bin
 ```
 
+### Windows Subsystem for Linux
```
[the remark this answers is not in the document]
ikifar2012: Fixed — the WSL section now sits after the Nightly block, so the `t3code-nightly-bin` paragraph stays under Arch Linux.

apps/desktop/src/wsl/DesktopWslEnvironment.ts
```diff
@@ -216,6 +244,99 @@ const runWslShell = (
...
+  const safeRuntimeId = sanitizeWslRuntimeId(runtimeId);
+  return [
+    "set -eu",
+    'runtime_parent="$HOME/.t3/runtime"',
+    `runtime_root="$runtime_parent/${safeRuntimeId}"`,
+    `ready_marker="$runtime_root/${WSL_RUNTIME_READY_MARKER}"`,
+    "runtime_is_ready() {",
+    '  [ -f "$ready_marker" ] &&',
+    '    [ -f "$runtime_root/apps/server/dist/bin.mjs" ] &&',
+    '    [ -f "$runtime_root/node_modules/node-pty/package.json" ]',
```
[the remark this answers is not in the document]
ikifar2012: Fixed — the ready marker is no longer trusted on its own. `runtime_is_ready` requires the native payload (a `node_modules/node-pty/prebuilds/linux-*/pty.node` next to its `t3code-wsl-node-pty.json` marker) and the sha256 the install recorded for `apps/server/dist/bin.mjs`, so a deleted or truncated payload is a cache miss and reinstalls.

For the case only the probe can see — payload present but unloadable — a fatal staged `ensureNodePty` failure now falls through to the mounted tree, and when the mount works `invalidateRuntime` revokes the marker so the next launch reinstalls. Executed corruption tests cover both the reinstall path and the fall-back-when-unfixable path.

apps/desktop/src/wsl/DesktopWslEnvironment.ts
```diff
@@ -216,6 +244,99 @@ const runWslShell = (
...
+
+export const buildWslRuntimePruneScript = (runtimeId: string): string => {
+  const safeRuntimeId = sanitizeWslRuntimeId(runtimeId);
+  return [
+    "set -eu",
+    'runtime_parent="$HOME/.t3/runtime"',
+    `current_runtime="$runtime_parent/${safeRuntimeId}"`,
+    '[ -d "$runtime_parent" ] || exit 0',
+    'previous_runtime=""',
+    'for candidate in "$runtime_parent"/*; do',
```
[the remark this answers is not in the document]
ikifar2012: Fixed — a backend launched from a cache has that cache's path in its argv (the entry is `<runtime>/apps/server/dist/bin.mjs`), so the running process is itself the lease and exiting releases it. `buildWslRuntimePruneScript` checks `/proc/[0-9]*/cmdline` and skips any cache in use, so A and B both survive C's arrival; the sweep exits early where `/proc` is not available rather than guessing. The install path takes the same check and moves an in-use tree aside instead of deleting it under a live backend.

apps/desktop/src/backend/DesktopBackendConfiguration.ts
```diff
@@ -467,31 +515,48 @@ const resolveWslStartConfig = Effect.fn("desktop.backendConfiguration.resolveWsl
...
+  const archivePath = environment.path.join(environment.resourcesPath, WSL_RUNTIME_ARCHIVE_NAME);
+  const archiveHashPath = environment.path.join(
+    environment.resourcesPath,
+    WSL_RUNTIME_ARCHIVE_HASH_NAME,
+  );
+  const hasArchive = environment.isPackaged
+    ? yield* fileSystem.exists(archivePath).pipe(Effect.orElseSucceed(() => false))
+    : false;
+  const archiveHash = hasArchive
+    ? yield* fileSystem.readFileString(archiveHashPath).pipe(
```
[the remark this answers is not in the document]
ikifar2012: Fixed — the archive is hashed and compared now. On a cache miss the install runs `sha256sum` on the archive and refuses to extract unless it matches the recorded digest (a warm launch exits before that line, so reuse does not pay for it), and after extraction it checks the tree's own `.t3code-wsl-runtime-content-id` manifest against the id the cache directory is named for before promoting. Executed tests cover a mismatched digest and a mismatched content id, and assert neither leaves anything behind under `~/.t3/runtime`.

apps/desktop/src/wsl/DesktopWslEnvironment.ts
```diff
@@ -216,6 +260,222 @@ const runWslShell = (
...
+    '  for candidate in "$1"/node_modules/node-pty/prebuilds/linux-*/pty.node; do',
+    '    [ -f "$candidate" ] || continue',
+    '    [ -f "${candidate%/*}/t3code-wsl-node-pty.json" ] || continue',
+    "    return 0",
+    "  done",
+    "  return 1",
+    "}",
+    "runtime_is_ready() {",
+    '  [ -f "$ready_marker" ] &&',
+    '    [ -f "$runtime_root/apps/server/dist/bin.mjs" ] &&',
```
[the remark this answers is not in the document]
ikifar2012: Fixed in 36ab20a32 (pushed just before this review landed, against cb5d9dd) — the ready marker now holds the sha256 of `apps/server/dist/bin.mjs`, written only when the install has verified the archive bytes and the extracted tree's content-id manifest, and every warm reuse re-hashes the entry against it. Hashing the ~7MB entry is single-digit milliseconds inside the distro, once per launch.

A zero-byte or otherwise altered entry is a cache miss: it reinstalls when the archive is there, and exits non-zero with no `runtimeRoot` when it is not, which is what sends the backend to the mounted tree. Both are executed regressions that truncate a real installed fixture and rerun the script.

scripts/build-desktop-artifact.ts
```diff
@@ -2267,6 +2337,140 @@ const stageWslNodePtyPrebuild = Effect.fn("stageWslNodePtyPrebuild")(function* (
...
+  archivePath,
+  ...WSL_RUNTIME_ARCHIVE_EXCLUDED_PREFIXES.map((prefix) => `--exclude=${prefix}*`),
+  WSL_RUNTIME_CONTENT_ID_MANIFEST,
+  ...WSL_RUNTIME_ARCHIVE_CONTENT_ROOTS,
+];
+
+// The runtime's canonical identity: sha256 over every shipped file's posix
+// path and bytes, walked in sorted order. It is computed from the staged tree
+// rather than the archive, and deliberately ignores mtimes, ownership, and
+// mode, so two builds of the same server payload agree even though their
```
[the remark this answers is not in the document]
ikifar2012: Fixed in c73b467 and 75b0fc8 (pushed just before this review landed, against cb5d9dd) — the id now hashes each entry's type, a normalized semantic mode, and either the file's content hash or the symlink's target, while still ignoring mtime and ownership. The mode collapses to the executable bit only (`644`/`755`), since owner and group bits carry no meaning once tar restores the tree as whoever ran the install, and a Windows builder reports no executable bits at all so every regular file there lands in the same bucket and the id stays stable.

Regressions cover both directions: equal ids for equivalent trees with different incidental metadata, and different ids for an executable-bit change, a content-root mode change, and a symlink versus a regular file holding the same bytes.

UtkarshUsername (approved):

PR #2546, opened 2026-05-06
port desktop app to Effect

juliusmarminge:
## Summary
- Reworked desktop backend readiness probing around Effect-based services and effects, with temporary promise shims preserved for existing call sites.
- Added Effect/Vitest coverage for readiness polling, request timeouts, abort handling, and backend startup race behavior.
- Updated the desktop main process to use typed `URL` handling for backend endpoints and the new readiness error type.

## Testing
- `bun fmt`
- `bun lint`
- `bun typecheck`
- `bun run test` for `apps/desktop/src/backendReadiness.test.ts` and `apps/desktop/src/backendStartupReadiness.test.ts`

---

> [!NOTE]
> **High Risk**
> Large refactor of the Electron desktop main process, backend spawning/readiness, and lifecycle handling; mistakes could prevent the app from starting or shut down cleanly. Also changes runtime configuration/paths and protocol handling, which can impact packaging and local file access.
> 
> **Overview**
> Ports the desktop Electron main process to an Effect-based architecture, introducing layered services for app identity/environment/config, lifecycle/shutdown coordination, observability/tracing, and Electron wrappers (app/dialog/menu/protocol).
> 
> Reworks desktop backend startup into Effect-managed spawning with fd-based bootstrap payloads, HTTP readiness polling, structured child-process output logging, and automatic restart scheduling/cancellation logic.
> 
> Cleans up legacy promise/imperative utilities and tests (e.g., `backendPort`, readiness helpers, settings/persistence/dialog helpers) in favor of new Effect/Vitest test coverage, and updates build hygiene (.gitignore consolidation) plus bumps `electron` to `41.5.0`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35835a7a84759ebf2731272728570c5fcc332ef5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Port desktop app to Effect by replacing imperative Electron code with layered Effect-TS services
> - Rewrites [main.ts](https://github.com/pingdotgg/t3code/pull/2546/files#diff-31a471f6ef958ceff6e87ee910f4bc4f7bbc4986f797f159353b328a5916f7cb) from imperative Electron startup code to a declarative Effect-TS layer composition, using `NodeRuntime.runMain` to execute a `DesktopApp.program` with all services provided via layers.
[...]

PR #1729, opened 2026-04-04
Fix Windows PATH hydration and repair

Badbird5907:
- hydrate desktop and server PATH handling for Windows CLI tools
- probe shells without profiles first, then fall back to profile-loaded env when node is missing
- share command availability and PATH repair utilities across desktop and server

## What Changed

- Added a Windows startup fallback that loads the user’s PowerShell profile only when the initial no-profile environment still cannot find node
- Made Windows environment repair update the full set of relevant environment variables, not just PATH
- This only affects the packaged Windows `.exe` startup path and does not affect running the CLI from an existing shell, such as `npx t3`

## Why

This adds support for windows setups using environment managers like [fnm](https://github.com/Schniz/fnm) that initialize node through the user's powershell profile instead of PATH.
I was also able to confirm this still works on systems with a stock node setup.

Before:
image

After:
image

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] I included before/after screenshots for any UI changes
- [x] I included a video for animation/interaction changes

---

> [!NOTE]
> **Medium Risk**
> Changes Windows environment hydration and terminal shell selection, which can affect process spawning and tool discovery at runtime (especially in packaged apps). Risk is moderate due to cross-platform branching and reliance on PowerShell probing, but changes are well-covered by new tests.
> 
> **Overview**
> **Improves Windows startup environment repair** by adding shared utilities in `@t3tools/shared/shell` to probe PowerShell for env vars, merge/dedupe `PATH` (including known CLI install dirs), and *only* load the user profile when `node` still isn’t resolvable (also capturing `FNM_*` vars).
> 
> **Wires the Windows repair path into app entrypoints**: `apps/desktop` `syncShellEnvironment` and `apps/server` `fixPath` now apply the full env patch on `win32` and return early; unsupported platforms remain no-ops.
> 
> **Adjusts Windows terminal spawning behavior**: terminal manager now prefers `pwsh.exe` (with `-NoLogo`) and falls back through absolute Windows PowerShell and `cmd.exe`, with tests updated/added; PTY adapter now wraps spawn failures as structured `PtySpawnError`.
> 
> **Centralizes command availability + build tweaks**: `isCommandAvailable` is moved to shared shell utils and re-exported from `server/open`, and the desktop artifact build on Windows disables `npmRebuild`, disables signing/editing when unsigned, and uses `bun install --omit optional` plus a fallback-installed `electron-builder` invocation.
> 
[...]

PR #5771, opened 2026-08-09
fix(desktop): skip the PowerShell profile probe when node is already resolvable

### LINKED ISSUE #4403: [Bug]: Unconditional PowerShell profile discovery delays Windows desktop startup
### Before submitting

### Area

`apps/desktop`

### Steps to reproduce

1. On Windows, ensure `pwsh.exe` is installed. The affected machine also has Windows PowerShell 5.1 available as `powershell.exe`.
2. Ensure `node.exe` is already available from the inherited/no-profile `PATH`:
   ```powershell
   pwsh.exe -NoProfile -NonInteractive -Command "(Get-Command node).Source"
   ```
3. Use a normal PowerShell profile that performs environment initialization. No artificial delay or `Start-Sleep` is required for this reproduction.
4. Fully exit T3 Code and verify no prior T3 Code/Electron process remains, to keep this separate from #4395.
5. Start the packaged desktop app normally.
6. Inspect:
   ```text
   %USERPROFILE%\.t3\userdata\logs\desktop.trace.ndjson*
   ```
7. Find the `desktop.shellEnvironment.installWindowsEnvironment` and `readWindowsEnvironment` spans.
8. Repeat several cold launches.

On the affected machine, every observed startup performs both:

1. a `-NoProfile` environment read; and
2. a profile-loaded environment read.

This happens even though `node.exe` is already available after the first read.

### Expected behavior

Windows environment hydration should first construct a baseline environment without loading user profiles.

If that baseline is sufficient to launch the required runtime, profile discovery should be skipped. This was the behavior introduced in #1729: load the PowerShell profile only when `node` is not available from the baseline environment.

If profile discovery is still required—for example, to obtain FNM variables or additional profile-defined paths—the complete discovery operation should have one total startup budget. Trying another PowerShell candidate should consume the remaining budget rather than start a fresh timeout.

A profile failure or timeout should remain best-effort: preserve the baseline environment and continue startup.

### Actual behavior

The affected implementation unconditionally performs the no-profile read and then the profile-loaded read:

- [`DesktopShellEnvironment.ts` lines 297–353](https://github.com/pingdotgg/t3code/blob/9a0a07167f0623c3a7db0ffeff2e3939760309df/apps/desktop/src/shell/DesktopShellEnvironment.ts#L297-L353)
- [`DesktopApp.ts` lines 229–243](https://github.com/pingdotgg/t3code/blob/9a0a07167f0623c3a7db0ffeff2e3939760309df/apps/desktop/src/app/DesktopApp.ts#L229-L243)

Environment hydration also occurs before user-data resolution, lifecycle registration, the single-instance configuration, and `app.whenReady()`.

Each read tries the following candidates sequentially:

```text
pwsh.exe
powershell.exe
```

Each candidate receives a new five-second timeout. Because both no-profile and profile-loaded reads use this loop, failures can consume up to four separate timeout budgets, plus process-termination overhead.

#### Measured packaged-app startup traces

Twenty-three successful Windows desktop starts were measured between 2026-07-15 and 2026-07-23:

| Startup phase | Count | Minimum | Median | p95 | Maximum | Mean |
|---|---:|---:|---:|---:|---:|---:|
| No-profile read | 23 | 1.16 s | 2.06 s | 3.31 s | 3.32 s | 2.11 s |
| Profile-loaded read | 23 | 2.21 s | 4.20 s | 9.70 s | 11.35 s | 4.34 s |
| Total shell hydration | 23 | 3.54 s | 6.31 s | 12.21 s | 13.35 s | 6.45 s |

Every trace contained exactly two aggregate environment reads in this order: no-profile, then profile-loaded.

Two of the 23 starts reached profile-probe timeouts:

```text
trace 8be0a1df…
pwsh.exe profile probe:       timed out
powershell.exe profile probe: timed out
profile phase:                11,354.5 ms
total shell hydration:        13,348.0 ms

trace c03677f6…
pwsh.exe profile probe:       timed out
powershell.exe fallback:      succeeded
profile phase:                 9,700.2 ms
total shell hydration:        12,206.7 ms
```

Across these traces, shell hydration accounted for a median 98.5% of the main-process `desktop.startup` span.

These measurements cover only main-process startup through the bootstrap request. They do **not** measure the additional time until the renderer, backend, or environment reconnect is usable; those layers can make the user-visible wait longer.

#### Direct profile/no-profile comparison

Five alternating executions of the same `pwsh.exe` environment-capture shape used by the app produced:

```text
-NoProfile:
  minimum:  671.6 ms
  median: 1,014.7 ms
  maximum: 1,738.5 ms

Profile loaded:
  minimum: 3,516.0 ms
  median: 3,881.5 ms
  maximum: 4,736.5 ms
```

For this environment:

```text
PATH identical:              yes
node available in both:      yes
FNM_DIR present:             no
FNM_MULTISHELL_PATH present: no
```

The profile-loaded probe therefore added several seconds without changing any of the values collected by the app.

#### Suspected regression

PR #1729 intentionally implemented a conditional fallback:

1. read `PATH` without loading profiles;
2. merge inherited and known CLI paths;
3. return immediately if `node` is available;
4. load the profile only if `node` remains unavailable.

Its tests asserted that the environment reader was invoked only once when `node` was already available.

The unconditional profile read appears to have been introduced during the large Effect port in #2546. That PR does not document an intended change to the Windows profile-loading policy, so this looks like a behavioral regression, although maintainer confirmation would be useful.

One open design question is whether the unconditional read was intended to collect `FNM_DIR`, `FNM_MULTISHELL_PATH`, or the complete profile `PATH` even when `node` already exists. If so, that requirement should be explicit and the work should still be bounded or removed from the synchronous startup path.

#### Possible acceptance criteria

- When the no-profile baseline satisfies the documented runtime requirements, no profile-loaded PowerShell process is started.
- When a profile is required, all PowerShell candidates share one total deadline.
- A timeout preserves and installs the usable baseline environment.
- Tests cover:
  - runtime already available without a profile;
  - runtime only available after profile loading;
  - a hanging `pwsh.exe` candidate;
  - fallback to `powershell.exe` within the remaining budget;
  - preservation of required FNM variables.
- Traces record whether profile discovery was required, which candidate ran, its outcome, and the remaining or total budget.

#### Related work checked

- #2951 removed the `cmd.exe` wrapper but did not change profile-loading policy.
- #3383 added structured failures and a five-second **per-command** timeout, not a total discovery budget.
- #4291 and #4221 bound server-side editor discovery; they do not affect desktop shell hydration.
- #4228 improves Windows process diagnostics but does not affect this startup path.
- #4395 tracks rejected duplicate launches leaving Electron processes and is intentionally separate from this issue.
- #3611 concerns a WSL/runtime-version startup loop rather than desktop PowerShell discovery.

I did not find an open issue or PR specifically tracking unconditional Windows profile discovery or its additive candidate timeouts.

### Impact

`Major degradation or frequent failure`

The delay occurs on every observed launch of the affected installation. It postpones all later desktop initialization and can add more than ten seconds before renderer/backend startup delays are considered.

### Version or commit

```text
T3 Code 0.0.29-nightly.20260722.878
Commit 9a0a07167f0623c3a7db0ffeff2e3939760309df
Electron 41.5.0
```

### Environment

```text
Windows 11 x64, build 26200.8875
PowerShell 7.6.4
Windows PowerShell 5.1
Node.js 22.14.0, already available from the no-profile PATH
Timezone: America/Bogota (UTC-05)
```

### Logs or stack traces

```text
Observed startup count: 23
Observation period: 2026-07-15 through 2026-07-23

No-profile read median:     2,064.0 ms
Profile-loaded read median: 4,201.4 ms
Total hydration median:     6,305.3 ms
Total hydration p95:       12,206.7 ms
Total hydration maximum:   13,348.0 ms

Profile timeouts observed: 2
Starts executing both profile modes: 23 of 23
```

Secrets, profile contents, and absolute user paths have been omitted.

No screenshot is needed for the startup-only behavior. A redacted trace extract can be attached if maintainers need the underlying span events.

### Workaround

Temporarily simplifying expensive PowerShell profile initialization reduces the delay, but this changes the user’s shell configuration and is not a reliable application-level workaround.

Placing `node.exe` on the inherited or no-profile `PATH` does not prevent the current implementation from loading the profile.

CDVolvik:
## Problem

On Windows, `installWindowsEnvironment` runs two PowerShell probes on every desktop start: one with `-NoProfile`, then one that loads the user profile. Each probe walks `pwsh.exe` and then `powershell.exe`, and every candidate gets its own 5 second timeout, so a slow or failing profile can burn up to four timeout budgets before the window appears.

The profile probe was never meant to be unconditional. #1729 added it as a fallback for when node is missing, and `resolveWindowsEnvironment` in `packages/shared/src/shell.ts` still works that way:

```ts
if (yield* commandAvailable("node", { env: baselineEnv })) {
  return baselinePatch
}
```

The desktop copy lost that gate in #2546 when the app was ported to Effect, so both probes now run every time.

## Fix

- build the baseline PATH from the no-profile probe, the known CLI directories, and the inherited environment
- return early when node already resolves from that baseline
- fall through to the profile probe only when it does not, so fnm users still get `FNM_DIR` and `FNM_MULTISHELL_PATH`
- resolve `CommandAvailability` in `make` alongside the other services, so the check is injectable in tests

PATH precedence is unchanged: profile, then known CLI directories, then the no-profile probe, then the inherited environment.

## Validation

- `vp test run src/shell/DesktopShellEnvironment.test.ts` in `apps/desktop` (10 passed)
- `vp run --filter @t3tools/desktop typecheck`
- `vp lint --report-unused-disable-directives` (clean; the one warning it reports is pre-existing in `apps/web`)

The new test fails with `expected 2 to equal 1` when the gate is removed, so it covers the regression rather than just the current behavior.

I did not measure packaged-app startup traces, so the probe count is covered by the test rather than by a trace comparison.

Closes #4403

---

> [!NOTE]
> **Low Risk**
> Localized Windows shell hydration logic with shared parity to `resolveWindowsEnvironment`; profile fallback remains when node is missing.
> 
> **Overview**
> **Windows desktop startup** no longer always runs the slow PowerShell **profile** probe. After the existing **no-profile** PATH probe, known CLI directories, and inherited `PATH` are merged into a baseline, the app uses shared **`CommandAvailability`** to check whether `node` resolves; if it does, it **returns early** and skips the profile probe (and fnm env vars unless needed).
> 
> When `node` is still missing, behavior matches before: profile probe runs to pick up profile `PATH`, `FNM_DIR`, and `FNM_MULTISHELL_PATH`, with the same PATH merge order (profile ahead of baseline). **`CommandAvailability`** is resolved in `make` and injected into `installIntoProcess` so tests can stub it.
> 
> Tests gain a **`nodeAvailable`** hook and a case that asserts only one PowerShell invocation (`-NoProfile`) when node is already available.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ccb618c6e66e3130d69e91bb1671f69f7f8fc5aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Skip PowerShell profile probe on Windows when `node` is already resolvable
> - On Windows, a baseline PATH is now computed by merging known CLI directories, a `-NoProfile` PATH probe, and the existing process PATH before any profile probing occurs.
> - If `node` is resolvable in the baseline environment, the full PowerShell profile probe is skipped entirely, avoiding slow or side-effectful profile execution.
> - When `node` is not resolvable, the existing profile-loading behavior is preserved and FNM-related variables (`FNM_DIR`, `FNM_MULTISHELL_PATH`) are still populated.
> - `DesktopShellEnvironment.make` now explicitly depends on `Path.Path` and `CommandAvailability` services to support the new resolution check.
> - Behavioral Change: baseline PATH is always set before profile probing, which changes the PATH order on Windows even when the probe runs.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ccb618c.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/shell/DesktopShellEnvironment.test.ts
+++ b/apps/desktop/src/shell/DesktopShellEnvironment.test.ts
@@ -10,6 +10,8 @@ import * as Stream from "effect/Stream";
 import * as ChildProcess from "effect/unstable/process/ChildProcess";
 import * as ChildProcessSpawner from "effect/unstable/process/ChildProcessSpawner";
 
+import { CommandAvailability } from "@t3tools/shared/shell";
+
 import * as DesktopEnvironment from "../app/DesktopEnvironment.ts";
 import * as DesktopShellEnvironment from "./DesktopShellEnvironment.ts";
 
@@ -69,7 +71,11 @@ function runShellEnvironment(input: {
   readonly platform: NodeJS.Platform;
   readonly handler: (command: ChildProcess.Command) => string;
   readonly failure?: PlatformError.PlatformError;
+  readonly nodeAvailable?: boolean;
 }) {
+  const commandAvailabilityLayer = Layer.succeed(CommandAvailability, () =>
+    Effect.succeed(input.nodeAvailable ?? false),
+  );
   const environmentLayer = Layer.succeed(
     DesktopEnvironment.DesktopEnvironment,
     DesktopEnvironment.DesktopEnvironment.of({
@@ -91,7 +97,14 @@ function runShellEnvironment(input: {
   }).pipe(
     Effect.provide(
       DesktopShellEnvironment.layer.pipe(
-        Layer.provide(Layer.mergeAll(environmentLayer, NodeServices.layer, spawnerLayer)),
+        Layer.provide(
+          Layer.mergeAll(
+            environmentLayer,
+            NodeServices.layer,
+            spawnerLayer,
+            commandAvailabilityLayer,
+          ),
+        ),
       ),
     ),
   );
@@ -197,7 +210,7 @@ describe("DesktopShellEnvironment", () => {
     }),
   );
 
-  it.effect("loads PowerShell profile environment on Windows", () =>
+  it.effect("loads PowerShell profile environment on Windows when node is missing", () =>
     Effect.gen(function* () {
       const env: NodeJS.ProcessEnv = {
         PATH: "C:\\Windows\\System32",
@@ -209,6 +222,7 @@ describe("DesktopShellEnvironment", () => {
       yield* runShellEnvironment({
         env,
         platform: "win32",
+        nodeAvailable: false,
         handler: (command) => {
           if (command._tag !== "StandardCommand") return "";
           const loadProfile = !command.args.includes("-NoProfile");
@@ -241,6 +255,50 @@ describe("DesktopShellEnvironment", () => {
         env.FNM_MULTISHELL_PATH,
         "C:\\Users\\testuser\\AppData\\Local\\fnm_multishells\\123",
       );
+    }),
+  );
+
+  it.effect("skips the PowerShell profile probe on Windows when node is already resolvable", () =>
+    Effect.gen(function* () {
+      const env: NodeJS.ProcessEnv = {
+        PATH: "C:\\Windows\\System32",
+        APPDATA: "C:\\Users\\testuser\\AppData\\Roaming",
+        LOCALAPPDATA: "C:\\Users\\testuser\\AppData\\Local",
+        USERPROFILE: "C:\\Users\\testuser",
+      };
+      const commands: ChildProcess.Command[] = [];
+
+      yield* runShellEnvironment({
+        env,
+        platform: "win32",
+        nodeAvailable: true,
+        handler: (command) => {
+          commands.push(command);
+          if (command._tag !== "StandardCommand") return "";
+          return command.args.includes("-NoProfile")
+            ? envOutput({ PATH: "C:\\Custom\\Bin;C:\\Windows\\System32" })
+            : envOutput({ PATH: "C:\\Profile\\Node;C:\\Windows\\System32" });
+        },
+      });
+
+      assert.equal(commands.length, 1);
+      assert.isTrue(
+        commands[0]?._tag === "StandardCommand" && commands[0].args.includes("-NoProfile"),
+      );
+      assert.equal(
+        env.PATH,
+        [
+          "C:\\Users\\testuser\\AppData\\Roaming\\npm",
+          "C:\\Users\\testuser\\AppData\\Local\\Programs\\nodejs",
+          "C:\\Users\\testuser\\AppData\\Local\\Volta\\bin",
+          "C:\\Users\\testuser\\AppData\\Local\\pnpm",
+          "C:\\Users\\testuser\\.bun\\bin",
+          "C:\\Users\\testuser\\scoop\\shims",
+          "C:\\Custom\\Bin",
+          "C:\\Windows\\System32",
+        ].join(";"),
+      );
+      assert.isUndefined(env.FNM_DIR);
     }),
   );
 
--- a/apps/desktop/src/shell/DesktopShellEnvironment.ts
+++ b/apps/desktop/src/shell/DesktopShellEnvironment.ts
@@ -4,9 +4,11 @@ import * as Effect from "effect/Effect";
 import * as FileSystem from "effect/FileSystem";
 import * as Layer from "effect/Layer";
 import * as Option from "effect/Option";
+import * as Path from "effect/Path";
 import * as Schema from "effect/Schema";
 import * as ChildProcess from "effect/unstable/process/ChildProcess";
 import * as ChildProcessSpawner from "effect/unstable/process/ChildProcessSpawner";
+import { CommandAvailability } from "@t3tools/shared/shell";
 
 import * as DesktopEnvironment from "../app/DesktopEnvironment.ts";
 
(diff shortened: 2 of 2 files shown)


CDVolvik:
Closing this one. #5878 landed the concurrency fix on the same function and removes most of what this PR was for: with both probes running under `Effect.all`, the serial cost is gone, and by the numbers in that PR the profile probe is the faster of the two (2066ms against 2718ms), so skipping it no longer buys measurable startup time.

What remains is one fewer PowerShell spawn. That is real but small, and I cannot put a measurement behind it, so it is not worth carrying a conflicting branch through a rewritten file. If profile load does turn out to dominate on machines with heavy profiles, that is a separate PR with numbers attached.

PR #5806, opened 2026-08-09
fix(usage): usage no longer misses custom provider-instance homes

### LINKED ISSUE #5805: [Bug]: Usage page misses Codex usage from provider instances with a custom CODEX_HOME
## Description

The usage page reports Claude Code usage only, even with heavy daily Codex use, when Codex is configured through `providerInstances` (multiple instances, each with its own `CODEX_HOME` / shadow home) rather than the legacy single-instance settings field.

Usage page reporting Claude only despite daily Codex use

## Reproduction

1. Configure one or more Codex instances in `providerInstances` with a custom `CODEX_HOME` (the legacy `providers.codex` field stays empty, which is what the settings UI produces for instance-based setups).
2. Run Codex turns; rollouts land under the configured home's `sessions` directory.
3. Open the usage page: Codex shows zero usage.

## Version or commit

Reproduced on `main @ 963ebf5b` (screenshot above is from that commit); also present in the current nightly.

## Impact

Minor in the scheme of things — the usage page is brand new and this only matters for multi-instance setups. Within that scope the numbers are silently wrong (a whole provider omitted, no error shown), which seemed worth reporting before the page reaches a wider audience.

## Cause

`UsageService.resolveTranscriptDirs` resolves transcript directories only from the legacy `settings.providers.<kind>` blobs. For instance-based setups that blob decodes to defaults, so the scan reads `~/.codex/sessions` only. Instances that use the default home still show up; any instance whose home differs from the default (custom `CODEX_HOME`, shadow home on a custom shared home) is never scanned.

I have a fix ready and will attach a PR.

zenDevNomad:
## What Changed

The usage scan now enumerates every configured provider instance instead of only the legacy `settings.providers.<kind>` blobs:

- A new `usageHomes.ts` lists each instance's config with the same precedence the provider registry uses (`deriveProviderInstanceConfigMap`): explicit `providerInstances` entries win, and the legacy blob only fills the default slot while no entry claims that id.
- Each instance's home resolves with the runtime's exact rules: a configured `homePath` wins; otherwise the provider's home variable from the instance's spawn environment applies (server environment as base, per-instance entries overriding, absolute paths only); never for shadowed homes, where the server always sets the variable itself.
- Resolved directories are canonicalised (`realPath`) and de-duplicated, so instances sharing a home — shadow overlays symlink `sessions` into the shared home — are scanned exactly once.

Server-only. No contract change (`sources` already carries several directories per provider) and no UI code change.

## Why

Configuring Codex through `providerInstances` (e.g. work + personal instances, each with its own `CODEX_HOME`) leaves the legacy `providers.codex` blob empty, so the scan read `~/.codex/sessions` — a directory those instances never write to — and the usage page reported no Codex usage at all.

Closes #5805.

One known edge is left for a follow-up (#5738): buckets are still keyed per provider kind, so two environments whose *partially* overlapping directory sets get de-duplicated client-side can still over-count in the overlap. Fixing that needs per-directory bucket attribution, which is a contract change and deliberately out of scope here.

## UI Changes

No UI code changes, but the visible effect on an instance-based setup (four Codex instances across two homes):

| Before | After |
| --- | --- |
| Codex missing entirely | Codex usage reported |
| Usage page with Claude only | Usage page with Codex and Claude |

## Testing

- New unit tests for the instance enumeration (legacy fallback, default-slot precedence including cross-driver claims, spawn-environment inheritance/override, absolute-path gating) and the home-path precedence.
- `vp test run src/usage/`: 40 passed; `tsgo --noEmit` and `vp lint` clean.
- Verified live against a real four-instance setup (two homes, two shadow overlays, legacy blob empty): both screenshots above are from the same state directory and transcripts, before/after only differing by this change on top of current `main`.

## Checklist

- [ ] I included a video for animation/interaction changes — not applicable

---

> [!NOTE]
> **Medium Risk**
> Changes which transcript directories are scanned and aggregated usage totals for multi-instance setups; logic is server-only with tests but mis-resolved paths could still skew reported usage.
> 
> **Overview**
> **Usage scanning** no longer relies only on legacy `settings.providers` blobs. It walks every Claude/Codex **provider instance**, resolves each home with the same rules as spawn (`homePath`, `CODEX_HOME` / `CLAUDE_CONFIG_DIR`, shadow homes), and **dedupes** transcript dirs via `realPath` so shared/shadow homes are counted once.
> 
> Adds **`usageHomes.ts`** (`listProviderHomeCandidates`, `scanHomePath`) aligned with provider registry precedence, plus unit tests. **`resolveTranscriptDirs`** in `UsageService` loops instances, skips invalid configs, and can return multiple dirs per provider (fixes missing Codex usage when instances live only under `providerInstances`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ed33dde6edc90f2f90052a50238181340989a0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `resolveTranscriptDirs` to scan transcript homes for every configured provider instance
> - Previously, `resolveTranscriptDirs` resolved a single transcript directory per provider. It now enumerates all configured Claude and Codex instances via the new `listProviderHomeCandidates` utility in [usageHomes.ts](https://github.com/pingdotgg/t3code/pull/5806/files#diff-37f9e7f04c437ed2bf12f56af3a05c2d1ed664cb845a2317bd015de3168c4d1a).
> - For each instance, the home path is derived using `scanHomePath`, which mirrors runtime precedence between configured `homePath` and environment-variable-provided home paths.
> - Instances with undecodable configs are skipped; results are de-duplicated by provider and canonical real path to avoid counting the same transcripts twice.
> - A legacy provider blob fallback is included when no instance claims the default slot, preserving backward compatibility.
> - Behavioral Change: `resolveTranscriptDirs` now returns zero or more entries per provider (instead of exactly one), so usage counts may change for users with multiple configured instances or non-default home paths.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0ed33dd.</sup>
> 
>

### DIFF
--- a/apps/server/src/usage/UsageService.ts
+++ b/apps/server/src/usage/UsageService.ts
@@ -14,6 +14,8 @@
 import * as NodeOS from "node:os";
 
 import {
+  ClaudeSettings,
+  CodexSettings,
   USAGE_CONTRACT_VERSION,
   type UsageProviderKind,
   type UsageSource,
@@ -37,6 +39,7 @@ import { ServerConfig } from "../config.ts";
 import * as ServerSettings from "../serverSettings.ts";
 import { resolveClaudeHomePath } from "../provider/Drivers/ClaudeHome.ts";
 import { resolveCodexHomeLayout } from "../provider/Drivers/CodexHomeLayout.ts";
+import { listProviderHomeCandidates, scanHomePath } from "./usageHomes.ts";
 import { UsageAggregator } from "./usageAggregation.ts";
 import { parseRateTable, type RateTable } from "./usagePricing.ts";
 import {
@@ -84,6 +87,9 @@ const encodeRatesCache = Schema.encodeEffect(
 const ScanCacheJson = Schema.fromJsonString(Schema.Unknown as unknown as Schema.Codec<unknown>);
 const decodeScanCacheFile = Schema.decodeUnknownEffect(ScanCacheJson);
 const encodeScanCacheFile = Schema.encodeEffect(ScanCacheJson);
+
+const decodeClaudeSettings = Schema.decodeUnknownEffect(ClaudeSettings);
+const decodeCodexSettings = Schema.decodeUnknownEffect(CodexSettings);
 
 export class UsageService extends Context.Service<
   UsageService,
@@ -196,7 +202,11 @@ export const make = Effect.gen(function* () {
       return nestedExists ? nested : path.join(homePath, "projects");
     });
 
-  /** Resolves the transcript directory for each provider. */
+  /**
+   * Resolves the transcript directories for each provider, one per configured
+   * provider instance. Instances that share a home (shadow homes symlink
+   * `sessions` back into the shared home) collapse to a single directory.
+   */
   const resolveTranscriptDirs = Effect.fn("UsageService.resolveTranscriptDirs")(function* () {
     // A settings failure must surface as an error: swallowing it here would
     // present "zero usage from every provider" as a valid answer.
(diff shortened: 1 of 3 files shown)


Jardo-51:
Is this still alive? It's been open a while, and #8470 has since arrived at the same fix from a different direction, so it would be good to know which one to get behind.

Related question about scope: as written, every instance's transcripts pool into one series per provider. #6920 asks for the instances to be shown as separate, labelled series, which is the half your own #5738 implements — `homePath` + `homeLabel` on `UsageBucket`, `label` on `UsageSource`, one series per home named after the configured instance. Since #5738 is parked waiting on this PR, is the plan still to land them in sequence, or would you rather fold the labelling into this one so the feature arrives whole?

t3dotgg:
> [!NOTE]
>

Closing in favor of [#8470](https://github.com/pingdotgg/t3code/pull/8470). It uses the provider registry's environment rules for usage scans and already carries this branch's server and instance home precedence. Review continues there. The physical-path deduplication case is recorded on the retained PR so two configured paths to the same transcript directory do not count it twice.

PR #5808, opened 2026-08-09
fix(client-runtime): stop counting the workflow coordinator as a working agent

### LINKED ISSUE #5807: [Bug]: Background-work banner counts the workflow coordinator as a working agent
### Before submitting

### Area

apps/web

### Steps to reproduce

1. Open a thread and run any workflow that spawns members (e.g. `Workflow` with 2 phases, one agent each).
2. Let the first member finish while the second is still running.
3. Look at the three places the app reports agent counts:
   - inline spawn card in the transcript
   - Agents right panel
   - background-work banner above the composer

Deterministic repro without a provider (client-side fold only):
- Seed `projection_thread_activities` with a `local_workflow` coordinator (`task.started`, `taskType: "local_workflow"`, `agentKind: "agent"`) plus two members carrying `parentAgentId`, one `task.completed` and one left running.
- Unit repro: `deriveAgentPanelModel` with that roster returns `liveCount: 2`.

### Expected behavior

All three surfaces agree on the number of agents actually working. With one member running and one settled, the banner should read **"1 agent working in the background"** and the panel footer **"1 working · 1 settled"**.

### Actual behavior

The banner reads **"2 agents working in the background"** and the Agents panel footer reads **"2 working · 1 settled"**, while the inline card correctly reads **"Spine · 1 working"**.

**Root cause:** `deriveAgentPanelModel` in `packages/client-runtime/src/state/subagentRuntime.ts` tallies status buckets over every entry in `source`, which includes the workflow coordinator (`kind === "workflow"`). The coordinator stays `running` for the whole run, so `liveCount = runningCount + waitingCount` is one higher than the number of agents doing work.

The inline card (`apps/web/src/components/chat/MessagesTimeline.tsx`, `AgentSpawnCtaRow`) builds its list from `workflowGroup.phases.flatMap(members)` — coordinator excluded — which is why it is the only correct one.

Second site with the same cause: `apps/web/src/components/AgentsPanel.tsx` renders `● {runningCount + waitingCount} working` in the footer.

Note the same function already applies a coordinator-exclusion rule for `totalTokens` a few lines below (`if (agent.kind !== "workflow" || members.length === 0)`); it was never applied to the status counts.

**Scale:** with N concurrent workflows the banner is off by N. A workflow whose members have all settled between phases reads "1 agent working" when zero are.

### Impact

Cosmetic issue

### Version or commit

main @ a20923ce4

### Environment

macOS 15 (Darwin 22.6.0), Chromium 1161, Node 24.16.0, T3 Code web dev server

### Logs or stack traces

```shell

```

Image

### Workaround

None. The inline spawn card and the per-phase counts in the Agents panel are correct; only the composer banner and the panel footer overcount.

artieeg:
Closes #5807

## What changed

`deriveAgentPanelModel` counted the workflow coordinator as one of the agents doing work. A workflow running a single member reported two:

- composer banner: **"2 agents working in the background"**
- Agents panel footer: **"2 working · 1 settled"**
- inline spawn card: **"Spine · 1 working"** ← the only correct one

The tally loop in `packages/client-runtime/src/state/subagentRuntime.ts` iterated every roster entry, including the coordinator (`kind === "workflow"`), which stays `running` for the whole run. So `liveCount = runningCount + waitingCount` was always one higher than the number of agents actually working.

`liveCount` now applies the same coordinator-exclusion rule the `totalTokens` sum four lines below already used: skip a coordinator that has members. A memberless coordinator still counts, so a workflow that has not spawned its first member — or is between phases — does not read as zero agents.

`runningCount` / `waitingCount` still count every roster entry, preserving the existing bucket-sum invariant (`idle + running + waiting + settled === roster.length`). `AgentsPanel`'s footer switches to `liveCount` so both surfaces agree with the inline card.

## Why it should exist

The banner is the only visible Stop affordance for background work, so its count is what a user reads to decide whether anything is still running. It disagreed with the two other surfaces in the same view. With N concurrent workflows the banner is off by N, and a workflow whose members have all settled between phases reads "1 agent working" when zero are.

Note the inline card is unaffected and stays correct: "Kicked off 2 subagents" is the batch size (how many were spawned), not a live count.

## Before / after

Identical fixture, one member running and one settled. Only the two counts differ.

| | Banner | Panel footer |
|---|---|---|
| Before | `2 agents working in the background` | `● 2 working · 1 settled` |
| After | `1 agent working in the background` | `● 1 working · 1 settled` |

Unchanged in both: inline card `Kicked off 2 subagents · ride-shelf-v2 — Spine · 1 working`, panel header `1/2 settled`, `RECON 1 done`, `SPINE 1 active · 0 done`.

Before:
image

After:
image

## Tests

Two cases added to `subagentRuntime.test.ts`:

- coordinator excluded — `liveCount` is 1 while `runningCount` is 2
- memberless coordinator still counts as 1, so a mid-spawn workflow does not read as zero

Verified red/green: with the source change reverted the first test fails `expected 2 to be 1`; with it applied 48/48 pass. `vp lint` and `tsgo --noEmit` clean for both packages.

## Scope

3 files, +35 / −5. No behavior change outside the two count readouts.

---

> [!NOTE]
> **Low Risk**
> UI-only count logic in client-runtime with targeted tests; no auth, data, or server behavior changes.
> 
> **Overview**
> **`liveCount` no longer treats workflow coordinators as working agents** when they already have member rows, so the composer banner, panel footer, and inline cards agree (e.g. one running member reads as **1 working**, not 2).
> 
> In `deriveAgentPanelModel`, `liveCount` is computed in the roster loop instead of `runningCount + waitingCount`. Live agents (`running`, `pending`, `waiting`) count toward `liveCount` unless they are a **workflow coordinator with members**—the same rule already used for `totalTokens`. A coordinator with **no** members still counts, so mid-spawn or between-phase runs do not show zero.
> 
> The **Agents panel** footer now displays `model.liveCount` directly. **`runningCount` / `waitingCount`** are unchanged and still sum to roster size with idle and settled.
> 
> Tests cover coordinator exclusion vs `runningCount`, and memberless coordinators still contributing `liveCount` of 1.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17a1a6feb9b76910e8d3fa7f0da29b72a8c9e59f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> 
> 
> [!NOTE]
> ### Fix `liveCount` in `deriveAgentPanelModel` to exclude workflow coordinators that have members
> - `liveCount` previously equaled `runningCount + waitingCount`, which included workflow coordinators even when they duplicate work already counted via their members.
> - `deriveAgentPanelModel` now computes `liveCount` independently: non-workflow agents count if running, pending, or waiting; coordinators only count when they have no members.
> - The `AgentsPanel` footer now uses `model.liveCount` directly instead of summing `runningCount + waitingCount`.
> - Behavioral Change: the working-agents count shown in the UI will be lower in workflows where coordinators have active members.
> >
> 
> >
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 17a1a6f.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/AgentsPanel.tsx
+++ b/apps/web/src/components/AgentsPanel.tsx
@@ -568,10 +568,8 @@ export function AgentsPanel({
       </ScrollArea>
       <footer className="flex items-center justify-between border-t border-border/60 px-3 py-1.5 font-mono text-[.7rem] text-muted-foreground">
         <span className="flex items-center gap-2">
-          {model.runningCount + model.waitingCount > 0 ? (
-            <span className="text-info-foreground">
-              ● {model.runningCount + model.waitingCount} working
-            </span>
+          {model.liveCount > 0 ? (
+            <span className="text-info-foreground">● {model.liveCount} working</span>
           ) : null}
           {model.idleCount > 0 ? <span>{model.idleCount} idle</span> : null}
           {model.settledCount > 0 ? <span>{model.settledCount} settled</span> : null}
--- a/packages/client-runtime/src/state/subagentRuntime.test.ts
+++ b/packages/client-runtime/src/state/subagentRuntime.test.ts
@@ -390,6 +390,28 @@ describe("deriveAgentPanelModel", () => {
     );
   });
 
+  it("excludes the workflow coordinator from the live count", () => {
+    // The composer banner and the panel footer read liveCount. A workflow
+    // running one member must read "1 agent working", not 2 — the
+    // coordinator is the workflow itself, not a worker.
+    const model = deriveAgentPanelModel({ agents: roster });
+    // wf-1 (coordinator) + wf-1:wf:1 (member) are both live in the tally...
+    expect(model.runningCount).toBe(2);
+    // ...but only the member is an agent doing work.
+    expect(model.liveCount).toBe(1);
+  });
+
+  it("counts a memberless workflow coordinator as its own live work", () => {
+    // Dynamic spawns mean a coordinator can be live before its first member
+    // exists; dropping it outright would report zero agents mid-run.
+    const model = deriveAgentPanelModel({
+      agents: fold([
+        activity("task.started", { taskId: "wf-2", taskType: "local_workflow", title: "solo" }),
+      ]),
+    });
+    expect(model.liveCount).toBe(1);
+  });
+
   it("keeps direct spawns in first-seen order as their activity changes", () => {
     const directRoster = fold([
       activity("task.started", { taskId: "direct-a", title: "First" }, "2026-08-01T11:00:00.000Z"),
--- a/packages/client-runtime/src/state/subagentRuntime.ts
+++ b/packages/client-runtime/src/state/subagentRuntime.ts
@@ -825,11 +825,21 @@ export function deriveAgentPanelModel({
   let idleCount = 0;
   let settledCount = 0;
   let totalTokens = 0;
+  let liveCount = 0;
   for (const agent of source) {
+    const live =
+      agent.status === "running" || agent.status === "pending" || agent.status === "waiting";
     if (agent.status === "running" || agent.status === "pending") runningCount += 1;
     else if (agent.status === "waiting") waitingCount += 1;
     else if (agent.status === "idle") idleCount += 1;
     else settledCount += 1;
+    // A coordinator is the workflow itself, not a worker: counting it would
+    // report one more agent than is doing work ("2 agents working" for a
+    // single running member). Same exclusion rule as totalTokens below —
+    // a coordinator with no members is still its own unit of live work.
+    if (live && (agent.kind !== "workflow" || (members.get(agent.id) ?? []).length === 0)) {
+      liveCount += 1;
+    }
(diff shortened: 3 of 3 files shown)


juliusmarminge:
Closing as superseded by #6672, which merged and fixes the same underlying issue (#5807).

Both PRs patch `deriveAgentPanelModel`'s coordinator double-count. #6672's fix is a strict superset of this one: it hoists a single `continue` that excludes a workflow coordinator (when it has members) from the whole status tally loop — `runningCount`, `waitingCount`, `idleCount`, `settledCount`, `liveCount`, and `totalTokens` — whereas this PR only corrects `liveCount` and left `runningCount`/`waitingCount` still counting the coordinator (relying on the panel switching to `liveCount` for display).

Rebasing this branch onto current `main` produces a real conflict in `subagentRuntime.ts`/`subagentRuntime.test.ts` against #6672's already-merged fix — confirming the overlap. Thanks for the report and the fix here; #6672 covers it. 🙏

PR #5855, opened 2026-08-09
fix(server): assistant streaming no longer rescans thread history

### LINKED ISSUE #5719: [Bug]: Assistant streaming causes full-thread projection scans for every text delta
## Steps to reproduce

1. Run T3 Code 0.0.32 with a mature thread containing substantial message and activity history.
2. Enable **Settings → Assistant output → Stream assistant messages**.
3. Start a Codex turn that streams a moderately long response.
4. Observe CPU and RSS for the T3 server process (not only the provider subprocess).
5. Compare with assistant streaming disabled.

The degradation grows with both the number of streamed response chunks and the amount of existing thread history.

## Expected behavior

Streaming an assistant response should have approximately constant per-delta projection cost. Assistant text deltas should not repeatedly load and decode complete message, proposed-plan, activity, and approval collections.

## Actual behavior

Each provider assistant-text delta is dispatched as a durable `thread.message.assistant.delta` command and becomes a `thread.message-sent` event.

Every event is then applied serially to all nine projectors. The threads projector handles every `thread.message-sent` event by invoking `refreshThreadShellSummary`, which reloads all messages, proposed plans, activities, and pending approvals for that thread.

Sanitized measurements from a real 0.0.32 database:

- 3,610 logical messages produced 128,972 `thread.message-sent` events
- Average: 35.7 persisted chunk events per logical message
- Maximum observed: 2,436 events for one 11,646-character message
- Busiest thread: 25,878 orchestration events
- With approximately one new orchestration event per second, the T3 server sustained 110–165% CPU
- Server RSS oscillated between approximately 1.4 GB and 2.6 GB

The observed cost therefore scales with streamed chunks × accumulated thread history. Repeatedly loading and decoding the full activity collection also creates significant allocation and garbage-collection pressure.

### Relevant source paths

- `apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts`: when `enableAssistantStreaming` is true, each assistant delta is individually dispatched (currently around lines 1638–1683).
- `apps/server/src/orchestration/Layers/ProjectionPipeline.ts`: `refreshThreadShellSummary` loads all four thread collections (currently around lines 547–589).
- The same projector calls that refresh for every `thread.message-sent` event (currently around lines 845–860).
- Every event runs nine projectors serially, with a transactional projector cursor update for each one (currently around lines 1601–1692).

This appears to undermine the intent of #1647, which removed an earlier thread-wide scan from message projection.

Issue #2761 is adjacent but appears distinct: that report concerns large WebSocket snapshots and reconnect behavior, while this report concerns synchronous persistence/projection work during active streaming.

### Suggested direction

The narrowest fix may be to avoid `refreshThreadShellSummary` for assistant `thread.message-sent` events. Assistant text cannot change `latestUserMessageAt`, pending approvals, pending user input, or actionable proposed-plan state.

More generally, shell-summary fields could be maintained incrementally or through targeted aggregate queries for the event types that can actually affect each field. A useful regression contract would be: projecting an assistant delta performs a constant number of queries independent of thread history and does not query activities, proposed plans, or approvals.

## Impact

Major degradation or frequent failure.

## Version or commit

T3 Code Alpha 0.0.32.

## Environment

macOS on Apple Silicon, desktop app, Codex app-server provider, assistant streaming enabled.

## Logs or stack traces

No raw logs or database artifacts are attached because they can contain private agent content. The measurements above contain aggregate counts only.

## Workaround

Disable **Stream assistant messages**. The server checks this setting while processing deltas, so subsequent text is handled by the existing buffered path without terminating the provider session. Starting a fresh thread reduces the history-dependent portion of the cost but does not address the per-delta transaction amplification.

cheruvian:
Fixes #5719.

## Problem

The threads projector handled `thread.message-sent`, `thread.activity-appended` and four other event types with one blanket `refreshThreadShellSummary`, which reloaded the thread's entire messages, proposed plans, activities and pending approvals to recompute four summary columns. With assistant streaming enabled every provider text delta is a durable `thread.message-sent` event, so projection work scaled with streamed chunks times accumulated thread history.

## Fix

Each of the four shell-summary fields is owned by exactly one projection, and assistant text can change none of them. The projector now refreshes only the fields an event can actually invalidate:

- `latestUserMessageAt` — user messages only, advanced in place as a running maximum, since messages are append-only outside revert
- `pendingApprovalCount` — approval activities and `thread.approval-response-requested`
- `pendingUserInputCount` — user-input activities and `thread.user-input-response-requested`
- `hasActionableProposedPlan` — plan upserts and the events that move `latestTurnId`
- `thread.reverted` still re-derives everything, since it is the only path that removes projected rows

Fields are recomputed from the owning projection rather than incremented. That matters for rebuilds: each projector keeps its own cursor and bootstraps over the whole stream in turn, so the threads projector sees the upstream projections already fully caught up — recomputation converges to the correct value where counter arithmetic would drift.

Streaming, event volume and event ordering are unchanged. Only the cost of projecting each event drops.

## Measurements

The new test installs a SQL client decorator that records every statement the pipeline issues. For 8 streamed deltas:

| | statements | full-thread scans |
|---|---|---|
| before | 160 (20/delta) | 32 (4/delta) |
| after | 112 (14/delta) | 0 |

The same test pins history-independence: identical statement counts against a 4-activity and a 200-activity thread. A third test appends events without projecting them, runs `bootstrap`, and asserts the replayed summary matches live projection.

Worth noting for expectations: buffered delivery is the default now, so the ~36 events per message in the issue report only apply to `enableLegacyTokenStreaming`. This change still helps every setup, because `thread.activity-appended` fires constantly in both modes and was paying the same full-thread scans.

## Verification

- `vp test run apps/server/src/orchestration apps/server/src/persistence apps/server/src/relay` — 45 files, 298 tests passed
- `vp run --filter t3 typecheck` — clean
- targeted lint and format — clean

## Risks

- **Replay/rebuild** — the threads projector bootstraps after the others have caught up, so it sees future rows at every event. Recomputation converges correctly under that ordering; incrementing would not. Covered by the bootstrap test.
- **Reconnect** — no behavior change. The old blanket refresh incidentally self-healed on `thread.session-set`; that is gone, but every writer of the underlying rows now triggers its own field refresh, and only this pipeline writes these columns.
- **Concurrent activity updates** — projectors run serially with the owning projector ahead of the threads projector, so a recomputed counter always reads post-write state. One residual gap: if an existing `activityId` were re-upserted with a different `kind`, from a user-input kind to a non-user-input one, the counter would not be revisited. No provider path does that today.

Model and harness: Claude Opus 5 (1M context) in Claude Code.

---

> [!NOTE]
> **Medium Risk**
> Changes core orchestration projection semantics (transactions, cursor invariants, summary derivation) that affect every live event; risk is mitigated by extensive new cost/rebuild/revert tests and stricter upsert guards.
> 
> **Overview**
> **Stops assistant streaming deltas from reloading whole thread collections** when updating `projection_threads` shell fields. Shell refresh is now **field-scoped** via `threadShellSummaryFieldsForEvent` / `refreshThreadShellSummary(..., fields)` so assistant `thread.message-sent` events only bump `updatedAt` (user messages still advance `latestUserMessageAt` incrementally); approvals, user-input, plans, and revert paths refresh only the summaries they can invalidate.
> 
> **Reworks live `projectEvent`** into one SQL transaction: enforce all projector cursors at `sequence - 1`, run **only subscriber** projectors per event (`projectorNamesForEvent`), then advance **every** cursor in a single `projection_state` batch (`upsertMany`). Bootstrap/replay still uses per-projector transactions.
> 
> **Hardens projection upserts** when IDs collide across threads: conditional `ON CONFLICT` updates for messages, activities, plans, and pending approvals; message/approval apply paths ignore conflicting thread or role; `thread.reverted` prunes pending approvals tied to removed turns.
> 
> Adds **`ProjectionPipeline.summaryCost.test.ts`** (SQL statement recording, history-independence, targeted summary scans, bootstrap/revert parity) and extends **`ProjectionPipeline.test.ts`** for rollback cursor invariants and bootstrap resume after partial failure.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit baa977a0e1370d65e9c5e4579a828ce50e1ef439. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix assistant streaming to avoid rescanning thread history on each delta event
> - Projection of live events now applies all subscriber projectors and advances all cursors atomically in a single transaction, replacing per-projector individual transactions.
> - Adds `projectorNamesForEvent` routing so only projectors subscribed to a given event type run `apply`; non-subscriber projectors still advance their cursor.
> - Adds `assertLiveProjectorCursors` to enforce that all projector cursors are at `sequence - 1` before applying a live event, failing fast on out-of-order projection.
> - Thread shell-summary refreshes are now field-targeted via `threadShellSummaryFieldsForEvent`, so events like assistant streaming deltas no longer trigger full thread-collection scans.
> - Upsert guards added to projection tables (messages, activities, pending approvals, proposed plans) to skip updates when `thread_id`, `role`, or `kind` conflict with a different owner.
> - `ProjectionStateRepository` gains a `upsertMany` method for batching all projector cursor advances in one SQL statement.
> - Risk: `projectEvent` now fails before applying if any projector cursor is not at the expected prior sequence, which is a new hard precondition not previously enforced.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized baa977a.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProjectionPipeline.summaryCost.test.ts
+++ b/apps/server/src/orchestration/Layers/ProjectionPipeline.summaryCost.test.ts
@@ -77,6 +77,23 @@ const threadCollectionScanTables = (statements: ReadonlyArray<string>) =>
     );
   });
 
+const projectionStateWrites = (statements: ReadonlyArray<string>) =>
+  statements.filter((statement) =>
+    statement.replace(/\s+/g, " ").trim().toLowerCase().includes("insert into projection_state"),
+  );
+
+const PROJECTOR_NAMES = [
+  "projection.projects",
+  "projection.threads",
+  "projection.thread-messages",
+  "projection.thread-proposed-plans",
+  "projection.thread-activities",
+  "projection.thread-sessions",
+  "projection.thread-turns",
+  "projection.checkpoints",
+  "projection.pending-approvals",
+] as const;
+
 const at = (seconds: number) =>
   `2026-08-09T12:00:${String(seconds).padStart(2, "0")}.000Z` as const;
 
@@ -382,15 +399,17 @@ const seedThread = Effect.fn("seedThread")(function* (input: {
     eventStore
       .append(event)
       .pipe(Effect.flatMap((savedEvent) => projectionPipeline.projectEvent(savedEvent)));
-  const recordProjection = (event: AnyEventInput) =>
+  const recordProjectionResult = (event: AnyEventInput) =>
     Effect.gen(function* () {
       const savedEvent = yield* eventStore.append(event);
       recordedStatements.length = 0;
       yield* projectionPipeline.projectEvent(savedEvent);
       const statements = [...recordedStatements];
       recordedStatements.length = 0;
-      return statements;
+      return { sequence: savedEvent.sequence, statements };
     });
+  const recordProjection = (event: AnyEventInput) =>
+    recordProjectionResult(event).pipe(Effect.map(({ statements }) => statements));
   const projectAndAssert = (event: AnyEventInput, expected: TargetProjectionExpectation) =>
     recordProjection(event).pipe(
       Effect.flatMap((statements) => assertTargetProjection({ threadId, statements, ...expected })),
@@ -427,7 +446,14 @@ const seedThread = Effect.fn("seedThread")(function* (input: {
     { concurrency: 1, discard: true },
   );
 
-  return { appendAndProject, projectAndAssert, projectId, recordProjection, threadId };
+  return {
+    appendAndProject,
+    projectAndAssert,
+    projectId,
+    recordProjection,
+    recordProjectionResult,
+    threadId,
+  };
 });
 
 it.layer(TestLayer)("OrchestrationProjectionPipeline shell-summary cost", (it) => {
@@ -436,7 +462,7 @@ it.layer(TestLayer)("OrchestrationProjectionPipeline shell-summary cost", (it) =
     readonly historyLength: number;
   }) {
     const sql = yield* SqlClient.SqlClient;
-    const { appendAndProject, recordProjection, threadId } = yield* seedThread(input);
+    const { appendAndProject, recordProjectionResult, threadId } = yield* seedThread(input);
     const messageId = MessageId.make(`message-${input.slug}-assistant`);
 
     yield* appendAndProject(
@@ -453,7 +479,7 @@ it.layer(TestLayer)("OrchestrationProjectionPipeline shell-summary cost", (it) =
       }),
     );
 
-    const statements = yield* recordProjection(
+    const { sequence, statements } = yield* recordProjectionResult(
       messageEvent({
         slug: input.slug,
         label: "assistant-delta",
@@ -487,9 +513,21 @@ it.layer(TestLayer)("OrchestrationProjectionPipeline shell-summary cost", (it) =
       FROM projection_threads
       WHERE thread_id = ${threadId}
     `;
+    const projectionStateRows = yield* sql<{
+      readonly projector: string;
+      readonly lastAppliedSequence: number;
+      readonly updatedAt: string;
+    }>`
+      SELECT
+        projector,
+        last_applied_sequence AS "lastAppliedSequence",
+        updated_at AS "updatedAt"
+      FROM projection_state
+      ORDER BY projector ASC
+    `;
     recordedStatements.length = 0;
 
-    return { messageRows, statements, threadRows };
+    return { messageRows, projectionStateRows, sequence, statements, threadRows };
   });
 
   const measureTaskProgress = Effect.fn("measureTaskProgress")(function* (input: {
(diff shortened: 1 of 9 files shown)


Defmon3:
We investigated a Windows T3 instance where agent responses were delayed by minutes even though the provider CLI responded normally outside T3.

Server traces showed thread-shell summary refreshes repeatedly rereading growing thread histories. During the incident, a single refresh took as long as **4.97 seconds**, creating a backlog that delayed provider events and UI updates.

After applying this PR:

- the focused projection-cost suite passed **7/7**;
- a regression scenario containing **14,625 activities** kept the SQL work bounded instead of growing with thread history;
- a fresh isolated **10-subagent** run measured the same refresh at no more than approximately **0.6 ms**.

This confirms the PR addresses a major bottleneck we observed. We are not claiming it resolves every possible source of T3 latency.

t3dotgg:
> [!NOTE]

Closing in favor of merged [#8150](https://github.com/pingdotgg/t3code/pull/8150), which fixes assistant-message and routine-activity history scans with a smaller change. Further summary optimizations can be proposed separately if profiling shows a remaining problem. Your work is credited in #8150.

PR #5875, opened 2026-08-09
feat(mobile): auto-start new task flow in most recently active project

### LINKED ISSUE #5785: [Feature]: Start mobile New Task in the most recently active project
## Before submitting

- [x] I am describing a concrete problem or use case.

## Area

`apps/mobile` new-task flow.

## Problem or use case

Opening New Task always begins at the project picker even when one project was used moments ago. Repeatedly creating tasks in the same project requires selecting that project each time.

## Proposed solution

When New Task opens and there is a valid recently active project, start on that project composer. Keep the existing back action so the user can return to the project picker, and fall back to the picker when no recent project is available.

## Why this matters

It removes a repetitive navigation step from the common case while preserving explicit project selection.

## Smallest useful scope

Use the existing recent-project ordering already available to the mobile client. Do not persist a new preference or change project sorting. Apply the default only when starting a fresh new-task flow.

## Alternatives considered

Persisting a separate default-project preference adds settings and state. Always showing the picker is predictable but slower for repeated work.

## Risks or tradeoffs

Automatic navigation can surprise users who expect the picker. The selected project must still exist in the active environment, and back navigation must remain obvious and reliable.

## Contribution

mohamedmastouri-hue:
Fixes #5785

---

> [!NOTE]
> **Low Risk**
> Navigation-only UX change in the new-task flow with explicit guards for share flows and empty catalog states.
> 
> **Overview**
> Opening **Choose project** for a normal new task now **auto-navigates** into `NewTaskDraft` for the default project (`projectScopes[0].projects[0]`, ordered by recent activity via the new-task flow provider) instead of leaving users on the picker.
> 
> A **one-shot** `useEffect` plus ref ensures navigation runs only once. **Incoming shares** and reserved share destinations are skipped so existing share auto-navigation is unchanged. If loading finishes with **no projects**, the ref is set without navigating so the empty state stays visible.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4bff2b9756b5dcc331b0fb69903faa97104c9745. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Auto-navigate to the most recently active project when opening the new task flow
> In [NewTaskRouteScreen](https://github.com/pingdotgg/t3code/pull/5875/files#diff-ff6990d7f3ffc05e93955c8a5c7bf5a3b77e8139e45c8b63e3a296596cd2d29e), adds a one-time auto-navigation effect that sends the user directly to `NewTaskDraft` for the first available project when the screen loads without an incoming share or reserved destination project. A ref guard (`defaultSelectionNavigatedRef`) ensures the effect fires only once per screen mount. If loading completes with no projects, the empty state is shown instead.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4bff2b9.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/features/threads/NewTaskRouteScreen.tsx
+++ b/apps/mobile/src/features/threads/NewTaskRouteScreen.tsx
@@ -112,6 +112,39 @@ export function NewTaskRouteScreen({ route }: StaticScreenProps<NewTaskRoutePara
       ) ?? null)
     : null;
 
+  const defaultSelectionNavigatedRef = useRef(false);
+
+  useEffect(() => {
+    if (defaultSelectionNavigatedRef.current) {
+      return;
+    }
+
+    if (incomingShare || reservedDestinationProject) {
+      // Incoming shares have their own auto-navigation path; don't fight it.
+      defaultSelectionNavigatedRef.current = true;
+      return;
+    }
+
+    if (projectScopes.length > 0) {
+      defaultSelectionNavigatedRef.current = true;
+      const defaultProject = projectScopes[0]?.projects[0];
+      if (defaultProject) {
+        navigation.navigate("NewTaskSheet", {
+          screen: "NewTaskDraft",
+          params: {
+            environmentId: defaultProject.environmentId,
+            projectId: defaultProject.id,
+            title: defaultProject.title,
+          },
+        });
+      }
+    } else if (!projectEmptyState.loading) {
+      // If we finished loading and there are truly no projects, we shouldn't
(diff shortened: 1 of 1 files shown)


t3dotgg:
> [!NOTE]

This makes New Task immediately open the most recently active project and removes the project choice step from one route. We are keeping the current explicit project selection behavior.

PR #5889, opened 2026-08-09
fix(desktop): select the WSL route source address

### LINKED ISSUE #5211: [Bug]: Desktop stuck on "Connecting to WSL" — getDistroIp picks unreachable Docker bridge IP instead of eth0
### Summary

### Environment

### Repro

### Root cause
```ts
```

### Suggested fix

### Workaround

nickmatteo:
## What Changed

- Resolve the selected WSL distro's preferred IPv4 source with `ip -4 route get 1.1.1.1` instead of assuming the first `hostname -I` address is reachable.
- Parse the exact `src` token with strict IPv4 octet validation.
- Fall back to `hostname -I` only when it yields exactly one valid IPv4 address; ambiguous Docker-bridge output now returns no address rather than guessing.
- Keep the existing total timeout, scoped process cleanup, NAT direct-address behavior, and mirrored-mode loopback mapping.
- Add command-level and parser regression coverage for the reported Docker ordering, missing route source, unavailable `ip`, invalid octets, ambiguous fallback, and timeout behavior.

## Why

`hostname -I` enumerates addresses from every interface without a stable reachability order. With Docker running inside a mirrored WSL distro, bridge addresses can precede the route-selected host address, causing the desktop to poll an unreachable bridge indefinitely.

Asking the kernel for its preferred route source avoids interface-order assumptions without hardcoding Docker or RFC1918 ranges.

Closes #5211.

## Validation

- `vp test run apps/desktop/src/wsl/DesktopWslEnvironment.test.ts` — 38 passed
- `vp test run apps/desktop/src/backend/DesktopBackendConfiguration.test.ts -t "resolveWsl preserves existing WSLENV entries when forwarding backend secrets"` — 1 passed
- `vp run --filter @t3tools/desktop typecheck`
- `vp lint apps/desktop/src/wsl/DesktopWslEnvironment.ts apps/desktop/src/wsl/DesktopWslEnvironment.test.ts --report-unused-disable-directives`
- `vp fmt --check apps/desktop/src/wsl/DesktopWslEnvironment.ts apps/desktop/src/wsl/DesktopWslEnvironment.test.ts`
- Local Ubuntu 24.04 route lookup returned `src 172.20.42.37`, matching the parser contract.

The local manual check used default WSL NAT. The reporter's exact mirrored-networking-plus-in-distro-Docker topology remains the external acceptance boundary.

## Checklist

- [x] No UI, animation, or interaction changes

> [!NOTE]
> ### Fix `getDistroIpImpl` to select the WSL route source address instead of hostname
> - `getDistroIpImpl` now runs `ip -4 route get 1.1.1.1` first and extracts the `src` field via the new `parseWslRouteSource` helper, falling back to `hostname -I` only when the route command fails or returns a non-zero exit code.
> - `parseSingleWslHostnameAddress` replaces the previous loose regex, accepting the fallback only when `hostname -I` returns exactly one valid IPv4 address.
> - IPv4 validation is now done via `isIpv4Address` (checks each octet ≤255) instead of a regex pattern.
> - The timeout constant switches from `USER_HOME_TIMEOUT` to `DISTRO_IP_TIMEOUT` (5 seconds).
> - Behavioral Change: previously the first IPv4-like token from `hostname -I` was accepted; now multiple addresses from `hostname -I` yield `none`, and non-zero exit codes or spawn errors also yield `none`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 34a6876.</sup>
> 
>


CDVolvik:
Verified on Windows 11, Node 24: 38/38 pass on this branch. Reverting only `DesktopWslEnvironment.ts` to main while keeping the new tests fails 10 of them, including "uses the route-selected src instead of the first hostname address" and "uses one total timeout and does not start the fallback after it expires", so the suite covers the regression rather than the current behavior.

This is unit-level only. I did not run the packaged app against a real WSL distro, so I can't confirm the end-to-end Docker-bridge case from #5211.

CDVolvik (commented):
Re-verified on Ubuntu WSL (this machine, 2026-08-15).

`DesktopWslEnvironment.test.ts`: 38/38 on the PR. Replacing the source with `origin/main` fails 10/38 — all `is not a function`, because `getDistroIpImpl` / `parseWslRouteSource` / `parseSingleWslHostnameAddress` do not exist on main. That proves the tests are new. It does not prove they catch a wrong-address pick in the old `hostname -I` path.

Live on this distro the two answers agree:

```
ip -4 route get 1.1.1.1  → src 172.25.119.119
hostname -I              → 172.25.119.119
```

Only one address, so the Docker-bridge ordering bug does not reproduce here (Docker Desktop is a separate distro). The parser tests still look right: exact `src` token, reject `prefsrc`, reject a multi-address fallback.

The fallback-to-none when `hostname -I` is ambiguous is the part I would keep even if the route lookup stays.

Previous unit pass on 8/10 still holds.

t3dotgg:
> [!NOTE]
>

Closing in favor of [#8438](https://github.com/pingdotgg/t3code/pull/8438). It compares WSL address candidates with Windows interfaces, which avoids accepting an Internet route source that belongs to an unreachable WSL VPN. Review continues there. This branch's strict IPv4 parsing, single-address fallback, and shared timeout cases are recorded on the retained PR, which still needs native topology verification.

PR #5919, opened 2026-08-10
feat(web): add "Clear inactive" to the Agents panel

### LINKED ISSUE #5918: [Feature]: Let the Agents panel clear settled and idle agents
### Area

apps/web

### Problem or use case

On a long-lived thread the Agents panel accumulates every subagent it has ever spawned, and there is no way to hide any of them. Finished agents (completed / failed / cancelled / interrupted) and idle ones stay in the list forever, so the one agent that is actually working ends up buried under history.

This gets worse the more useful a thread is. A thread that has run several workflows can show a dozen-plus dead rows above the single running agent, and scanning for live status turns into a hunt.

Idle is the annoying case specifically. An idle agent is not finished, so it never reads as "done", but it is also not doing anything — it just sits there looking like it might still matter.

### Proposed solution

Add a **Clear inactive** button to the Agents panel footer. It hides every agent in the current thread that is not actively working — both settled and idle. Agents in pending, running, or waiting stay visible.

Nothing is destroyed. The clear is a per-thread timestamp cutoff:

- The footer shows **N hidden · Show all** while a cutoff is active, so it is always reversible.
- The empty state becomes "No active agents" with the hidden count and a **Show all** button.
- The cutoff lives in localStorage, keyed per thread.
- Any agent that starts working again reappears on its own, because its updatedAt moves past the cutoff.
- Workflow groups clear atomically — if any member is still working or newer than the cutoff, the whole coordinator/member group stays visible, so you never see a workflow with holes in it.

Explicitly out of scope: this is presentation-only. Clearing does not cancel an agent, delete provider history, or touch orchestration state. No server changes.

### Why this matters

Anyone running multi-agent threads for more than a few minutes. The Agents panel's job is to tell you what is happening *now*, and right now it competes with an unbounded, unfilterable log of what already happened. A one-click way to collapse to "just the live ones" makes the panel usable on exactly the threads where it matters most.

### Smallest useful scope

A single **Clear inactive** button that hides non-working agents for the current thread, with **Show all** to undo it. Client-side only, no persistence across devices, no per-agent dismissal, no auto-clearing.

### Alternatives considered

- **Clear only terminal agents.** Simpler, but leaves idle agents on screen — which is most of the clutter in practice, so the button would look broken.
- **Per-agent dismiss (an × on each row).** More precise, but it is a click per agent when the common case is "hide all the dead ones at once".
- **Auto-hide after a timeout.** Removes control from the user and makes agents vanish while you are reading them.
- **Collapsible "Earlier" section.** Some grouping already exists; it still leaves everything on screen and does not help when the history is long.
- **Reloading the page / switching threads.** Doesn't help — the panel rebuilds from the same state.

### Risks or tradeoffs

- **Idle is treated as clearable, which is a judgment call.** An idle agent is resumable, so someone could reasonably argue it should stay. The reappear-on-resume behavior is what makes this safe, and **Show all** is always there.
- **Cutoff state is per browser.** Clearing on desktop does not clear on mobile. Acceptable for a view preference.
- **The stored map grows** one entry per thread you have ever cleared, so it needs a cap (oldest dropped first) to avoid unbounded localStorage growth on long-lived installs.
- **Interaction with #5529** (resumed subagent stays in Earlier and stops updating). A cutoff based on updatedAt depends on that field advancing on resume; if it does not, a resumed agent could stay hidden. Worth checking these together.

### Examples or references

Related: #5529, #5219.

### Contribution

I would be open to helping implement this.

LordMerc:
Closes #5918

## What Changed

Adds a **Clear inactive** button to the Agents panel footer. It hides every
agent in the current thread that is not actively working — both settled
(completed / failed / cancelled / interrupted) and idle. Only `pending`,
`running`, and `waiting` agents stay.

Cleared rows are not gone: the footer shows **N hidden · Show all**, and the
empty state offers the same escape hatch. The cutoff is a timestamp stored in
`localStorage`, scoped per thread and capped at 200 entries.

Workflow groups stay atomic — if any member of a workflow is still working or
newer than the cutoff, the whole coordinator/member group remains visible.

## Why

On a long-lived thread the Agents panel accumulates finished and idle
subagents indefinitely, and there was no way to hide them. The working agent
you actually care about ends up buried in history.

Idle is deliberately treated as clearable. An idle agent is
resting-but-resumable, and in the panel it piles up exactly like a terminal
row — it is precisely the clutter being cleared. Gating only on terminal
statuses leaves the button dead in the common case.

## UI Changes

- Footer gains a **Clear inactive** button, disabled when there is nothing to
  clear.
- Footer gains a **N hidden · Show all** link once a cutoff is active.
- Empty state reads "No active agents" with a hidden count and a **Show all**
  button, instead of the generic "No agents yet".

*Before/after screenshots attached below.*
image

image

## Verification

Run under Node 24.19.0 (the repo requires `^24.13.1`), re-run after rebasing
onto `upstream/main`.

```
packages/client-runtime $ npx vp test run src/state/subagentRuntime.test.ts
  Test Files  1 passed (1)
       Tests  51 passed (51)

apps/web $ npx vp test run src/components/ChatView.logic.test.ts
  Test Files  1 passed (1)
       Tests  41 passed (41)

packages/client-runtime $ npx tsgo --noEmit      # exit 0
apps/web              $ npx tsgo --noEmit        # exit 0
$ npx vp lint <changed files>                    # clean
$ npx vp fmt --check <changed files>             # clean
$ git diff --check                               # clean
```

New coverage: idle rows are clearable; invalid and undateable timestamps fail
safe; the model reference is preserved when nothing is hidden; workflow groups
clear atomically; the persisted map respects its 200-entry cap and round-trips
through its schema.

Manually verified in a dev build against a real thread: a panel holding two
idle agents and one working agent cleared to just the working agent, and
**Show all** restored the other two.

## Scope / Safety

This is a **presentation-level change only**. Clearing agents does **not**:

- cancel or interrupt any agent,
- delete provider history or transcripts,
- mutate orchestration state or emit any event.

Nothing server-side is touched — no commands, deciders, projectors, or provider
adapters. The cutoff lives entirely in client `localStorage`, is per thread, and
is reversible via **Show all**.

An agent that resumes after being cleared reappears on its own, because its
`updatedAt` advances past the cutoff. This is covered by unit test; it was not
exercised manually in the UI.

## Checklist

- [x] Focused tests pass for every changed file
- [x] Typecheck clean in both affected packages
- [x] Lint and format clean on the diff
- [x] `git diff --check` clean
- [x] Manually verified in a dev build
- [x] No server, event, or orchestration changes
- [ ] Before/after screenshots attached

---

Implemented with GPT-5.6 Sol through the Codex harness in T3 Code.

---

> [!NOTE]
> **Low Risk**
> UI-only filtering and local persistence; no auth, server, or agent lifecycle changes beyond presentation-layer timestamp rules.
> 
> **Overview**
> Adds **Clear inactive** to the Agents panel so users can hide non-working agents (settled and idle) per thread without touching orchestration or server state. A per-thread ISO cutoff is stored in `localStorage` (capped at 200 threads), with **Show all** and footer/empty-state copy when rows are hidden.
> 
> `applyAgentPanelClearedAt`, `hasClearableAgents`, and `latestAgentActivityAt` in client-runtime filter the panel model by cutoff, treat workflow groups atomically, and derive the clear timestamp from visible agent `updatedAt` (not the client clock). `foldSubagentActivities` now accepts `sessionUpdatedAt` and only advances `updatedAt` forward so late completions, session death, and coordinator cascades do not leave cleared rows permanently hidden.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2128e0ad3b9e33e86cf5e32e8b1fde711d304d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add "Clear inactive" button to the Agents panel with per-thread persistence
> - Adds a "Clear inactive" button to [AgentsPanel](https://github.com/pingdotgg/t3code/pull/5919/files#diff-10464442e3a11a770b2b1a5718e7a452453347c492f6aa0533f267063e2c6494) that hides agents whose last activity is at or before a cutoff timestamp; workflows are kept or hidden atomically.
> - The cutoff is stored per-thread in localStorage via `agentsClearedAtByThread`, bounded to the newest entries by `boundAgentsClearedAtByThread` to cap storage growth.
> - A "Show all" affordance appears when rows are hidden, and the hidden count is shown in the footer; status counts and total tokens reflect only visible agents.
> - `foldSubagentActivities` in [subagentRuntime.ts](https://github.com/pingdotgg/t3code/pull/5919/files#diff-b28278fe62e259afab164f07d2e56800ca49f60942f6b12da5f748e9415c8259) now tracks `updatedAt` more accurately: terminal rows enriched post-completion advance `updatedAt`, and session-death interruptions are stamped at `sessionUpdatedAt` to remain visible relative to a prior clear.
> - Behavioral Change: agents interrupted by session death now carry a server-derived timestamp that affects their visibility when a cleared-at cutoff is active.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a2128e0.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/AgentsPanel.tsx
+++ b/apps/web/src/components/AgentsPanel.tsx
@@ -18,15 +18,18 @@ import type {
   RuntimeSubagent,
 } from "@t3tools/client-runtime/state/subagentRuntime";
 import {
+  applyAgentPanelClearedAt,
   formatSubagentModelLabel,
   formatSubagentTokenCount,
+  hasClearableAgents,
 } from "@t3tools/client-runtime/state/subagentRuntime";
 import type { EnvironmentId, ThreadId } from "@t3tools/contracts";
 import { Bot, Braces, Check, ChevronDown, ChevronRight, X } from "lucide-react";
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 
 import { cn } from "~/lib/utils";
 import { orchestrationEnvironment } from "~/state/orchestration";
+import { Button } from "~/components/ui/button";
 import { ScrollArea } from "~/components/ui/scroll-area";
 
 /**
@@ -522,20 +525,45 @@ export function AgentsPanel({
   model,
   environmentId = null,
   threadId = null,
+  clearedAt = null,
+  onClearInactive,
+  onShowAll,
 }: {
   model: AgentPanelModel;
   environmentId?: EnvironmentId | null;
   threadId?: ThreadId | null;
+  clearedAt?: string | null;
+  onClearInactive?: () => void;
+  onShowAll?: () => void;
 }) {
-  if (!model.hasAgents) {
+  const { model: visibleModel, hiddenCount } = useMemo(
+    () => applyAgentPanelClearedAt(model, clearedAt),
+    [clearedAt, model],
+  );
+  const canClearInactive = hasClearableAgents(visibleModel);
+
+  if (!visibleModel.hasAgents) {
     return (
       <div className="flex h-full flex-col items-center justify-center gap-2 p-6 text-center">
         <Bot aria-hidden className="size-6 text-muted-foreground/60" />
-        <p className="text-sm font-medium">No agents yet</p>
-        <p className="max-w-56 text-xs text-muted-foreground">
-          When this thread spawns subagents or runs a workflow, they show up here with live status,
-          activity, and token usage.
+        <p className="text-sm font-medium">
+          {hiddenCount > 0 ? "No active agents" : "No agents yet"}
         </p>
+        <p className="max-w-56 text-xs text-muted-foreground">
+          {hiddenCount > 0
+            ? `${hiddenCount} cleared ${hiddenCount === 1 ? "agent is" : "agents are"} hidden. New spawns still show up here.`
+            : "When this thread spawns subagents or runs a workflow, they show up here with live status, activity, and token usage."}
+        </p>
+        {hiddenCount > 0 && onShowAll ? (
+          <Button
+            variant="ghost"
+            size="xs"
+            onClick={onShowAll}
+            className="text-muted-foreground hover:text-foreground"
+          >
+            Show all
+          </Button>
+        ) : null}
       </div>
     );
   }
@@ -544,7 +572,7 @@ export function AgentsPanel({
     <div className="flex h-full min-h-0 flex-col">
       <ScrollArea className="min-h-0 flex-1">
         <div className="flex flex-col gap-2 p-2">
-          {model.workflows.map((group) => (
+          {visibleModel.workflows.map((group) => (
             <WorkflowSection
               key={group.workflow.id}
               group={group}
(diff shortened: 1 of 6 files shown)


LordMerc:
Addressed all three review findings in a2128e0ad:

- **macroscope (late `task.completed` enrichment)**: the already-terminal branch now advances `updatedAt` (monotonically, via a new `laterTimestamp` helper, so an out-of-order duplicate can't move a stamp backwards). A row cleared before its result arrived now reappears with the result.
- **Cursor Bugbot (interruption hidden by a prior clear)**: session-derived interruption now stamps `updatedAt` from the session record's `updatedAt` (fallback: newest folded activity), and the coordinator→member cascade can no longer stamp a member in the past. Agents that were working at clear time reappear as `interrupted` when their session dies. Note: the cited lines pointed at `clearedByCutoff`/`applyAgentPanelClearedAt`; the actual transition sites were the cascade and the `sessionLive === false` loop — fixed there.
- **macroscope (client wall-clock cutoff)**: the cutoff is now `latestAgentActivityAt(model)` — the newest server-stamped `updatedAt` on screen — instead of `new Date()`, eliminating cross-machine clock skew. Since the hide test is `<=`, the clear hides exactly the rows visible at clear time and nothing stamped later, which composes with the two fixes above.

Six regression tests added (57 passing in `subagentRuntime.test.ts`); typecheck/lint/format clean.

juliusmarminge:

**Reason:** Clear-inactive Agents feature never landed; issue bulk-closed
**Related:** #5918

PR #5920, opened 2026-08-10
fix(usage): unreachable environments no longer block the usage tab

### LINKED ISSUE #6363: [Bug]: usage tab keeps loading until every connected machine is online, otherwise it loads forever
## Summary

The Usage tab never settles while any connected machine (environment) is offline or not serving. It shows the loading skeleton until every environment answers — and since a down machine never answers (and never errors out), the page keeps loading forever until that machine comes back online.

## Repro

1. Have two (or more) machines/environments connected to the account (e.g. via T3 Connect), each with usage data.
2. Shut one machine down (or otherwise make it unreachable / not serving).
3. Open the Usage tab (from the sidebar or `/usage`).
4. Observed: the device strip shows the down machine as "still scanning" (`<label>…`) and the page stays on the loading skeleton indefinitely. It never shows partial totals and never reports an error for the down machine.
5. Bring the down machine back online (e.g. run `npx t3 serve` on it).
6. Now the Usage tab settles and renders totals.

## Expected

The Usage tab should settle within a bounded time even when a connected machine is unreachable. A machine that cannot report should be treated as failed/partial (surfaced through the coverage notice, like a machine that "could not report usage"), rather than keeping the page in the loading state forever. Totals should render from the machines that did answer, with the missing machine called out.

## Notes from code exploration

- `apps/web/src/state/usage.ts` derives `isPending` / `isPartial` from environments whose `summary === null && error === null` ("still reporting"). A down environment never transitions to `error` in the observed flow, so `stillReporting > 0` forever.
- `apps/web/src/components/usage/UsagePage.tsx` holds the whole page on the skeleton while `isPending || isPartial` — "Environments that are still answering never reach this notice; the page shows the loading skeleton until every one is terminal."
- The `usageSummary` RPC atom (`packages/client-runtime/src/state/server.ts`) has no explicit failure path/timeout that would turn an unreachable environment's pending request into an error state.

flamboh:
> [!NOTE]
> 🤖 **GPT-6 on behalf of Oliver**

Resolves #6363, #6045

## What Changed

- Usage fetching now times out per environment after 30 seconds, and the usage tab renders as soon as the first environment answers instead of waiting for all of them. 

- The device strip and coverage notice show a timed-out or failed environment as unavailable while reachable environments still contribute their numbers.
- **Refresh** retries the unavailable ones.
- When every environment fails or times out, web and mobile show a "No device reported usage" notice instead of a zero-totals dashboard. Refresh scans again, and a late answer restores the totals on its own.

## Why

- An environment that never connects, like a Tailscale box while Tailscale is off, used to pin the whole tab on the loading skeleton forever. A timeout plus an optional retry lets you see your usage without first fixing every environment you have configured.

- This does reverse an earlier decision.
    - The page used to hold all content until every environment settled so totals never jumped as devices landed. That tradeoff assumed every environment settles. 
    - When one never does, holding the page means showing nothing indefinitely, which is worse than numbers that tick up. 

- The device strip stays up while environments are still reporting, so you can tell a partial total from a final one.

## UI Changes

SCR-20260809-nwgr
SCR-20260809-omrs
SCR-20260809-ongd

## Checklist

---

Implemented by GPT-5.6 Sol and Fable 5.1 via Codex and T3 Code. Rebased by GPT-5.6 Sol; test trimming by GPT-6 through Codex.

> [!NOTE]
> ### Fix unreachable environments blocking usage tab by adding 30s query timeout and shared unreachable state
> - Environment usage queries now fail with a timeout after 30 seconds instead of staying pending indefinitely; a timed-out query can be retried via explicit refresh or automatic reconnect
> - Usage state derivation (merge, pending, partial, unreachable) moves into the shared `client-runtime` package so web and mobile use the same logic via `deriveUsageState` and `environmentUsageStatus`
> - Web and mobile Usage pages now render the first compatible environment's totals while others still scan, and show a no-report message when all environments settle without a usable answer
> - Connection-state comparator in `environmentQueryConnectionEquals` dedupes equivalent transient, connected, and failed states so query subscribers no longer receive no-op connection changes
> - Behavioral Change: `UsageView` gains a required `isUnreachable` boolean field on both platforms; any caller of `useUsage` that mocks or destructures the return must include it. `UsageCoverageNotice` row keys change from labels to environment IDs, which may affect snapshots or tests relying on label-based keys.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 028f6ac.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/features/usage/UsageRouteScreen.tsx
+++ b/apps/mobile/src/features/usage/UsageRouteScreen.tsx
@@ -92,8 +92,8 @@ export function UsageRouteScreen() {
   );
 
   // The pull spinner tracks re-scans of environments that have answered
-  // before. The initial scan renders its own placeholder, and an unreachable
-  // environment stays pending forever — neither may pin the spinner on.
+  // before. The initial scan renders its own placeholder, and the query
+  // deadline prevents an unreachable environment from pinning the spinner on.
   const refreshingUsage = environments.some((entry) => entry.isPending && entry.summary !== null);
   const showingLimits = tab === "limits";
   // One ScrollView serves both tabs, so the offset would otherwise carry over
--- a/apps/mobile/src/features/usage/UsageRouteScreen.tsx
+++ b/apps/mobile/src/features/usage/UsageRouteScreen.tsx
@@ -185,7 +185,7 @@ export function UsageRouteScreen() {
               </Text>
             ) : isUnreachable ? (
               <Text className="py-16 text-center text-base text-foreground-muted">
-                No device reported usage. Reconnect a device or pull to scan again.
+                No device reported usage.
               </Text>
             ) : (
               <>
--- a/apps/mobile/src/features/usage/UsageRouteScreen.tsx
+++ b/apps/mobile/src/features/usage/UsageRouteScreen.tsx
@@ -64,7 +64,7 @@ export function UsageRouteScreen() {
   const [metric, setMetric] = useState<UsageChartMetric>("cost");
   const { days: windowDays, window } = windowSelection;
   const isPast24Hours = windowDays === 1;
-  const { merged, environments, isPending, isPartial, refresh } = useUsage(window);
+  const { merged, environments, isPending, isPartial, isUnreachable, refresh } = useUsage(window);
   const limits = useRefreshLimits();
 
   const days = useMemo(
@@ -182,6 +182,10 @@ export function UsageRouteScreen() {
             ) : environments.length === 0 ? (
               <Text className="py-16 text-center text-base text-foreground-muted">
                 Connect an environment to see usage.
+              </Text>
+            ) : isUnreachable ? (
+              <Text className="py-16 text-center text-base text-foreground-muted">
+                No device reported usage. Reconnect a device or pull to scan again.
               </Text>
             ) : (
               <>
--- a/apps/mobile/src/state/usage.ts
+++ b/apps/mobile/src/state/usage.ts
@@ -65,6 +65,11 @@ export interface UsageView {
    * improve by waiting on them, so they must not read as "still reporting".
    */
   readonly isPartial: boolean;
+  /**
+   * True when environments exist but none answered: every one failed or timed
+   * out. Pages show a no-report state instead of zero totals.
+   */
+  readonly isUnreachable: boolean;
   readonly refresh: () => void;
 }
 
@@ -119,6 +124,7 @@ export function useUsage(input: UsageSummaryInput): UsageView {
     environments,
     isPending: usage.isPending,
     isPartial: usage.isPartial,
+    isUnreachable: usage.isUnreachable,
     refresh,
   };
 }
--- a/apps/mobile/src/state/usage.ts
+++ b/apps/mobile/src/state/usage.ts
@@ -10,29 +10,22 @@
  * @module state/usage
  */
 import { useAtomValue } from "@effect/atom-react";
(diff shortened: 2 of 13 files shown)


t3dotgg:
> [!NOTE]
>

Keep the all-unavailable behavior from [#8196](https://github.com/pingdotgg/t3code/pull/8196) at its [head commit](https://github.com/pingdotgg/t3code/commit/e03ed66dd12258a6b20036f27a9001400ee2b0a8) before this fix is considered complete. Both web and mobile must show a no-report state with Retry when every environment fails or times out, not a confirmed zero-usage dashboard. Retain its cases for a late answer restoring totals, Refresh starting a new attempt, an environment added after the prior deadline, and returning to a previously selected window. The source files are `UsagePage.test.tsx`, web `state/usage.test.tsx`, and the `isUnreachable` branches in both clients. This records the required update. The replacement does not yet implement it.

flamboh:
> [!NOTE]
> 🤖 **Fable 5.1 on behalf of Oliver**

Carried over the all-unavailable behavior from #8196 and rebased onto main.

When every environment has failed or timed out, both web and mobile now show a "No device reported usage" notice with the coverage list instead of the zero-totals dashboard. `deriveUsageState` in `client-runtime` derives an `isUnreachable` flag, and both hooks pass it through.

The four retained cases from #8196 map onto this branch as follows. The timeout lives in the query atom here rather than in hook state, so the deadline mechanics are covered by the runtime tests and the hook needs no timer of its own.

- **Late answer restoring totals**: `runtime.test.ts` "recovers automatically when an environment reconnects after timing out"; `usage.test.ts` checks the flag clears once a summary lands.
- **Refresh starting a new attempt**: `runtime.test.ts` "restarts a timed-out query after an explicit refresh"; `usage.test.ts` checks a waiting-after-failure query reads as pending again.
- **Environment added after the deadline**: each environment's query starts its own 30 second clock on first read, so a new environment cannot inherit an expired deadline. `usage.test.ts` checks an added pending environment drops the unreachable state.
- **Returning to a previously selected window**: a timed-out result is cached like any other settled result, so revisiting a window shows the unavailable state for that window and Refresh starts a new attempt. This differs from #8196, which restarted its hook-level grace period on every revisit. Caching the failure the same way as a success is simpler and consistent with how the rest of the page behaves.

I did not port #8196's `state/usage.test.tsx` harness, since it existed to drive a hook-level timer that no longer exists.

Also exercised in a browser against two local servers: baseline, all environments down then Refresh (no-report notice, no zero headline), automatic recovery on restart without a click, one of two environments down (totals kept, dead one marked unavailable), and revisiting a window. All behaved as described.

Mnigos (approved):
Tested this on a real two-environment setup: paired a second server into the client and then killed it, so the client kept a remembered environment that never answers, which is the #6045 case. On main the Usage page sits on the skeleton with "1 device still scanning" forever. On this branch the dead environment shows up in the coverage notice as unable to report and the totals, chart and breakdown from the live environment render normally. Also ran the touched test files locally, and reverting runtime.ts to main fails exactly the three timeout tests, so they pin the behavior.

Image

Mnigos:
One housekeeping note: the description resolves #6363, which was closed as a duplicate of #6045, so it might be worth pointing it at #6045 so the open issue closes on merge.

juliusmarminge:
Superseded by #9860 — same fix: render usage as environments respond instead of blocking on unreachable ones.

PR #5941, opened 2026-08-10
fix(server): stop the provider title mirror from overwriting real thread titles

### LINKED ISSUE #5245: [Bug]: OpenCode provider overwrites user-set thread titles repeatedly (provider title mirror bypasses canReplaceThreadTitle)
### What happens

A thread backed by the **OpenCode** provider has its user-set title repeatedly overwritten with OpenCode's auto-generated session title. I create a thread with an explicit title (`thread.create` with `title`), never send `titleSeed` or `regenerateTitle`, and the title is replaced within seconds — and then again, and again. Measured 13 and 38 overwrites on two threads in a single session.

`thread.meta.update` (the "rename" path) does hold the title, but only until the next provider event, so re-asserting it is a race that can't be won.

Threads on `claudeAgent`, `grok` and `cursor` keep their titles correctly. Only OpenCode (and, more quietly, Codex) is affected.

### Root cause

`canReplaceThreadTitle` (`apps/server/src/orchestration/Layers/ProviderCommandReactor.ts:165`) is not on this code path. Its only two call sites (`:1052`, `:800`) gate T3's *own* first-turn title generator. The provider-title mirror never consults it:

1. `apps/server/src/provider/Layers/OpenCodeAdapter.ts:199-205` reads `title` from every OpenCode `session.updated` event.
2. `OpenCodeAdapter.ts:785-825` emits it as `thread.metadata.updated` with `payload.name`, guarded only by `if (title)`.
3. `apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts:1707-1714` dispatches that straight into `thread.meta.update` — **no guard, no dedup, no comparison against the thread's current title**:

```ts
if (event.type === "thread.metadata.updated" && event.payload.name) {
  yield* orchestrationEngine.dispatch({
    type: "thread.meta.update",
    commandId: yield* providerCommandId(event, "thread-meta-update"),
    threadId: thread.id,
    title: event.payload.name,
  });
}
```

The repetition count comes from the OpenCode side: OpenCode's `patch()` publishes `session.updated` carrying the **entire** session record — including `title` — on *every* session mutation, including `touch()` on each prompt (`packages/opencode/src/session/session.ts:735`, `:752`). Their sync layer documents this as deliberate back-compat behaviour. So one generated title gets re-applied on every unrelated session change.

`ClaudeAdapter`, `GrokAdapter` and `CursorAdapter` never emit `thread.metadata.updated`, which is exactly why their titles survive. `CodexAdapter.ts:712-732` has the same unguarded path but fires only on the dedicated, infrequent `thread/name/updated` notification.

Also relevant: T3 never gives OpenCode a title. `OpenCodeAdapter.ts:1303-1312` calls `client.session.create({ permission })` with no `title`, so OpenCode auto-names the session and T3 mirrors that name back onto a thread the user already named.

### Suggested fix

Two options, not mutually exclusive:

1. **Pass the thread title at session creation** (smallest, no behaviour change elsewhere). OpenCode's `CreateInput` accepts `title` (`packages/opencode/src/session/session.ts:260-270`), and its titler backs off from any non-default title (`packages/opencode/src/session/prompt.ts:199-200`, `isDefaultTitle`). Sending the title at `OpenCodeAdapter.ts:1303` means OpenCode never generates one and the mirror becomes a self-consistent no-op.

2. **Guard the shared mirror** at `ProviderRuntimeIngestion.ts:1707` with `canReplaceThreadTitle` (currently module-private in `ProviderCommandReactor.ts`). ~5 lines, and it closes the same hole for Codex — though that is a deliberate semantic call, since a Codex thread would stop following Codex-side renames once T3 has a non-default title.

A `titleLocked` / `titleSource` field, or an `expectedTitle` compare-and-swap on `thread.meta.update` (mirroring the existing `expectedBranch` at `apps/server/src/orchestration/decider.ts:639-644`), would be the more thorough fix — there is currently no way for any client to express "this title is mine". Grepping `titleLocked|userTitle|pinnedTitle|titleSource|titleOwner` returns zero matches.

### Steps to reproduce

1. Create a thread on an OpenCode instance with an explicit non-default title (e.g. `My custom name`).
2. Send it a prompt.
3. Watch the sidebar: the title is replaced by OpenCode's generated name, and re-replaced on each subsequent session mutation.
4. Renaming the thread manually holds only until the next provider event.

### Version

Reproduced on nightly `v0.0.32-nightly.20260802.980`. Code references above are from commit `e60821f0e0d82a5d671ca3b94719c49d333921c8`. OpenCode side read at `anomalyco/opencode@1882c33`; running `opencode` binary reports `1.18.11`.

repparw:
## Summary

The provider title mirror (`thread.metadata.updated` -> `thread.meta.update`) in `ProviderRuntimeIngestion` dispatched every provider title with **no guard and no `titleSeed`**, bypassing `canReplaceThreadTitle` entirely. When OpenCode (or Codex) emits a session title, the mirror unconditionally overwrote the thread title.

Concretely with OpenCode:
1. OpenCode mints a placeholder title at `session.create` (`New session - <ts>`) and re-emits it on every `session.updated`.
2. The mirror copies it into the thread (`canReplaceThreadTitle` would have allowed it anyway while the thread is titled `New thread`).
3. Once the thread carries the placeholder, the real generated title is **rejected** in `maybeGenerateThreadTitleForFirstTurn` because the current title no longer equals the title seed. The thread is stuck with the placeholder.

## Changes

- **Pass the thread title at `session.create`** (`ProviderSessionStartInput.title` -> `OpenCodeAdapter`). OpenCode backs off titling any non-default title, so it never mints a placeholder in the first place.
- **Guard the shared mirror with `canReplaceThreadTitle`** in `ProviderRuntimeIngestion`: a provider title is only applied while the thread still has its default title. This closes the same hole for Codex and any other provider.
- **Stop mirroring OpenCode's default placeholder titles** in the adapter, covering older OpenCode builds whose `session.create` ignores the provided title.
- Extracted `canReplaceThreadTitle` / `DEFAULT_THREAD_TITLE` into `orchestration/threadTitles.ts`, shared by the reactor and the ingestion layer (no import cycle).

## Tests

- Adapter: title passed to `session.create` when provided; placeholder `New session - <ts>` titles are not mirrored while real ones are.
- Ingestion: provider title mirrored only while thread title is the default; rejected once the thread has a real title.
- Existing title-generation and mirror tests updated/kept green.

Fixes #5245

---

> [!NOTE]
> **Medium Risk**
> Changes thread-title behavior across providers and OpenCode session creation; wrong guards could leave default titles or still accept unwanted renames, but scope is limited to metadata mirroring with tests.
> 
> **Overview**
> Fixes provider-emitted session titles **unconditionally overwriting** thread names (notably OpenCode’s `New session - <timestamp>` placeholder), which could block server-side title generation.
> 
> **Provider runtime ingestion** now applies `thread.metadata.updated` → `thread.meta.update` only when `canReplaceThreadTitle` allows it (default **"New thread"** or title matching a seed). **`DEFAULT_THREAD_TITLE`** and **`canReplaceThreadTitle`** move to shared **`threadTitles.ts`** for the reactor and ingestion layer.
> 
> **Session start** passes the current thread **`title`** through **`ProviderSessionStartInput`** into **`providerService.startSession`**; **OpenCode** sends it to **`session.create`** so placeholders are less likely. The adapter still **drops** auto-generated **`New session -`** / **`Child session -`** titles before emitting metadata updates.
> 
> Tests cover mirror gating, OpenCode title forwarding, and placeholder filtering; the broad P1 ingestion test no longer expects provider rename when the harness thread already has a real title.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5afe24ea4af529900a16ddb8466700b359499096. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Stop provider title mirroring from overwriting user-set thread titles
> - Adds a `canReplaceThreadTitle` guard in [`ProviderRuntimeIngestion.ts`](https://github.com/pingdotgg/t3code/pull/5941/files#diff-22a6c9818b956463b848a73a5b3e787b44a144bd003b1cc25fa2b1b3110cdea7) so provider-originated title updates are only applied when the thread still has the default title (`'New thread'`).
> - Filters out OpenCode's auto-generated placeholder titles (matching `New session - <ISO timestamp>` or `Child session - <ISO timestamp>`) in [`OpenCodeAdapter.ts`](https://github.com/pingdotgg/t3code/pull/5941/files#diff-a4267388b15e5043ae1899beded972cbee9e52533ca3d9d65d438a91a9f7fbe7) so they are never mirrored as thread metadata updates.
> - Passes the current thread title to `providerService.startSession` and forwards it to `client.session.create` in OpenCode, so new sessions inherit the existing thread title.
> - Extracts shared utilities `DEFAULT_THREAD_TITLE` and `canReplaceThreadTitle` into [`threadTitles.ts`](https://github.com/pingdotgg/t3code/pull/5941/files#diff-c640b24f6e23d548aae27e3674ad646053c184a7492c78d299fd7d2ea5f89d2b).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5afe24e.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts
@@ -40,6 +40,10 @@ import {
 } from "../Services/ProviderCommandReactor.ts";
 import { forkParked, ServerActivation } from "../../serverActivation.ts";
 import {
+  canReplaceThreadTitle,
+  DEFAULT_THREAD_TITLE,
+} from "../threadTitles.ts";
+import {
   resolveSourceControlWriterModelSelection,
   ServerSettingsService,
 } from "../../serverSettings.ts";
@@ -91,7 +95,6 @@ const turnStartKeyForEvent = (event: ProviderIntentEvent): string =>
 const HANDLED_TURN_START_KEY_MAX = 10_000;
 const HANDLED_TURN_START_KEY_TTL = Duration.minutes(30);
 const DEFAULT_RUNTIME_MODE: RuntimeMode = "full-access";
-const DEFAULT_THREAD_TITLE = "New thread";
 const MAX_REGENERATION_ATTACHMENTS = 4;
 const MAX_THREAD_TITLE_CONTEXT_CHARS = 8_000;
 const MAX_FIRST_USER_TITLE_CONTEXT_CHARS = 2_000;
@@ -225,18 +228,6 @@ export function providerErrorLabelFromInstanceHint(input: {
   return providerErrorLabel(
     input.instanceId ?? input.modelSelectionInstanceId ?? input.sessionProvider,
   );
-}
-
-function canReplaceThreadTitle(currentTitle: string, titleSeed?: string): boolean {
-  const trimmedCurrentTitle = currentTitle.trim();
-  if (trimmedCurrentTitle === DEFAULT_THREAD_TITLE) {
-    return true;
-  }
-
-  const trimmedTitleSeed = titleSeed?.trim();
-  return trimmedTitleSeed !== undefined && trimmedTitleSeed.length > 0
-    ? trimmedCurrentTitle === trimmedTitleSeed
-    : false;
 }
 
 function findProviderAdapterRequestError(
@@ -626,6 +617,7 @@ const make = Effect.gen(function* () {
         ...(preferredProvider ? { provider: preferredProvider } : {}),
         providerInstanceId: desiredInstanceId,
         ...(effectiveCwd ? { cwd: effectiveCwd } : {}),
+        ...(thread.title ? { title: thread.title } : {}),
         modelSelection: desiredModelSelection,
         ...(input?.resumeCursor !== undefined ? { resumeCursor: input.resumeCursor } : {}),
         runtimeMode: desiredRuntimeMode,
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts
@@ -39,10 +39,7 @@ import {
   type ProviderCommandReactorShape,
 } from "../Services/ProviderCommandReactor.ts";
 import { forkParked, ServerActivation } from "../../serverActivation.ts";
-import {
-  canReplaceThreadTitle,
-  DEFAULT_THREAD_TITLE,
-} from "../threadTitles.ts";
+import { canReplaceThreadTitle, DEFAULT_THREAD_TITLE } from "../threadTitles.ts";
 import {
   resolveSourceControlWriterModelSelection,
   ServerSettingsService,
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
@@ -221,7 +221,10 @@ describe("ProviderRuntimeIngestion", () => {
     }
   });
 
-  async function createHarness(options?: { serverSettings?: Partial<ServerSettings> }) {
+  async function createHarness(options?: {
+    serverSettings?: Partial<ServerSettings>;
+    threadTitle?: string;
+  }) {
     const workspaceRoot = makeTempDir("t3-provider-project-");
     NodeFS.mkdirSync(NodePath.join(workspaceRoot, ".git"));
     const provider = createProviderServiceHarness();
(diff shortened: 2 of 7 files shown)


repparw:
The only failing check is the Vercel marketing preview, which reports 'Authorization required to deploy' — a fork-PR restriction that needs a maintainer to authorize the Vercel app. All code checks (Check, Test, Macroscope, Bugbot, native) pass.

PR #5945, opened 2026-08-10
feat(composer): show provider account usage in the input area

gfsaaser24:
> **Up front: this is a very large feature PR (~2,650 lines), which your contributing guide says not to open.** I am opening it anyway because the feature is finished, tested, and stable rather than a sketch, and it seemed more useful to you as working code than as an issue describing it. It is a UX addition, not a bug fix. **Please close it without ceremony if it is not a direction you want** — no reply needed, and I will not re-open or follow up.

## What Changed

Adds account usage meters to the composer.

**Contract** — a provider-agnostic `ProviderUsageLimits` in `packages/contracts`, shaped `{ id, label, usedPercent, resetsAt }`, so clients map over a list instead of branching per provider.

**Server** — an in-memory `ProviderUsageLimitsStore` fed by two sources:

- `ProviderUsageIngestion` consumes the `account.rate-limits.updated` runtime events the Claude and Codex adapters **already emit** and nothing currently reads. Zero extra network calls.
- `ProviderUsageRefresher` does a debounced (60s/instance) pull for Claude only, since Claude reports usage only on request. It reads the token the `claude` CLI already stored and calls the OAuth usage endpoint, which reports utilization **without spending message quota**. Codex needs no pull — it volunteers numbers over the app-server connection.

Provider-shaped payloads are flattened in `usageLimits.ts`. Every normalizer is total: anything unparseable yields no windows rather than an error.

**Web** — a ring per window beside the send button, collapsing to the worst window when the composer is narrow. Hover gives exact percentages and reset times.

**Mobile** — one toolbar button instead of three 44pt touch targets, listing the windows on tap.

Providers that do not report usage (Cursor, Grok, OpenCode) render nothing at all, so switching threads between providers costs no layout shift.

## Why

Right now you find out you are out of quota when a turn stops halfway through. The data to prevent that is already flowing — both adapters emit rate-limit events at the end of every turn and nothing consumes them.

Design decisions worth flagging, since they are the ones you would push back on:

- **Nothing is persisted.** A usage number is only interesting while it is roughly current, and both feeds repopulate within one turn of a restart.
- **The store only announces when a number actually moves.** `updatedAt` is stamped server-side on every reading, so publishing on the stamp would wake the registry and push the entire provider array — every model, capability, and skill — to every connected client once per turn, for no visible change.
- **No tickers.** Relative labels ("resets in 2h 15m") are stamped when the popover opens, not on an interval. A continuously repainting composer pegs the GPU on high-refresh displays.
- **Every failure path draws nothing.** No credentials, locked keychain, expired token, network blip — all resolve to blank meters. There is no error a user could act on here.

## UI Changes

Hovering a meter in the web composer:

Claude usage popover in the composer, showing Session at 0%, Weekly at 17%, and Fable at 16% with reset times, beside the send button

Before is the same composer with no circles and no popover — the meters are purely additive, and providers that report no usage still render exactly that.

No motion or transitions were added, so there is no video.

## Testing

Branched off `main` at `3d74474f`. Focused tests for the normalizers, the store's merge/debounce/announce behavior, the ingestion seam, the token reader, and the web components. Typecheck and lint clean across server, web, mobile, contracts, and client-runtime.

Two pre-existing failures in `ProviderRegistry.test.ts` and `server.test.ts` reproduce identically on unmodified `main` on my machine — they look like Windows-only arg-quoting artifacts in the spawn mocks (`Unexpected args: ^"--version^"`), unrelated to this change.

## Checklist

- [ ] This PR is small and focused — **it is not; stated plainly above**
- [x] I included a screenshot for the UI change
- [ ] I included a video for animation/interaction changes — no motion added

---

> [!NOTE]
> **Medium Risk**
> Large cross-cutting feature touching provider registry, OAuth credential reads, and macOS keychain subprocesses; failures are designed to degrade to empty meters rather than block turns.
> 
> **Overview**
> Adds **account quota usage** to the composer on web and mobile, driven by a new optional `usageLimits` field on `ServerProvider` and a `serverRefreshProviderUsage` RPC that schedules a background pull and returns immediately.
> 
> On the server, a **volatile in-memory store** merges full vs partial readings from turn-time `account.rate-limits.updated` events (Claude/Codex) and from **Claude-only on-demand pulls** (OAuth token from the CLI credentials path or macOS keychain, debounced per instance). **Codex** also seeds usage during status probe via `account/rateLimits/read` with a tight timeout. The **provider registry** decorates snapshots from the store, republishes on usage-only changes without disk writes, strips usage from status cache persistence, and **clears stored usage** when an instance is rebuilt or removed.
> 
> **Web** shows ring meters beside send (compact = worst window); **mobile** uses one gauge menu in the toolbar. `ControlPillMenu` gains a render-prop child so Android triggers stay tappable. Shared formatting/selection lives in `@t3tools/client-runtime/state/provider-usage`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 75a8a88532fbaca232ca548af9a747fd5d9d68b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add provider account usage meters to the composer input area
> - Adds per-provider usage windows (percent used, reset time) to the web and mobile composer toolbars, shown as circular ring meters that turn critical above 90%.
> - Introduces a server-side in-memory `ProviderUsageLimitsStore` that tracks usage per provider instance, populated via runtime turn events (`ProviderUsageIngestionLive`) and on-demand HTTP fetches (`ProviderUsageRefresher`) for Claude and Codex.
> - Adds a new `serverRefreshProviderUsage` WebSocket RPC that schedules a background usage refresh without blocking the caller; client-side calls are deduplicated with single-flight keying.
> - Usage is refreshed on composer mount, on window/app focus, and when the usage popover is opened; server-side refreshes are debounced to once per 60 seconds per instance.
> - Usage data is decorated onto `ServerProvider` snapshots in the registry and propagated to clients without disk persistence or waiting for full status probes.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 75a8a88.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/components/AppSymbol.tsx
+++ b/apps/mobile/src/components/AppSymbol.tsx
@@ -36,6 +36,7 @@ import {
   IconFolder,
   IconFolderOpen,
   IconFolderPlus,
+  IconGauge,
   IconGitBranch,
   IconHammer,
   IconGitMerge,
@@ -80,6 +81,7 @@ import { SymbolView as ExpoSymbolView, type SFSymbol, type SymbolViewProps } fro
 
 const ANDROID_ICON_BY_SF_SYMBOL: Partial<Record<SFSymbol, Icon>> = {
   "arrow.branch": IconGitBranch,
+  "gauge.with.dots.needle.bottom.50percent": IconGauge,
   "arrow.clockwise": IconRefresh,
   "arrow.down.circle": IconArrowDownCircle,
   "arrow.right.circle": IconArrowRightCircle,
--- a/apps/mobile/src/components/ControlPill.tsx
+++ b/apps/mobile/src/components/ControlPill.tsx
@@ -86,9 +86,18 @@ export function ControlPill(props: {
 // Android renders the token-styled AndroidAnchoredMenu, since the native
 // AppCompat popup can't be themed past its stock animation, metrics, and
 // submenu chrome.
+/**
+ * A render-function child keeps the trigger interactive on Android.
+ *
+ * The plain-node form there is wrapped in a `pointerEvents="none"` view so
+ * the anchor Pressable owns the tap, which silently swallows any `onPress`
+ * the child declared. Callers that need their own press side effects take
+ * `open` and call it themselves. On iOS the native menu opens on tap and the
+ * child's handlers already run, so `open` is a no-op there.
+ */
 export function ControlPillMenu(
   props: Omit<ComponentProps<typeof MenuView>, "children" | "themeVariant"> & {
-    readonly children: ReactNode;
+    readonly children: ReactNode | ((open: () => void) => ReactNode);
     readonly className?: string;
   },
 ) {
@@ -133,7 +142,12 @@ export function ControlPillMenu(
   }
 
   const { className: _className, ...menuProps } = props;
-  let children = menuProps.children;
+  // iOS opens the native menu from the tap itself, so the render-function
+  // form has nothing to trigger.
+  let children: ReactNode =
+    typeof menuProps.children === "function"
+      ? menuProps.children(() => undefined)
+      : menuProps.children;
   // In long-press mode the wrapped pressable still receives the touch (the
   // patched MenuView button is touch-transparent) and RN's Fabric touch
   // handler is never cancelled by the in-tree UIContextMenuInteraction, so a
--- a/apps/mobile/src/features/threads/ThreadComposer.tsx
+++ b/apps/mobile/src/features/threads/ThreadComposer.tsx
@@ -1,6 +1,7 @@
 import { isLiquidGlassSupported, LiquidGlassView } from "@callstack/liquid-glass";
 import type {
   EnvironmentId,
+  ProviderUsageWindow,
   MessageId,
   ModelSelection,
   OrchestrationThreadShell,
@@ -18,6 +19,7 @@ import type { ReactNode } from "react";
 import { memo, useCallback, useEffect, useMemo, useRef, useState, type RefObject } from "react";
 import {
   ActivityIndicator,
+  AppState,
   Image,
   Platform,
   Pressable,
@@ -51,7 +53,15 @@ import {
   ComposerToolbarScroller,
   ComposerToolbarTrigger,
 } from "../../components/ComposerToolbarTrigger";
-import { ControlPill } from "../../components/ControlPill";
+import { ControlPill, ControlPillMenu } from "../../components/ControlPill";
+import {
+  formatUsagePercent,
+  formatUsageResetLabel,
+  formatUsageUpdatedAtLabel,
+  pickWorstUsageWindow,
+} from "@t3tools/client-runtime/state/provider-usage";
+import { serverEnvironment } from "../../state/server";
+import { useAtomCommand } from "../../state/use-atom-command";
 import { ProviderIcon } from "../../components/ProviderIcon";
 import type { DraftComposerImageAttachment } from "../../lib/composerImages";
 import { buildModelOptions, groupByProvider } from "../../lib/modelOptions";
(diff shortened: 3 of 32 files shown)


gfsaaser24:
Pushed `55e6796a`. Four of the bot findings were real and are fixed; seven are wrong, with reasoning below so you don't have to re-derive it.

## Fixed

**Android usage menu never ran its own `onPress`** (Macroscope + Cursor Bugbot, independently — good catch). `AndroidAnchoredMenu` wraps a plain child in a `pointerEvents="none"` view under the anchor `Pressable`, so the trigger's handler was unreachable and opening the menu neither restamped the relative labels nor requested a reading. The same component's render-function branch keeps children interactive, so `ControlPillMenu` now accepts `(open) => ReactNode` as well and the usage trigger uses it. `open` is a no-op on iOS, which opens natively. Purely additive — existing `ReactNode` callers are untouched.

**Web popover attributed one provider's quota to another.** The meters read `selectedInstanceId` (which prefers the live model-picker choice) while the title read the thread's persisted provider, so between picking a model and sending, the new provider's numbers sat under the old provider's name. Added a display name derived from the same instance the windows come from.

**`claimRefreshSlot` wedged on a backwards wall clock.** `now - lastRefreshAt` goes negative after an NTP correction and read as "still inside the debounce window," blocking refreshes until the clock caught up. Now requires a non-negative elapsed. Regression test added with `TestClock.setTime`.

**`fetchClaudeUsage` timeout didn't cover the body.** The 10s bound wrapped `client.execute` only, so a 2xx response with a trickled body parked the refresh fiber indefinitely. The timeout now wraps execute-and-decode together.

## Not fixed, because the bots are wrong

**`Equal.equals(existing.windows, merged.windows)` "compares by reference"** — this is incorrect on effect 4. Two separate bots have now made this call, both reasoning from effect 3 semantics. In the pinned `effect@4.0.0-beta.103`, `Equal.ts` dispatches plain arrays to `compareArrays` and plain objects to `compareRecords`; the comparison is deep. The existing test *"does not announce a change when only the timestamp moved"* asserts exactly the behavior the bot claims is broken, and it passes. Nothing to change.

**Effect service conventions ×4** — the suggestion is to collapse the tag, construction, and layer into one canonical module and inline the `*Shape` interfaces. That is the opposite of the convention in the directory these services were added to: `apps/server/src/provider/` currently holds 14 services all built the same way, `Services/<Name>.ts` with the tag and a `<Name>Shape` interface, `Layers/<Name>.ts` with construction and `<Name>Live` — `ClaudeAdapter`, `CodexAdapter`, `ProviderRegistry`, `ProviderService`, `ProviderSessionReaper`, and so on. Adopting the bot's shape would make these two the only inconsistent services in the subtree. Happy to move the whole directory if you'd rather standardize the other way, but that is not this PR.

**"Refresh debounce burned without a token"** — the sequence is real but the current ordering is the safer one. Claiming the slot before reading credentials is what stops ambient UI triggers (window focus, thread switch, meter hover) from driving an unbounded number of macOS `security` keychain invocations while signed out, each of which can raise an interactive prompt. A signed-out user waiting up to 60s for meters to appear after logging in is the cheaper side of that trade.

Tests, typecheck, and lint are clean on the new commit. The two pre-existing failures I mentioned in the description still reproduce on unmodified `main` on my machine and are unrelated.

gfsaaser24:
revisions inbound. leave open please.

gfsaaser24:
Superseded by #6080 — the same feature re-cut as one clean commit on current `main`, with all the review-bot findings from this PR folded in. Closing to keep the history readable.

PR #5968, opened 2026-08-10
feat(server): add a Kiro CLI provider driver

NatoNathan:
Adds [Kiro CLI](https://kiro.dev) as a sixth built-in provider. Kiro speaks ACP, so this follows the Cursor/Grok shape: a new `KiroDriver` with its own probe, adapter, text generation, settings, docs, and web/mobile metadata. It ships **disabled by default** behind the existing "Early Access" badge, so nothing changes for anyone who does not turn it on.

`No ticket: upstream open-source contribution.`

## Why copy-adapt Grok instead of writing a generic ACP driver

That was the first design I tried, and I rejected it. A generic "point me at any ACP agent" driver cannot carry what a real driver needs: vendor extension notifications, an auth probe (`kiro-cli whoami`), a fast model catalogue probe, branding, and per-agent quirks. It also creates an untested matrix of agents we would implicitly claim to support. I would rather ship one driver that is actually exercised against the real CLI. I am happy to file a separate design issue for a generic ACP path if maintainers want it.

## Shared ACP layer changes

Three additive changes were needed. Each one leaves Cursor and Grok byte-identical in behaviour, and I verified that by running their suites (38/38).

1. **`AcpSessionRuntime.authMethodId` is now optional.** `kiro-cli` advertises `authMethods: []` and answers `authenticate` with `-32601`. The runtime previously called `authenticate` unconditionally, which would have failed startup. Cursor (`cursor_login`) and Grok both still pass an id and are unaffected. I deliberately did **not** swallow `-32601` globally, since that would change their behaviour.
2. **A `ReasoningDelta` parsed-event variant** plus `makeAcpReasoningDeltaEvent`. `agent_thought_chunk` was previously dropped by a `default: break`. I added a new variant rather than a `streamKind` field on `ContentDelta`, again to avoid touching Cursor/Grok.
3. **Kind-based permission helpers** (`selectAcpPermissionOptionIdByKind`, `selectAcpAutoApprovedPermissionOptionId`) exported from `AcpAdapterSupport.ts`. Permission option ids are agent-defined: Cursor uses dashed ids, Kiro uses `allow_once` / `allow_always` / `reject_once`. Grok already had a private copy of this logic; I left it alone since the instruction was not to refactor Grok, but a future shared-ACP-core pass should delete the duplicate.

Everything Kiro-specific lives in `acp/KiroAcpSupport.ts` (spawn, auth, models) and `acp/KiroAcpExtension.ts` (the `_kiro.dev/*` surface), so the shared layer stays vendor-neutral.

## What works

Model picker, with Kiro's live catalogue:

Model picker showing Kiro models

A real turn:

A completed Kiro turn

Approvals, using Kiro's own option ids:

Kiro command approval prompt

Approved command output

Settings, disabled by default until toggled (account email redacted):

Providers settings with Kiro enabled

Also working: streaming text and thinking, tool calls, interrupt, in-session model switching (Kiro honours `session/set_model`, so no new thread is needed unlike Grok), session resume via `session/load`, todo lists rendered as plans, and the four text-generation operations (commit message, PR content, branch name, thread title).

## Findings from the real CLI, and things I deliberately did not build

These are all recorded in code comments so the next person does not have to rediscover them.

- **`_kiro.dev/commands/execute` and `/commands/options` kill the agent process.** Clean exit 0, no JSON-RPC error, across seven payload shapes on 2.16.2. The driver never calls them. Typing `/tools` as ordinary prompt text works instead.
- **Slash commands are not surfaced.** `_kiro.dev/commands/available` arrives per session, but T3's only slash-command surface is the per-instance probe snapshot. Wiring a per-thread channel is new contracts + server + web plumbing, i.e. a separate PR. Typing commands already works today; there is just no autocomplete.
- **Usage is not mapped to token counts.** `_kiro.dev/metadata` reports a context-usage percentage and credit metering, not tokens, and `ThreadTokenUsageSnapshot` requires `usedTokens`. Deriving token counts from a percentage would invent numbers shown to users.
- **Elicitation is not wired.** I advertised both elicitation modes and `kiro-cli` never sent `elicitation/create` — it asks clarifying questions as plain assistant text. I removed the copied `pendingUserInputs` machinery rather than ship a handler that can never fire.
- **No plan/implement mode mapping.** Kiro's ACP "modes" are agent configs (`kiro_default`, `kiro_planner`, …), not interaction modes. Mapping `kiro_planner` onto plan mode would promise a restriction Kiro does not enforce, so agent choice stays a per-instance setting.
- **`KIRO_HOME` does not separate logins.** It scopes agents, settings, and sessions only; `whoami` still returns the real account. Multiple Kiro instances share one login, and the settings copy plus user docs say so.
- **`rollbackThread` reports unsupported**, exactly like Grok. A silent no-op would let a checkpoint revert restore files while the agent still remembers the reverted turns.

Follow-ups worth their own issues: a per-thread slash-command channel, a percentage/credit-shaped usage contract, and an MCP authorization prompt surface (`_kiro.dev/mcp/oauth/*` is logged as a warning for now — I could never trigger one, so a UI path would be untestable).

## Tested

- New suites: `KiroAdapter` (20), `KiroProvider` (11), `KiroAcpSupport` (17), `KiroAcpExtension` (14), `KiroTextGeneration` (6), `ProviderInstanceRegistryHydration` (4) — all against the shared mock agent, extended with default-off flags (`T3_ACP_FAIL_AUTHENTICATE`, `T3_ACP_EMIT_THOUGHT_CHUNKS`, `T3_ACP_EMIT_KIRO_TODOS`, `T3_ACP_MODEL_IDS`) so existing Cursor/Grok tests are untouched.
- Regression: Cursor + Grok adapters, ACP runtime/events/support, contracts settings, web provider settings — 116 passing on the rebased branch.
- `typecheck` clean for server, web, mobile, contracts. `vp lint` 0 warnings / 0 errors. `vp fmt --check` clean.
- Manual pass in the real web client against `kiro-cli` 2.16.2 with an IAM Identity Center login: everything in the screenshots above.

Not run: the repo-wide suite (CI owns it). One pre-existing unrelated failure I did not touch: `apps/web/src/terminal/ghostty/runtimeAbi.test.ts` cannot import `ghostty-vt.wasm?inline`.

## Not done on purpose

`apps/marketing` is untouched. Adding Kiro to the hero needs a brand mark and marketing copy, which felt like a maintainer decision rather than something to guess at while the provider is early access and off by default. Happy to add it if you tell me what you want there.

---

Written with Kiro CLI (Claude Opus 5) driving T3 Code.

> [!NOTE]
> ### Add Kiro CLI as a built-in provider driver over ACP
> - Adds a new `kiro` provider driver ([KiroDriver.ts](https://github.com/pingdotgg/t3code/pull/5968/files#diff-cde6efc51f62dda368d05dfd9a74068357637644276fc99008574e6f6b9def26)) that spawns `kiro-cli acp` as an ACP subprocess, handling session lifecycle, model selection, permission approvals, and event streaming.
> - Implements provider health probing ([KiroProvider.ts](https://github.com/pingdotgg/t3code/pull/5968/files#diff-5146d5c53327286c986136e51e4eea966db561c11b0806bd84787d790159e242)) via `kiro-cli --version`, `whoami`, and `chat --list-models`, publishing a model catalog with fallback to an `auto` model.
> - Adds `KiroSettings` schema with `binaryPath`, `homePath`, `agent`, and `customModels` fields; `enabled` defaults to `false`.
> - Adds a `KiroTextGeneration` service that drives commit message, PR content, branch name, and thread title generation over a short-lived ACP session with a 180s timeout, excluding reasoning chunks from output.
> - Registers Kiro in model aliases, provider display names, web provider picker, mobile icon, and settings UI with an Early Access badge.
> - Extends the ACP mock agent to support configurable model IDs, optional todo/thought-chunk emission, and simulated `authenticate` failures for testing.
> - Risk: `authenticate` is now skipped entirely when no auth method is advertised by the agent, changing ACP startup behavior for all ACP-based drivers.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized ef5fb69. 29 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>


ankitcharolia:
@NatoNathan did you try kiro provider integration with agent version 3 ? Just asking because they have changed whole authentication mechanism from v3 and that would break kiro provider in t3code 

Try this ACP compatible Kiro gateway : https://github.com/ankitcharolia/kiro-gateway

t3dotgg:
> [!NOTE]

Closing this PR after an automated pass over open pull requests. Adds an unsupported first-party provider or source-control host without a maintainer request.

PR #5996, opened 2026-08-10
fix(web): prevent Windows titlebar shift after restore

thiscallnet:
## What Changed

Stabilizes the Windows desktop titlebar layout across minimize → restore transitions.

The last valid native window-controls geometry is preserved while the Window Controls Overlay is temporarily unavailable during restore, then refreshed once the restored window has settled.

Diff, Usage, Settings, and right-panel headers now use the same stabilized titlebar geometry. Other platforms keep the existing WCO behavior.

## Why

On Windows desktop, restoring T3 Code after minimizing could briefly render titlebar content in the wrong position before snapping into place, including content shifting into the native minimize/maximize/close button area.

Preserving the last known-good geometry through the restore transition avoids rendering that intermediate layout while still allowing the native controls geometry to update normally afterwards.

### Before

https://github.com/user-attachments/assets/add6d3e6-59b6-431e-9efc-2f6e36fd3e00

### After

https://github.com/user-attachments/assets/9754c964-5274-41f0-baf4-b2d785902eaa

## Checklist

---

> [!NOTE]
> **Low Risk**
> Desktop UI/layout only on Windows Electron; no auth, data, or API changes. Risk is limited to titlebar geometry regressions on other platforms if WCO sync behavior diverges.
> 
> **Overview**
> Fixes a **Windows desktop** glitch where titlebar content jumped or overlapped the system buttons briefly after **minimize → restore**.
> 
> **`syncDocumentWindowControlsOverlayClass`** now keeps the last good native control geometry on Windows when the Window Controls Overlay is temporarily unavailable during restore: it can leave the **`wco`** class applied, writes **`--workspace-native-controls-width`** from **`getTitlebarAreaRect`**, and skips that preservation in fullscreen via new **win32** fullscreen IPC from **`DesktopWindow`**.
> 
> **CSS** adds **`.electron-windows`** / **`.wco`** variables (fixed **40px** topbar height, native-control insets) so layout does not depend on transitional **`env(titlebar-area-*)`** values during restore.
> 
> **Headers** (diff panel, settings, usage, right-panel sheet) switch from hardcoded **`h-[52px]`** / **`env(...)`** padding to **`workspace-topbar`** and **`--workspace-native-controls-inset`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 021bfb1ff3caab200ba14e426525880177690ca9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix Windows titlebar shift after restore by preserving WCO layout during overlay hide
> - Revamps `syncDocumentWindowControlsOverlayClass` in [`windowControlsOverlay.ts`](https://github.com/pingdotgg/t3code/pull/5996/files#diff-0b4bc2286c6db8c805258eaec400a41bb45075d438330415672e821a0c99d9bf) to track fullscreen state and overlay geometry separately, keeping the `wco` class active on Windows when the overlay becomes temporarily hidden (e.g. after restore from maximized) to prevent layout shift
> - Adds CSS variables (`--workspace-topbar-height`, `--workspace-native-controls-inset`, etc.) under `.electron-windows` in [`index.css`](https://github.com/pingdotgg/t3code/pull/5996/files#diff-70ad657a634ce55957729ff8d1c93f697cc7e3ca2a5cb5fe6b00c90d7d515fcd) so components can size/pad consistently without hardcoded heights or `calc()` expressions
> - Extends full-screen event listeners in [`DesktopWindow.ts`](https://github.com/pingdotgg/t3code/pull/5996/files#diff-d92229c509071d1e9d3595d187fe2f1f13c394b22e1d9d2070ed6fb99678f429) to cover `win32` in addition to `darwin`, so the renderer knows when to drop the preserved `wco` class during fullscreen
> - Updates header rows in diff panel, settings, usage, and project settings pages to use `workspace-topbar` and `var(--workspace-native-controls-inset)` instead of fixed heights and calc-based padding
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 021bfb1.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/window/DesktopWindow.ts
+++ b/apps/desktop/src/window/DesktopWindow.ts
@@ -553,7 +553,7 @@ export const make = Effect.gen(function* () {
       runFork(flushBoundsPersist);
     });
 
-    if (environment.platform === "darwin") {
+    if (environment.platform === "darwin" || environment.platform === "win32") {
       window.on("enter-full-screen", () => {
         window.webContents.send(WINDOW_FULLSCREEN_STATE_CHANNEL, true);
       });
--- a/apps/web/src/components/DiffPanelShell.tsx
+++ b/apps/web/src/components/DiffPanelShell.tsx
@@ -13,7 +13,7 @@ function getDiffPanelHeaderRowClassName(mode: DiffPanelMode) {
     "flex items-center justify-between gap-2",
     mode === "embedded" ? "px-2" : "px-4",
     shouldUseDragRegion
-      ? "drag-region h-[52px] border-b border-border wco:h-[env(titlebar-area-height)] wco:pr-[calc(100vw-env(titlebar-area-width)-env(titlebar-area-x)+1em)]"
+      ? "workspace-topbar drag-region border-b border-border wco:pr-[var(--workspace-native-controls-inset)]"
       : "surface-subheader",
   );
 }
--- a/apps/web/src/components/settings/ProjectSettingsPanel.tsx
+++ b/apps/web/src/components/settings/ProjectSettingsPanel.tsx
@@ -187,7 +187,7 @@ export function ProjectSettingsPage({ projectKey }: { projectKey: string }) {
         {isElectron && (
           <div
             className={cn(
-              "drag-region flex h-[52px] shrink-0 items-center px-5 transition-[padding-left] duration-200 ease-linear motion-reduce:transition-none wco:h-[env(titlebar-area-height)] wco:pr-[calc(100vw-env(titlebar-area-width)-env(titlebar-area-x)+1em)]",
+              "workspace-topbar drag-region px-5 transition-[padding-left] duration-200 ease-linear motion-reduce:transition-none wco:pr-[var(--workspace-native-controls-inset)]",
               COLLAPSED_SIDEBAR_TITLEBAR_INSET_CLASS,
             )}
           >
--- a/apps/web/src/components/usage/UsagePage.tsx
+++ b/apps/web/src/components/usage/UsagePage.tsx
@@ -81,7 +81,7 @@ export function UsagePage() {
         {isElectron && (
           <div
             className={cn(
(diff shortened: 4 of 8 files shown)


CDVolvik (commented):
The `.electron-windows` height override lands, but it wins on source order rather than specificity, and that is worth making explicit.

```css
.wco               { --workspace-topbar-height: env(titlebar-area-height, 52px); }  /* ~line 121 */
.electron-windows  { --workspace-topbar-height: 40px; }                             /* ~line 1407 */
```

Both are single-class selectors, so they tie on specificity and the later one wins. That is the intent, since the comment says the WCO height is transitional during restore, which is the whole bug. But it means the fix depends on the `.electron-windows` block staying below the `.wco` block in the file. Move it, split the stylesheet, or have a bundler reorder layers, and the value silently reverts to `env(titlebar-area-height, 52px)`, which is the transitional value this PR exists to avoid. Writing it as `.electron-windows.wco` for the height too would make it two classes to one and take the ordering out of it.

Second thing, the 40 is duplicated across languages:

```ts
const TITLEBAR_HEIGHT = 40;   // apps/desktop/src/window/DesktopWindow.ts:23
```
```css
/* Matches DesktopWindow TITLEBAR_HEIGHT; WCO height can be transitional during restore. */
--workspace-topbar-height: 40px;
```

The comment is accurate today, and nothing keeps it accurate. Change `TITLEBAR_HEIGHT` and the CSS drifts silently, and the symptom that comes back is the exact shift this PR fixes. Since the main process already sets the `electron-windows` class, it could set the height as a custom property at the same time and delete the literal, or failing that a test asserting the two agree would at least make the drift loud.

Third, worth calling out in the description rather than changing: swapping the hardcoded `h-[52px]` for `workspace-topbar` in `DiffPanelShell`, `ProjectSettingsPanel` and `UsagePage` also pulls in the new 40px on Windows, and every element already using `workspace-topbar` goes 52 to 40 on Windows as well. ChatView's topbar, settings, usage, pull requests, `NoActiveThreadState` and the sidebar chrome header all shift by 12px. That is presumably the point, since 40 is the real titlebar height, but a reader of the title would expect a restore-time fix rather than a Windows-wide 12px change to every top bar, and reviewers checking screenshots should know to look beyond the restore case.

Adding `win32` to the `enter-full-screen` / `leave-full-screen` wiring looks straightforwardly right.

PR #6001, opened 2026-08-10
feat(threads): pin a thread from the row quick actions

inayayousfi:
Pinning took a context menu on web and a long-press on mobile, while snooze sat one click away on both. This closes that gap on the row itself.

## Web

An icon-only pin button joins the card's hover cluster, leftmost, so the order reads pin, snooze, Settle. Settle keeps the right edge it has always had and snooze keeps its slot, so neither learned target moves.

It shows only on unpinned cards. A pinned card already carries its own unpin glyph beside the project label, so a second control there would be two ways to say one thing. Slim rows are untouched: the cluster lives in the card branch only.

The button reuses the sidebar's existing pin command, so it stays silent on success and toasts only on failure, exactly as the context menu entry does.

## Mobile

The swipe tray took a `primaryAction` / `secondaryAction` prop pair and a width constant fixed at two slots. It now takes an ordered `actions` list, so it can hold three buttons.

The inbox card's tray becomes Pin, Settle, Snooze. Pin is appended on the **inside**, so Settle and Snooze keep their existing distance from the screen edge and no learned gesture moves. A full swipe still commits the lifecycle action and never pins.

Pin rides the card only. Slim rows are settled or snoozed, and pinning one clears that state server-side, which is too big a jump for a swipe.

## Risk, and what bounds it

The tray is shared with the v1 thread list and the archive screen, neither of which gains an action here. The reveal stagger used to be two hardcoded ranges; it is now a formula over the slot index whose one and two action cases reproduce those exact numbers. `thread-swipe-layout.test.ts` pins them down, so a future change to the formula cannot silently shift the two lists that were not meant to move.

The layout maths moved into `thread-swipe-layout.ts` so it can be asserted without loading React Native.

## Verification

- `tsc --noEmit` clean in both apps
- `vp test run` in mobile: 102 files, 633 tests passing
- `vp test run --project unit` in web: 221 files, 2002 tests passing
- `vp lint` clean

---

> [!NOTE]
> **Medium Risk**
> Touches the shared mobile swipe component and every list that uses it, but two-action geometry is regression-tested and pin/full-swipe behavior is explicitly gated so existing gestures should stay stable.
> 
> **Overview**
> **Pin** is exposed on the thread row without opening a menu: web sidebar cards get a leftmost pin icon in the hover cluster (only when pinning is supported and the thread is not already pinned), wired to the existing `attemptPin` flow. On mobile Thread List v2 **card** rows, swipe reveals **Pin · Settle · Snooze** when eligible—pin is inserted innermost so settle/snooze stay at the same distance from the screen edge and a full swipe still commits the lifecycle action, not pin.
> 
> **`ThreadSwipeable`** no longer takes `primaryAction`, `secondaryAction`, `onDelete`, or `threadTitle`. Callers pass an ordered **`actions`** array (each action carries its own **`backgroundColor`**) plus optional **`fullSwipeIndex`**. Delete is built via **`deleteSwipeAction`** so label/color/a11y stay consistent. Tray width and reveal stagger move to **`thread-swipe-layout.ts`**, with **`thread-swipe-layout.test.ts`** locking the one- and two-action geometry so archive and v1 lists do not shift when a third slot exists.
> 
> **`resolveThreadListV2SwipeActions`** now returns **`actions`** and **`fullSwipeIndex`** (and still **`secondary`** for snooze menu wiring); pinning rules are covered by new unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f21a3fe84e342e980af39d3424891294df7837f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add pin quick action to thread row swipe tray and sidebar
> - Adds a Pin action to the swipe tray in V2 thread list rows when pinning is supported, the variant is 'card', and the thread is not already pinned; the pin action appears as the innermost (rightmost) swipe button.
> - Adds a 'Pin thread' quick-action button to sidebar thread rows in [Sidebar.tsx](https://github.com/pingdotgg/t3code/pull/6001/files#diff-12c9f3c619e6ab8dda238e965323ac96e62f3084e3adfb0671b1a6459aa61ba1), visible only on supported servers for unpinned threads.
> - Redesigns the `ThreadSwipeable` API to accept an ordered `actions` array and a `fullSwipeIndex` instead of separate primary/secondary props, allowing an arbitrary number of swipe actions with consistent per-action background colors.
> - Extracts swipe tray geometry into a new [thread-swipe-layout.ts](https://github.com/pingdotgg/t3code/pull/6001/files#diff-2f0ac68b81ce1572411a4d9ed6f1ffba83952a3909ae0ab1c94e09e971ca71ea) module and standardizes action colors via new exports in [thread-swipe-actions.tsx](https://github.com/pingdotgg/t3code/pull/6001/files#diff-49d535c68199a21330fbd5e7c019f130a9b0b760aec90d7ea8a45cad38789eb9).
> - Behavioral Change: all existing swipe rows (archived, v1 list, v2 list) are migrated to the new `actions` array API; `onDelete`, `primaryAction`, `secondaryAction`, and `threadTitle` props are removed from `ThreadSwipeable`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f21a3fe.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/features/archive/ArchivedThreadsScreen.tsx
+++ b/apps/mobile/src/features/archive/ArchivedThreadsScreen.tsx
@@ -28,7 +28,11 @@ import { ProjectFavicon } from "../../components/ProjectFavicon";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
 import { relativeTime } from "../../lib/time";
 import { useThemeColor } from "../../lib/useThemeColor";
-import { ThreadSwipeable } from "../home/thread-swipe-actions";
+import {
+  deleteSwipeAction,
+  SWIPE_ACTION_LIFECYCLE_COLOR,
+  ThreadSwipeable,
+} from "../home/thread-swipe-actions";
 import {
   createNativeMailSearchToolbarItem,
   NATIVE_MAIL_SEARCH_TOOLBAR_SUPPORTED,
@@ -421,18 +425,23 @@ function ArchivedThreadRow(props: {
         borderBottomRightRadius: props.isLast ? 20 : 0,
         overflow: "hidden",
       }}
+      actions={[
+        {
+          accessibilityLabel: `Unarchive ${props.thread.title}`,
+          backgroundColor: SWIPE_ACTION_LIFECYCLE_COLOR,
+          icon: "arrow.uturn.backward",
+          label: "Unarchive",
+          onPress: props.onUnarchive,
+        },
+        deleteSwipeAction({ onDelete: props.onDelete, threadTitle: props.thread.title }),
+      ]}
+      // Delete sits at the screen edge and is what a full swipe commits, as
+      // it did when the tray built that action implicitly.
+      fullSwipeIndex={1}
       fullSwipeWidth={windowWidth - 32}
-      onDelete={props.onDelete}
       onSwipeableClose={props.onSwipeableClose}
       onSwipeableWillOpen={props.onSwipeableWillOpen}
-      primaryAction={{
-        accessibilityLabel: `Unarchive ${props.thread.title}`,
-        icon: "arrow.uturn.backward",
-        label: "Unarchive",
-        onPress: props.onUnarchive,
-      }}
       simultaneousWithExternalGesture={props.simultaneousSwipeGesture}
-      threadTitle={props.thread.title}
     >
       {() => (
         <View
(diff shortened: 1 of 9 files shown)


t3dotgg:
> [!NOTE]

Closing this PR after an automated pass over open pull requests. Thread pin and unpin actions already exist in current sidebar menus.

PR #6002, opened 2026-08-10
feat(threads): empty-inbox block with a new-thread button

### LINKED ISSUE #10207: [Bug]: I cannot add a tread (New session/conversation) if the project is Empty
### Before submitting

### Area

apps/desktop

### Steps to reproduce

If i create a new Workspace / project and i do not start chating and go to another project. There's no way to start a chat in the project created without a way to add a conversation. (If we can it's very not obvious how to do it).
It's look like we cannot start a new conversation in the project. Why? If the conversation is too big and we want to start a new one what do we do? Start a new project? I really do not understand the logic behind this. Just add a File+ button to add as new session we want in a project

### Expected behavior

adding tread / session in any project. 
Just add a File+ button to add as new session we want in a given project

### Actual behavior

Just very simple UI and logical behavior like adding conversations. 
teste it by Opening un new project and go to another project without beginning a conversation

### Impact

Blocks work completely

### Version or commit

the latest today

### Environment

Windows 11

### Logs or stack traces

```shell

```

_No response_

### Workaround

_No response_

inayayousfi:
Clearing your inbox showed nothing. The old empty state only appeared when every group was empty at once, so a cleared inbox with any settled or snoozed history rendered as blank space.

## What lands

The block takes the inbox's own slot, above the snoozed and settled shelves, so those keep rendering underneath it. On mobile that means it is a list row, not a `ListEmptyComponent`: the list is rarely empty even when the inbox is.

It appears when the inbox, the pinned block and the drafts are all empty. Pinned threads and drafts count as live work, since the block would contradict itself sitting under them.

## The copy

- `You're all caught up`, or `You're all caught up in <project>` when the list is filtered to one
- `Everything's handled. Enjoy the quiet.` renders **only** when something is actually settled or snoozed, so a brand-new project is not congratulated on work it never had
- Button: `New thread`

Both clients suppress the block while a search is running. Zero matches is not the same statement as an empty inbox, and the existing search states already cover it.

## One home for the strings

The rule and all three strings live in `packages/client-runtime/src/state/threadInbox.ts`, exported as `state/thread-inbox`. That sits beside `state/thread-settled`, which already shares a display helper (`snoozeWakeLabel`) with both clients.

Neither app spells a string out. That duplication is exactly what let the old `"No threads yet"` copy end up written in four separate files.

## Per client

**Web** replaces its two thread-copy branches with the block. `"No projects yet"` keeps its own state at the foot of the list, since with no projects there is nothing to create a thread in.

**Mobile** renders it on both the phone list and the tablet sidebar. The button opens the new-task project picker rather than guessing, because an unfiltered list has no project to target. The tablet sidebar gains an `onStartNewTask` prop, threaded from the adaptive layout, since it had no new-task entry point at all.

## Verification

- `tsc --noEmit` clean in both apps
- `vp run --filter @t3tools/client-runtime test`: 47 files, 599 tests passing, including 9 new ones covering the rule and all four copy branches
- `vp test run` in mobile: 101 files, 629 tests passing, including 6 new ones covering placement above the shelves and every case that withholds the block
- `vp test run --project unit` in web: 221 files, 2002 tests passing
- `vp lint` clean
- Each new string greps to exactly one production-code location

---

> [!NOTE]
> **Low Risk**
> UI-only empty-state and list composition changes with shared copy; no auth, data, or API changes. Risk is mainly wrong gating (search/loading) or CTA navigation edge cases.
> 
> **Overview**
> When the inbox is clear but snoozed or settled history remains, thread lists no longer show blank space. A new **empty-inbox** row/block sits in the inbox slot **above** snoozed and settled shelves, with a **New thread** action.
> 
> **Shared logic** in `client-runtime` (`state/thread-inbox`): `isInboxClear` (active + pinned + drafts must be zero; parked shelves do not count) and `emptyInboxCopy` (scoped headline, optional “parked” detail only when something is snoozed/settled).
> 
> **Web** adds `SidebarEmptyInboxBlock` and drops the old “No threads yet” / per-project empty copy in favor of this block; **no projects** still uses the foot “Add project” state.
> 
> **Mobile** v2 list builder accepts optional `emptyInbox` and inserts `v2-empty-inbox`; `ThreadListV2EmptyInbox` renders it on home and tablet sidebar. The block is **suppressed** during search, while connections load, or with zero projects. Tablet sidebar gets `onStartNewTask` → `NewTaskSheet` project picker when there is no scoped project.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb11bc179244f3f42a5a9b0ed867f9f7b4461436. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add empty-inbox block with a new-thread button to thread list sidebar and home screen
> - Adds a new `ThreadListV2EmptyInbox` component (mobile) and `SidebarEmptyInboxBlock` component (web) that render a headline, optional parked-items detail, and a "New thread" CTA button when the inbox is clear.
> - Introduces a shared [`threadInbox`](https://github.com/pingdotgg/t3code/pull/6002/files#diff-7c7138602ad056cc1912db363282c5f07b8ef103f9adbaeea08b718c87b77cb8) module in `client-runtime` with `isInboxClear`, `emptyInboxCopy`, and related constants used by both platforms.
> - The empty-inbox block is inserted into the list above snoozed/settled items when there are no active, pinned, or draft threads, and is suppressed during search, while connections are loading, or when no projects exist.
> - Copy is project-scoped when a single project is selected, and the detail line reflects whether parked (snoozed/settled) items exist.
> - The new-thread action navigates to the `NewTaskSheet` project picker on mobile and calls `onNewThread` on web.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bb11bc1.</sup>
> 
>

Closes #10207

### DIFF
--- a/apps/mobile/src/features/home/HomeScreen.tsx
+++ b/apps/mobile/src/features/home/HomeScreen.tsx
@@ -721,12 +721,22 @@ export function HomeScreen(props: HomeScreenProps) {
         settledShelfExpanded,
         settledShelfHeaderIndex: threadListV2Layout.settledShelfHeaderIndex,
         snoozeLabelNow: `${nowMinute}:00.000Z`,
-        // Suppressed while searching: no matches is not the same statement
-        // as an empty inbox, and the search states already cover it.
-        emptyInbox: hasSearchQuery ? null : { projectName: v2ScopedProjectGroup?.title ?? null },
+        // Same gate as the tablet sidebar, kept identical on purpose. This
+        // screen also returns a full-page empty state before the list ever
+        // renders, which already covers loading, but the block must not
+        // depend on that guard staying where it is:
+        //   - searching, where no matches is not "you are all caught up"
+        //   - loading, where an empty list means "not yet", not "all done"
+        //   - no projects, where its button would open an empty picker
+        emptyInbox:
+          hasSearchQuery || props.catalogState.isLoadingConnections || props.projects.length === 0
+            ? null
+            : { projectName: v2ScopedProjectGroup?.title ?? null },
       }),
     [
       hasSearchQuery,
+      props.catalogState.isLoadingConnections,
+      props.projects.length,
       settledShelfExpanded,
       snoozedShelfExpanded,
       threadListV2Layout,
--- a/apps/mobile/src/features/home/HomeScreen.tsx
+++ b/apps/mobile/src/features/home/HomeScreen.tsx
@@ -46,6 +46,7 @@ import {
   ThreadListV2PendingRow,
   ThreadListV2Row,
   ThreadListV2SettledShelfHeader,
+  ThreadListV2EmptyInbox,
   ThreadListV2SnoozedShelfHeader,
 } from "../threads/thread-list-v2-items";
 import {
@@ -720,8 +721,18 @@ export function HomeScreen(props: HomeScreenProps) {
         settledShelfExpanded,
         settledShelfHeaderIndex: threadListV2Layout.settledShelfHeaderIndex,
         snoozeLabelNow: `${nowMinute}:00.000Z`,
+        // Suppressed while searching: no matches is not the same statement
+        // as an empty inbox, and the search states already cover it.
+        emptyInbox: hasSearchQuery ? null : { projectName: v2ScopedProjectGroup?.title ?? null },
       }),
-    [settledShelfExpanded, snoozedShelfExpanded, threadListV2Layout, v2PendingTasks],
+    [
+      hasSearchQuery,
+      settledShelfExpanded,
+      snoozedShelfExpanded,
+      threadListV2Layout,
+      v2PendingTasks,
+      v2ScopedProjectGroup,
+    ],
   );
 
   const renderV2Item = useCallback(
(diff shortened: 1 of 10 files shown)


inayayousfi:
Fixed in ae9b3fdf. Both bots were right, and it was a real user-visible bug.

The block is a list row, so it makes the list non-empty and suppresses `ListEmptyComponent`. On the tablet sidebar that component owns the `Loading threads…` state, so a cold start showed `You're all caught up` with a `New thread` button before a single thread had loaded.

The gate now excludes all three cases that component still owns: searching, loading, and no projects. The last one also stops the button opening a project picker with nothing in it, which web already special-cased and the tablet sidebar did not.

The phone screen was already safe, since it returns a full-page empty state before the list renders at all. It gets the same gate anyway, so the rule reads identically on both surfaces rather than depending on that unrelated guard staying where it is.

One thing the bots did not raise, worth stating plainly: the **web** sidebar has no loading signal in scope at all, so it cannot be gated the same way. That window is pre-existing shape rather than a regression here, since the old `No threads yet` copy had the identical exposure, but this change does make the wording during it more confident. Plumbing workspace loading state into the web sidebar felt like the wrong thing to bolt onto this PR. Happy to do it here or in a follow-up, whichever you prefer.

Issue #5978, opened 2026-08-10
[Bug]: second-instance handler drops deep-link URL, breaking SSO login when app is already running


spundone:
## Description

When T3 Code is already running and a deep link (`t3code://...`) is opened — e.g. following an SSO login callback from the browser — the link is silently swallowed. The app window is focused/revealed, but the URL itself is never processed, so login never completes.

## Root cause

In `apps/desktop/src/app/DesktopClerk.ts`:

```ts
yield* electronApp.on("second-instance", () => {
  void runPromise(
    Effect.gen(function* () {
      const mainWindow = yield* electronWindow.currentMainOrFirst;
      if (Option.isSome(mainWindow)) {
        yield* electronWindow.reveal(mainWindow.value);
      }
    }),
  );
});
```

Electron's `second-instance` event is emitted as `(event, argv, workingDirectory)`. `argv` contains the command-line arguments of the second launch, including any deep-link URL (e.g. the SSO callback). This handler ignores that argument entirely — it only reveals the main window. As a result, any protocol URL delivered while the app is already running (SSO callbacks, OAuth redirects, etc.) is dropped on the floor.

This only works today when the app is *not* already running, since in that case the URL arrives as a normal startup argument and gets processed on initial launch — which is why the failure mode looks intermittent ("it worked once, then stopped").

## Repro

1. Launch T3 Code, leave it running.
2. From a terminal, run `xdg-open "t3code://app/anything"` (or trigger any real deep link, e.g. an SSO callback from the browser).
3. Observe: a second Electron process briefly starts, logs `before-quit received`, and exits. The already-running window is focused (if not already focused) but nothing else happens — the URL is never acted on.
4. Confirmed via desktop trace logs: no `second-instance`, `protocol`, or the invoked URL ever appears anywhere in `desktop.trace.ndjson` for that session.

Workaround: fully quit the app before clicking an SSO/login link, so it launches fresh and receives the URL as a startup argument instead of via `second-instance`.

## Suggested fix

Parse the deep-link URL out of `argv` in the `second-instance` handler (same extraction logic used for the initial-launch argv) and dispatch it through whatever handler processes protocol URLs at startup, in addition to revealing the window.

## Secondary/related observation

[...]

Issue #4625, opened 2026-07-27
[Feature]: Deep link (t3code://checkout-pr) to open the PR-checkout dialog from outside the app


achtan:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I am describing a concrete problem or use case, not just a vague idea.

### Area

apps/desktop

### Problem or use case

My old terminal workflow was a one-liner: paste a PR link or branch name → land in a worktree for it. T3 Code already has the in-app equivalent (paste a PR URL into the BranchToolbar branch selector → `PullRequestThreadDialog` → local/worktree mode), which is great — but it can only be triggered by clicking through the UI on a fresh draft thread.

There is no outside-in entry point. Coming from a PR page in the browser, a Slack message, or a terminal, I can't hand T3 Code a PR reference and have it open the checkout flow. I have to switch to the app, find/create a draft thread, open the branch picker, and paste.

### Proposed solution

A deep link that opens the PR-checkout flow with the reference pre-filled, e.g.:

```
t3code://checkout-pr?ref=https://github.com/org/repo/pull/123
```

Behavior: focus/launch the app, open (or create) a draft thread in the current project, and open `PullRequestThreadDialog` with `initialReference` set — the user just picks local vs. worktree.

Most of the plumbing already exists:

- The Electron app already registers a protocol client (`setAsDefaultProtocolClient` in `apps/desktop/src/electron/ElectronApp.ts`) — missing only the URL routing for this action.
- `PullRequestThreadDialog` already accepts an `initialReference` prop.
- Reference parsing (`apps/web/src/pullRequestReference.ts`) already normalizes GitHub/GitLab/Azure URLs, `#123`, and even pasted `gh pr checkout 123` commands.
- The backend RPC `git.preparePullRequestThread` does the resolution + branch + worktree work.

### Why this matters

Enables composition with everything outside the app: browser extensions, Raycast/Alfred, and trivial shell functions like `w() { open "t3code://checkout-pr?ref=$1"; }`. For PR-review-heavy workflows, "paste link → worktree thread" becomes one action instead of five clicks, and T3 Code stays the owner of worktree lifecycle (no external `git worktree` juggling, avoids #1714-style mismatches).

### Smallest useful scope

Handle the single deep-link route: parse `ref`, focus the app, open the PR dialog pre-filled in the active project's draft thread. No CLI, no branch-name variant, no thread auto-start — the dialog's existing UX handles the rest.

### Alternatives considered

[...]

PR #6008, opened 2026-08-10
feat(desktop): open a thread from a t3code:// deep link

### LINKED ISSUE #4996: [Feature]: Open a specific desktop thread via t3code://
### Before submitting

### Area

apps/desktop

### Problem or use case

T3 Code already exposes thread deep links as `/threads/<environmentId>/<threadId>`, and the mobile app uses `t3code://threads/<environmentId>/<threadId>` for widgets and notification navigation.

On macOS, T3 Code is registered as the handler for the `t3code` scheme, but opening such a URL does not currently navigate the desktop app to the matching thread. The desktop app can be brought to the foreground, but the user still has to manually locate the conversation.

This prevents integrations such as a hardware controller, menu-bar status app, notification action, or other local automation from taking the user directly to the thread that needs attention.

### Proposed solution

Support a documented desktop deep link for an existing thread:

```
t3code://threads/<environmentId>/<threadId>
```

When received, the desktop app should reveal/focus its main window and navigate through the existing typed renderer route:

```
/$environmentId/$threadId
```

With Electron's hash history this is represented as `#/<environmentId>/<threadId>` internally; the external/mobile contract can remain `t3code://threads/...`.

The behavior should work when the app is already running and when it is launched through the link.

### Why this matters

This brings desktop behavior in line with the existing mobile deep-link model and enables fast "attention -> exact conversation" workflows.

For example, I am building a small read-only MIDI controller integration: each pad represents a live T3 thread and its LED reflects its state. Pressing a pad can already focus T3 Code on macOS; opening the corresponding thread would make the interaction immediate and useful without relying on brittle GUI scripting or storing credentials.

### Smallest useful scope

- Handle `open-url` on macOS.
- Handle the protocol URL passed to a second desktop instance on Windows/Linux.
- Parse only the `t3code://threads/<environmentId>/<threadId>` shape.
- Reveal the main window and forward the validated destination to the renderer.
- Navigate via the existing thread route after the renderer is ready.
- Ignore malformed or unsupported URLs safely.

No server API or authentication change is required.

### Alternatives considered

- AppleScript/UI automation to search for and click a thread: fragile, locale-dependent, and requires Accessibility permission.
- Focusing the app only: this works today but still leaves the user to locate the correct chat.
- Opening the remote web UI: does not provide the native desktop workflow.

### Risks or tradeoffs

The implementation needs to handle cold-start timing, duplicate URL events, and invalid or unavailable environment/thread IDs. The deep-link format should be treated as a stable public contract once documented.

The desktop renderer uses hash history, so the handoff should navigate through the renderer rather than assuming arbitrary external `t3code://` URLs are already renderer routes.

### Examples or references

- `packages/shared/src/agentAwareness.ts` builds `/threads/<environmentId>/<threadId>`.
- `apps/mobile/src/widgets/AgentActivity.tsx` turns that route into `t3code://threads/<environmentId>/<threadId>`.
- `apps/web/src/main.tsx` uses hash history under Electron.
- `apps/desktop/src/app/DesktopClerk.ts` already handles a second desktop instance by revealing the main window.
- Related: #4625 proposes a different desktop deep-link (`t3code://checkout-pr`) and identifies the same missing external URL-routing layer.

### Contribution

edwluo:
## Summary

- **Problem**: `t3code://threads/<environmentId>/<threadId>` links are already produced by `buildAgentAwarenessDeepLink` and consumed by the mobile app, but opening one on desktop does nothing beyond (at most) focusing the window — the user still has to find the thread by hand.
- **Why it matters**: it blocks exactly the integrations #4996 asks for — a menu-bar app, notification action, hardware controller or any local automation that wants to take the user straight to the thread that needs attention.
- **What changed**: the main process now listens for `open-url` (macOS) and reads the URL out of `argv` (Windows/Linux, both cold start and `second-instance`), validates it, and hands a typed target to the renderer, which navigates through the existing `/$environmentId/$threadId` route.
- **What did NOT change (scope boundary)**: no changes to Clerk/OAuth handling, to `DesktopClerk`'s existing `second-instance` reveal listener, to the deep-link *producer*, to mobile, or to the packaged Linux `.desktop` file. The external URL contract is unchanged — this only teaches desktop to consume what mobile already consumes.

## On CONTRIBUTING

I read `CONTRIBUTING.md` before opening this, so to address its points directly:

- **Issue first**: #4996 has been open since 30 July asking for exactly this, filed by someone else. This is an implementation of an existing request, not a drive-by feature or a scope expansion.
- **Size**: 430 lines, of which ~130 are the test file and a good share of the rest is comments explaining the platform differences. The behavioural change is one new module plus one method on `DesktopWindow`.
- **UI evidence**: there is no visual change to capture — no new surface, no motion, no transition. The observable difference is that a URL now navigates. If you would still like a screen recording of the navigation, I am happy to attach one.
- **Shrinking**: if this is still too much at once, I can cut it down to macOS `open-url` only and drop the Windows/Linux argv paths, or split the pure parser out as its own PR. Say which you prefer rather than closing it, and I will resubmit.

Entirely fine if the answer is "not now" — no obligation implied.

## Change Type / Scope

- [ ] Bug fix  [x] Feature  [ ] Refactor  [ ] Docs  [ ] Security  [ ] Chore
- Scope: apps/desktop (main + preload), apps/web (renderer navigation), packages/contracts (bridge type)

## Linked Issue/PR

- Closes #4996
- Related #5978 — that report is about `second-instance` dropping the URL for *SSO callbacks*. This PR adds a second `second-instance` listener for thread links and does not touch the Clerk path, so #5978 is not fixed here.
- Related #4625 — the target type is a discriminated union so a `checkout-pr` link kind can be added without re-parsing raw URLs in the renderer.

## What changed

**Before** — nothing consumes the URL. `DesktopClerk` registers the only `second-instance` listener and ignores its arguments:

```ts
yield* electronApp.on("second-instance", () => {
  void runPromise(
    Effect.gen(function* () {
      const mainWindow = yield* electronWindow.currentMainOrFirst;
      if (Option.isSome(mainWindow)) {
        yield* electronWindow.reveal(mainWindow.value);
      }
    }),
  );
});
```

There is no `open-url` listener at all, so on macOS the URL never reaches the app.

**After** — a new `DesktopDeepLinkRouter` owns link delivery, and `DesktopWindow` gains `dispatchDeepLink`:

```ts
// DesktopDeepLinkRouter.register
yield* electronApp.on<[{ preventDefault: () => void }, string]>("open-url", (event, url) => {
  event.preventDefault();
  handle(url, "open-url");
});

yield* electronApp.on<[unknown, readonly string[], string]>("second-instance", (_event, argv) => {
  const url = DesktopDeepLink.findDeepLinkInArgv(argv ?? [], schemes);
  if (url !== null) handle(url, "second-instance");
});

// Windows/Linux cold start: the launching URL is already in our own argv.
const launchUrl = DesktopDeepLink.findDeepLinkInArgv(process.argv, schemes);
if (launchUrl !== null) handle(launchUrl, "launch-argv");
```

Delivery reuses the `dispatchMenuAction` shape (existing window → otherwise `ensureMain`, wait for `did-finish-load` when the frame is still loading, then `reveal`), plus one addition that menu actions do not need:

```ts
if (Option.isNone(existingWindow) && !(yield* Ref.get(backendReadyRef))) {
  yield* Ref.set(pendingDeepLinkRef, Option.some(target));   // held, not dropped
  return;
}
```

`handleBackendReady` then replays it once via `Ref.getAndSet(pendingDeepLinkRef, Option.none())`. Without this, a link that *launches* the app would be discarded, which is half of what #4996 asks for.

Renderer navigates through the typed route rather than a raw hash:

```tsx
void navigate({
  to: "/$environmentId/$threadId",
  params: { environmentId: target.environmentId, threadId: target.threadId },
});
```

## Test plan

- [x] `DesktopDeepLink.test.ts` — 15 cases over the two pure functions.
- [x] Happy path: documented shape parses; both registered schemes accepted; scheme/host case-insensitive; segments percent-decoded.
- [x] Rejection: foreign schemes (`https://`, `t3codex://`); other hosts, explicitly including `t3code://app/...` (the renderer bundle) and `t3code://oauth/callback` — so this cannot swallow links that belong to other handlers.
- [x] Rejection: wrong segment count, blank/whitespace segments, malformed percent-encoding, non-URL input.
- [x] Rejection: segments that decode back into `/`, `?` or `#` — a crafted link must not address a different route than its two visible segments suggest.
- [x] argv extraction: finds the link among unrelated args, trims shell whitespace, takes the first of several, returns null when absent.

### Environment

- OS: macOS 26 (Apple silicon)
- Runtime: Node 24
- App: built from this branch

### Steps

1. Open any thread and note its `environmentId` / `threadId`.
2. Navigate elsewhere in the app (or quit it entirely).
3. Run `open "t3code://threads/<environmentId>/<threadId>"`.
4. Before: the window is focused at most, and the thread is not opened. After: the app reveals and navigates to that thread, whether it was already running or was launched by the link.

Three existing `DesktopWindow` stubs (`DesktopLifecycle.test.ts`, `DesktopBackendPool.test.ts`, `DesktopApplicationMenu.test.ts`) each gained one line, because they `satisfies` the service type and it now has one more method. That is the whole reason those three files appear in the diff.

Commands run in this working tree:

```
apps/desktop     tsgo --noEmit        no errors
packages/contracts tsgo --noEmit      no errors
apps/web         tsgo --noEmit        no errors
apps/desktop     vp test run          59 files, 459 tests passed
```

## Evidence

- [x] Unit tests for both pure functions
- [x] Mutation-checked: weakening `isCleanSegment` to accept any non-empty string turns exactly two cases red ("rejects segments that decode back into path separators", "rejects blank or whitespace-padded segments") and leaves the other 13 green — the guards are actually exercised, not incidentally satisfied.
- [ ] Trace/log snippets
- [ ] Screenshot/recording — no visual change; see note above
- [ ] Perf numbers — N/A

## Security Impact

- New permissions/capabilities? No.
- Secrets/tokens handling changed? No.
- New/changed network calls? No.
- Command/tool execution surface changed? No — the URL is parsed, never executed, and only two id strings are forwarded.
- Data access scope changed? No. The parser is deliberately strict: only the app's own registered scheme, only the `threads` host, exactly two segments, and no segment that decodes back into a path separator. Everything else returns `null` and falls through to existing behaviour.

## Human Verification

- **Verified**: unit tests for the parsing and argv-extraction logic; typecheck.
- **NOT verified**: I have not run a packaged build end-to-end on macOS/Windows/Linux, so the Electron event wiring itself (`open-url` firing, `second-instance` argv contents, the cold-start replay through `handleBackendReady`) is reasoned from the existing `dispatchMenuAction` path rather than observed. If you would like me to attach a recording from a local build before merging, say the word.

## Failure Recovery

- How to revert: revert this PR's single commit — no migrations, no persisted state, no schema change.
- Files to restore: `apps/desktop/src/{app/DesktopApp.ts,ipc/channels.ts,main.ts,preload.ts,window/DesktopWindow.ts}`, `apps/web/src/components/AppSidebarLayout.tsx`, `packages/contracts/src/ipc.ts`; delete `apps/desktop/src/app/DesktopDeepLink*.ts`.
- Bad symptoms to watch: a `t3code://` link that used to reach another handler (OAuth) no longer doing so — the parser rejects any host other than `threads`, and there is a test for the `oauth` and `app` hosts specifically.

## Risks and Mitigations

- **Risk**: swallowing links intended for the Clerk/OAuth flow.
  - Mitigation: this adds a listener rather than replacing one; every listener still runs. Non-`threads` hosts return `null` and are ignored, with tests covering `t3code://app/...` and `t3code://oauth/callback`.
- **Risk**: the held deep link being replayed more than once across a backend restart.
  - Mitigation: `Ref.getAndSet(..., Option.none())` takes the value, so a later `handleBackendReady` has nothing to replay.
- **Risk**: navigating to a thread the user cannot see.
  - Mitigation: navigation goes through the existing typed route, so the route's own loading/missing states (`resolveThreadRouteRenderState`) apply unchanged.

---

🤖 Authored with [Claude Code](https://claude.com/claude-code). The parsing rules and the hold-until-ready behaviour were derived by reading the existing `dispatchMenuAction` / `handleBackendReady` paths in this repo; tests were written to fail without the change.

> [!NOTE]
> ### Add `t3code://` deep link routing to open threads in `DesktopApp`
> - `DesktopDeepLinkRouter` parses `t3code://threads/{envId}/{threadId}` URLs from macOS `open-url`, Windows/Linux `second-instance`, and cold-start `process.argv`.
> - `DesktopWindow.dispatchDeepLink` holds the parsed target until the backend is ready, then sends it over `DEEP_LINK_CHANNEL` to the renderer.
> - [AppSidebarLayout.tsx](https://github.com/pingdotgg/t3code/pull/6008/files#diff-b580a936d60a70bebcccf38ac9a82ca5f4440c99445dbccb0c176ebe5dc1d301) subscribes to `window.desktopBridge.onDeepLink` and navigates to `/$environmentId/$threadId`.
> - Behavioral Change: `DesktopWindow` interface extended with `dispatchDeepLink`; test mock layers updated to include this method.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 699f356.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Touches Electron protocol events (`open-url`, `second-instance`, argv) and IPC into the renderer. Parsing is strict and OAuth/app hosts are rejected, but a mistake could steal other scheme handlers or navigate to the wrong thread.
> 
> **Overview**
> Desktop now opens `t3code://threads/<environmentId>/<threadId>` links (the same contract mobile already uses) instead of only focusing the window.
> 
> A new router listens on macOS `open-url` (registered before `ready`) and on Windows/Linux via launch argv plus `second-instance`. URLs are parsed strictly (registered scheme, `threads` host, exactly two clean segments) so OAuth and renderer-bundle URLs are ignored. `DesktopWindow` holds a pending target until the backend is ready, then sends it over `desktop:deep-link`.
> 
> The preload buffers the typed target until the renderer subscribes; `AppSidebarLayout` navigates to `/$environmentId/$threadId`. Clerk/OAuth handling is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 699f356f89ff11d03cc866fe94e5b3f5bb4782bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/desktop/src/app/DesktopApp.ts
+++ b/apps/desktop/src/app/DesktopApp.ts
@@ -18,6 +18,7 @@ import * as DesktopWindow from "../window/DesktopWindow.ts";
 import * as DesktopBackendPool from "../backend/DesktopBackendPool.ts";
 import * as DesktopEnvironment from "./DesktopEnvironment.ts";
 import * as DesktopLifecycle from "./DesktopLifecycle.ts";
+import * as DesktopDeepLinkRouter from "./DesktopDeepLinkRouter.ts";
 import * as DesktopLinuxUrlHandler from "./DesktopLinuxUrlHandler.ts";
 import * as DesktopObservability from "./DesktopObservability.ts";
 import * as DesktopPreReadyPlatform from "./DesktopPreReadyPlatform.ts";
@@ -224,6 +225,7 @@ const startup = Effect.gen(function* () {
   const electronApp = yield* ElectronApp.ElectronApp;
   const lifecycle = yield* DesktopLifecycle.DesktopLifecycle;
   const linuxUrlHandler = yield* DesktopLinuxUrlHandler.DesktopLinuxUrlHandler;
+  const deepLinkRouter = yield* DesktopDeepLinkRouter.DesktopDeepLinkRouter;
   const clerk = yield* DesktopClerk.DesktopClerk;
   const shellEnvironment = yield* DesktopShellEnvironment.DesktopShellEnvironment;
   const desktopSettings = yield* DesktopAppSettings.DesktopAppSettings;
@@ -287,6 +289,9 @@ const startup = Effect.gen(function* () {
   yield* applicationMenu.configure;
   yield* updates.configure;
   yield* linuxUrlHandler.register;
+  // Registered before bootstrap so a link that launched the app is picked up
+  // and held, rather than being missed while the backend starts.
+  yield* deepLinkRouter.register;
   yield* bootstrap.pipe(Effect.catchCause((cause) => fatalStartupCause("bootstrap", cause)));
 }).pipe(Effect.withSpan("desktop.startup"));
 
--- a/apps/desktop/src/app/DesktopApp.ts
+++ b/apps/desktop/src/app/DesktopApp.ts
@@ -273,6 +273,9 @@ const startup = Effect.gen(function* () {
   yield* appIdentity.configure;
   yield* lifecycle.register;
   yield* clerk.configure;
+  // macOS can emit open-url before ready, and does not preserve that URL in
+  // argv. Register now; DesktopWindow holds it until the backend can open.
+  yield* deepLinkRouter.register;
 
   yield* electronApp.whenReady.pipe(
     Effect.withSpan("desktop.electron.whenReady"),
@@ -289,9 +292,6 @@ const startup = Effect.gen(function* () {
   yield* applicationMenu.configure;
   yield* updates.configure;
   yield* linuxUrlHandler.register;
-  // Registered before bootstrap so a link that launched the app is picked up
-  // and held, rather than being missed while the backend starts.
-  yield* deepLinkRouter.register;
   yield* bootstrap.pipe(Effect.catchCause((cause) => fatalStartupCause("bootstrap", cause)));
 }).pipe(Effect.withSpan("desktop.startup"));
 
--- a/apps/desktop/src/app/DesktopDeepLink.test.ts
+++ b/apps/desktop/src/app/DesktopDeepLink.test.ts
@@ -1,0 +1,110 @@
+import { describe, expect, it } from "vite-plus/test";
+
+import { findDeepLinkInArgv, parseDeepLink } from "./DesktopDeepLink.ts";
+
+const SCHEMES = ["t3code", "t3code-dev"] as const;
+
+describe("parseDeepLink", () => {
+  it("parses the documented thread link shape", () => {
+    expect(parseDeepLink("t3code://threads/env-1/thread-42", SCHEMES)).toEqual({
+      kind: "thread",
+      environmentId: "env-1",
+      threadId: "thread-42",
+    });
+  });
+
+  it("accepts every registered scheme, including the development one", () => {
+    for (const scheme of SCHEMES) {
+      expect(parseDeepLink(`${scheme}://threads/env/thread`, SCHEMES)).toEqual({
+        kind: "thread",
+        environmentId: "env",
+        threadId: "thread",
+      });
+    }
+  });
+
+  it("is case insensitive for the scheme and host", () => {
+    expect(parseDeepLink("T3CODE://THREADS/env/thread", SCHEMES)).toEqual({
+      kind: "thread",
+      environmentId: "env",
+      threadId: "thread",
+    });
+  });
+
+  it("percent-decodes segments", () => {
+    expect(parseDeepLink("t3code://threads/env%20one/thread%2Btwo", SCHEMES)).toEqual({
+      kind: "thread",
+      environmentId: "env one",
+      threadId: "thread+two",
+    });
+  });
+
+  it("rejects foreign schemes so other protocol handlers are never hijacked", () => {
+    expect(parseDeepLink("https://threads/env/thread", SCHEMES)).toBeNull();
+    expect(parseDeepLink("t3codex://threads/env/thread", SCHEMES)).toBeNull();
+  });
+
+  it("rejects other hosts, including the renderer bundle URL", () => {
+    expect(parseDeepLink("t3code://app/#/env/thread", SCHEMES)).toBeNull();
+    expect(parseDeepLink("t3code://oauth/callback", SCHEMES)).toBeNull();
+  });
+
+  it("requires exactly two path segments", () => {
+    expect(parseDeepLink("t3code://threads/env", SCHEMES)).toBeNull();
+    expect(parseDeepLink("t3code://threads/env/thread/extra", SCHEMES)).toBeNull();
+    expect(parseDeepLink("t3code://threads", SCHEMES)).toBeNull();
+  });
+
+  // A decoded separator would let a crafted link address a different route than
+  // the one the two segments appear to describe.
+  it("rejects segments that decode back into path separators", () => {
+    expect(parseDeepLink("t3code://threads/env%2F..%2Fother/thread", SCHEMES)).toBeNull();
+    expect(parseDeepLink("t3code://threads/env/thread%3Fx%3D1", SCHEMES)).toBeNull();
+    expect(parseDeepLink("t3code://threads/env/thread%23frag", SCHEMES)).toBeNull();
+  });
+
+  it("rejects blank or whitespace-padded segments", () => {
+    expect(parseDeepLink("t3code://threads/%20/thread", SCHEMES)).toBeNull();
+    expect(parseDeepLink("t3code://threads/env/%20thread", SCHEMES)).toBeNull();
+  });
+
+  it("rejects malformed percent-encoding instead of throwing", () => {
+    expect(parseDeepLink("t3code://threads/%E0%A4%A/thread", SCHEMES)).toBeNull();
+  });
+
+  it("rejects input that is not a URL at all", () => {
+    expect(parseDeepLink("", SCHEMES)).toBeNull();
+    expect(parseDeepLink("not a url", SCHEMES)).toBeNull();
+  });
+});
+
+describe("findDeepLinkInArgv", () => {
+  // Electron passes the full argv of the second launch, so the URL is never the
+  // first element in practice.
+  it("finds the link among unrelated arguments", () => {
+    expect(
+      findDeepLinkInArgv(
+        ["/path/to/T3 Code", "--allow-file-access", "t3code://threads/env/thread"],
+        SCHEMES,
+      ),
+    ).toBe("t3code://threads/env/thread");
+  });
+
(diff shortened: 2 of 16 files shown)


Moinax:
Ran this branch from a packaged Linux build (CachyOS, AppImage) — the wiring holds end to end, cold-start argv path included. One number worth having before it lands, because it bounds what can be built on top.

On Linux a caller's only entry point is `xdg-open`, which launches **a second full copy of the app** whose entire job is to take the single-instance lock, emit `second-instance` and exit. Measured here: **~1.4s** from `xdg-open` to the thread being on screen, roughly half of it the AppImage's squashfs mount. Fine for a link clicked in a browser — but #4996 asks for notification actions and status-bar integrations, and a menu-bar item that takes 1.4s to switch threads reads as broken.

A local listener in the main process removes that second process entirely. I added one on top of this branch:

```ts
// in DesktopDeepLinkRouter, alongside open-url / second-instance
const server = NodeNet.createServer((connection) => {
  /* accumulate, first line wins, capped */ handle(url, "socket");
});
server.listen(socketPath, () => NodeFS.chmodSync(socketPath, 0o600));
```

`$XDG_RUNTIME_DIR/t3code-deeplink.sock`, `/tmp/<uid>-…` when that is unset, distinct name in dev. Same URL string, same `handle()`, same parser — one more source on the funnel you already built, not a second contract.

**~1.4s → ~10ms**, same machine, same link. In practice the tab switch is indistinguishable from a click inside the app.

Shape, briefly:

- The 0600 mode in a user-owned directory *is* the authentication — the same trust boundary `xdg-open` already assumes. Nothing is read back, and the only reachable action is the navigation `handle` performs for links from anywhere else.
- Non-fatal by construction: `acquireRelease` unlinks a stale socket before `listen` and closes + unlinks on release; the whole registration is wrapped so any failure logs a warning and leaves `xdg-open` as the path.
- Linux/macOS in the form I wrote it. Windows would want a named pipe, or simply not have it — `open-url`/argv still work.

Happy to open it as a follow-up once this merges, or to hand you the diff to fold in here — whichever you prefer. Either way the measurement is the part worth keeping: without something like it, every Linux deep link pays 1.4s.

MaanavD:
@Moinax @edwluo anything you need help w/ to get this merged? Would love to get it in for a stream deck integration I've been making!

t3dotgg:
> [!NOTE]

This adds `t3code://` parsing, buffering, and routing, but its URL does not carry both environment and thread identity. [#8246](https://github.com/pingdotgg/t3code/pull/8246) defines that full contract and the renderer readiness flow, so we are closing this older parser.

PR #6041, opened 2026-08-10
fix(server): let a draft retry after its worktree fails to create

sonyabytes:
## Problem

When a draft's first send is in **New worktree** mode and `git worktree add` fails, every retry from that draft fails with `Thread 'X' already exists and cannot be created twice.` — including after switching the draft to **Current checkout** — until the user abandons the draft and starts a new one. The failed attempt also leaves the surviving draft hidden from the sidebar.

## Fix

The bootstrap flow dispatches `thread.create` before preparing the worktree and rolls back with `thread.delete` (a soft delete) when preparation fails. Two small changes:

- **server**: `requireThreadAbsent` now treats soft-deleted threads as absent, mirroring the existing project invariant, so the draft's pre-allocated thread id can be re-created on retry.
- **web**: the draft route only marks a draft as promoted once its thread has actually started, so a rolled-back bootstrap no longer sticks `promotedTo` on the draft and hides it from the sidebar.

## Verification

- New unit test: re-creating a soft-deleted thread id passes `requireThreadAbsent` (`commandInvariants.test.ts`).
- Reproduced live with a forced `git worktree add` failure: fail → retry in worktree mode → clean rollback again; switch to Current checkout → send succeeds on the same thread id (event log: `created → deleted → created → deleted → created`), and the normal worktree flow still creates worktrees when nothing is blocking it.

Fixed by Claude Fable 5 running in Claude Code.

> [!NOTE]
> ### Fix draft retry after worktree creation failure by treating soft-deleted threads as absent
> - Updates `requireThreadAbsent` in [commandInvariants.ts](https://github.com/pingdotgg/t3code/pull/6041/files#diff-426dec65d7876b98b993600840e4b58dcca662d84f3a23e884ea72cc2783e6a0) to treat threads with a non-null `deletedAt` as absent, allowing `thread.create` commands to succeed when a prior thread was soft-deleted.
> - Fixes premature draft promotion in [_chat.draft.$draftId.tsx](https://github.com/pingdotgg/t3code/pull/6041/files#diff-81aad35683c56f7b68b13e851e82c9b2677cce9d19548633e91b9b93798e030a) by gating `markPromotedDraftThreadByRef` on `serverThreadStarted`, preventing promotion when the server thread exists but hasn't started yet.
> - Together these allow a draft to retry successfully after its worktree fails to create, leaving a soft-deleted thread behind.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 9402a15. 2 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/apps/server/src/orchestration/commandInvariants.test.ts
+++ b/apps/server/src/orchestration/commandInvariants.test.ts
@@ -201,6 +201,37 @@ describe("commandInvariants", () => {
     ).rejects.toThrow("already exists");
   });
 
+  it("treats soft-deleted threads as absent for create flows", async () => {
+    const deletedThread = readModel.threads[0]!;
+    const readModelWithDeletedThread: OrchestrationReadModel = {
+      ...readModel,
+      threads: [{ ...deletedThread, deletedAt: now }, ...readModel.threads.slice(1)],
+    };
+
+    await Effect.runPromise(
+      requireThreadAbsent({
+        readModel: readModelWithDeletedThread,
+        command: {
+          type: "thread.create",
+          commandId: CommandId.make("cmd-4"),
+          threadId: deletedThread.id,
+          projectId: ProjectId.make("project-a"),
+          title: "recreated",
+          modelSelection: {
+            instanceId: ProviderInstanceId.make("codex"),
+            model: "gpt-5-codex",
+          },
+          interactionMode: DEFAULT_PROVIDER_INTERACTION_MODE,
+          runtimeMode: "full-access",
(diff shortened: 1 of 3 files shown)


sonyabytes:
checked that other similar prs will be fixing it in favor of the orchestrator v2 (4647). closing

Issue #4820, opened 2026-07-29
[Feature]: Mark WSL environments in their label, since WSL inherits the Windows host name


inayayousfi:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I am describing a concrete problem or use case, not just a vague idea.

### Area

apps/server

### Problem or use case

A WSL environment and the Windows machine hosting it show up under the same name, so the connection list has two entries I cannot tell apart.

This is because WSL2 defaults its Linux hostname to the Windows computer name. On my machine:

```
$ cat /etc/hostname
internet-gyal-terminal

$ cmd.exe /c "echo %COMPUTERNAME%"
INTERNET-GYAL-T          # same name, truncated to the 15-char NetBIOS limit
```

The label resolver (`apps/server/src/environment/ServerEnvironmentLabel.ts`) walks a fallback chain, and on a stock WSL distro every friendly step misses:

1. `/etc/machine-info` → `PRETTY_HOSTNAME`. Not present on a default install.
2. `hostnamectl --pretty`. Returns empty even with `systemd=true` in `/etc/wsl.conf`.
3. `HostProcessHostname`, i.e. `NodeOS.hostname()` (`packages/shared/src/hostProcess.ts`). This is what actually wins, and it is the mirrored Windows name.

That label goes straight onto the `ExecutionEnvironmentDescriptor` (`apps/server/src/environment/ServerEnvironment.ts:127-134`) and is what every client renders, plus what the relay stores as `environmentLabel` (`infra/relay/src/environments/EnvironmentLinks.ts`).

Two collisions follow from this, not one:

- WSL vs its own Windows host.
- Two WSL distros on the same host, which also both inherit the same Windows name.

I could not find a rename affordance in Connections settings, so there is no in-app way out. The only fixes are outside the app: change the Linux hostname, or write `PRETTY_HOSTNAME` into `/etc/machine-info`.

### Proposed solution

When the server detects it is running under WSL **and** the label came from the plain hostname fallback, disambiguate the label with the WSL distro:

```
internet-gyal-terminal (WSL: archlinux)
```

The distro name is what makes this work for the multi-distro case too, not just for WSL vs Windows.

[...]

PR #6053, opened 2026-08-10
Fix for issue 4820

mohamedmastouri-hue:
Fixes issue #4820

---

> [!NOTE]
> **Low Risk**
> Display-only labeling change with file read and env lookup; no auth, security, or data-path impact.
> 
> **Overview**
> **Server environment labels** now distinguish Windows Subsystem for Linux when friendly host names are unavailable and the resolver falls back to the hostname.
> 
> On Linux, a new **`isWslEnvironment`** check reads `/proc/sys/kernel/osrelease` for `microsoft` or `wsl`. If WSL is detected, the label becomes `hostname (WSL: <distro>)` when **`WSL_DISTRO_NAME`** is set and non-empty, or `hostname (WSL)` otherwise. **`HostProcessEnvironment`** is wired in for distro name lookup; tests stub **`HostProcessEnvironment`** via an extended **`withHostPlatform`** helper and cover both named and generic WSL cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84346fe3a1074089ab9398e047887367eebfe2a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Append WSL distro info to server environment label when running under WSL
> - Adds `isWslEnvironment` in [`ServerEnvironmentLabel.ts`](https://github.com/pingdotgg/t3code/pull/6053/files#diff-b1aacbc192ffaacdfdd1ba5d7cbd3815df9976714c71601e826526144c769e59) that detects WSL by checking `/proc/sys/kernel/osrelease` for 'microsoft' or 'wsl' (case-insensitive); file errors are logged at debug level and treated as non-WSL.
> - Updates `resolveServerEnvironmentLabel` to append `(WSL: <distro>)` when `WSL_DISTRO_NAME` is set, or `(WSL)` when not, for Linux hosts running under WSL.
> - Adds tests covering both cases using a mocked `WslLayer` and `HostProcessEnvironment`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 84346fe.</sup>
> 
>

### DIFF
--- a/apps/server/src/environment/ServerEnvironmentLabel.test.ts
+++ b/apps/server/src/environment/ServerEnvironmentLabel.test.ts
@@ -286,6 +286,19 @@ describe("resolveServerEnvironmentLabel", () => {
   );
   it.effect("appends WSL distro name when WSL_DISTRO_NAME is present", () =>
     Effect.gen(function* () {
+      runMock.mockReturnValueOnce(
+        Effect.succeed({
+          stdout: "",
+          stderr: "",
+          code: ChildProcessSpawner.ExitCode(1),
+          timedOut: false,
+          stdoutTruncated: false,
+          stderrTruncated: false,
+          stdoutInvalidUtf8: false,
+          stderrInvalidUtf8: false,
+        }),
+      );
+
       const WslLayer = Layer.merge(
         ProcessRunnerTest,
         FileSystem.layerNoop({
@@ -311,6 +324,19 @@ describe("resolveServerEnvironmentLabel", () => {
 
   it.effect("appends generic WSL label when WSL_DISTRO_NAME is absent", () =>
     Effect.gen(function* () {
+      runMock.mockReturnValueOnce(
+        Effect.succeed({
+          stdout: "",
+          stderr: "",
+          code: ChildProcessSpawner.ExitCode(1),
+          timedOut: false,
+          stdoutTruncated: false,
+          stderrTruncated: false,
+          stdoutInvalidUtf8: false,
+          stderrInvalidUtf8: false,
+        }),
+      );
+
       const WslLayer = Layer.merge(
         ProcessRunnerTest,
         FileSystem.layerNoop({
--- a/apps/server/src/environment/ServerEnvironmentLabel.test.ts
+++ b/apps/server/src/environment/ServerEnvironmentLabel.test.ts
@@ -7,7 +7,7 @@ import * as PlatformError from "effect/PlatformError";
 import * as References from "effect/References";
 import * as Schema from "effect/Schema";
 import * as ChildProcessSpawner from "effect/unstable/process/ChildProcessSpawner";
-import { HostProcessHostname, HostProcessPlatform } from "@t3tools/shared/hostProcess";
+import { HostProcessEnvironment, HostProcessHostname, HostProcessPlatform } from "@t3tools/shared/hostProcess";
 import { vi } from "vite-plus/test";
 
 import * as ProcessRunner from "../processRunner.ts";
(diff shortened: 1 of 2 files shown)


CDVolvik:
Verified on Linux/WSL against this PR's head (tested 2026-08-11):

`apps/server` · `src/environment/ServerEnvironmentLabel.test.ts` → **10/10 pass**.

I hit the same bug first-hand on a stock Ubuntu WSL2 install before this PR existed: `/etc/hostname` and `os.hostname()` both return the Windows computer name (`GalacticCow` / `GALACTICCOW`), `/etc/machine-info` is absent, and `hostnamectl --pretty` is empty, so the label chain lands on the mirrored Windows name. Two distros on one host are also indistinguishable without the distro name in the label.

A few notes from reading the approach:

1. **Detecting WSL via `/proc/sys/kernel/osrelease`** is stronger than env-only when `WSL_DISTRO_NAME` is unset, and the generic `hostname (WSL)` fallback is the right shape for that case.
2. **Suffix only on the plain-hostname path** (after friendly/pretty resolution fails) is the important constraint — a deliberate `PRETTY_HOSTNAME` should stay untouched. Your tests cover the hostname-fallback cases; a pin that "pretty name does not get a WSL suffix" would make that invariant visible to the next editor.
3. **Reuse of `HostProcessEnvironment`** matches how `externalLauncher` already reads `WSL_DISTRO_NAME`. Worth keeping the test helper defaulting env to empty so Linux CI and WSL hosts cannot diverge on accidental real `process.env`.

I am not filing a competing PR for #4820.

CDVolvik (commented):
Verified on Linux (WSL2 Ubuntu, Node 24).

Branch as submitted:

```
✓ src/environment/ServerEnvironmentLabel.test.ts (10 tests)
  Tests  10 passed (10)
```

Swapping only `ServerEnvironmentLabel.ts` for the current `main` copy, keeping this PR's tests:

```
× appends WSL distro name when WSL_DISTRO_NAME is present
  AssertionError: expected 'WINDOWS-HOST' to be 'WINDOWS-HOST (WSL: Ubuntu-22.04)'
× appends generic WSL label when WSL_DISTRO_NAME is absent
  AssertionError: expected 'WINDOWS-HOST' to be 'WINDOWS-HOST (WSL)'

  Tests  2 failed | 8 passed (10)
```

Both new tests fail without the source change and pass with it, and the other eight are untouched, so they pin the added behaviour specifically.

Worth adding that the premise holds on a real setup, which is the part unit tests cannot show. On a Windows host running a WSL2 distro, `hostname` inside the distro returns the Windows machine name verbatim, and `WSL_DISTRO_NAME` is populated. So before this change the two environments are genuinely indistinguishable by label, and `WSL_DISTRO_NAME` is the right discriminator to reach for. That matches what #4820 describes.

One thing to confirm: `WSL_DISTRO_NAME` is set by the WSL init path, so it is present in an interactive shell and in anything launched from one, but it is not guaranteed for a process started through other entry points. The generic `(WSL)` fallback covers that, which is the right call, though it does mean the distro name is best-effort rather than always present.

inayayousfi:
Written by inayayousfi, typed by gpt-5.6-sol running in OpenCode.
Every call here is inayayousfi's, and no agent acted on its own.

PR #7267 is a related broader naming implementation that is now ready for review. It lets users assign a server-owned name from web, desktop, or mobile, including for remote environments, then propagates that name to connected clients and T3 Connect. It handles the WSL collision once the user chooses distinct names, and it also covers hostname collisions outside WSL.

This PR still provides something #7267 does not: automatic WSL detection and a zero-configuration distro suffix. I hope this focused fix can land alongside the manual rename support.

PR #6064, opened 2026-08-10
fix(web): file tree no longer blanks on whitespace-padded duplicate paths

### LINKED ISSUE #5501: [Bug]: Files panel throws "Duplicate path" and renders nothing when a workspace has two filenames differing only by trailing whitespace
Image

## Before submitting

## Area

`apps/server`

## Steps to reproduce

A workspace that contains two distinct files whose names differ only by trailing whitespace kills
the entire Files panel. Both names are legal on macOS and Linux, and they show up in real repos
after a bad `touch "$(pbpaste)"` or a copy/paste into a shell.

```bash
python3 - <<'EOF'
import os, subprocess
R = "/tmp/dup-path-repro"
os.makedirs(os.path.join(R, "pkg"))
subprocess.run(["git", "init", "-q"], cwd=R, check=True)
open(os.path.join(R, "pkg", "notes.txt"), "w").write("normal file\n")
# Same name plus a trailing newline: a second, real file. Note this cannot be
# created with `touch $'pkg/notes.txt\n'` — the shell strips trailing newlines.
open(os.path.join(R, "pkg", "notes.txt\n"), "w").write("evil twin\n")
EOF
```

1. Open `/tmp/dup-path-repro` in T3 Code.
2. Open the Files panel.

The workspace only needs **one** such file anywhere in the tree for the whole panel to break.

## Expected behavior

The file tree renders. Either both files appear as separate rows, or the whitespace-suffixed one is
dropped — but the panel stays usable.

## Actual behavior

The Files panel renders nothing and throws:

```
Error: Duplicate path: "pkg/notes.txt"
    at yn.appendPreparedPath (t3code://app/assets/FilePreviewPanel-<hash>.js:1:15031)
    at yn.appendPreparedPaths (t3code://app/assets/FilePreviewPanel-<hash>.js:1:10832)
    at Fa.resetPaths (t3code://app/assets/FilePreviewPanel-<hash>.js:1671:17085)
    at Ml.resetPaths (t3code://app/assets/FilePreviewPanel-<hash>.js:1671:97106)
```

There is no recovery from inside the app — refresh, reopen, and restart all hit the same throw. The
only fix is deleting the offending file from a terminal.

## Root cause

The server dedupes entries **before** the normalization that creates the collision.

1. `apps/server/src/workspace/WorkspaceSearchIndex.ts:281` — `withDirectoryAncestors` dedupes with
   `new Map(entries.map((entry) => [entry.path, entry]))`. The index reports both files, and
   `"pkg/notes.txt" !== "pkg/notes.txt\n"`, so **both survive**.
2. `apps/server/src/workspace/WorkspaceSearchIndex.ts:435` — `list()` sorts with `localeCompare`,
   which treats `\n` as ignorable. The pair compares equal and lands **adjacent**.
3. `packages/contracts/src/project.ts:28` — `ProjectEntry.path` is `TrimmedNonEmptyString`, and
   `packages/contracts/src/baseSchemas.ts:6` runs `value.trim()` on both decode and encode. The
   trailing newline is stripped here, so the client receives two **byte-identical, adjacent** paths.
4. `apps/web/src/components/files/FileBrowserPanel.tsx:118` maps them into `treePaths` and `:262`
   calls `model.resetPaths(treePaths)`. `@pierre/trees`' path-store builder throws on its
   adjacent-duplicate guard (`dist/path-store/src/builder.js:383`).

Confirmed against the library in isolation with exactly the paths the client receives:

```js
const paths = ["pkg/", "pkg/notes.txt", "pkg/notes.txt"];
builder.appendPreparedPaths(paths.map(parseInputPath));
// THREW: Duplicate path: "pkg/notes.txt"
```

## Failing test

Drop this in `apps/server/src/workspace/` and run
`bun run test src/workspace/WorkspaceSearchIndex.duplicatePathRepro.test.ts` from `apps/server`:

```ts
import { expect, it } from "@effect/vitest";
import { ProjectEntry } from "@t3tools/contracts";
import * as Effect from "effect/Effect";
import * as Schema from "effect/Schema";
import { execFileSync } from "node:child_process";
import { mkdtempSync, mkdirSync, writeFileSync } from "node:fs";
import { tmpdir } from "node:os";
import { join } from "node:path";

import * as WorkspaceSearchIndex from "./WorkspaceSearchIndex.ts";

/**
 * A workspace can legitimately contain two distinct files whose names differ
 * only by surrounding whitespace ("notes.txt" and "notes.txt\n"). The index
 * dedupes on the raw name, but ProjectEntry.path is a TrimmedNonEmptyString,
 * so both entries collapse to the same string once they cross the wire — and
 * the file tree throws `Duplicate path: "..."` on the adjacent pair.
 */
function makeWorkspace(): string {
  const root = mkdtempSync(join(tmpdir(), "t3-dup-path-"));
  execFileSync("git", ["init", "-q"], { cwd: root });
  mkdirSync(join(root, "pkg"));
  writeFileSync(join(root, "pkg", "notes.txt"), "");
  writeFileSync(join(root, "pkg", "notes.txt\n"), "");
  return root;
}

it.effect("does not emit two entries that collapse to the same trimmed path", () =>
  Effect.scoped(
    Effect.gen(function* () {
      const root = makeWorkspace();
      const index = yield* WorkspaceSearchIndex.make(root);
      const { entries } = yield* index.list();

      const rawPaths = entries.map((entry) => entry.path).filter((p) => p.includes("notes.txt"));
      // Both files are indexed as distinct strings, so the server-side dedupe
      // (a Map keyed on the raw path) keeps both.
      expect(rawPaths).toHaveLength(2);

      // ProjectEntry trims on encode/decode, which is what the client receives.
      const encode = Schema.encodeUnknownSync(Schema.Array(ProjectEntry));
      const wirePaths = encode(entries).map((entry) => entry.path);
      const duplicates = wirePaths.filter((path, i) => wirePaths.indexOf(path) !== i);

      expect(duplicates).toEqual([]);
    }),
  ),
);
```

Current output:

```
AssertionError: expected [ 'pkg/notes.txt' ] to deeply equal []
```

## Suggested fix

Normalize in `toProjectEntry` (`WorkspaceSearchIndex.ts:138`) so `withDirectoryAncestors` dedupes on
the same string the client will see. That is the actual bug and it is a one-line change.

Optionally also drop adjacent duplicates when building `treePaths`
(`FileBrowserPanel.tsx:118`), so a single odd filename can never take down the whole panel again.

## Impact

Major degradation or frequent failure — the Files panel is completely unusable for the affected
workspace until the file is deleted outside the app.

## Version or commit

`main @ a483337a0`

## Environment

macOS 15 (Darwin 25.5.0), Apple Silicon, desktop app.

## Workaround

Delete the whitespace-suffixed file:

```bash
python3 -c "import os; os.remove('pkg/notes.txt\n')"
```

To find offenders in a workspace:

```bash
python3 - <<'EOF'
import os
for root, dirs, files in os.walk('.'):
    dirs[:] = [d for d in dirs if d not in ('.git', 'node_modules', '.venv', '__pycache__')]
    for n in files + dirs:
        if n != n.strip():
            print(repr(os.path.join(root, n)))
EOF
```

gertzgal:
Fixes #5501

## What Changed

`WorkspaceSearchIndex` no longer emits entries the wire contract cannot represent. `ProjectEntry.path` is a `TrimmedNonEmptyString` (trimmed on encode), so two on-disk paths like `pkg/notes.txt` and `pkg/notes.txt\n` collapse to the same wire path; the client's file tree throws `Duplicate path` on the pair and the whole window dies to an error boundary.

- **Server** (`apps/server/src/workspace/WorkspaceSearchIndex.ts`): a path is wire-representable iff every segment survives trimming. Checking per segment (not just the whole string) keeps every ancestor prefix of an emitted entry emittable too, so subtrees under an unrepresentable directory are dropped consistently rather than emitted as orphans of a directory the server refuses to send.
- **Client** (`apps/web/src/components/files/fileBrowserTreePaths.ts`): the entry→tree-path mapping is extracted into a small pure helper that also de-dupes defensively, so one bad entry from an older/other server can never take the whole Files panel down again.

Both layers are covered by tests (`WorkspaceSearchIndex.wirePaths.test.ts` against the real file walker in a temp dir, plus unit tests for `buildTreePaths`).

## Why

A single whitespace-padded filename anywhere in a workspace — trivially created by e.g. `touch "$(pbpaste)"` — currently kills the entire app with `Duplicate path: "pkg/notes.txt"`, with no recovery from inside the app (refresh/reopen/restart all re-throw; you have to delete the file from a terminal). Repro script and full root-cause chain are in #5501.

Dropping unrepresentable paths server-side is the right layer: such paths can't be addressed by any path-based request after the schema trims them, so emitting them can only ever produce collisions. The client-side de-dupe is defense in depth for a component whose failure mode is "the panel throws and renders nothing."

## UI Changes

Both screenshots are against the issue's repro workspace (`/tmp/dup-path-repro` containing `pkg/notes.txt` and `pkg/notes.txt\n`), on a clean dev profile.

**Before** — opening the Files panel takes down the whole window:

before-duplicate-path-crash

**After** — the panel renders; the whitespace-suffixed twin is dropped:

after-files-panel-renders

## Checklist

- [ ] I included a video for animation/interaction changes (N/A — no motion changes)

> [!NOTE]
> ### Fix file tree blanking on whitespace-padded duplicate paths in the file browser
> - Adds `buildTreePaths` in [fileBrowserTreePaths.ts](https://github.com/pingdotgg/t3code/pull/6064/files#diff-04f73a2750a43d870b2dd7eeb17be002531ec9c350a11e6123dcf04fc7cbf24e) to convert `ProjectEntry` items to tree paths, appending trailing slashes for directories and deduplicating via a `Set` to prevent the tree model from crashing on duplicate paths.
> - Adds `isWireRepresentablePath` and `toEntryPath` helpers in [WorkspaceSearchIndex.ts](https://github.com/pingdotgg/t3code/pull/6064/files#diff-c11793234fa6d81f052147cd8ae0b62e4e60366837e340f42279225a76704093) to filter out any file, directory, or project entries whose path segments contain leading or trailing whitespace before emitting them over the wire.
> - `FileBrowserPanel` now uses `buildTreePaths` instead of the local `treePath` helper.
> - Behavioral Change: paths with whitespace-padded segments are now silently dropped from the file tree rather than passed through, which may hide files with unusual names.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized da79dab. 3 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/apps/server/src/workspace/WorkspaceSearchIndex.ts
+++ b/apps/server/src/workspace/WorkspaceSearchIndex.ts
@@ -137,26 +137,39 @@ function parentPathOf(input: string): string | undefined {
   return separatorIndex === -1 ? undefined : input.slice(0, separatorIndex);
 }
 
-function toProjectEntry(item: MixedItem): ProjectEntry | null {
-  const normalizedPath = trimDirectorySeparator(toPosixPath(item.item.relativePath));
-  if (!normalizedPath) {
+// ProjectEntry.path is trimmed by the wire schema, so a path with leading or
+// trailing whitespace would collapse onto its trimmed twin after encoding
+// (crashing the client's file tree on the duplicate) and cannot be addressed
+// by any path-based request. A path is wire-representable iff every segment
+// survives trimming — checking per segment (not just the whole string) keeps
+// every ancestor prefix of a kept path representable too, so files inside an
+// unrepresentable directory are dropped along with it rather than emitted as
+// orphans of a directory the server refuses to send.
+function isWireRepresentablePath(path: string): boolean {
+  return path.split("/").every((segment) => segment === segment.trim());
+}
+
+function toEntryPath(relativePath: string): string | null {
+  const normalizedPath = trimDirectorySeparator(toPosixPath(relativePath));
+  if (!normalizedPath || !isWireRepresentablePath(normalizedPath)) {
     return null;
   }
+  return normalizedPath;
+}
 
-  return {
-    path: normalizedPath,
-    kind: item.type,
-  };
+function toProjectEntry(item: MixedItem): ProjectEntry | null {
+  const path = toEntryPath(item.item.relativePath);
+  return path ? { path, kind: item.type } : null;
 }
 
 function toFileEntry(item: FileItem): ProjectEntry | null {
-  const normalizedPath = trimDirectorySeparator(toPosixPath(item.relativePath));
-  return normalizedPath ? { path: normalizedPath, kind: "file" } : null;
+  const path = toEntryPath(item.relativePath);
+  return path ? { path, kind: "file" } : null;
 }
 
 function toDirectoryEntry(item: DirItem): ProjectEntry | null {
-  const normalizedPath = trimDirectorySeparator(toPosixPath(item.relativePath));
-  return normalizedPath ? { path: normalizedPath, kind: "directory" } : null;
+  const path = toEntryPath(item.relativePath);
+  return path ? { path, kind: "directory" } : null;
 }
 
 function mapFileSearchResult(
--- a/apps/server/src/workspace/WorkspaceSearchIndex.wirePaths.test.ts
+++ b/apps/server/src/workspace/WorkspaceSearchIndex.wirePaths.test.ts
@@ -1,0 +1,68 @@
+import * as NodeServices from "@effect/platform-node/NodeServices";
+import { ProjectEntry } from "@t3tools/contracts";
+import { expect, it } from "@effect/vitest";
+import * as Effect from "effect/Effect";
+import * as FileSystem from "effect/FileSystem";
+import * as Path from "effect/Path";
+import * as Schema from "effect/Schema";
+
+import * as WorkspaceSearchIndex from "./WorkspaceSearchIndex.ts";
+
+/**
+ * ProjectEntry.path is a TrimmedNonEmptyString, so a path segment with leading
+ * or trailing whitespace ("notes.txt\n") collapses onto its trimmed twin once
+ * it crosses the wire. The client's file tree throws `Duplicate path: "..."`
+ * on the resulting adjacent pair and renders nothing. The index must therefore
+ * only emit entries whose full ancestor chain the wire contract can represent.
+ */
+const listEntriesOf = Effect.fn(function* (files: ReadonlyArray<string>) {
+  const fileSystem = yield* FileSystem.FileSystem;
+  const path = yield* Path.Path;
+  const root = yield* fileSystem.makeTempDirectoryScoped({ prefix: "t3-dup-path-" });
+  for (const relativePath of files) {
+    const absolutePath = path.join(root, relativePath);
+    yield* fileSystem.makeDirectory(path.dirname(absolutePath), { recursive: true });
+    yield* fileSystem.writeFileString(absolutePath, "");
+  }
+  const index = yield* WorkspaceSearchIndex.make(root);
+  return yield* index.list();
+});
+
+function encodedPaths(entries: ReadonlyArray<ProjectEntry>): ReadonlyArray<string> {
+  const encode = Schema.encodeUnknownSync(Schema.Array(ProjectEntry));
+  return encode(entries).map((entry) => entry.path);
+}
+
+it.effect("drops a file whose name collapses onto a sibling's after trimming", () =>
+  Effect.gen(function* () {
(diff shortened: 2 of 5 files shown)


gertzgal:
will get back to it.

PR #6080, opened 2026-08-11
feat(composer): show provider account usage in the composer

gfsaaser24:
> **Up front: this is a very large feature PR (~2,650 lines), which your contributing guide says not to open.** I am opening it anyway because the feature is finished, tested, and stable rather than a sketch, and it seemed more useful to you as working code than as an issue describing it. It is a UX addition, not a bug fix. **Please close it without ceremony if it is not a direction you want** — no reply needed, and I will not re-open or follow up.

Supersedes #5945 — same feature, re-cut as a single clean commit on current `main` with every review-bot finding from that PR folded in (Android menu tap, provider-name mismatch, backwards-clock debounce, body-read timeout, stale usage across instance rebuilds with an in-flight-pull guard, percent/reset alias fallback).

## What Changed

Adds account usage meters to the composer.

**Contract** — a provider-agnostic `ProviderUsageLimits` in `packages/contracts`, shaped `{ id, label, usedPercent, resetsAt }`, so clients map over a list instead of branching per provider.

**Server** — an in-memory `ProviderUsageLimitsStore` fed by two sources:

- `ProviderUsageIngestion` consumes the `account.rate-limits.updated` runtime events the Claude and Codex adapters **already emit** and nothing currently reads. Zero extra network calls.
- `ProviderUsageRefresher` does a debounced (60s/instance) pull for Claude only, since Claude reports usage only on request. It reads the token the `claude` CLI already stored and calls the OAuth usage endpoint, which reports utilization **without spending message quota**. Codex needs no pull — it volunteers numbers over the app-server connection.

Provider-shaped payloads are flattened in `usageLimits.ts`. Every normalizer is total: anything unparseable yields no windows rather than an error.

**Web** — a ring per window beside the send button, collapsing to the worst window when the composer is narrow. Hover gives exact percentages and reset times.

**Mobile** — one toolbar button instead of three 44pt touch targets, listing the windows on tap.

Providers that do not report usage (Cursor, Grok, OpenCode) render nothing at all, so switching threads between providers costs no layout shift.

## Why

Right now you find out you are out of quota when a turn stops halfway through. The data to prevent that is already flowing — both adapters emit rate-limit events at the end of every turn and nothing consumes them.

Design decisions worth flagging, since they are the ones you would push back on:

- **Nothing is persisted.** A usage number is only interesting while it is roughly current, and both feeds repopulate within one turn of a restart.
- **The store only announces when a number actually moves.** `updatedAt` is stamped server-side on every reading, so publishing on the stamp would wake the registry and push the entire provider array — every model, capability, and skill — to every connected client once per turn, for no visible change.
- **No tickers.** Relative labels ("resets in 2h 15m") are stamped when the popover opens, not on an interval. A continuously repainting composer pegs the GPU on high-refresh displays.
- **Every failure path draws nothing.** No credentials, locked keychain, expired token, network blip — all resolve to blank meters. There is no error a user could act on here.

## UI Changes

Hovering a meter in the web composer:

Claude usage popover in the composer, showing Session at 0%, Weekly at 17%, and Fable at 16% with reset times, beside the send button

Before is the same composer with no circles and no popover — the meters are purely additive, and providers that report no usage still render exactly that.

No motion or transitions were added, so there is no video.

## Testing

Branched off `main` at `e5c82d7`. Focused tests for the normalizers, the store's merge/debounce/announce behavior, the ingestion seam, the token reader, and the web components. Typecheck and lint clean across server, web, mobile, contracts, and client-runtime.

Two pre-existing failures in `ProviderRegistry.test.ts` and `server.test.ts` reproduce identically on unmodified `main` on my machine — they look like Windows-only arg-quoting artifacts in the spawn mocks (`Unexpected args: ^"--version^"`), unrelated to this change.

## Checklist

- [ ] This PR is small and focused — **it is not; stated plainly above**
- [x] I included a screenshot for the UI change
- [ ] I included a video for animation/interaction changes — no motion added

---

> [!NOTE]
> **Medium Risk**
> Touches provider registry, credential reads (Claude OAuth/keychain), and live snapshot publishing; failures are designed to degrade to blank meters, but instance-rebuild races and multi-account config paths need careful review.
> 
> **Overview**
> Introduces **account usage meters** in the composer so users see subscription quota before a turn fails. The feature spans contracts, server plumbing, shared client helpers, and web/mobile UI.
> 
> **Contracts & RPC** — Adds `ProviderUsageLimits` / `ProviderUsageWindow` on `ServerProvider` and a read-scoped `server.refreshProviderUsage` WebSocket RPC that schedules a pull and returns immediately; fresh numbers arrive on the provider snapshot stream.
> 
> **Server** — A volatile `ProviderUsageLimitsStore` merges **full** vs **partial** readings, debounces pulls per instance (60s), clears usage when an instance is rebuilt, and pushes only `usageLimits` changes to clients without persisting or re-upserting full snapshots. **ProviderUsageIngestion** consumes existing `account.rate-limits.updated` turn events (Claude/Codex). **ProviderUsageRefresher** pulls Claude via stored OAuth credentials and the usage API (no message quota). Codex usage is read during status probe (`account/rateLimits/read`, bounded timeout) and normalized in `usageLimits.ts`. **ProviderRegistry** decorates snapshots from the store and strips usage from disk cache.
> 
> **Clients** — Shared formatting/selection in `@t3tools/client-runtime/state/provider-usage`. **Web**: `UsageLimitsMeters` ring(s) beside send, compact mode shows worst window, refresh on focus/hover/open. **Mobile**: single gauge toolbar menu via `ControlPillMenu` render-function child (Android tap fix) and `AppState` focus refresh. **ControlPillMenu** accepts `children` as `(open) => ReactNode` so Android triggers can call `open()` after their own `onPress`.
> 
> Docs: user guide and glossary entry for usage windows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cee4fe4d468912e728cdd50fb73716ffab5d3214. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Show provider account usage meters in the composer and mobile thread view
> - Adds `ProviderUsageWindow` and `ProviderUsageLimits` schemas to `ServerProvider` snapshots, populated from Claude OAuth usage endpoint and Codex `account/rateLimits/read` during provider probes.
> - Introduces `ProviderUsageLimitsStore` (in-memory, per-instance, debounced) and `ProviderUsageIngestionLive` to merge usage data from both probe snapshots and live turn events without extra network calls.
> - Adds `ProviderUsageRefresher` service and a new `server.refreshProviderUsage` WebSocket RPC so clients can request on-demand usage pulls; results appear on the provider snapshot stream.
> - Renders animated SVG usage rings in the web composer footer (`UsageLimitsMeters`) and a `ControlPillMenu` in the mobile composer; both refresh on mount, window/app focus, and menu open.
> - Client-side formatting utilities (`formatUsagePercent`, `formatUsageResetLabel`, `formatUsageUpdatedAtLabel`, `pickWorstUsageWindow`) handle countdown vs. absolute reset labels and staleness display.
> - Risk: usage data is no longer persisted to disk; stored usage is cleared on instance rebuild, so meters are blank until the first successful refresh after restart.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cee4fe4.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/components/AppSymbol.tsx
+++ b/apps/mobile/src/components/AppSymbol.tsx
@@ -36,6 +36,7 @@ import {
   IconFolder,
   IconFolderOpen,
   IconFolderPlus,
+  IconGauge,
   IconGitBranch,
   IconHammer,
   IconGitMerge,
@@ -80,6 +81,7 @@ import { SymbolView as ExpoSymbolView, type SFSymbol, type SymbolViewProps } fro
 
 const ANDROID_ICON_BY_SF_SYMBOL: Partial<Record<SFSymbol, Icon>> = {
   "arrow.branch": IconGitBranch,
+  "gauge.with.dots.needle.bottom.50percent": IconGauge,
   "arrow.clockwise": IconRefresh,
   "arrow.down.circle": IconArrowDownCircle,
   "arrow.right.circle": IconArrowRightCircle,
--- a/apps/mobile/src/components/ControlPill.tsx
+++ b/apps/mobile/src/components/ControlPill.tsx
@@ -86,9 +86,18 @@ export function ControlPill(props: {
 // Android renders the token-styled AndroidAnchoredMenu, since the native
 // AppCompat popup can't be themed past its stock animation, metrics, and
 // submenu chrome.
+/**
+ * A render-function child keeps the trigger interactive on Android.
+ *
+ * The plain-node form there is wrapped in a `pointerEvents="none"` view so
+ * the anchor Pressable owns the tap, which silently swallows any `onPress`
+ * the child declared. Callers that need their own press side effects take
+ * `open` and call it themselves. On iOS the native menu opens on tap and the
+ * child's handlers already run, so `open` is a no-op there.
+ */
 export function ControlPillMenu(
   props: Omit<ComponentProps<typeof MenuView>, "children" | "themeVariant"> & {
-    readonly children: ReactNode;
+    readonly children: ReactNode | ((open: () => void) => ReactNode);
     readonly className?: string;
   },
 ) {
@@ -133,7 +142,12 @@ export function ControlPillMenu(
   }
 
   const { className: _className, ...menuProps } = props;
-  let children = menuProps.children;
+  // iOS opens the native menu from the tap itself, so the render-function
+  // form has nothing to trigger.
+  let children: ReactNode =
+    typeof menuProps.children === "function"
+      ? menuProps.children(() => undefined)
+      : menuProps.children;
   // In long-press mode the wrapped pressable still receives the touch (the
   // patched MenuView button is touch-transparent) and RN's Fabric touch
   // handler is never cancelled by the in-tree UIContextMenuInteraction, so a
--- a/apps/mobile/src/features/threads/ThreadComposer.tsx
+++ b/apps/mobile/src/features/threads/ThreadComposer.tsx
@@ -1,6 +1,7 @@
 import { isLiquidGlassSupported, LiquidGlassView } from "@callstack/liquid-glass";
 import type {
   EnvironmentId,
+  ProviderUsageWindow,
   MessageId,
   ModelSelection,
   OrchestrationThreadShell,
@@ -18,6 +19,7 @@ import type { ReactNode } from "react";
 import { memo, useCallback, useEffect, useMemo, useRef, useState, type RefObject } from "react";
 import {
   ActivityIndicator,
+  AppState,
   Image,
   Platform,
   Pressable,
@@ -51,7 +53,15 @@ import {
   ComposerToolbarScroller,
   ComposerToolbarTrigger,
 } from "../../components/ComposerToolbarTrigger";
-import { ControlPill } from "../../components/ControlPill";
+import { ControlPill, ControlPillMenu } from "../../components/ControlPill";
+import {
+  formatUsagePercent,
+  formatUsageResetLabel,
+  formatUsageUpdatedAtLabel,
+  pickWorstUsageWindow,
+} from "@t3tools/client-runtime/state/provider-usage";
+import { serverEnvironment } from "../../state/server";
+import { useAtomCommand } from "../../state/use-atom-command";
 import { ProviderIcon } from "../../components/ProviderIcon";
 import type { DraftComposerImageAttachment } from "../../lib/composerImages";
 import { buildModelOptions, groupByProvider } from "../../lib/modelOptions";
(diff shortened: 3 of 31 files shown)


gfsaaser24:
Addressed the Macroscope findings in `9280026e` (amended in place to keep the PR at one commit):

**Fixed**

- **Generation snapshot ordering** (`ProviderUsageRefresher.ts`) — the clear-generation is now snapshotted *before* any credential work, so a rebuild landing between the token read and the pull can no longer slip an old account's reading past the guard.
- **`clear` atomicity** (`ProviderUsageLimits.ts`) — readings, generations, and the refresh-debounce stamps now live in one `Ref`, so `clear` drops all three in a single `Ref.modify`. A concurrent `claimRefreshSlot` can no longer interleave and leave a rebuilt instance debounced for 60s.
- **Ingestion generation guard** (`ProviderUsageIngestion.ts`) — turn events now pass through the same `ifGenerationIs` guard as the OAuth pull, so a stale event from a torn-down session cannot repopulate a cleared instance.
- **Custom-instance display name** (`ChatComposer.tsx`) — the usage popover title prefers the status entry's own `displayName` before falling back to the driver-level resolution, so a custom instance's quota is no longer labeled with the default instance's name.
- **Relative `CLAUDE_CONFIG_DIR`** (`ProviderUsageRefresher.ts`) — the credential resolution now passes the workspace cwd (the same cwd the spawned CLI runs with), so a relative config-dir override resolves to the directory the CLI actually uses.

**Not changed, deliberately**

- **`updatedAt`-only readings not republishing** — this is the design trade-off described in the PR body: `updatedAt` is stamped server-side on every reading, so publishing stamp-only changes would push the full provider array to every client once per turn for no visible change. The stamp is still stored and rides out on the next real change or snapshot publish.
- **iOS menu presenting with last-render labels** — the native `UIMenu` presents synchronously from the tap and cannot be updated after presentation, so the restamp-on-open pattern that works for the JS-rendered Android menu has no iOS equivalent short of rebuilding the actions every render (which the memo exists to avoid). Worst case is minute-granularity relative labels lagging the last re-render; taking that over a per-render rebuild of the toolbar.

gfsaaser24:
Follow-up in `cee4fe4d`: the usage-change subscriber no longer round-trips a previously read snapshot through `upsertProviders`. It now rewrites only the `usageLimits` decoration on whatever snapshot is current, in a single `Ref.modify`, publishing when the array actually changed. That removes the race Macroscope flagged (a concurrent live refresh could have its newer status/capabilities rolled back by the stale re-upsert), and it also makes the earlier clear-boomerang hazard structurally impossible — the subscriber reads the store instead of re-seeding it from a snapshot.

Bil0000:
What about adding this in the same popover beside the context?

t3dotgg:
> [!NOTE]

This branch adds provider account usage to the composer through a separate provider-limit service. [#1732](https://github.com/pingdotgg/t3code/pull/1732) is the retained end-to-end provider-limit implementation, so we are closing this parallel path.

PR #6098, opened 2026-08-11
fix(desktop): prevent shared database ownership

### LINKED ISSUE #6097: [Bug]: Desktop starts a second backend against the background service database
### Before submitting

### Area

apps/desktop

### Steps to reproduce

### Expected behavior

### Actual behavior

```
database is locked
```

### Root cause

### Suggested fix

### Impact

- Desktop can fail to open.

### Version or commit

### Environment

### Workaround

matheustimbo:
## What

- add a desktop bootstrap preflight for the selected T3 state directory
- read the existing `server-runtime.json` owner record before scanning for another port
- confirm both the recorded PID and the T3 environment descriptor before blocking startup
- surface a specific recovery message instead of opening the shared SQLite database
- preserve normal startup for missing, malformed, dead, or non-T3 runtime records

## Why

When the background service already listens on port 3773, the desktop currently scans to 3774 but still launches its embedded backend with the same default T3 home. Both long-lived servers then open `userdata/state.sqlite`, causing `database is locked` HTTP 500s and duplicate resource usage.

This enforces a single live backend owner for a desktop state directory. A SQLite busy timeout can reduce transient contention, but it cannot make duplicate schedulers, command processors, and WebSocket streams safe.

## Impact

Users who launch the desktop while a background service owns the same T3 home now get a clear startup error with safe recovery instructions. They can stop the service or use a separate `T3CODE_HOME` and connect to the running environment.

## Checks

- `pnpm exec vp test run apps/desktop/src/backend/DesktopBackendDatabaseOwner.test.ts apps/desktop/src/app/DesktopAppErrors.test.ts` — 2 files, 6 tests passed
- `pnpm exec vp run --filter @t3tools/desktop typecheck` — passed
- targeted `vp fmt --check` — passed
- `git diff --cached --check` — passed

Fixes #6097

Model: GPT-5 Codex
Harness: Codex desktop app

> [!NOTE]
> ### Prevent desktop backend startup when another T3 server owns the database
> - Adds `DesktopBackendDatabaseOwner.ensureDesktopBackendDatabaseAvailable` in [DesktopBackendDatabaseOwner.ts](https://github.com/pingdotgg/t3code/pull/6098/files#diff-3bdcade13d6ef3b65a89a5404267c0a2a6d0e6d3215f7bd69c1c25c3844cae07), a pre-start guard that checks whether a previously recorded server process is still alive and serving a live T3 backend.
> - The guard reads a persisted runtime owner record from the state directory, checks PID liveness via `process.kill(pid, 0)`, and probes `/.well-known/t3/environment` on the recorded origin with a 2s timeout.
> - If all checks pass (process alive, endpoint responds 200), startup is aborted with a structured `DesktopBackendDatabaseOwnedError` containing the origin, PID, and state directory.
> - This guard is called in [DesktopApp.ts](https://github.com/pingdotgg/t3code/pull/6098/files#diff-267cf20af5c499f356b88a4bc3d3dec61aa6c7a3398f93f8836975b3bcf764e9) during bootstrap, before the backend starts.
> - Risk: if the liveness probe is slow or the state file is stale but not cleaned up, startup may be blocked unnecessarily.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 971ce3f. 2 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/apps/desktop/src/app/DesktopApp.ts
+++ b/apps/desktop/src/app/DesktopApp.ts
@@ -15,6 +15,7 @@ import * as DesktopAppIdentity from "./DesktopAppIdentity.ts";
 import * as DesktopClerk from "./DesktopClerk.ts";
 import * as DesktopApplicationMenu from "../window/DesktopApplicationMenu.ts";
 import * as DesktopWindow from "../window/DesktopWindow.ts";
+import * as DesktopBackendDatabaseOwner from "../backend/DesktopBackendDatabaseOwner.ts";
 import * as DesktopBackendPool from "../backend/DesktopBackendPool.ts";
 import * as DesktopEnvironment from "./DesktopEnvironment.ts";
 import * as DesktopLifecycle from "./DesktopLifecycle.ts";
@@ -153,6 +154,11 @@ const bootstrap = Effect.gen(function* () {
   if (environment.isDevelopment && Option.isNone(environment.configuredBackendPort)) {
     return yield* new DesktopDevelopmentBackendPortRequiredError();
   }
+
+  yield* DesktopBackendDatabaseOwner.ensureDesktopBackendDatabaseAvailable({
+    stateDir: environment.stateDir,
+    joinPath: environment.path.join,
+  });
 
   const backendPortSelection = yield* resolveDesktopBackendPort(environment.configuredBackendPort);
   const backendPort = backendPortSelection.port;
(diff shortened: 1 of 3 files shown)


t3dotgg:
> [!NOTE]

Closing this implementation because a live server that misses the two-second probe still allows a second backend to open the same database. Two simultaneous launches can also both pass the preflight. The duplicate-owner bug is valid, but the fix needs atomic ownership instead of treating an HTTP timeout as proof that the database is free.

PR #5871, opened 2026-08-09
feat(web): render color swatches next to color literals in chat markdown

mohamedmastouri-hue:
Fixes #5815

---

> [!NOTE]
> **Low Risk**
> Presentation-only chat markdown change; swatch colors come from regex-matched literals, with no auth or data-path impact.
> 
> **Overview**
> Chat messages now show a small **color preview chip** beside recognized color literals (hex, `rgb`/`rgba`, `hsl`/`hsla`) in rendered markdown.
> 
> A new **`remarkColorSwatches`** remark plugin walks text nodes, splits out matches, and emits spans tagged with `data-color-swatch`. The sanitize schema allows those span attributes (plus `className`/`style`), and a custom **`span`** renderer paints a bordered swatch using `backgroundColor` while keeping the literal text visible.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3cb7fc0e3450a8a55347ed0ce88fe00c0b068f94. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Render color swatches next to color literals in chat markdown
> - Adds a `remarkColorSwatches` remark plugin to [ChatMarkdown.tsx](https://github.com/pingdotgg/t3code/pull/5871/files#diff-70b5bb003244414202733c39403081a84415e222927a1ef415599958115c0c05) that detects hex and functional color literals (e.g. `#fff`, `rgb()`, `hsl()`) in markdown text nodes and wraps them in annotated `colorSwatch` span nodes.
> - A custom `span` renderer displays a small inline colored square next to the color text when the `data-color-swatch` attribute is present.
> - The sanitize schema is extended to allow `span` elements to retain `className`, `style`, and `dataColorSwatch` attributes so the swatches survive sanitization.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 3cb7fc0. 1 file reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

PR #6099, opened 2026-08-11
feat(web): render inline hex color swatches in chat messages

### LINKED ISSUE #5815: [Feature]: Render color swatches next to color literals in chat markdown
### Before submitting

### Area

apps/desktop

### Problem or use case

When a thread discusses colors, the values arrive as bare inline code. A line like "`#FFB3C6` on cream `#FDF6E3` is a small step in value" tells you nothing about what either color actually looks like, so reviewing a palette, a set of theme tokens, or a diff of design values means copying each literal out into a separate tool one at a time.

Code editors solved this a long time ago — VS Code ships color decorators, JetBrains shows a gutter swatch, Zed renders inline swatches, Sublime has ColorHelper, Neovim has nvim-colorizer. AI chat clients generally have not followed, so the transcript is the one place in the workflow where a color is still just text.

### Proposed solution

In the chat transcript, when an inline code span is *exactly* a CSS color literal, render a small swatch chip immediately before the text — the same convention editors use.

- Supported notations: 3/4/6/8-digit hex, and `rgb() / rgba() / hsl() / hsla() / hwb() / lab() / lch() / oklab() / oklch()`.
- Translucent colors (`#FFB3C680`, `rgba(20 24 40 / 0.35)`) sit on a checkerboard so the alpha reads.
- The swatch is decorative and `aria-hidden`, so copying the message out is unchanged.
- Spans that merely *contain* a color stay plain: `#FFB3C6 on cream`, `red`, `abcdef`.
- Existing file-path chips keep precedence, so `apps/web/src/index.css` is unaffected.

### Why this matters

Anyone doing design, theming, or CSS work with an agent can read a palette straight out of the conversation instead of round-tripping every value through a color picker. It is most useful exactly when the agent is explaining a set of related colors — the case where copying each one out is slowest.

It is also a small differentiator: this is standard in editors and absent from essentially every AI chat client.

### Smallest useful scope

Hex literals only, in the web chat markdown renderer (`apps/web/src/components/ChatMarkdown.tsx`). Hex covers the large majority of colors that appear in prose. Functional notations and the alpha checkerboard can follow.

### Alternatives considered

Prior art for the interaction: VS Code color decorators, JetBrains gutter swatches, Zed inline swatches, Sublime ColorHelper, nvim-colorizer.

Considered and rejected: matching bare CSS color keywords (`red`, `tan`, `salmon`). They appear in prose far more often than as color values, so the false-positive rate is not worth it.

Today's workaround is pasting each literal into a color picker.

### Risks or tradeoffs

- **False positives.** Requiring the span to be exactly a color literal, and skipping bare keywords, keeps this narrow.
- **Line height.** Making the `<code>` an atomic inline (`inline-flex`) causes its padding and line-height to count toward the line box, so every line holding a color gets taller than its neighbours. The swatch has to be an inline-block inside a still-inline `<code>`, the way `.chat-markdown-link-favicon` already handles link favicons.
- **Injection.** Whatever is matched ends up in a style attribute, so the value needs to be shape-validated by regex (not merely trusted) before it is used as a fill.
- **Mobile.** `apps/mobile` uses a separate native markdown pipeline (`modules/t3-markdown-text`) and would need its own change; this request is web only.

### Examples or references

Before / after — screenshots below.

### Contribution

pranav100000:
## What Changed

Render a small color-swatch chip beside a hex color literal (e.g. `#FF5733`) in assistant messages — in paragraphs, list items, headings, and table cells. Hex inside code spans / fenced blocks is left untouched, and 3/6/8-digit hex is supported.

## Why

Closes #5815. Hex colors in chat are plain text today, so you can't tell `#0F172A` from `#1E293B` at a glance. This reuses the existing inline-markdown decoration path (no new rendering pipeline), and deliberately keeps chips out of table cells' copied text so CSV copy stays faithful.

## UI Changes

Hex literals in a chat message now show a color swatch inline:

Hex color swatches in a chat message

## Checklist

### DIFF
--- a/apps/web/src/components/ChatMarkdown.tsx
+++ b/apps/web/src/components/ChatMarkdown.tsx
@@ -44,6 +44,7 @@ import rehypeSanitize, { defaultSchema } from "rehype-sanitize";
 import remarkBreaks from "remark-breaks";
 import remarkGfm from "remark-gfm";
 import { remarkGithubAlerts } from "../markdown-github-alerts";
+import { HexColorSwatch, parseHexColorLiteral } from "./chat/HexColorInlineText";
 import { renderSkillInlineMarkdownChildren } from "./chat/SkillInlineText";
 import { CHAT_FILE_TAG_CHIP_CLASS_NAME, FileTagChipContent } from "./chat/FileTagChip";
 import { PierreEntryIcon } from "./chat/PierreEntryIcon";
@@ -1634,6 +1635,15 @@ function ChatMarkdown({
             resolveInlineCodeFileLinkMeta(codeText, cwd);
           if (fileLinkMeta) {
             return fileLinkChip(fileLinkMeta, `\`${codeText}\``);
+          }
+          const hexColor = parseHexColorLiteral(codeText);
+          if (hexColor) {
+            return (
+              <code {...props} className={className}>
+                {children}
+                <HexColorSwatch color={hexColor} />
+              </code>
+            );
           }
         }
         return (
--- a/apps/web/src/components/ChatMarkdown.tsx
+++ b/apps/web/src/components/ChatMarkdown.tsx
@@ -1480,9 +1480,39 @@ function ChatMarkdown({
       );
     };
 
+    // Headings and table cells sit outside the paragraph/list-item route, so a hex
+    // literal in one renders plain. Decorate them for colors only: passing no skills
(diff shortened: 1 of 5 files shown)


pranav100000:
Closing this — there are already open PRs for #5815 (#5816 and #5871), so a third isn't needed. Deferring to those. Thanks!

PR #6068, opened 2026-08-10
fix(desktop): harden Windows and WSL support

rookepoole:
## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] I included before/after screenshots for any UI changes
- [x] I included a video for animation/interaction changes

---

> [!NOTE]
> **High Risk**
> Touches core Windows/WSL startup, networking, packaging, and release gates; regressions could break packaged WSL backends or Windows CI/release builds despite broad new test coverage.
> 
> **Overview**
> Adds **Windows/WSL CI and release gates**: native Windows checks, Ubuntu 22.04 ABI-floor builds that audit `node-pty` + bundled Linux Node, Ubuntu 24.04 smoke validation, and a Windows+real-WSL2 job that runs integration tests and **packaged portable lifecycle smoke** with artifact uploads.
> 
> **WSL backend behavior** shifts from binding `0.0.0.0` and host-side port scans to **distro-specific bind addresses**, **Linux-side TCP port allocation** (`fixed` for WSL-only primary, `wsl-auto` with ephemeral fallback for secondaries), **WSL2-only** and **packaged Node + glibc floor** preflight, and stricter failures for missing bundled runtime. Adds **structured WSL diagnostics**, IPC **retry** / diagnostic export, backend **`probeReady`** and failure hooks, and **Windows suspend/resume** reconciliation for backends.
> 
> **Windows reliability**: `taskkill /T` for dev Electron shutdown, Electron repair via upstream `install.js`, and **retried atomic file writes** for settings/catalog/logs. **WSL folder picker** resolves the distro from the running backend and blocks cross-distro UNC selections.
> 
> Release packaging now requires successful WSL ABI audit + validation before Windows artifacts consume the full `wsl-prebuild/` runtime (not just `pty.node`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35e03310eafa3d168bbc85ef0a4f4334867d3696. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Harden Windows and WSL2 support with filesystem retries, diagnostics, and ABI-audited runtime packaging
[...]

Issue #5522, opened 2026-08-06
[Bug]: Desktop stops probing backend readiness after the 60s budget, leaving "Connecting to WSL…" stuck forever even though the backend becomes healthy


lgwacker:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

1. Windows 11 + WSL2, enable **WSL backend only** in Settings > Connections.
2. Launch the desktop app under a slow WSL cold boot (server bundle loading over `/mnt/c`, Defender re-scanning freshly installed files, or right after an app update relaunches).
3. Wait on the "Connecting to WSL…" splash.

### Expected behavior

The desktop should keep probing readiness while the backend process is still alive. When the backend eventually becomes healthy (it can take 70–80s+ over the WSL/Windows filesystem bridge, and even longer right after an update), the app should connect — or at minimum surface an error and a recovery path instead of an indefinite splash.

### Actual behavior

The readiness probe is **one-shot**: `DesktopBackendManager.runBackendProcess` (apps/desktop/src/backend/DesktopBackendManager.ts:566) forks a single `waitForHttpReady` call with a fixed 60s budget. When that budget expires, `onReadinessFailure` only logs, and nothing ever probes again — even though the child process stays alive and becomes healthy ~20s later. The "Connecting to WSL…" splash spins forever with no retry and no error. This is exactly the failure documented in #4535 (73–81s boot against a 60s budget), and it is worst on the first launch after an update (cold Windows file cache + Defender re-scan push the boot well past the budget).

The backend is fine — it starts, listens, and answers HTTP 200 continuously; nothing is listening for it anymore.

### Impact

Blocks work completely

### Version or commit

main @ abc1234 (reproducible on 0.0.28+; readiness code unchanged since)

### Environment

Windows 11 x64, WSL2 (Ubuntu 24.04), T3 Code desktop app, WSL backend only

### Logs or stack traces

```
"httpReadiness.timedOut",{"baseUrl":"http://172.31.213.87:3773/",
 "requestUrl":"http://172.31.213.87:3773/.well-known/t3/environment",
 "timeoutMs":60000,"attempts":54,
 "lastFailure":{"cause":{"_tag":"BackendTimeoutError",
 "message":"Timed out waiting for backend readiness ..."}}}
```

[...]

PR #6129, opened 2026-08-11
fix(desktop): WSL backend connects with Docker bridges and slow cold starts

### LINKED ISSUE #5211: [Bug]: Desktop stuck on "Connecting to WSL" — getDistroIp picks unreachable Docker bridge IP instead of eth0

rehanhaider:
## What Changed

- `getDistroIp` now prefers the default route’s source address, with `hostname -I` as the fallback.
- `DEFAULT_BACKEND_READINESS_TIMEOUT` increases from 1 to 3 minutes.

Scope: two files, 19 lines, two commits. No tests added—the changes are one shell string and one constant.

### Why two commits and not one?
Both commits fix the same user-visible hang, so they ship together. 
- The address bug is what #5211 reports and the address commit closes it on its own; 
- the readiness budget is what kept my machine hanging afterwards. 

Both causes the exact same symptoms to the user. 

Happy to split if you'd rather take them separately.

## Why

On Windows, the desktop can remain on “Connecting to WSL…” when Docker is running inside the distro. The backend starts, listens, and answers readiness checks, but the app targets the wrong address and times out too early. Either bug can cause the hang.

### Wrong address

`getDistroIp` previously selected the first IPv4 address from `hostname -I`, which lists all bound addresses in interface order. With Docker running, a `br-*` or `docker0` bridge can appear first, directing the renderer to an address Windows cannot route to.

This also breaks mirrored networking. `isLocalHostIpv4` in `DesktopBackendConfiguration` maps the backend to loopback only when the reported address belongs to a Windows interface. A Docker bridge does not match, so mirrored-mode users lose the intended loopback path.

The new command asks the kernel for the default route’s source address using `ip -4 route get 192.0.2.1`. This avoids interface-name matching and RFC1918 heuristics without adding another spawn. The route source is printed before `hostname -I`, allowing the existing first-match parser and IPv4 validation to remain unchanged.

`hostname -I` remains the fallback, so distros without iproute2 or a default route behave as they do today. `192.0.2.1` is reserved by RFC 5737, and this is only a routing-table lookup—no packet is sent.

### Readiness timed out too early

Readiness currently receives a single 60-second budget. Loading the server bundle and dependencies across `/mnt/c` regularly takes longer. When the probe expires, the backend process remains alive but is never marked ready, leaving the splash visible even if startup completes seconds later.

The probe polls every 100 ms and resolves immediately when the backend responds, so increasing the budget does not slow fast launches.

## Verification

Verified on Windows 11 with WSL2 mirrored networking and three active Docker bridges—the topology reported in #5211:

```text
hostname -I   ->  172.18.0.1  172.17.0.1  172.19.0.1  192.168.1.5
route source  ->  192.168.1.5
```

Before the timeout change, with the address fix already applied:

- Two launches abandoned readiness at exactly 60,000 ms.
- The backend was listening and answering HTTP 200 at 98 seconds.

After both changes:

- isLocalHostIpv4 matches 192.168.1.5 to the host Wi-Fi adapter.
- Readiness targets http://127.0.0.1:<port>.
- The backend returns HTTP 200 and the desktop connects.

The address fix is evidenced by the route lookup above: hostname -I returns the bridge first, so the previous first-match selection resolves to 172.18.0.1, which Windows cannot route to. Measurements were taken against a dev build (dev:desktop), which also waits on the Vite server, so the 98s figure is not directly a packaged-build number.

Validation:

```sh
vp test run apps/desktop/src/wsl/DesktopWslEnvironment.test.ts apps/desktop/src/backend/DesktopBackendManager.test.ts
```

All 52 tests passed. Typecheck and lint are clean for both changed files.

## Related Work

This area is active:

- #5889 is a draft addressing the same IP-selection bug with two parser helpers, an octet validator, a new timeout constant, and a second sequential spawn. This PR keeps the single spawn, existing timeout, IPv4 gate, and parser; only the command changes.
- #5998 and #6068 also modify address selection as part of larger rewrites.
- #5526 replaces the one-shot readiness probe with retries.
- #5769 addresses slow startup by caching the runtime on the WSL filesystem.

There could be more robustness in the solution but I have preferred simple, tested, and workable solution instead. 

Closes #5211 
Refs #4535 and #5522.

## Checklist

- [x] No UI, animation, or interaction changes

> [!NOTE]
> 
> 
> [!NOTE]
> ### Fix WSL distro IP detection and extend backend readiness timeout
> - Updates `getDistroIpImpl` in [DesktopWslEnvironment.ts](https://github.com/pingdotgg/t3code/pull/6129/files#diff-eb443cdb82c5c1542b51d3b5bd796cd760f44161c7cd5f8ead2affd539c22d1d) to prefer the IPv4 source address of the default route (`ip -4 route get 192.0.2.1`) before falling back to `hostname -I`, which is more reliable in multi-interface WSL environments.
> - Increases `DEFAULT_BACKEND_READINESS_TIMEOUT` in [DesktopBackendManager.ts](https://github.com/pingdotgg/t3code/pull/6129/files#diff-45848e60bcec64dd7239fc34379d23407c47d0ade3f351e89a4efc7f6b42757f) from 1 minute to 3 minutes to accommodate slower WSL startup times.
> - Behavioral Change: the IP returned by `getDistroIpImpl` may differ in environments with multiple network interfaces.
> >
> 
> >
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cce7a8b.</sup>
> 
>


rehanhaider:
@codex review

CDVolvik (commented):
Checked the IP half on a real WSL2 distro. The new command emits two lines, the route source then the `hostname -I` list, and the existing parser is

```ts
```

which splits on all whitespace including newlines and takes the first match, so the route source wins and the `hostname -I` tail stays a fallback exactly as intended. Worth stating explicitly in the comment, because the fix silently depends on that parser taking the first match across the whole output rather than parsing a single line. If anyone later "tidies" it to read one line or to prefer the last match, this regresses without a test noticing.

The fallback chain also holds up: no iproute2 means `sed` gets nothing and `hostname -I` carries it, and no default route means no ` src ` field and the same fallback. Both degrade to today's behaviour rather than to nothing.

The timeout half is the part I would push back on:

```ts
-const DEFAULT_BACKEND_READINESS_TIMEOUT = Duration.minutes(1);
+// One-shot probe; a WSL backend over /mnt/c regularly needs 60-98s to answer.
+const DEFAULT_BACKEND_READINESS_TIMEOUT = Duration.minutes(3);
```

The justification is specific to a WSL backend running off `/mnt/c`, but the constant is the default for every backend. After this, a local backend that is simply dead keeps the app in "connecting" for three minutes instead of one, on machines that never had the problem. That is a real cost paid by the majority to fix a case the code can already distinguish, since the WSL path is a separate branch and could carry its own readiness timeout.

Worth flagging too that #5042 and #5769 both go after the cause rather than the symptom, by getting the backend off `/mnt/c` and onto the Linux filesystem. If either lands, the 60-98s figure this constant is sized against disappears, and a tripled global timeout is left behind as something nobody remembers to put back. A WSL-scoped override would not have that problem.

t3dotgg:
> [!NOTE]

Closing this combined revision in favor of reviewing the focused WSL work in [#5889](https://github.com/pingdotgg/t3code/pull/5889) and the recovery work in [#5998](https://github.com/pingdotgg/t3code/pull/5998). Raising the shared readiness timeout to three minutes also delays non-WSL failures and does not fix recovery. Both alternatives are still open, so the WSL cases must remain tracked.

PR #6155, opened 2026-08-11
fix(web): move command palette list a page at a time with PageUp/PageDown

Falkesand:
## What Changed

- PageUp and PageDown now move the highlighted row in command palette lists by one screenful instead of doing nothing.
- Only vertical single-column lists are handled. A grid (`cols > 1`) or a horizontal list keeps the key's default behaviour, so no key is swallowed where a straight-line jump would land in an unrelated column.
- The step is measured from the laid-out positions of the rows rather than by summing row heights, so group headings, separators, margins and gaps all count towards a screenful.
- The highlight stops on the first and last row instead of wrapping around.

The behaviour is added by a patch on `@base-ui/react`, registered the same way as the other patches in `patches/`. Base UI owns the list navigation and has no case for the page keys, and its public API exposes no way to set the highlighted index: `onItemHighlighted` only reports changes, and `actionsRef` carries just `unmount`. Both the ESM and CommonJS builds that ship in the package are patched, so every bundle gets the same behaviour.

## Why

The folder list in the "New project" dialog is the clearest case. It lists every directory in the current path, so it is routinely longer than the popup, and the only way through it was one arrow press per row. The other dialogs built on the same palette (file picker, project content search, favicon picker) had the same gap.

Home and End already work, so the page keys were the only missing step size.

Reported as #6159.

## UI Changes

No visual change; this is keyboard behaviour only.

Measured on the "New project" folder list with 31 folders, a 330px list viewport and 32px rows, so one page is 10 rows:

| Key | Highlighted row |
| --- | --- |
| ArrowDown | 0 |
| PageDown | 10 -> 20 -> 30 |
| PageDown at the end | stays on 30 |
| PageUp | 20 -> 10 -> 0 |
| PageUp at the start | stays on 0 |

The highlighted row was scrolled into view at every step. Checked on Windows in the dev build, both with real key presses and with dispatched key events.

Measured again on a two-group list (Actions and Projects, nine rows, one 28px heading between the groups, same 330px viewport) to check that headings count: summing row heights gives a step of nine rows, measuring positions gives eight. The eighth row is the one the heading pushes off screen.

## Checklist

- [x] UI changes are not applicable - no visual change, keyboard behaviour only
- [ ] I included a video for animation/interaction changes - not included; the behaviour is described with measured row indices above

### DIFF
--- a/patches/@base-ui__react@1.5.0.patch
+++ b/patches/@base-ui__react@1.5.0.patch
@@ -1,0 +1,236 @@
+diff --git a/esm/floating-ui-react/hooks/useListNavigation.js b/esm/floating-ui-react/hooks/useListNavigation.js
+index e7adca5a55c3906462b320b9488e1ff07642d32f..c3ff8b2c356240078aface81583ee2f71ee99981 100644
+--- a/esm/floating-ui-react/hooks/useListNavigation.js
++++ b/esm/floating-ui-react/hooks/useListNavigation.js
+@@ -16,6 +16,69 @@ import { activeElement, contains, getFloatingFocusElement, getTarget, isTypeable
+ import { enqueueFocus } from "../utils/enqueueFocus.js";
+ import { isVirtualClick, isVirtualPointerEvent, stopEvent } from "../utils/event.js";
+ export const ESCAPE = 'Escape';
++const PAGE_UP = 'PageUp';
++const PAGE_DOWN = 'PageDown';
++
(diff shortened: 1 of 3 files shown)


t3dotgg:
> [!NOTE]

This implements PageUp and PageDown by carrying a 236-line local Base UI patch in both ESM and CommonJS files. The optional navigation change does not justify maintaining that dependency patch.

PR #6190, opened 2026-08-11
fix(desktop): timestamps follow the OS locale instead of en-US

### LINKED ISSUE #6178: [Bug]: Desktop "Timestamp format: Locale" always renders US 12-hour times because the packaged app ships only the en-US Chromium locale
### Before submitting

### Area

apps/desktop

### Steps to reproduce

1. Use a machine whose OS region is a 24-hour locale but whose language is English — e.g. macOS with `AppleLocale = en_GB` and `AppleLanguages = ("en-GB")`.
2. Install and launch the packaged desktop app (0.0.33).
3. Leave **Settings → Timestamp format** on its default, **Locale**.
4. Look at any message timestamp in a thread.

### Expected behavior

With **Timestamp format: Locale**, times follow the operating system's locale. On an `en-GB` machine that is `15:44`.

### Actual behavior

Times render in US 12-hour form — `3:44 PM` — on every surface, regardless of the OS locale.

The cause is that the packaged desktop app ships only the `en-US` Chromium locale:

```
// scripts/build-desktop-artifact.ts:629
export const DESKTOP_ELECTRON_LANGUAGES = ["en-US"] as const;
```

which is passed straight to electron-builder at `scripts/build-desktop-artifact.ts:1542`. The installed bundle confirms the trim — exactly one locale pak survives:

```
$ ls "/Applications/T3 Code (Alpha).app/Contents/Resources/"
app-update.yml   app.asar   app.asar.unpacked   en.lproj   icon.icns   resource-monitor
```

Chromium picks its application locale from the *available* pak files, and the renderer's default `Intl` locale follows that application locale rather than the OS regional settings. With every pak but `en-US` stripped, `navigator.language` and `Intl.DateTimeFormat().resolvedOptions().locale` are pinned to `en-US` inside the desktop renderer.

That is exactly the path the timestamp helper takes. `getTimestampFormatOptions` deliberately omits `hour12` for the `"locale"` setting and defers to the runtime locale (`apps/web/src/timestampFormat.ts:13-15`), and the formatter is constructed with `undefined` as its locale argument (`apps/web/src/timestampFormat.ts:35-36`). On desktop that `undefined` can only ever resolve to `en-US`, so `"locale"` silently means "US 12-hour".

`DEFAULT_TIMESTAMP_FORMAT` is `"locale"` (`packages/contracts/src/settings.ts:19`), so every desktop user outside the US gets the wrong hour cycle out of the box.

The same `undefined`-locale assumption also affects other desktop-rendered date strings, e.g. `timeStyle: "short"` in `apps/web/src/components/settings/ConnectionsSettings.tsx:144` and `apps/web/src/components/clerk/MobileClientsUserProfilePage.logic.ts:3`, and the snooze preset labels in `packages/client-runtime/src/state/threadSettled.ts:300`.

### Impact

Cosmetic issue

### Version or commit

0.0.33 (packaged desktop app); reproduced against `main` @ 6676f9c83

### Environment

macOS 26.6.1 (build 25G76), Apple Silicon, Electron 41.5.0, `AppleLocale = en_GB`, `AppleLanguages = ("en-GB")`, `AppleICUForce24HourTime` unset. Desktop app 0.0.33.

### Workaround

Set **Settings → Timestamp format** to **24-hour** explicitly. That branch passes `hour12: false` outright (`apps/web/src/timestampFormat.ts:17-20`), so it produces `15:44` even under the pinned `en-US` locale. It does not help the surfaces that bypass the setting, and it does not fix date ordering or other locale-sensitive strings.

### Possible fix

Two independent options, either of which resolves the report:

1. **Feed the real OS locale to the renderer.** `app.getSystemLocale()` and `app.getPreferredSystemLanguages()` report the true OS values and are unaffected by the pak trim (unlike `app.getLocale()`, which returns the trimmed application locale). Pass one of those through to the renderer and use it as the explicit first argument to `Intl.DateTimeFormat` instead of `undefined`. This keeps the bundle small and fixes every `undefined`-locale call site at once.
2. **Ship more locale paks** by widening `DESKTOP_ELECTRON_LANGUAGES`. Simpler, but trades install size for the fix and only helps locales that are explicitly listed.

Option 1 looks like the smaller and more complete change. Happy to open a PR for it if that is wanted — flagging first per CONTRIBUTING.

brzzdev:
Fixes #6178.

## Problem

On the packaged desktop app, the default **Timestamp format: Locale** setting always renders US 12-hour times — `3:44 PM` on a machine whose region asks for `15:44`.

The app ships only the `en-US` Chromium locale pak (`DESKTOP_ELECTRON_LANGUAGES`, `scripts/build-desktop-artifact.ts:629`), and Chromium resolves its default `Intl` locale from the application locale rather than from the OS. Passing `undefined` to `Intl.DateTimeFormat` in the renderer therefore means `en-US` however the machine is configured, so the `"locale"` branch of `getTimestampFormatOptions` — which deliberately omits `hour12` and defers to the runtime — silently means "US 12-hour" on desktop. `DEFAULT_TIMESTAMP_FORMAT` is `"locale"`, so this is the out-of-the-box experience for every desktop user outside the US.

Measured inside the packaged renderer (0.0.33, macOS set to `en_GB`):

```js
{
  intlDefaultLocale: "en-US",          // ← should be en-GB
  navigatorLanguages: ["en-US", "en-GB"],
  renderedLikeT3:  "4:04 PM",          // Intl.DateTimeFormat(undefined, { hour: "numeric", minute: "2-digit" })
  withExplicitEnGB: "16:04",
  timeZone: "Europe/London"            // ← correct
}
```

The time zone resolving correctly while the locale does not is the tell: the OS is reachable from the process, and it is specifically the application locale that the pak trim pins.

## Fix

Read the OS locale in the main process with `app.getSystemLocale()` — which the pak trim does not affect, unlike `app.getLocale()` — expose it over the existing desktop bridge, and format timestamps with it. The bridge member follows `getAppBranding`: a sync `sendSync` getter, so the cached, synchronous formatter needs no async plumbing.

Web is unchanged: with no bridge present the resolver returns `undefined` and the browser keeps resolving its own locale, which is already correct there.

`resolveTimestampLocale` normalizes POSIX-style tags (`en_GB` → `en-GB`) and falls back to the runtime default rather than throwing if the host ever reports something ICU rejects, since the formatter backs every timestamp in the UI.

## Before / after

Chat timestamps, rendered by the real client under each locale. Same fixture, same crop; only the locale the formatter receives differs.

| Before (`en-US` — what desktop shows today) | After (OS locale `en-GB`) |
| --- | --- |
|  |  |

A note on how these were produced, so the evidence is not overstated: the bug does not reproduce in a dev run, because `electronLanguages` is an electron-builder packaging option and `pnpm start` has every locale pak. These are the web client driven under a locale-pinned Chromium context — `en-US` for before, the OS locale `en-GB` for after — which is exactly the pair of locales the formatter receives on either side of this change. The IPC half is covered by types and by the `getAppBranding` precedent it mirrors, not by these images.

## Surfaces

- **Desktop** — the only surface affected, and the only one changed.
- **Web** — unaffected; browsers resolve their own locale correctly, and the resolver returns `undefined` without a bridge.
- **Mobile** — unaffected; React Native has no Chromium pak trim.
- **Contracts** — `DesktopBridge` gains one optional member.
- **Docs** — no user docs describe timestamp formatting, so none changed.

## Out of scope

Three other call sites format with an implicit locale and are wrong on desktop for the same reason. Left alone to keep this to one concern; happy to follow up:

- `apps/web/src/components/settings/ConnectionsSettings.tsx:144`
- `apps/web/src/components/clerk/MobileClientsUserProfilePage.logic.ts:3`
- `packages/client-runtime/src/state/threadSettled.ts:300` (snooze preset labels)

## Testing

`resolveTimestampLocale` is unit tested for the tag it picks and for the hour cycle that results, including the `en-GB` → `15:44` / `en-US` → `3:44 PM` pair. `ElectronApp.test.ts` covers the OS-locale passthrough and the POSIX normalization. Targeted `vp test run` on both files (25 and 8 passing), plus typecheck and lint on the three touched packages.

Four other `ElectronApp` test stubs gained the new `systemLocale` member. Those four are among 22 desktop test files that fail to import in my environment on `main` — an unrelated `vi.mock`/`electron` issue, identical before and after this change — so they are type-correct but I could not execute them locally. `ElectronApp.test.ts` itself is not affected and runs green.

---

---

> [!NOTE]
> **Low Risk**
> Localized display-only timestamp formatting on desktop; web unchanged. Fallbacks avoid throws on bad locale tags.
> 
> **Overview**
> Fixes packaged desktop timestamps always rendering as **US 12-hour** when **Timestamp format: Locale** is selected, because Chromium’s default `Intl` locale is pinned to `en-US` (only that locale pak ships).
> 
> The main process now reads **`Electron.app.getSystemLocale()`**, normalizes POSIX tags (`en_GB` → `en-GB`), and exposes the result via a new sync **`getSystemLocale`** IPC path on `desktopBridge` (mirroring `getAppBranding`).
> 
> In **`timestampFormat.ts`**, **`resolveTimestampLocale`** validates the host tag and feeds it into cached `Intl.DateTimeFormat` instances so wall-clock times follow the OS (e.g. `15:44` for `en-GB`). Without a bridge (web), behavior is unchanged (`undefined` → browser default). Invalid tags fall back instead of breaking the UI.
> 
> **`DesktopBridge`** gains optional **`getSystemLocale`**. Desktop test stubs pick up **`systemLocale`** on `ElectronApp`; new unit tests cover locale resolution and Electron passthrough/normalization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 244ca169ca860663e7229c1b6e9f176c9c06a9f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix timestamp formatting to use OS locale instead of en-US
> - Adds `systemLocale` to the `ElectronApp` service, reading `Electron.app.getSystemLocale()` and normalizing POSIX-style tags (e.g. `en_GB` → `en-GB`).
> - Exposes the locale to the renderer via a new synchronous IPC method (`getSystemLocale`) and `window.desktopBridge.getSystemLocale()`.
> - Adds `resolveTimestampLocale` in [timestampFormat.ts](https://github.com/pingdotgg/t3code/pull/6190/files#diff-3c87b91789abf1f3644a1a66f0da1d791ac92e439c18f8a651d177cd2ed1601b) to validate the host-provided locale using `Intl.DateTimeFormat.supportedLocalesOf`, falling back to the runtime default on invalid or missing values.
> - All `Intl.DateTimeFormat` timestamp formatters now use the resolved OS locale instead of an implicit `en-US` default.
> - Behavioral Change: timestamps in the desktop app will now render according to the OS locale (e.g. 24-hour clock for `en-GB`), which may differ visually from previous output.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 244ca16.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/app/DesktopAppIdentity.test.ts
+++ b/apps/desktop/src/app/DesktopAppIdentity.test.ts
@@ -40,6 +40,7 @@ const makeElectronAppLayer = (calls: ElectronAppCalls) =>
   Layer.succeed(ElectronApp.ElectronApp, {
     metadata: Effect.die("unexpected metadata read"),
     name: Effect.succeed("T3 Code"),
+    systemLocale: Effect.succeed("en-US"),
     whenReady: Effect.void,
     quit: Effect.void,
     exit: () => Effect.void,
--- a/apps/desktop/src/app/DesktopLifecycle.test.ts
+++ b/apps/desktop/src/app/DesktopLifecycle.test.ts
@@ -21,6 +21,7 @@ describe("DesktopLifecycle", () => {
       const electronAppLayer = Layer.succeed(ElectronApp.ElectronApp, {
         metadata: Effect.die("unexpected metadata read"),
         name: Effect.succeed("T3 Code"),
+        systemLocale: Effect.succeed("en-US"),
         whenReady: Effect.void,
         quit: Effect.void,
         exit: () => Effect.void,
--- a/apps/desktop/src/electron/ElectronApp.test.ts
+++ b/apps/desktop/src/electron/ElectronApp.test.ts
@@ -8,6 +8,7 @@ const {
   autoUpdaterRemoveListenerMock,
   exitMock,
   getAppPathMock,
+  getSystemLocaleMock,
   getVersionMock,
   isDefaultProtocolClientMock,
   onMock,
@@ -29,6 +30,7 @@ const {
   autoUpdaterRemoveListenerMock: vi.fn(),
   exitMock: vi.fn(),
   getAppPathMock: vi.fn(() => "/app"),
+  getSystemLocaleMock: vi.fn(() => "en-GB"),
   getVersionMock: vi.fn(() => "1.2.3"),
   isDefaultProtocolClientMock: vi.fn(() => false),
   onMock: vi.fn(),
@@ -60,6 +62,7 @@ vi.mock("electron", () => ({
       setIcon: setDockIconMock,
     },
     getAppPath: getAppPathMock,
+    getSystemLocale: getSystemLocaleMock,
     getVersion: getVersionMock,
     isDefaultProtocolClient: isDefaultProtocolClientMock,
     isPackaged: true,
@@ -108,6 +111,23 @@ describe("ElectronApp", () => {
         resourcesPath: process.resourcesPath,
         runningUnderArm64Translation: false,
       });
+    }).pipe(Effect.provide(ElectronApp.layer)),
+  );
+
+  it.effect("reads the OS locale through the service", () =>
+    Effect.gen(function* () {
+      const electronApp = yield* ElectronApp.ElectronApp;
+
+      assert.strictEqual(yield* electronApp.systemLocale, "en-GB");
+    }).pipe(Effect.provide(ElectronApp.layer)),
+  );
+
+  it.effect("normalizes POSIX-style locale identifiers that Intl rejects", () =>
+    Effect.gen(function* () {
+      getSystemLocaleMock.mockImplementationOnce(() => "en_GB");
+      const electronApp = yield* ElectronApp.ElectronApp;
+
+      assert.strictEqual(yield* electronApp.systemLocale, "en-GB");
     }).pipe(Effect.provide(ElectronApp.layer)),
   );
 
--- a/apps/desktop/src/electron/ElectronApp.ts
+++ b/apps/desktop/src/electron/ElectronApp.ts
@@ -126,7 +126,10 @@ export const make = ElectronApp.of({
     };
   }),
   name: Effect.sync(() => Electron.app.name),
-  systemLocale: Effect.sync(() => Electron.app.getSystemLocale()),
+  // macOS derives this from NSLocale, which uses POSIX-style identifiers
+  // (`en_GB`). `Intl` rejects those outright rather than normalizing them, so
+  // the tag is normalized here rather than in the renderer that consumes it.
+  systemLocale: Effect.sync(() => Electron.app.getSystemLocale().replace(/_/g, "-")),
   whenReady: Effect.gen(function* () {
     const isPackaged = Electron.app.isPackaged;
     yield* Effect.tryPromise({
--- a/apps/desktop/src/electron/ElectronApp.ts
+++ b/apps/desktop/src/electron/ElectronApp.ts
@@ -43,6 +43,13 @@ export class ElectronApp extends Context.Service<
   {
     readonly metadata: Effect.Effect<ElectronAppMetadata, ElectronAppMetadataReadError>;
     readonly name: Effect.Effect<string>;
+    /**
+     * The OS locale, read from the operating system rather than from Chromium's
+     * resolved application locale — the packaged app ships only the `en-US`
+     * locale pak, so `app.getLocale()` and the renderer's `Intl` default are
+     * pinned to `en-US` however the machine is configured.
+     */
+    readonly systemLocale: Effect.Effect<string>;
     readonly whenReady: Effect.Effect<void, ElectronAppWhenReadyError>;
     readonly quit: Effect.Effect<void>;
     readonly exit: (code: number) => Effect.Effect<void>;
@@ -119,6 +126,7 @@ export const make = ElectronApp.of({
     };
   }),
   name: Effect.sync(() => Electron.app.name),
+  systemLocale: Effect.sync(() => Electron.app.getSystemLocale()),
   whenReady: Effect.gen(function* () {
     const isPackaged = Electron.app.isPackaged;
     yield* Effect.tryPromise({
--- a/apps/desktop/src/ipc/DesktopIpcHandlers.ts
+++ b/apps/desktop/src/ipc/DesktopIpcHandlers.ts
@@ -34,6 +34,7 @@ import {
   getAppBranding,
   getLocalEnvironmentBootstraps,
   getLocalEnvironmentBearerToken,
+  getSystemLocale,
   getWindowFullscreenState,
   openExternal,
   probeRemoteEditors,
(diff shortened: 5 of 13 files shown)


brzzdev:
Pushed a follow-up commit for the review. Rebased onto latest `main`.

**Fixed — mixed-language tooltips.** Correct, and it was a regression this PR introduced. `formatChatTimestampTooltip` builds `${time}, ${day}${ordinalSuffix} ${month} ${year}`, where the ordinal suffix and the day-before-month order are hardcoded English, so localizing the month alone really does yield `15:44, 4th Juni 2026`. `monthNameFormatter` is back on the runtime locale with a comment recording why, which also drops a line this fix never needed. Localizing that whole label is a separate change.

**Fixed — adapter boundary.** POSIX normalization (`en_GB` → `en-GB`) moved to `ElectronApp.systemLocale`, where the macOS `NSLocale` behavior that causes it belongs, and the packaging exposition left the renderer. The renderer still falls back rather than throwing on a tag it cannot use — that guard is not platform knowledge, it is that this formatter backs every timestamp in the UI and a `RangeError` there would blank all of them.

**Fixed — desktop coverage.** Two tests added to `ElectronApp.test.ts`, for the passthrough and for the normalization. Worth correcting one point: that file is not among the desktop tests that fail to import, and it runs green locally (6 → 8). The import failures are a separate pre-existing `vi.mock`/`electron` issue affecting 22 files identically on `main`; four of them happen to hold `ElectronApp` stubs that needed the new member.

**Not doing — user documentation.** `docs/user/` has no page covering timestamps, appearance, or the Settings surface this lives on, so there is nothing to update; the rule points at behavior changes a user would notice, and this restores the behavior the existing setting already promises. Writing a new user-docs page to carry a bug fix is a second concern, and I would rather you tell me you want one than have it arrive unasked.

**Not doing — shared `ElectronApp` test builder.** Agreed on the smell, but collapsing four unrelated test files into a shared stub is exactly the "also" that gets a PR split. Happy to do it separately if you want it.

Verification after the change: `ElectronApp.test.ts` 8 passing, `timestampFormat.test.ts` 25 passing, typecheck and lint clean on the three touched packages. The before/after images are unaffected — they show `formatShortTimestamp` in the chat timeline, not the tooltip.

juliusmarminge:
The current head still has the date-prefix locale bug identified by Macroscope and Bugbot. I prepared and verified the focused fix, but `maintainerCanModify` is disabled and GitHub rejected the guarded push to the fork.

Please update both `numericDateFormatter` and `numericDateWithYearFormatter` in `apps/web/src/timestampFormat.ts` to construct `Intl.DateTimeFormat(timestampLocale, ...)`, and add coverage proving an `en-GB` desktop host renders an older chat timestamp as `12/08 15:44` rather than mixing `8/12` with a 24-hour time.

Verified locally against head `244ca169ca860663e7229c1b6e9f176c9c06a9f0`: 39 focused tests passed; targeted lint and format passed; web, desktop, and contracts typechecks passed. Auto-merge should remain disabled until the actionable review is fixed; GitHub currently rejects the disable-auto-merge mutation for this PR.

PR #6197, opened 2026-08-11
fix(server): recover threads whose worktree was removed

mackinleysmith:
## Problem

A thread whose git worktree is removed becomes permanently unusable, even though nothing about the conversation was actually lost.

The failure needs two things to line up, and neither is rare:

1. **The provider session drops.** Auto-update restarts call `stopAll`; the idle reaper stops sessions after 30 minutes. Both are meant to be survivable — that is what the persisted resume cursor is for.
2. **The thread's worktree is gone.** `recoverSessionForThread` replays the persisted cwd verbatim, with no check that it still exists.

Alone, each is harmless. Together the resume fails, and because provider CLIs key their transcript store off the cwd, the failure surfaces as a generic `session not found` — which `toSessionError` then relabelled as `ProviderAdapterSessionNotFoundError: Unknown claudeAgent adapter thread`. That reads like the thread is gone. It isn't: the transcript is still on disk and the full message history is still in our own projection. Nothing clears the stale resume cursor either (`ProviderSessionDirectory.upsert` deliberately preserves it), so every later turn repeats the identical doomed resume.

Observed on a real host: a 17-minute-old thread, never deleted or archived, died on the next turn after an auto-update restart. Its worktree was gone; its transcript and all 12 of its messages were intact. Across that machine, **140 of 166 live threads** pointed at a worktree that no longer existed — including threads with 167, 144, and 126 messages.

## What this changes

Four commits, each independently reviewable:

| Commit | What |
|---|---|
| `stop reporting non-session failures as unknown sessions` | `toSessionError` matched a bare `"not found"` substring. The SDK uses that same wording for a missing CLI binary and for missing files under the resumed cwd, so install problems and dead directories were both reported as unknown threads. Now requires the message to be about a session. |
| `add worktree prune to the git driver` | A directory removed without `git worktree remove` leaves an admin entry claiming the path, which makes `git worktree add` refuse it. `git worktree prune` only drops entries whose directory is already gone. |
| `refuse to resume a session into a missing working directory` | Stat the persisted cwd before resuming; fail with a typed `ProviderSessionWorkspaceMissingError` instead of launching a doomed CLI. A failed stat is treated as present, so a transient error cannot strand a healthy thread. |
| `recreate a thread's worktree before starting a turn` | Turn start prunes and recreates the worktree from the thread's branch before the session is resolved. Best-effort — on failure the turn proceeds so existing handling reports the real reason. Either outcome appends a thread activity, because a recreated worktree means uncommitted work in the old one is gone and that should be visible. |

## Testing

- `typecheck` clean.
- 260 tests pass across `ProviderService`, `ProviderCommandReactor`, `GitVcsDriverCore`, `GitWorkflowService`, and `server.test.ts`.
- Both new tests were verified to **fail without their fix** — each production change was disabled in turn and the corresponding test went red, then restored.

## Notes for reviewers

- `ProviderService` now requires `FileSystem`. Production wiring already provided it; three test layers did not. Fixing that surfaced a latent issue where `NodeServices.layer` was passed as a sibling inside `Layer.mergeAll`, which builds in parallel and would not have satisfied the dependency — moved beneath via `Layer.provideMerge`.
- `appendWorkspaceActivity` deliberately duplicates the shape of `appendProviderFailureActivity` rather than generalising it, because the latter hardcodes `tone: "error"` and a closed `kind` union that would have to widen across every existing call site.

## What this does not fix

**It does not stop worktrees from being deleted.** I could not determine the cause — there is no cleanup timer, no script, and no automatic removal in the codebase (`removeWorktree` is only reachable from the thread-delete flow). This makes the symptom survivable rather than absent; the deletion itself deserves its own investigation.

Recovery is also limited to threads whose branch still exists. On the host above that was 136 of 140 — but a thread whose branch is also gone will still fail, now with a clear typed error instead of a misleading one.

---

> [!NOTE]
> **Medium Risk**
> Touches turn-start orchestration, provider session resume, and git worktree operations on the critical path for every turn with a worktree; behavior is mostly best-effort repair plus explicit typed failures when repair cannot run.
> 
> **Overview**
> Threads could become stuck after a provider restart when the thread’s **git worktree directory was gone**: resume replayed the persisted cwd and provider CLIs reported misleading “session not found” errors.
> 
> **Turn start** now runs **`ensureThreadWorkspace`** before resolving the provider session: if the worktree path is missing or not a directory, it **`git worktree prune`**s stale entries, **recreates the worktree** from the thread branch (best-effort), and appends **`provider.workspace.restored`** thread activity on success or failure.
> 
> **Provider resume** paths (`startSession` with a resume cursor and `recoverSessionForThread`) **`stat` the cwd** and fail with **`ProviderSessionWorkspaceMissingError`** instead of launching the adapter when the workspace is unusable; the resume cursor is preserved.
> 
> The git stack gains **`pruneWorktrees`** (`git worktree prune`). The Claude adapter only maps **session-specific** “not found” patterns to session errors. Tests and layers are updated so **`FileSystem` / `NodeServices`** satisfy session-recovery checks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e0229a27068080d9e1f9c65b510c78becc6e6c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Recover provider threads whose git worktree directory has been removed
> - Before starting a provider turn, `ProviderCommandReactor` now calls `ensureThreadWorkspace`, which checks if the thread's worktree path exists and is a directory, runs `git worktree prune` to drop stale admin entries, then recreates the worktree from the thread branch if missing. A workspace restoration activity is appended to the thread on success or failure.
> - `ProviderService` now guards session resume paths with `assertResumeWorkspaceExists`, failing with the new `ProviderSessionWorkspaceMissingError` when the persisted cwd is missing or not a directory. Fresh sessions without a resume cursor are unaffected.
> - `GitWorkflowService` and `GitVcsDriver` gain a `pruneWorktrees(cwd)` method that runs `git worktree prune` with a 15s timeout.
> - The Claude adapter's `toSessionError` now matches session-not-found errors against specific regex patterns instead of a substring check, preventing misclassification of unrelated 'not found' errors (e.g. missing CLI binary or files under the resumed cwd).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9e0229a.</sup>
> 
>


mackinleysmith:
Confirmed and fixed in `ec24a7fa` — thanks, this was a real gap and more significant than medium.

Turn start does not reach `recoverSessionForThread` at all in the scenario this PR targets. `buildSendTurnRequestForThread` calls `ensureSessionForThread` first; after a restart there is no live adapter session, so `existingSessionThreadId` is null and it goes straight to `startProviderSession` → `providerService.startSession`, which resumes from the persisted cursor. By the time `sendTurn` runs, `hasSession` is true, so the recovery path never fires. The guard was sitting on the less-travelled branch.

It was masked because `ensureThreadWorkspace` recreates the worktree before the turn, so the guard is only needed when that repair fails — branch also deleted, which was 4 of 140 cases on the host I profiled. Real, just narrow.

Fix extracts the check into `assertResumeWorkspaceExists` and applies it to `startSession` too, gated on `effectiveResumeCursor !== undefined` so a fresh session with a not-yet-provisioned workspace is unaffected. New test covers the start-session path specifically; verified it fails with the guard disabled.

The doc-comment point was also correct — it claimed the reactor catches this error and retries. It does neither. Rewritten to describe what actually happens.

mackinleysmith:
Worked through all four findings — three valid and fixed in `2f0e095d`, one I believe is incorrect.

**Valid — `resolveProject` unguarded (High).** Confirmed, and it was worse than described because I widened it. `hasHandledTurnStartRecently` caches the dedupe key *before* any work, and `processDomainEventSafely` only logs a warning — so a failure escaping this best-effort helper drops the turn with no failure activity for the full 30 minute TTL. A `resolveProject` call already existed in this handler, but only on the first-turn branch; putting one on every turn made a narrow failure mode general. Guarded as suggested.

**Valid — `exists()` accepts a non-directory (High).** Correct. Now stats and requires `type === "Directory"`. One refinement on the suggested fix: a naive `stat` + `orElseSucceed(true)` inverts the guard, because `stat` on a *missing* path fails and would be read as "fine" — which is the case the guard exists for. Only `NotFound`/`BadArgument` count as unusable; other failures still pass so a permissions or transient error cannot strand a healthy thread. Added a test with a plain file at the worktree path; verified it fails against the old `exists()` check.

**Valid — null resume cursor (Low, from Bugbot).** A persisted cursor is `null` rather than `undefined` when there is no resume state, so `!== undefined` treated a non-resuming start as a resume. Now excludes both.

**Not valid — `git worktree prune` needs `--expire now` (High).** Git prunes entries whose directory is already gone unconditionally; the expiry threshold applies to other kinds of stale entry, not this one. Tested directly on git 2.50.1:

```
$ git worktree add /tmp/wtt/wt2 -b feat2 && rm -rf /tmp/wtt/wt2
$ git worktree prune -v --dry-run
Removing worktrees/wt2: gitdir file points to non-existent location
```

and a plain `prune` followed by `git worktree add` at the same path succeeds. Adding `--expire now` would widen what gets pruned beyond this method's purpose, so I have left it as is — happy to revisit if you have a case where the plain form is insufficient.

mackinleysmith:
Valid — fixed in `c2cec3ae`.

Confirmed by convention rather than just reasoning: this file re-raises interrupts in **eight** other places (`processDomainEventSafely`, the title-regeneration handlers, and others), and my three `catchCause` sites in `ensureThreadWorkspace` were the only ones that did not. Since the helper runs inline on the turn-start fiber, a shutdown or worker drain mid-restore was flattened into a warning (or `false` for `createWorktree`) and the fiber continued into session start — which for a PR about session lifecycle is a bad way to lose a shutdown.

All three now re-raise via `Effect.failCause` when `Cause.hasInterruptsOnly(cause)`.

Scoped deliberately to the `catchCause` sites: the two `orElseSucceed` recoveries in the same helper (`fileSystem.exists`, `resolveProject`) recover from a *typed failure* only and never observe an interrupt, so they did not need changing. Same for the `Effect.catch` on `stat` in `ProviderService`.

Also rebased onto the `main` merge that landed on the branch. Re-verified against it: typecheck clean, 262 tests passing across the six affected files.

t3dotgg:
> [!NOTE]

Closing in favor of [#7839](https://github.com/pingdotgg/t3code/pull/7839). The missing-worktree recovery from this PR was extracted and merged there. Any remaining session-resume or error-classification changes can be proposed separately.

PR #6233, opened 2026-08-11
fix(web): keep interrupted environment queries pending

serabi:
## What Changed

- Prevent cancelled requests from appearing as errors.
- Keep showing real errors when something actually goes wrong.
- Add tests for both cases.

## Why

T3 Code was cancelling requests when reconnecting and I was seeing a confusing internal message:

`All fibers interrupted without error`

This made the Pull Requests page look broken even though retrying often worked. This PR adds an update to show the normal loading state until a new result arrives.

I tested this both before and after with a couple projects. The logs showed several cancelled pull-request requests, followed by a successful response. After the fix, the error screen did not appear.

## UI Changes

There is no new design. The app now shows its existing loading state instead of an internal error.

Screenshot from before:

Could not load pull requests error

## Checklist

> [!NOTE]
> ### Fix `useEnvironmentQuery` to treat interrupted failures as pending
> When an environment query is interrupted (e.g. due to cancellation), `useEnvironmentQuery` previously surfaced the interruption as an error. It now checks `Cause.hasInterruptsOnly` on failure causes and, if true, treats the result as pending with no error instead. Only genuine (non-interrupted) failures expose an error message. Tests in [query.test.ts](https://github.com/pingdotgg/t3code/pull/6233/files#diff-e270f494d4a812045f777068cd8cf4145957f720a4c688bb08570b6937e35cf3) cover both cases.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0c80679.</sup>
> 
>

### DIFF
--- a/apps/web/src/state/query.test.ts
+++ b/apps/web/src/state/query.test.ts
@@ -1,0 +1,35 @@
+import * as Cause from "effect/Cause";
+import { AsyncResult, Atom } from "effect/unstable/reactivity";
+import { describe, expect, it, vi } from "vite-plus/test";
+
+const harness = vi.hoisted(() => ({
+  result: null as AsyncResult.AsyncResult<unknown, unknown> | null,
+  refresh: vi.fn(),
+}));
+
+vi.mock("@effect/atom-react", () => ({
+  useAtomRefresh: () => harness.refresh,
+  useAtomValue: () => harness.result,
+}));
+
+import { useEnvironmentQuery } from "./query";
+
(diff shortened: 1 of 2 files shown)


t3dotgg:
> [!NOTE]

Closing as covered by [#8117](https://github.com/pingdotgg/t3code/pull/8117). The shared query runtime now waits through reconnects and runs the query again when the session is ready. That addresses the reported interruption without making every interruption look like permanent loading in the web wrapper.

PR #6248, opened 2026-08-12
fix(connection): keep Windows reconnects responsive

ProtonDev-sys:
## Problem

On Windows, the desktop client could become unreliable during busy turns or reconnects: requests crossed the 15-second warning threshold, the client displayed repeated reconnect banners, and newly submitted messages could fail.

The trace showed three compounding causes:

- local trace append/rotation used synchronous filesystem calls on the Node.js event loop;
- reconnect hydration spawned repeated Git/favicon work and revalidated every mounted asset URL at once;
- lazy provider/settings PubSub streams left a startup window where a change could be published before the consumer subscribed.

## Fix

- Add serialized asynchronous append and rotation to `RotatingFileSink`, and queue trace flushes without blocking the event loop.
- Cache/coalesce repository identity and favicon discovery, increase bounded cold-snapshot repository resolution from 4 to 16, and cap client asset URL RPC fan-out at 8.
- Acquire managed-provider and settings subscriptions before forking their consumers, then start the consumer fibers immediately.

This keeps the existing event-sourced contracts intact and bounds pressure at the adapter/client edges.

## Validation

Based on upstream `main` at `560d4a456`.

- Focused regression suite: **7 files, 110 tests passed**.
- Changed-file formatting: **16/16 passed**.
- Changed-file lint: **0 warnings, 0 errors**.
- Package-scoped typechecks passed for `t3`, `@t3tools/client-runtime`, and `@t3tools/shared` (only existing suggestions outside this diff).
- The desktop dev runner built and launched the web, server, and Electron surfaces against an isolated T3 home.
- A 100,000-record trace stress run completed in **363 ms** with **0.8 ms maximum timer lag**; all retained NDJSON records were valid and every rotated file stayed within the configured limit.

### Local reconnect E2E

The real Electron dev shell and a paired local renderer ran against the same isolated backend. I submitted a real Codex turn, terminated only that isolated backend process, verified the desktop shell automatically restarted it on the same port in about five seconds, and then submitted both Low and Ultra post-reconnect turns.

| Turn | Result | Duration | Checkpoint |
| --- | --- | ---: | --- |
| Initial | `E2E-CONNECTION-OK` | 5.818 s | ready |
| Post-reconnect | `E2E-RECONNECT-OK` | 7.893 s | ready |
| Post-reconnect Ultra | `E2E-ULTRA-RECONNECT-OK` | 2.878 s | ready |

All **41 orchestration command receipts were accepted**, with **0 error receipts**. The client showed none of `Some requests are slow`, `Failed to connect`, `Reconnecting`, or `disconnected` after the reconnect and Ultra turn.

## Surface notes

- The asset concurrency guard lives in `packages/client-runtime`, so web, desktop, and mobile share it.
- Server-side changes apply to the same WebSocket path used by local, remote/relay, and tunnel connections; local desktop was the integrated path exercised here.
- Managed subscription handling is provider-generic.
- No wire-contract or user-workflow changes were required, so there is no user-facing documentation change.

Implemented and validated with GPT-5.6 Sol in T3 Code through the Codex harness.

### DIFF
--- a/apps/server/src/orchestration/Layers/ProjectionSnapshotQuery.ts
+++ b/apps/server/src/orchestration/Layers/ProjectionSnapshotQuery.ts
@@ -350,7 +350,12 @@ const makeProjectionSnapshotQuery = Effect.gen(function* () {
   const threadPlanProgress = yield* ThreadPlanProgressService;
   const sql = yield* SqlClient.SqlClient;
   const repositoryIdentityResolver = yield* RepositoryIdentityResolver.RepositoryIdentityResolver;
-  const repositoryIdentityResolutionConcurrency = 4;
+  // A cold shell snapshot may contain hundreds of workspaces. Git process
+  // startup is comparatively expensive on Windows, so a four-wide queue can
+  // hold the initial WebSocket subscription open for tens of seconds. Keep the
+  // fan-out bounded while allowing enough parallelism to finish before the
+  // client's slow-request threshold.
+  const repositoryIdentityResolutionConcurrency = 16;
   const resolveRepositoryIdentitiesForProjects = Effect.fn(
     "ProjectionSnapshotQuery.resolveRepositoryIdentitiesForProjects",
   )(function* (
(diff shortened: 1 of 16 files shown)


ProtonDev-sys:
Closing this zero-diff PR because its head repository was an independent mirror rather than a GitHub fork. Recreating it from the proper fork so GitHub can compute and review the patch.

Issue #5900, opened 2026-08-09
[Bug]: Windows port discovery re-spawns a doomed PowerShell/WMI probe every 3 s (the port-scanner half of #4182, not fixed by #2679)


dominicwild:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

> Closest existing report is #4182, which mentioned this probe among several CPU problems. It was closed as fixed by #2679, but that PR replaced the process-diagnostics polling only; the changed-files list contains no `PortScanner.ts`, and the code below is unchanged on current `main`. Filing this as a focused issue for the remaining path.

### Area

apps/server

### Steps to reproduce

1. On a Windows machine, time the exact probe command from `apps/server/src/preview/PortScanner.ts` (line 243):
   ```powershell
   Measure-Command { powershell.exe -NoProfile -NonInteractive -Command 'Get-NetTCPConnection -State Listen -ErrorAction Stop | ForEach-Object { $processName = (Get-Process -Id $_.OwningProcess -ErrorAction SilentlyContinue).ProcessName; Write-Output "$($_.LocalAddress)|$($_.LocalPort)|$($_.OwningProcess)|$processName" }' }
   ```
   The bug triggers whenever this exceeds the scanner's 5 s timeout (`WINDOWS_LISTENER_TIMEOUT_MS`, line 55). On my machine it takes **48.2 s** with 89 listening sockets.
2. If it finishes under 5 s on your machine, open artificial listeners in a separate PowerShell window and re-run the timing. The cost is per listener (roughly half a second each on my machine), so a few hundred is plenty:
   ```powershell
   $listeners = 1..300 | ForEach-Object { $l = [Net.Sockets.TcpListener]::new([Net.IPAddress]::Loopback, 0); $l.Start(); $l }
   ```
   Verified here: adding these 300 listeners took the probe from 48.2 s (89 listeners) to **215.2 s** (390 listeners), i.e. linear in listener count.
3. Start T3 Code (Alpha) and leave it open with a session active.
4. Watch process creation (Process Explorer), or count probe launches in the classic "Windows PowerShell" event log, where event 400 records the full command line:
   ```powershell
   (Get-WinEvent -FilterHashtable @{LogName='Windows PowerShell'; Id=400; StartTime=(Get-Date).AddHours(-1)} |
     Where-Object { $_.Message -match 'Get-NetTCPConnection' }).Count
   ```

### Expected behavior

[...]

PR #6254, opened 2026-08-12
perf(server): stop Windows port discovery thrashing WMI

RioPlay:
## Problem

On busy Windows machines, preview port discovery re-spawns a PowerShell listener probe every few seconds. The probe called `Get-Process -Id` once per listening socket, which often exceeded the 5s timeout. Timed-out probes were killed mid-WMI query and retried forever, driving sustained WMI/CPU load (#5900).

## Fix

- Build a single PID→name map with one `Get-Process` per probe instead of per connection.
- After a probe timeout, cool the expensive path off for ~60s and use the existing common-port fallback so a doomed probe is not re-spawned every poll.

## Verification

- `vp test run apps/server/src/preview/PortScanner.test.ts` (6 passed)
- Live on Windows: Preview **Local servers** listed listening ports with process names (map-once path), including an intentional `localhost:3000` listener.

## Checklist

- [x] No UI changes

---
Model: Grok 4.5 · Harness: Grok Build / T3 Code

---

> [!NOTE]
> **Medium Risk**
> Changes Windows process probing and port discovery behavior; during cooldown or concurrent scans, results may omit process names via the common-port fallback.
> 
> **Overview**
> Stops Windows preview port discovery from repeatedly spawning expensive PowerShell/WMI probes that were timing out and thrashing CPU (#5900).
> 
> Replaces per-listener `Get-Process -Id` with a single `WINDOWS_LISTENER_COMMAND` that builds the PID→name map once. Adds **single-flight** gating and a **60s wall-clock cooldown** after timeouts so concurrent `scan()` / poll / retain paths fall back to common ports instead of stacking probes. Uses `acquireUseRelease` so the in-flight flag clears even if the probe fiber is interrupted.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 920f8251d9c63d0b7e78995ae4bcb3b65f5e6fbf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Stop Windows port discovery from thrashing WMI with per-listener `Get-Process` calls
> - Replaces the per-connection `Get-Process -Id` PowerShell calls with a single PID-to-name map built once per probe via the new `WINDOWS_LISTENER_COMMAND` in [PortScanner.ts](https://github.com/pingdotgg/t3code/pull/6254/files#diff-0f6b3292e660acf982893ba130c26e1ffa8f91ff09032cce9b07d55074e296f4).
> - Adds single-flight enforcement so at most one PowerShell probe runs at a time; concurrent `scan()` calls skip the probe and fall back to common-port results while the probe is in flight.
> - Arms a wall-clock cooldown (`WINDOWS_LISTENER_COOLDOWN_MS`) after a `ProcessTimeoutError` so subsequent scans skip PowerShell until the cooldown expires.
> - The in-flight flag is cleared via `acquireUseRelease` so fiber interruption cannot leave the gate permanently locked.
> - Tests covering all four behaviors (single map build, cooldown, single-flight, interrupted probe) are added in [PortScanner.test.ts](https://github.com/pingdotgg/t3code/pull/6254/files#diff-98bc68e3fec0e5eecd2296d96c5534119745d09603b318b5c1f35045a7325af9).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 920f825.</sup>
> 
>

### DIFF
--- a/apps/server/src/preview/PortScanner.test.ts
+++ b/apps/server/src/preview/PortScanner.test.ts
@@ -122,6 +122,101 @@ effectIt.layer(TestPortDiscoveryLive)("PortDiscovery integration (TCP probe fall
   );
 });
 
+effectIt("Windows listener probe builds the process-name map once", () =>
+  Effect.gen(function* () {
+    let seenCommand: string | undefined;
+    const layer = PortScanner.layer.pipe(
+      Layer.provide(
+        Layer.mergeAll(
+          Layer.succeed(ProcessRunner.ProcessRunner, {
+            run: (input) => {
+              seenCommand = input.args.at(-1);
+              return Effect.succeed({
+                stdout: "127.0.0.1|5173|4242|node\n",
+                stderr: "",
+                code: 0,
+                timedOut: false,
+                stdoutTruncated: false,
+                stderrTruncated: false,
+                stdoutInvalidUtf8: false,
+                stderrInvalidUtf8: false,
+              });
+            },
+          }),
+          Layer.succeed(Net.NetService, {
+            canListenOnHost: () => Effect.succeed(true),
+            isPortAvailableOnLoopback: () => Effect.succeed(true),
+            reserveLoopbackPort: () => Effect.succeed(40_000),
+            findAvailablePort: (preferred) => Effect.succeed(preferred),
+          }),
+          Layer.succeed(HostProcessPlatform, "win32"),
+        ),
+      ),
+    );
+
+    const servers = yield* Effect.gen(function* () {
+      const scanner = yield* PortScanner.PortDiscovery;
+      return yield* scanner.scan();
+    }).pipe(Effect.provide(layer), Effect.scoped);
+
+    expect(seenCommand).toBe(PortScanner.WINDOWS_LISTENER_COMMAND);
+    // Regression for #5900: never call Get-Process -Id per listener.
+    expect(seenCommand).toContain("$m = @{}");
+    expect(seenCommand).not.toMatch(/Get-Process\s+-Id/);
+    expect(servers).toEqual([
+      {
+        host: "localhost",
+        port: 5173,
+        url: "http://localhost:5173",
+        processName: "node",
+        pid: 4242,
+        terminal: null,
+      },
+    ]);
+  }),
+);
+
+effectIt("Windows listener probe cools down after a timeout", () =>
+  Effect.gen(function* () {
+    let probeRuns = 0;
+    const layer = PortScanner.layer.pipe(
+      Layer.provide(
+        Layer.mergeAll(
+          Layer.succeed(ProcessRunner.ProcessRunner, {
+            run: () => {
+              probeRuns += 1;
+              return Effect.fail(
+                new ProcessRunner.ProcessTimeoutError({
+                  command: "powershell.exe",
+                  argumentCount: 4,
+                  timeoutMs: 5_000,
+                }),
+              );
+            },
+          }),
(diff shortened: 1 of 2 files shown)


RioPlay:
Addressing issues found by macroscopeapp and cursor... will submit changes soon.

RioPlay:
Pushed a follow-up commit that addresses the Macroscope/Bugbot findings on the first commit:

- **Wall-clock cooldown** (windowsListenerCooldownUntilMs) instead of a tick counter, so retain/scan/poll cannot burn the cool-off early
- **Single-flight claim** before spawning PowerShell, so overlapping poll + retain cannot both run the probe
- Timeout still arms the ~60s cool-off; concurrent callers skip to the common-port fallback while a probe is in flight

Tests: `vp test run apps/server/src/preview/PortScanner.test.ts` (7 passed), including concurrent `scan()` single-flight.

CDVolvik (commented):
Collapsing the per-listener `Get-Process -Id` into one map is clearly right, that was the quadratic part. Two things about the new gate though.

**The cooldown only arms on timeout, and spawn failure is the case that repeats forever.**

```ts
ProcessSpawnError: recoverWindowsProbeFailure,
...
ProcessTimeoutError: (error) =>
  armWindowsListenerCooldown.pipe(Effect.zipRight(recoverWindowsProbeFailure(error))),
```

A timeout is transient, and it gets the 60s cool-off. A spawn failure is usually not transient — `powershell.exe` blocked by policy, AppLocker, a stripped PATH — and it gets nothing, so the 3s poll retries the spawn indefinitely. That is a steadier drip of process creation than the timeout case this PR is fixing. Whatever the cool-off is worth after a timeout, it is worth more after a spawn error.

**While skipping, the scan publishes the common-port list rather than holding the last snapshot.**

`skip` returns `null`, and `null` falls through to `probeCommonPorts()`. So during the 60s cooldown, and during any single-flight overlap, every tick publishes only `COMMON_DEV_PORTS` instead of the real listener set. At a 3s interval that is roughly twenty ticks where a dev server on a non-common port drops out of the snapshot and then returns. Listeners see it as the server disappearing rather than as degraded discovery.

Returning `state.lastSnapshot` on `skip` would keep the last good answer instead of replacing it with a narrower one. That also makes the distinction meaningful: common ports as a genuine fallback when there is nothing better, last snapshot when there is.

Smaller notes:

- `Date.now()` rather than `Clock` means the cooldown window cannot be driven by `TestClock`, so the 60s behaviour is hard to cover. The comment explains the wall-clock intent, and `Clock.currentTimeMillis` is also wall-clock under the live layer, so it would keep the intent and stay testable.
- The title says WMI, but `Get-NetTCPConnection` is still a CIM cmdlet, so the WMI round trip stays and only its frequency drops. Worth wording so nobody later reads this as WMI having been removed from the path.
- The branch is currently CONFLICTING against main.

The `acquireUseRelease` comment explaining why it is not claim-plus-ensuring is a nice touch, that interrupt window is easy to miss.

t3dotgg:
> [!NOTE]
>

Closing in favor of [#9520](https://github.com/pingdotgg/t3code/pull/9520), which replaces Windows PowerShell listener discovery with native enumeration. Review continues there, including Windows verification. That removes the polling path this branch tries to reduce. The listener PID mapping, interrupted scan recovery, and non-common-port fallback cases are recorded on the retained PR.

PR #6268, opened 2026-08-12
fix(scripts): run worktree setup without POSIX shell

RioPlay:
## Problem

Worktree setup from `t3.json` is typed into a terminal via `ProjectSetupScriptRunner`. The command used `ln -sf`, `&&`, and `` expansion, which fails under Windows PowerShell/cmd — so new worktrees on Windows never get deps, env files, or a warm web cache.

## Fix

Replace the shell one-liner with `node scripts/setup-worktree.ts` that:

1. Runs `vp i` with the worktree as cwd
2. Links project-root `.env` files into the worktree (copy if symlink is denied)
3. Warms the web dep cache

Does not delete a worktree `.env` when the project root has no source yet. Parent dirs are created as needed.

## Test plan

- [x] `vp test run scripts/setup-worktree.test.ts` (link / copy / skip-missing / same-path)
- [ ] Create a worktree on Windows and confirm setup terminal runs without `ln` errors
- [ ] Confirm `.env` appears in the worktree (link or copy) when present on the project root

---

Model: Grok 4.5 · Harness: Grok Build

---

> [!NOTE]
> **Medium Risk**
> Touches worktree bootstrap and `.env` linking/copying with filesystem staging/rollback; a bug could leave worktrees without deps or with wrong env files, especially on Windows.
> 
> **Overview**
> Replaces the POSIX-only worktree setup command in `t3.json` (`vp i && ln -sf ...`) with `node scripts/setup-worktree.ts`, so setup works under PowerShell/cmd on Windows.
> 
> The new script installs deps, links `.env` and `infra/relay/.env` from the project root into the worktree (falling back to copy when symlinks are denied), then warms the web dep cache. Linking uses staging + backup/rename so a failed mid-flight write does not wipe a local worktree `.env`, and missing sources are skipped instead of deleting the destination.
> 
> Adds unit coverage for path resolution, link/copy, same-path/junction skips, missing sources, and non-file sources.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60918659d093c34beff103bfa0fe1fea2764aaf2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Replace inline shell worktree setup with a Node script in `scripts/setup-worktree.ts`
> - Replaces the inline shell command in [t3.json](https://github.com/pingdotgg/t3code/pull/6268/files#diff-ed2426d5f9a505e4adcba4a25328883302241ea69c3132391a87bc8dc01b80ba) with `node scripts/setup-worktree.ts`, removing the dependency on a POSIX shell.
> - The new script resolves `projectRoot` and `worktree` from environment variables, installs dependencies via `vp i`, links or copies configured `.env` files, and warms the web dep cache.
> - Env file linking uses an atomic staging/rename flow: symlink when possible, copy as fallback, and never deletes an existing worktree env when the source is missing.
> - Behavioral Change: env file handling now skips when source and destination resolve to the same path (symlink/junction-aware), replacing the previous unconditional symlink creation.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6091865.</sup>
> 
>

### DIFF
--- a/scripts/setup-worktree.test.ts
+++ b/scripts/setup-worktree.test.ts
@@ -91,4 +91,27 @@ describe("setup-worktree", () => {
       NodeFs.rmSync(root, { recursive: true, force: true });
     }
   });
+
+  it("keeps the worktree env file when staging the replacement fails", () => {
+    const root = makeTempDir("setup-worktree-fail-root-");
+    const worktree = makeTempDir("setup-worktree-fail-wt-");
+    try {
+      // Source path exists but is a directory so symlink("file")/copyFile both fail.
+      NodeFs.mkdirSync(NodePath.join(root, ".env"));
+      const destination = NodePath.join(worktree, ".env");
+      NodeFs.writeFileSync(destination, "LOCAL=1\n", "utf8");
+
+      expect(() =>
+        linkOrCopyEnvFile({
+          projectRoot: root,
+          worktree,
+          relativePath: ".env",
+        }),
+      ).toThrow();
+      expect(NodeFs.readFileSync(destination, "utf8")).toBe("LOCAL=1\n");
+    } finally {
+      NodeFs.rmSync(root, { recursive: true, force: true });
+      NodeFs.rmSync(worktree, { recursive: true, force: true });
+    }
+  });
 });
--- a/scripts/setup-worktree.test.ts
+++ b/scripts/setup-worktree.test.ts
@@ -1,0 +1,94 @@
+// @effect-diagnostics nodeBuiltinImport:off
+import * as NodeFs from "node:fs";
+import * as NodeOs from "node:os";
+import * as NodePath from "node:path";
+
+import { describe, expect, it } from "vite-plus/test";
+
+import {
+  linkOrCopyEnvFile,
+  linkProjectEnvFiles,
+  resolveWorktreePaths,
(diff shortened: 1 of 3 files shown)


CDVolvik (commented):
Windows verification (PowerShell 5.1 + Node 22.14, Developer Mode on).

The old `t3.json` command is broken here exactly as described:

```
ln : The term 'ln' is not recognized as the name of a cmdlet, function,
script file, or operable program.
```

`scripts/setup-worktree.test.ts`: 6/6. The link path works on this box — `symlinkSync(..., "file")` succeeded and `lstat` reports a symlink (so the copy fallback is unexercised here). `spawnSync("node", ["-e", "..."], { shell: true })` does pass the args through `cmd.exe` on this Node, so the `vp i` / `node …/warm-dep-cache.ts` spawn shape is fine.

I did not run the full setup (`vp i` + cache warm) in a real worktree. The unchecked items in the test plan are still the missing piece.

One small gap: the junction/same-path test `return`s if `symlinkSync(..., "junction")` throws, so CI on a box without that permission never sees `skipped-same-path`. The missing-source and not-a-file cases still pin the "don't delete the worktree .env" rule.

PR #6270, opened 2026-08-12
fix(web): replace whitespace in new ref names with dashes

### LINKED ISSUE #6163: [Feature]: Replace white spaces with dashes when creating a branch
### Before submitting

### Area

apps/web

### Problem or use case

I want to create the branch from the bottom-right corner of the chat view. I usually give it a name, e.g. "new branch", and I don't care if characters are a valid git branch.

### Proposed solution

Replace invalid characters (like spaces) with dashes.

### Why this matters

This is a simple change that allows faster branch creation, less headache.

### Smallest useful scope

Spaces are replaced by dashes.

### Alternatives considered

An alternative would be marking the input field in red, saying "Invalid git branch name".

### Risks or tradeoffs

Edge cases: trying to create a valid branch name with replacements, ends up with a invalid branch name.

### Examples or references

_No response_

### Contribution

jorj-pineda:
Creating a ref from the branch picker passed the typed name straight through to git, which rejects every ref name containing ASCII whitespace. Typing a name like "new branch" therefore always failed with a raw git error, so the picker accepted input it could never act on.

The typed name now has runs of whitespace collapsed to a single dash, turning "new branch" into "new-branch". The picker's "Create new ref" entry shows the sanitized name, so it no longer promises one name and creates another.

Scoped deliberately to whitespace. Case is preserved because git ref names are case sensitive, and existing dashes are left alone because consecutive dashes are valid. Names that are invalid for other reasons still surface the git error unchanged. Only the whitespace git actually rejects is replaced, ASCII space and the ASCII control characters. Names containing whitespace git accepts, such as U+00A0, are left untouched.

Closes #6163

## Before / after

| | Typed | Picker shows | Result |
| --- | --- | --- | --- |
| Before | `new branch` | `Create new ref "new branch"` | Failed to create and switch ref. |
| After | `new branch` | `Create new ref "new-branch"` | ref `new-branch` created and checked out |

**Before** (on `main`): typing "new branch" fails:
https://github.com/user-attachments/assets/90b1b4ef-2e2b-4c4c-94df-9fe2f8939647

**After** (this PR): the same input creates `new-branch`:
https://github.com/user-attachments/assets/63e94eae-c0de-43bf-8022-60fb02daca20

## Verification

- `apps/web` unit tests: 65 passed, including 9 new cases covering `sanitizeNewRefName`
- Targeted lint, typecheck (`tsgo --noEmit`) and `vp fmt --check` all pass
- Every sanitized output asserted in the tests was independently checked against `git check-ref-format`
- Verified end to end in the web client: typing "new branch" in the ref picker creates and switches to `new-branch`

> [!NOTE]
> ### Replace whitespace with dashes in new git ref names in the branch picker
> - Adds `sanitizeNewRefName` in [`BranchToolbar.logic.ts`](https://github.com/pingdotgg/t3code/pull/6270/files#diff-22eaeeb6cc7f6e680a66d29924a8aa6c56a51450578462389a4b649102ea655f) that trims input and collapses ASCII whitespace runs (spaces, tabs, newlines, etc.) into a single dash, leaving non-ASCII characters and existing dashes untouched.
> - The branch picker's search now matches refs against the sanitized query, so typing "new branch" surfaces "new-branch" in results.
> - Server-side ref queries, collision detection, and the "Create new ref" label all use the sanitized form of the user's input.
> - When creating a new ref, the name is sanitized before submission, preventing invalid git ref names containing spaces.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d2241dd.</sup>
> 
>

---

> [!NOTE]
> **Low Risk**
> Scoped UX fix in the web branch picker with thorough unit tests; no auth, persistence, or server contract changes beyond ref query strings.
> 
> **Overview**
> The branch picker no longer sends typed ref names with ASCII whitespace straight to git. **`sanitizeNewRefName`** trims input and collapses runs of spaces, tabs, and other ASCII control whitespace into a single dash (e.g. `new branch` → `new-branch`), while leaving case, slashes, user-typed dashes, and non-ASCII whitespace git accepts unchanged.
> 
> **Create new ref** uses the sanitized name for the mutation, the picker label, duplicate detection, and the paginated ref search query. Client-side filtering via **`shouldIncludeBranchPickerItem`** also matches refs when the search text sanitizes to an existing ref name, so spaced queries still surface `new-branch`.
> 
> Unit tests cover sanitization edge cases and picker filtering behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2241ddd47d38fb900edf92b887313f406629b8e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/web/src/components/BranchToolbar.logic.test.ts
+++ b/apps/web/src/components/BranchToolbar.logic.test.ts
@@ -744,6 +744,19 @@ describe("shouldIncludeBranchPickerItem", () => {
     ).toBe(true);
   });
 
+  // A partial query has to reach the ref it would have been created as, so
+  // searching "hello w" still finds an existing hello-world.
+  it("surfaces a ref from a partial query containing a space", () => {
+    expect(
+      shouldIncludeBranchPickerItem({
+        itemValue: "hello-world",
+        normalizedQuery: "hello w",
+        createBranchItemValue: null,
+        checkoutPullRequestItemValue: null,
+      }),
+    ).toBe(true);
+  });
+
   it("excludes refs matching neither the raw nor the sanitized query", () => {
     expect(
       shouldIncludeBranchPickerItem({
--- a/apps/web/src/components/BranchToolbar.logic.test.ts
+++ b/apps/web/src/components/BranchToolbar.logic.test.ts
@@ -730,11 +730,36 @@ describe("shouldIncludeBranchPickerItem", () => {
       }),
     ).toBe(false);
   });
+
+  // Typing a spaced name must still surface the ref it would have been created
+  // as, or the picker shows nothing at all for that query.
+  it("surfaces an existing ref matching the sanitized query", () => {
+    expect(
+      shouldIncludeBranchPickerItem({
+        itemValue: "new-branch",
+        normalizedQuery: "new branch",
+        createBranchItemValue: null,
+        checkoutPullRequestItemValue: null,
+      }),
+    ).toBe(true);
+  });
+
+  it("excludes refs matching neither the raw nor the sanitized query", () => {
+    expect(
+      shouldIncludeBranchPickerItem({
+        itemValue: "main",
+        normalizedQuery: "new branch",
+        createBranchItemValue: null,
+        checkoutPullRequestItemValue: null,
+      }),
+    ).toBe(false);
+  });
 });
 
-// Git rejects every ref name containing whitespace, so a typed name like
-// "new branch" can only ever fail. Sanitizing can therefore turn a failing
-// name into a working one but can never break a name that works today.
+// Git rejects ASCII space and the ASCII control characters in ref names, so a
+// typed name like "new branch" can only ever fail. Replacing exactly those can
+// turn a failing name into a working one without touching a name git already
+// accepts, including one holding non-ASCII whitespace such as U+00A0.
 describe("sanitizeNewRefName", () => {
   it("replaces a space with a dash", () => {
     expect(sanitizeNewRefName("new branch")).toBe("new-branch");
--- a/apps/web/src/components/BranchToolbar.logic.test.ts
+++ b/apps/web/src/components/BranchToolbar.logic.test.ts
@@ -748,9 +748,15 @@ describe("sanitizeNewRefName", () => {
     expect(sanitizeNewRefName("  new branch  ")).toBe("new-branch");
   });
 
-  it("replaces tabs and non-breaking spaces pasted from other apps", () => {
+  it("replaces tabs, which git rejects just like spaces", () => {
     expect(sanitizeNewRefName("new\tbranch")).toBe("new-branch");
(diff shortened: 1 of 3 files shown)


jorj-pineda:
Thanks — two of the three findings were real and are now fixed. Every review above ran against `111359af5`; the branch is three commits past that, so here is where each stands:

| Finding | Status |
| --- | --- |
| Macroscope: over-replaces Unicode whitespace | Fixed in `542ef842c` |
| Macroscope: collision check uses the raw query | Fixed in `d36efad56` |
| Cursor: collision check "still looks up the raw trimmed query" | Fixed in `d36efad56`, which predates that review |

The Unicode finding was a good catch and the original code was wrong. Checked against git directly:

| Character | `git check-ref-format --branch` |
| --- | --- |
| U+0020 space | rejected |
| U+0009 tab | rejected |
| U+00A0 no-break space | **accepted** |
| U+2009 thin space | **accepted** |

`\s` matched all four, so a name git would happily create was being silently rewritten. The sanitizer now replaces only ASCII space and the ASCII control characters, with tests asserting U+00A0 and U+2009 survive untouched.

Cursor's comment raised a second point that was still live after the collision fix: the existing ref also got filtered out of the list, so a query whose sanitized form already exists showed "No refs found" — create suppressed, matching ref hidden. Fixed in `782578157` with tests.

The approvability verdict above was produced before the first of these fixes was pushed, so it reflects the pre-fix commit.

arturo-source:
Something you might want to consider for future PRs: there are another invalid character that can be replaced by a dash.

```
const GIT_REF_FORBIDDEN_CHARS = /[\x00-\x1f\x7f ~^:?*[\\]+/g;
```

And the branch name cannot start with dash `-`.

arturo-source:
Something you might want to consider:

1. I create a branch "hello world"
2. The branch created is "hello-world"
3. Later I search for "hello w"

Even if I search with an space, the branch `hello-world` should appear as an option of an already created branch.

jorj-pineda:
@arturo-source 
Thanks for ur input! Both r useful.

On the search case: that one already works as of `ae8e65aa7`, which landed after you looked. The ref list is filtered server side with a substring match, so the raw `hello w` was dropping `hello-world` from the response before the client ever saw it. The query is sanitized now too, so the server matches on `hello-w`. Added `ac70c59d2` covering exactly your scenario, and it passes without any further source change.

On the forbidden characters: you're right that the full set is `[\x00-\x1f\x7f ~^:?*[\\]` plus the rule that a name cannot start with `-`. I'd rather keep those out of this PR. #6163 asked specifically for spaces, and handling the rest turns a small input fix into a partial reimplementation of `check-ref-format`, which deserves its own review. Names invalid for those reasons still surface the git error today, same as before.

Happy to open a separate issue for the wider character set if that's useful.

PR #6307, opened 2026-08-12
feat(web): let change request links open in the browser

Bil0000:
## What

Adds a Source Control setting for where a link to a pull or merge request opens: **T3 Code** or your **browser**.

Today any link to a change request on a repository this workspace has checked out is claimed by the app — GitHub and Bitbucket pull requests, GitLab merge requests, Azure DevOps pull requests, self-hosted installs included. That is great if you live in the Pull Requests page, and annoying if your review workflow lives in a browser tab you already have open.

**Settings → Source Control → Links → Change request links.** Defaults to T3 Code, so nothing changes unless you pick otherwise.

## What it covers

Set to Browser, every claimed link goes back to being an ordinary link:

- links an agent writes into a thread
- links in the sidebar and the branch picker
- the chat header's shortcut to the thread's own change request

The right panel's own pull request tab is deliberately unaffected — picking that surface by name is not a link, so it keeps doing what you asked it to do.

## How

One guard in `useOpenChangeRequestLink`, which every link path already routes through, so there is no per-caller version of this behaviour to keep in sync. The header shortcut is the one non-link path; it drops `onOpenPullRequest` and falls back to the browser exactly as it does when the server has no pull request capability.

The setting is client-side (`ClientSettings.changeRequestLinkTarget`), like the other "where does this open" preferences — it is about this client, not the server. It is a named two-option target rather than a boolean because a switch has to be read against its label to know what "off" does, while the dropdown says it outright and the description under it changes to match.

## Tests

- `packages/contracts/src/settings.test.ts` — defaults to `app`, both targets round-trip through the schema and the patch, and an unknown target is rejected
- `apps/desktop/src/settings/DesktopClientSettings.test.ts` — fixture updated for the new key

Affected packages: 66 tests passing, typecheck clean, lint and format clean.

CleanShot 2026-08-12 at 2  37 27

> [!NOTE]
> ### Add setting to open change request links in the browser instead of the app
> - Adds a new `changeRequestLinkTarget` client setting (`"app"` or `"browser"`, defaulting to `"app"`) defined in [settings.ts](https://github.com/pingdotgg/t3code/pull/6307/files#diff-82d7525e9c2a6a61cd5d416ceaeba40992140373eb7ac00fa3a91c2edd00bc3c).
> - Adds a UI control in [SourceControlSettings.tsx](https://github.com/pingdotgg/t3code/pull/6307/files#diff-9f2e55c161725a6350a5fe7b23c2b7ada03b60855000614a443c6253ad64b625) under Source Control settings to toggle this preference, with a reset-to-default button.
> - The `useOpenChangeRequestLink` hook in [openPullRequestLink.ts](https://github.com/pingdotgg/t3code/pull/6307/files#diff-953140e01f1329a36fe8f5606c9027215b588e76537644ab9819a9a618d000ae) now skips in-app interception when the setting is `"browser"`, letting the browser handle the link.
> - The chat header's `onOpenPullRequest` callback is only wired when the setting targets the app.
> - The new setting is indexed in settings search under "Change request links".
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c76844e.</sup>
> 
>


ParthJadhav:
Much needed !

PR #3573, opened 2026-06-26
Harden VCS polling and CLI shutdown handling

bmdavis419:
## Summary
- Bound Git/VCS refresh waiters and status broadcaster state so interrupted or stalled refreshes do not pile up.
- Added bounded child-process shutdown and explicit CLI signal escalation so stuck shutdowns eventually force exit.
- Tightened provider event log lifecycle and reduced tracing overhead on hot native log paths.
- Released idle client VCS subscriptions immediately to avoid unnecessary polling work.

## Testing
- `vp check`
- `vp run typecheck`
- `vp run lint` if available, otherwise not run
- Added/updated unit coverage for child-process shutdown, shutdown signal escalation, VCS refresh behavior, and provider event logging

> [!NOTE]
> ### Harden VCS polling and CLI shutdown handling with bounded process spawning and LRU status cache
> - **VCS status cache**: `VcsStatusBroadcaster` now maintains a bounded 128-entry LRU cache; entries are evicted when the last subscriber disconnects, and active pollers protect their entries from eviction.
> - **Upstream refresh**: `GitVcsDriverCore` serializes concurrent git fetches per shared object store, uses hardened `--no-write-fetch-head` fetch flags, backs off 30s on failure, and surfaces fetch errors via a new `refreshStatusUpstream` service method on `GitVcsDriver` and `GitWorkflowService`.
> - **CLI shutdown escalation**: `installNodeShutdownSignalEscalation` installs SIGINT/SIGTERM handlers that force-exit with a conventional code on a second signal or after a 10s timeout; installed and cleaned up in [`bin.ts`](https://github.com/pingdotgg/t3code/pull/3573/files#diff-49e5d719869710e6cd7800547c130d0f20fcf22e14bb44b6ea8db436bb9a5a5f).
> - **Bounded child process spawner**: `BoundedChildProcessSpawner` wraps the existing spawner to send SIGTERM on scope close, escalate to SIGKILL after `termGraceMs`, and detach delegate finalizers so they don't block caller scope closure; activated via `BoundedChildProcessSpawner.layer()` in the CLI runtime.
> - **Event NDJSON logging**: `EventNdjsonLogStore` replaces per-stream writers with a shared store supporting retention by age/size, oversized-record omission notices, and a configurable batch window (now 1000ms); native and canonical loggers share state in `ProviderEventLoggers`.
> - **VCS status atom**: client-side VCS status subscription atom now sets `idleTtlMs: 0`, tearing down immediately when it has no consumers.
[...]

PR #6318, opened 2026-08-12
fix(server): [Windows] bound child process shutdown

Defmon3:
## What Changed

On Windows, wrap the CLI child-process spawner with bounded shutdown behavior.

When a scoped child process does not stop after `SIGTERM`, T3 waits up to two seconds before escalating to forced process-tree termination. It then waits one additional second before allowing shutdown to continue with a warning.

The wrapper also:

- bounds explicit `handle.kill()` calls;
- preserves `unref()` as an opt-out from scope-owned termination;
- leaves non-Windows process handling unchanged.

## Why

A child process or descendant that fails to exit can otherwise prevent the T3 server from shutting down and leave its console host running.

This revisits the Windows shutdown behavior discussed in #3573 against the current server implementation. The change is limited to Windows and uses Effect's existing Windows process-tree termination behavior.

## Verification

Run on Windows with Node 24:

```powershell
vp test run apps/server/src/cli/boundedChildProcessSpawner.test.ts
vp run --filter t3 typecheck
```

The focused test suite covers:

- natural child exit;
- `unref()` behavior;
- bounded explicit termination with forced escalation;
- honoring an explicit `forceKillAfter` grace period;
- stale exit-event state after the process is already gone;
- scope closure when the underlying finalizer does not return;
- termination of a real Windows parent process and its descendant.

Expected result:

- Seven tests pass.
- Server typecheck passes.

## Checklist

Implemented with GPT-5.6 Sol through the Codex harness.

> [!NOTE]
> ### Add bounded child process shutdown on Windows with SIGTERM/SIGKILL grace periods
> - Introduces `BoundedChildProcessSpawner` in [boundedChildProcessSpawner.ts](https://github.com/pingdotgg/t3code/pull/6318/files#diff-18cf7a10f780abb762046c9a45964767a6a17708cfd9992210d4ff1374dff5fa), which wraps the existing `ChildProcessSpawner` service to enforce a timed shutdown sequence: send an initial signal (default SIGTERM), wait a configurable grace period, escalate to SIGKILL if still running, and log a warning if the process still does not stop.
> - On Windows, [server.ts](https://github.com/pingdotgg/t3code/pull/6318/files#diff-a8f34e8f17024480f0ca046c8ea4a8af3b525fe68940b703ba496fa4aa4ddfbd) merges `BoundedChildProcessSpawner.layer` into `PlatformServicesLive`, replacing the default spawner with the bounded implementation.
> - Explicit `kill(options)` calls while the caller scope is open are forwarded unchanged to the delegate; bounded shutdown only triggers on scope close or re-reference after scope close.
> - Behavioral Change: on Windows, child processes now receive SIGTERM before SIGKILL on shutdown rather than being killed immediately.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 97dd9fa.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Touches core process lifecycle on Windows only, but mis-timed kills or scope/finalizer ordering could affect CLI children during shutdown or after unref.
> 
> **Overview**
> **On Windows only**, the server wraps Effect’s `ChildProcessSpawner` with **`BoundedChildProcessSpawner`** so scoped children cannot block shutdown indefinitely.
> 
> When the caller scope closes (or a child is re-referenced after close), the wrapper runs **bounded termination**: **SIGTERM**, wait up to **2s** (`termGraceMs`), then **SIGKILL** with a **1s** grace (`killGraceMs`). Signals are sent in a detached fiber so parent scope finalizers are not held on a stuck delegate cleanup. **`unref()`** still opts out of scope-driven kills; explicit **`kill()`** while the scope is open is forwarded unchanged.
> 
> `PlatformServicesLive` in **`server.ts`** composes this layer on top of Bun/Node platform services when `isHostWindows` is true; non-Windows behavior is unchanged.
> 
> Adds **`boundedChildProcessSpawner.test.ts`** (mocked lifecycle cases plus a Windows integration test for parent + descendant process trees).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97dd9fa05c15fe04f45596d8a297ccc558461191. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/server/src/cli/boundedChildProcessSpawner.test.ts
+++ b/apps/server/src/cli/boundedChildProcessSpawner.test.ts
@@ -1,0 +1,323 @@
+import { NodeServices } from "@effect/platform-node";
+import { expect, it } from "@effect/vitest";
+import { HostProcessPlatform } from "@t3tools/shared/hostProcess";
+import * as Deferred from "effect/Deferred";
+import * as Effect from "effect/Effect";
+import * as Exit from "effect/Exit";
+import * as Fiber from "effect/Fiber";
+import * as Layer from "effect/Layer";
+import * as Option from "effect/Option";
+import * as PlatformError from "effect/PlatformError";
+import * as Scope from "effect/Scope";
+import * as Sink from "effect/Sink";
+import * as Stream from "effect/Stream";
+import { TestClock } from "effect/testing";
+import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process";
+
+import * as BoundedChildProcessSpawner from "./boundedChildProcessSpawner.ts";
+
+const makeHandle = (input: {
+  readonly pid: number;
+  readonly exitCode: Effect.Effect<ChildProcessSpawner.ExitCode>;
+  readonly isRunning: Effect.Effect<boolean>;
+  readonly kill?: ChildProcessSpawner.ChildProcessHandle["kill"];
+  readonly unref?: ChildProcessSpawner.ChildProcessHandle["unref"];
+}) =>
+  ChildProcessSpawner.makeHandle({
+    pid: ChildProcessSpawner.ProcessId(input.pid),
+    exitCode: input.exitCode,
+    isRunning: input.isRunning,
+    kill: input.kill ?? (() => Effect.void),
+    unref: input.unref ?? Effect.succeed(Effect.void),
+    stdin: Sink.drain,
+    stdout: Stream.empty,
+    stderr: Stream.empty,
+    all: Stream.empty,
+    getInputFd: () => Sink.drain,
+    getOutputFd: () => Stream.empty,
+  });
+
+const isProcessAlive = (pid: number) => {
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch (cause) {
+    return !(cause instanceof Error && Reflect.get(cause, "code") === "ESRCH");
+  }
+};
+
+const readFixturePids = (line: string) => {
+  const value: unknown = JSON.parse(line);
+  if (typeof value !== "object" || value === null) {
+    throw new Error("Expected the child fixture to print an object");
+  }
+  const pid = Reflect.get(value, "pid");
+  const descendantPid = Reflect.get(value, "descendantPid");
+  if (typeof pid !== "number" || typeof descendantPid !== "number") {
+    throw new Error("Expected numeric child fixture pids");
+  }
+  return { pid, descendantPid };
+};
+
+const waitForProcessesToStop = Effect.fn("waitForProcessesToStop")(function* (
+  pids: ReadonlyArray<number>,
+  timeoutMs: number,
+) {
+  let remainingMs = timeoutMs;
+  while (pids.some(isProcessAlive)) {
+    if (remainingMs <= 0) return false;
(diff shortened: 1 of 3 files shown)


CDVolvik (commented):
The test that covers the behaviour this PR is named for does not report as skipped on other platforms, it reports as passed:

```ts
it.effect("kills a Windows child process tree without blocking scope close", () => {
  if (process.platform !== "win32") return Effect.void;
```

Running the file on Linux:

```
 ✓ closes the delegate scope after a natural child exit                        5ms
 ✓ preserves unref as an opt-out from scope-owned termination                   1ms
 ✓ bounds explicit handle kill calls and escalates to SIGKILL                   2ms
 ✓ respects an explicit force-kill grace period                                 4ms
 ✓ treats a missing pid as stopped when the delegate exit event is stale        1ms
 ✓ returns from scope close before the delegate finalizer can block             1ms
 ✓ kills a Windows child process tree without blocking scope close              0ms

 Tests  7 passed (7)
```

The last line is a green check on a run where nothing executed. Anyone reading CI concludes the tree kill is covered. `it.effect.skipIf(process.platform !== "win32")` reports it as skipped instead, which is the honest signal and costs nothing.

That matters more than usual here, because the tree kill is the only part that cannot be inferred from the other six tests. Those all drive a synthetic handle whose `kill` just pushes a string into an array, so they verify the escalation state machine rather than that anything dies. On Windows `kill()` terminates the process it targets, not its descendants, so whether `descendantPid` actually stops depends entirely on what the delegate spawner does underneath. Has the fixture test been run on a Windows machine? If it has, saying so in the PR body would help, since CI cannot show it.

Smaller thing: `shutdown` takes `childScope` and never uses it.

```ts
const shutdown = Effect.fn("BoundedChildProcessSpawner.shutdown")(function* (
  handle: ChildProcessSpawner.ChildProcessHandle,
  childScope: Scope.Closeable,
  isReferenced: () => boolean,
) {
  if (isReferenced()) {
    yield* kill(handle);
  }
});
```

The scope is closed by the `exitCode` fork instead, so the parameter looks like a leftover from an earlier shape. Worth dropping so it does not read as a missed cleanup path.

Also worth a note: `sendSignal` forks the kill detached and `Effect.ignore`s the result, so a `kill` that fails outright is indistinguishable from one that is merely slow. Both land on the same `logWarning` after the full grace window. Given the whole point is bounding shutdown, a failed signal is probably worth logging at the moment it fails rather than only after the timeout.

Defmon3:
Thank you, updated the PR

Defmon3:
Closing because we are no longer able to reproduce the shutdown hang against current upstream's process handler.

We ran 10 isolated Windows checks on Node 24.13.1 using Effect 4.0.0-rc.115, the version used by upstream main at 9ea9c3d5d. Scope closure, forceKillAfter, and explicit kill all completed without a hang or timeout, including cases with parent and descendant processes.

These checks exercised the process handler, not a full server or provider shutdown, so this does not prove every shutdown path is unaffected. We can revisit this if a concrete reproduction becomes available.

PR #6324, opened 2026-08-12
fix(mobile): prevent OTA update restart crashes

juliusmarminge:
## Problem

Applying an OTA update called `Updates.reloadAsync()` mid-session, seconds after the Home screen mounted — tearing down the JS runtime while native surfaces (terminal, composer editor, review diff), live WebSocket connections, and first-render hydration were still active. That teardown window is where the post-OTA crashes came from.

## Fix

- After an update downloads, a native alert offers **Install Now** or **Later**. "Later" defers the restart to the next backgrounding, where nothing is rendering and the restart is invisible.
- A deferred restart re-verifies it is still safe right before firing: the app must actually be backgrounded (flush latency or an iOS suspend can push the continuation into the next foreground session) and not merely paused behind an app-initiated handoff — on Android the image picker covers the activity and reports as backgrounded. Otherwise it re-arms for the next backgrounding.
- Rollback directives (`eas update:rollback`) skip the prompt and apply immediately — they exist to pull a broken bundle.
- Before any restart, debounced composer-draft writes and in-flight outbox writes are flushed, and both stores now write atomically (temp file + rename) so an interrupted write can never leave truncated JSON.
- New foreground re-check after 15+ minutes backgrounded, so long-resident apps pick up updates published while in memory. A failed deferred restart returns the prompt to later checks instead of suppressing them for the session.
- The hidden Settings 5-tap check keeps its immediate restart (`applyMode: "immediate"`).

Covered by 22 unit tests over the check/prompt/defer/rollback/flush state machine plus the handoff guard; drafts and outbox suites pass unchanged. Docs: new "The Mobile App" section in `docs/user/updating.md`.

Built by Claude Fable 5 via Claude Code.

> [!NOTE]
> ### Fix OTA update crashes by coordinating restarts with state flush and app lifecycle
> - Adds a deferred install flow: downloaded updates are installed automatically when the app backgrounds, or the user is prompted after an extended foreground session.
> - Introduces `flushComposerDrafts` and `flushThreadOutbox` so in-flight writes are drained before restarting; automatic installs abort and re-arm on flush failure to avoid data loss.
> - Adds `isSafeToRestartInBackground` gating (checks `AppState` and foreground handoffs) so restarts never fire during image picker or other OS-covered surfaces.
> - Adds a `ready` state to `AppUpdateCheckState` and surfaces an "Update ready" label in [SettingsRouteScreen](https://github.com/pingdotgg/t3code/pull/6324/files#diff-2f7a2e66b652454e7fd1b30226a5146d595dfe05945b34d90ffd5f3a5037cb1e) when a background-downloaded update is waiting.
> - Adds `startAppUpdateForegroundRecheck` in [HomeRouteScreen](https://github.com/pingdotgg/t3code/pull/6324/files#diff-c2d4821e3a79b2c54784238ebdee5378f310422e4a79169b0bc6fc95081dacf5) to trigger periodic update checks after meaningful background gaps, suppressed when an install is already pending.
> - Risk: restarts now flush composer drafts and the thread outbox; a flush error silently re-arms the deferred install rather than proceeding, which may delay updates indefinitely if persistence is broken.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 63fe6d1.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Changes when and how the JS runtime tears down (reload, persistence flush, AppState), which can affect crashes and data loss if edge cases are missed; behavior is heavily tested but touches critical session state.
> 
> **Overview**
> **Mobile OTA updates no longer call `reloadAsync()` right after download.** Default behavior downloads in the background, moves to a **ready** state, and installs on the next real backgrounding after flushing composer drafts and the thread outbox. A long foreground session (30 minutes) can trigger an **Install now / Later** alert; declining keeps the deferred install armed.
> 
> **Restart safety:** Injectable `AppUpdateEnvironment` coordinates flush, background-only deferred installs (with re-arm when unsafe or flush fails), a single `installInProgress` lock for racing paths, and **immediate** `applyMode` for the Settings hidden version check. Rollbacks still restart without deferral; failed automatic flushes block silent restarts but user-requested installs may proceed. **Foreground recheck** after 15+ minutes backgrounded picks up updates for long-lived sessions.
> 
> **Persistence:** `writeFileAtomically` for drafts and outbox; `flushComposerDrafts` / `flushThreadOutbox` run before reload. Image picker uses `beginForegroundHandoff` so Android “background” during picker does not trigger an update restart. Settings shows **Update ready** when a manual check joins an in-flight background download. User docs add a **The Mobile App** OTA section.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63fe6d1eb495c4f54a77ac14f3b6ee2e24b7f618. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/user/updating.md
+++ b/docs/user/updating.md
@@ -70,4 +70,11 @@ If a step fails:
 3. For a command-line server, relaunch it with `npx t3@<client-version>`, replacing
    `<client-version>` with the client version shown in the warning.
 
+## The Mobile App
+
+The mobile app keeps itself current on its own. When it finds a new version, it downloads it in the
+background and then asks whether to install it right away. Choosing **Later** is safe: the update
+installs automatically the next time you leave the app, and unsent drafts and queued messages are
+saved before the restart either way.
+
 For remote connection setup and access troubleshooting, see [Remote Access](./remote-access.md).
--- a/docs/user/updating.md
+++ b/docs/user/updating.md
@@ -73,8 +73,9 @@ If a step fails:
 ## The Mobile App
 
 The mobile app keeps itself current on its own. When it finds a new version, it downloads it in the
-background and then asks whether to install it right away. Choosing **Later** is safe: the update
-installs automatically the next time you leave the app, and unsent drafts and queued messages are
-saved before the restart either way.
+background and installs it automatically the next time you leave the app. Unsent drafts and queued
+messages are saved before the restart. Only if the app stays open long enough that the update never
+gets that chance does it ask whether to install right away; choosing **Later** is safe and keeps the
+automatic install armed.
 
 For remote connection setup and access troubleshooting, see [Remote Access](./remote-access.md).
--- a/apps/mobile/src/features/home/HomeRouteScreen.tsx
+++ b/apps/mobile/src/features/home/HomeRouteScreen.tsx
@@ -11,7 +11,7 @@ import { useSavedRemoteConnections } from "../../state/use-remote-environment-re
 import { useAdaptiveWorkspaceLayout } from "../layout/AdaptiveWorkspaceLayout";
 import { WorkspaceEmptyDetail } from "../layout/WorkspaceEmptyDetail";
 import { WorkspaceSidebarToolbar } from "../layout/workspace-sidebar-toolbar";
-import { checkForAppUpdateOnLaunch } from "../updates/app-updates";
+import { checkForAppUpdateOnLaunch, startAppUpdateForegroundRecheck } from "../updates/app-updates";
 import { AndroidHomeFabLayout } from "./AndroidHomeFab";
 import { HomeScreen } from "./HomeScreen";
 import { HomeHeader } from "./HomeHeader";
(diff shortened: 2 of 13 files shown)


docs/user/updating.md
```diff
@@ -70,4 +70,11 @@ If a step fails:
 3. For a command-line server, relaunch it with `npx t3@<client-version>`, replacing
    `<client-version>` with the client version shown in the warning.
 
+## The Mobile App
+
+The mobile app keeps itself current on its own. When it finds a new version, it downloads it in the
+background and then asks whether to install it right away. Choosing **Later** is safe: the update
+installs automatically the next time you leave the app, and unsent drafts and queued messages are
+saved before the restart either way.
```
juliusmarminge: we should prefer auto-updating silently. download in background, and update next launch should be the default I tihnk. only prompt them if we've had one downloaded and they haven't backgrounded the app for a bit

PR #6356, opened 2026-08-12
fix(server): pick up newly installed provider CLIs on Windows refresh

### LINKED ISSUE #6352: Windows: newly installed provider CLI stays missing until app restart
On Windows, if T3 starts without a provider CLI on PATH and you install that CLI while the app is already open, Refresh providers still shows Not found. Restarting the app finds it.

### Repro
1. Uninstall Codex CLI (or start T3 when it is not installed).
2. Launch T3. Settings → Providers → Codex is red / Not found.
3. Install Codex: `powershell -ExecutionPolicy ByPass -c "irm https://chatgpt.com/codex/install.ps1 | iex"`
4. Click Refresh provider status. Still Not found.
5. Quit and reopen T3. Codex is found / authenticated.

### Expected
Refresh after install should pick up the new CLI without restarting.

### Notes
- Same pattern should apply to other CLIs that land on User PATH or under `%LOCALAPPDATA%\Programs\...` after launch.
- Uninstall → Refresh → red works fine; the broken case is install mid-session after a cold start without the CLI.
- Related: #4846 (User PATH ignored for Claude). That was closed after adding `~\.local\bin`, but this is the mid-session PATH / refresh case.

### Environment
- Windows 10/11
- Codex CLI via official `install.ps1`
- Reproduced on current release build; fixed locally by re-reading User/Machine PATH (and refreshing PATH) before provider re-probe

fraineralex:
Fixes #6352

On Windows, if T3 starts without a provider CLI and you install it mid-session, Refresh still shows Not found until restart. The process PATH was frozen at launch, and the Windows env probe read process PATH instead of User/Machine.

- Call `fixPath()` before provider refresh
- Read Machine + User PATH in the Windows shell probe
- Include `%LOCALAPPDATA%\Programs\OpenAI\Codex\bin` in known Windows CLI dirs

### Demo
Cold start without Codex → red → install → Refresh → green (no restart).

https://github.com/user-attachments/assets/1d83a854-40a9-45ab-95a3-3bc3c07c307b

Model: Grok 4.5

### DIFF
--- a/apps/desktop/src/shell/DesktopShellEnvironment.test.ts
+++ b/apps/desktop/src/shell/DesktopShellEnvironment.test.ts
@@ -229,6 +229,7 @@ describe("DesktopShellEnvironment", () => {
           "C:\\Windows\\System32",
           "C:\\Users\\testuser\\AppData\\Roaming\\npm",
           "C:\\Users\\testuser\\AppData\\Local\\Programs\\nodejs",
+          "C:\\Users\\testuser\\AppData\\Local\\Programs\\OpenAI\\Codex\\bin",
           "C:\\Users\\testuser\\AppData\\Local\\Volta\\bin",
           "C:\\Users\\testuser\\AppData\\Local\\pnpm",
           "C:\\Users\\testuser\\.local\\bin",
--- a/apps/desktop/src/shell/DesktopShellEnvironment.ts
+++ b/apps/desktop/src/shell/DesktopShellEnvironment.ts
@@ -201,7 +201,12 @@ const knownWindowsCliDirs = (env: NodeJS.ProcessEnv): ReadonlyArray<string> => [
   ...trimNonEmpty(env.LOCALAPPDATA).pipe(
     Option.match({
       onNone: () => [],
-      onSome: (value) => [`${value}\\Programs\\nodejs`, `${value}\\Volta\\bin`, `${value}\\pnpm`],
+      onSome: (value) => [
+        `${value}\\Programs\\nodejs`,
+        `${value}\\Programs\\OpenAI\\Codex\\bin`,
+        `${value}\\Volta\\bin`,
+        `${value}\\pnpm`,
+      ],
     }),
   ),
   ...trimNonEmpty(env.USERPROFILE).pipe(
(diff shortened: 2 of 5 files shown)


CDVolvik (commented):
Windows box check on `d00be9516`.

`packages/shared/src/shell.test.ts` + `DesktopShellEnvironment.test.ts`: 38/38 on the PR. Checking out `origin/main` over those two source files (keeping the PR tests) fails 5/38 — the new Machine/User command assertion and the Codex bin path. Those tests pin what they claim.

The stated bug is real. `[Environment]::GetEnvironmentVariable('PATH')` with no target is Process scope, so a User-PATH write after T3 started is invisible. Reading Machine+User from the registry is the right no-profile fix. On this machine that registry read is a superset of process PATH except five dirs (`nodejs`, `Volta`, `pnpm`, `.bun`, `scoop`) — exactly `knownWindowsCliDirs`, which you still prepend.

The problem is the same command is used for `loadProfile: true`. After this change the profile probe and the no-profile probe emit the same PATH source. `$env:PATH` mutations from a profile (fnm prepending `FNM_MULTISHELL_PATH`) cannot be observed. `FNM_DIR` / `FNM_MULTISHELL_PATH` are still captured from process scope, but PATH no longer includes the shim the profile prepends. Desktop still merges `profile.PATH` first, but it is the same string as `noProfile.PATH`.

The new tests mock `execFile`; they do not pin that the two probes differ. A mock that returns `C:\Profile\Node` still passes regardless of the command text.

Suggested split: keep Machine+User for `-NoProfile`, and keep process `$env:PATH` for `loadProfile: true` (the profile has already run). `Programs\OpenAI\Codex\bin` in `knownWindowsCliDirs` is fine.

`fixPath()` on every `refresh`/`refreshInstance` is already win32-gated inside, but on Linux/mac it now re-hydrates login-shell PATH on every provider refresh. No `ProviderRegistry` test covers that.

CDVolvik (changes requested):
Formalizing my earlier review as changes-requested — head is unchanged (d00be95) and the core issue stands:

Machine+User registry PATH is the right fix for the `loadProfile: false` probe, but the same command string is used for the `loadProfile: true` path — so profile-driven prepends (fnm's `FNM_MULTISHELL_PATH` shims, any `$env:Path` mutation in the profile) vanish from the one probe whose job is to capture them. That regresses exactly the users the profile probe exists for.

Scoping the registry-read command to the no-profile probe only (or taking the #4896 shape, which joins Process+Machine+User) resolves it. Detail in my earlier review on this commit.

t3dotgg:
> [!NOTE]
>

Closing in favor of [#8465](https://github.com/pingdotgg/t3code/pull/8465). It addresses Windows PATH refresh while preserving profile-added paths and updating provider environments used for later sessions. Review continues there. This branch's registry PATH change also replaces the profile probe's PATH, which loses entries that profiles add. The Codex install-directory fallback is recorded on the retained PR.

Issue #5734, opened 2026-08-08
[Bug]: branch name helper session starts to work on the main task


menzels:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

started app 
put prompt "look at gh issue 109 and fix it"
claude opus 5 high

### Expected behavior

the branch name generator should not start to work on the issue.

## What happened

t3code spawns a small Haiku 4.5 helper session at thread creation to produce a branch name. and its only prompt was:

> You generate concise git branch names. Return a JSON object with key: branch. Rules: … User message: **look at gh issue 109 and fix it**

Instead of just answering with JSON, it treated the embedded user message as its task — and it had the full tool set plus full-access permissions in *my* worktree. Its timeline (UTC):

| 20:14:46 | `gh issue view 109` |
| 20:14:52–58 | reads `calculate-afa.ts` and the spec |
| 20:15:04–26 | 5 `Edit` calls — the `ancillaryAdjustment` change and the spec rewrite (this is what broke my `Edit`s out from under me) |
| 20:15:27–20:16:18 | `devenv up &`, then `devenv shell pnpm … test` — full backend suite, "All 330 tests pass" |
| 20:16:20 | `git checkout .devcontainer/devcontainer.json` — that's what reverted the symlink noise |
| 20:16:31 | `git add … && git commit` → `bc2aceb` |
| 20:16:35 | finally emits `StructuredOutput {"branch": "fix afa ancillary costs conversion"}` |
| 20:16:36 | t3code's `GitVcsDriver.renameBranch` uses that name |

The commit at 20:16:31 landed 15 seconds *after* my first turn ended, which is why it looked like it came from nowhere.

## It's systematic, not a one-off

Every thread on this issue got the same treatment — the branch-namer did the work in each worktree:

- `t3code-a952502c` (mine): `b3291028`, 8 edits, committed `bc2aceb`
- `t3code-ab2636ef`: `41f5229d`, 11 edits, committed `7c4940b` at 22:13
- `t3code-4e6baf23`: `da90157e`, 8 edits, committed

[...]

Issue #5359, opened 2026-08-04
[Bug]: Thread title and branch-name generation fail silently when the text generation provider is unhealthy


gustavohariel:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

### Steps to reproduce

1. Have both Codex and Claude enabled as providers. Leave **Settings → Text generation model** untouched, i.e. on its default (`instanceId: "codex"`, `gpt-5.6-luna`).
2. Make the Codex CLI unusable **without disabling the provider** — e.g. corrupt `~/.codex/config.toml` so the app-server refuses to load it. In my case a concurrent write left a duplicate table + an orphaned key, and the provider probe recorded:
   ```
   "status": "error",
   "message": "Codex app-server provider probe failed: failed to reload config: ~/.codex/config.toml:7:1: duplicate key."
   ```
3. Start a new thread with any first message.
4. Observe the thread title and, for worktree threads, the generated branch name.
5. Right-click the thread → **Regenerate title**.

### Expected behavior

Either:

- text generation falls back to a provider that is actually usable (Claude was `status: "ready"` and authenticated the whole time), **or**
- the failure is surfaced the way `generateCommitMessage` surfaces it (`Text generation failed in generateCommitMessage: …`, see #5182) — a toast or a thread-activity entry.

### Actual behavior

Nothing happens, and nothing tells you why.

- Thread titles permanently stay as the raw `titleSeed` (the first message text), so the feature looks like it is simply not implemented.
- Worktree branch names permanently stay as the `t3code-<hash>` fallback — all 20 of my worktrees are `t3code-09b397af`, `t3code-0a41d30a`, …
- **Regenerate title** clears the pending state and writes no title. In my event store: a `client` `thread.meta-updated` with `"regenerateTitle": true` at `12:24:03.721Z`, then a `server` `thread.meta-updated` with `"titleRegeneration": null` at `12:24:08.553Z` and no title. No error anywhere in the UI.
- Meanwhile commit-message generation through the *same* broken provider does show an error, so the app is inconsistent about the same underlying failure.

Two things combine to make this silent:

[...]

Issue #3044, opened 2026-06-11
[Feature]: Name the worktree (with AI) before creating it


tyteen4a03:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I am describing a concrete problem or use case, not just a vague idea.

### Area

apps/web

### Problem or use case

Right now, T3 code defaults all worktrees to have the format `t3code/<sha>`. While this is fine within T3 code, I use an external Git client to manage worktrees and have difficulty linking threads to their worktrees.

### Proposed solution

Provide an option to name the worktree manually or with AI, before creating it.

### Why this matters

This will make worktrees more identifiable for other clients.

### Smallest useful scope

Provide an option to name the worktree manually, before creating it.

### Contribution

- [ ] I would be open to helping implement this.

msaspence:
And/or the ability to rename it via a tool after it has been created

Issue #967, opened 2026-03-12
Modifiable worktree prefix.


sbansal1999:
We should add a functionality on the app level allowing the user to enter a worktree prefix instead of having it always as `t3code`.

danielo515:
Same for branch prefix...

juliusmarminge:
Duplicate of #272

Same ask — make the hardcoded `t3code/` worktree branch prefix configurable. Consolidating into #272 (maintainer-acknowledged, with a PR linked).

PR #6428, opened 2026-08-13
Feat/custom worktree branch name

### LINKED ISSUE #6421: [Feature]: Fully custom branch name when starting a new worktree thread
### Before submitting

### Area

apps/web

### Problem or use case

New worktree threads get an auto-generated branch, `t3code/<hex>`, and a directory derived from it, `t3code-<hex>`. Neither tells me anything.

There is an LLM step that renames the branch after the first message, but it is best-effort and depends on a working text-generation CLI. When that CLI is missing or unhealthy it fails silently and every worktree keeps its hex name. #5359 reports 20 worktrees all named `t3code-<hex>`.
A scenario I ran into is I only have Claude, but the branch name LLM was configured to use Codex so would never work.

Outside T3 it can get a bit confusing as well. In an external Git client, a window switcher, or a list of open editor windows, every worktree looks identical and I can't tell which thread a given worktree belongs to without opening it.

Often the case, I already know what the branch should be called before I send the first message (picking up a ticket in Jira or Linear for example), and some workplaces put restrictions on branch names which makes using T3 Code a little cumbersome

### Proposed solution

An optional text field next to the worktree controls in the composer, to name the branch before the thread is created.

- Leave it empty → today's behaviour
- Type a name → it is used verbatim as the branch, and the worktree directory derives from it as it already does. Typing `feat/login-fix` gives branch `feat/login-fix` and directory `feat-login-fix`.
- Sanitize as you type, rather than rejecting on submit: strip quotes, replace characters Git refuses in a ref, collapse repeated separators, block leading dots per path segment, cap the length.
- Preserve case and dots. A hand-typed name should be used as typed — `Feature.V2` should survive.
    - Im aware that these are stripped currently, but my thinking here is if you care about the full name of the branch, you may also care about weather capital letters should be in there or not
- A typed name is used and wont trigger be overwritten by the LLM rename

### Why this matters

It makes worktrees identifiable everywhere outside T3, which is where the confusion actually happens.

It gives a deterministic path to a readable branch that needs no text-generation provider at all — useful for anyone hitting #5359, and for anyone who would simply rather not spend an LLM call on naming.

It also side-steps #5734: when a name is typed, the branch-naming helper session is never spawned for that thread. That is not a fix for the underlying bug, but it is a usable workaround today.

Right now the workaround is to open the terminal drawer and `git checkout -b` by hand, which several people in #272 describe doing.

### Smallest useful scope

One optional text input on the new-worktree composer, applied at creation time

- No persistence, no per-project default, no template
- No renaming after creation
- No change to the generated name when the field is left empty

### Alternatives considered

- Opening the terminal drawer and running `git checkout -b` manually. Works, but it is a detour for something the composer could ask once.
- Configurable branch prefixes (PRs #3954 and #6037). Orthogonal, not a substitute: those set an automatic default, this names one specific branch
- PR #1110 implemented a "Custom Branch Name" override and was closed for drift, not on merit, with an explicit invitation to resubmit against current main.

### Risks or tradeoffs

- Sanitization has to match what Git actually refuses in a ref, and be forgiving enough to type into — collapsing separators live rather than erroring on every keystroke.
- Name collisions with an existing branch need handling at the input, not as a failure after the thread is created.
- This does not satisfy #272 and the prefix cluster. Those want a persistent automatic rule for org branch policies; retyping a required prefix on every thread is not an answer to that. It should be framed as complementary, not as closing them.

### Examples or references

Related issues: #3044 (this request, including the "before creating it" timing), #5359 (silent rename failure leaves every worktree hex-named), #5734 (the branch-name helper session going off task), #967 (comments asking for legible worktree directory names).

Adjacent: #272, #5275 and #5965 want configurable naming defaults — related but a different shape.

Near-freebie: #6163 asks for whitespace to be replaced with dashes when creating a branch. The same sanitizer would close it if it were also wired to the branch selector's create-branch flow.

Prior art: PR #1110 (closed, "Custom Branch Name" override), PRs #3954 and #6037 (open, prefix settings).

This is what ive been running in a fork (placeholder vs valid input)
Image
Image

Im not _sold_ on this UI, but it works well for my use cases

### Contribution

MrSimmmons:
## What Changed

Allows the setting of a custom branch name when creating a new worktree from a new thread

## Why

Fixes #6421, also adds a workflow / fixes a potential bug for #3044, #5359 #5734 and #967

Was initially developed on top of https://github.com/pingdotgg/t3code/pull/6427 to round out some worktree features I found missing when comparing to other apps like VS Code, but its a pretty good stand alone feature addition so this gets its own PR

## UI Changes

Before:
Screenshot 2026-08-13 at 4 03 25 PM

After:
Screenshot 2026-08-13 at 2 33 01 PM
Screenshot 2026-08-13 at 2 33 26 PM

Happy to make changes to the UI if the idea seems sound :D

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches first-message worktree creation and send gating; incorrect collision logic or stale async checks could block sends or allow invalid branch names, though server-side git would still reject duplicates.
> 
> **Overview**
> Adds an optional **custom branch name** field in the composer context strip when starting a **new thread** in **New worktree** mode (before the first send, no existing worktree). The name is sanitized as you type, persisted on the draft (`worktreeBranchName`), and used in `prepareWorktree.branch` instead of the random `t3code/…` temporary name when it passes validation.
> 
> **Conflict checks** call `listRefs` with a new **`queryMode: "collision"`** on the Git VCS driver, which returns local refs that would collide with the typed name (exact match, parent, or child under Git’s ref rules—not substring search). The input shows inline errors and blocks send while the check is in flight or if a collision is found; `ChatView` mirrors those gates with thread errors.
> 
> Supporting changes: **`sanitizeWorktreeBranchNameInput`** / **`normalizeWorktreeBranchName`** in shared git helpers (with tests), **`queryMode`** on `VcsListRefsInput`, and minor toolbar layout tweaks (`shrink` / `flex-auto`) so the new control fits beside existing selectors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64bb942fa7d29d06aa25609fd28b7788799f1d76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add custom worktree branch name input with collision validation
> - Users can now type and validate a custom branch name when creating a worktree from a draft thread, via the new [BranchToolbarWorktreeNameInput](https://github.com/pingdotgg/t3code/pull/6428/files#diff-7d2cab4bf1f5f3c2730ec81571feb2a631a04c012c6936af577c9d33da39a870) component rendered inside [BranchToolbar](https://github.com/pingdotgg/t3code/pull/6428/files#diff-698af5019ef8a26d8e1653085b9e61762e35f389e79e62d402760255625b0a10)
> - Input is sanitized per-keystroke with `sanitizeWorktreeBranchNameInput` and normalized with `normalizeWorktreeBranchName` to comply with Git ref rules
> - Adds a `collision` query mode to `listRefs` ([VcsListRefsInput](https://github.com/pingdotgg/t3code/pull/6428/files#diff-2282301d92a7a9ddc7e8d88b986c9628fdf70497dc6a09fe80eb8d46d6a769e7)) backed by the new `refCollidesWithName` helper in [GitVcsDriverCore](https://github.com/pingdotgg/t3code/pull/6428/files#diff-1c3e2b9763c6526006485633fb677c83dc54cc1ea5563a1e51aab707c35854d8), returning only refs that would collide with the given name
> - [ChatView](https://github.com/pingdotgg/t3code/pull/6428/files#diff-4b49e092ccd43be0f0de24abe85ba522e09f04288a5d84253b0263e1a389400e) defers sends while validation is in-flight, blocks sends on collisions, and skips LLM-based branch naming when a custom name is used
> - The typed name persists in the draft session via a new `worktreeBranchName` field in [composerDraftStore](https://github.com/pingdotgg/t3code/pull/6428/files#diff-fe25f0ac6b3f15bcb25522be5eddf9c2db913c035393f9a4dc019a9afda80f8a)
> - Risk: `filterBranchesForListQuery` now accepts an optional `queryMode` parameter; callers that omit it retain the existing substring behavior, but any out-of-tree consumers of `VcsListRefsInput` should be aware of the new optional field
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 64bb942. 5 files reviewed, 3 issues evaluated, 3 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> <details>
> <summary>apps/web/src/components/BranchToolbarWorktreeNameInput.tsx — 0 comments posted, 2 evaluated, 2 filtered</summary>
>
> - [line 42](https://github.com/pingdotgg/t3code/blob/64bb942fa7d29d06aa25609fd28b7788799f1d76/apps/web/src/components/BranchToolbarWorktreeNameInput.tsx#L42): `normalizeWorktreeBranchName` leaves the special branch name `HEAD` unchanged, so the collision query can report it as available and the send path uses it for `git worktree add -b`. Git rejects `HEAD` as a branch name, meaning typing `HEAD` passes the new validation but causes worktree creation to fail. <b>[ Already posted ]</b>
> - [line 67](https://github.com/pingdotgg/t3code/blob/64bb942fa7d29d06aa25609fd28b7788799f1d76/apps/web/src/components/BranchToolbarWorktreeNameInput.tsx#L67): A failed `listRefs` lookup is treated as `checked` and therefore produces the `available` state at lines 65-78. If the lookup failed while a local branch with the requested name already exists, the send path accepts that status and attempts `git worktree add -b`; Git then rejects the colliding ref, so the UI shows the name as available but the first send fails instead of asking the user to retry validation or choose another name. <b>[ Already posted ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/ChatView.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 5622](https://github.com/pingdotgg/t3code/blob/64bb942fa7d29d06aa25609fd28b7788799f1d76/apps/web/src/components/ChatView.tsx#L5622): The send gate only validates a custom name when `worktreeBranchNameStatus` is truthy. Immediately after entering a name (before the input's effect publishes its `checking` status), `draftWorktreeBranchName` is non-null but the status is still `null`, so the send proceeds and line 5879 falls back to a generated branch. Thus a user can submit a first worktree message quickly and silently lose the requested custom branch name instead of waiting for validation. <b>[ Already posted ]</b>
> </details>
>
>
> </details>
>


t3dotgg:
> [!NOTE]

This adds custom worktree branch names across server, contracts, web, and mobile. We are not adding this new worktree naming path through the current backlog.

PR #6505, opened 2026-08-13
fix(web): preserve inline attachment image bounds

myNameArnav:
## Problem
Inline user-image previews used `object-cover`, which cropped tall screenshots and cut into their rounded corners.

## Fix
Use `object-contain` so the preview preserves the uploaded image bounds while the existing rounded, overflow-hidden wrapper continues to clip the corners cleanly.

## Verification
- `vp run --filter @t3tools/web test`
- `vp lint apps/web/src/components/chat/MessagesTimeline.tsx`
- Browser visual comparison with the supplied screenshot completed

Model: GPT-5.6-Luna | Harness: OpenCode in T3 Code


myNameArnav:
Closed after fresh browser reproduction showed the proposed object-fit change produced identical before/after rendering; no effective fix was shipped.

PR #6596, opened 2026-08-14
fix(server): usage counts every configured provider profile

bwmp:
## What Changed

`UsageService` resolved exactly two transcript directories, from the legacy `providers.claudeAgent` and `providers.codex` settings. It now resolves one directory per configured provider profile: the legacy blob plus every `providerInstances` entry on the `claudeAgent` / `codex` driver, each decoded through that driver's own settings schema.

The resolution moved to `apps/server/src/usage/usageTranscriptDirs.ts` so it can be tested without standing up the service.

Three rules keep the totals honest:

- Precedence mirrors `deriveProviderInstanceConfigMap` — the legacy blob is a mirror of the driver's default instance slot, so it is dropped once an explicit entry claims that id. Without this a migrated config would be scanned under two homes.
- Results are de-duplicated by resolved path. Instances legitimately share a home (Codex shadow homes share their `CODEX_HOME` by design), and scanning one twice would double count every token in it.
- The Claude transcript layout follows the home's provenance rather than a filesystem probe. `resolveClaudeHome` now reports whether the path was configured or defaulted to the OS home: a configured home *is* the config dir, so transcripts sit directly below it, and only the defaulted home nests them under `.claude`. Probing `<home>/.claude/projects` cannot tell those apart, so a profile configured at the OS home scanned the default profile's transcripts instead of its own.

`UsageSource` is already fingerprinted by resolved home path, so several sources per provider need no contract or client change. No UI change.

## Why

Anyone running more than one account of a provider configures a second `providerInstances` entry with its own `CLAUDE_CONFIG_DIR`, and that home keeps its own transcripts. The scan never looked at those instances, so the Usage page silently reported the default account's usage and omitted the rest — with no indication that coverage was partial.

On a three-Claude-account machine the resolver went from one Claude directory to three:

```
claude  ~/.claude-work/projects        22 transcripts
claude  ~/.claude-personal/projects    88 transcripts
claude  ~/.claude/projects            251 transcripts
codex   ~/.codex/sessions             216 transcripts
```

## Checklist

- [x] I included a video for animation/interaction changes (none)

---

Verified with `vp test run src/usage src/provider/Drivers/ClaudeHome.test.ts` (50 passing, 8 new in `usageTranscriptDirs.test.ts` covering multi-home scanning, both Claude layouts, default-id precedence, shared-directory de-duplication, undecodable instance configs, and other drivers), plus `tsgo --noEmit`, lint, and format on the changed scope.

---

> [!NOTE]
> **Medium Risk**
> Changes how usage paths are resolved (multi-home, Claude layout semantics), which can shift reported totals for multi-account or edge-case Claude configs; logic is covered by new unit tests.
> 
> **Overview**
> **Usage scanning** no longer reads only the legacy `providers.claudeAgent` / `providers.codex` homes. It walks **one transcript directory per configured profile**—legacy settings plus every `providerInstances` entry for those drivers—via new `resolveUsageTranscriptDirs`.
> 
> **Claude layout** stops using a filesystem probe (`~/.claude/projects` vs `projects`). `resolveClaudeHome` now exposes whether the home was **explicitly configured**; configured homes use `projects` directly under the config dir, while the default OS home still uses `.claude/projects`. That fixes profiles pointed at `~` from inheriting the default install’s transcripts.
> 
> **Correct totals** follow registry-style rules: drop the legacy blob when an instance owns the driver’s default id, **de-duplicate** shared paths (e.g. Codex shadow homes), skip undecodable instance configs, and ignore other drivers. `UsageService` delegates to the new module; user docs note multi-instance coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e983d52e1caaa649e39c012e8dd062fe64be60f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix usage counting to include every configured provider profile
> - Introduces `resolveUsageTranscriptDirs` in [usageTranscriptDirs.ts](https://github.com/pingdotgg/t3code/pull/6596/files#diff-8c486578dd6dc056a75c8e3f29f9741e5a72cdd3c0986264a16427f36ca7026f) to enumerate transcript directories for all configured provider instances, including multiple instances of the same provider with separate home directories.
> - Adds `resolveClaudeHome` in [ClaudeHome.ts](https://github.com/pingdotgg/t3code/pull/6596/files#diff-73ba6e23eba4250edbecbe8239d07a8f0556ddef26400793eb1b7432199ea29d) that returns both the resolved path and whether it was explicitly configured, replacing a filesystem probe for Claude layout selection with config-based provenance.
> - `UsageService.resolveTranscriptDirs` now delegates to `resolveUsageTranscriptDirs`, de-duplicating results by provider+path to avoid double-counting shared directories (e.g. Codex shadow homes).
> - Behavioral Change: Claude transcript layout (`.claude/projects` vs `projects` directly under home) is now determined by whether the home was explicitly configured rather than by probing for directory existence.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e983d52.</sup>
> 
>


t3dotgg:
> [!NOTE]

[#5806](https://github.com/pingdotgg/t3code/pull/5806) remains open as the active path for scanning every configured provider home. The explicit Claude home provenance case and the disabled-profile history case in this PR remain useful as references while the common scanner is reviewed.

PR #6611, opened 2026-08-14
fix(web): move settings back action to the top

### LINKED ISSUE #4303: [Feature]: Move Settings Back control to the top
### Before submitting

### Area

apps/web

### Problem or use case

When I open Settings from the sidebar, the section list is at the top and Back sits in the footer. After spending time in Settings, leaving with the button feels awkward. My attention is on the top/content, and I have to hunt down to the footer to quit.

Escape works for me, but that does not help if you use the on-screen Back control, or if you do not know the shortcut.

Version: `0.0.29-nightly.20260722.878 (9a0a07167f06)`

### Proposed solution

Put Back at the top of the Settings sidebar, with the section list or chrome.

### Why this matters

Leaving Settings should be obvious without a shortcut. The button should sit where people already look.

### Smallest useful scope

Move the existing Back control to the top.

### Alternatives considered

Using Escape. Fine as a shortcut, not a substitute for clear button placement.

### Risks or tradeoffs

_No response_

### Examples or references

_No response_

### Contribution

hcaiano:
## What Changed

Moved the existing Back action from the Settings footer to the top of the Settings sidebar, before search and section navigation.

The action keeps the same behavior. Secondary account actions remain in the footer, whose structure is unchanged.

Closes #4303.

## Why

Back is one of the main actions in Settings. After changing a setting, users often want to return to what they were doing.

Keeping Back in the footer makes it harder to find and requires moving the pointer away from the Settings navigation every time. Placing it at the top keeps navigation actions together and makes the exit visible without relying on the Escape shortcut.

## UI Changes

| Before | After |
| --- | --- |
| Back action in the Settings footer | Back action at the top of the Settings sidebar |

## Validation

- Confirmed that Back returns to the previous page.
- Confirmed that the fallback navigation still returns to the main page.
- Confirmed that Escape navigation is unchanged.
- Confirmed that secondary footer actions remain in place.
- `pnpm --filter @t3tools/web exec tsgo --noEmit`
- `vp lint apps/web/src/components/settings/SettingsSidebarNav.tsx`
- `vp fmt --check apps/web/src/components/settings/SettingsSidebarNav.tsx`
- `git diff --check`

## Checklist

- [x] I included before/after screenshots
- [x] No video is needed because this change has no motion or timing changes

Created with GPT-5.6 Sol via the Codex harness.

### DIFF
--- a/apps/web/src/components/settings/SettingsSidebarNav.tsx
+++ b/apps/web/src/components/settings/SettingsSidebarNav.tsx
@@ -191,6 +191,14 @@ export function SettingsSidebarNav({ pathname }: { pathname: string }) {
     <>
       <SidebarContent className="overflow-x-hidden">
         <SidebarGroup className="gap-2 p-[var(--sidebar-content-inset)]">
+          <SidebarMenu>
+            <SidebarMenuItem>
+              <SidebarMenuButton onClick={handleBackClick}>
+                <ArrowLeftIcon />
+                <span>Back</span>
+              </SidebarMenuButton>
+            </SidebarMenuItem>
+          </SidebarMenu>
           <div className="flex h-8 items-center gap-2 rounded-md px-2 py-1.5 text-sm font-medium text-sidebar-muted-foreground hover:bg-sidebar-row-hover hover:text-sidebar-foreground">
             <SearchIcon className="size-4 shrink-0 text-sidebar-muted-foreground/80" />
             <Input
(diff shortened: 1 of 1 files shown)


t3dotgg:
> [!NOTE]

Moving the Back action above search is a small layout preference, and the current navigation still exposes the action. We do not want to keep a separate implementation for this placement change.

PR #6608, opened 2026-08-14
perf: make streaming projection and activity appends incremental

x1xhlol:
## What Changed

**Server (`ProjectionPipeline`):**
- `thread.message-sent` no longer triggers a full `refreshThreadShellSummary`; it updates the projection row directly, bumping `updatedAt` and advancing `latestUserMessageAt` only for newer user messages.
- `thread.activity-appended` also updates the row directly and only refreshes the shell summary for lifecycle activity kinds that can change it (`user-input.requested/resolved`, `approval.requested/resolved`, and their provider failure events). Ordinary streaming activities (commands, deltas, logs) skip the recomputation entirely.
- Proposed-plan and explicit approval/user-input response events keep the full refresh. Every event that mutates `latestTurnId`, proposed plans, or pending-approval rows still refreshes, so the gated kinds are exactly the ones the summary derives from.

**Client (`threadReducer`):**
- Appending an activity takes a fast path (plain append + `updatedAt` bump) only when the current array is known to be reducer-produced and therefore `activityOrder`-sorted, the event is in order relative to the tail, and its id is unseen. A `WeakMap`-backed id set keyed by array identity provides both the sortedness proof and O(1) dedup; it moves forward with each append.
- Everything else — snapshot-loaded arrays (DB order), out-of-order arrivals, re-deliveries, resolvable context-window supersession — falls back to the previous filter + re-sort, which repairs ordering once and seeds the index so subsequent in-order appends are O(1).

**Tests:**
- `ProjectionPipeline`: streaming messages/activities preserve `latestUserMessageAt` and pending counters; `user-input.requested` and approval lifecycle kinds still update the pending counts.
- `threadReducer`: out-of-order insertion, re-delivered activity replacement (including right after fast-path appends), and snapshot-order repair (null-sequence rows in DB order get re-sorted before the fast path engages).

## Why

Streaming a turn generates a high-frequency stream of `thread.activity-appended` and `thread.message-sent` events. Each one previously:

- recomputed the full thread shell summary on the server (queries across messages/activities per event), and
- re-filtered and re-sorted the whole activity array on the client per delta (O(n log n) per event, so O(n² log n) across a long turn).

[...]

Issue #4597, opened 2026-07-27
[Bug]: Headless nightly server crashes with JavaScript heap out of memory after ~23.5 hours


Prgm-code:
### Before submitting

- [x] I searched existing issues and did not find a report with this exact environment and crash signature.
- [x] I included enough detail to reproduce or investigate the problem.

Possibly related reports: #4178, #2761
apps/server

### Steps to reproduce

1. On an Ubuntu VM, start the published nightly server with:

   `npx --yes t3@nightly serve --host "::" --port 3773`

2. Connect to the headless server using the remote web client.
3. Use the server normally with the persisted T3 state under `~/.t3`.
4. Leave the server running continuously.

I do not yet have a smaller deterministic sequence beyond long-running use. In this occurrence, V8 reported an elapsed runtime of `84,604,456 ms` immediately before the fatal error, equivalent to approximately 23 hours, 30 minutes, and 4 seconds.

### Expected behavior

The headless server should remain available during long-running use and its memory consumption should remain bounded.

A large persisted history should not cause the Node process to reach the V8 heap limit and abort.

### Actual behavior

After approximately 23.5 hours, the Node process reached a heap size close to 2 GiB. Mark-Compact garbage collection reclaimed very little memory and the process terminated with:

`FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`

The process then exited with `Aborted (core dumped)`, making the remote T3 server unavailable until it was restarted manually.

Restarting the same command successfully restored the service with the existing state.

A read-only inspection after the restart showed:

- Restarted process RSS after approximately 29 minutes: 756,416 KiB (~738.7 MiB)
- SQLite state database: 181,133,312 bytes (~172.7 MiB)
- Projection threads: 39 total, 2 deleted, 0 archived
- Projected thread activities: 25,601 rows
- Activity JSON payload size: 44,106,381 bytes (~42.1 MiB)
- Orchestration events: 31,800 rows
- Event JSON payload size: 62,380,841 bytes (~59.5 MiB)
- Largest thread: 5,999 activities and approximately 9.35 MiB of activity JSON payload
- Messages: 1,503 rows and approximately 0.88 MiB of text

No `NODE_OPTIONS` or custom `--max-old-space-size` value was configured.

### Impact

Blocks work completely

### Version or commit

`t3 0.0.29-nightly.20260725.899`

Started through `npx --yes t3@nightly`.

### Environment

[...]

PR #6613, opened 2026-08-14
perf(server): stop reading a thread's whole activity timeline per event

patroza:
## The problem

`refreshThreadShellSummary` loads every activity row a thread has ever produced — payloads included — to compute a single integer, `pendingUserInputCount`.

Activity payloads are the tool timeline, so the cost of one refresh scales with the thread's entire history. Across the 402 threads on a heavily-used instance, comparing what the refresh reads against what the derivation needs:

| thread | rows read | of which needed | bytes read | of which needed |
|---|---|---|---|---|
| median | 277 | 0 | 2.8 MB | 0 |
| p95 | 1,190 | 0 | 60.4 MB | 0 |
| p99 | 4,279 | 0 | 123.0 MB | 0 |
| max | 11,137 | 0 | 493.3 MB | 0 |

The "needed" column is zero for 397 of 402 threads, and that is not a quirk of this dataset: `pendingUserInputCount` derives only from `user-input.requested`, `user-input.resolved` and `provider.user-input.respond.failed`, which exist only when a provider stops mid-turn to ask the user something. Even on the 5 threads here that *have* had one, they account for **2–4 rows out of thousands** — 1.3 KB out of 3.5 MB on the largest.

So the read is not merely oversized on one unusual thread; it is reading the whole timeline to find a handful of rows that are usually not there at all.

## This is already reported

- **#5719** — *Assistant streaming causes full-thread projection scans for every text delta* — names `refreshThreadShellSummary` and this exact reload.
- **#4597** — *Headless nightly server crashes with JavaScript heap out of memory after ~23.5 hours* — the symptom. That report died at ~23.5 h; on the instance measured here, systemd recorded a 25 h run peaking at **31.8 G** with **28.8 G read from disk**, and shorter runs at 32.4 G and 36.8 G against a raised `--max-old-space-size`. At the cap the process sits in back-to-back full GCs and every connected client stalls until it is restarted.
- On #5719, `dain` also observed cross-thread head-of-line blocking: one busy thread delaying unrelated turns for minutes. That matches what this looks like from a client — everything slow, not one screen.

## How this relates to #5855 and #6608

Both of those reduce **how often** the refresh runs. This PR reduces **what it costs when it runs**. They compose; none of them replaces another:

| | #5855 | #6608 | this PR |
|---|---|---|---|
| Skip refresh for assistant deltas | ✅ | ✅ | — |
| Skip refresh for streaming activity kinds | — | ✅ | — |
| Make the remaining refreshes cheap | — | — | ✅ |

Neither #5855 nor #6608 changes the read itself — both still call `projectionThreadActivityRepository.listByThreadId({ threadId })` and pull the full payload set. With either merged, every *lifecycle* event (`user-input.requested/resolved`, approval events, proposed-plan upserts, `thread.session-set`, `thread.turn-diff-completed`) still pays the whole-thread read. On the instance above that is up to 467 MB per event, for one integer.

Merging this alongside them means the refreshes that remain are also cheap. If the maintainers prefer #6608's shape, this still applies unchanged on top of it — the two touch different lines.

## The change

`derivePendingUserInputCountFromActivities` only reacts to three kinds — `user-input.requested`, `user-input.resolved` and `provider.user-input.respond.failed` — and `continue`s past everything else. The read now filters on exactly those three, with the kinds declared next to the deriver so the two stay in step.

Measured against the same database. The busiest thread, warm, before schema decode and the sort that follows:

```
before:  10,652 rows      467.4 MB     348.8 ms
after :       0 rows        0.000 MB       6.9 ms
```

And on the threads where the filtered read is *not* empty — the fix's worst case:

```
3,778 rows / 3.5 MB → 2 rows / 1.3 KB     7.3 ms → 1.9 ms
1,576 rows / 1.8 MB → 3 rows / 1.3 KB     2.3 ms → 0.8 ms
1,226 rows / 1.6 MB → 2 rows / 1.4 KB     1.9 ms → 0.6 ms
```

Deployed, the traced projection step moved:

```
applyThreadsProjection   p50 1,707 ms → 1.5 ms     p95 4,814 ms → 17.4 ms     max 15,010 ms → 35.1 ms
```

94% of orchestration events there reach this path (`thread.activity-appended` alone), averaging ~356/hour and peaking at 2,613 in one hour.

`ProjectionThreadActivityRepository` gains `listByThreadIdAndKinds`, which reuses the row decoding of `listByThreadId`, keeps its exact ordering, and short-circuits an empty kind list without issuing a query. `listByThreadId` is unchanged and still used everywhere else.

No behaviour change: the rows removed from the read are ones the deriver already skipped.

## Why it went unnoticed for so long

The code landed in #1973 (2026-04-13) and has not been touched in the 1,239 commits since. It only bites at the tail — on the instance measured, the median thread holds 306 activity rows (~2.8 MB), which is unnoticeable; the p99 holds 4,279 rows / 123 MB. It needs long-lived threads *and* a server process that stays up for days, which is not the common desktop profile.

Adjacent fixes have repeatedly addressed the same underlying volume on the **serve** path — #4622 pruning activity payloads over the wire, #4788 gzipping snapshots, #5482 dropping MCP tool results from thread payloads, #5147 bounding catch-up replay. This is the same volume problem on the **projection** path.

## Tests

Five repository tests cover kind filtering, ordering parity with the unfiltered list, payload-decoding parity, thread isolation, and the empty-kinds short circuit.

One `ProjectionPipeline` test is added, because the existing suite could not catch the failure mode this change introduces. The suite asserted `pendingUserInputCount` only where it settles back to `0`, so a filter that dropped every row would still have passed. The new test appends a `user-input.requested` activity alongside unrelated tool noise and asserts the count is **1**.

Verified by mutation: renaming the filtered kinds makes only the new test fail — the other 22 in that file still pass.

Equivalence was also checked against real data. Replaying the deriver over all 402 threads on the instance above, once from the full activity list and once from the filtered list, produced identical counts for every thread.

Not a UI change, so no screenshots.

### DIFF
--- a/apps/server/src/orchestration/Layers/ProjectionPipeline.test.ts
+++ b/apps/server/src/orchestration/Layers/ProjectionPipeline.test.ts
@@ -2086,6 +2086,125 @@ it.layer(BaseTestLayer)("OrchestrationProjectionPipeline", (it) => {
     }),
   );
 
+  it.effect("counts an unresolved user-input request on the thread shell", () =>
+    Effect.gen(function* () {
+      const projectionPipeline = yield* OrchestrationProjectionPipeline;
+      const eventStore = yield* OrchestrationEventStore;
+      const sql = yield* SqlClient.SqlClient;
+      const appendAndProject = (event: Parameters<typeof eventStore.append>[0]) =>
+        eventStore
+          .append(event)
+          .pipe(Effect.flatMap((savedEvent) => projectionPipeline.projectEvent(savedEvent)));
+
+      yield* appendAndProject({
+        type: "project.created",
+        eventId: EventId.make("evt-open-user-input-1"),
+        aggregateKind: "project",
+        aggregateId: ProjectId.make("project-open-user-input"),
+        occurredAt: "2026-02-26T13:00:00.000Z",
+        commandId: CommandId.make("cmd-open-user-input-1"),
+        causationEventId: null,
+        correlationId: CorrelationId.make("cmd-open-user-input-1"),
+        metadata: {},
+        payload: {
+          projectId: ProjectId.make("project-open-user-input"),
+          title: "Project Open User Input",
+          workspaceRoot: "/tmp/project-open-user-input",
+          defaultModelSelection: null,
+          scripts: [],
+          createdAt: "2026-02-26T13:00:00.000Z",
+          updatedAt: "2026-02-26T13:00:00.000Z",
+        },
+      });
+
+      yield* appendAndProject({
+        type: "thread.created",
+        eventId: EventId.make("evt-open-user-input-2"),
+        aggregateKind: "thread",
+        aggregateId: ThreadId.make("thread-open-user-input"),
+        occurredAt: "2026-02-26T13:00:01.000Z",
+        commandId: CommandId.make("cmd-open-user-input-2"),
+        causationEventId: null,
+        correlationId: CorrelationId.make("cmd-open-user-input-2"),
+        metadata: {},
+        payload: {
+          threadId: ThreadId.make("thread-open-user-input"),
+          projectId: ProjectId.make("project-open-user-input"),
+          title: "Thread Open User Input",
+          modelSelection: {
+            instanceId: ProviderInstanceId.make("codex"),
+            model: "gpt-5-codex",
+          },
+          runtimeMode: "approval-required",
+          interactionMode: "default",
+          branch: null,
+          worktreePath: null,
+          createdAt: "2026-02-26T13:00:01.000Z",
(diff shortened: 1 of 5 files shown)


t3dotgg:
> [!NOTE]

Closing this because its server optimization is already on `main`. [#8988](https://github.com/pingdotgg/t3code/pull/8988) limits shell-summary activity reads to user-input lifecycle rows, and [#8150](https://github.com/pingdotgg/t3code/pull/8150) skips shell-summary refreshes for routine streamed activity.

[#9032](https://github.com/pingdotgg/t3code/pull/9032) records this PR in the related-work credits.

PR #6636, opened 2026-08-14
feat(web): accept file drops across the chat workspace

### LINKED ISSUE #6474: [Feature]: Accept image drops across the active workspace
### Area

apps/web

### Problem or use case

Dragging an image onto the text input attaches it. Dropping it elsewhere in the active workspace does nothing, even though the current thread is the clear destination.

### Proposed solution

The drop zone should cover the active workspace: everything below the window chrome and to the right of the left sidebar. Dropping an image anywhere in that area should attach it to the current thread. The chrome and sidebar should keep their existing drag behavior.

### Why this matters

The current drop target is unnecessarily small, so valid image drops are easy to miss.

### Smallest useful scope

Expand the existing image drop target in the shared web and desktop workspace. Keep the current image types, limits, and validation.

### Alternatives considered

I can use the upload button or aim directly at the text input, but both are slower than dropping into the active workspace.

### Risks or tradeoffs

The larger target should not intercept dragged text, links, or files outside the active workspace.

### Examples or references

[#1790](https://github.com/pingdotgg/t3code/issues/1790) confirms that image drag and drop already exists. This request only expands its drop target.

### Contribution

dbalders:
## Problem

Files only attach when they are dropped directly on the composer, so valid drops elsewhere in the active conversation area are easy to miss.

## Solution

Make the chat column one continuous file-drop target and show a clear “Drop files to attach” overlay while a file drag is active. Drops are forwarded to the existing composer attachment path, so its supported types, limits, validation errors, and focus behavior remain authoritative.

The target deliberately excludes the window chrome, left sidebar, and separate right panel so their existing drag behavior is unchanged. Text, link, and other non-file drags are ignored. Cancelled drags and thread changes clear the overlay.

This expands the drop target only; it does not add new attachment formats or overlap the separate non-image attachment work in #3927.

Fixes #6474

Screenshot 2026-08-14 at 1 11 20 PM

### DIFF
--- a/apps/web/src/components/ChatView.tsx
+++ b/apps/web/src/components/ChatView.tsx
@@ -141,6 +141,7 @@ import { closePreviewSession } from "./preview/closePreviewSession";
 import { ThreadPreviewMiniPlayer } from "./preview/ThreadPreviewMiniPlayer";
 import { subscribePreviewAction } from "./preview/previewActionBus";
 import { getConfiguredPreviewUrls } from "./preview/previewEmptyStateLogic";
+import { makeWorkspaceFileDropHandlers } from "./chat/workspaceFileDrop";
 import {
   selectThreadPreviewMiniPlayer,
   usePreviewMiniPlayerStore,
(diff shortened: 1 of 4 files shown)


CDVolvik (commented):
`workspaceFileDrop.test.ts`: 4/4.

Drops were composer-only; whole-workspace drop is what people try first. MERGEABLE. I did not drag a file in the running app.

Issue #6027, opened 2026-08-10
[Bug]: AskUserQuestion multiSelect commits on first click — cannot select multiple options


bayoumymac:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/web

### Steps to reproduce

1. Have an agent call AskUserQuestion with a question configured `multiSelect: true` ("select all that apply") and 3–4 options.
2. Click one option.

### Expected behavior

Options toggle on click; a Continue/Done action submits the accumulated set of selections.

### Actual behavior

Clicking a single option immediately commits the answer and advances to the next question. There is no way to toggle multiple options and then confirm — multi-select silently degrades to single-select and the agent receives one selection when the user intended several.

### Impact

Major degradation or frequent failure — the agent silently receives wrong/incomplete answers, and neither side can tell.

### Environment

t3code desktop app, macOS 26, Claude provider.

### Notes

Originally this issue also reported the mobile multi-question view being unscrollable; that is already tracked in #5009 (iOS) and #5264 (Android), so this issue is now scoped to the multiSelect behavior only.

Sumstead72:
This still reproduces on 0.0.34 nightly (Windows 11, Claude provider) — first click on a `multiSelect: true` option commits that single option and advances. Re-filed with current detail as #6344, since this issue is closed.

Issue #2707, opened 2026-05-15
[Bug]: Select one or more options question tool call only allowing a single selection


winegardnersam:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/web

### Steps to reproduce

Using claude give a prompt requiring the model to ask a multiple choice question.

Select one of the answers

### Expected behavior

Selecting an answer while check the checkbox and allow more selections to occur, eaving the window open.

### Actual behavior

The window closes as soon as an answer is selected.

### Impact

Minor bug or occasional failure

### Environment

macOS

### Logs or stack traces

```shell

```

AngeloTadeucci:
+1 on this

on latest 0.0.25-nightly.20260606.480

skint007:
+1 for this also.
on stable 0.0.28

eggins:
+1 — still reproducing on `0.0.29-nightly.20260725.899`. Multi-select questions from Claude still close on the first option clicked instead of staying open for additional selections.

juliusmarminge:
<img width="3024" height="1898" alt="Image" src="https://github.com/user-attachments/assets/394ea760-05ba-4ba0-b582-8d1dea7465a0" />

PR #6646, opened 2026-08-14
fix(web): keep multi-select questions open after the first click

### LINKED ISSUE #6344: [Bug]: AskUserQuestion multiSelect still commits on first click — #6027 closed with no fix, reproduces on 0.0.34
### Before submitting

> Re-filing rather than commenting because the prior report is closed and
> unlikely to be seen. See "Relationship to #6027 and #2707" below.

### Area
apps/web

### Steps to reproduce
1. Have the Claude provider call `AskUserQuestion` with a question configured
   `multiSelect: true` and 3–4 options.
2. Observe the panel renders the subtitle "Select one or more options".
3. Click any option.

### Expected behavior
Options toggle on and off, accumulating a set. A "Submit answers" action commits
all selected options together, and the model receives the full set.

### Actual behavior
The first click immediately commits that single option and advances to the next
question. There is no chance to toggle a second option. Multi-select silently
degrades to single-select, and the model receives one answer where several were
intended — with nothing on either side indicating the loss.

Notably the subtitle is correct: the client *does* read `multiSelect` for the
"Select one or more options" copy, but the option click handler still commits and
advances as if the question were single-select. A "Submit answers" button is
present in the footer but the click handler fires before it can be used.

### Impact
Major degradation or frequent failure. The wrong answer is silently substituted —
the agent proceeds on incomplete input, and neither the user nor the model can
detect that the other options were dropped.

### Version or commit
0.0.34 nightly. The installed executable reports ProductVersion `0.0.34.0`; the
full nightly build string is not exposed in the installed files.

### Environment
T3 Code desktop app (Nightly channel), Windows 11, Claude provider.

### Relationship to #6027 and #2707
This is the same defect as **#6027**, which was closed as completed on 2026-08-10
— 44 minutes after it was filed, with no linked commit, no pull request, and no
comment. **#2707** before it was closed the same way. It still reproduces on
0.0.34 on Windows, so I do not believe a fix for the toggle behavior ever landed;
#6027 appears to have been closed against #2707's earlier closure rather than
against a change.

Cross-platform note: #6027 reported this on macOS and #2707 on macOS, so this is
not Windows-specific.

### Workaround
Answer the questions one at a time across several turns, or ask the agent to
re-pose the question as free text and type the multiple answers into the
"Type your own answer" field.

RaitP1:
## What Changed

`apps/web/src/components/chat/ComposerPendingUserInputPanel.tsx`: `handleOptionSelection` is a plain `useCallback` with `[activeQuestion, onToggleOption]` instead of `useEffectEvent`, and the number-key effect lists it as a dependency.

Fixes #6344 (re-file of #6027 and #2707).

## Why

A question with `multiSelect: true` commits and advances on the first click whenever it is
**not** the first question of the prompt, so multi-select degrades to single-select and the
model receives one answer where several were intended.

The multi-select branch already exists. It never runs, because the guard reads a stale
question. `ComposerPendingUserInputCard` is wrapped in `memo`, and React applies effect-event
implementations only for plain function-component fibers: in `commitBeforeMutationEffects`,
tag `0` assigns `ref.impl = nextImpl`, while `15` (`SimpleMemoComponent`) and `11`
(`ForwardRef`) fall through to `break`. The `useEffectEvent` closure therefore freezes at the
first render, and the `multiSelect` check keeps reading question 1 for the whole prompt.

The click passes `activeQuestion.id` from the *render* closure, which is current, so the
answer is still stored against the right question. Only the guard is stale, so a later
multi-select question takes the single-select path and arms the 200 ms auto-advance timer.
Mirror case: put a multi-select question first, and later single-select questions never advance.

Reproduces on `upstream/main` at 9885a845c. macOS, desktop app, Claude provider. No test
accompanies the fix — the `apps/web` unit project renders with `renderToStaticMarkup` and has
no DOM environment, so a click cannot be simulated.

## UI Changes

Same window, theme, viewport, and question set in both clips. The only difference is this commit.

**Before** — one click on the multi-select step commits the whole prompt:

https://github.com/user-attachments/assets/b417a53c-bed5-489a-a4d5-8d8fc5712d3b

**After** — the multi-select step holds every pick and waits:

https://github.com/user-attachments/assets/41f2fa16-2e03-4a7f-80b4-ef9eef84cd1d

## Checklist

---

Written with Claude Opus 5 (1M context) in Claude Code.

---

> [!NOTE]
> **Low Risk**
> Localized chat composer interaction fix with no auth, data, or API surface changes.
> 
> **Overview**
> Fixes pending user-input prompts where **multi-select** questions behaved like single-select (auto-advancing after one pick) whenever they were not the first question in the flow.
> 
> In `ComposerPendingUserInputPanel`, **`handleOptionSelection`** is switched from **`useEffectEvent`** to **`useCallback`** with `[activeQuestion, onToggleOption]` so the handler always sees the current question’s `multiSelect` flag. The number-key shortcut **`useEffect`** now depends on **`handleOptionSelection`** so keyboard and click paths stay aligned.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a051e4e1a4c7c9fa47ac783ac24092661e49b5d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix multi-select questions staying open after first click in chat composer
> Replaces `useEffectEvent` with `useCallback` for `handleOptionSelection` in [ComposerPendingUserInputPanel.tsx](https://github.com/pingdotgg/t3code/pull/6646/files#diff-4d79f14a0bb3d90ecac6af885c54e9a485bcd9510c4a339f6bfdbcde568a0b0e), adding `[activeQuestion, onToggleOption]` as dependencies. The previous implementation caused multi-select dropdowns to close after a single selection because the stale handler reference wasn't triggering a re-bind of the keyboard shortcut effect.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a051e4e.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/chat/ComposerPendingUserInputPanel.tsx
+++ b/apps/web/src/components/chat/ComposerPendingUserInputPanel.tsx
@@ -1,5 +1,5 @@
 import { type ApprovalRequestId } from "@t3tools/contracts";
-import { memo, useEffect, useEffectEvent, useRef, useState } from "react";
+import { memo, useCallback, useEffect, useRef, useState } from "react";
 import { type PendingUserInput } from "../../session-logic";
 import {
   derivePendingUserInputProgress,
@@ -109,21 +109,24 @@ const ComposerPendingUserInputCard = memo(function ComposerPendingUserInputCard(
     };
   }, []);
 
-  const handleOptionSelection = useEffectEvent((questionId: string, optionLabel: string) => {
-    if (activeQuestion?.multiSelect) {
+  const handleOptionSelection = useCallback(
+    (questionId: string, optionLabel: string) => {
+      if (activeQuestion?.multiSelect) {
+        onToggleOption(questionId, optionLabel);
+        return;
+      }
+      setOptimisticSingleSelect({ questionId, optionLabel });
       onToggleOption(questionId, optionLabel);
-      return;
-    }
-    setOptimisticSingleSelect({ questionId, optionLabel });
-    onToggleOption(questionId, optionLabel);
-    if (autoAdvanceTimerRef.current !== null) {
-      window.clearTimeout(autoAdvanceTimerRef.current);
-    }
-    autoAdvanceTimerRef.current = window.setTimeout(() => {
-      autoAdvanceTimerRef.current = null;
-      onAdvanceRef.current();
-    }, 200);
-  });
+      if (autoAdvanceTimerRef.current !== null) {
+        window.clearTimeout(autoAdvanceTimerRef.current);
+      }
+      autoAdvanceTimerRef.current = window.setTimeout(() => {
+        autoAdvanceTimerRef.current = null;
+        onAdvanceRef.current();
+      }, 200);
+    },
+    [activeQuestion, onToggleOption],
+  );
 
   // Keyboard shortcut: number keys 1-9 select corresponding options when focus is
   // outside editable fields. Multi-select prompts toggle options in place; single-
@@ -154,7 +157,7 @@ const ComposerPendingUserInputCard = memo(function ComposerPendingUserInputCard(
     };
     document.addEventListener("keydown", handler);
     return () => document.removeEventListener("keydown", handler);
-  }, [activeQuestion, isCollapsed, isResponding]);
+  }, [activeQuestion, handleOptionSelection, isCollapsed, isResponding]);
 
   if (!activeQuestion) {
     return null;


RaitP1:
Prompt used in the video test:
```
Call the AskUserQuestion tool exactly once with the JSON below. Do not change any
text, label, description, or the multiSelect flags. Do not run any other tool.
After I answer, reply with only "done".

{
  "questions": [
    {
      "header": "Single",
      "question": "Step 1 of 3 — single-select. Click one option. It must jump to step 2 at once.",
      "multiSelect": false,
      "options": [
        { "label": "Alpha", "description": "Click either option. Both do the same thing." },
        { "label": "Beta", "description": "Click either option. Both do the same thing." }
      ]
    },
    {
      "header": "Multi",
      "question": "Step 2 of 3 — multi-select. Click two or three options. They must all stay lit.",
      "multiSelect": true,
      "options": [
        { "label": "One", "description": "Click this one." },
        { "label": "Two", "description": "Click this one too. It must not replace the first." },
        { "label": "Three", "description": "Add a third if you want." },
        { "label": "Four", "description": "Add a fourth if you want." }
      ]
    },
    {
      "header": "Mirror",
      "question": "Step 3 of 3 — single-select after a multi-select step. One click must finish the flow.",
      "multiSelect": false,
      "options": [
        { "label": "All three steps worked", "description": "Step 2 held several picks and this step commits at once." },
        { "label": "Step 2 still jumped away", "description": "The first click on step 2 moved the panel forward." },
        { "label": "Step 2 held, this step is stuck", "description": "Step 2 worked, but this step does not commit on one click." }
      ]
    }
  ]
}
```

CDVolvik (commented):
Multi-select already returned before the 200ms auto-advance. Switching `useEffectEvent` to `useCallback` and putting `handleOptionSelection` in the keydown effect deps is what keeps a later option click from seeing a stale single-select handler.

No test in the diff. MERGEABLE. I did not click a multi-select prompt in the app.

juliusmarminge (approved):
Re-approved after the verified conflict resolution and focused component checks.

PR #6649, opened 2026-08-14
fix(desktop): app zoom no longer zooms the preview browser

juliusmarminge:
Zooming the app UI (View → Zoom In/Out) also zoomed whatever the preview browser was showing, so there was no way to shrink T3's chrome without shrinking the previewed page with it — which is what Tanner asked for: "zoom T3's UI out, but not the browser".

Chromium hands every guest `<webview>` the embedder's zoom level, and `PreviewManager` then read that level back off the guest, so an app zoom silently became the preview's own zoom (and the zoom indicator reported it).

The tab's zoom factor is now the only source of truth: it is asserted onto the guest whenever a webview attaches, re-asserted after every app zoom, and never read back off the guest.

## Before / after

App zoomed out two steps, same window, same page. The readout is the previewed page's own `devicePixelRatio` and viewport.

**Before** — the page shrinks with the UI (`devicePixelRatio` drops to 1.67):

before

**After** — the UI shrinks, the page stays at true scale (`devicePixelRatio` stays 2.00; the guest just gets a narrower viewport as the panel gets physically smaller):

after

At 100% both builds are identical, and the preview's own zoom controls are unchanged:

baseline

## Notes

- Desktop-only: the preview browser is an Electron `<webview>`, so web and mobile are unaffected.
- One behaviour change worth calling out: a zoom the user makes *inside* the guest (ctrl+wheel) is no longer mirrored into the tab state. That mirror was the leak — it could not tell an inherited app zoom from a real one — and T3's own zoom controls remain the way to zoom the preview.
- Verified in the desktop app (screenshots above) plus focused tests in `Manager.test.ts` and `DesktopWindow.test.ts`.

Model: Claude Opus 5 (1M context). Harness: Claude Code inside T3 Code.

---

> [!NOTE]
> **Medium Risk**
> Touches desktop preview webview lifecycle and zoom synchronization; behavior change for guest-reported zoom, but scoped to Electron preview and covered by new tests.
> 
> **Overview**
> **App UI zoom no longer changes the previewed page’s scale.** Chromium was inheriting the main window zoom into guest `<webview>`s, and `PreviewManager` then mirrored `getZoomFactor()` back into tab state—so View → Zoom In/Out affected both chrome and the preview.
> 
> Tab **`zoomFactor` is now authoritative**: `syncWebContentsState` stops reading zoom from the guest; **`assertTabZoom`** pushes the stored factor onto the active webContents (including after async attach and on duplicate `registerWebview`); new **`reapplyZoom`** resets every tab’s guest after app zoom. **`DesktopWindow.zoomMain`** calls `reapplyZoom()` after each main-window zoom step.
> 
> **Behavior change:** in-guest zoom (e.g. ctrl+wheel) is no longer synced into tab state; preview zoom stays on T3’s own controls. Tests cover tab-owned zoom, `reapplyZoom`, re-register paths, and desktop window integration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a46468dc59a53a30a595ea8cc942327c93711b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix app window zoom from affecting preview browser tab zoom
> - Embedded preview guests inherit the host window's zoom level, causing app zoom to unintentionally rescale preview content.
> - `PreviewManager` now tracks each tab's own zoom factor and asserts it onto the guest on registration, re-registration, and navigation, rather than reading the guest's zoom back into state.
> - A new `reapplyZoom()` method on `PreviewManager` iterates all tabs and pushes their zoom to attached guests; `DesktopWindow.zoomMain` calls this after each window zoom change.
> - Behavioral Change: the manager no longer adopts zoom from guests — guest zoom is always overwritten by the tab's stored value on attach.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0a46468.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/preview/Manager.test.ts
+++ b/apps/desktop/src/preview/Manager.test.ts
@@ -979,7 +979,10 @@ describe("PreviewManager", () => {
     ),
   );
 
-  effectIt.effect("mirrors Electron's effective zoom across registration and navigation", () =>
+  // The guest reports whatever zoom level Chromium handed it from the app
+  // window, so the tab's own zoom is the source of truth in both directions:
+  // asserted onto every guest, never read back off one.
+  effectIt.effect("keeps the tab's own zoom instead of the guest's reported zoom", () =>
     withManager((manager) =>
       Effect.gen(function* () {
         let effectiveZoom = 0.9;
@@ -1025,18 +1028,13 @@ describe("PreviewManager", () => {
         yield* manager.createTab("tab_zoom");
         yield* manager.registerWebview("tab_zoom", 42);
 
-        expect(states.at(-1)?.zoomFactor).toBe(0.9);
-        expect(setZoomFactor).not.toHaveBeenCalled();
+        expect(states.at(-1)?.zoomFactor).toBe(1);
+        expect(setZoomFactor).toHaveBeenCalledWith(1);
 
-        effectiveZoom = 1.25;
-        listeners.get("did-navigate")?.();
-        yield* Effect.yieldNow;
-
-        expect(states.at(-1)?.zoomFactor).toBe(1.25);
-        expect(setZoomFactor).not.toHaveBeenCalled();
-
-        zoomReadable = false;
-        url = "https://example.com/after-zoom-read-failed";
+        // An app zoom leaves the guest reporting the inherited level. Navigating
+        // must not adopt it as the preview's zoom.
+        effectiveZoom = 0.8;
+        url = "https://example.com/after-app-zoom";
         listeners.get("did-navigate")?.();
         yield* Effect.yieldNow;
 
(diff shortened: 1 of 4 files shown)


juliusmarminge:
Addressed in `0a46468` — both bots flagged the same thing from two angles, so I fixed it at the source rather than patch-by-patch.

**What was wrong.** Three places applied a tab snapshot taken *before* an await: the `registerWebview` fast path, the full attach path (which also wrote its pre-`attachListeners` factor into the committed state), and `reapplyZoom`. A zoom action landing in that window could be rolled back on the guest, and the fast path additionally wrote its stale factor back over tab state.

**Fix.** One helper, `assertTabZoom(tabId)`, reads the tab *and* resolves its guest at the moment it applies, and every post-await application goes through it. The fast path no longer republishes unchanged tab state (that write could only ever clobber), the attach path no longer writes its own `zoomFactor` into the committed state, and a swapped-in guest settles on the committed factor after the commit lands.

**On reachability, so the record is straight:** I could not build a failing test for the fast-path window — every step between the read and the apply is synchronous, so the fiber never suspends there and the interleaving does not occur today. It is reachable in the attach path, which does suspend. I kept the change anyway because it removes the hazard class and is smaller than what it replaces, but I am not claiming a test proves the race. What the new test does cover is behaviour this PR introduces and nothing else covered: re-registering the already-attached guest re-asserts the tab's zoom without republishing state. It fails against `main`.

Re-verified end to end in the desktop app on the updated build — previewed page holds `devicePixelRatio` 2.00 across four zoom-out steps, zoom-in, and reset, while its viewport tracks the panel. 74 tests, lint, and typecheck green.

Issue #5550, opened 2026-08-06
[Feature]: Add supported retention and compaction for unbounded local SQLite growth


Jules-Astier:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I am describing a concrete problem or use case, not just a vague idea.

### Area

apps/server

### Problem or use case

T3 Code's local SQLite database grows without a supported retention or compaction path. Settling, archiving, or deleting a task does not appear to reclaim its persisted activity/event payloads; settling currently behaves like a UI/state override rather than a storage lifecycle operation.

On one macOS installation using `0.0.32-nightly.20260804.997`, `~/.t3/userdata/state.sqlite` reached **3.7 GiB** with:

- 171 threads
- 389,181 orchestration events
- 388,131 command receipts
- 350,773 projected activity rows
- only 1,928 turns

A small number of long-running/tool-heavy threads dominated storage. One thread alone held about **215.8 MiB** in projected activities. The database remained valid (`PRAGMA quick_check = ok`), so this was live retained data, not corruption or free-page bloat.

The size eventually contributed to V8 heap OOM crashes during hydration. The desktop UI reported these as a failed local environment connection to `127.0.0.1:3773`, obscuring the storage problem.

An offline, backup-first maintenance pass that retained only the newest 500 projected activities per thread removed 300,475 projection rows and reduced the database from **3.7 GiB to 2.3 GiB** after `VACUUM`. It preserved all events, receipts, messages, turns, and threads. That still leaves the append-only event copy unbounded.

### Proposed solution

Add a supported storage-retention and compaction system for the local SQLite database:

1. Bound projected activity history per thread (for example, retain the newest 500 by default), with pagination/lazy loading if older activity remains available.
2. Compact or snapshot the append-only event history so obsolete activity/message deltas and command receipts do not grow forever.
3. Preserve unresolved approvals, user-input requests, current session state, final messages, checkpoints, and other state required to reconstruct the thread correctly.
4. Apply a documented lifecycle policy to settled, archived, and deleted threads. Settling could queue eligible compaction, or the UI should clearly state that it has no storage effect.
5. Run checkpoint/`VACUUM` safely when worthwhile, outside latency-sensitive paths.
6. Add a diagnostics/settings view showing database size, largest threads, retained activity/event counts, and a backup-first “Compact now” action.
7. Make retention thresholds configurable and provide a dry-run estimate before destructive compaction.

### Why this matters

[...]

PR #6675, opened 2026-08-15
perf(server): persist the wire projection for streaming tool.updated data

### LINKED ISSUE #6673: [Bug]: Streaming tool.updated persists the full accumulated rawOutput on every delta — O(N²) bytes and rows per tool call
### Summary

Each streaming `tool.updated` event persists the full accumulated tool output, not the delta. A tool call with N updates writes O(N²) bytes and creates one activity row per update. The 180-char truncation applies only to `detail`; `data` passes through verbatim.

Context: t3 0.0.32, Node 24, Linux LXC. Measured 2026-08-13 to 2026-08-15 on two self-hosted instances.

### Mechanism

1. `AcpSessionRuntime.handleSessionUpdate` merges each ACP `toolCallUpdate` into a per-call state. `mergeToolCallState` spreads `...previous.data` forward, so `data.rawOutput` and `data.content` always hold the full output produced so far.
2. `shouldEmitToolCallUpdate` compares only the 180-char `detail` string. A chatty tool (progress bar, build log) changes `detail` every few milliseconds, so it passes the gate on every chunk.
3. `ProviderRuntimeIngestion` truncates `detail` to 180 chars but passes `data` through verbatim.
4. Each emission becomes one orchestration event. The server writes the payload twice: into `orchestration_events.payload_json` and into `projection_thread_activities.payload_json`. Each update carries a fresh `activity_id`, so the `ON CONFLICT (activity_id) DO UPDATE` upsert never collapses the stream.

### Measured

- One tool call produced 2,720 `tool.updated` events totalling 357 MB of payload.
- One 65 KB tool result persisted 238.7 MB of surviving rows: 3,772× amplification. The payload length series for the call rises monotonically from 3,819 B to 66,356 B across 2,226 updates.
- One thread reached 3,232 activity rows / 351 MB from render jobs with progress-bar output.
- One instance wrote 60.4 GB of physical disk in 27 hours to persist 18,612 events.

### Why this matters

#6608 / #6613 / #5855 fix the read side (`refreshThreadShellSummary` rescans). This is the write side, and the two multiply: every extra emitted event is also one more full-thread projection pass on unpatched versions. It is also the main input volume behind the unbounded growth in #5550.

### Suggested fix

Cap or drop `data.rawOutput` and `data.content` on non-terminal `tool.updated` emissions, the same way `detail` is capped. Keep the full payload on `tool.completed` only. A follow-up could derive a stable `activity_id` from the `toolCallId` so the existing upsert collapses a stream into one row.

We have a patch for the first part and will open a PR against this issue.

mInrOz:
Fixes #6673. Related: #6608 / #5855 (read side of the same event flood), #5550 (retention), #5651 (assistant message write path). Decision date: 2026-08-15.

## Outcome

Non-terminal `tool.updated` activities now persist the same projected `data` that clients receive on the wire. `projectActivityPayload` already runs in `ws.ts` and in `http.ts`, so no client can receive the full `data` of any activity today. This PR applies the same projection at ingestion time, before the payload reaches `orchestration_events` and `projection_thread_activities`.

This PR does not change: terminal `tool.completed` payloads (still persisted in full), the event count per tool call, any schema, any existing rows, or any client-visible payload.

## Before and after

| Failure mode before | Mechanism after |
| --- | --- |
| A streaming update persists the full accumulated `rawOutput`/`content`, so one tool call with N updates writes O(N²) bytes, twice. Measured: one 65 KB tool result persisted 238.7 MB across 2,226 updates (3,772×); one tool call produced 357 MB of payload. | A non-terminal update persists the projected form: `toolCallId`, `kind`, `command`, changed files, and a one-line `rawOutput` summary. Bytes per update are bounded (~1 KB), so a tool call writes O(N) bytes. |
| The event store grows without bound from streaming payloads (~235 MB/day measured on one instance), which multiplies the full-thread rescan cost fixed by #6608. | Streaming updates no longer carry payload bulk into the event store. Terminal payloads remain the only large rows. |

## Decisions taken

- **Reuse `projectActivityPayload` instead of adding a size cap.** A cap needs a constant and keeps a second copy of the slimming rules. Reuse keeps one invariant: for streaming updates, persistence stores what clients can receive. The projection's own doc comment ("full payloads remain in persistence") described the pre-#6673 intent; this PR revises that intent for non-terminal updates only.
- **`item.completed` stays untouched.** The terminal payload is the record of what the tool produced. It is one row per call, so it does not amplify.
- **Event count is not reduced.** Debouncing emissions or reusing a stable `activity_id` per `toolCallId` changes what clients observe and belongs in a separate change. Deferred; noted in #6673 as a follow-up.
- **Interrupted tool calls keep only projected data for their last update.** Accepted: clients could never render more than the projected form anyway.

## Finding disposition (#6673)

- O(N²) bytes per tool call, stored twice → fixed here, regression-tested.
- One activity row per update (O(N) rows) → deferred; requires an id-semantics change (see #6673 suggested follow-up).
- `shouldEmitToolCallUpdate` gates only on `detail` → not changed; with bounded payloads the extra emissions cost little, and fewer emissions would change client behavior.

## Verification

Ran on Linux, Node v24.18.1, pnpm 11.10.0:

- `pnpm exec vp test run src/orchestration/Layers/ProviderRuntimeIngestion.activity.test.ts` — 4 passed (2 new: projected persistence for `tool.updated`; full payload kept for `tool.completed`).
- `pnpm exec vp test run src/orchestration/Layers/ProviderRuntimeIngestion.test.ts src/orchestration/Layers/ProviderRuntimeIngestion.approval.test.ts src/orchestration/ActivityPayloadProjection.test.ts` — 52 passed.
- `pnpm exec tsgo --noEmit` in `apps/server` — no errors (pre-existing suggestions in unrelated files only).

Not run: the full monorepo suite (left to CI) and a production soak (moves to the rollout gate below).

## Rollout and proof gates

We run t3 on self-hosted instances with the exact workload from #6673 (render jobs with progress-bar output). Ordered steps to produce production proof:

1. Deploy a build of this branch to an idle instance.
2. Run a chatty render job to completion.
3. Compare `SELECT count(*), sum(length(payload_json)) FROM projection_thread_activities WHERE thread_id = ?` against an unpatched run of the same job. Expected: similar row count, payload bytes down by >100×.
4. Report the numbers on this PR.

Rollback: reinstall the released npm build. The change is write-path only; there is no schema change and old rows are untouched, so nothing durable is lost by rolling back.

---

> [!NOTE]
> **Medium Risk**
> Write-path change to persisted activity payloads for streaming tool updates; behavior aligns with existing ws/http projection but could affect any internal reader that expected full in-progress data in the DB.
> 
> **Overview**
> Fixes **O(N²) storage** from streaming tool runs by persisting the same slim **`data`** clients already get on the wire for non-terminal **`tool.updated`** activities, instead of the full accumulated stdout/content on every chunk.
> 
> **`runtimeEventToActivities`** now runs **`projectActivityPayload`** on **`item.updated`** tool lifecycle events before rows hit the event store and **`projection_thread_activities`**. **`item.completed`** is unchanged and still stores the full terminal payload. Event count and client-visible payloads are unchanged; only what gets written on the ingestion path for in-progress updates.
> 
> Regression tests assert projected **`tool.updated`** **`data`** (bounded size, one-line **`rawOutput`**) and full **`streamingData`** on **`tool.completed`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 66d0ce4d898989977dfad94b855eddd6e90ff0f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Project wire payload for streaming `tool.updated` activities to avoid quadratic growth
> Streaming tool output accumulates over time, so persisting the full `data` payload on every `item.updated` event causes storage to grow quadratically with stream length. `runtimeEventToActivities` in [ProviderRuntimeIngestion.ts](https://github.com/pingdotgg/t3code/pull/6675/files#diff-22a6c9818b956463b848a73a5b3e787b44a144bd003b1cc25fa2b1b3110cdea7) now wraps `item.updated` tool activities with `projectActivityPayload`, which truncates raw output and omits content fields. Terminal `item.completed` events are unaffected and still persist the full payload.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 66d0ce4.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.activity.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.activity.test.ts
@@ -82,3 +82,63 @@ describe("runtimeEventToActivities task progress", () => {
     expect(usagePayload).not.toHaveProperty("status");
   });
 });
+describe("runtimeEventToActivities tool streaming persistence", () => {
+  const accumulatedStdout = [
+    "first line of output",
+    ...Array.from({ length: 500 }, (_, index) => `Capturing frame ${index}/9028`),
+  ].join("\n");
+  const streamingData = {
+    toolCallId: "tool-call-1",
+    kind: "execute",
+    command: "blender --render",
+    rawOutput: { stdout: accumulatedStdout },
+    content: [{ type: "content", content: { type: "text", text: accumulatedStdout } }],
+  };
+
+  it("persists tool.updated with the wire projection of data, not the accumulated stream", () => {
+    const event = {
+      ...base,
+      type: "item.updated",
+      eventId: EventId.make("evt-tool-streaming-updated"),
+      payload: {
+        itemType: "command_execution",
+        status: "inProgress",
+        title: "Render",
+        detail: accumulatedStdout,
+        data: streamingData,
+      },
+    } satisfies ProviderRuntimeEvent;
+
+    const activities = runtimeEventToActivities(event);
+
+    expect(activities).toHaveLength(1);
+    const payload = activities[0]?.payload as Record<string, unknown>;
+    const data = payload.data as Record<string, unknown>;
+    expect(payload.status).toBe("inProgress");
+    expect(data.toolCallId).toBe("tool-call-1");
+    expect(data.command).toBe("blender --render");
+    expect(data.rawOutput).toEqual({ content: "first line of output" });
+    expect(data.content).toBeUndefined();
+    expect(JSON.stringify(data).length).toBeLessThan(1_000);
+  });
+
+  it("persists the full terminal payload on tool.completed", () => {
+    const event = {
+      ...base,
+      type: "item.completed",
+      eventId: EventId.make("evt-tool-streaming-completed"),
+      payload: {
+        itemType: "command_execution",
+        status: "completed",
+        title: "Render",
+        data: streamingData,
+      },
+    } satisfies ProviderRuntimeEvent;
+
+    const activities = runtimeEventToActivities(event);
+
+    expect(activities).toHaveLength(1);
+    const payload = activities[0]?.payload as Record<string, unknown>;
+    expect(payload.data).toEqual(streamingData);
+  });
+});
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts
@@ -41,6 +41,7 @@ import {
   ProviderRuntimeIngestionService,
   type ProviderRuntimeIngestionShape,
 } from "../Services/ProviderRuntimeIngestion.ts";
+import { projectActivityPayload } from "../ActivityPayloadProjection.ts";
 import { forkParked } from "../../serverActivation.ts";
 import { ServerSettingsService } from "../../serverSettings.ts";
 
@@ -785,8 +786,15 @@ export function runtimeEventToActivities(
       if (!isToolLifecycleItemType(event.payload.itemType)) {
         return [];
       }
+      // A streaming update's `data` carries the full tool output accumulated
+      // so far (adapters merge state forward), and a new activity is emitted
+      // per chunk, so persisting `data` verbatim writes O(N²) bytes per tool
+      // call into both the event store and the projection table. No reader
+      // needs it: ws.ts and http.ts apply `projectActivityPayload` before any
+      // payload reaches a client. Persist the projected form for non-terminal
+      // updates; `item.completed` below still persists the full payload.
       return [
-        {
+        projectActivityPayload({
           id: event.eventId,
           createdAt: event.createdAt,
           tone: "tool",
@@ -804,7 +812,7 @@ export function runtimeEventToActivities(
           },
           turnId: toTurnId(event.turnId) ?? null,
           ...maybeSequence,
-        },
+        }),
       ];
     }
 


mInrOz:
Rollout-gate results from our idle instance (Linux LXC, Node v24.19.0, 2026-08-15).

### A/B benchmark through the real persistence stack

We ran one identical synthetic tool call through `runtimeEventToActivities` → `OrchestrationEventStore.append` → `OrchestrationProjectionPipeline.projectEvent`, on this branch (66d0ce4d) and on its parent (f0ebc628). The harness is an untracked test file on the repo's own test layers (`SqlitePersistenceMemory` + `OrchestrationEventStoreLive` + `OrchestrationProjectionPipelineLive`), so the SQL write path is the production code path. Workload: 300 `item.updated` events, each adds 10 output lines; final `rawOutput` is 76,892 bytes; one terminal `item.completed` carries the full output.

| metric | parent f0ebc628 | this branch | change |
| --- | ---: | ---: | ---: |
| `projection_thread_activities` bytes | 24,018,619 | 190,135 | 126× less |
| `orchestration_events` bytes | 24,074,799 | 246,315 | 98× less |
| total persisted bytes | 48.1 MB | 436 KB | 110× less |
| bytes persisted / final output | 625× | 5.7× | — |
| append+project wall time | 8,719 ms | 979 ms | 8.9× faster |
| activity rows | 301 | 301 | unchanged |

The row count is unchanged, as intended: this PR bounds bytes, not event count. The wall-time gain is a side effect of smaller JSON through the synchronous SQL path.

### Deploy smoke

`pnpm run build:bundle` on this branch; ran `dist/bin.mjs serve` against a fresh `--base-dir` with the released web client assets linked in. Result: "Migrations ran successfully", "Listening on http://127.0.0.1:3782", `GET /` and `GET /api/health` return 200. Stopped after the probe.

PR #7068, opened 2026-08-15
fix(codex): frame large app-server records linearly

### LINKED ISSUE #5389: [Bug]: app-server stdin reader is quadratic in line length — OOM crash on large tool payloads
## Steps to reproduce

1. Run a self-hosted server (`t3` serve) with the Codex provider.
2. Start a thread and have the agent use the browser preview to fetch and read
   web pages, so `tool.completed` results contain full page content.
3. Let it run for a while — in my case roughly an hour of ordinary browsing work
   with a single agent.

I don't have a minimal deterministic repro, sorry. The trigger seems to be a
single JSON-RPC line on the app-server stdio pipe that is large (megabytes) and
therefore arrives across many chunks. Any tool result big enough to be delivered
in many reads should do it; page content is just the easiest way to get there.

## Expected behavior

Reading a large message off the app-server pipe costs time proportional to the
size of the message.

## Actual behavior

The server's CPU goes to ~100% of a core and heap climbs until Node aborts:

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0xe42d60 node::OOMErrorHandler(char const*, v8::OOMDetails const&)
 2: 0x121ddd0 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, ...)
 3: 0x121e0a7 v8::internal::V8::FatalProcessOutOfMemory(...)
 5: 0x1465299 v8::internal::Heap::CollectGarbage(...)
 6: 0x1439948 v8::internal::HeapAllocator::AllocateRawWithLightRetrySlowPath(...)
 7: 0x143a875 v8::internal::HeapAllocator::AllocateRawWithRetryOrFailSlowPath(...)
```

and systemd records:

```
t3code.service: Main process exited, code=dumped, status=6/ABRT
t3code.service: Failed with result 'core-dump'.
t3code.service: 8.8G memory peak
```

The clients disconnect when the process dies. systemd restarted it and it
happened again on the same workload.

## Where it comes from

`packages/effect-codex-app-server/src/protocol.ts` (v0.0.31, around line 354),
in the stdin reader:

```ts
Ref.modify(remainder, (current) => {
  const combined = current + chunk;
  const lines = combined.split("\n");
  const nextRemainder = lines.pop() ?? "";
  return [lines.map((line) => line.replace(/\r$/, "")), nextRemainder] as const;
})
```

Every chunk concatenates onto the whole pending buffer and re-splits all of it.
For one line arriving across N chunks that's O(n²) in both copying and
allocation. A ~50 MB line in 64 KB chunks works out to roughly 20 GB of copying
to receive 50 MB.

It only shows up when a single line is large. Normal coding tool results are
small enough that the accumulated buffer never gets big, which I assume is why
this hasn't come up before.

## Evidence

V8 CPU profile of the process (`Profiler.start`, 25 s, sampled at 100 µs) while
reproducing, aggregated by self time:

```
96.8%  (anonymous)  bin.mjs:68926   -> protocol.ts:354 via the shipped sourcemap
 2.5%  (garbage collector)
 0.0%  everything else
```

`process.memoryUsage()` at the same moment:

```
rss           1955.7 MB
heapUsed      1743.3 MB
external         3.8 MB
arrayBuffers     0.2 MB
```

So it is JS heap, not buffers.

## Possible fix

Holding the pending bytes as the chunks that produced them and joining only when
a newline actually arrives makes line assembly linear — a chunk with no newline
becomes an array push instead of a full buffer copy:

```ts
const remainder = yield* Ref.make<ReadonlyArray<string>>([]);

Ref.modify(remainder, (current) => {
  if (!chunk.includes("\n")) {
    return [[] as ReadonlyArray<string>, [...current, chunk]] as const;
  }
  const combined = current.length === 0 ? chunk : current.join("") + chunk;
  const lines = combined.split("\n");
  const nextRemainder = lines.pop() ?? "";
  return [
    lines.map((line) => line.replace(/\r$/, "")),
    nextRemainder.length === 0 ? [] : [nextRemainder],
  ] as const;
})
```

The stream-end path also needs `Effect.map((pending) => pending.join(""))` before
handling the trailing partial line.

I've been running this on a local fork since the crash. Under the same browsing
workload the profile no longer shows this function at all, heap now sawtooths and
recovers instead of climbing, and the OOM hasn't recurred. Package typecheck and its 20 tests pass, and the
server suite (1771 passed / 7 skipped) is unaffected. Happy to open a PR if
that's useful, or equally happy for you to take the idea and do it your own way —
I know you're not looking for contributions right now.

## Note

I can see #2829 is rewriting the orchestration layer, and I had a look to check
whether this was already covered. As far as I can tell this code is in the
app-server protocol package rather than the orchestration layer, so it may
outlive V2 — but you'd know far better than I would.

## Workaround

None that I found short of patching. Avoiding large tool results (not using the
browser preview on content-heavy pages) keeps lines small enough to stay clear
of it.

clintebbesen:
Fixes #5389.

The Codex app-server protocol reader repeatedly concatenated and rescanned an unfinished JSONL record for every incoming chunk. Large records therefore incurred quadratic copying and could exhaust CPU and memory.

This change adds one incremental line-framing owner that retains unfinished records as fragments and joins them only at a newline or EOF. Both protocol stdin and Codex stderr now use it while preserving existing newline, CRLF, empty-line, parse-error, stream-failure, and EOF behavior.

Regression coverage includes a 20 MiB record split into irregular chunks, multiple records, CRLF, empty records, newline-boundary splits, final unterminated records, protocol EOF routing, and termination order. On the exact 20 MiB workload, the candidate completed in 7–13 ms versus 8,105 ms for the prior accumulator; the test rejects implementations exceeding 2,000 ms. Focused tests, package/server typechecks, targeted lint/formatting, and done-check passed.

Fork coordination: https://github.com/clintebbesen/t3code/issues/2

Verification environment: Codex harness, gpt-5.6-sol.

---

> [!NOTE]
> **Medium Risk**
> Changes the Codex app-server JSONL read path and stderr framing; behavior is intended to match prior semantics but mistakes could drop or mis-parse wire messages.
> 
> **Overview**
> Fixes pathological CPU/memory use when Codex sends **large JSONL records** split across many stdin chunks. The old reader kept **re-concatenating and re-scanning** the whole unfinished line on every chunk (quadratic work); very large lines could stall or exhaust resources.
> 
> This PR adds **`makeLineFramer`** in `effect-codex-app-server`, exported as `./lineFramer`. It buffers incomplete lines as **fragments**, emits only on `\n`, strips trailing `\r`, and exposes **`finish()`** for a final unterminated line at EOF. **Protocol stdin** and **Codex session stderr** both use it instead of local `remainder` + `split("\n")` logic.
> 
> Regression tests cover LF/CRLF, empty lines, arbitrary chunk boundaries, a **20 MiB** fragmented record (must complete under 2s), CRLF messages split across chunks, and routing the **last line without a trailing newline** before input stream end (notification before termination).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3de1fe92b989e804a540725222f9c32641943217. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Replace ad-hoc line buffering with `makeLineFramer` in codex app-server protocol and runtime
> - Introduces [`makeLineFramer`](https://github.com/pingdotgg/t3code/pull/7068/files#diff-e5e4811ebd80c4f45e1712c3557fae7c295dd5628a146ceffe48c8dd633b79e5), a reusable framer that incrementally buffers string chunks and emits complete lines split on `\n` with trailing `\r` trimmed.
> - Replaces `Ref`-based remainder accumulators in both [`protocol.ts`](https://github.com/pingdotgg/t3code/pull/7068/files#diff-c2eebe727f23e219c63497172bee07fd14e21634903ec9de0e7d63bfc80571fd) and [`CodexSessionRuntime.ts`](https://github.com/pingdotgg/t3code/pull/7068/files#diff-e29bf409c8e737a2d4e655f7e9e7def9c78c008d98a0a23575baa5e714c4e12c) with `lineFramer.push()` / `lineFramer.finish()` calls.
> - Exports the new module as the `effect-codex-app-server/lineFramer` subpath in [`package.json`](https://github.com/pingdotgg/t3code/pull/7068/files#diff-e02c7fc3e76b0053bbf910f67f429902bbcbaaaf7785daf78c787889f40a3172).
> - Adds tests covering LF, CRLF, empty lines, unterminated final lines, and large fragmented lines.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3de1fe9.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/CodexSessionRuntime.ts
+++ b/apps/server/src/provider/Layers/CodexSessionRuntime.ts
@@ -32,6 +32,7 @@ import * as Stream from "effect/Stream";
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process";
 import * as CodexClient from "effect-codex-app-server/client";
 import * as CodexErrors from "effect-codex-app-server/errors";
+import { makeLineFramer } from "effect-codex-app-server/lineFramer";
 import * as CodexRpc from "effect-codex-app-server/rpc";
 import * as EffectCodexSchema from "effect-codex-app-server/schema";
 
@@ -1621,34 +1622,25 @@ export const makeCodexSessionRuntime = (
       Effect.forkIn(runtimeScope),
     );
 
-    const stderrRemainderRef = yield* Ref.make("");
+    const stderrLineFramer = makeLineFramer();
     yield* child.stderr.pipe(
       Stream.decodeText(),
       Stream.runForEach((chunk) =>
-        Ref.modify(stderrRemainderRef, (current) => {
-          const combined = current + chunk;
-          const lines = combined.split("\n");
-          const remainder = lines.pop() ?? "";
-          return [lines.map((line) => line.replace(/\r$/, "")), remainder] as const;
-        }).pipe(
-          Effect.flatMap((lines) =>
-            Effect.forEach(
-              lines,
-              (line) => {
-                const classified = classifyCodexStderrLine(line);
-                if (!classified) {
-                  return Effect.void;
-                }
-                return emitEvent({
-                  kind: "notification",
-                  threadId: options.threadId,
-                  method: "process/stderr",
-                  message: classified.message,
-                });
-              },
-              { discard: true },
-            ),
-          ),
+        Effect.forEach(
+          stderrLineFramer.push(chunk),
+          (line) => {
+            const classified = classifyCodexStderrLine(line);
+            if (!classified) {
+              return Effect.void;
+            }
+            return emitEvent({
+              kind: "notification",
+              threadId: options.threadId,
+              method: "process/stderr",
+              message: classified.message,
+            });
+          },
+          { discard: true },
         ),
       ),
       Effect.forkIn(runtimeScope),
--- a/packages/effect-codex-app-server/package.json
+++ b/packages/effect-codex-app-server/package.json
@@ -19,6 +19,10 @@
       "types": "./src/protocol.ts",
       "import": "./src/protocol.ts"
     },
+    "./lineFramer": {
+      "types": "./src/lineFramer.ts",
+      "import": "./src/lineFramer.ts"
+    },
     "./errors": {
       "types": "./src/errors.ts",
       "import": "./src/errors.ts"
--- a/packages/effect-codex-app-server/src/lineFramer.test.ts
+++ b/packages/effect-codex-app-server/src/lineFramer.test.ts
@@ -1,0 +1,56 @@
+import { assert, it } from "@effect/vitest";
+
+import { makeLineFramer } from "./lineFramer.ts";
+
+const collectLines = (chunks: ReadonlyArray<string>) => {
+  const framer = makeLineFramer();
+  const lines = chunks.flatMap((chunk) => framer.push(chunk));
+  const finalLine = framer.finish();
+  return finalLine === undefined ? lines : [...lines, finalLine];
+};
+
+it("preserves LF, CRLF, empty-line, and unterminated-line semantics across chunks", () => {
+  assert.deepEqual(collectLines(["alpha\r", "\n", "\nb", "eta\n", "gamma", "\r", "\ndelta"]), [
+    "alpha",
+    "",
+    "beta",
+    "gamma",
+    "delta",
+  ]);
+});
+
+it("handles chunk boundaries immediately before, after, and within line endings", () => {
+  assert.deepEqual(collectLines(["one", "\n", "two\n", "\r", "\n", "three\r", "\nfour"]), [
+    "one",
+    "two",
+    "",
+    "three",
+    "four",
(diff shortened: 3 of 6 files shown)


t3dotgg:
> [!NOTE]
>

Closing as superseded by [#8605](https://github.com/pingdotgg/t3code/pull/8605), which is on main. The protocol reader now stores partial records as fragments and joins them at a newline or EOF, removing the repeated copying behind the reported failure. The separate stderr optimization and extra framing cases remain available in this branch.

PR #7094, opened 2026-08-15
fix(web): remember the right panel width per thread

### LINKED ISSUE #5548: [Bug]: Right panel width is shared globally instead of saved per thread
## Problem

The resizable right panel uses one shared width across all threads. Resizing it while viewing one thread immediately changes the panel width used by every other thread.

This is awkward when different threads need different layouts—for example, one thread may need a wide browser or diff view while another only needs a narrow terminal or file panel.

## Steps to reproduce

1. Open the right panel in thread A and resize it to be narrow.
2. Switch to thread B and resize its right panel to be wide.
3. Switch back to thread A.

## Actual behavior

Thread A uses the latest width selected in thread B. The most recently selected width is shared globally.

## Expected behavior

Each thread should remember its own right-panel width. Returning to thread A should restore thread A's previous width, while thread B should retain its own width.

The width should remain persisted across application reloads, matching the existing persisted, thread-scoped right-panel surface state. New threads can continue to use the default width.

## Environment

- T3 Code desktop 0.0.31
- macOS

This likely affects the shared web client as well because the width is stored under one global browser-storage key rather than keyed by the scoped thread identity.

Rishet11:
## What Changed

Resizing the right panel in one thread resized it in every other thread. Make it narrow in thread A, switch to thread B and pull it wide, come back to A, and A is now wide too. Threads that want different layouts (a wide diff in one, a narrow terminal in another) fight over one width.

The panel width is persisted in browser storage, and the key had no thread in it, so every thread read and wrote the same entry.

The key is now scoped to the thread. Each thread remembers its own width, it survives a reload, and a thread that has never been resized still gets the default.

## Why

`PreviewPanelShell` already supports this. It takes an optional `widthStorageKey` and falls back to the shared one:

```ts
storageKey: props.widthStorageKey ?? PREVIEW_PANEL_WIDTH_STORAGE_KEY,
```

Its own comment explains the intent: "Callers embedding this shell for a different surface (e.g. the pull requests page) should pass their own key so resizing one panel doesn't clobber the other's remembered width." The pull requests page does exactly that with `widthStorageKey="t3code:pull-request-panel-width"`.

The chat view never passed one, so it kept the shared default. This passes a per-thread key, using `activeThreadKey`, the same scoped key the right panel already uses for its surface state (`byThreadKey`). That keeps width consistent with everything else remembered per thread.

The key builder is a small module of its own rather than an inline template string, so it can be unit tested. The panel itself needs a rendered chat view to exercise, and this directory already keeps its small logic in tested modules next to the components (`previewAutomationClientId.ts`, `fileExplorerLabel.ts`, `addBrowserSurface.ts`).

Only the inline panel is touched. The sheet presentation does not use a resizable width - `PreviewPanelShell` applies `width` only when `isInline && !maximized` - so there is nothing to scope there.

## UI Changes

n/a - no layout or styling change. The panel is the same panel at the same default width; it just stops overwriting other threads' widths.

## Checklist

- [x] I included before/after screenshots for any UI changes (n/a, see above)

## Verification

```
vp test run apps/web/src/components/preview/rightPanelWidthStorageKey.test.ts   # 3 passed
vp fmt --check <the three touched files>                                        # clean
```

The three tests cover what the report asks for: two threads get different keys, the same thread gets a stable key across calls (so the width survives a reload), and the key stays under the existing `t3code:preview-panel-width` prefix.

One thing I could not run, so I am flagging it rather than claiming it: a full `@t3tools/web` typecheck. My checkout is behind `main` on several unrelated files, so `tsgo` reports errors for symbols that exist upstream but not locally, none of them on the lines this PR changes. What I did verify against current `main`: `activeThreadKey` is in scope at the call site, `RightPanelTabs.tsx` is byte-identical to `main` so its `widthStorageKey?: string` prop is current, and `PreviewPanelShell.tsx` still reads `props.widthStorageKey ?? PREVIEW_PANEL_WIDTH_STORAGE_KEY`. The diff against `main` is 4 added lines in `ChatView.tsx` and one new module, with no deletions. CI will be the real check on the full typecheck.

Fixes #5548

---

> [!NOTE]
> **Low Risk**
> Small UI persistence change with no auth, data, or API impact; only affects which localStorage key stores panel width.
> 
> **Overview**
> **Fixes shared right-panel width across chat threads** by persisting resize width per thread instead of one global `localStorage` entry.
> 
> `ChatView` now passes `widthStorageKey: rightPanelWidthStorageKey(activeThreadKey)` into inline `RightPanelTabs`, which forwards it to `PreviewPanelShell` (same pattern as the pull-requests page). A small `rightPanelWidthStorageKey` helper builds keys like `t3code:preview-panel-width:<threadKey>`; unit tests cover per-thread isolation, stability on reload, and the storage prefix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 31f36345eb8951a8f205cca10ab131a5669eb332. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Remember right panel width per thread in ChatView
> The right panel width is now stored and restored per thread rather than globally. A new `rightPanelWidthStorageKey` utility generates a namespaced localStorage key (`t3code:preview-panel-width:<threadKey>`) for each thread, passed to `RightPanelTabs` via `widthStorageKey` when `activeThreadKey` is defined.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 31f3634. 2 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/apps/web/src/components/ChatView.tsx
+++ b/apps/web/src/components/ChatView.tsx
@@ -141,6 +141,7 @@ import { closePreviewSession } from "./preview/closePreviewSession";
 import { ThreadPreviewMiniPlayer } from "./preview/ThreadPreviewMiniPlayer";
 import { subscribePreviewAction } from "./preview/previewActionBus";
 import { getConfiguredPreviewUrls } from "./preview/previewEmptyStateLogic";
+import { rightPanelWidthStorageKey } from "./preview/rightPanelWidthStorageKey";
 import { makeWorkspaceFileDropHandlers } from "./chat/workspaceFileDrop";
 import {
   selectThreadPreviewMiniPlayer,
@@ -6623,6 +6624,9 @@ function ChatViewContent(props: ChatViewProps) {
         <RightPanelTabs
           mode="inline"
           maximized={rightPanelMaximized}
+          {...(activeThreadKey
+            ? { widthStorageKey: rightPanelWidthStorageKey(activeThreadKey) }
+            : {})}
           surfaces={rightPanelState.surfaces}
           activeSurfaceId={activeRightPanelSurface?.id ?? null}
           pendingSurfaceIds={pendingFileSurfaceIds}
--- a/apps/web/src/components/preview/rightPanelWidthStorageKey.test.ts
+++ b/apps/web/src/components/preview/rightPanelWidthStorageKey.test.ts
@@ -1,0 +1,24 @@
+import { describe, expect, it } from "vite-plus/test";
+
+import { rightPanelWidthStorageKey } from "./rightPanelWidthStorageKey";
+
+describe("rightPanelWidthStorageKey", () => {
+  it("gives two threads separate keys so one resize does not move the other", () => {
+    const threadA = rightPanelWidthStorageKey("local/thread-a");
+    const threadB = rightPanelWidthStorageKey("local/thread-b");
+
+    expect(threadA).not.toBe(threadB);
+  });
+
+  it("returns the same key for the same thread so the width survives a reload", () => {
+    expect(rightPanelWidthStorageKey("local/thread-a")).toBe(
+      rightPanelWidthStorageKey("local/thread-a"),
+    );
+  });
+
+  it("stays under the shared preview-panel prefix", () => {
+    expect(rightPanelWidthStorageKey("local/thread-a")).toBe(
+      "t3code:preview-panel-width:local/thread-a",
+    );
+  });
+});
--- a/apps/web/src/components/preview/rightPanelWidthStorageKey.ts
+++ b/apps/web/src/components/preview/rightPanelWidthStorageKey.ts
@@ -1,0 +1,8 @@
+/**
+ * Scopes the remembered right-panel width to one thread. Without a thread in the
+ * key every thread reads and writes the same width, so resizing the panel in one
+ * thread resizes it in all of them.
+ */
+export function rightPanelWidthStorageKey(threadKey: string): string {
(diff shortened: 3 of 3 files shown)


Rishet11:
Closing this one — the UI Consistency check is right and my fix does not actually work.

`useResizableWidth` only reads localStorage in its lazy `useState` initializer, and the panel is not remounted when you switch threads. So changing the key per thread does nothing on its own: the previous thread's width stays on screen, and the next drag saves it under the new thread's key. That is worse than the bug I was trying to fix.

The real fix has to re-seed the width inside `useResizableWidth` when its `storageKey` changes. I have that working locally, but it changes a hook the pull request panel and sidebar also use, and I could not find a way to regression test it — the web tests render with `renderToStaticMarkup`, which mounts once and cannot re-render with a different prop, so nothing would catch a regression here later.

There is also the second point from the check: namespacing the key drops the width people already have saved, so panels would reset to 540px once.

I would rather not land a change to a shared hook that I cannot test. Happy to reopen with the hook fix if you would prefer that, and if there is a rendering harness I missed for testing hooks, please point me at it and I will redo it properly.

Sorry for the noise.

PR #7123, opened 2026-08-15
fix(mobile): accept a pairing code however it was typed

### LINKED ISSUE #4889: [Bug]: Android manual pairing rejects formatted codes and can report a false timeout
### Before submitting

### Area

`apps/mobile` (not currently listed in the issue template's Area options)

### Steps to reproduce

1. Run a headless T3 server and expose it through Tailscale Serve HTTPS.
2. Create a one-time pairing credential. The CLI emits an uppercase, unhyphenated 12-character value.
3. In the official Android app, open **Add Environment**.
4. Enter the reachable HTTPS backend in **Host**.
5. Enter the credential in the hyphenated format suggested by the field placeholder (for example, `XXXX-XXXX-XXXX`).
6. Tap **Add environment**.
7. Retry with the raw unhyphenated code.
8. As a control, paste the complete `https://<host>/pair#token=<raw-token>` URL into **Host**, leave **Pairing code** empty, and submit.

### Expected behavior

The manual pairing-code path should work reliably.

Either:

- normalize presentation separators/whitespace and case before exchange, or
- show the exact server-generated format and clearly require it.

The surfaced error should also match the actual failed request.

### Actual behavior

- Hyphenated manual entry is sent verbatim and the server returns `401 auth_invalid`.
- A retry with raw code also failed during this reproduction.
- The app subsequently reported a timeout after `10000ms`, despite the corresponding server-side `/oauth/token` request returning `401` in approximately 2.7ms.
- Pasting the complete pairing URL into **Host** with **Pairing code** empty succeeded immediately.

The backend and transport were independently verified: environment discovery returned `200`, Tailscale connectivity was healthy, and a disposable server-generated credential exchanged successfully through the same public HTTPS `/oauth/token` endpoint.

### Impact

Major degradation or frequent failure.

The advertised manual host + pairing-code flow is unreliable. QR/full-URL pairing remains a workaround.

### Version or commit

- Server: `t3 v0.0.32-nightly.20260729.951`
- Mobile: official Android client installed from Google Play on 2026-07-29

### Environment

- Device: Pixel 9a
- Client OS: Android
- Server OS: Linux x64
- Transport: Tailscale Serve HTTPS on port 443
- Mobile HTTP user agent observed server-side: `okhttp/4.9.2`

### Logs or stack traces

```text
GET  /.well-known/t3/environment -> 200
POST /oauth/token                  -> 401 (about 2.7ms)
Mobile UI                          -> timed out after 10000ms

Control:
POST /oauth/token with a disposable server-generated credential through
the same Tailscale HTTPS endpoint -> 200
```

Secrets, pairing credentials, environment IDs, trace IDs, and Tailnet hostnames have been omitted.

### Relevant source observation

The Android pairing field currently shows the placeholder `abc-123-xyz`, while `buildPairingUrl()` only applies `.trim()` to the code before putting it into the URL fragment. It does not remove presentation separators or normalize case.

Relevant files:

- `apps/mobile/src/features/connection/ConnectionsNewRouteScreen.tsx`
- `apps/mobile/src/features/connection/pairing.ts`

### Workaround

Paste the complete server-generated `/pair#token=...` URL into the **Host** field, leave **Pairing code** empty, then add the environment. Scanning the server-generated QR code should follow the same successful path.

Rishet11:
## What Changed

Pairing codes are issued from `PAIRING_TOKEN_ALPHABET = "23456789ABCDEFGHJKLMNPQRSTUVWXYZ"` at 12 characters (`PairingGrantStore.ts:260-261`) — uppercase, no separators, no `0`/`1`/`O`/`I`. The code fields send whatever was typed straight through and the server compares it exactly, so `abc-123-xyz` comes back `401 auth_invalid`. The mobile field's own placeholder is `abc-123-xyz`, which invites precisely the format that fails.

A shared `normalizePairingCode` (strip whitespace and dashes, upper-case) is now applied at the two fields that take a pairing code and nothing else:

- `ConnectionsNewRouteScreen.tsx` — `handleCodeChange`, the mobile Add Environment field
- `ConnectionsSettings.tsx` — the Pairing code input in web/desktop Settings

Since the issued alphabet has no lowercase, dashes or spaces, normalizing cannot turn one valid code into another or accept anything the server would reject. It only repairs input that could never have been valid.

## Why these two call sites and not the shared parser

The tidy-looking place is `resolveRemotePairingTarget` — one function, every surface at once. That is wrong here, because **machine-issued tokens reach the same `pairingCode` field**, and those are not drawn from the pairing alphabet:

- `ConnectionsSettings.tsx` `handleSavedBackendHostChange` splits a pairing URL pasted into the *host* box and writes its token into the pairing-code state. On submit the host no longer parses as a URL, so it falls through the same branch.
- `parsePairingUrlFields` returns `pairingCode: hostedPairingRequest.token` for a hosted link, and hosted tokens are issued outside this repo.
- On mobile, `handleQrScan` writes the scanned token into `codeInput`, which submit then re-reads.

Both patched call sites are *typing* handlers, so a scan, a paste-split or a deep link never passes through them. `remote.test.ts` carries a guard for this: it asserts a mixed-case token survives `resolveRemotePairingTarget` untouched through both the `pairingUrl` and the `pairingCode` field, so a later refactor that "simplifies" this into the parser fails the suite instead of silently breaking hosted pairing.

## A third call site I added and reverted

Worth stating plainly, since the review history shows it and two resolved-by-deletion comments still point at the file. An earlier commit also normalized the credential field on the local `/pair` page. That was wrong and CI caught it: that field is not pairing-code-only — `describeSupportedMethods` advertises it for a one-time token **or** a desktop bootstrap credential, and the bootstrap credential is lowercase hex (`/^[0-9a-f]{48}$/i`, see `DesktopBackendConfiguration.test.ts:152`). Since `PairingGrantStore.consume` matches exactly, upper-casing would have made every pasted bootstrap credential fail authentication. The field is also explicitly `autoCapitalize="none"`.

It is gone as of `2ddc8f0`; `PairingRouteSurface.tsx` is untouched by this PR. Calling it out because "this input holds a credential" is not the same as "this input holds a pairing code", and that distinction is the whole reason the fix sits where it does.

## UI Changes

n/a — no layout or component change. The visible difference is that the two code fields now show the canonical form as you type: entering `abcd-2345-efgh` leaves `ABCD2345EFGH` in the box. Worth knowing when reviewing, since it is a live transform rather than a submit-time one — deliberate, so what the user sees is what gets sent.

## Checklist

## Verification

```
vp test run packages/shared/src/remote.test.ts                                  # 17 passed (15 before)
vp test run apps/mobile/src/features/connection/pairing.test.ts                 #  7 passed, unchanged
vp test run apps/web/src/components/settings/ConnectionsSettings.logic.test.ts  #  8 passed, unchanged
vp run --filter @t3tools/shared typecheck                                       # clean
vp run --filter @t3tools/mobile typecheck                                       # clean
vp run --filter @t3tools/web typecheck                                          # no errors in the touched files
vp lint / vp fmt --check on all five touched files                              # clean
```

Two new cases in `remote.test.ts`: one pins the repair across the forms a person actually types (`abcd2345efgh`, `ABCD-2345-EFGH`, `abcd 2345 efgh`, ` AbCd-2345 EfGh ` all give `ABCD2345EFGH`), the other is the anti-regression guard described above.

**What I could not verify.** The two call sites are one-line changes inside UI event handlers and there is no seam to unit test them through — `apps/mobile` has no React Native component test infrastructure, and `ConnectionsSettings.tsx` is a large component whose colocated test covers extracted logic rather than rendered input. So the normalizer is covered by tests and its wiring is not; the wiring is two one-liners and is best judged by reading them. I also have no Android device here, so the end-to-end pairing round trip is reasoned from the alphabet and the call sites rather than observed.

**Not addressed:** the issue's second claim, the false timeout. I could not tie the reported `10000ms` to any constant in the mobile connection path and did not want to guess. It may be the auth failure above surfacing late rather than an independent bug, but I have not established that, so this PR does not touch it and the issue should stay open for it.

Fixes #4889

### DIFF
--- a/apps/mobile/src/features/connection/ConnectionsNewRouteScreen.tsx
+++ b/apps/mobile/src/features/connection/ConnectionsNewRouteScreen.tsx
@@ -1,6 +1,7 @@
 import { CameraView, useCameraPermissions } from "expo-camera";
 import { NativeHeaderToolbar, NativeStackScreenOptions } from "../../native/StackHeader";
 import { StackActions, useNavigation, type StaticScreenProps } from "@react-navigation/native";
+import { normalizePairingCode } from "@t3tools/shared/remote";
 import { AsyncResult } from "effect/unstable/reactivity";
 import { useCallback, useEffect, useRef, useState } from "react";
 import { Alert, Linking, Platform, ScrollView, View } from "react-native";
@@ -78,7 +79,10 @@ export function ConnectionsNewRouteScreen({
   }, []);
 
   const handleCodeChange = useCallback((value: string) => {
-    setCodeInput(value);
+    // Only what the user types is repaired. A scanned or deep-linked code is
+    // written straight to state and never reaches this handler, which matters
+    // because those can carry a hosted token from a different alphabet.
+    setCodeInput(normalizePairingCode(value));
   }, []);
 
   const openScanner = useCallback(async () => {
--- a/apps/web/src/components/settings/ConnectionsSettings.tsx
+++ b/apps/web/src/components/settings/ConnectionsSettings.tsx
@@ -86,7 +86,11 @@ import { Button } from "../ui/button";
 import { Empty, EmptyDescription, EmptyHeader, EmptyMedia, EmptyTitle } from "../ui/empty";
 import { AnimatedHeight } from "../AnimatedHeight";
 import { Textarea } from "../ui/textarea";
-import { getPairingTokenFromUrl, setPairingTokenOnUrl } from "../../pairingUrl";
+import {
+  getPairingTokenFromUrl,
+  normalizePairingCode,
+  setPairingTokenOnUrl,
+} from "../../pairingUrl";
 import { readHostedPairingRequest } from "../../hostedPairing";
 import {
   createServerPairingCredential,
@@ -2423,7 +2427,9 @@ export function ConnectionsSettings() {
           <span className="mb-1.5 block text-xs font-medium text-foreground">Pairing code</span>
           <Input
             value={savedBackendPairingCode}
-            onChange={(event) => setSavedBackendPairingCode(event.target.value)}
+            onChange={(event) =>
+              setSavedBackendPairingCode(normalizePairingCode(event.target.value))
+            }
             placeholder="PAIRCODE"
             disabled={isAddingSavedBackend}
             spellCheck={false}
--- a/apps/web/src/pairingUrl.ts
+++ b/apps/web/src/pairingUrl.ts
@@ -1,5 +1,6 @@
 export {
   getPairingTokenFromUrl,
+  normalizePairingCode,
   setPairingTokenOnUrl,
   stripPairingTokenFromUrl,
 } from "@t3tools/shared/remote";
--- a/packages/shared/src/remote.test.ts
+++ b/packages/shared/src/remote.test.ts
@@ -5,6 +5,7 @@ import {
   RemoteBackendUrlMissingError,
   RemotePairingTokenMissingError,
   RemotePairingUrlInvalidError,
+  normalizePairingCode,
   resolveRemotePairingTarget,
 } from "./remote.ts";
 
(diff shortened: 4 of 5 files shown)


Rishet11:
Both findings are the same real bug and they are right — fixed in `2ddc8f0` by dropping that call site.

The `/pair` credential field is not a pairing-code-only input. `describeSupportedMethods` advertises it for a one-time token **or** a desktop bootstrap credential, and the bootstrap credential is lowercase hex (`/^[0-9a-f]{48}$/i`, per `DesktopBackendConfiguration.test.ts:152`). `PairingGrantStore.consume` matches exactly, so upper-casing a pasted bootstrap credential would have made it fail with "Authentication failed." every time. The field is also explicitly `autoCapitalize="none"`, which was the hint I walked past.

That is precisely the case the helper's own doc comment tells callers to stay out of, and I put a call site there anyway. It came from a review suggestion that the local `/pair` page was a missed surface for this bug; it is a different field taking a different credential type, and I should have checked what it accepts before wiring it up rather than trusting the shape of the fix.

Normalization now applies only to the two code-only fields:

- `ConnectionsSettings.tsx` — the Settings "Pairing code" input
- `ConnectionsNewRouteScreen.tsx` — `handleCodeChange`, the mobile Add Environment field

Both take only a code issued from `PAIRING_TOKEN_ALPHABET`, and both are typing handlers, so a scanned QR, a pasted-and-split pairing URL, or a deep link never passes through them.

Verification on this commit:

```
vp test run packages/shared/src/remote.test.ts     # 17 passed (17)
vp run --filter @t3tools/shared typecheck          # clean
vp run --filter @t3tools/mobile typecheck          # clean
vp run --filter @t3tools/web typecheck             # no errors in the touched files
vp fmt --check on all five touched files           # clean
```

The `remote.test.ts` guard against normalizing tokens that arrive through `resolveRemotePairingTarget` is unchanged and still passing — it just did not cover this field, since `/pair` submits through `submitServerAuthCredential` rather than that function.

shivamhwp:

The body explicitly leaves the false timeout from #4889 unresolved but still ends with `Fixes #4889`. Please replace that closing keyword with a related-issue reference so landing the code normalization does not close the remaining timeout report.

PR #7125, opened 2026-08-15
fix(desktop): stop overwriting a custom dock icon on launch

### LINKED ISSUE #7117: [Bug]: setDockIcon() overrides the user's custom dock icon on launch
### Before submitting

### Area

apps/desktop

### Steps to reproduce

1. In Finder, select the T3 Code app and Press Cmd+I.
2. Drag a PNG onto the icon in the Get Info window.
3. Confirm that the app in Finder and in the Dock have the new Icon
4. Launch T3 Code

### Expected behavior

T3 Code keeps the custom app icon in the dock when launched.

### Actual behavior

T3 Code overwrites the custom app icon in the dock when launched when launched, replacing it with its default app icon.

### Impact

Cosmetic issue

### Version or commit

v0.0.33

### Environment

macOS 15.7.5

### Logs or stack traces

```shell

```

_No response_

### Workaround

_No response_

Rishet11:
## What Changed

`DesktopAppIdentity.configure` called `setDockIcon` on every macOS launch:

```ts
if (environment.platform === "darwin") {
  const iconPaths = yield* assets.iconPaths;
  yield* Option.match(iconPaths.png, { onNone: () => Effect.void, onSome: electronApp.setDockIcon });
}
```

It now runs only when the app is unpackaged.

## Why

In a packaged build `resolveSourceTreeIconPath` returns `undefined` (`DesktopAssets.ts:81` bails on `environment.isPackaged`), so `resolveIconPath("png")` falls through to `resolveResourcePath("icon.png")` and finds the icon bundled in `Contents/Resources`. `iconPaths.png` is therefore always `Some` in a packaged build, and `app.dock.setIcon` fires on every launch.

That call has nothing to add there. A packaged bundle already declares its icon through `Info.plist`, so the dock shows the right artwork without it. What the call does do is overwrite the dock tile at runtime, which discards a custom icon the user attached to the app themselves via Finder's Get Info — the symptom in the issue, and it comes back after every relaunch.

Unpackaged runs still need it: Electron shows a generic icon for a plain `electron .` run, which is what the source-tree branch of `resolveIconPath` exists to feed. That path is untouched.

## UI Changes

n/a — no interface change. The visible effect is on the macOS dock tile: a user's custom icon now survives a relaunch of a packaged build.

## Checklist

## Verification

```
vp test run apps/desktop/src/app/DesktopAppIdentity.test.ts   # 4 passed (3 before)
vp lint / vp fmt --check on both touched files                # clean
```

`DesktopAppIdentity.test.ts` already drives `configure` through fake `ElectronApp` / `DesktopAssets` / `DesktopEnvironment` layers, so this needs no real Electron instance. Two changes there:

- the existing `configures app identity from the environment commit override` case runs under the default `isPackaged: true` environment and asserted `setDockIcon` had been called with `/icon.png`. That assertion encoded the bug, so it now asserts no call was made.
- a new case pins the behaviour that must stay: with `isPackaged: false`, `setDockIcon` is still called with the resolved png.

Reverting the source change fails the first with `expected [ '/icon.png' ] to deeply equal []`.

**On the flipped assertion.** I want to flag that plainly rather than bury it, since changing an existing expectation is how a real intention gets erased. I read it as describing what the code did rather than asserting a packaged build should re-set its own icon: the same `environment.isPackaged` gate is already used a few lines up in this file, `DesktopAssets` deliberately excludes `icns` and all source-tree lookups from packaged builds, and no comment or changelog ties the dock call to packaged runs. If it was in fact deliberate — a branded dock tile for nightly against stable, say — then this PR is wrong and I would rather hear that than have it merged.

**Not verified:** I cannot build and run a packaged macOS bundle here, so "the bundle icon still shows once we stop calling `setDockIcon`" rests on `Info.plist`/`CFBundleIconFile` behaviour and the asset-resolution code above, not on something I watched happen. Worth a reviewer with a packaged build confirming the dock still looks right on a clean install.

Fixes #7117

---

> [!NOTE]
> **Low Risk**
> Small, platform-specific startup behavior change with explicit tests; no auth, data, or security surface.
> 
> **Overview**
> **Packaged macOS builds no longer call `app.dock.setIcon` on startup**, so a custom dock icon set via Finder survives relaunch. The dock icon is still set during **unpackaged** dev runs, where Electron would otherwise show a generic tile.
> 
> `DesktopAppIdentity.configure` now gates the Darwin `setDockIcon` path with `!environment.isPackaged`, matching how bundled icons are already declared in `Info.plist`. Tests were updated to expect no dock call when packaged (default test env) and to assert unpackaged runs still pass the resolved PNG path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b29d47e9b87c7a3061ddf294e097521811fc40f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Stop overwriting a custom dock icon for packaged desktop builds on macOS
> Previously, `DesktopAppIdentity.configure` called `setDockIcon` on all macOS builds, overwriting the bundle's native icon. Now `setDockIcon` is only called when the app is running unpackaged and a PNG icon path is available.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6b29d47.</sup>
> 
>


Rishet11:
Can I get merge please if everything is perfect

Issue #7113, opened 2026-08-15
[Bug]: OpenCodeAdapter leaks pending permission requests on session teardown — thread becomes permanently unsettleable


simonechecchia:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

### Summary

`OpenCodeAdapter` never settles its pending permission requests when a session tears down. Any approval still open when the session stops is stranded in `projection_pending_approvals` with `status = 'pending'` forever, so `projection_threads.pending_approval_count` stays above zero and the thread can never be settled:

> This thread still needs attention. Resolve or interrupt it first, then try again.

There is nothing left to resolve. The session is `stopped`, the turns that requested the approvals are long dead, and no UI surface can clear them — the only way out is hand-editing `state.sqlite`.

This is the OpenCode counterpart of #5119 (`ClaudeAdapter` leaking pending *user inputs* on teardown), which was fixed. The same parity gap exists here for *approvals*: `CursorAdapter` and `GrokAdapter` both call `settlePendingApprovalsAsCancelled` on teardown, and `OpenCodeAdapter` has no equivalent.

It is made much easier to hit by #4795 — permissions that map to `requestType: "unknown"` render no approval UI, so the user never sees the request in the first place and the turn dies with it still open. But the leak is a separate defect: it strands *any* pending permission on teardown, and fixing the UI mapping alone would not clear the rows that are already stuck.

### Steps to reproduce

1. Start a thread on the OpenCode provider in **auto** runtime mode.
2. Get the agent to use a tool whose permission kind is outside `bash` / `read` / `edit` — `grep` and `task` both do it. In auto mode these are guaranteed to prompt (see the permission arrays below), and none of them render an approval UI (#4795), so the request opens invisibly and is never answered.
3. Let the session end — the turn erroring out does it, and so does interrupting or a server restart.
4. Try to settle the thread → refused with the message above, with no pending approval visible anywhere in the UI.

Auto mode is what makes this reliably reproducible. The permission array T3 sends to OpenCode differs by runtime mode, and only `full-access` appends a catch-all allow:

```jsonc
// auto — ends here, so the leading `*: ask` governs every unlisted permission kind
[{"permission":"*","pattern":"*","action":"ask"}, ..., {"permission":"question","pattern":"*","action":"allow"}]

// full-access — same array plus a trailing catch-all
[...]

Issue #5454, opened 2026-08-05
[Bug]: Pending user-input popup can never be dismissed after the provider session dies (e.g. server restart)


NeilTheFisher:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/web + apps/server

### Summary

If the provider session for a thread dies while a `request_user_input` / `AskUserQuestion` popup is pending — e.g. the user restarts the server, or the provider process exits — the popup becomes permanently stuck:

- **The message can't be submitted.** Submitting appends a `provider.user-input.respond.failed` activity whose detail is `"No active provider session is bound to this thread."`, but the box stays open. Every submit just adds another failure.
- **The popup can never be dismissed.** The client only closes a pending user-input request on a `user-input.resolved` activity or when a failure detail matches the "stale pending request" pattern. That detail does not match, so the request stays open indefinitely.

Related but distinct issues: #4060 (desktop, stuck on user input GUI) and #5119 (Claude adapter leaking pending requests on teardown). This one is provider-agnostic and also triggers on a plain server restart.

### Steps to reproduce

1. Have an agent call `request_user_input` so the question card renders in the composer.
2. Restart the T3 server (or let the provider process exit) while the card is pending — the thread session becomes `stopped`.
3. Type an answer and submit → error `No active provider session is bound to this thread.`, and the card stays open.
4. The card can never be closed, and the thread stays unsettleable.

### Expected behavior

Once the provider session is gone, the pending request is unanswerable and should be treated as terminal (like the existing `Unknown pending user-input request` stale path), so the popup closes and the thread can be settled or restarted.

### Fix

PR https://github.com/pingdotgg/t3code/pull/5453 treats the `"No active provider session is bound to this thread."` respond-failure detail as a stale pending request in the matchers shared by the web, mobile, server decider, and projection pipeline.

kevinavos:
Hit the same stuck-composer failure on a **pending approval** (not user-input) with the **Cursor** provider.

### What happened

T3 Code 0.0.33 on Linux. Thread was Cursor `auto-smart` in Full access. An MCP approval for `t3-code-preview_navigate` stayed in the composer after the Cursor ACP session died (`provider.stopAll` on server restart; projected session `stopped`).

Cancel / Approve both fail with:

```text
Provider approval response failed
No active provider session is bound to this thread.
```

[...]

Issue #5119, opened 2026-07-31
[Bug]: ClaudeAdapter leaks pending AskUserQuestion requests on session teardown — thread becomes unanswerable and unsettleable


AaronAbuUsama:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

### Summary

If a Claude session dies while an `AskUserQuestion` card is pending — mine died from running out of inference credits mid-turn — the thread deadlocks permanently:

- **The card can't be answered.** Every submit fails with `No active provider session is bound to this thread.`
- **Switching the model doesn't help.** Same error.
- **The thread can't be settled or snoozed.** `canSettle` refuses while `hasPendingUserInput` is true, and nothing left alive can clear it.

**Cause: `ClaudeAdapter.stopSessionInternal` settles pending approvals on teardown but not pending user inputs.** The other adapters settle both.

### Steps to reproduce

1. Start a thread on the Claude provider.
2. Get the agent to call `AskUserQuestion` so the card renders in the composer.
3. Kill the session while the card is pending — running out of credits does it; `adapter.stopSession` reproduces it directly.
4. Answer the card → error. Switch model, answer again → same error. Try to settle the thread → refused.

### Expected behavior

Killing a session settles its pending user-input requests, the same way it already settles pending approvals. The card resolves, the thread unblocks, and a normal message restarts the session on the selected model.

### Actual behavior

The request stays open forever. It cannot be answered, dismissed, or settled around.

### Impact

Blocks work completely

### Version or commit

0.0.31 (e6987965f659) — T3 Code (Alpha), desktop

### Environment

macOS 26.1 (25B5042k), Apple M1 Pro (arm64), T3 Code (Alpha) desktop 0.0.31, provider: Claude / Fable, trigger: inference credits exhausted mid-turn

### Logs or stack traces

```shell
Provider user input response failed
No active provider session is bound to this thread.
```

### Root cause

`apps/server/src/provider/Layers/ClaudeAdapter.ts:3029-3055` — teardown drains one map and not the other:

```ts
for (const [requestId, pending] of context.pendingApprovals) {
  yield* Deferred.succeed(pending.decision, "cancel");
  // ... emits request.resolved
}
context.pendingApprovals.clear();

// context.pendingUserInputs (declared :191, populated :3268) is never touched.
```

[...]

PR #7131, opened 2026-08-15
fix(opencode): settle pending questions before teardown

inayayousfi:
Written by inayayousfi, typed by openai/gpt-5.6-sol running in OpenCode.
Every call here is inayayousfi's, and no agent acted on its own.

## Problem

OpenCode keeps unanswered question requests in memory when a turn or session stops. The client continues rendering that request after the provider can no longer answer it, which leaves the composer stuck in question mode.

## Fix

Reject pending OpenCode questions before interrupting or stopping a session, then emit the existing resolved event and clear the local request. The same cleanup now runs during session replacement, bulk shutdown, and adapter shutdown.

Provider rejection gets a five-second budget. A failure or timeout is logged, but teardown continues so the Stop action cannot hang.

## Related work

Related to #5454, which covers pending user input after a provider session dies.

This addresses the pending-question counterpart noted in #7113. Pending OpenCode permission cleanup remains outside this pull request.

This also complements #5453. That pull request recovers requests after a dead-provider response fails, while this change prevents OpenCode from leaving the question open during teardown.

The Claude counterpart was reported in #5119.

## Testing

- `vp test run apps/server/src/provider/Layers/OpenCodeAdapter.test.ts` (37 tests passed)
- `vp fmt --check apps/server/src/provider/Layers/OpenCodeAdapter.ts apps/server/src/provider/Layers/OpenCodeAdapter.test.ts`
- `git diff --check`

OpenCode may still publish its native rejection event after local resolution. That can produce a duplicate resolved activity, but pending state remains closed.

> [!NOTE]
> ### Settle pending questions before session teardown in OpenCodeAdapter
> - Adds `OpenCodeQuestionState` to track pending and in-flight questions per session, replacing the old flat `pendingQuestions` map.
> - `interruptTurn`, `stopSession`, `stopAll`, and the layer finalizer now call a new `settlePendingQuestions` step that rejects pending questions (with a 5-second timeout) and emits `user-input.resolved` before aborting.
> - `respondToUserInput` atomically claims a question and moves it to `replyingQuestions`; it fails if settlement is active or the question is unknown, and restores the question on reply failure.
> - New questions are silently dropped when settlement is in progress (`settlementCount > 0`).
> - Behavioral Change: sessions being replaced or stopped now emit `user-input.resolved` for all outstanding questions before the remote abort, and `session.exited` is only emitted when the session entry is actually removed from the map.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 073397c.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Touches provider lifecycle, user-input events, and concurrent interrupt/stop/reply paths; behavior is heavily tested but wrong settlement ordering could still leave stale UI or duplicate resolved events.
> 
> **Overview**
> Fixes the composer staying in question mode when a turn or session ends while OpenCode still has an unanswered question.
> 
> **Pending question settlement** runs before `session.abort` on interrupt, stop, bulk shutdown, session replacement, and adapter shutdown. It calls `question.reject` (5s timeout, failures logged) and emits `user-input.resolved` so local state closes even if the provider is gone.
> 
> **Concurrency** moves question tracking into a `Ref` with pending vs in-flight replies and a `settlementCount` gate: new questions are ignored during teardown, replies are blocked while settlement runs, and in-flight replies are not double-rejected on interrupt.
> 
> **Teardown** wraps `stopOpenCodeContext` in an uninterruptible mask with injectable pre-abort work, and tightens session-map removal so a slow stop cannot drop a replacement session.
> 
> Tests cover interrupt/stop ordering, reject failure/timeout, reply vs interrupt races, late events during settlement, and concurrent interrupts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 073397ccb4cb8b8ff2f879630dbcc6410788c1c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/server/src/provider/Layers/OpenCodeAdapter.test.ts
+++ b/apps/server/src/provider/Layers/OpenCodeAdapter.test.ts
@@ -64,9 +64,18 @@ const runtimeMock = {
     authHeaders: [] as Array<string | null>,
     abortCalls: [] as string[],
     questionRejectCalls: [] as string[],
+    questionReplyCalls: [] as string[],
     lifecycleCalls: [] as string[],
     questionRejectError: null as Error | null,
     questionRejectHangs: false,
+    questionReplyHangs: false,
+    lateSubscribedEvent: null as unknown,
+    releaseLateSubscribedEvent: null as (() => void) | null,
+    onLateSubscribedEventWait: null as (() => void) | null,
+    onLateSubscribedEventConsumed: null as (() => void) | null,
+    onQuestionReject: null as (() => void) | null,
+    onQuestionReply: null as (() => void) | null,
+    releaseLateEventOnQuestionReject: true,
     closeCalls: [] as string[],
     revertCalls: [] as Array<{ sessionID: string; messageID?: string }>,
     promptCalls: [] as Array<unknown>,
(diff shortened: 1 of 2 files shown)


t3dotgg:
> [!NOTE]

We are keeping OPEN [#7173](https://github.com/pingdotgg/t3code/pull/7173) as the review path for settling pending OpenCode interactions during teardown. We do not need a second branch for the same defect.

Issue #6418, opened 2026-08-13
[Bug]: Absolute Windows file links render without a usable destination


teloverge:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/web

### Steps to reproduce

On Windows, create one file outside the active workspace:

```
New-Item -ItemType Directory -Force D:\tmp\t3-link-repro
Set-Content -LiteralPath D:\tmp\t3-link-repro\example.md -Value '# Link target'
```

Then render in a thread message with direct Windows drive destination:

    [Open the example](D:/tmp/t3-link-repro/example.md)
  
Bare Windows path, matching output that agents sometimes produce:

    M:\batches\issue-40-v1\docs\prompt.md
  
Repeat with a path under the active workspace so both sides of the workspace boundary are covered.

Similar to: #6360 desktop thread file links do nothing instead of opening the file panel
but that is a macOS bug with different file paths. However, they may share same cause.

### Expected behavior

Windows drive paths should use the existing chat file-link behavior:

- A file inside the active workspace opens in T3 Code's file panel.
- A file outside the active workspace falls back to the configured editor.
- The shortened chip label may be used for display, but the tooltip and Copy full path action retain the complete resolved path.

### Actual behavior

There are two related failures:

  - A direct Markdown destination such as D:/tmp/t3-link-repro\example.md renders with link styling but has no usable href.
  - A bare path such as M:\batches\...\prompt.md remains plain text and is not interactive.

This reproduces for paths inside and outside the active project/worktree. The raw provider transcript and persisted message contain the complete paths, so the paths are not truncated by the provider, event store, or projection.

The direct-drive case is lost between Markdown parsing and the custom anchor renderer. rehype-sanitize interprets D: as a URL protocol and removes the href before React Markdown calls urlTransform. Calling rewriteMarkdownFileUriHref() only from urlTransform is therefore too late.

The bare-path case has a separate trigger: GFM autolinking does not recognize filesystem paths, so no link node is created at all.

T3 already contains code paths that can handle both targets once the message author uses a specific syntax:

  - file:///D:/... survives sanitization and is later handled by rewriteMarkdownFileUriHref().
  - An absolute Windows path inside an inline-code span is recognized by the existing inline file-link resolver.

[...]

Issue #5158, opened 2026-08-01
[Bug]: Markdown file links break when the destination uses angle brackets or contains parentheses


hey-jj:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/web

### Steps to reproduce

1. Create the target files:

   ```
   mkdir -p "/tmp/link repro" && touch "/tmp/link repro/manifest.tsv" "/tmp/a(1).txt"
   ```

2. Open any thread and send a message containing exactly these three lines:

   ```
   [angle](</tmp/link repro/manifest.tsv>)
   [parens](/tmp/a(1).txt)
   [control](/tmp/link%20repro/manifest.tsv)
   ```

3. Click each link in the sent message.

Sent user messages render through the same ChatMarkdown component as agent replies, so the repro needs no agent output. The first and third lines point at the same file and differ only in how the destination is written.

### Expected behavior

All three lines render as file link chips. Clicking each one opens the file.

### Actual behavior

The `angle` and `parens` lines render as plain underlined anchors. Clicking them does nothing. The `control` line renders as a file link chip and opens the file.

File link classification runs a separate regex over the raw message text, and that regex accepts fewer destination forms than the markdown parser does. `MARKDOWN_LINK_HREF_PATTERN` at `apps/web/src/components/ChatMarkdown.tsx:750` is:

```
/\[[^\]]*]\(([^)\s]+)(?:\s+["'][^"']*["'])?\)/g
```

The capture group has no angle-bracket branch and stops at the first `)`. For `[angle](</tmp/link repro/manifest.tsv>)` it extracts nothing. For `[parens](/tmp/a(1).txt)` it extracts the truncated `/tmp/a(1`. CommonMark accepts both forms, angle brackets permit spaces in the destination and bare destinations may contain balanced parentheses, so react-markdown renders the links with hrefs `/tmp/link%20repro/manifest.tsv` and `/tmp/a(1).txt`. Neither href has an entry in `markdownFileLinkMetaByHref` (built at `ChatMarkdown.tsx:1279`), the lookup at `ChatMarkdown.tsx:1453` misses, and the `!fileLinkMeta` branch renders a plain anchor with `target="_blank"`. Parenthesized names matter in practice because duplicate downloads produce files like `report (1).pdf`, and the paths agents emit routinely contain spaces.

### Impact

Minor bug or occasional failure

### Version or commit

[...]

PR #7189, opened 2026-08-16
fix(web): keep windows drive paths clickable in chat markdown

CDVolvik:
A markdown link whose destination is an absolute Windows path renders as styled text with nothing behind it (#6418).

## What changed

Both chat render paths drop the destination before the anchor can use it:

- **Assistant messages** leave `parseRawHtml` at its default of `true`, so `rehype-sanitize` runs, reads `D:` as an unknown protocol, and removes the href — before `urlTransform` is reached.
- **User messages** pass `false`, skip the sanitizer, and React Markdown's `defaultUrlTransform` blanks the same destination for the same reason.

Either way the anchor arrives without an href, so it cannot match an entry in `markdownFileLinkMetaByHref` even though that map has already resolved the path from the raw message text.

Both pipelines run remark, so the destination is normalized there: `toFilesystemLinkUrl` maps a drive path to the `file:` URL the sanitize schema already allows, and the existing `rewriteMarkdownFileUriHref` maps it back on the way out. `resolveMarkdownFileLinkTarget` already handled these paths, so this is only about getting them that far.

Two smaller pieces come with it:

- Link keys go through the same function, so `M:\dir\file.md` in the source and the `M:/dir/file.md` the renderer receives resolve to one key rather than missing each other on the separator.
- The anchor resolves a drive destination itself when the metadata map misses. That map is built by scanning the message text for inline links with a plain destination, so balanced parentheses, nested brackets and bracketed destinations with spaces are absent from it; without the fallback those would render as ordinary anchors pointing at a drive path with no file action.

`CHAT_MARKDOWN_SANITIZE_SCHEMA` is exported so the tests can render both pipelines with the real configuration; that needs an explicit type annotation because its inferred type is not nameable outside the module.

## Before / after

Before and after rendering of a Windows drive markdown link

Rendered from the real `ChatMarkdown` component with the app's built stylesheet. "Before" is the same component with this PR's source hunks reverted to `origin/main`. The file-type glyph inside the chip is a runtime SVG sprite the offline render does not load, so it shows as blank space.

Emitted markup for `See [Open](D:/tmp/example.md) for the notes.`:

| pipeline | before | after |
| --- | --- | --- |
| assistant (`parseRawHtml: true`) | `<a target="_blank" rel="noopener noreferrer">Open</a>` | `<a href="D:/tmp/example.md" class="… chat-markdown-file-link">… example.md</a>` |
| user (`parseRawHtml: false`) | `<a href="" target="_blank" rel="noopener noreferrer">Open</a>` | same as above |

## Scope

Markdown destinations only. A bare `M:\dir\file.md` written as plain text is still not linked, because GFM does not autolink filesystem paths and adding that is a separate change. If you read #6418 to include the bare-path case, this does not close it outright.

## Testing

- `apps/web` unit project: **2574 tests across 262 files pass**
- **63 tests** in the three touched files, including new tests that render both pipelines through `renderToStaticMarkup` with the real sanitize schema, and that assert an unsafe scheme is still dropped
- `tsgo --noEmit` clean; `vp lint` 0 warnings / 0 errors on the changed files; `vp fmt --check` clean
- Rebased on `origin/main` @ `4cb676cc1`

---

> [!NOTE]
> **Low Risk**
> Localized markdown rendering and URL handling for Windows drive paths; unsafe schemes remain rejected and behavior is covered by pipeline tests for both assistant and user render modes.
> 
> **Overview**
> Fixes **#6418**: markdown links like `[Open](D:/tmp/example.md)` used to render without a usable `href` because `D:` was treated as an unknown protocol—**assistant** messages lost it in `rehype-sanitize`, and **user** messages got a blank href from `defaultUrlTransform`.
> 
> A new **`remarkFilesystemLinkDestinations`** plugin runs on both chat pipelines and rewrites drive (and reference-definition) destinations to allowed **`file:`** URLs; existing **`rewriteMarkdownFileUriHref`** still turns them back into plain paths on output. **`toFilesystemLinkUrl`** centralizes that mapping and normalizes `\` vs `/` so metadata keys match what the renderer sees.
> 
> **`ChatMarkdown`** wires the plugin in, keys link metadata through the same helper, and **falls back** to `resolveMarkdownFileLinkMeta` at render time when the message-text regex scan missed the destination (parentheses, spaced `<...>` paths, etc.). **`CHAT_MARKDOWN_SANITIZE_SCHEMA`** is exported (with an explicit type) so tests can mirror both `parseRawHtml` paths; new unit tests assert drive links work, `javascript:` stays dropped, and https links are unchanged.
> 
> **Scope:** markdown link destinations only—not autolinking bare paths in plain text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7ddc78670eff4905d268dadea2f78198953025fd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix Windows drive paths to remain clickable as file links in chat markdown
> - Adds a new remark plugin, `remarkFilesystemLinkDestinations` in [markdown-filesystem-links.ts](https://github.com/pingdotgg/t3code/pull/7189/files#diff-8e1bbe40206dff44d051563aa22cde9021714be313a5e4257933be170ab4f9c0), that rewrites Windows drive-style link destinations (e.g. `D:/path/file.md`, backslash variants) to `file:///` URLs before sanitization strips them.
> - Adds `toFilesystemLinkUrl` in [markdown-links.ts](https://github.com/pingdotgg/t3code/pull/7189/files#diff-010f94b966be6604cb228ce225d1f3ce41b15470b641a970ec5ed07cd2ed2040) to convert Windows drive paths to `file:` URLs and normalize path separators.
> - Updates `normalizeMarkdownLinkHrefKey` in [ChatMarkdown.tsx](https://github.com/pingdotgg/t3code/pull/7189/files#diff-70b5bb003244414202733c39403081a84415e222927a1ef415599958115c0c05) to key drive destinations via their `file:` form so lookups into the pre-scanned metadata map succeed for both slash and backslash variants.
> - The anchor renderer in `ChatMarkdown` now falls back to resolving file link metadata on the fly when the pre-scan misses a link shape.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7ddc786.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/ChatMarkdown.tsx
+++ b/apps/web/src/components/ChatMarkdown.tsx
@@ -73,11 +73,13 @@ import {
   serializeTableElementToMarkdown,
 } from "../markdown-clipboard";
 import { remarkNormalizeListItemIndentation } from "../markdown-list-indentation";
+import { remarkFilesystemLinkDestinations } from "../markdown-filesystem-links";
 import {
   normalizeMarkdownLinkDestination,
   resolveInlineCodeFileLinkMeta,
   resolveMarkdownFileLinkMeta,
   rewriteMarkdownFileUriHref,
+  toFilesystemLinkUrl,
   type MarkdownFileLinkMeta,
 } from "../markdown-links";
 import { readLocalApi } from "../localApi";
@@ -175,7 +177,7 @@ export function orderedListGutterStyle(
   return { "--list-gutter": `${digits + 1}ch` };
 }
 
-const CHAT_MARKDOWN_SANITIZE_SCHEMA = {
+export const CHAT_MARKDOWN_SANITIZE_SCHEMA: NonNullable<Parameters<typeof rehypeSanitize>[0]> = {
   ...defaultSchema,
   attributes: {
     ...defaultSchema.attributes,
@@ -187,12 +189,13 @@ const CHAT_MARKDOWN_SANITIZE_SCHEMA = {
     ...defaultSchema.protocols,
     href: [...(defaultSchema.protocols?.href ?? []), "file"],
   },
-} satisfies Parameters<typeof rehypeSanitize>[0];
+};
 
 const CHAT_MARKDOWN_REMARK_PLUGINS = [
   remarkGfm,
   remarkGithubAlerts,
   remarkNormalizeListItemIndentation,
+  remarkFilesystemLinkDestinations,
   remarkPreserveCodeMeta,
   remarkTagInlineCode,
 ] satisfies NonNullable<ReactMarkdownOptions["remarkPlugins"]>;
@@ -202,6 +205,7 @@ const CHAT_MARKDOWN_REMARK_PLUGINS_WITH_BREAKS = [
   remarkGithubAlerts,
   remarkNormalizeListItemIndentation,
   remarkBreaks,
+  remarkFilesystemLinkDestinations,
   remarkPreserveCodeMeta,
   remarkTagInlineCode,
 ] satisfies NonNullable<ReactMarkdownOptions["remarkPlugins"]>;
@@ -915,7 +919,10 @@ function extractMarkdownLinkHrefs(text: string): string[] {
 }
 
 function normalizeMarkdownLinkHrefKey(href: string): string {
-  const normalizedHref = normalizeMarkdownLinkDestination(href);
+  // Drive destinations are keyed through the same `file:` form the remark plugin
+  // rewrites them to, so a `M:\dir\file.md` in the source text and the
+  // `M:/dir/file.md` the renderer receives resolve to one key.
+  const normalizedHref = toFilesystemLinkUrl(href) ?? normalizeMarkdownLinkDestination(href);
   return rewriteMarkdownFileUriHref(normalizedHref) ?? normalizedHref;
 }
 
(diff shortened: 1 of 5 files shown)


thelonewander3r:
Confirming this on Windows against a real thread, and this branch fixes it cleanly.

## Repro (T3 Code `0.0.33`, Windows 11)

An agent listed the files it had just created. The stored message text is a normal markdown list, percent-encoded because the paths contain spaces (names below are placeholders; the spaces, the `&` and the percent-encoding are verbatim):

```markdown
- [Notes - August 2026.md](C:/Users/me/docs/20%20Projects/Alpha%20%26%20Beta/Notes%20-%20August%202026.md)
- [IMG_0001.jpg](C:/Users/me/docs/20%20Projects/Alpha%20%26%20Beta/media/IMG_0001.jpg)
```

Every link renders blue and underlined, and clicking one does not open the file panel.

Tracing `0.0.33`'s bundle matches your description exactly: `rehype-sanitize` reads `C:` as an unknown protocol and strips the href before `urlTransform` runs, and on the `parseRawHtml: false` path `defaultUrlTransform` returns `""` for the same reason. `resolveMarkdownFileLinkTarget` handles these paths fine — the anchor just never receives a destination to key off, so `markdownFileLinkMetaByHref` is never consulted even though it already holds the resolved entry.

## Verified against 7ddc786

I re-ran the gates shipped in `0.0.33` (`hast-util-sanitize@5.0.2` `safeProtocol`, `react-markdown@10.1.0` `defaultUrlTransform`) with and without this branch's `toFilesystemLinkUrl` plus the anchor fallback:

| destination | main | this PR |
| --- | --- | --- |
| `C:/…/Alpha%20%26%20Beta/Notes%20-%20August%202026.md` | dead, `href=""` | file chip |
| `<C:/Users/x/My File.md>` | dead, `href=""` | file chip |
| `C:\Users\x\file.md` | dead, `href=""` | file chip |
| `C:/src/main.ts:42` | dead, `href=""` | file chip |
| `</Users/x/My File.md>` | dead link | dead link |
| `/Users/x/File.md` | file chip | file chip |

The percent-decoding and the `&` both survive the round trip into the resolved target.

## Two notes

**1. The fallback covers more than the description claims.** Because the anchor re-resolves when the text scan misses, bracketed destinations containing spaces also start working — on Windows drive paths that is the Windows half of #5158, which currently reads as untouched by this PR. Worth stating explicitly so it is clear what is left: the POSIX half (`</Users/x/My File.md>`) still renders dead, so #5158 should stay open after this lands.

**2. The fallback gate is drive-only.** It is conditioned on `toFilesystemLinkUrl(normalizedHref)`, which tests `WINDOWS_DRIVE_PATH_PATTERN` but not `WINDOWS_UNC_PATH_PATTERN`. So a UNC destination the text scan cannot read skips the fallback, even though `resolveMarkdownFileLinkTarget` already resolves UNC paths. Deliberate scope call, or worth widening that gate?

Neither blocks it. Rewriting at the syntax tree so both pipelines are covered by one change is the right shape here, versus patching the sanitizer and `urlTransform` separately.

t3dotgg:
> [!NOTE]

Closing in favor of merged [#8081](https://github.com/pingdotgg/t3code/pull/8081), which fixes Windows drive-path file links in chat without weakening Markdown URL sanitization. Your work is credited in #8081.

PR #7249, opened 2026-08-16
fix(server): refresh Claude context meter after compact

### LINKED ISSUE #4650: [Bug]: context meter ratchets up and never reflects /compact — task_progress writes cumulative tokens into usedTokens via Math.max
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Start a thread with the **claude-code** backend (Claude Opus 5, context mode `High · 1M`).
2. Run it long enough that the context meter fills up (a long orchestration session, ~90%+).
3. Run `/compact` and let it finish successfully — the thread continues with a compacted transcript.
4. Watch the context meter in the composer.

### Expected behavior

Right after a successful `/compact`, the context meter drops to the post-compaction size (the summary plus the unsummarized tail — a small fraction of the window).

### Actual behavior

The meter stays pinned at its pre-compaction value (still red/full) even though the compaction demonstrably succeeded — the agent kept working from a summarized transcript.

It stays wrong across **at least one full completed turn** after the compaction, and then, at some later and not-obviously-related moment, snaps down to the correct low value on its own with no user action. In my session the jump happened several minutes later, in the middle of an unrelated turn. So this is a *lurching* meter rather than a permanently frozen one — the same shape as the "stuck at 5%, then suddenly 100%" report in #2034, just in the other direction.

This looks like the same *symptom* as #2807 / #2034, but #2034 was closed as completed on 2026-06-20 and this is reproducing on the **2026-07-27 nightly**, so it is either a regression or a remaining gap in that fix. Filing separately rather than reopening, because the code path below is a different one from the cumulative-vs-active accounting that #2034 fixed.

### Root cause (read out of the shipped bundle, `t3@0.0.29-nightly.20260727.915`, `dist/bin.mjs`)

The post-compaction reset has exactly one trigger, and it is conditional:

```js
case "compact_boundary":
  yield* emitThreadTokenUsage(
    context,
    compactBoundaryTokenUsageSnapshot(message, context.lastKnownContextWindow, context.lastKnownTotalProcessedTokens),
    { rawMethod: "claude/system/compact_boundary", rawPayload: message }
  );
```

`compactBoundaryTokenUsageSnapshot` bails out unless the SDK supplied `post_tokens`:

```js
function compactBoundaryTokenUsageSnapshot(message, contextWindow, totalProcessedTokens) {
  const metadata = message.compact_metadata;
  if (!metadata || typeof metadata !== "object" || Array.isArray(metadata)) return;
  const compactMetadata = metadata;
  const postTokens = finiteNonNegativeInteger(compactMetadata.post_tokens);
  if (postTokens === void 0 || postTokens <= 0) return;   // <-- silent no-op
  ...
}
```

…and the emitter drops an absent snapshot on the floor:

```js
const emitThreadTokenUsage = Effect.fn("emitThreadTokenUsage")(function* (context, usage, options) {
  if (!usage) return;   // <-- no thread.token-usage.updated is ever sent
  ...
});
```

So when `compact_metadata.post_tokens` is missing or `0`, **no `thread.token-usage.updated` event is emitted at all** and the UI simply keeps rendering the last pre-compaction reading. There is no error, no fallback, and nothing in the transcript that hints the meter is stale.

There *is* an authoritative pull that would answer this correctly — `queryCurrentContextUsage`, which calls `context.query.getContextUsage()` and also refreshes `context.lastKnownContextWindow`:

```js
const queryCurrentContextUsage = Effect.fn("queryCurrentContextUsage")(function* (context, totalProcessedTokens) {
  if (!context.query.getContextUsage) return;
  const usage = yield* Effect.promise(async () => { try { return await context.query.getContextUsage?.(); } catch { return; } });
  if (!usage) return;
  context.lastKnownContextWindow = usage.maxTokens;
  return normalizeClaudeContextUsageApiSnapshot(usage, totalProcessedTokens);
});
```

But its only call site in the bundle is the **turn-completion** path (next to `maxClaudeContextWindowFromModelUsage(result?.modelUsage)`), not the compact boundary. That matches the observed behavior: `/compact` is a standalone action, so nothing re-queries the real usage at the moment the meter goes stale.

### Suggested fix

On `compact_boundary`, fall back to the authoritative pull instead of returning silently:

```js
case "compact_boundary": {
  const snapshot =
    compactBoundaryTokenUsageSnapshot(message, context.lastKnownContextWindow, context.lastKnownTotalProcessedTokens)
    ?? (yield* queryCurrentContextUsage(context, context.lastKnownTotalProcessedTokens));
  yield* emitThreadTokenUsage(context, snapshot, { rawMethod: "claude/system/compact_boundary", rawPayload: message });
  ...
}
```

That keeps the cheap metadata path as the fast case and only pays for the query when the SDK omitted `post_tokens`. If neither source yields a value, emitting an explicitly *unknown* state would still beat rendering a confidently wrong one.

### Scope of what I verified

- ✓ Verified by reading `dist/bin.mjs` of the installed nightly: the three code paths quoted above, and that `queryCurrentContextUsage` has no `compact_boundary` call site.
- 🔬 Not verified: whether the Claude Code SDK actually omits `compact_metadata.post_tokens` in this case, or sends `0`. I could not capture the raw `compact_boundary` payload from the running session. Either input reaches the same silent no-op, so the branch is worth hardening regardless — but a maintainer with payload logging can confirm which one it is.
- 🔬 Not verified: **which** event eventually corrects the meter. I expected the turn-completion path (the only `queryCurrentContextUsage` call site I found), but the reading was still stale after the first turn completed post-compaction and only corrected later. Either there is a second update path I did not find in the bundle, or that call site is reached less often than I assumed. This does not affect the `compact_boundary` no-op above, which is independently readable in the source — but it does mean the delay is not fully explained by it.
- I checked the **server** side only; I did not inspect how `apps/web` renders the meter once the event does arrive.

### Impact

### Version or commit

`t3@0.0.29-nightly.20260727.915` (resolved via `npx t3`; `package.json` points at `pingdotgg/t3code`, directory `apps/server`)

### Environment

Linux (WSL2, kernel 6.6.87.2-microsoft-standard-WSL2) · backend: claude-code · model: Claude Opus 5, context mode `High · 1M` · local checkout

### Workaround

None that makes the meter trustworthy. Ignore it for a while after a `/compact` — it does eventually correct itself without intervention. I initially assumed the *next* completed turn would fix it (that is where `queryCurrentContextUsage` runs), but in my session it was still stale after that turn finished, so I cannot name the event that actually releases it. There is no in-app way to distinguish a stale reading from a real one in the meantime.

matsvarn:
Fixes #4650.

## What changed

`compact_boundary` only emitted `thread.token-usage.updated` when `compact_metadata.post_tokens` was a finite integer > 0. SDK `post_tokens` is optional, so a successful `/compact` often sent no usage event and the composer meter kept the pre-compaction `usedTokens`.

This keeps that cheap path, then falls back once to the existing `queryCurrentContextUsage` (`getContextUsage`) when metadata is missing, zero, or unusable. Compacted thread state is still emitted if neither source yields a snapshot; no usage value is invented.

`task_progress` / `task_notification` were writing cumulative `total_tokens` into `usedTokens` through `Math.max`. That cannot decrease, so a correct post-compact reading got ratcheted back up — worse with subagent progress that also carries `tool_uses` / `duration_ms`. Those events now keep the last active `usedTokens` and only advance `totalProcessedTokens` when a prior reading exists.

```
apps/server/src/provider/Layers/ClaudeAdapter.ts      |  73 +--
apps/server/src/provider/Layers/ClaudeAdapter.test.ts | 510 +++++++++++++++++----
2 files changed, 460 insertions(+), 123 deletions(-)
```

## Why it should exist

After `/compact` the meter stayed red/full, then could snap down and back up later with no user action. The UI already follows the latest `usedTokens`; the stale value was coming from the adapter. Teaching the web meter about Claude payloads would still leave `lastKnownTokenUsage` wrong.

## Test

`vp test run src/provider/Layers/ClaudeAdapter.test.ts` is 76/76.

- valid `post_tokens` does not call `getContextUsage`
- unusable metadata (`undefined` / `{}` / `0` / `"18000"` / `-1`) calls it once
- missing or throwing query emits compacted state and no fabricated usage
- compact then cumulative `task_progress` keeps `usedTokens: 18000`
- first `task_progress` no longer invents `usedTokens`

No UI change, so no before/after images.

---

> [!NOTE]
> **Medium Risk**
> Touches core Claude session lifecycle, token metering, and cleanup on close failures; behavior changes are broad though scoped to the adapter layer with strong test coverage.
> 
> **Overview**
> Fixes stale or ratcheting context-meter readings after `/compact` and during resume by changing how **ClaudeAdapter** derives and emits `thread.token-usage.updated`.
> 
> **Compaction:** Valid `compact_metadata.post_tokens` still drives usage without calling `getContextUsage`; unusable or oversized metadata falls back once to `getContextUsage`, and **no usage is fabricated** if both paths fail. Compacted thread state still emits.
> 
> **Task progress:** `task_progress` / `task_notification` no longer map cumulative totals into active `usedTokens`; they only bump `totalProcessedTokens` when a prior active reading exists, so post-compact meters are not pushed back up.
> 
> **Resume & model changes:** After a resume handshake, authoritative usage is refreshed (thread-only, no bogus `turn.completed`). Model/context-window transitions bump `modelRevision`, gate emissions while `setModel` runs, rebase cached usage for smaller windows, and retry resume refresh after success or failure.
> 
> **Session lifecycle:** `sendTurn` is serialized per session (`Semaphore` + `sendTurnLocked`); provider controls race a stop signal. Failed `query.close()` now **finishes cleanup** (session removed, sends rejected) instead of leaving a “ready” session alive; `session.exited` can report `exitKind: 'error'`.
> 
> Tests in `ClaudeAdapter.test.ts` are expanded heavily to lock in these behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 952478cecba233ecb6324201c68b2dd1530f5ac4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Refresh Claude context meter after compact in `ClaudeAdapter`
> - Fixes token usage tracking after compaction in [ClaudeAdapter.ts](https://github.com/pingdotgg/t3code/pull/7249/files#diff-e696bea66caf21d357bea5ea814bcbb18d1816288cc94247c855823f50fce6dc) by reworking `normalizeClaudeTaskProgressTokenUsage` to only advance `totalProcessedTokens` monotonically, preventing active token inflation or regression.
> - Introduces `ClaudeTokenUsageObservation` to tag usage emissions with `modelRevision` and suppress them during in-flight model transitions.
> - Makes model and context-window changes atomic transitions in `sendTurnLocked`, using a `sendSemaphore` to serialize sends and a `stopSignal` to coordinate shutdown.
> - Validates compact boundary metadata against the known context window in `compactBoundaryTokenUsageSnapshot`, falling back to an authoritative usage query when metadata is missing, invalid, or exceeds the window.
> - Behavioral Change: `task_progress` events no longer emit `thread.token-usage.updated`; `stopSessionInternal` now uses a `stopSignal` and an `exitPolicy` ("always", "on-error", "never") to conditionally emit `session.exited`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 952478c.</sup>
> 
>


Kanaliseren (commented):
Thanks for tackling this — separating active context usage from cumulative processed tokens is the right direction, and the compact-boundary fallback addresses the core failure in #4650.

I found three cases worth addressing before merge:

1. **Validate positive `post_tokens` against the context window.** `compactBoundaryTokenUsageSnapshot` accepts every finite positive value. The shared snapshot builder then clamps oversized values to `maxTokens`, so an invalid or stale-window value can pin the meter at 100% and prevent the authoritative `getContextUsage()` fallback. Please treat values above the known context window as unusable and query instead, with a focused test.

2. **Do not let the fallback block the compacted state indefinitely.** The compact-boundary handler awaits `getContextUsage()` before emitting `thread.state.changed`. Rejection and absence are handled, but a control request that never settles can suppress the compacted event and block later stream processing. The state event should be emitted independently, or the query should have a bounded timeout.

3. **Consider an authoritative refresh when initializing or resuming a Claude session.** We observed a real session whose T3 meter showed roughly 3M/full while Claude `/context` reported 116.6k/1m; running `/context` immediately corrected T3. This PR refreshes at compact boundaries and turn completion, but it does not appear to guarantee that a resumed session is correct before either event occurs. A one-time `getContextUsage()` sync on attach/resume would cover that path.

Non-blocking: the five compact-boundary tests repeat most of the same session/stream choreography. A small helper would make the behavior matrix easier to read.

The PR is also currently conflicted with `main` in `ClaudeAdapter.test.ts`, so it will need a rebase or conflict resolution regardless.

matsvarn:
Addressed all three review cases:

- Oversized `post_tokens` now triggers the authoritative usage query when the context window is known.
- Per-turn model changes now update the known context window.
- The one-second timeout preserves the compacted event and later event order.
- The resume handshake refreshes usage without emitting `turn.completed`.

I merged current `origin/main` normally in `d46bd542d`. The follow-up fix is `a69e2bd4d`.

Verified locally:

- Claude adapter tests: 90 passed.
- Server typecheck: passed.
- Targeted lint: passed with one existing warning.
- Formatting check: passed for both changed files.
- `git diff --check`: passed.

I did not verify the full CI suite.

matsvarn:
Update for final head `980e884e`. This supersedes the earlier `a69e2bd` and 90-test note.

- Resume handshakes now refresh usage without completing a real user turn.
- Unusable compact metadata uses the bounded one-second authoritative fallback.
- Oversized `post_tokens` values are rejected before clamping. Equality with the known limit remains valid.
- Same-model selections preserve API-learned windows. Real model or window changes clear or rebase stale usage state.
- Current `origin/main` was integrated with normal merge commits. No rebase was used.

Local proof:

- Claude adapter suite: 108 tests passed.
- Format check passed.
- Targeted lint passed with one inherited `unicorn(no-useless-spread)` warning.
- Server typecheck passed.
- `git diff --check` passed.
- Two independent reviews found no remaining issues.

No live Claude SDK session was tested. CI proof remains pending until CI completes.

matsvarn:
@Kanaliseren, this update addresses the three cases in your review.

Final head: `952478cecba233ecb6324201c68b2dd1530f5ac4`.

- Oversized compact `post_tokens` values are rejected. A bounded authoritative query supplies the fallback.
- An unresponsive query cannot block the compacted state or later stream events.
- Resume performs one authoritative refresh. Model revisions reject stale observations.

Additional review fixes preserve API-learned windows and compaction policy during fallbacks. Terminal close cleanup now emits the correct failed-replacement exit. Send and model transitions are serialized.

I merged current `main` normally through `f925d639`. The final merge had no conflict. I did not rebase or force-push.

Local proof:

- ClaudeAdapter tests: 112 passed.
- New main schema test: 1 passed.
- Server and Codex schema package typechecks: passed.
- Format and diff checks: passed.
- Targeted lint exited 0 with inherited warnings.
- Diff and worktree checks are clean.
- Two independent final reviews found no issues.

Remote proof:

- Macroscope correctness and Effect checks pass.
- Cursor Bugbot passes with no comments on the final SHA.
- CodeRabbit skipped its review.
- Macroscope approvability finished neutral because the review and vouch state is incomplete.

Limits: Tests use a mocked Claude SDK. Local Node is 26.7.0, but the repository requires 24.13.1. No live SDK timing test was run.

Please review and vouch if this is ready.

juliusmarminge:

**Reason:** 2790-line ClaudeAdapter rewrite; compact/usage refresh already evolved on main and in #8610.
**Related:** #8610 + main compactedSinceLatestAssistantUsage

PR #7265, opened 2026-08-16
fix(server): bound usage transcript lines

### LINKED ISSUE #7258: Usage scan crashes server on oversized Codex JSONL record
## Summary

Opening the Usage page can terminate the entire `t3@latest` Node process when a Codex session rollout contains one very large JSONL record.

In the observed case:

- **7 days** worked.
- **30 days** and **90 days** showed zero usage and then crashed the server.
- The longer windows included an older Codex rollout with one 516.47 MiB line.
- That line was a non-usage `event_msg -> patch_apply_end -> delete` record.
- Removing only that irrelevant record made both 30-day and 90-day scans complete successfully.

## Environment

- Windows
- Node.js v24.15.0
- `t3@latest`, launched with the local web server
- Codex transcripts under the normal `.codex/sessions` directory

## Crash

```text
node:internal/readline/interface:652
        this[kLine_buffer] += string;
                              ^

RangeError: Invalid string length
    at [_normalWrite] [as _normalWrite] (node:internal/readline/interface:652:31)
    at ReadStream.ondata (node:internal/readline/interface:263:23)
    at ReadStream.emit (node:events:509:28)
    at addChunk (node:internal/streams/readable:563:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:514:3)
    at Readable.push (node:internal/streams/readable:394:5)
    at node:internal/fs/streams:293:14
    at FSReqCallback.wrapper [as oncomplete] (node:fs:670:5)
```

## Root cause

`apps/server/src/usage/usageTranscriptReader.ts` reads each transcript with:

```ts
NodeReadline.createInterface({
  input: NodeFS.createReadStream(filePath, { encoding: "utf8" }),
  crlfDelay: Infinity,
});
```

Node's `readline` accumulates an entire line into a JavaScript string before T3 can apply `mightCarryUsage`. A sufficiently large record exceeds V8's maximum string length and throws from the stream's data callback. The surrounding `try/catch` does not prevent the process-level crash.

Measured transcript:

- Total file size before workaround: 551,526,950 bytes
- JSONL records: 3,403
- Oversized record: line 3,387
- Oversized record length: 541,561,318 bytes
- Record kind: `event_msg -> patch_apply_end -> delete`
- File remained correctly newline-delimited

## Workaround validation

After preserving the original transcript and removing only that non-usage line:

- Active transcript size: 9,965,632 bytes
- Remaining records: 3,402
- Largest remaining line: 106,482 bytes
- 30-day scan: 753 files, 1,559,096,780 bytes, 32,808 usage records, 0 failed files
- 90-day scan: 753 files, 1,559,126,792 bytes, 32,811 usage records, 0 failed files

Both scans used T3 Code's current `listTranscriptFiles` and `readTranscriptRecords` implementation directly.

## Expected behavior

An oversized or malformed transcript record should not terminate the T3 server. The Usage response should either:

1. skip an oversized record that cannot affect usage, or
2. skip/report the affected transcript while returning partial usage.

## Suggested direction

Use a bounded byte/chunk line reader instead of unbounded Node `readline`. It should stop buffering a record after a conservative limit, drain through its newline without decoding the whole record, and continue scanning later records.

A focused test could expose an injectable low line limit so the regression can be covered without committing a hundreds-of-megabytes fixture.

Possibly related but not equivalent: #996 reports crashes around a heavy thread, without this Usage-page stack or transcript-reader path.

ifBars:
Oversized JSONL tool-output records can make the usage reader return `null` for an entire transcript, losing usage records that follow them.

Bound each record to 64 MiB before decoding and skip oversized records through their next newline. This extends the incremental reader from #9024 while preserving its result contract, guard hash, Codex reducer state, and byte-exact resume position. An unfinished oversized tail remains outside the committed position so a later scan can consume its newline and subsequent usage.

Validation:

- All 91 usage tests pass, including the eight existing resume tests and six new bounded-record cases.
- Targeted lint, formatting, and TypeScript checks pass.
- Two new assertions fail against unmodified main.
- A synthetic 517 MiB tool-output record makes main return `null`; the updated reader preserves later usage and successfully resumes after an append, with about 132 MiB peak RSS.

Fixes #7258.

Updated with GPT-6 via Codex.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transcript processing for very large lines, preventing oversized records from causing failures.
  * Oversized or incomplete records are skipped safely while subsequent valid usage records continue to be processed.
  * Preserved parsing progress and session/model state when resuming reads.
  * Added support for accurate byte-based limits across Unicode text and different line-ending formats.


shivamhwp:

Please rebase this onto the incremental reader from #9024. `readTranscriptRecords` now accepts a `TranscriptParsePosition` and returns `records`, `tailRecords`, `position`, and `resumed`; its callers no longer consume a plain record array. Replacing that implementation with this version would discard warm-scan resumption and break the usage service's result contract.

The oversized-record problem remains on main, though its failure mode has changed. A 517 MiB tool-output line now makes the reader return `null` for the whole transcript, losing later usage, rather than escaping the old readline callback. The bounded reader here preserves the later usage record.

Carry the byte limit into the current parser while retaining its guard hash, Codex reducer state, and byte-exact resume offset. When an oversized line ends, advance past its newline before processing later records. Keep an unterminated oversized tail outside the committed resume position so appending its newline and later usage can still be handled correctly. Retain the current resume tests and add oversized records on both sides of the resume boundary.

ifBars:
@shivamhwp Addressed in d1b28accf. I integrated current main (including #9024) with a merge to preserve the existing branch history, and moved the 64 MiB limit into the incremental parser.

The reader retains the `TranscriptParsePosition` argument and `{ records, tailRecords, position, resumed }` result, guard hash, and Codex reducer state. Oversized terminated lines advance the committed byte offset past their newline; oversized unfinished tails remain outside that position and are re-read on the next scan.

All eight existing resume tests are retained. Six new cases cover oversized records before and after resumption, LF/CRLF offsets, unfinished tails on cold and resumed scans, exact byte-limit handling with UTF-8, and skipped Codex state changes. All 91 usage tests pass, as do targeted lint, formatting, and TypeScript checks. Two new assertions fail against unmodified main.

A synthetic 517 MiB tool-output record reproduces main returning `null` (about 1,622 MiB peak RSS). The updated reader retains the usage before and after it, then resumes successfully after an append (about 132 MiB peak RSS, 548 ms for the scan and append check on this Windows machine).

Issue #5942, opened 2026-08-10
[Bug]: Workflow/background agent tokens inflate main agent Context Window clock (parent should only show own context)


fuggysense:
## Summary

When a **workflow** spins up **background agents**, the main (orchestrator) chat’s **Context Window** meter attributes **child/workflow usage** to the parent. The clock becomes an unreliable signal for how full the **parent’s real context** is.

This report is **display / token accounting only** — I am **not** seeing child tool/stream spam injected into the parent transcript. Just the numbers on the parent Context Window widget.

## Expected

- Parent **Context Window** (`%` + `used/limit`, e.g. `x / 1m`) reflects **only** the main agent’s actual session context.
- Background / workflow agents have **their own** usage, or at least must **not** accumulate into the parent’s context meter.
- **Total processed** may be project- or thread-wide if that’s intentional — but if so, it should be labeled separately from “Context Window” so it isn’t read as parent fill.

## Actual

With background agents running under a workflow, the parent composer shows something like:

- **Context Window: 100% · 1m/1m** (full bar)
- **Total processed: 1.6m**
- UI: **“3 agents working in the background”**
- Model: **Claude Opus 5**

The parent feels nowhere near a full 1M context of its own messages; the meter climbs as workflow agents work, which makes the clock useless for deciding whether the orchestrator needs compaction / a fresh thread.

## Steps to reproduce

1. Open a project thread in T3 Code (desktop Alpha).
2. Use **Claude** (Opus 5) as the main agent.
3. Start a **workflow** that spawns multiple background agents (e.g. 3 agents working in the background).
4. Watch the **Context Window** chip on the **main** composer while those agents run.
5. Observe parent context % / used tokens rise toward the limit even though only the children are doing the heavy work.

## Environment

| | |
|---|---|
| App | T3 Code (Alpha) desktop |
| Version | **0.0.32** (`com.t3tools.t3code`) |
| OS | macOS |
| Provider | **Claude** (Claude Opus 5) |
| Surface | Main thread composer → Context Window + Total processed |

## What this is / isn’t

| | |
|---|---|
| **Is** | Parent context meter / clock mis-attributing workflow or background-agent tokens |
| **Is not** | Parent transcript polluted with child tool calls / stream spam (not observed) |
| **Is not** | Asking for a new product feature — this is incorrect accounting for the existing Context Window UI |

## Why it matters

[...]

PR #7287, opened 2026-08-17
fix(claude): report subagent model and effort

### LINKED ISSUE #7281: Agents panel shows the parent session's model/effort for every subagent instead of the subagent's own
## Summary

Every row in the Agents panel displays the **parent session's** model and effort rather than what the subagent actually ran on. With a session on `opus-5 · xhigh`, eight subagents spanning four different models all rendered as `opus-5 · xhigh`.

The token counts and titles on those rows are correct — only model/effort are wrong — so it reads as authoritative rather than as a placeholder. That's what makes it costly: it silently misreports model routing, and anyone reasoning from the panel will draw wrong conclusions about which model handled what.

## Reproduction

Requires custom subagents that declare `model:`/`effort:` in frontmatter (`~/.claude/agents/*.md`).

1. Run a session on a model/effort distinct from your agents' — here `opus-5` at `xhigh`.
2. Spawn agent types whose definitions pin different models, e.g. `model: haiku` / `effort: low`, `model: sonnet` / `effort: medium`, `model: fable` / `effort: high`.
3. Open the Agents panel.

**Expected** — each row shows that agent's own model/effort.
**Actual** — every row shows `opus-5 · xhigh`.

## Evidence

Eight probes, each `Reply with the single word ACK`. "Panel" is what the sidebar rendered; "actual" is from the SDK's own run records at `~/.claude/projects/<project>/<session>/subagents/agent-*.jsonl`, where `message.model` on each assistant record is the API's report of which model served the request:

| agent type | agent definition | panel showed | actually ran |
| --- | --- | --- | --- |
| `lookup-fast` | haiku / low | `opus-5 · xhigh` | `claude-haiku-4-5-20251001`, no effort |
| `implement-fast` | opus / low | `opus-5 · xhigh` | `claude-opus-5` / `low` |
| `general` | sonnet / medium | `opus-5 · xhigh` | `claude-sonnet-5` / `medium` |
| `review-standard` | sonnet / high | `opus-5 · xhigh` | `claude-sonnet-5` / `high` |
| `deep-reasoning` | opus / high | `opus-5 · xhigh` | `claude-opus-5` / `high` |
| `critical-review` | opus / xhigh | `opus-5 · xhigh` | `claude-opus-5` / `xhigh` ✅ |
| `long-horizon` | fable / high | `opus-5 · xhigh` | `claude-fable-5` / `high` |
| `inherit` | (inherits parent) | `opus-5 · xhigh` | `claude-opus-5` / `xhigh` ✅ |

The two correct rows are the two that coincidentally match the session — `critical-review` happens to share it, and `inherit` is supposed to.

Same eight agents in both columns: panel token counts match the records exactly (8.8k↔8895, 16.7k↔16684, 22.2k↔22151, 12.0k↔11960, 16.5k↔16537, 7.7k↔7707, 16.7k↔16729, 16.4k↔16421). Two independent signals confirm different models really ran: the haiku probe emitted `out=73` tokens where every Opus/Sonnet probe emitted `out=5`, and it carries no `effort` field at all — Haiku 4.5 doesn't take reasoning effort, yet the panel labels it `xhigh`.

## Cause

`apps/server/src/provider/Layers/ClaudeAdapter.ts:3194-3202` seeds both fields at launch:

```ts
const model =
  trimmedString(launchInput?.model) ?? trimmedString(context.session.model ?? undefined);
const rawLaunchEffort = launchInput?.effort;
const effort =
  trimmedString(rawLaunchEffort) ??
  (typeof rawLaunchEffort === "number" && Number.isFinite(rawLaunchEffort)
    ? String(rawLaunchEffort)
    : context.currentEffort);
```

The comment above it — "absent ones inherit the session's selection (SDK behavior)" — holds only for agent types that genuinely inherit (`general-purpose`, `claude`, `inherit`). For a file-defined agent, the SDK resolves model and effort from the agent's own frontmatter and pushes an effort layer for that subagent; the `Agent` tool input carries no override, so this fallback attributes the session's values to an agent that never used them. The adapter can't see the agent definition, so the launch-time seed is unavoidably a guess — the issue is that the guess is presented as fact.

**Effort is never corrected.** The model has a refinement path at `ClaudeAdapter.ts:2876-2883`, which overwrites `owningAgent.model` from the authoritative `message.model` on subagent assistant snapshots, and `taskLinkageFor` (`:1034-1035`) propagates both onward. There is no equivalent for effort anywhere — `context.currentEffort` is the only source it ever has.

For model, the refinement path exists but did not land for these direct spawns (all eight kept the seeded value). I haven't diagnosed why; the `agentIdForParentToolUse(…, assistantParentToolUseId)` lookup at `:2878` seems the place to look, and these probes were short-lived (1–2s), so a snapshot ordering or matching issue is plausible. Flagging it as observed rather than asserting the mechanism.

## Suggested direction

- Give effort the same snapshot-refinement treatment model has, or read both from the subagent run records, which are authoritative.
- Render unknown-until-refined as absent rather than falling back to session values — a blank is honest, a wrong model is not.
- Don't synthesize an effort for models that don't support one: the `lookup-fast` record has no `effort` field, and `xhigh` on Haiku 4.5 is not a value that can exist.
- Worth a regression test that a subagent whose definition pins a model differing from the session renders the pinned one.

## Environment

t3code on macOS 26 (Darwin 27.0.0) · Claude Code 2.1.233 · session `claude-opus-5` at `xhigh`

lnieuwenhuis:
## Problem

Claude file-defined subagents were seeded with the parent session's model and effort. The Agents panel therefore showed the parent settings for every subagent, and authoritative model snapshots that arrived late never reached the client.

## Fix

- leave subagent model and effort unknown unless the Agent launch explicitly overrides them
- reconcile authoritative model snapshots before or after `task.started`
- capture each subagent's effort through the SDK's typed `SubagentStart` and `SubagentStop` hooks
- emit late identity corrections and bound pending refinements to avoid unbounded session growth
- cover parent non-inheritance, event ordering, terminal-task correction, explicit overrides, and cache eviction

## Verification

- `pnpm exec vp test run apps/server/src/provider/Layers/ClaudeAdapter.test.ts` (74 passed)
- `pnpm exec vp test run packages/client-runtime/src/state/subagentRuntime.test.ts` (48 passed)
- targeted format, lint, diff, and changed-file secret checks

Fixes #7281

Model: GPT-5.6 Sol
Harness: Codex in T3 Code

---

> [!NOTE]
> **Medium Risk**
> Changes runtime event payloads and task identity logic in the Claude provider path; incorrect linkage could affect UI/analytics for subagents, but scope is localized to adapter event emission with heavy test coverage.
> 
> **Overview**
> Fixes **misattributed subagent model/effort** in the Claude adapter: file-defined agents no longer inherit the parent session’s selection when the Agent tool omits overrides. Identity is seeded only from explicit launch input, then refined from subagent assistant snapshots and SDK **`SubagentStart` / `SubagentStop`** hooks (effort).
> 
> **Late or early identity** is handled via a bounded pending-refinement map (model + effort, cap 256): snapshots that beat `task_started` apply on start; snapshots or hooks that arrive after terminal rows emit **`task.updated`** corrections. Error frames with model `<synthetic>` are ignored. Repeated **`task_started`** on resume keeps the original Agent `tool_use_id` so snapshots still link correctly.
> 
> Tests add `invokeSubagentEffortHook` and broaden scenarios (non-inheritance, ordering, post-completion correction, resume, synthetic models, explicit overrides, eviction).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f67bd5580cd182871adb6f16a73053e10bbaee0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix subagent model and effort attribution in `ClaudeAdapter`
> - Subagent `task.started` events no longer infer model or effort from the parent session; they use explicit Agent-tool overrides, buffered runtime refinements, or existing task metadata.
> - Adds `SubagentStart`/`SubagentStop` hook callbacks to capture effort metadata, either buffering it for unregistered tasks or emitting `task.updated` corrections for registered tasks.
> - Replaces the model-only pending buffer with a model-and-effort refinement buffer capped at 256 entries with FIFO eviction, keyed by a generic identity string.
> - Filters the synthetic error-frame model sentinel out of subagent attribution so progress retains the real model.
> - Risk: `task.started` now reports unknown model/effort when no explicit override or runtime refinement is available — check `makeClaudeAdapter` in [ClaudeAdapter.ts](https://github.com/pingdotgg/t3code/pull/7287/files#diff-e696bea66caf21d357bea5ea814bcbb18d1816288cc94247c855823f50fce6dc) for the fallback removal.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1f67bd5.</sup>
> 
>

## Summary by CodeRabbit

- **Improvements**
  - Improved accuracy of subagent model and effort attribution.
  - File-defined agents now display their own metadata instead of inheriting values from the parent session.
  - Updated task information when more authoritative metadata becomes available.
  - Preserved attribution when tasks are resumed.
  - Unspecified effort remains blank unless explicitly provided.


lnieuwenhuis:
Merged current `main` and resolved the conflicts with #7583, which landed a narrower version of the snapshot/task_started race fix (model only, keyed by parent_tool_use_id). This branch's pending-refinement buffer subsumes that behavior, so I kept the refinement implementation and removed the now-redundant `pendingTaskModels` map.

Two notes for reviewers:
- #7583's new test ("a subagent snapshot that beats task_started still wins over the seed") asserted that a subagent with no explicit effort override inherits the parent session's effort. That inheritance is exactly the misattribution this PR removes, so the assertion now expects `undefined`; the model assertions (including propagation into `task.progress`) are unchanged, and effort attribution is covered by the SubagentStart-hook tests here.
- Everything else merged cleanly. All 82 ClaudeAdapter tests pass locally.

SamGu-NRX:
Independent reproduction, plus one data point on `AgentInfo.model` that may be useful here.

**Prevalence.** I compared the label on every `local_agent` task in my local event store against the `model:` each agent file declares, normalizing effort and `[1m]` suffixes. Over 241 tasks: **192 wrong at `task_started` (79%)**, and **35 still wrong at rest (14%)** after snapshot refinement had its chance. One task recorded `"model": "claude-opus-5"` on `task.started` and `"model": "gpt-5.6-luna"` on `task.progress` seconds later, for the same task id, which matches your reading that routing was correct and only the label was wrong.

The 79%→14% gap is the part I found interesting: refinement does most of the work, so the residual is concentrated in children that never stream an assistant snapshot the parent sees.

**One possible addition.** `query.supportedAgents()` returns `AgentInfo[]`, and `AgentInfo.model` is documented as "Model alias this agent uses. If omitted, inherits the parent's model" (`sdk.d.ts:97-109`). It resolves off the cached initialization result rather than a per-task control request. Probing my installed CLI 2.1.245:

```
claude               (inherits)
Explore              (inherits)
general-purpose      (inherits)
Plan                 (inherits)
statusline-setup     sonnet
theo-composer        cursor/composer-2.5
```

That would give a launch-time value between an explicit `Agent` override and the parent fallback, so a definition-pinned agent starts with its own alias rather than blank, while genuinely inheriting agents still fall through. It is an alias, not the resolved API id, so your snapshot path would still be the authority that corrects it.

Worth noting the built-in defaults are version-sensitive: on the SDK-bundled 2.1.170 `Explore` reports `haiku`, but on 2.1.245 it inherits. So this helps custom agents much more than stock ones, and it does not reach effort at all, since `AgentInfo` has no effort field. It may well not be worth the extra surface here.

**Composition.** I have a separate fix for a different bug in this file, the parent context meter counting subagent tokens (#5942, PR #8453). I applied your branch and mine together on today's `main`: they typecheck, and the suite passes at 85 tests once three shared tests are reconciled. Only the test file collides; the source changes are independent. Happy to rebase mine whenever yours lands.

VendettaGhub:
Independent Windows test report for **1f67bd5580cd182871adb6f16a73053e10bbaee0**. I checked this PR because we use a local Claude SDK / hybrid-router setup and had patched the same parent-model/effort attribution problem locally.

### Results

- Unmodified PR checkout: `pnpm exec vp test run apps/server/src/provider/Layers/ClaudeAdapter.test.ts packages/client-runtime/src/state/subagentRuntime.test.ts` — **133 passed**, 2 files.
- Additional local, synthetic matrix: parameterized the existing “uses a subagent snapshot that arrives before task.started” test across five snapshot model identifiers (`gpt-6-astra`, `gpt-5.6-sol`, `gpt-5.6-luna`, `claude-opus-5`, `claude-fable-5-1`) and five hook effort strings (`low`, `medium`, `high`, `xhigh`, `max`). All **25 combinations passed**. With that one test expanded to 25, the same two-file run passes **157 tests**.
- `pnpm exec vp run --filter t3 typecheck` — passed with the matrix in place.
- `git diff --check` — passed.

The matrix emits an assistant snapshot with `parent_tool_use_id`, invokes the existing `invokeSubagentEffortHook(harness, "SubagentStart", "task-model", effort)` helper, then emits the matching `task_started`. It asserts that `task.started` preserves the exact snapshot model and hook effort instead of the parent session's `claude-opus-4-6` / `max` selection. The original suite also exercises unknown metadata, explicit overrides, late corrections and resumed-task identity.

### Boundaries

These are **adapter/client-runtime tests with synthetic SDK input**, not real provider calls or a rendered sidebar check. The matrix treats model names and effort values as metadata: it does not establish that every model supports every effort, that a hybrid router applies them, or that the live CLI emits each tested hook payload. No production T3 instance was replaced or restarted, and this is not a latest-main integration/rebase test.

Environment: Windows x64, Node **25.8.0**, pnpm **11.10.0**, lockfile-installed SDK **0.3.170**. Node is outside the repository's declared `^24.13.1` range, so this is supplementary Windows evidence, not a replacement for supported-runtime CI.

For our use case, preserving runtime-reported model IDs and effort is preferable to upstreaming our local frontmatter/alias display workaround. I did not find an additional issue in the tested metadata path that warrants a competing PR.

Test report prepared with Codex Desktop (local configuration: GPT-6 Astra, medium).

juliusmarminge:

PR #7293, opened 2026-08-17
feat(grok): show Grok skills in the composer $ picker

dascapytal1559:
## What Changed

The Grok provider status check now discovers skills by running `grok inspect --json` (best-effort, 5s cap, concurrent with the existing ACP model discovery) and maps them onto the existing `ServerProviderSkill` contract, so the composer's `$` picker works in Grok threads. Skills Grok itself hides from users — `[skills].disabled` entries and `user-invocable: false` skills — map to disabled and stay out of the picker.

No contract or client changes: `skills` already exists on the provider snapshot schema, so web, desktop, and mobile follow over the wire. The probe runs in the workspace directory (threaded from `ServerConfig`, same as `ClaudeDriver`) so local and repo-scope skills resolve like a live Grok session. Any probe failure (missing binary, timeout, non-zero exit, malformed JSON) yields an empty list and never degrades the snapshot.

## Why

The Grok provider snapshot never populated `skills` — the `$` picker always showed "No skills found" in Grok threads even though the Grok CLI discovers and uses skills fine. Claude discovers skills via filesystem scan and Codex via its app-server RPC; Grok had nothing. Reusing grok's own `inspect` report (rather than reimplementing its multi-root scan: local/repo/user roots, Claude/Cursor compat dirs, bundled and plugin skills, config overrides, name dedup) keeps the picker exactly in sync with what a live session loads — the same approach as trusting the Codex app-server's `skills/list`.

Tests: parser mapping, disabled/non-invocable/malformed entries, fake-binary probe behavior (cwd propagation, failure modes), and an end-to-end status-check test; also validated against real `grok inspect --json` output (grok 1.0.4, 37 skills across user/project/bundled scopes).

## UI Changes

No rendering code changed — the existing picker just receives data now.

Before (Grok thread, `$` typed):

before

After:

after

## Checklist

- [ ] I included a video for animation/interaction changes (n/a — no motion changes)

Work done by Claude Fable 5 via Claude Code.

### DIFF
--- a/apps/server/src/provider/Drivers/GrokDriver.ts
+++ b/apps/server/src/provider/Drivers/GrokDriver.ts
@@ -89,6 +89,7 @@ export const GrokDriver: ProviderDriver<GrokSettings, GrokDriverEnv> = {
       const httpClient = yield* HttpClient.HttpClient;
       const serverSettings = yield* ServerSettingsService;
       const eventLoggers = yield* ProviderEventLoggers;
+      const { cwd } = yield* ServerConfig;
       const processEnv = mergeProviderInstanceEnvironment(environment);
       const continuationIdentity = defaultProviderContinuationIdentity({
         driverKind: DRIVER_KIND,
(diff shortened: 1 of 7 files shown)


petekblaz:
Need this merged, cant really use grok currently since skills cant be invoked :(

willsantiagomedina:
Please merge this. I need this badly

PR #7357, opened 2026-08-18
fix(web): revalidate file previews on reopen

### LINKED ISSUE #5779: [Bug]: Files panel serves file content cached at app startup — external and agent edits never appear until full app restart
The files panel renders file content from what appears to be a snapshot taken at app startup. Any change made to a file on disk after that — by an agent, a script, or another program — is not reflected in the panel, and no user action short of restarting the app refreshes it.

**Repro (macOS, Darwin 25.4.0, desktop app):**

1. Open a markdown file from the workspace in the files panel.
2. Modify the file on disk from outside the panel (echo/printf append, `sed`, or an agent's file-edit tool — all behave identically).
3. The panel keeps rendering the old content.

**Things that do NOT refresh it (all tested):**
- Closing and reopening the file tab
- `touch` on the file
- Deleting the file and recreating it at the same path (new inode)
- Edits made via an agent's file tools vs plain shell writes — no difference

**The only thing that refreshes it:** quitting and relaunching the app.

**Expected:** the panel re-reads from disk on file change (watcher), or at minimum on tab reopen.

**Impact:** any workflow where an agent or external tool edits files while the user reads them in the panel — the user sees stale content with no indication it is stale. In our case this cost a debugging round on "the edit didn't happen" when the edit was fine on disk the whole time.

### LINKED ISSUE #5866: File preview can keep serving cached content after an agent modifies the file
## What happened

I opened a Markdown file in the file preview panel and switched it to the readable/formatted view. An agent then modified that same file. From that point on the panel kept showing the old content — the agent's changes never appeared, and the diff showed nothing either. Reopening the file showed the pre-edit content.

The project was a local folder with no remote at the time. I pushed it to GitHub thinking that might matter, and the behaviour was unchanged — which in hindsight fits, since nothing about this looks git-related.

## What I found in the code

Reading at `02f4ce56`, there's a path where a file's cached content can outlive the file itself.

`useProjectFileQuery` returns the optimistic entry in preference to the query result:

- `apps/web/src/components/files/projectFilesQueryState.ts:194` — `data: optimisticFile?.data ?? data`
- same again in `resolveProjectFileQueryData`, `projectFilesQueryState.ts:107`

So while an optimistic entry exists for `(environmentId, cwd, relativePath)`, it masks whatever is actually on disk.

That entry is cleared in exactly one place — the success branch after a save is confirmed:

- `projectFilesQueryState.ts:93-94` — `if (result._tag === "Success" && appAtomRegistry.get(atom) === confirmed) { appAtomRegistry.set(atom, null) }`

If that re-read doesn't succeed, or the atom moved on in between, the entry stays. I couldn't find anything that drops it afterwards, and in particular I couldn't find any invalidation of a file's query when something *outside the editor* — an agent tool call — writes that file.

Two things that look related:

- `clearProjectFileQueryData` (`projectFilesQueryState.ts:110`) does exactly this job, and has no callers outside its own definition and `projectFilesQueryState.test.ts`.
- `confirmedAgainst` is written at `:54` and `:85` but never read anywhere. It reads like a staleness check that wasn't finished.

## What I have not verified

I traced this from the source; I have not reproduced it from a clean checkout with a deterministic trigger, and I don't know what made the confirm step fail in my case. So treat the mechanism as a strong hypothesis rather than a diagnosis. It does explain every part of what I saw, including it persisting across reopening the file.

## Happy to send a PR

I didn't open one because the fix is a decision rather than a line: the right place to drop optimistic state could be on agent tool completion, on a file-watcher event, or on any external write, and that's your call rather than mine. If you tell me which invalidation point you want, I'll keep it small and focused.

ralphsmith80:
## What this does

File previews now re-read the workspace file when the preview mounts instead of trusting a fresh client cache entry. Confirmed editor content remains protected while a read is in flight or fails, but a later successful disk read can no longer be hidden by old optimistic state.

Fixes #5779.
Fixes #5866.

## Verification

- `pnpm exec vp test run apps/web/src/components/files/projectFilesQueryState.test.ts` (3 tests passed)
- Targeted Vite+ lint and formatting checks for both changed files

## Intentional decisions

This fix revalidates on preview mount and keeps the existing short-lived query cache for other consumers. Live filesystem watching and workspace-tree rescans remain separate concerns.

Generated with GPT-5.6 in the Codex harness through T3 Code.

---

> [!NOTE]
> **Medium Risk**
> Touches shared file query/optimistic merge logic used by previews and t3.json reads; wrong timing could flash stale content or drop drafts, but changes are localized with new unit tests.
> 
> **Overview**
> **File previews** no longer show stale disk content when reopening a file that still has a fresh in-memory cache entry. `useProjectFileQuery` now triggers a refresh on mount when the query atom already holds settled success data, so the preview re-reads from the workspace instead of skipping I/O.
> 
> **Confirmed optimistic editor content** stays visible while a background revalidation is in flight or fails. New helpers (`hasSettledProjectFileRevalidation`, `shouldRevalidateProjectFileOnMount`) gate when server data replaces optimistic data and when mount-time refresh runs. Optimistic state is cleared only after a successful, non-waiting read that differs from what was confirmed—not immediately on confirm in all cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a59f1731e44f66cb56cec9d9682305f8aa419b4d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix file preview revalidation on reopen in `useProjectFileQuery`
> - `useProjectFileQuery` now triggers a cache refresh on mount when a settled cached result exists, ensuring file previews reflect disk state when reopened rather than serving stale cached data.
> - Adds `hasSettledProjectFileRevalidation` to detect when a revalidation read has completed successfully relative to a prior confirmed result, and `shouldRevalidateProjectFileOnMount` to determine if a mount should trigger revalidation.
> - Once revalidation settles, the optimistic entry is cleared (only if it hasn't been replaced by a newer draft) and confirmed data is returned instead of the optimistic draft.
> - Behavioral Change: file previews that previously showed cached optimistic data on reopen will now show confirmed disk data after revalidation completes.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a59f173.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/files/projectFilesQueryState.test.ts
+++ b/apps/web/src/components/files/projectFilesQueryState.test.ts
@@ -1,13 +1,16 @@
 import type { ProjectReadFileResult } from "@t3tools/contracts";
 import { EnvironmentId } from "@t3tools/contracts";
+import { AsyncResult } from "effect/unstable/reactivity";
 import { afterEach, describe, expect, it, vi } from "vite-plus/test";
 
 import {
   clearProjectFileQueryData,
   confirmProjectFileQueryData,
   getOptimisticProjectFileQueryData,
+  hasSettledProjectFileRevalidation,
   resolveProjectFileQueryData,
   setProjectFileQueryData,
+  shouldRevalidateProjectFileOnMount,
 } from "./projectFilesQueryState";
 
 const environmentId = EnvironmentId.make("environment-project-files-query-test");
@@ -47,5 +50,45 @@ describe("project files queries", () => {
     expect(
       confirmProjectFileQueryData(environmentId, "/repo", "convex.json", '{"nodeVersion":"22"}'),
     ).toBe(true);
+  });
+
+  it("keeps confirmed optimistic data until a later read settles successfully", () => {
+    const confirmedAgainst = AsyncResult.success({
+      relativePath: "convex.json",
+      contents: '{"nodeVersion":"20"}',
+      byteLength: 20,
+      truncated: false,
+    });
+    const refreshed = {
+      relativePath: "convex.json",
+      contents: '{"nodeVersion":"22"}',
+      byteLength: 20,
+      truncated: false,
+    } satisfies ProjectReadFileResult;
+
+    expect(
+      hasSettledProjectFileRevalidation(
+        confirmedAgainst,
+        AsyncResult.success(refreshed, { waiting: true }),
+      ),
+    ).toBe(false);
+    expect(
+      hasSettledProjectFileRevalidation(confirmedAgainst, AsyncResult.success(refreshed)),
+    ).toBe(true);
+  });
+
+  it("revalidates settled cached file data when the preview mounts", () => {
+    const cached = {
+      relativePath: "convex.json",
+      contents: '{"nodeVersion":"20"}',
+      byteLength: 20,
+      truncated: false,
+    } satisfies ProjectReadFileResult;
+
+    expect(shouldRevalidateProjectFileOnMount(AsyncResult.initial(false))).toBe(false);
+    expect(shouldRevalidateProjectFileOnMount(AsyncResult.success(cached, { waiting: true }))).toBe(
+      false,
+    );
+    expect(shouldRevalidateProjectFileOnMount(AsyncResult.success(cached))).toBe(true);
   });
 });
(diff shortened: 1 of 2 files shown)


ralphsmith80:
This might be useful, but my agent got crazy submitting PRs and it's not something I have personally noticed.

Issue #4928, opened 2026-07-30
[Bug]: Claude shows "could not verify authentication status" on managed Bedrock installs


preetpatel:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

### Steps to reproduce

This shows up on managed Amazon Bedrock installs, where the credentials come from MDM/AWS rather than `claude auth login`.

1. Run Claude against Bedrock with managed credentials.
2. Open T3 and let the Claude provider status check run.
3. Look at the auth status.

`checkClaudeProviderStatus` decides auth state purely from the SDK capability-init probe. On managed Bedrock the probe comes back without any account info, so even though Claude is authenticated and working, T3 falls back to a warning and says it can't verify auth.

I've patched this locally: when the capability probe returns nothing, fall back to running `claude auth status` (with a 10s timeout), parse the JSON/text output to work out authenticated / unauthenticated / unknown, pull out `apiProvider` / `authMethod` / `email` when they're present, and label the Bedrock provider. "not logged in" style output maps to a clear error telling you to run `claude auth login`, and a timeout gets its own message.

### Expected behavior

Claude should show as authenticated and labelled "Amazon Bedrock", since the credentials are valid and Claude itself uses them fine.

### Actual behavior

T3 shows a warning — "Could not verify Claude authentication status from initialization result." — so a working provider looks unverified / unavailable. This doesnt stop requests from working.. But the model selector is disabled 

### Impact

Major degradation or frequent failure

### Version or commit

main @ b125b7635

### Environment

apps/server, Claude Agent provider. Managed Amazon Bedrock installs where SDK init returns no capabilities/account info.

### Logs or stack traces

```shell

```

### Screenshots, recordings, or supporting files

<img width="730" height="477" alt="Image" src="https://github.com/user-attachments/assets/2590356a-4653-4105-8a4b-321b2aae28c8" />

### Workaround

None from the UI — the provider actually works, the status is just wrong.

hadifarnoud:
I have a session working with claude code and at the same time t3 codes says auth issue

<img width="748" height="539" alt="Image" src="https://github.com/user-attachments/assets/f8ab9f3d-771d-427b-8078-f33b1558ed99" />

PR #7363, opened 2026-08-18
fix(desktop): inherit Claude Bedrock environment

huangkairan:
## Background

Related to #4928.

When Claude Code uses Amazon Bedrock, both the provider status probe and actual sessions depend on AWS credential environment variables. When T3 Code is launched from a GUI entry point such as Finder, Electron does not naturally inherit `AWS_BEARER_TOKEN_BEDROCK`, the AWS region, or the Claude Code Bedrock flags from the user's login shell. The existing desktop environment hydration only restored variables such as PATH, locale, and SSH settings.

As a result, Claude Code configuration could still identify Bedrock as the active backend while the T3 server child process lacked the credentials needed to call it. The provider probe could appear unverified, and actual requests could fail.

Issue #4928 also describes managed Bedrock environments where SDK initialization returns no account information. This PR does not change the server-side authentication fallback strategy; it specifically fixes credential inheritance at the desktop process boundary.

## Changes

- Restore a controlled allowlist of AWS Bedrock credentials, region settings, and Claude Code Bedrock flags from the login shell on macOS and Linux.
- Restore the same variables from the PowerShell profile on Windows while preserving values already inherited by the desktop process.
- Forward these variables to the Windows WSL backend through the existing `WSLENV` mechanism.
- Add behavior coverage for macOS, Windows, and Windows-to-WSL propagation.

## Verification

- `git diff --check` passes.
- Targeted Vitest cases were added, but could not be executed because the current environment repeatedly failed while downloading the monorepo dependencies.

Model: GPT-5.6 Sol
Harness: T3 Code Codex

### DIFF
--- a/apps/desktop/src/backend/DesktopBackendConfiguration.test.ts
+++ b/apps/desktop/src/backend/DesktopBackendConfiguration.test.ts
@@ -553,6 +553,40 @@ describe("DesktopBackendConfiguration", () => {
     }).pipe(Effect.scoped, Effect.provide(NodeServices.layer)),
   );
 
+  it.effect("resolveWsl forwards Claude Bedrock credentials across the WSL boundary", () =>
+    Effect.gen(function* () {
+      const previousBearerToken = process.env.AWS_BEARER_TOKEN_BEDROCK;
+      const previousRegion = process.env.AWS_REGION;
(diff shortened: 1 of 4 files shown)


huangkairan:
Hi, @t3dotgg @juliusmarminge  could you please review this PR when you have a moment? Thanks!

PR #7373, opened 2026-08-18
feat(swift-ios): link a build to its source T3 thread

saphid:
## What this adds

A development build can now say which T3 thread produced it, and hand you back to that thread.

Settings → About gains one **Source thread** row with three states:

- **Not recorded** — the build embedded no thread identity. Noninteractive, with a plain explanation. This is what every stock build shows.
- **Recorded but unresolved** — a thread was recorded, but this device cannot resolve it to exactly one local thread. Noninteractive, showing the build-time title (or a shortened identifier) plus "Not available on this device."
- **Openable** — the recorded thread exists here. Tapping it closes Settings and selects that existing thread.

## How the identity is embedded

Three optional Info.plist values expand from build settings: `T3SourceThreadID`, `T3SourceThreadEnvironmentID`, `T3SourceThreadTitle`, from `T3CODE_SOURCE_THREAD_ID` / `_ENVIRONMENT_ID` / `_TITLE`. Nothing is fetched at runtime.

The three settings are declared with **empty-string defaults** on the T3Code app target's Debug and Release configurations, so the knob is discoverable in the project. An unset setting expands to an empty string, and a misconfigured build can leave a literal `$(NAME)` placeholder behind — both are treated as "not recorded", so the feature fails closed and a stock build is unchanged. A build records its thread only when the value is passed explicitly, e.g. `xcodebuild … T3CODE_SOURCE_THREAD_ID=<uuid>`.

## Identity is preserved; nothing is created

`BuildSourceThreadResolver` mirrors `PlatformRouteResolver.thread`: match on the thread's `id` **or** its `wireID`, scope to the recorded environment, and refuse an ambiguous match when no environment was recorded. Resolution yields the thread's own local id, which is then handed to the existing `WorkspaceView.openThread` — the same selection path deep links already use. No creation path is touched, and an unresolvable record stays noninteractive rather than opening an error or an unrelated destination.

The rule is restated rather than imported because the Features layer does not reference the Platform layer anywhere in this target. That duplication is commented at the site and covered by tests on both sides of every branch; folding the two into one shared helper would be a reasonable follow-up if the deep-link rule changes.

## Scope

6 files, +418/−7:

| File | |
|---|---|
| `Features/Settings/BuildSourceThread.swift` | new — metadata reader, resolver, presentation states |
| `Features/Settings/SettingsView.swift` | the About row plus an `openThread` handoff parameter |
| `Features/Workspace/WorkspaceView.swift` | passes the handoff into the existing `openThread` |
| `Resources/Info.plist` | three `$(T3CODE_SOURCE_THREAD_*)` values |
| `T3Code.xcodeproj/project.pbxproj` | three empty user-defined settings, app target Debug + Release only |
| `Tests/FeatureTests/BuildSourceThreadTests.swift` | new — 12 tests |

The pbxproj change is additive and contained: three settings inserted alphabetically into the app target's two config blocks. No version, bundle id, team, entitlement, target, or file-membership change. Only the app target uses `Resources/Info.plist` — the widget and share extensions have their own — so the new keys do not reach them.

## Verification

**Focused tests — 12 passed, 0 failed, real exit 0.** `xcodebuild test -only-testing:T3CodeTests/BuildSourceThreadTests`, Debug, isolated temporary DerivedData, `** TEST SUCCEEDED **`; the count was confirmed from the result bundle rather than the legacy XCTest counter line. Coverage: plist reading and trimming; absent / empty / placeholder / non-string metadata; each optional field falling away independently; the not-recorded state; opening by local id; opening by `wireID` returning the **local** id; thread absent on device; the same id in a different environment never substituted; an ambiguous match without a recorded environment refused; an unambiguous one still opening; newest `updatedAt` winning; a blank title falling back to a shortened identifier.

**Simulator proof.** All three states were exercised against a disposable local server seeded with one project and one thread, and three raw (unedited, uncropped) screenshots were captured and compared against the stated behaviour. Two details worth naming: the openable row displays the **live** thread title rather than the string recorded at build time, which is the visible evidence that it resolved the thread instead of echoing its own metadata; and immediately after the tap the server projections still held exactly one thread and zero turns, confirming nothing was created.

**On-device acceptance.** Installed and accepted on the maintainer's physical device as personal Test build 88.

**Independent cross-provider review.** Not yet run. A GPT-5.6 Sol high review is scheduled against **this exact head** (`3acbf240420e0d1349814baf2caec29f74349954`) once provider quota resets at 2026-08-20T03:29Z. It has not happened yet, so no findings from it are claimed here; I will post the outcome on this PR.

Coordination trace: T3 thread B9EF7113-8094-49A1-A908-ED434CDAF91A · saphid/t3code-personal#85

---

> [!NOTE]
> **Low Risk**
> Settings-only navigation with fail-closed metadata parsing and conservative thread matching; no auth, sync, or thread-creation changes.
> 
> **Overview**
> Dev and Test builds can now **embed which T3 thread produced them** via optional `T3CODE_SOURCE_THREAD_*` build settings that land in **Info.plist** (`T3SourceThreadID`, environment, title). Empty or unresolved `$(…)` values are ignored so stock builds behave as **not recorded**.
> 
> **Settings → About** adds a **Source thread** row: static copy when nothing was recorded or the thread cannot be resolved on this device; when resolution succeeds, the row is tappable, **closes Settings**, and selects that thread through the same **`openThread`** path as deep links—using a Features-layer resolver that mirrors **`PlatformRouteResolver`** (id/wire id, environment scoping, ambiguity refusal, newest `updatedAt`).
> 
> Xcode gains empty defaults for the three settings on the app target; **12 unit tests** cover plist parsing and resolver branches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c2c2897a321b871df3b00ed55b4a4f44a0b49ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Link a build to its source T3 thread in the Settings view
> - Adds `BuildSourceThread`, which reads build-time source-thread metadata (`T3SourceThreadID`, `T3SourceThreadEnvironmentID`, `T3SourceThreadTitle`) from `Info.plist` at runtime, rejecting empty or unresolved `$(...)` placeholders.
> - Adds `BuildSourceThreadResolver`, which matches the recorded thread ID against a `FeatureSnapshot`, handling environment filtering, wireID-to-localID mapping, ambiguity rejection, and recency-based selection.
> - Adds a "Source thread" row to the Settings About section: tappable (closes Settings and opens the thread) when the thread is resolvable, or a static `SettingsValueRow` with status text otherwise.
> - Extends `SettingsNavigationRow` to support multi-line trailing values and `SettingsValueRow` to show an optional secondary detail string.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0c2c289.</sup>
> 
>


t3dotgg:
This seems weird and fragile to put in a hot path. Don't love the idea of this context existing at all in the app, why should it have any idea what thread made it when most of the builds that matter happen with no thread involved at all?

Is this just to make it easier to open the thread you're working on in the build itself? Not sure that use case is important enough to me, I actively DON'T want to use the app to work on changes on the dev build

saphid:
Yes — opening the producing thread from the personal build was the intended use case. Agreed that it does not justify adding build-specific context to this path, especially when normal builds have no producing thread. I am withdrawing this approach and will not spend another proof/build round on it.

saphid:
> This seems weird and fragile to put in a hot path. Don't love the idea of this context existing at all in the app, why should it have any idea what thread made it when most of the builds that matter happen with no thread involved at all?
> 
> Is this just to make it easier to open the thread you're working on in the build itself? Not sure that use case is important enough to me, I actively DON'T want to use the app to work on changes on the dev build

I made this for myself, but actually I do want something like this if we are keeping the session data stored somewhere, this is becoming a more common use case to store session data alongside the PR that it made, but happy to put more thought into exactly how to do that cleaner and better. Closing this for now

PR #7376, opened 2026-08-18
fix(server): settle turns when provider interrupts fail

### LINKED ISSUE #7368: Stop button leaves a thread stuck "Working" forever when the provider's abort fails
**Symptom**

Clicking the red stop button in the composer appears to do nothing. The thread keeps showing "Working for NNm" indefinitely — in my case 54 minutes — and never returns to an idle state. Nothing is surfaced to explain why. The only way out is restarting the app, which kills the thread.

**What's actually happening**

The button itself is fine — it isn't disabled and the click handler does fire. The failure is downstream, and it's silent at every layer:

1. `apps/server/src/orchestration/Layers/ProviderCommandReactor.ts` → `processTurnInterruptRequested` calls `providerService.interruptTurn({ threadId })` with no error handling (on `main` today this is the bare call). If it fails, the thread's session is left with its `activeTurnId` still set, so the UI keeps rendering "Working".
2. The opencode adapter's `interruptTurn` issues `session.abort` against the OpenCode server. If that server is gone or never came up, the call fails.
3. On the web side the interrupt action is dispatched with `reportFailure: false`, so the error never reaches the user.

Net effect: provider abort fails → the reactor propagates the error → nothing clears the active turn → the UI shows a permanently running turn, with no error shown.

**Why the abort was failing for me**

My OpenCode server wasn't starting reliably. The banner read:

> Failed to execute OpenCode CLI health check: Timed out waiting for OpenCode server start after 5000ms.

That 5s bound was `DEFAULT_OPENCODE_SERVER_TIMEOUT_MS` in `apps/server/src/provider/opencodeRuntime.ts`. On my machine the opencode binary is ~143MB and takes ~1.9s just for `--version`, so a cold start under load regularly exceeded 5s. **This part is already fixed on `main`** — the constant is now `30_000` — so I'm only mentioning it as the trigger that exposed the stuck-thread bug, not as something to change.

The stuck-thread behaviour is the real issue, and it will reproduce for any provider whose abort fails for any reason, not just opencode.

**Reproduce**

1. Use a provider whose backing CLI/server can die independently (opencode is easiest).
2. Start a long-running turn.
3. Kill the provider's server process out from under it, or let its start time out.
4. Press the stop button.

Expected: the turn stops, or an error explains why it couldn't.
Actual: the button does nothing visible and the thread shows "Working" forever.

**Suggested fix**

Make the interrupt path fail-safe in `processTurnInterruptRequested`: if `interruptTurn` fails, append a provider-failure activity, force `stopSession` for the thread, and clear the session's `activeTurnId` (the existing turn-start-failure helper already does exactly this). The thread then returns to idle and the next turn starts a fresh session, instead of the UI being stuck until an app restart.

Separately, the interrupt action dispatching with `reportFailure: false` means a user gets no signal at all when a stop fails — worth reconsidering for this action specifically.

**What I ran**

T3 Code (Alpha) **0.0.28** (build commit `ecb35f758399`, 2026-07-15) when first diagnosed; I am now on **v0.0.34-nightly.20260817.1120**, macOS arm64, provider = OpenRouter via the opencode driver (DeepSeek V4 Flash). Diagnosed on that build; I re-read `ProviderCommandReactor.ts` on `main` today (2026-08-18) and the interrupt call there still has no error handling, so I believe it's still present. I have not reproduced it end-to-end on a current nightly.

I applied the fail-safe described above locally and the stop button now reliably returns the thread to idle even when the provider's abort fails.

StiensWout:
The stop button could leave a thread showing Working forever when the provider abort failed. The asynchronous reactor left the active turn set, while the provider failure never became visible to the user.

Interrupt failure is now fail-safe: T3 records a provider failure activity, attempts to stop the provider session, and clears the projected active turn even if session cleanup also fails.

## Verification

- pnpm exec vp test run apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
- pnpm exec vp fmt --check apps/server/src/orchestration/Layers/ProviderCommandReactor.ts apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
- pnpm exec vp lint --report-unused-disable-directives apps/server/src/orchestration/Layers/ProviderCommandReactor.ts apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
- pnpm exec vp run --filter t3 typecheck
- git diff --check

Fixes #7368

Implemented with GPT-5.6 Sol through Codex in T3 Code.

---

> [!NOTE]
> **Medium Risk**
> Changes orchestration session state on interrupt failure paths; behavior is localized to ProviderCommandReactor with test coverage, but affects stop-button UX and projected thread state.
> 
> **Overview**
> Fixes threads that could stay in **Working** forever when **Stop** failed because the provider abort errored and the reactor never cleared `activeTurnId` or surfaced the failure.
> 
> **Turn interrupt** handling in `ProviderCommandReactor` is now fail-safe: if `interruptTurn` fails, the reactor tries `stopSession`, then uses the shared `setThreadSessionError` helper (renamed from the turn-start-only variant) to set session **error**, clear **activeTurnId**, and set **lastError**, and it records a `provider.turn.interrupt.failed` activity. Recovery still runs when session stop also fails.
> 
> A regression test covers the case where both interrupt and stop fail against an unavailable OpenCode server.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4eb952e4e73bcc2180a8a218cb3238a3012a4648. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Handle provider turn interrupt failures by settling thread session with error state
> - When `providerService.interruptTurn` fails, the handler now attempts `providerService.stopSession`, marks the thread session as error with the failure detail, and appends a `provider.turn.interrupt.failed` activity.
> - If `stopSession` also fails, a warning is logged with both the stop cause and the original interrupt cause; the thread is still settled.
> - Renames the internal `setThreadSessionErrorOnTurnStartFailure` utility to `setThreadSessionError` and reuses it across both turn start and interrupt failure paths in [ProviderCommandReactor.ts](https://github.com/pingdotgg/t3code/pull/7376/files#diff-0630e9ad70d5dcd1c8586d29297f9b7d5f5436cc8e10fc88dc22acc511f1450d).
> - Interrupt-only causes (e.g. fiber cancellation) still short-circuit as interrupts without triggering recovery.
> - Behavioral Change: provider turn interrupts that previously propagated errors now produce a settled error session state instead.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4eb952e.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
@@ -2486,6 +2486,79 @@ describe("ProviderCommandReactor", () => {
     });
   });
 
+  it("settles the thread when provider interrupt and session cleanup fail", async () => {
+    const harness = await createHarness();
+    const now = "2026-01-01T00:00:00.000Z";
+    harness.interruptTurn.mockImplementationOnce(
+      () =>
+        Effect.fail(
+          new ProviderAdapterRequestError({
+            provider: "opencode",
+            method: "session.abort",
+            detail: "OpenCode server is unavailable.",
+          }),
+        ) as never,
+    );
+    harness.stopSession.mockImplementationOnce(
+      () =>
+        Effect.fail(
+          new ProviderAdapterRequestError({
+            provider: "opencode",
+            method: "session.stop",
+            detail: "OpenCode server is unavailable.",
+          }),
+        ) as never,
+    );
+
+    await harness.runEffect(
+      harness.engine.dispatch({
+        type: "thread.session.set",
+        commandId: CommandId.make("cmd-session-set-before-failed-interrupt"),
+        threadId: ThreadId.make("thread-1"),
+        session: {
+          threadId: ThreadId.make("thread-1"),
+          status: "running",
+          providerName: "opencode",
+          runtimeMode: "approval-required",
+          activeTurnId: asTurnId("turn-1"),
+          lastError: null,
+          updatedAt: now,
+        },
+        createdAt: now,
+      }),
+    );
+
+    await harness.runEffect(
+      harness.engine.dispatch({
+        type: "thread.turn.interrupt",
+        commandId: CommandId.make("cmd-turn-interrupt-provider-failure"),
+        threadId: ThreadId.make("thread-1"),
+        turnId: asTurnId("turn-1"),
+        createdAt: now,
+      }),
+    );
+
+    await harness.drain();
+
+    expect(harness.interruptTurn).toHaveBeenCalledWith({ threadId: "thread-1" });
+    expect(harness.stopSession).toHaveBeenCalledWith({ threadId: "thread-1" });
+    const readModel = await harness.readModel();
+    const thread = readModel.threads.find((entry) => entry.id === ThreadId.make("thread-1"));
+    expect(thread?.session).toMatchObject({
+      status: "error",
+      activeTurnId: null,
+      lastError: "OpenCode server is unavailable.",
+    });
+    expect(
+      thread?.activities.find((activity) => activity.kind === "provider.turn.interrupt.failed"),
+    ).toMatchObject({
+      turnId: "turn-1",
+      payload: {
+        detail: "OpenCode server is unavailable.",
+      },
+    });
+  });
+
   it("starts a fresh session when only projected session state exists", async () => {
     const harness = await createHarness();
     const now = "2026-01-01T00:00:00.000Z";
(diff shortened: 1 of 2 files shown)


t3dotgg:
> [!NOTE]

Closing in favor of [#7412](https://github.com/pingdotgg/t3code/pull/7412). It already fixes the same provider interrupt failure and clears the stuck thread state.

Issue #5436, opened 2026-08-05
[Bug]: Mobile: message queued while agent is working is never sent if the app goes to background — only delivered on next app open


iamtoled:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/mobile

### Steps to reproduce

1. Start a thread and wait for the agent to be actively working on a turn.
2. Send a follow-up message while the turn is in progress — it gets queued (locally, it seems).
3. Lock the phone or switch to another app.
4. Wait for the agent to finish its turn.

### Expected behavior

The queued message is delivered to the agent once the current turn settles, regardless of whether the mobile app is foregrounded. Queuing should be owned by the server/backend, not the client.

### Actual behavior

The queued message stays pending indefinitely. It is only actually sent when I reopen the app. If queuing happens client-side, it silently defeats the purpose of queuing on mobile: you queue a follow-up, put your phone away, and come back later to find the agent idle and the message never delivered.

### Related

Not a duplicate of #231 / #4550 (those are feature requests for Queue/Steer modes) — this is about delivery reliability of the existing queue behavior when the mobile app is backgrounded.

nhorto:
Partial fix landed in #6543 — the foreground half of this is fixed on `main`, but the backgrounded case this issue is actually about is not.

## What #6543 fixed

The delivery rule in `apps/mobile/src/state/thread-outbox-model.ts` used to be:

```ts
return input.environmentConnected && !input.threadBusy ? "send" : "wait";
```

#6543 dropped the `threadBusy` term, so a message now goes out immediately while a turn is running instead of being held on the device.

That matters more than the steer-vs-queue framing, because the `"wait"` branch in `useThreadOutboxDrain` schedules no retry — it is a bare `continue` (`use-thread-outbox-drain.ts:319-321`). Nothing wakes it. It only re-evaluates when a React dependency changes, so a message that hit `"wait"` sat there until an unrelated re-render shook it loose.

`/compact` hit this hard: the Claude adapter reports compaction as `"waiting"` (`ClaudeAdapter.ts:3119`), which `ProviderRuntimeIngestion` maps to session status `"running"`. The busy gate therefore stayed engaged for the entire compaction, and the follow-up message stayed on the device until the thread was reopened.

[...]

PR #7405, opened 2026-08-18
refactor(mobile): drop the ignored busy input from outbox delivery

fomoles:
## What Changed

`resolveThreadOutboxDeliveryAction` no longer takes `threadBusy`. The drain no longer computes it at the call site, and the delivery-action tests no longer pass it.

No behavior change: every input the resolver accepted before returns the same action.

## Why

#6543 made connectivity the delivery boundary for the mobile outbox and dropped the `!input.threadBusy` term that was the only reader of that field. What stayed behind is a signature that still advertises a busy gate:

- `thread-outbox-model.ts` declares `readonly threadBusy: boolean`.
- `use-thread-outbox-drain.ts` recomputes `thread?.session?.status === "running" || thread?.session?.status === "starting"` on every drain pass and passes it in.
- `thread-outbox.test.ts` passes `threadBusy: true` in the steering cases, which reads as coverage for "a running turn does not hold a queued message back" but asserts nothing about it.

Read today, the resolver looks like a running turn can still gate a queued message. It cannot. Removing the input makes the contract say what the function does, and a short comment records why connectivity alone decides delivery so the gate is not reintroduced by inspection.

Called out as a leftover in #5436.

## Checklist

- [x] No UI changes

## Verification

- `vp test run apps/mobile/src/state/thread-outbox.test.ts` — 20 passed
- `vp run typecheck` from `apps/mobile` — clean
- `vp lint` on the three changed files — clean
- `vp fmt --check` on the three changed files — clean

Every `threadBusy` reference in the repo is in the three files changed here, and the removed field had no reader, so the resolver returns the same action for every input it accepted before.

Authored with Claude Opus 5 via Claude Code.

### DIFF
--- a/apps/mobile/src/state/thread-outbox-model.ts
+++ b/apps/mobile/src/state/thread-outbox-model.ts
@@ -148,12 +148,15 @@ export function threadOutboxRetryDelayMs(attempt: number): number {
 
 export type ThreadOutboxDeliveryAction = "wait" | "remove" | "send";
 
+/**
+ * Connectivity is the delivery boundary: a running turn does not hold a queued
+ * message back, because a message sent during one steers it.
+ */
 export function resolveThreadOutboxDeliveryAction(input: {
   readonly isCreation: boolean;
   readonly threadExists: boolean;
   readonly shellStatus: EnvironmentShellStatus;
   readonly environmentConnected: boolean;
-  readonly threadBusy: boolean;
 }): ThreadOutboxDeliveryAction {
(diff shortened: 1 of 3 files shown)


t3dotgg:
> [!NOTE]

Closing this standalone cleanup in the backlog pass. [#6543](https://github.com/pingdotgg/t3code/pull/6543) already changed delivery behavior so busy threads can receive steering messages. This only removes the unused argument and matching test inputs. It can be included in the next outbox change.

Issue #7128, opened 2026-08-15
[Bug]: Finished threads stay Working after delayed task progress


matsvarn:
## Steps to reproduce

1. Start a provider task in a thread.
2. Record `task.updated` with status `idle`.
3. Deliver a delayed `task.progress` for the same task without a status.
4. Read `ThreadBackgroundLiveness` for the thread.

Minimal unit reproduction:

```ts
const liveness = ThreadBackgroundLiveness.make();
liveness.recordTaskLiveness({
  threadId: "thread",
  taskId: "task",
  taskType: undefined,
  status: undefined,
  kind: "started",
});
liveness.recordTaskLiveness({
  threadId: "thread",
  taskId: "task",
  taskType: undefined,
  status: "idle",
  kind: "updated",
});
liveness.recordTaskLiveness({
  threadId: "thread",
  taskId: "task",
  taskType: undefined,
  status: undefined,
  kind: "progress",
});
```

## Expected behavior

The final liveness value is `null`. Descriptive progress must not restart an idle task.

## Actual behavior

The final value is `"working"`. The desktop and web clients show “Background work running” and a Working sidebar state. The agent list contains zero live agents.

## Impact

Minor bug or occasional failure.

## Version or commit

`0.0.34-nightly.20260813.1086`

The same liveness code is present on current `main`.

## Environment

macOS desktop app with the Codex provider.

## Supporting evidence

A real task became idle at `2026-08-14T20:29:34.717Z`. A delayed status-free progress event arrived at `2026-08-14T20:29:53.327Z`. The thread then stayed Working with zero live agents.

## Workaround

Restarting T3 Code clears the in-memory liveness registry. The stale state can return after another delayed progress event.

maslinedwin:
Fix is in #7172.

PR #7468, opened 2026-08-19
fix(server): status-free task updates no longer revive an idle task

shoemoney:
#7128 / #7172 stopped a delayed status-free `progress` event from putting an already-dropped task back in the live set. The same hole is still open for `updated`: the guard only tests `input.kind === "progress"`, so a status-free `updated` falls through to `drop()` + `bucket.add()` and revives a task that had gone idle. The sidebar status pill then stays pinned on "working" until the session exits.

That input is real, not hypothetical. `ClaudeAdapter`'s `task_updated` case emits `task.updated` with `...(status ? { status } : {})`, so no `status` field is sent whenever the patch carries only `description`, `error`, `end_time` or `is_backgrounded` — and `ProviderRuntimeIngestion` forwards it straight through as `kind: "updated"` with `status: undefined`.

The fix extends the existing guard to cover `updated` as well. `started` is deliberately left out so a genuine restart still revives the task.

Two tests: one for the bug, one pinning that `kind: "started"` after idle still returns `"working"` so the guard cannot over-correct.

Verified by mutation, not just by a green run — dropping the `updated` clause back out fails the new test with `expected 'working' to be null`, and restoring it returns 9/9 green. Full `apps/server` suite: 2614 passed, 7 skipped, exit 0.

---

> [!NOTE]
> **Low Risk**
> Small, targeted change to in-memory sidebar liveness with regression tests and no auth or persistence impact.
> 
> **Overview**
> Extends the **status-free lifecycle guard** in `ThreadBackgroundLiveness` so delayed **`updated`** events (no `status`, e.g. description-only patches from `ClaudeAdapter`) cannot re-add a task that already went **idle**, matching the existing behavior for **`progress`** (#7128, #7172).
> 
> **`kind: "started"`** is unchanged so a real restart still shows the thread as **working**. Adds tests for the `updated` bug and for restart-after-idle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit beb1fa64c2b775b27525f1665a1586351dbe15a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix status-free `updated` events to no longer revive an idle task
> In [`ThreadBackgroundLiveness`](https://github.com/pingdotgg/t3code/pull/7468/files#diff-b0c25dad649cb92e4b2e2045f18da810e4c2ecce6520501428293b115d9940cd), the no-restart guard previously only applied to `progress` events with no status. It now also applies to `updated` events with no status, treating them as description ticks rather than task revivals. Only `started` events can revive an idle task.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized beb1fa6.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/ThreadBackgroundLiveness.test.ts
+++ b/apps/server/src/orchestration/ThreadBackgroundLiveness.test.ts
@@ -28,6 +28,58 @@ describe("ThreadBackgroundLiveness", () => {
     expect(liveness.getThreadBackgroundLiveness("thread")).toBeNull();
   });
 
+  it("does not let status-free updated restart an idle task", () => {
+    const liveness = ThreadBackgroundLiveness.make();
+    liveness.recordTaskLiveness({
+      threadId: "thread",
+      taskId: "task",
+      taskType: undefined,
+      status: undefined,
+      kind: "started",
+    });
+    liveness.recordTaskLiveness({
+      threadId: "thread",
+      taskId: "task",
+      taskType: undefined,
+      status: "idle",
+      kind: "updated",
+    });
+    liveness.recordTaskLiveness({
+      threadId: "thread",
+      taskId: "task",
+      taskType: undefined,
+      status: undefined,
+      kind: "updated",
+    });
+    expect(liveness.getThreadBackgroundLiveness("thread")).toBeNull();
+  });
+
+  it("a genuine restart (kind: started) still revives an idle task", () => {
+    const liveness = ThreadBackgroundLiveness.make();
+    liveness.recordTaskLiveness({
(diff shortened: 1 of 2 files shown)


shivamhwp:
gpt-5.6-sol on behalf of shivamhwp.

Thank you for catching this task-liveness edge case. We are closing this PR because #7172 has already landed the same protection on main. Status-free progress no longer revives an idle task, and main includes a focused test for that exact behavior.

The fix is present and tested, so this branch is no longer needed.

PR #7499, opened 2026-08-19
fix(server): resolve partial clone repository identities

marinushv:
## What

Resolve repository identity for Git partial clones whose `git remote -v` fetch line includes transport metadata such as `[blob:none]`.

## Why

The parser previously required `(fetch)` to be the end of the line. For partial clones, that left `repositoryIdentity` empty, so projects on those machines could not group with normal clones of the same repository.

## Validation

- `vp test apps/server/src/project/RepositoryIdentityResolver.test.ts` (8 passed)

---

> [!NOTE]
> **Low Risk**
> Narrow parser change in project identity resolution with a targeted test; no auth, data, or broad API surface impact.
> 
> **Overview**
> **Repository identity resolution** now tolerates Git partial-clone metadata on `git remote -v` lines (e.g. `[blob:none]` after `(fetch)`), so those workspaces get the same canonical identity as full clones.
> 
> `parseRemoteFetchUrls` in `RepositoryIdentityResolver` extends the line regex to accept zero or more bracketed annotations after the fetch/push direction while still extracting remote name and URL. A new integration test configures `remote.origin.promisor` and `partialclonefilter` and asserts the expected GitHub canonical key and `origin` locator.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit adbf94c81b0ee64f1da092fa7b12bc14e28d13a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/server/src/project/RepositoryIdentityResolver.test.ts
+++ b/apps/server/src/project/RepositoryIdentityResolver.test.ts
@@ -61,6 +61,26 @@ it.layer(NodeServices.layer)("RepositoryIdentityResolverLive", (it) => {
     }).pipe(Effect.provide(RepositoryIdentityResolver.layer)),
   );
 
+  it.effect("resolves a promisor remote reported with a partial-clone annotation", () =>
+    Effect.gen(function* () {
+      const fileSystem = yield* FileSystem.FileSystem;
+      const cwd = yield* fileSystem.makeTempDirectoryScoped({
+        prefix: "t3-repository-identity-partial-clone-test-",
+      });
+
+      yield* git(cwd, ["init"]);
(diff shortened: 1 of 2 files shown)


lnieuwenhuis:
Maintainer verification: the tolerant fetch/push suffix is conservative — name/url/direction capture is unchanged and normal lines parse exactly as before. One test-strength caveat (non-blocking): the regression relies on the installed git annotating remote -v with [blob:none]; if a git ever omits that annotation the test passes without exercising the regex. A synthetic-stdout unit case would make it airtight, but I would not hold the PR for it.

PR #7535, opened 2026-08-19
fix(server): skip disabled text generation fallbacks

### LINKED ISSUE #7533: [Bug]: Text generation falls back to a disabled provider — fallback reads legacy `providers` map instead of `providerInstances`
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Leave **Settings → Text generation model** unset (`textGenerationModelSelection` absent/null in `~/.t3/userdata/settings.json`).
2. In **Settings → Providers**, disable Codex. Keep Claude (and/or OpenCode) enabled and authenticated. The UI writes this to `providerInstances.codex.enabled = false`:
   ```json
   "providerInstances": {
     "codex": { "driver": "codex", "enabled": false, "config": { ... } }
   }
   ```
   Note there is no top-level `providers` key in the settings file.
3. Do not install the Codex CLI, or make sure `codex` is not on `PATH`.
4. Start a new thread and send a first message.

### Expected behavior

Text generation uses one of the providers that is actually enabled (Claude / OpenCode). Disabling a provider in Settings should mean it is never selected for anything.

### Actual behavior

Every text generation call targets the **disabled** Codex provider and dies immediately:

```
maybeGenerateThreadTitleForFirstTurn
  CodexTextGeneration.generateThreadTitle
    runCodexJson.runCodexCommand
      TextGenerationError: Text generation failed in generateThreadTitle:
      Failed to spawn Codex CLI process
  -> logWarning "provider command reactor failed to generate or rename thread title"
```

Fails in ~7 ms, is swallowed by `Effect.catchCause`, and the thread keeps its raw `titleSeed` forever. Same for `generateBranchName`, `generateCommitMessage`, and `generatePrContent`.

### Root cause

The two halves of the provider-enabled check read different maps.

`resolveTextGenerationProvider` (`apps/server/src/serverSettings.ts:234`) correctly detects that the configured selection is disabled, because `isModelSelectionProviderEnabled` (`packages/shared/src/serverSettings.ts:34`) consults the **new** `providerInstances` map first:

```ts
const instanceConfig = settings.providerInstances[selection.instanceId];
if (instanceConfig !== undefined) {
  return resolveProviderInstanceEnabled(instanceConfig);
}
```

It then hands off to `fallbackTextGenerationProvider` (`apps/server/src/serverSettings.ts:240`), which scans the **legacy** map instead:

```ts
const fallbackEntry = Object.entries(settings.providers).find(([, provider]) => provider.enabled);
```

`settings.providers` is not written by the Providers UI (which only patches `providerInstances`), so it decodes from schema defaults, where `codex` is the first key (`packages/contracts/src/settings.ts:659-660`) and `CodexSettings.enabled` defaults to `true` (`packages/contracts/src/settings.ts:312-313`).

Net effect: the fallback re-selects Codex, i.e. exactly the provider that was just rejected as disabled. With `providerInstances` populated by the UI and `providers` left at defaults, the fallback can never pick anything but Codex.

### Impact

### Version or commit

main @ f2d5fc91 (verified the code path is unchanged on `origin/main`); reproduced on T3 Code (Alpha) desktop build

### Environment

macOS 15 (Darwin 25.5.0), Claude CLI v2.1.235 + OpenCode v1.18.18 enabled, Codex disabled and not installed

### Logs or stack traces

From `~/.t3/userdata/logs/server.trace.ndjson`:

```
{"name":"CodexTextGeneration.generateThreadTitle","durationMs":7.34,
 "exit":{"_tag":"Failure","cause":"TextGenerationError: Text generation failed in
 generateThreadTitle: Failed to spawn Codex CLI process ..."}}
{"name":"maybeGenerateThreadTitleForFirstTurn","events":[[
 "provider command reactor failed to generate or rename thread title",
 {"threadId":"...","cwd":"...","cause":"TextGenerationError: ..."}]]}
```

After setting `textGenerationModelSelection` explicitly, the same thread regenerates fine:

```
processThreadTitleRegenerationSafely   Success
regenerateThreadTitle                  Success
ClaudeTextGeneration.generateThreadTitle  Success
thread.title.regeneration.complete     Success
```

### Workaround

Set the selection explicitly so the fallback never runs (app must be quit; it rewrites `settings.json` on exit):

```json
"textGenerationModelSelection": { "instanceId": "claudeAgent", "model": "claude-haiku-4-5" }
```

`claudeAgent` has no `providerInstances` entry, so the enabled check falls through to the legacy default of `true` and is accepted. Existing threads keep their seed titles until regenerated one by one.

### Suggested fix

Make `fallbackTextGenerationProvider` use the same precedence as `isModelSelectionProviderEnabled`: iterate `providerInstances` first (respecting `resolveProviderInstanceEnabled`), and only consult `settings.providers` for drivers with no instance entry. A regression test with `providerInstances.codex.enabled = false` and `providers` left at defaults should assert the fallback does not return `codex`.

Related but distinct: #5359 covers the same functions failing silently when the selected provider is *enabled but unhealthy*. This report is about a provider that is *explicitly disabled* still being selected.

StiensWout:
Disabling Codex in provider settings could still send title, branch, commit, and PR text generation to Codex because fallback selection read the legacy provider defaults.

Fallback selection now prefers enabled explicit provider instances and only uses a legacy default when that default instance has no explicit entry. Regression coverage includes the reported disabled-Codex state and an enabled custom-instance fallback.

Fixes #7533

Made with gpt-5.6-sol in T3 Code using Codex.

---

> [!NOTE]
> **Medium Risk**
> Changes server settings resolution for text generation routing; wrong fallback logic could send generation to the wrong provider, but scope is limited to fallback when the current selection is disabled.
> 
> **Overview**
> Fixes **text generation model fallback** so disabling a provider instance (e.g. Codex) no longer routes title/branch/commit/PR generation through that provider when the current selection becomes invalid.
> 
> `fallbackTextGenerationProvider` now **prefers the first enabled explicit `providerInstances` entry** (via `resolveProviderInstanceEnabled`) and only uses the legacy `providers` map when there is **no** explicit instance for that driver’s default instance id. Fallback selections set both **instance id** and **default model for that instance’s driver**, not just the driver name.
> 
> Regression tests cover a disabled default Codex instance falling back to Claude, and an enabled custom Claude instance winning over disabled Codex.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0d2c63985c81e9f25b36f6d78e40f24fbab9e2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix text generation fallback to skip disabled provider instances
> - Updates `fallbackTextGenerationProvider` in [serverSettings.ts](https://github.com/pingdotgg/t3code/pull/7535/files#diff-ec910be99e19e00c0e09b15484059b34164d71de8d0026f4ba98471bbbb3680e) to check `providerInstances` first, using `resolveProviderInstanceEnabled` to skip disabled instances.
> - Falls back to legacy `settings.providers` entries only when no corresponding `providerInstances` entry exists for that driver's default instance ID.
> - The resolved fallback now carries a concrete `instanceId` and derives the default model by driver via `DEFAULT_TEXT_GENERATION_MODEL_BY_PROVIDER`.
> - Behavioral Change: A disabled default provider instance is no longer selected as a fallback; an enabled instance is preferred when multiple instances exist.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a0d2c63.</sup>
> 
>

### DIFF
--- a/apps/server/src/serverSettings.test.ts
+++ b/apps/server/src/serverSettings.test.ts
@@ -1,6 +1,7 @@
 import * as NodeServices from "@effect/platform-node/NodeServices";
 import {
   DEFAULT_SERVER_SETTINGS,
+  DEFAULT_TEXT_GENERATION_MODEL_BY_PROVIDER,
   ProviderDriverKind,
   ProviderInstanceId,
   ServerSettings,
@@ -322,6 +323,62 @@ it.layer(NodeServices.layer)("server settings", (it) => {
           model: "openai/gpt-5.5",
         });
       }).pipe(Effect.provide(makeServerSettingsLayer())),
+  );
+
+  it.effect("does not fall back to a disabled default provider instance", () =>
+    Effect.gen(function* () {
+      const serverSettings = yield* ServerSettingsModule.ServerSettingsService;
+      const codexId = ProviderInstanceId.make("codex");
+      const claudeDriver = ProviderDriverKind.make("claudeAgent");
+      const claudeModel = DEFAULT_TEXT_GENERATION_MODEL_BY_PROVIDER[claudeDriver];
+      assert.isDefined(claudeModel);
+
+      const next = yield* serverSettings.updateSettings({
+        providerInstances: {
+          [codexId]: {
+            driver: ProviderDriverKind.make("codex"),
+            enabled: false,
+            config: {},
+          },
+        },
+      });
+
+      assert.deepEqual(next.textGenerationModelSelection, {
+        instanceId: ProviderInstanceId.make("claudeAgent"),
+        model: claudeModel,
+      });
+    }).pipe(Effect.provide(makeServerSettingsLayer())),
+  );
+
+  it.effect("prefers an enabled provider instance when falling back", () =>
+    Effect.gen(function* () {
+      const serverSettings = yield* ServerSettingsModule.ServerSettingsService;
+      const codexId = ProviderInstanceId.make("codex");
+      const claudeId = ProviderInstanceId.make("claude_openrouter");
+      const claudeDriver = ProviderDriverKind.make("claudeAgent");
+      const claudeModel = DEFAULT_TEXT_GENERATION_MODEL_BY_PROVIDER[claudeDriver];
+      assert.isDefined(claudeModel);
+
+      const next = yield* serverSettings.updateSettings({
+        providerInstances: {
+          [codexId]: {
+            driver: ProviderDriverKind.make("codex"),
+            enabled: false,
+            config: {},
+          },
+          [claudeId]: {
+            driver: claudeDriver,
+            enabled: true,
+            config: {},
+          },
+        },
+      });
+
+      assert.deepEqual(next.textGenerationModelSelection, {
+        instanceId: claudeId,
+        model: claudeModel,
+      });
+    }).pipe(Effect.provide(makeServerSettingsLayer())),
   );
 
   it.effect("preserves enabled text generation selections for non-built-in drivers", () =>
--- a/apps/server/src/serverSettings.ts
+++ b/apps/server/src/serverSettings.ts
@@ -15,11 +15,13 @@ import {
   DEFAULT_TEXT_GENERATION_MODEL_BY_PROVIDER,
   DEFAULT_MODEL_BY_PROVIDER,
   DEFAULT_SERVER_SETTINGS,
+  defaultInstanceIdForDriver,
   type ModelSelection,
   type ProviderInstanceConfig,
   type ProviderInstanceEnvironmentVariable,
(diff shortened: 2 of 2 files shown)


StiensWout:
[gpt-5.6-sol] RESPONDING ON BEHALF OF WOUT:
---
This is the intended behavior change from #7533, not an unverified side effect. The server settings service tests exercise both fallback paths directly: a disabled explicit Codex default falls through to the unshadowed enabled Claude default, and an enabled custom provider instance wins before legacy defaults. The full CI test job and the focused local test pass on the latest commit.

PR #7540, opened 2026-08-19
fix(server): stop losing OpenCode providers to truncated CLI output

### LINKED ISSUE #7539: OpenCode providers silently dropped when `opencode models --verbose` output is large
### What happens

Providers I configure in `~/.config/opencode/opencode.json` don't show up in T3 Code's model picker, even though OpenCode itself lists them fine. In my case two local llama.cpp providers (`llama-local`, `llama-remote`) and `google-vertex` are missing, and the OpenCode card says "3 upstream providers connected" when there are 6.

`~/.t3/caches/opencode.json` holds exactly the first 29 model slugs of `opencode models --verbose`, in output order, stopping at `google/gemini-3.1-pro-preview-customtools` — byte 32540 of the 110KB the CLI actually prints. Everything after that offset is gone. Providers are dropped purely because they're printed last.

### Why

`loadInventoryFromCli` folds over the child's stdout with `collectStreamAsString`. That backpressures the process once the OS pipe fills. OpenCode's CLI ends in `process.exit()` without waiting for stdout to flush, so it drops whatever is still queued and exits 0 (their bug, and a recurring one — anomalyco/opencode#29330, #14948, #29866, #30831). We then parse a prefix and report a partial inventory as if it were complete: `parseModelsCliOutput` silently discards the half-written trailing record, and `providerModelsFromSettings` filters to `providerList.connected`, so providers with no surviving model vanish without a warning.

Reading the same command with plain `child_process` and a `data` handler gets all 110916 bytes every time — we only lose data because we read slowly.

### Reproduce

Any OpenCode config big enough that `opencode models --verbose` exceeds the pipe buffer (mine is 110KB). It's a race, so it bites on a cold OpenCode cache, when the CLI is slowest:

```
opencode models --verbose | wc -c                      # 110916
opencode models --verbose | { sleep 6; cat; } | wc -c   #  56221, exit 0
```

Deterministic version of what our reader does — child writes 200KB and exits without flushing, consumer takes 30ms per chunk:

```
unbuffered fold: 58000 / 200000
```

### Expected

Providers configured in OpenCode show up in the picker. Failing that, a truncated inventory should surface as an error instead of silently showing a subset.

NitayRabi:
Closes #7539.

Providers I configure in OpenCode were missing from the picker — two local llama.cpp providers and `google-vertex`, with the card reporting "3 upstream providers connected" out of 6. The cache held exactly the first 29 model slugs of `opencode models --verbose`, ending at byte 32540 of the 110KB the CLI prints. Providers were dropped only because they're printed last.

Two things go wrong:

**We read too slowly.** `collectStreamAsString` folds straight over the child's stdout, which backpressures it once the OS pipe fills. OpenCode's CLI ends in `process.exit()` without waiting for stdout to flush, so it drops the queued tail and exits 0. Buffering the stream unbounded fixes it — and costs nothing, since the fold accumulates the whole output anyway. Reading the same command with plain `child_process` gets all 110916 bytes every time.

**We don't notice.** `parseModelsCliOutput` silently discarded the half-written trailing record, so a cut looked like a clean, shorter listing. It now reports `truncated`, which reuses the existing retry and otherwise fails instead of handing back an inventory that's missing providers. Bodyless output (`opencode models` without `--verbose`) is not flagged.

The upstream flush bug is real and recurring (anomalyco/opencode#29330, #14948, #29866, #30831) but we shouldn't be the ones losing the data.

### Testing

- `vp test run src/provider/ src/stream/` — 521 passed, 6 skipped
- Deterministic check: child writes 200KB then exits without flushing, consumer takes 30ms/chunk — unbuffered fold gets 58000/200000, `collectStreamAsString` gets 200000/200000 across repeated runs
- Against the real CLI on a cold cache: unbuffered truncated at 32540 (matching the bad cache byte-for-byte), buffered got all 110916 in 5/5 runs
- All 6 providers now appear

---

> [!NOTE]
> **Medium Risk**
> Changes OpenCode provider inventory loading behavior (retry + hard failure on persistent truncation); broader `collectStreamAsString` buffering affects all provider CLI capture but is low-risk given full-output collection anyway.
> 
> **Overview**
> Fixes **missing OpenCode providers** in the model picker when `opencode models --verbose` stdout is cut off but still exits 0.
> 
> **`collectStreamAsString`** now drains child stdout through an **unbounded buffer** before folding, so a slow consumer does not backpressure the pipe and lose the tail when the CLI exits without flushing.
> 
> **`parseModelsCliOutput`** returns a **`truncated`** flag when verbose output ends on an incomplete slug/JSON pair (including a trailing slug with no body or bad final JSON). **Slug-only** listings without `--verbose` are not treated as truncated.
> 
> **`loadInventoryFromCli`** **retries** the models command after 1s on truncation (same as transient failures) and returns **`OpenCodeRuntimeError`** if output is still incomplete, instead of caching a partial provider list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0d0093e45254ee46c148aa1c28daae184dbd258. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix truncated CLI output causing OpenCode providers to be silently dropped
> - Adds a `truncated` flag to `parseModelsCliOutput` in [opencodeRuntime.ts](https://github.com/pingdotgg/t3code/pull/7540/files#diff-b36408ac400332f121d5c4711ccfe4fd4049772d0f302343ee303670f4198ea8), set when output ends mid-record (trailing slug with no JSON body, or unparseable final JSON).
> - `loadInventoryFromCli` now retries the models command after 1s when output is truncated, and returns an `OpenCodeRuntimeError` if truncation persists, rather than proceeding with a partial inventory.
> - [providerSnapshot.ts](https://github.com/pingdotgg/t3code/pull/7540/files#diff-62373f674410fb63637739f0efeb07b9ef79b579393ae2adf51567de0b812b26) buffers the incoming stream unbounded before collection to prevent partial reads.
> - Behavioral Change: previously, truncated model output was silently accepted as a complete inventory; now it triggers a retry and an error on repeated truncation.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c0d0093.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/opencodeRuntime.cliParsers.test.ts
+++ b/apps/server/src/provider/opencodeRuntime.cliParsers.test.ts
@@ -83,6 +83,49 @@ describe("parseModelsCliOutput", () => {
     NodeAssert.ok(provider.models["claude-haiku-4-5"]);
   });
 
+  it("flags output cut after a slug line as truncated", () => {
+    // How a slow reader sees it: the CLI exits 0 mid-listing, right after a header.
+    const stdout = [
+      "anthropic/claude-sonnet-4-5",
+      JSON.stringify({ id: "claude-sonnet-4-5", providerID: "anthropic", name: "Sonnet 4.5" }),
+      "openai/gpt-4o",
+      "",
+    ].join("\n");
+
+    const result = parseModelsCliOutput(stdout);
+    NodeAssert.equal(result.truncated, true);
+  });
+
+  it("flags output cut inside a model body as truncated", () => {
+    const stdout = [
+      "anthropic/claude-sonnet-4-5",
+      JSON.stringify({ id: "claude-sonnet-4-5", providerID: "anthropic", name: "Sonnet 4.5" }),
+      "openai/gpt-4o",
+      '{ "id": "gpt-4o", "providerID": "op',
+    ].join("\n");
+
+    const result = parseModelsCliOutput(stdout);
+    NodeAssert.equal(result.truncated, true);
+  });
+
+  it("does not flag complete output as truncated", () => {
+    const stdout = [
+      "anthropic/claude-sonnet-4-5",
+      JSON.stringify({ id: "claude-sonnet-4-5", providerID: "anthropic", name: "Sonnet 4.5" }),
+      "",
+    ].join("\n");
+
+    NodeAssert.equal(parseModelsCliOutput(stdout).truncated, false);
+    NodeAssert.equal(parseModelsCliOutput("").truncated, false);
+  });
+
+  it("does not flag a bodyless listing as truncated", () => {
+    // `opencode models` without `--verbose` prints slugs only.
+    const stdout = ["anthropic/claude-sonnet-4-5", "openai/gpt-4o", ""].join("\n");
+
+    NodeAssert.equal(parseModelsCliOutput(stdout).truncated, false);
+  });
+
   it("handles Windows-style CRLF line endings", () => {
     const stdout =
       "anthropic/claude-sonnet-4-5\r\n" +
--- a/apps/server/src/provider/opencodeRuntime.ts
+++ b/apps/server/src/provider/opencodeRuntime.ts
@@ -211,6 +211,12 @@ export function parseModelsCliOutput(stdout: string): {
     { readonly id: string; readonly name: string; readonly models: { [key: string]: Model } }
   >;
   readonly connected: ReadonlyArray<string>;
+  /**
+   * The output ended mid-record, so providers printed after the cut are missing. The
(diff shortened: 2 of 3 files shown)


t3dotgg:
> [!NOTE]

Closing this as replaced by [#8480](https://github.com/pingdotgg/t3code/pull/8480). Production OpenCode inventory now uses the `provider.list`, `app.agents`, and `app.skills` HTTP APIs, so it no longer depends on truncated CLI output or a retry in the CLI parser. This closure is part of an automated pass based on our investigation.

PR #7542, opened 2026-08-19
fix(server): load MCP servers declared by Claude plugins

paul-vd:
## Problem

A Claude plugin that ships an `.mcp.json` looks installed but its tools never appear. The Agent SDK reports the plugin in its `system/init` handshake and loads its skills, commands, and agents — but only the interactive CLI registers the MCP servers the plugin declares. Every SDK consumer hits this, so a user with the Linear or Vercel plugin enabled gets no Linear or Vercel tools in T3, while the same plugin works in `claude` at the terminal.

Reproduced outside T3 with the same CLI binary — `claude -p --output-format stream-json` reports five plugins and zero `plugin:*` MCP servers, with or without `--setting-sources`, and with `--plugin-dir` too. Nothing T3 passes causes it.

## Fix

`ClaudePlugins.ts` resolves the servers from disk — `enabledPlugins` merged across the user, project, and local settings files, install paths from `plugins/installed_plugins.json`, then each plugin's `.mcp.json` (both the documented `mcpServers` wrapper and the bare map, since official plugins ship both). `ClaudeAdapter` applies them with `Query.setMcpServers` once the session is up.

Three details that are load-bearing:

- **Names must be `plugin:<plugin>:<server>`.** Stored MCP OAuth tokens are keyed by server name, so registering the same URL under any other name comes up `needs-auth` instead of reusing the token the user already has. The launch-time `mcpServers` option can't be used at all here — it silently drops names starting with `plugin:`.
- **Registration runs detached.** The control call doesn't resolve until every named server has connected — a consistent ~6s, and it's the first call that pays it. That's not worth adding to a turn, and the tools sit behind tool search, which reads live state.
- **`${CLAUDE_PLUGIN_ROOT}` is expanded here.** The control-channel path bypasses the CLI's plugin loader, which normally does that substitution. Verified by capturing the child's argv: it receives the literal placeholder. Five official marketplace plugins (`fakechat`, `imessage`, `discord`, `telegram`, `laravel-boost`) build their command with it and would otherwise start in a directory that doesn't exist.

`t3-code` is deliberately left out of the call: the SDK merges dynamic servers rather than replacing them, and re-sending a connected server tears down and reconnects its transport.

`resolveClaudeConfigDirPath` moved from `ClaudeSkills.ts` to `ClaudeHome.ts` so both discoverers share one config-dir precedence rule.

## Verification

With the Linear plugin enabled, the session goes from no `plugin:*` server at all to `plugin:linear:linear` connected, with all 63 `mcp__plugin_linear_linear__*` tools present in its tool list. `reloadPlugins()` was tried as the alternative and rejected — it registers the servers but they never reach the session's tool list.

No UI surface, so no screenshots. Provider-shaped but Claude-specific: plugins are a Claude Code concept, so Codex, Cursor, Grok, and OpenCode need no decision here. A user with no plugin declaring an MCP server takes zero extra work.

Six new tests: server naming across both `.mcp.json` shapes, `${CLAUDE_PLUGIN_ROOT}` expansion in command/args/env, disabled/absent/server-less plugins, project settings overriding user settings, malformed manifests and missing config dirs, plus adapter-level coverage that the call carries the resolved servers and no `t3-code`, and that it doesn't happen at all when nothing declares a server. The adapter test's fake `setMcpServers` is a prototype method reading `this`, mirroring the SDK, so passing an unbound reference fails the test the way it fails in production.

`vp test run` on the four touched suites: 90 passed. Targeted typecheck and lint clean.

Model: Claude Opus 5. Harness: T3 Code.

> [!NOTE]
> ### Load MCP servers declared by Claude plugins into the Claude adapter
> - Adds `discoverClaudePluginMcpServers` in [ClaudePlugins.ts](https://github.com/pingdotgg/t3code/pull/7542/files#diff-d2ff0667449a162bc8bd2e960e4f0f6cd7c05a28496d9e643b240b3eaf32fd10) to enumerate MCP servers from enabled Claude plugins by reading their `.mcp.json` manifests, expanding `${CLAUDE_PLUGIN_ROOT}`, and filtering disabled plugins and project-level overrides.
> - After creating a Claude query runtime, `ClaudeAdapterLive.makeClaudeAdapter` in [ClaudeAdapter.ts](https://github.com/pingdotgg/t3code/pull/7542/files#diff-e696bea66caf21d357bea5ea814bcbb18d1816288cc94247c855823f50fce6dc) asynchronously calls `queryRuntime.setMcpServers` to register discovered plugin servers, keyed as `plugin:<plugin>:<server>`.
> - Extracts `resolveClaudeConfigDirPath` from [ClaudeSkills.ts](https://github.com/pingdotgg/t3code/pull/7542/files#diff-df655ed95ed759ff78376524f2ceda1b436b639952ddab7e7e19f618a43e77c3) into [ClaudeHome.ts](https://github.com/pingdotgg/t3code/pull/7542/files#diff-73ba6e23eba4250edbecbe8239d07a8f0556ddef26400793eb1b7432199ea29d) so it can be shared across both drivers.
> - The `t3-code` server is excluded from the plugin registration call to avoid disrupting the existing connection.
> - Risk: Plugin server registration is detached and best-effort — failures log warnings but do not surface errors to the caller.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cca292e.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Drivers/ClaudeHome.ts
+++ b/apps/server/src/provider/Drivers/ClaudeHome.ts
@@ -34,6 +34,34 @@ export const makeClaudeEnvironment = Effect.fn("makeClaudeEnvironment")(function
   };
 });
 
+/**
+ * Resolve the Claude config directory the CLI would use, matching the
+ * precedence the spawned CLI sees: the instance's `homePath` (exported as
+ * `CLAUDE_CONFIG_DIR` by `makeClaudeEnvironment`), then a `CLAUDE_CONFIG_DIR`
+ * already present in the process environment, then `~/.claude`.
+ */
+export const resolveClaudeConfigDirPath = Effect.fn("resolveClaudeConfigDirPath")(function* (
+  config: Pick<ClaudeSettings, "homePath">,
+  environment: NodeJS.ProcessEnv,
+  cwd?: string,
+): Effect.fn.Return<string, never, Path.Path> {
+  const path = yield* Path.Path;
+  const homePath = config.homePath.trim();
+  if (homePath.length > 0) {
+    return path.resolve(expandHomePath(homePath));
+  }
+  // No tilde expansion here: the spawned CLI receives this env var verbatim
+  // (env vars are never shell-expanded), so a literal `~` must stay literal
+  // for discovery to scan the same directory the runtime would. A relative
+  // value is resolved against the workspace cwd — the subprocess's own cwd —
+  // for the same reason.
+  const environmentConfigDir = environment.CLAUDE_CONFIG_DIR?.trim() ?? "";
+  if (environmentConfigDir.length > 0) {
+    return cwd ? path.resolve(cwd, environmentConfigDir) : path.resolve(environmentConfigDir);
+  }
+  return path.join(NodeOS.homedir(), ".claude");
+});
+
 export const makeClaudeContinuationGroupKey = Effect.fn("makeClaudeContinuationGroupKey")(
   function* (config: Pick<ClaudeSettings, "homePath">): Effect.fn.Return<string, never, Path.Path> {
     const resolvedHomePath = yield* resolveClaudeHomePath(config);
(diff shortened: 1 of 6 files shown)


juliusmarminge:

**Reason:** Abandoned conflicted draft; Claude plugin MCP load never left draft.

PR #7545, opened 2026-08-19
Add `t3 env add`: pre-seed a legacy saved-environment catalog, headlessly

jvkassi:
## Why

`t3 project` exists because GUIs cannot add a project to a remote environment, so the CLI is the only way in. This fills a related gap: nothing headless can add an *environment* to a saved-environment catalog either.

## What this does (revised after review - see below)

- New `t3 env add` command, taking the same `--pairing-url` / `--host` + `--pairing-code` inputs as the existing pairing flow.
- Reuses `preparePairingRegistration` from `@t3tools/client-runtime/connection` for the probe + bearer bootstrap.
- **Narrower scope than originally written**: this only pre-seeds the legacy `saved-environments.json` format, and only makes sense *before* a desktop app has ever launched on a machine. It refuses (with an explanatory error) if a real `connection-catalog.json` already exists next to the target path, since that's the actual catalog a launched desktop app reads - a single blob encrypted whole via Electron's `safeStorage` (OS keychain-backed), which this command has no access to and cannot safely write into.
- Existing records in the file are preserved as opaque JSON, never decoded against a narrowing schema, so fields this command doesn't model (`encryptedBearerToken`, etc.) on *other* records survive untouched.
- No bearer token is persisted by default; `--print-token` prints it once for a caller to store wherever it already stores other secrets.

## Review history

The first version of this PR had two real, independently-flagged bugs (see PR comments): it targeted the wrong catalog file entirely (silent no-op on any already-migrated desktop install), and it silently dropped `encryptedBearerToken` on unrelated records via a lossy schema round-trip. Both are fixed as of the latest commit - see the comment thread for the specific fixes and the updated commit message for detail.

## What I could not verify

Still no local build in the environment I authored this in (`bun install` doesn't resolve the workspace `catalog:` deps here), so nothing has run against real `vp run typecheck`/`test` or CI. I cross-checked every API against live call sites elsewhere in the repo this round rather than assuming, but that's not a substitute for CI actually running - please treat it accordingly.

## Re: CONTRIBUTING.md

Still offering this small and easy to close - genuinely fine to redirect to Ideas discussions if that's a better fit than a PR for a not-currently-accepting-contributions project.

---

> [!NOTE]
> **Medium Risk**
> Writes local catalog files and runs pairing/bootstrap over the network; mitigations include refusing when the encrypted catalog exists, opaque merge to avoid data loss, and atomic writes.
> 
> **Overview**
> Adds **`t3 env add`** under the main CLI so provisioning can register a remote environment without opening the desktop pairing UI. It runs the same **`preparePairingRegistration`** path as clients (pairing URL or host + code), then upserts a **`PersistedSavedEnvironmentRecord`** into the legacy **`saved-environments.json`** catalog with atomic temp-file writes.
> 
> **Scope is intentionally narrow:** the command **refuses** if sibling **`connection-catalog.json`** already exists (post-migration desktop catalog), because it cannot write Electron **`safeStorage`**-encrypted data. Existing catalog rows are kept as **opaque JSON** so fields like **`encryptedBearerToken`** are not stripped when adding one environment; only the new row is schema-validated. Optional **`--print-token`** prints the bootstrapped bearer once without persisting it.
> 
> Wires **`envCommand`** into **`bin.ts`** and adds unit tests for upsert, catalog round-trip, and the encrypted-catalog guard.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b80570ad1973e3f98db450a477e10e0df580740. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add `t3 env add` CLI subcommand to pre-seed a legacy saved-environment catalog headlessly
> - Adds a new `env add` subcommand to the `t3` CLI ([env.ts](https://github.com/pingdotgg/t3code/pull/7545/files#diff-d73565d0f8292b769a1d9d258046782bb7da814bc070ee3ad62190dd6e3acf39)) that performs network pairing registration and writes the resulting record to the legacy saved-environments catalog without user interaction.
> - Catalog reads return an empty document when the file is absent; writes are atomic via a PID-suffixed temp file and rename, with directory auto-creation.
> - `upsertEnvironmentRecord` deduplicates by `environmentId`, preserving all other record fields.
> - A preflight check (`assertLegacyCatalogIsSafeToWrite`) blocks writes to the legacy catalog if an encrypted sibling (`connection-catalog.json`) already exists.
> - Risk: if an encrypted catalog exists at the derived sibling path, the command fails with a typed error rather than overwriting or merging.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6b80570.</sup>
> 
>

### DIFF
--- a/apps/server/src/bin.ts
+++ b/apps/server/src/bin.ts
@@ -9,6 +9,7 @@ import * as NetService from "@t3tools/shared/Net";
 import packageJson from "../package.json" with { type: "json" };
 import { authCommand } from "./cli/auth.ts";
 import { connectCommand } from "./cli/connect.ts";
+import { envCommand } from "./cli/env.ts";
 import { pairCommand } from "./cli/pair.ts";
 import { hasCloudPublicConfig } from "./cloud/publicConfig.ts";
 import { sharedServerCommandFlags } from "./cli/config.ts";
@@ -53,6 +54,7 @@ export const makeCli = ({ cloudEnabled = hasCloudPublicConfig } = {}) =>
       serveCommand,
       pairCommand,
       authCommand,
+      envCommand,
       projectCommand,
       serviceCommand,
       servicePreflightCommand,
--- a/apps/server/src/cli/env.test.ts
+++ b/apps/server/src/cli/env.test.ts
@@ -7,10 +7,16 @@ import * as NodeServices from "@effect/platform-node/NodeServices";
 import type { PersistedSavedEnvironmentRecord } from "@t3tools/contracts";
 import { assert, describe, expect, it } from "@effect/vitest";
 import * as Effect from "effect/Effect";
+import * as Exit from "effect/Exit";
 import * as FileSystem from "effect/FileSystem";
 import * as Path from "effect/Path";
 
-import { readCatalog, upsertEnvironmentRecord, writeCatalog } from "./env.ts";
+import {
+  assertLegacyCatalogIsSafeToWrite,
+  readCatalog,
+  upsertEnvironmentRecord,
+  writeCatalog,
+} from "./env.ts";
 
 const makeRecord = (
   overrides: Partial<PersistedSavedEnvironmentRecord> = {},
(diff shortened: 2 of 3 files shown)


jvkassi:
Pushed a fix (6b80570) for the real bugs flagged here - thank you both, Macroscope and Bugbot each independently caught the same root problem from different angles, which is a pretty strong signal it was real.

**Bugbot's finding was the more fundamental one**: I had targeted `saved-environments.json`, but that's a legacy format only read once, to migrate into the real catalog (`connection-catalog.json`, encrypted whole via Electron's `safeStorage`) - and only when that encrypted file doesn't exist yet. On any already-migrated desktop install, my writes were a silent no-op. Fixed by checking for a sibling `connection-catalog.json` first and refusing with an explanatory error if one exists, rather than pretending to work. This does make the command's honest scope narrower than the PR description originally claimed: it only helps pre-seed a catalog before a desktop app's first launch on a machine, since writing into the encrypted format from outside Electron isn't something this can safely do at all. Updated the PR description to reflect that.

**Macroscope's finding** (decoding every existing record through `PersistedSavedEnvironmentRecordSchema` drops `encryptedBearerToken` and anything else that schema doesn't model, for every OTHER record in the file, not just the new one) was real and independent of the above - fixed by keeping existing records as opaque JSON objects that never get decoded against a narrowing schema; only the new record being inserted is schema-validated. Added a test asserting an unrelated record's unknown field survives a merge untouched.

Also applied the error-modelling suggestion (operation discriminator on both read/write errors, matching `DesktopSavedEnvironmentsReadError`/`WriteError`/`DocumentDecodeError`).

Still unverified against a real `vp run typecheck`/`test` in the environment I authored this in (same caveat as the original PR) - I cross-checked every API I used against live call sites elsewhere in the repo this time (`fileSystem.exists`, `Schema.decodeUnknownEffect`, `Exit.isFailure`, etc.) rather than assuming, but that's not a substitute for CI actually running.

Albro3459:
I'm not saying this PR is pointless, but the original claim "GUIs cannot add a project to a remote environment" is no longer true. You can add a project to a remote environment through both the web and through the iOS app.

jvkassi:
I'll think it through more. 
Thanks for feedback

PR #7599, opened 2026-08-19
fix(desktop): bound backend shutdown wait during quit

### LINKED ISSUE #7598: [Bug]: Holding Cmd+Q can silently hang the app instead of quitting
### Before submitting

### Area

apps/desktop

### Steps to reproduce

1. Have an active agent or terminal session running in T3 Code desktop.
2. Get into a state where a backend or session doesn't exit cleanly when asked to stop (see Related issues below for how that happens).
3. Quit the app by holding Cmd+Q, as the app requires.
4. Release Cmd+Q after the hold completes.

### Expected behavior

The app quits.

### Actual behavior

The window disappears, but the process keeps running in the background. The app never fully quits. I have to force-quit it by hand.

### Impact

### Version or commit

0.0.34-nightly.20260819.1133

### Environment

macOS 25.6.1 (darwin arm64), Node v24.18.0, T3 Code (Nightly) desktop app

### Logs or stack traces

None. This is a silent hang; nothing errors or logs.

### Workaround

Force-quit the app.

### Diagnosis

The quit path (`apps/desktop/src/app/DesktopApp.ts`, the `scopedProgram` finalizer) stops every backend instance in the pool before calling `app.quit()`. That stop call had no timeout. If any one backend didn't exit cleanly, the wait never ended, and `app.quit()` never ran.

Before PR #6562 ("close the window before quit cleanup"), the window stayed open during this wait. You could at least see something was stuck. After #6562, the window closes first, so the same hang now looks like nothing happened at all. #6562 didn't cause this on purpose. It just removed the one visible sign of it.

### Related issues

#7369 (closed, fixed by #7397, different bug: hold-detection itself misfired, not a duplicate)

Fix: https://github.com/pingdotgg/t3code/pull/7599 — bounds the backend-stop wait to 5 seconds, matching the same pattern already used in `apps/desktop/src/updates/DesktopUpdates.ts`.

### Filed by

Claude (Sonnet 5) via Claude Code, at the reporter's request.

ishaanko:
## Problem

The quit-time finalizer stopped every backend instance in the pool with no timeout. If any one backend didn't exit cleanly, the wait never ended, and the app never called `app.quit()`. Since PR #6562 now destroys the window before this wait, the hang became invisible: holding Cmd+Q looked like it did nothing, and the only way out was a force quit.

Fixes #7598

## Fix

Bound the wait to 5 seconds, matching the identical pattern already used for this same operation in `apps/desktop/src/updates/DesktopUpdates.ts`. On timeout, the close keeps running in the background instead of blocking quit, so a slow-but-healthy shutdown still finishes cleanly, and a genuinely hung one no longer blocks the app from exiting.

Extracted the finalizer's logic into an exported `stopAllPoolInstances` function so it has a real seam to test against.

## Testing

- New test in `DesktopBackendManager.test.ts` calls the real `stopAllPoolInstances` export with a mocked pool whose instances hang past the timeout, and asserts the call still completes on time. Verified by hand that removing the timeout argument makes this test hang, confirming it actually catches a regression.
- `vp test run apps/desktop/src/backend/DesktopBackendManager.test.ts`: 26 passed
- `vp run --filter @t3tools/desktop typecheck`: clean

---
Model: Claude Sonnet 5. Harness: Claude Code.

---

> [!NOTE]
> **Medium Risk**
> Changes desktop quit/shutdown behavior for all pooled backends; bounded wait reduces hang risk but timed-out teardown may still run in the background.
> 
> **Overview**
> Fixes quit hangs when a backend pool instance never finishes teardown (e.g. after the window is destroyed first), so **Cmd+Q could appear to do nothing** until force quit.
> 
> **Quit finalizer** now goes through exported **`stopAllPoolInstances`**, which stops every pooled backend **concurrently** with a **5s timeout per `instance.stop`**—same pattern as `DesktopUpdates.ts`. After the budget, quit keeps moving instead of waiting forever; slow but healthy shutdown can still finish in the background.
> 
> Adds a **regression test** that calls the real `stopAllPoolInstances` with a mocked pool whose instances hang past the timeout and asserts the call completes on time.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6187110d9c21db292f4bae4db70d9f372b2f9f48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Bound backend shutdown wait to 5 seconds per instance during desktop app quit
> - Extracts a new `stopAllPoolInstances` function in [DesktopApp.ts](https://github.com/pingdotgg/t3code/pull/7599/files#diff-267cf20af5c499f356b88a4bc3d3dec61aa6c7a3398f93f8836975b3bcf764e9) that stops each instance from `DesktopBackendPool` with a 5-second timeout, running stops concurrently.
> - Replaces the inline teardown loop in `scopedProgram`'s finalizer with a call to `stopAllPoolInstances`, preventing the app from hanging indefinitely on quit if a backend fails to stop.
> - Adds a test in [DesktopBackendManager.test.ts](https://github.com/pingdotgg/t3code/pull/7599/files#diff-bce4b7f6b285d268f04bc702d29a17a6a4cf8c42adc9454dff19950eb60e7661) that simulates hanging teardown and verifies the effect completes after the timeout.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6187110.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved application shutdown reliability when backend services fail to stop promptly.
  - The app now proceeds with shutdown after a five-second limit instead of waiting indefinitely.
  - Backend cleanup continues in the background and completes when services become available again.

### DIFF
--- a/apps/desktop/src/app/DesktopApp.ts
+++ b/apps/desktop/src/app/DesktopApp.ts
@@ -1,4 +1,5 @@
 import * as Cause from "effect/Cause";
+import * as Duration from "effect/Duration";
 import * as Effect from "effect/Effect";
 import * as Option from "effect/Option";
 import * as Ref from "effect/Ref";
@@ -141,6 +142,20 @@ const handleFatalStartupError = Effect.fn("desktop.startup.handleFatalStartupErr
 
 const fatalStartupCause = <E>(stage: string, cause: Cause.Cause<E>) =>
   handleFatalStartupError(stage, Cause.pretty(cause)).pipe(Effect.andThen(Effect.failCause(cause)));
+
+export const stopAllPoolInstances = Effect.fn("desktop.app.stopAllPoolInstances")(
+  function* (): Effect.fn.Return<void, never, DesktopBackendPool.DesktopBackendPool> {
+    // Stop every backend in the pool with a timeout to guarantee the quit
+    // path makes progress even if a backend hangs during teardown.
+    const pool = yield* DesktopBackendPool.DesktopBackendPool;
+    const instances = yield* pool.list;
+    yield* Effect.forEach(
+      instances,
+      (instance) => instance.stop({ timeout: Duration.seconds(5) }),
+      { concurrency: "unbounded" },
+    );
+  },
+);
 
 const bootstrap = Effect.gen(function* () {
   const pool = yield* DesktopBackendPool.DesktopBackendPool;
@@ -313,18 +328,12 @@ const scopedProgram = Effect.scoped(
     const shutdown = yield* DesktopShutdown.DesktopShutdown;
 
     yield* Effect.addFinalizer(() =>
-      Effect.gen(function* () {
-        const pool = yield* DesktopBackendPool.DesktopBackendPool;
-        // Stop every backend in the pool, not just the primary. The
-        // electronApp.quit() path can race ahead of the layer-scope
-        // cascade, so leaving the WSL instance for its parent scope
-        // finalizer means it gets hard-killed by the OS instead of
-        // receiving SIGTERM + grace. Stops run concurrently.
-        const instances = yield* pool.list;
-        yield* Effect.forEach(instances, (instance) => instance.stop(), {
-          concurrency: "unbounded",
-        });
-      }).pipe(Effect.ensuring(shutdown.markComplete)),
+      // Stop every backend in the pool, not just the primary. The
+      // electronApp.quit() path can race ahead of the layer-scope
+      // cascade, so leaving the WSL instance for its parent scope
+      // finalizer means it gets hard-killed by the OS instead of
+      // receiving SIGTERM + grace.
+      stopAllPoolInstances().pipe(Effect.ensuring(shutdown.markComplete)),
     );
 
     yield* startup;
--- a/apps/desktop/src/backend/DesktopBackendManager.test.ts
+++ b/apps/desktop/src/backend/DesktopBackendManager.test.ts
@@ -4,6 +4,7 @@ import {
   DesktopTelemetryControlMessage,
 } from "@t3tools/contracts";
 import { assert, describe, it } from "@effect/vitest";
+import * as Cause from "effect/Cause";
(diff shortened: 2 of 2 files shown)


ishaanko:
[claude-fable-5] RESPONDING ON BEHALF OF ISHAAN

Approved a few days back and still applies cleanly on latest main. Glad to rebase or adjust if anything is needed before merge.

shivamhwp:

Please make the shutdown regression wait for both backend finalizers to start, then wait for both to finish after releasing `allowTeardown`. The shared `teardownStarted` Deferred only confirms one backend reached teardown, and advancing TestClock by 100 ms does not establish that either finalizer finished. Use per-instance Deferreds or queue notifications for those milestones.

Please also invoke `stopAllPoolInstances` from an `Effect.addFinalizer` scope in the regression. That matches the quit path's finalizer context instead of exercising only an ordinary interruptible call.

ishaanko:
Thanks, both points are addressed in 44ad0b15e.

**Per-instance milestones.** Each backend's process-scope finalizer now offers its instance name to a `teardownStarted` queue when it begins and to a `teardownFinished` queue when it ends. The test takes two entries from `teardownStarted` before it advances the clock, so both backends are proven to have reached teardown. After the quit finalizer returns, it asserts `teardownFinished` is still empty, releases `allowTeardown`, and then takes two entries from `teardownFinished`. No clock advance stands in for completion any more.

**Finalizer context.** The test now registers `stopAllPoolInstances` with `Effect.addFinalizer` inside an `Effect.scoped` block and lets the scope close run it, which matches how `DesktopApp` invokes it on quit.

I also rebased onto current `main`, since the old branch no longer shared history with it. Verified by hand that removing the timeout argument makes the test hang, so it still catches the regression.

Model: Claude Fable 5.1. Harness: Claude Code.

PR #7610, opened 2026-08-19
fix(server): unpushed branch no longer breaks thread creation

DraftProducts:
## What Changed

`dispatchBootstrapTurnStart` now treats the remote-tracking lookup as optional.
When `resolveRemoteTrackingCommit` fails, the worktree is cut from the local base
branch instead of taking down the whole bootstrap. Repos with no `origin` remote
already behave this way; this extends the same fallback to a base branch that has
no remote-tracking ref.

Tests: new bootstrap case in `server.test.ts` for a base branch with no
remote-tracking ref, asserting the worktree is cut from the local branch and that
no `thread.delete` is dispatched — the rollback is what burns the thread id. Full
`apps/server` suite (2611 passing) plus every other package, typecheck, lint, and
format pass. The two failures in the repo (`ProviderRegistry` codex re-probe and
`Net.findAvailablePort`) reproduce on `main` and are untouched by this change.

## Why

Creating a thread on a local-only branch with "start from origin" enabled fails,
and every retry then fails differently:

    Orchestration command invariant failed (thread.create):
    Thread '<id>' already exists and cannot be created twice.

That invariant error is a red herring. Bootstrap runs `git fetch origin` and then
resolves `refs/remotes/origin/<baseBranch>`. A branch that was never pushed has no
remote-tracking ref, so `git rev-parse --verify` exits non-zero and the whole
bootstrap fails. The rollback dispatches `thread.delete`, which soft-deletes the
thread but leaves its event stream in place, so `requireThreadAbsent` rejects the
retry. Each retry carries a fresh commandId, so receipt idempotency never dedupes
it either. The thread id is burned for good, and the git failure that started it
never reaches the user.

Falling back is the right call rather than surfacing the git error: "start from
origin" is a stored default rather than something chosen per thread, so honoring
it only when a remote base exists matches what the no-remote path already does.

Not addressed: any other bootstrap failure still rolls back into the same
unusable state. Making a soft-deleted thread id reusable is a decision about the
event-sourcing invariant, not this fix.

## Checklist

- [ ] I included before/after screenshots for any UI changes — N/A, server-only
- [ ] I included a video for animation/interaction changes — N/A, server-only

Changes by Claude Opus 5 running in Claude Code.

---

> [!NOTE]
> **Medium Risk**
> Changes bootstrap git/worktree resolution for thread creation; wrong fallback could point worktrees at unexpected refs, though scope is limited to the optional remote-tracking path.
> 
> **Overview**
> Fixes thread bootstrap failing (and burning the thread id on retry) when **start from origin** is enabled but the chosen base branch has no `origin` tracking ref.
> 
> In `dispatchBootstrapTurnStart`, after `fetchRemote`, **`resolveRemoteTrackingCommit`** is now wrapped in **`Effect.option`**. On success, the worktree still bases off the remote commit SHA; on failure, **`worktreeBaseRef`** stays on the local **`baseBranch`** (same behavior as repos without `origin`).
> 
> Adds a **`server.test`** case that stubs a failed remote-tracking resolve and asserts **`createWorktree`** is called with the local branch as **`refName`** while bootstrap commands complete without rollback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 490657a0b7a929ddf61a06a84cb1c3908e8afeb0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix thread creation when base branch has no remote-tracking ref
> When `startFromOrigin=true` and the base branch has no remote-tracking ref, `resolveRemoteTrackingCommit` now uses `Effect.option` so a failure returns `None` instead of aborting. If no remote-tracking commit is found, `worktreeBaseRef` is left unchanged and `createWorktree` falls back to the local base branch.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 490657a.</sup>
> 
>

### DIFF
--- a/apps/server/src/server.test.ts
+++ b/apps/server/src/server.test.ts
@@ -7598,6 +7598,120 @@ it.layer(NodeServices.layer)("server router seam", (it) => {
       }).pipe(Effect.provide(NodeHttpServer.layerTest)),
   );
 
+  it.effect(
+    "falls back to the local base branch when the base branch has no remote-tracking branch",
+    () =>
+      Effect.gen(function* () {
+        const dispatchedCommands: Array<OrchestrationCommand> = [];
+        const remoteExists = vi.fn(
+          (_: Parameters<GitVcsDriver.GitVcsDriver["Service"]["remoteExists"]>[0]) =>
+            Effect.succeed(true),
+        );
+        const fetchRemote = vi.fn(
+          (_: Parameters<GitVcsDriver.GitVcsDriver["Service"]["fetchRemote"]>[0]) => Effect.void,
+        );
+        const resolveRemoteTrackingCommit = vi.fn(
+          (
+            input: Parameters<
+              GitVcsDriver.GitVcsDriver["Service"]["resolveRemoteTrackingCommit"]
+            >[0],
+          ) =>
+            Effect.fail(
+              new GitCommandError({
+                operation: "GitVcsDriver.resolveRemoteTrackingCommit",
+                command: `git rev-parse --verify refs/remotes/origin/${input.refName}^{commit}`,
+                cwd: input.cwd,
+                detail: "fatal: Needed a single revision",
+              }),
+            ),
+        );
+        const createWorktree = vi.fn(
+          (_: Parameters<GitVcsDriver.GitVcsDriver["Service"]["createWorktree"]>[0]) =>
+            Effect.succeed({
+              worktree: {
+                refName: "t3code/bootstrap-refName",
+                path: "/tmp/bootstrap-worktree",
+              },
+            }),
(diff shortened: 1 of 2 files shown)


t3dotgg:
> [!NOTE]

Closing this catch-all fallback in favor of [#8241](https://github.com/pingdotgg/t3code/pull/8241). That open candidate checks whether the remote branch exists, while this version suppresses every remote-tracking resolution failure. The local-only branch case should stay covered without hiding unrelated git failures. The replacement is not merged yet.

PR #7632, opened 2026-08-20
feat(server): render model reasoning as timeline activities

jbbottoms:
Models that stream reasoning (Claude thinking blocks, Codex `item/reasoning/*`, OpenCode reasoning parts, local reasoning models behind OpenCode) currently show nothing in any T3 client — the thinking never renders on web, desktop, or mobile.

### Why

Adapters already emit reasoning. Ingestion drops it in two places:

- `content.delta` with `streamKind: "reasoning_text"` / `"reasoning_summary_text"` has no consumer — the assistant-delta path in `ProviderRuntimeIngestion` matches `assistant_text` exactly.
- `item.started/updated/completed` with `itemType: "reasoning"` fail the `isToolLifecycleItemType` gate in all three activity projections, so they project to `[]`.

(Separately, ClaudeAdapter emits thinking deltas without an `itemId` because `content_block_start` creates no block state for `thinking` blocks — this change works with or without that, by falling back to a per-turn buffer key.)

### What

Server-only, one file plus tests. In `ProviderRuntimeIngestion`:

- Buffer reasoning deltas per `(thread, itemId ?? turn)`.
- While streaming, emit a throttled `tool.updated` activity (summary/title "Thinking", `status: "inProgress"`, detail = tail of the accumulated text). The first delta emits immediately so the row appears; afterwards re-emit only every 600 buffered chars, mirroring the O(N²) persistence guard the tool `item.updated` path documents.
- Flush a `tool.completed` activity (summary/title "Thought", full text capped at 6000 chars) on `item.completed` for a reasoning item (richer of streamed text vs item detail wins, and the buffer is consumed so turn boundaries cannot double-flush), on `turn.completed` / `turn.aborted` / `request.opened` / `user-input.requested`, and on `session.exited` (which also drops the thread's buffers).

Reusing the existing tool activity shape (`tone: "tool"`, `toolCallId` for collapse) means web, desktop, and mobile all render, collapse, and expand the rows with zero client changes — streaming updates collapse into one row via `deriveToolLifecycleCollapseKey`, exactly like streaming tool output.

### Proof

- 3 new tests in `ProviderRuntimeIngestion.test.ts`: streaming + turn-boundary flush; reasoning `item.completed` flush picking the richer text and not double-flushing at the turn boundary; no reasoning activities for plain assistant text. Full file: 52/52 pass. `tsgo --noEmit`: clean.
- Live: OpenCode lane against a local Qwen3.8-27B — "Thinking" streams during the turn and a "Thought" row with the full reasoning body renders in the web timeline (screenshot in a comment below).

### Notes for review

- Reasoning summaries (`reasoning_summary_text`) share the item's buffer with full reasoning text; adapters that emit both for one item interleave. Codex GPT models emit only summaries, so in practice each item carries one kind.
- Two provider-side limits observed while testing, out of scope here: Claude Code CLI redacts thinking text in its stream-json output (`"thinking": ""` with token estimates only), and Codex reasoning items can arrive with empty `content`/`summary` when summaries are not enabled. In both cases this change correctly renders nothing.
- The 6000-char completed-detail cap and the "[reasoning truncated]" marker are deliberate: `payload.detail` persists to the projection table and the event store.

---

> [!NOTE]
> **Medium Risk**
> Adds in-memory buffering and new persisted activity payloads on a hot ingestion path; turn-scoping and caps limit wrong-flush and storage blow-up, but timeline volume and ordering still depend on provider event timing.
> 
> **Overview**
> **Provider runtime ingestion** now turns streamed model reasoning into thread timeline activities so web, desktop, and mobile can show “Thinking” / “Thought” without client changes.
> 
> Reasoning from `content.delta` (`reasoning_text` / `reasoning_summary_text`) and `item.completed` with `itemType: "reasoning"` is buffered per thread and item (or per turn when `itemId` is missing). While streaming, ingestion emits throttled **`tool.updated`** rows (summary **Thinking**, `inProgress`, tail-truncated detail); on flush it emits **`tool.completed`** (**Thought**, full text capped at 6k with an optional truncation marker). Flushes happen on reasoning item completion (preferring richer item detail over streamed text), on turn boundaries (`turn.completed` / `turn.aborted` and approval/user-input pauses) scoped to that turn so stale completions cannot wipe active reasoning, and on **`session.exited`** (buffers cleared).
> 
> Reusing the existing tool activity shape and `toolCallId` keeps collapse behavior aligned with streaming tools. **`ProviderRuntimeIngestion.test.ts`** adds coverage for streaming + turn flush, item.completed richness and no double-flush, superseded-turn boundaries, and no noise from plain assistant text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26afe446dbf53161863947dd0b6f9e06c0f68e65. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Render model reasoning as 'Thinking' and 'Thought' timeline activities
> - Adds `handleReasoningEvent` to [ProviderRuntimeIngestion.ts](https://github.com/pingdotgg/t3code/pull/7632/files#diff-22a6c9818b956463b848a73a5b3e787b44a144bd003b1cc25fa2b1b3110cdea7) to buffer `reasoning_text` deltas and emit periodic `tool.updated` ('Thinking') activities.
> - Emits a final `tool.completed` ('Thought') activity on `item.completed` or turn/session boundaries, using the richer of buffered or item text.
> - Enforces size limits via `REASONING_BUFFER_TEXT_CAP` (60000) and `REASONING_FINAL_DETAIL_LIMIT` (6000), adding truncation markers when text is capped.
> - Risk: Introduces `reasoningBuffersByThread` state that retains up to 60000 characters per active reasoning stream until flushed.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 26afe44.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
@@ -3623,4 +3623,190 @@ describe("ProviderRuntimeIngestion", () => {
     expect(thread.session?.status).toBe("error");
     expect(thread.session?.lastError).toBe("runtime still processed");
   });
+
+  it("projects reasoning deltas into a streaming thinking activity and flushes a thought on turn completion", async () => {
+    const harness = await createHarness();
+    const now = "2026-01-01T00:00:00.000Z";
+
+    harness.emit({
+      type: "content.delta",
+      eventId: asEventId("evt-reasoning-delta-1"),
+      provider: ProviderDriverKind.make("opencode"),
+      createdAt: now,
+      threadId: asThreadId("thread-1"),
+      turnId: asTurnId("turn-reasoning"),
+      itemId: asItemId("reasoning-part-1"),
+      payload: {
+        streamKind: "reasoning_text",
+        delta: "The user wants the answer in metric units.",
+      },
+    });
+    harness.emit({
+      type: "content.delta",
+      eventId: asEventId("evt-reasoning-delta-2"),
+      provider: ProviderDriverKind.make("opencode"),
+      createdAt: now,
+      threadId: asThreadId("thread-1"),
+      turnId: asTurnId("turn-reasoning"),
+      itemId: asItemId("reasoning-part-1"),
+      payload: {
+        streamKind: "reasoning_text",
+        delta: " Convert before replying.",
+      },
+    });
+    harness.emit({
+      type: "turn.completed",
+      eventId: asEventId("evt-reasoning-turn-completed"),
+      provider: ProviderDriverKind.make("opencode"),
+      createdAt: now,
+      threadId: asThreadId("thread-1"),
+      turnId: asTurnId("turn-reasoning"),
+      status: "completed",
+    });
+
+    const thread = await waitForThread(harness.readModel, (entry) =>
+      entry.activities.some(
+        (activity: ProviderRuntimeTestActivity) => activity.kind === "tool.completed",
+      ),
+    );
+
+    const streaming = thread.activities.find(
+      (activity: ProviderRuntimeTestActivity) => activity.id === "evt-reasoning-delta-1:reasoning",
+    );
+    const streamingPayload =
+      streaming?.payload && typeof streaming.payload === "object"
+        ? (streaming.payload as Record<string, unknown>)
+        : undefined;
+    expect(streaming?.kind).toBe("tool.updated");
+    expect(streaming?.summary).toBe("Thinking");
+    expect(streamingPayload?.toolCallId).toBe("reasoning-part-1");
+    expect(streamingPayload?.status).toBe("inProgress");
+    expect(streamingPayload?.detail).toBe("The user wants the answer in metric units.");
+
+    const completed = thread.activities.find(
+      (activity: ProviderRuntimeTestActivity) =>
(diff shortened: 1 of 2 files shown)


jbbottoms:
Live proof — patched server, OpenCode lane against a local Qwen3.8-27B. "Thinking" streamed during the turn; the completed "Thought" row below expands to the full reasoning body:

reasoning rendered in the web timeline

jbbottoms:
Both findings were real — fixed in 26afe446:

- Turn boundaries now flush only buffers attributed to the boundary's turn, so a late `turn.completed` from a superseded turn can no longer flush or wipe the active turn's reasoning (unattributed buffers still flush on any boundary; `session.exited` still flushes and drops everything).
- Buffer state now mutates only after the activity dispatch succeeds: `emittedLength` advances after the streaming append, and the buffer is deleted after the completion flush, so the logged-and-skipped failure path gets a natural retry at the next delta or boundary instead of silently losing the thought.

Added a regression test for the superseded-turn case (stale boundary leaves the active buffer intact, active boundary flushes it). 53/53 pass, typecheck clean.

juliusmarminge:
Superseded by #11784 (feat(chat): show provider thinking traces), which landed on main.

PR #7656, opened 2026-08-20
feat(copilot): add GitHub Copilot provider (via @github/copilot-sdk)

NSExceptional:
> **Draft / RFC.** Opening this early for maintainer feedback on integration and conventions. It builds, typechecks, and passes its unit tests, and I've smoke-tested real turns against the live `copilot` CLI — but I'm not confident it hits every T3 convention (multi-surface, receipts, provider parity), so I'd love guidance on what to tighten before it's merge-ready.

## What

Adds **GitHub Copilot** as a first-class provider, alongside Codex / Claude / Cursor / Grok / OpenCode. It's driven by GitHub's first-party [`@github/copilot-sdk`](https://www.npmjs.com/package/@github/copilot-sdk), which spawns and drives the installed `copilot` runtime binary over its typed JSON-RPC protocol (`RuntimeConnection.forStdio`) — the same engine the Copilot CLI/IDE use. No extra runtime download; it reuses whatever `copilot` the user already has (e.g. Homebrew).

## How

- **Provider / driver / adapter** under `apps/server/src/provider/`, plus a small `provider/sdk/` layer:
  - `CopilotSdkClient` — scoped Effect wrapper around `CopilotClient` (start on acquire, stop on release; one shared client per provider instance).
  - `CopilotSdkModels` — maps `client.listModels()` into per-model capabilities: **reasoning effort** from each model's `supportedReasoningEfforts`, and a **context-window tier** gated on the model's `longContext` billing block. Applied via `SessionConfig` / `session.setModel`.
  - `CopilotSdkRuntimeEvents` — translates SDK `SessionEvent`s into the canonical `ProviderRuntimeEvent` stream.
- **Session lifecycle:** per-thread `CopilotSession`, a callback→Effect event bridge (SDK is callback-based, unlike the async-iterable providers), permission requests wired into the existing approval flow, and `send` + `session.idle` turn handling.
- **Model discovery** via `client.listModels()`.
- **Git text generation** (commit messages, PR content, branch names, thread titles) uses the SDK's one-shot `sendAndWait`.
- **Contracts:** adds `copilot.sdk.event` / `copilot.sdk.permission` runtime sources and Copilot settings/model schemas; web wiring for the provider settings + model picker (reuses the generic option-descriptor UI).
- Resolves the `copilot` binary to an absolute path before spawning (a GUI-launched app inherits a minimal PATH), and passes env only on the stdio connection (the SDK rejects env in both places).

## Testing

- `tsgo` typecheck clean (contracts / server / web); targeted lint clean.
- Unit tests for the model→capability mapping, tunable resolution, and provider status parsing (74 provider + text-gen tests green).
- End-to-end smoke test against `copilot` 1.0.80: `createSession` → streaming `send` → `session.idle`, `setModel` with `long_context`, and the permission callback. `listModels()` returns per-model reasoning efforts / context tiers as expected.

## Notes / open questions

- Builds on earlier Copilot-provider groundwork by @its-hmny (credited via `Co-authored-by`); I ported it to the SDK and reworked model discovery + tunables.
- The SDK is callback-based; I bridge its events into an Effect queue. Happy to align that with how you'd prefer provider adapters to consume SDK streams.
- Mobile surface and docs aren't touched yet — guidance welcome on what's expected for a new provider.

_Authored with Claude Opus 4.8 via Claude Code._

> [!NOTE]
> ### Add GitHub Copilot provider via `@github/copilot-sdk`
> - Introduces a full `copilot` provider driver registered in [builtInDrivers.ts](https://github.com/pingdotgg/t3code/pull/7656/files#diff-7bd5dca89e0c2be78fa6adf17c285d16f158d8c8b671b62245d46b910a665c39), wiring together a SDK client, adapter, text generation, and provider snapshot pipeline.
> - [CopilotSdkClient.ts](https://github.com/pingdotgg/t3code/pull/7656/files#diff-ea0a23669eaaa0f0c665b504f46a05d620aad4637bf8e29c5a28021b424ff95f) wraps the SDK with Effect-based methods, binary path resolution, and scoped lifecycle management.
> - [CopilotAdapter.ts](https://github.com/pingdotgg/t3code/pull/7656/files#diff-d1fdd1ff48ff201b33512aada0688992230a399ec0b6482437aabdd0768bc6d9) manages per-thread sessions, streaming runtime events, permission requests, resume cursors, and in-session model switching.
> - [CopilotTextGeneration.ts](https://github.com/pingdotgg/t3code/pull/7656/files#diff-3f55717b198ae96e8c1c6f1acc682bb6da327ec1da696ff426e3b8edc4b53759) implements commit message, PR content, branch name, and thread title generation with schema-validated JSON and a 180s timeout.
> - [CopilotProvider.ts](https://github.com/pingdotgg/t3code/pull/7656/files#diff-a7ca86ea5e4e0462fc38932a09bfc0a947a45493deac41356edd4f39f7e9211f) probes the Copilot CLI for version/auth status and discovers models via the SDK, enriching snapshots asynchronously.
> - Contracts and UI are updated to recognize the `copilot` kind, default to `gpt-4.1` / `gpt-4.1-mini`, and surface the provider in pickers and settings dialogs.
> - Risk: adds `@github/copilot-sdk` dependency and disallows `koffi` builds in [pnpm-workspace.yaml](https://github.com/pingdotgg/t3code/pull/7656/files#diff-18ae0a0fab29a7db7aded913fd05f30a2c8f6c104fadae86c9d217091709794c); `COPILOT_DRIVER_KIND` defaults in [model.ts](https://github.com/pingdotgg/t3code/pull/7656/files#diff-cff20c6bccac2635a0e8b89343cf21815f9fc219bd536a90f66537a56eac7959) may affect model selection if overrides are absent.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 56d5f9a. 17 files reviewed, 16 issues evaluated, 2 issues filtered, 10 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> <details>
> <summary>apps/server/src/provider/Layers/CopilotAdapter.ts — 3 comments posted, 5 evaluated, 1 filtered</summary>
>
> - [line 447](https://github.com/pingdotgg/t3code/blob/56d5f9acdfe25f231f6235d1a979adabc1f59ed0/apps/server/src/provider/Layers/CopilotAdapter.ts#L447): The adapter constructs or accepts `nativeEventLogger`, but never invokes its `write` method. Therefore configuring `nativeEventLogPath` or injecting a native logger creates/manages the logger while every Copilot SDK event is silently omitted from the native event log. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/provider/Layers/CopilotProvider.ts — 2 comments posted, 3 evaluated, 1 filtered</summary>
>
> - [line 115](https://github.com/pingdotgg/t3code/blob/56d5f9acdfe25f231f6235d1a979adabc1f59ed0/apps/server/src/provider/Layers/CopilotProvider.ts#L115): `detectCopilotAuthFromEnvironment` uses nullish coalescing before checking whether a token is nonempty. If `COPILOT_GITHUB_TOKEN` is defined as `""` or whitespace while `GH_TOKEN` or `GITHUB_TOKEN` contains a valid token, the empty first value wins and the function returns `unknown` instead of `authenticated`. Select the first nonblank value rather than the first non-nullish one. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
>
> </details>
>

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -28,6 +28,7 @@
     "@effect/platform-node-shared": "catalog:",
     "@effect/sql-sqlite-bun": "catalog:",
     "@ff-labs/fff-node": "0.9.4",
+    "@github/copilot-sdk": "^1.0.11",
     "@opencode-ai/sdk": "^1.3.15",
     "@pierre/diffs": "catalog:",
     "effect": "catalog:",
--- a/apps/server/src/provider/Drivers/CopilotDriver.ts
+++ b/apps/server/src/provider/Drivers/CopilotDriver.ts
@@ -1,0 +1,191 @@
+/**
+ * CopilotDriver — `ProviderDriver` for GitHub Copilot, backed by the first-party
+ * `@github/copilot-sdk` (see CopilotAdapter / CopilotProvider). The SDK spawns
+ * and drives the installed `copilot` runtime binary over its typed JSON-RPC
+ * protocol; only the git text-generation path still uses the ACP transport.
+ *
+ * @module provider/Drivers/CopilotDriver
+ */
+import {
+  CopilotSettings,
+  ProviderDriverKind,
+  type ServerProvider,
+  type ServerProviderModel,
+} from "@t3tools/contracts";
+import * as Crypto from "effect/Crypto";
+import * as Effect from "effect/Effect";
+import * as FileSystem from "effect/FileSystem";
+import * as Path from "effect/Path";
+import * as Ref from "effect/Ref";
+import * as Schema from "effect/Schema";
+import { HttpClient } from "effect/unstable/http";
+import { ChildProcessSpawner } from "effect/unstable/process";
+
+import * as BackgroundPolicy from "../../background/BackgroundPolicy.ts";
+import { ServerConfig } from "../../config.ts";
+import { ServerSettingsService } from "../../serverSettings.ts";
+import { makeCopilotTextGeneration } from "../../textGeneration/CopilotTextGeneration.ts";
+import { ProviderDriverError } from "../Errors.ts";
+import { makeCopilotAdapter } from "../Layers/CopilotAdapter.ts";
+import {
+  buildInitialCopilotProviderSnapshot,
+  checkCopilotProviderStatus,
+  enrichCopilotSnapshot,
+} from "../Layers/CopilotProvider.ts";
+import { ProviderEventLoggers } from "../Layers/ProviderEventLoggers.ts";
+import { makeManagedServerProvider } from "../makeManagedServerProvider.ts";
+import {
+  defaultProviderContinuationIdentity,
+  type ProviderDriver,
+  type ProviderInstance,
+} from "../ProviderDriver.ts";
+import type { ServerProviderDraft } from "../providerSnapshot.ts";
+import { mergeProviderInstanceEnvironment } from "../ProviderInstanceEnvironment.ts";
+import {
+  makeProviderMaintenanceCapabilities,
+  makeStaticProviderMaintenanceResolver,
+  resolveProviderMaintenanceCapabilitiesEffect,
+} from "../providerMaintenance.ts";
+import {
+  haveProviderSnapshotSettingsChanged,
+  makeProviderSnapshotSettingsSource,
+  type ProviderSnapshotSettings,
+} from "../providerUpdateSettings.ts";
+
+const decodeCopilotSettings = Schema.decodeSync(CopilotSettings);
+
+const DRIVER_KIND = ProviderDriverKind.make("copilot");
+
+const UPDATE = makeStaticProviderMaintenanceResolver(
+  makeProviderMaintenanceCapabilities({
+    provider: DRIVER_KIND,
+    packageName: null,
+    updateExecutable: "copilot",
+    updateArgs: ["update"],
+    updateLockKey: "copilot-update",
+  }),
+);
+
+export type CopilotDriverEnv =
+  | BackgroundPolicy.BackgroundPolicy
+  | ChildProcessSpawner.ChildProcessSpawner
+  | Crypto.Crypto
+  | FileSystem.FileSystem
+  | HttpClient.HttpClient
+  | Path.Path
+  | ProviderEventLoggers
+  | ServerConfig
+  | ServerSettingsService;
+
+const withInstanceIdentity =
+  (input: {
+    readonly instanceId: ProviderInstance["instanceId"];
+    readonly displayName: string | undefined;
+    readonly accentColor: string | undefined;
+    readonly continuationGroupKey: string;
+  }) =>
+  (snapshot: ServerProviderDraft): ServerProvider => ({
+    ...snapshot,
+    instanceId: input.instanceId,
+    driver: DRIVER_KIND,
+    ...(input.displayName ? { displayName: input.displayName } : {}),
+    ...(input.accentColor ? { accentColor: input.accentColor } : {}),
+    continuation: { groupKey: input.continuationGroupKey },
+  });
+
+export const CopilotDriver: ProviderDriver<CopilotSettings, CopilotDriverEnv> = {
+  driverKind: DRIVER_KIND,
+  metadata: {
+    displayName: "GitHub Copilot",
+    supportsMultipleInstances: true,
+  },
+  configSchema: CopilotSettings,
+  defaultConfig: (): CopilotSettings => decodeCopilotSettings({}),
+  create: ({ instanceId, displayName, accentColor, environment, enabled, config }) =>
+    Effect.gen(function* () {
+      const crypto = yield* Crypto.Crypto;
+      const spawner = yield* ChildProcessSpawner.ChildProcessSpawner;
+      const httpClient = yield* HttpClient.HttpClient;
+      const eventLoggers = yield* ProviderEventLoggers;
+      const serverSettings = yield* ServerSettingsService;
+      const processEnv = mergeProviderInstanceEnvironment(environment);
+      const continuationIdentity = defaultProviderContinuationIdentity({
+        driverKind: DRIVER_KIND,
+        instanceId,
+      });
+      const stampIdentity = withInstanceIdentity({
+        instanceId,
+        displayName,
+        accentColor,
+        continuationGroupKey: continuationIdentity.continuationKey,
+      });
+      const effectiveConfig = {
+        ...config,
+        enabled,
+      } satisfies CopilotSettings;
+      const maintenanceCapabilities = yield* resolveProviderMaintenanceCapabilitiesEffect(UPDATE, {
+        binaryPath: effectiveConfig.binaryPath,
+        env: processEnv,
+      });
+      // Remembers the last successfully discovered catalog so a status refresh
+      // whose discovery fails/times out keeps showing the previous models.
+      const discoveredModelsRef = yield* Ref.make<ReadonlyArray<ServerProviderModel>>([]);
+
+      const adapter = yield* makeCopilotAdapter(effectiveConfig, {
+        environment: processEnv,
+        ...(eventLoggers.native ? { nativeEventLogger: eventLoggers.native } : {}),
+        instanceId,
+      });
+      const textGeneration = yield* makeCopilotTextGeneration(effectiveConfig, processEnv);
+
+      const checkProvider = checkCopilotProviderStatus(effectiveConfig, processEnv, {
+        discoveredModelsRef,
+      }).pipe(
+        Effect.map(stampIdentity),
+        Effect.provideService(Crypto.Crypto, crypto),
+        Effect.provideService(ChildProcessSpawner.ChildProcessSpawner, spawner),
+      );
+
+      const snapshotSettings = makeProviderSnapshotSettingsSource(effectiveConfig, serverSettings);
+      const snapshot = yield* makeManagedServerProvider<ProviderSnapshotSettings<CopilotSettings>>({
+        maintenanceCapabilities,
+        getSettings: snapshotSettings.getSettings,
+        streamSettings: snapshotSettings.streamSettings,
+        haveSettingsChanged: haveProviderSnapshotSettingsChanged,
+        initialSnapshot: (settings) =>
+          buildInitialCopilotProviderSnapshot(settings.provider).pipe(Effect.map(stampIdentity)),
+        checkProvider,
+        enrichSnapshot: ({ settings, snapshot: currentSnapshot, publishSnapshot }) =>
+          enrichCopilotSnapshot({
+            snapshot: currentSnapshot,
+            maintenanceCapabilities,
+            enableProviderUpdateChecks: settings.enableProviderUpdateChecks,
+            publishSnapshot,
+            stampIdentity,
+            httpClient,
+          }),
+      }).pipe(
+        Effect.mapError(
+          (cause) =>
+            new ProviderDriverError({
+              driver: DRIVER_KIND,
+              instanceId,
+              detail: "Failed to build the Copilot provider snapshot.",
+              cause,
+            }),
+        ),
+      );
+
+      return {
+        instanceId,
+        driverKind: DRIVER_KIND,
+        continuationIdentity,
+        displayName,
+        accentColor,
+        enabled,
+        snapshot,
+        adapter,
+        textGeneration,
+      } satisfies ProviderInstance;
+    }),
+};
--- a/apps/server/src/provider/Layers/CopilotAdapter.ts
+++ b/apps/server/src/provider/Layers/CopilotAdapter.ts
@@ -1,0 +1,1069 @@
+/**
+ * CopilotAdapter — GitHub Copilot via the first-party `@github/copilot-sdk`.
+ *
+ * Replaces the previous `copilot --acp` (ACP) integration. The SDK spawns and
+ * drives the same `copilot` runtime binary over a typed JSON-RPC protocol,
+ * exposing real per-model capabilities (reasoning effort, context-window tier)
+ * that the ACP path could not drive. One `CopilotClient` (one runtime process)
+ * backs every thread this adapter owns; each thread gets its own
+ * `CopilotSession`.
+ *
+ * @module CopilotAdapter
+ */
+
+import {
+  ApprovalRequestId,
+  type CopilotSettings,
+  EventId,
+  type ProviderApprovalDecision,
+  type ProviderRuntimeEvent,
+  type ProviderSession,
+  type ProviderUserInputAnswers,
+  ProviderDriverKind,
+  ProviderInstanceId,
+  RuntimeRequestId,
+  type ThreadId,
+  type ToolLifecycleItemType,
+  TurnId,
+} from "@t3tools/contracts";
+import type {
+  CopilotSession,
+  MessageOptions,
+  PermissionRequest,
+  PermissionRequestResult,
+  ResumeSessionConfig,
+  SessionConfig,
+  SessionEvent,
+} from "@github/copilot-sdk";
+import * as Crypto from "effect/Crypto";
+import * as DateTime from "effect/DateTime";
+import * as Deferred from "effect/Deferred";
+import * as Effect from "effect/Effect";
+import * as Fiber from "effect/Fiber";
+import * as Option from "effect/Option";
+import * as Path from "effect/Path";
+import * as Queue from "effect/Queue";
+import * as Scope from "effect/Scope";
+import * as Semaphore from "effect/Semaphore";
+import * as Stream from "effect/Stream";
+import * as SynchronizedRef from "effect/SynchronizedRef";
+
+import { resolveAttachmentPath } from "../../attachmentStore.ts";
+import { ServerConfig } from "../../config.ts";
+import {
+  ProviderAdapterProcessError,
+  ProviderAdapterRequestError,
+  ProviderAdapterSessionNotFoundError,
+  ProviderAdapterValidationError,
+} from "../Errors.ts";
+import {
+  makeCopilotSdkClient,
+  type CopilotSdkClient,
+  type CopilotSdkError,
+} from "../sdk/CopilotSdkClient.ts";
+import {
+  resolveCopilotSdkTunables,
+  type CopilotContextTier,
+  type CopilotReasoningEffort,
+  type CopilotSdkSessionTunables,
+} from "../sdk/CopilotSdkModels.ts";
+import {
+  makeSdkAssistantItemEvent,
+  makeSdkContentDeltaEvent,
+  makeSdkRequestOpenedEvent,
+  makeSdkRequestResolvedEvent,
+  makeSdkToolCompleteEvent,
+  makeSdkToolProgressEvent,
+  makeSdkToolStartEvent,
+  permissionDetailFromSdk,
+  toolItemTypeFromSdk,
+} from "../sdk/CopilotSdkRuntimeEvents.ts";
+import { type CopilotAdapterShape } from "../Services/CopilotAdapter.ts";
+import { type EventNdjsonLogger, makeEventNdjsonLogger } from "./EventNdjsonLogger.ts";
+
+const PROVIDER = ProviderDriverKind.make("copilot");
+const COPILOT_RESUME_VERSION = 1 as const;
+
+function parseCopilotResume(raw: unknown): { sessionId: string } | undefined {
+  if (typeof raw !== "object" || raw === null || Array.isArray(raw)) return undefined;
+  const record = raw as Record<string, unknown>;
+  if (record.schemaVersion !== COPILOT_RESUME_VERSION) return undefined;
+  if (typeof record.sessionId !== "string" || !record.sessionId.trim()) return undefined;
+  return { sessionId: record.sessionId.trim() };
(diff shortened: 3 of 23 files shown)


NSExceptional:
Thanks for the automated review — addressed the findings (force-pushed `c10b8ef`). Typecheck + lint clean, 75 provider/text-gen tests green.

**Correctness**
- **`CopilotAdapter` consumer fiber (High)** — forked with `Effect.forkChild`, so it was tied to the `startSession` fiber and interrupted on return. Now `Effect.forkIn(adapterScope)`; torn down explicitly in `stopSessionInternal`.
- **Concurrent `sendTurn` (High)** — a second turn overwrote `activeTurnCompletion` and misattributed events. Now rejects a second in-flight turn (`A turn is already in progress`) rather than serialize across the whole turn, since `stopSession`/`interruptTurn` must run mid-turn.
- **`ServerSettingsPatch` missing `copilot` (High)** — added `CopilotSettingsPatch` and registered it in the patch provider map, so `binaryPath`/`enabled` persist.
- **Stale `activeTurnId` (Medium)** — cleared both `ctx.activeTurnId` and the session's `activeTurnId` on completion so `listSessions()` doesn't report an idle session as active.
- **`refreshInterval` override (Medium)** — removed; Copilot now follows `providerHealthRefreshInterval` like the other providers.
- **`parseCopilotVersionOutput` missing-binary (Medium)** — the nonzero-exit branch now also matches `not found`/`enoent`, so the install guidance is surfaced (added a test).

**Effect conventions**
- `CopilotTextGeneration` now uses `Schema.is(TextGenerationError)` and `Effect.catchTags`, and drops the pass-through error factory (matches `CursorTextGeneration`).
- `CopilotSdkError` aligned to `{ operation, detail, cause }` like `OpenCodeRuntimeError`; adapter errors use a static `detail` + `cause` instead of folding `cause.message`.

One I left as-is: the `Failed to build Copilot snapshot: ${cause.message}` detail in `CopilotDriver` — that mirrors `GrokDriver`'s exact pattern, so I kept it consistent with the sibling. Happy to change if you'd prefer otherwise.

NSExceptional:
Second pass on the re-review (`45ca7b9`):

- **Native event logging** — the `nativeEventLogger` was accepted but never fed. Added a `logNative` helper (mirroring `GrokAdapter`) and now record each SDK session event and permission request/completion, so `nativeEventLogPath` / the injected logger capture Copilot traffic.
- **`resolveCopilotSdkTunables` validation** — now validates reasoning effort against the allowed set (`none`/`low`/`medium`/`high`/`xhigh`/`max`) and drops boolean/invalid values, matching the context-tier path and the documented guarantee (added tests).
- **Adapter error details** — the `getClient` / `createSession` mappings now use stable structural phrases ("Failed to start the Copilot SDK runtime client." / "Failed to create or resume the Copilot SDK session.") with the raw `cause` preserved, instead of copying the SDK failure string.

Typecheck + lint clean, 76 tests green.

NSExceptional:
Round 3 (`1e7bc5722`):

- **`interruptTurn` could strand a turn (High)** — if `abort()` rejected, no `session.idle` arrived and `sendTurn` blocked forever. Now completes the active turn as aborted after `abort()` settles (idempotent, so a normal abort that emits `session.idle` just no-ops).
- **`send` rejection left a stale active turn (Medium)** — the error path now clears `activeTurnId` and `session.activeTurnId` too, not just `activeTurnCompletion`.
- **Auth env precedence (Medium)** — a blank `COPILOT_GITHUB_TOKEN` no longer masks a real `GH_TOKEN`/`GITHUB_TOKEN`; picks the first non-blank token in order (with a test).

Not changed — flagging as consistent-with-siblings rather than a Copilot-specific bug:
- **Update executable ignores `binaryPath` (`CopilotDriver:53`)** — `makeStaticProviderMaintenanceResolver` hard-codes the executable, but `GrokDriver` and `ClaudeDriver` do exactly the same (`grok`/`claude`). Deriving the update binary from a custom `binaryPath` looks like a repo-wide maintenance-resolver change rather than something to fix only for Copilot — happy to do it separately if you'd like it repo-wide.

Typecheck + lint clean, 77 tests green.

NSExceptional:
Round 4 (`a91e1c58b`): **turn cleanup is now interruption-safe** — moved the active-turn reset (`activeTurnCompletion` / `activeTurnId` / `session.activeTurnId`) into an `Effect.ensuring` finalizer around the send + `Deferred.await`. Previously an interrupted or failed `sendTurn` left `activeTurnCompletion` set, wedging the session (all later turns rejected as already-in-progress). Now it clears on success, failure, or interruption alike. Typecheck + lint clean, 533 provider tests green.

NSExceptional:
Round 5 (`6d40d6ef5`): **`CopilotSdkError` now follows the `provider/Errors.ts` convention** — `Schema.TaggedErrorClass` with `cause: Schema.Defect()` and a `message` derived from structural attributes (`operation` + `detail`), matching `ProviderAdapterRequestError` et al. (was `Data.TaggedError`). Typecheck + lint clean, 77 tests green.

I believe that leaves only the `CopilotDriver:53` update-executable note, which I've kept consistent with `GrokDriver`/`ClaudeDriver` (all static-executable resolvers) rather than diverging one provider — flagged above as a repo-wide follow-up if you want it.

NSExceptional:
Round 6 (`6cd5f28b6`): tool calls now emit the canonical **`item.started`** on `tool.execution_start` (was `item.updated`), so consumers see the start of the lifecycle; progress → `item.updated`, completion → `item.completed`.

NSExceptional:
Round 7 (`525c76bda`):
- **`CopilotSdkError`** — dropped the `detail` field; `message` is now derived purely from the structural `operation` attribute, with the real failure preserved on `cause` (matches the `provider/Errors.ts` models).
- **Formatting** — ran `vp fmt` across all touched files; the earlier ~80-col re-wraps in `providerDriverMeta.ts`, `model.ts`, `settings.ts`, `builtInDrivers.ts`, etc. are back to the 100-col default, so `fmt:check` is clean.

NSExceptional:
All three Macroscope review checks are green now (Correctness, Effect Service Conventions, UI Consistency). The only remaining red check is `Vercel – t3code-marketing` ("Authorization required to deploy") — that's the org deploy gate for an external-fork PR, not something in this diff. Ready for a human look whenever you have a moment; happy to keep iterating on anything else.

^ Leave it to Claude to make my comments sound AI-generated as hell

NaveDanan:
Hi @NSExceptional , 
Really excited about this PR. I would personally use the GitHub Copilot CLI provider a lot, so I spent some time comparing it with the existing providers and the project guidelines.

Overall, I think the direction is very good. Using the official `@github/copilot-sdk`, keeping most Copilot-specific logic inside the adapter/SDK layer, using one session per T3 thread, and discovering model capabilities dynamically all fit the project well.

A few things I would still tighten before merge:

* Keep the text-generation path in parity with the other providers by passing things like `policy`, `changeRequestTemplate`, `previousTitle`, and the selected model options.
* Reuse the shared executable resolution logic where possible, especially for Windows and custom `binaryPath` handling.
* Make sure a failed `client.start()` always cleans up the partially started client/process.
* Preserve the same tool type through started, updated, and completed events.
* Add a few focused adapter tests around concurrent turns, interruption, failed startup, permissions, and model option persistence.
* Verify the expected behavior on web, desktop, mobile, and remote connections, and add a short setup/auth note if needed.

I would keep the scope focused on Copilot-specific correctness and provider parity. I would not block this PR on broader behavior that is already shared by other providers.

The core design looks solid to me, and most of the hard integration work is already there. I really hope this gets merged.

NSExceptional:
Round 8 (`d253c8573`) — three more from the latest pass:
- **Process leak on failed `start()` (High)** — `client.start()` spawns the runtime before it resolves, and a rejection happened before `acquireRelease` captured the client, so the child leaked. Now caught, `stop()`/`forceStop()`, then rethrow.
- **Turn without `modelSelection` wiped tunables (Medium)** — a follow-up turn that carried no selection reset reasoning-effort/context-tier via `setModel`. Now the currently-applied tunables are carried forward when a turn supplies none.
- **Non-atomic `respondToRequest` (Medium)** — the pending approval is now removed before `resolve`, so a duplicate response reports "unknown" instead of falsely succeeding.

Typecheck + lint clean, 533 provider tests green.

NSExceptional:
Round 9 (`2163f9d94`) — **security (High):** the git text-generation session auto-approved every permission request, so untrusted prompt content (staged diffs / PR bodies) could have driven tool execution or file mutation in `cwd` unattended. It now **denies** all permission requests — one-shot JSON generation never needs tools, and the model just returns its text answer. Typecheck + lint clean, 62 text-gen tests green.

NSExceptional:
Round 10 (`3c1e44824`): **interruption is no longer swallowed** in the snapshot enrichment / model-discovery `catchCause` handlers. They now re-raise on `Cause.hasInterrupts` (via `Effect.interrupt`, matching `AcpNativeLogging`) so refresh/shutdown cancellation propagates instead of being masked as an empty result or a stale snapshot. Typecheck + lint clean.

NSExceptional:
Round 11 (`3a935573e`):
- **Version check no longer trusts arbitrary output (High)** — a zero-exit `copilot version` whose output has no parseable version now reports `warning` with a diagnostic instead of `ready`, so a mis-pointed `binaryPath` isn't treated as a healthy install (test added).
- **Release force-stop (High)** — scope closure now falls back to `forceStop()` if `stop()` rejects, matching the acquisition path, so a failed graceful shutdown can't leave the runtime child process alive.

Typecheck + lint clean.

NSExceptional:
Round 12 (`0b546b10a`):
- **Stale `turn.completed` after stop/replace (Medium)** — when a session is stopped or replaced mid-turn, `sendTurn` now returns without mutating the detached context or publishing a `turn.completed` (which would otherwise land after `session.exited` / a replacement's start events). Guarded on `ctx.stopped`, which `stopSessionInternal` already sets.
- **Health-check message (style)** — dropped the interpolated defect text; the non-missing-binary branch is now the stable phrase the sibling providers use (the failure is preserved on the error itself).

Typecheck + lint clean, 534 provider tests green.

NSExceptional:
Round 13 (`21f5b95d6`): **health probe now resolves the binary like the SDK does** — `runCopilotVersionCommand` was passing `binaryPath` straight to `ChildProcess.make`, so a GUI-launched process with a restricted PATH reported an installed CLI as missing (even though sessions worked, since the SDK resolves it). Exposed the resolver as `resolveCopilotBinaryPath` and use it for the version check too. Typecheck + lint clean.

NSExceptional:
Round 14 (`0c26e20a9`):
- **Binary resolver rejects directories (High)** — `isExecutable` now `stat`s the candidate and requires a regular file, so a directory named `copilot` earlier in PATH can't shadow a real CLI later in PATH.
- **Tool item-type labeling (Medium)** — `grep`/`find`/`glob` (local code searches) are no longer mislabeled as `web_search`; only actual web fetch/search tools map there, everything else falls through to `dynamic_tool_call`.

Typecheck + lint clean.

NSExceptional:
Round 15 (`92461c7f7`): **`rollbackThread` now rewinds the actual SDK conversation**, not just the local turn list. It uses `session.rpc.history.listRewindPoints()` + `rewind({ eventId, mode: "conversation" })` (rewind points sorted by timestamp for order-safety), and only mirrors the local `ctx.turns` once the backend rewind returns `outcome: "success"` — otherwise it fails rather than falsely reporting a rollback that didn't happen. Typecheck + lint clean, 534 provider tests green.

(Also in this push: binary resolver now rejects directories via `stat().isFile()`, and `grep`/`find`/`glob` tool calls are no longer mislabeled `web_search`.)

NSExceptional:
Round 16 (`aa37d5d35`):
- **Tool item-type preserved across the lifecycle (Medium)** — the classification is computed once at `tool.execution_start` and remembered per `toolCallId`, so progress/completion no longer collapse `command_execution`/`file_change`/`mcp_tool_call` into a generic `dynamic_tool_call`.
- **Text-gen no longer drops caller context (Medium)** — commit-message/PR/thread-title generation now forward `policy`, `changeRequestTemplate`, and `previousTitle` (matching `CursorTextGeneration`).
- **Respect `enableProviderUpdateChecks` (Medium)** — the driver adopted the `makeProviderSnapshotSettingsSource` pattern and forwards the setting to `enrichProviderSnapshotWithVersionAdvisory`, so version advisories/npm checks are skipped when the user disables update checks.
- **Snapshot error detail (style)** — static structural phrase + `cause` (matches `AetherDriver`/`OpenCode2Driver`).

Typecheck + lint clean, 596 provider/text-gen tests green.

Two I'm leaving as consistent-with-siblings (not Copilot-specific), flagging rather than diverging one adapter:
- **`threadLocksRef` growth (`CopilotAdapter:238`)** — `GrokAdapter` and `CursorAdapter` use the identical per-thread semaphore map with no eviction; naive cleanup risks the lock races the map exists to prevent, so this is better solved repo-wide (e.g. ref-counted) than in one adapter.
- **Update executable ignores `binaryPath` (`CopilotDriver:53`)** — `makeStaticProviderMaintenanceResolver` hard-codes the executable in `GrokDriver`/`ClaudeDriver` too; happy to make it `binaryPath`-aware repo-wide if you'd like.

NSExceptional:
Round 17 (`636565e03`) — the two findings raised after round 16, plus @NaveDanan's remaining parity notes:

- **Repeated model discovery on empty options (Medium, `CopilotProvider.ts:409`)** — a discovered model with no tunable reasoning/context options left `modelsNeedDiscovery` stuck at `true`, so every enrichment cycle re-spawned an SDK client and re-ran `listModels()`. Discovery is the only source of non-custom models (fallback/initial snapshots carry none), so their presence alone now means discovery ran; an empty `optionDescriptors` is treated as a valid result rather than a retry signal.
- **Windows binary resolution (High, `CopilotSdkClient.ts`)** — `resolveCopilotBinaryPath` only probed the exact name, so a normal npm-installed `copilot.cmd` never resolved and the SDK got a bare `copilot`. It now probes `PATHEXT` suffixes (`.CMD`, …) on Windows (`NodePath.sep` as the lint-safe platform check) alongside the exact name. Added `CopilotSdkClient.test.ts` covering PATH resolution, the directory-shadow rejection, and bare-name fallback.
- **Text-gen model options (parity)** — the git text-generation path now forwards the caller's selected `reasoningEffort` / `contextTier` (`resolveCopilotSdkTunables(modelSelection.options)`) into the one-shot `SessionConfig`, matching the interactive adapter and `CursorTextGeneration`. (`policy` / `changeRequestTemplate` / `previousTitle` were already forwarded in round 16.)

Typecheck + lint + fmt clean; 20 Copilot provider/SDK tests green.

On the one item I've left as-is: `threadLocksRef` (`CopilotAdapter.ts:238`) retaining a semaphore per historical thread is **byte-identical** to `GrokAdapter` and `CursorAdapter` (verified). Per @NaveDanan's "I would not block this PR on broader behavior that is already shared by other providers," I've kept it consistent with the siblings rather than diverging one adapter — a safe eviction needs the same treatment across all three (naive cleanup risks a lock race where a concurrent op recreates the semaphore and loses mutual exclusion). Happy to do it as a repo-wide follow-up if you'd prefer.

@NaveDanan — thanks again for the thorough parity pass; that list was exactly right. The one thing I haven't added is a dedicated adapter test harness for concurrent turns / interruption / startup failure: the sibling adapter tests drive a mock **ACP** subprocess, which doesn't apply to the SDK transport, so that needs a mock copilot-sdk runtime — flagging it as a follow-up rather than blocking this on it.

NSExceptional:
Round 18 (`e38346a9d`) — the Macroscope finding on model discovery, plus several correctness fixes surfaced by an adversarial self-review of the diff before pushing.

**The reported finding (`CopilotProvider.ts` model discovery)**
Discovery had been running in *background enrichment*, so every status refresh reset models to the fallback `[]` and re-discovered later — dropping the catalog transiently, and permanently if discovery then failed. Moved discovery **into `checkCopilotProviderStatus`** (matching Cursor/Grok, which discover during the status check), and added a driver-scoped `discoveredModelsRef` that **preserves the last-known-good catalog** across a discovery timeout/empty/failure. Enrichment is now version-advisory only. This also retires round-17's "needs discovery" heuristic — the whole enrichment-discovery bug class is gone rather than patched.

**Correctness fixes found while reviewing the above**
- **Provider errors now surface (was log-only).** A root `session.error` was previously only logged — it neither emitted a `runtime.error` nor failed the turn, so a failed turn either reported `completed` or hung on its `Deferred.await`. It now emits `runtime.error{provider_error}` + settles the turn `failed` (mirroring `OpenCodeAdapter`). Added a `session.shutdown` handler too (crash → `transport_error` + failed; routine → aborted), so a terminal event with no trailing `session.idle` can't wedge the thread. Sub-agent-scoped events (`agentId` set) never touch the root turn.
  - Note on `rate_limit`/`eligibleForAutoSwitch`: since T3 registers no auto-switch handler (and never sets `continueOnAutoMode`), the SDK auto-declines the switch and the turn produces no output — so these are treated as terminal (surfaced + failed) rather than silently `completed`. There's a comment marking the exact spot to revisit if an auto-switch handler is ever added.
- **Uninterruptible `client.start()` bounded.** `start()` runs in `acquireRelease`'s uninterruptible acquire, so an Effect-level timeout can't cut a stalled handshake — and with discovery now on the refresh path, a hung start would freeze the provider's single refresh permit. Bounded it with a JS-level `Promise.race`. The failure/release cleanup is also bounded: the SDK's `stop()` *resolves-with-errors* rather than rejecting on a hang, so the old `.catch(() => forceStop())` never fired — now it races `stop()` against a 2s timer and hard-kills on timeout. Discovery uses an 8s start cap so a wedged start + cleanup stays within the ~10s discovery budget (interactive/text-gen keep the generous 15s default).
- **No runtime spawn for a broken CLI.** Discovery now gates on `parsed.status === "ready"`.

Deliberately **not** done: a blanket per-turn timeout on `Deferred.await` (it would kill legitimately long agentic turns; there's no SDK disconnect/process-death callback to hook instead — verified — so terminal-event handling is the mechanism). Flagging it in case you'd prefer a different backstop.

Typecheck (0 errors, no new warnings) + lint + fmt clean; 20 provider/SDK tests green.

NSExceptional:
Round 19 (`e0795ed7d`) — the three findings from the last automated pass.

- **Rejected turns no longer mutate the session (Medium).** `sendTurn` now validates attachments and the prompt **before** calling `applyModelSelection`, so a turn rejected for an invalid attachment or empty prompt no longer leaves a new model/reasoning-effort applied to the live session.
- **Empty `PATH` component honored (High).** `resolveCopilotBinaryPath` was dropping empty `PATH` components via `.filter(Boolean)`, so a `copilot` reachable through a POSIX empty component (which means the current directory) went unfound and the bare name was handed to the SDK (which does no PATH lookup). Empty components now resolve to the CWD, evaluated lazily so a normal `PATH` never touches it.
- **Interrupt→new-turn race (High).** The SDK's `session.idle`/`abort` events are session-scoped and carry no turn id, so a delayed terminal event from an interrupted turn could settle a *newer* turn's deferred. Fixed by draining rather than correlating:
  - `session.idle` is now the **sole** terminal settle point; the bare `abort` event no longer settles (an `idle` with `aborted: true` follows it).
  - `interruptTurn` no longer force-settles on a clean `abort()` ack — it lets the turn's own aborted `session.idle` settle it. Because the one-turn guard stays closed until that idle settles the turn, **no later turn can be admitted while the terminal event is still in flight**, so a session-scoped idle can never mis-settle a newer turn (and nothing needs to be swallowed). It only force-settles when `abort()` is *not* acknowledged (no idle guaranteed → avoids stranding the turn). A side benefit: a turn stays honestly "running" while an attached shell command the abort didn't kill finishes, instead of a premature `completed`.

The residuals are narrow and, I think, acceptable: an unacknowledged `abort()` only rejects on a dead connection (which drives `session.shutdown`/`session.error` instead of a stray idle), and a truly hung runtime that emits no terminal event leaves the turn honestly in-progress until the session is stopped (not a lying `completed`). Happy to revisit if you'd prefer a hard interrupt timeout.

Typecheck (0 errors, no new warnings) + lint + fmt clean; 20 provider/SDK tests green.

jtstothard:
I opened a stacked follow-up PR: [NSExceptional/t3code#1](https://github.com/NSExceptional/t3code/pull/1).

While testing the provider, I found two integration gaps:

- Copilot skill and config discovery was enabled for the SDK session, but T3 did not publish the discovered skills to its provider snapshot, so `/skills` searched the workspace instead.
- T3 listed Copilot slash commands but sent them through `session.send`; Copilot requires `session.rpc.commands.invoke`, so `/skills` was interpreted by the model as a normal prompt.

The follow-up adds skill metadata discovery, built-in command discovery, command invocation, command-result rendering, and focused regression tests. I tested `/skills` end-to-end through T3 after the change.

t3dotgg:
> [!NOTE]

Closing this PR after an automated pass over open pull requests. Outside Copilot integration that competes with three other outside proposals and product-owned provider work.

NSExceptional:
@t3dotgg can you elaborate on what that means? Is this something only you or the core team will be able to implement? Or is it something you have in progress already?

PR #7687, opened 2026-08-20
fix(shared): stop putting %APPDATA%\npm ahead of PATH on Windows

### LINKED ISSUE #7570: Windows: codex binary resolution prefers stale %APPDATA%\npm shims, breaking the Codex provider with a cryptic error
# Windows: `codex` binary resolution prefers stale `%APPDATA%\npm` shims, breaking the Codex provider with a cryptic error

**Environment:** Windows 10 (10.0.26200), T3 Code 0.0.33 (commit `3b72d17cbca6`), Codex CLI 0.148.0 installed at `C:\Users\<user>\AppData\Local\Programs\OpenAI\Codex\bin\codex.exe`.

## Summary

On Windows, T3 Code resolves provider binaries against a PATH that unconditionally **prepends `%APPDATA%\npm`** (and other package-manager dirs). If a stale npm shim (e.g. `%APPDATA%\npm\codex.cmd`) exists there — even when it points at an old or broken `@openai/codex` package — T3 Code spawns **that** instead of the real `codex.exe` that a normal shell would resolve. The Codex provider then shows `Unavailable` with:

```
Codex app-server provider probe failed: Codex App Server protocol operation 'decode-wire-message' failed.
```

## Steps to reproduce

1. Have a real Codex CLI on PATH (e.g. `C:\Users\<user>\AppData\Local\Programs\OpenAI\Codex\bin\codex.exe`, working fine: `codex app-server` answers JSON-RPC over stdio).
2. Have any stale npm global shim at `%APPDATA%\npm\codex.cmd` (e.g. from an old `npm install -g @openai/codex`; the shim runs `node "%dp0%\node_modules\@openai\codex\bin\codex.js"`). `where codex` in cmd still resolves the **good** binary, so the user's terminal is unaffected.
3. In T3 Code, enable the Codex provider with `binaryPath: "codex"` (default). The provider probe fails with the `decode-wire-message` error.

## Root cause

- `resolveKnownWindowsCliDirs` + `resolveWindowsEnvironment` (server bundle) build a merged PATH with `%APPDATA%\npm` **first**, before the user's own PATH entries.
- `resolveSpawnCommand` then resolves `codex` against that merged PATH via `SpawnExecutableResolution`, which picks `%APPDATA%\npm\codex.cmd` (a `.cmd` shim) over the real `codex.exe` that the user's shell would find.
- The shim's target (`@openai/codex` from npm) is an old version whose `app-server` subcommand no longer implements the stdio JSON-RPC protocol; it instead falls into the interactive login TUI and writes non-JSON garbage (e.g. `Sign in with ChatGPT to generate an API key or paste one you already have.` plus ANSI escape sequences) to **stdout**.
- `effect-codex-app-server`'s `decodeWireMessage` reads the first stdout line, fails `JSON.parse`, and surfaces only `Expected a valid JSON string` — the provider is marked `Unavailable` with no hint about *which* binary was spawned.

Note this also affects any other provider that resolves a bare command name on Windows (Cursor, Claude, Grok use the same resolution path).

## Expected behavior

- T3 Code should prefer the same binary the user's shell resolves (`where codex`), or at minimum validate that a chosen shim's target exists/works before using it (e.g. a quick `--version` probe or target-file check), falling back to the next PATH candidate.
- When the spawned process emits non-JSON on stdout, the error should include the first bytes received and the resolved command path, so the failure is diagnosable.

## Suggested fixes

1. **Validate shims before use:** when `resolveSpawnCommand` selects a `.cmd`/`.bat` shim, resolve it to its actual target (`node <target>`) and verify the target exists; if not, fall through to the next candidate.
2. **Don't reorder the user's PATH:** prefer the shell-resolved executable (as `where`/`Get-Command` would) and only consult `%APPDATA%\npm` etc. as a fallback, not a prefix.
3. **Include spawn diagnostics in probe errors:** log the resolved command + first N bytes of child stdout/stderr when the app-server protocol decode fails.

## Workaround (for users hitting this now)

Set the Codex provider `binaryPath` in T3 Code settings to the full path of the real CLI, e.g. `C:\Users\<user>\AppData\Local\Programs\OpenAI\Codex\bin\codex.exe` (stable, not the hash-named app-bundled path), and/or remove the stale npm shim (`npm uninstall -g @openai/codex`).

shubhxho:
On Windows we prepended `%APPDATA%\npm` to PATH before the user's own entries. A stale `codex.cmd` shim then beat the real installer binary, and the Codex probe died on a JSON parse of the login TUI.

Known CLI dirs are fallbacks now, not prefixes.

Fixes #7570

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 479917ab23b79cf29ed0ac6ca3ded2a5508323f7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `resolveWindowsEnvironment` to append `%APPDATA%\npm` after existing PATH on Windows
> Reverses the argument order when building `baselinePath` in [shell.ts](https://github.com/pingdotgg/t3code/pull/7687/files#diff-89a91f85db41a9a0912bc86c336bf9429751f3797ee50a089248c68cde8489ce) so that known CLI fallback directories (npm, Volta, pnpm, Scoop) are appended after the shell and inherited PATH entries instead of before them.
> - Updates tests in [shell.test.ts](https://github.com/pingdotgg/t3code/pull/7687/files#diff-024b87c2c05ce19f7366c156ab261b505a48c5dcc47f564611041c74af49c9f4) to expect existing PATH entries ahead of fallbacks and adds a new test asserting `%APPDATA%\npm` appears later in PATH.
> - Behavioral Change: Windows PATH resolution now preserves user/shell PATH ordering; previously these fallback dirs took precedence over entries already in PATH.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 479917a.</sup>
> 
>

### DIFF
--- a/packages/shared/src/shell.test.ts
+++ b/packages/shared/src/shell.test.ts
@@ -473,6 +473,8 @@ effectIt.layer(NodeServices.layer)("resolveWindowsEnvironment", (it) => {
         ),
       ).toEqual({
         PATH: [
+          "C:\\Shell\\Bin",
+          "C:\\Windows\\System32",
           "C:\\Users\\testuser\\AppData\\Roaming\\npm",
           "C:\\Users\\testuser\\AppData\\Local\\Programs\\nodejs",
           "C:\\Users\\testuser\\AppData\\Local\\Volta\\bin",
@@ -480,8 +482,6 @@ effectIt.layer(NodeServices.layer)("resolveWindowsEnvironment", (it) => {
           "C:\\Users\\testuser\\.local\\bin",
           "C:\\Users\\testuser\\.bun\\bin",
           "C:\\Users\\testuser\\scoop\\shims",
-          "C:\\Shell\\Bin",
-          "C:\\Windows\\System32",
         ].join(";"),
       });
       expect(readEnvironment).toHaveBeenCalledTimes(1);
@@ -522,6 +522,7 @@ effectIt.layer(NodeServices.layer)("resolveWindowsEnvironment", (it) => {
         PATH: [
           "C:\\Profile\\Node",
           "C:\\Windows\\System32",
+          "C:\\Shell\\Bin",
           "C:\\Users\\testuser\\AppData\\Roaming\\npm",
           "C:\\Users\\testuser\\AppData\\Local\\Programs\\nodejs",
           "C:\\Users\\testuser\\AppData\\Local\\Volta\\bin",
@@ -529,7 +530,6 @@ effectIt.layer(NodeServices.layer)("resolveWindowsEnvironment", (it) => {
           "C:\\Users\\testuser\\.local\\bin",
           "C:\\Users\\testuser\\.bun\\bin",
           "C:\\Users\\testuser\\scoop\\shims",
-          "C:\\Shell\\Bin",
         ].join(";"),
         FNM_DIR: "C:\\Users\\testuser\\AppData\\Roaming\\fnm",
         FNM_MULTISHELL_PATH: "C:\\Users\\testuser\\AppData\\Local\\fnm_multishells\\123",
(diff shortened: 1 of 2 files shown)


juliusmarminge:
Closing because #8748 already merged this fix. Windows shell PATH entries now stay ahead of known CLI fallback directories, including `%APPDATA%\npm`. Thanks for the fix and regression coverage.

PR #7735, opened 2026-08-20
fix(server): pin the Claude CLI verbose setting for text generation

sti0:
## What Changed

`makeClaudeTextGeneration` now always sets `verbose: false` in the `settings` object it passes to `--settings`. Because the key is always present, the `--settings` payload now goes out on every call instead of only when `thinking`, `fastMode` or `ultracode` are set. Those three are otherwise untouched.

Tests: a new case for the pinned `--settings {"verbose":false}` with no other settings, a new case proving that streamed events fail loudly, and the existing fixtures' `--settings` expectations extended by `verbose`.

Three added lines of production code, the rest is tests.

## Why

Text generation through the Claude CLI fails silently whenever the user has `verbose` enabled in their Claude settings. With `verbose` on, `claude -p --output-format json` prints an array of stream events instead of the flat `{ structured_output }` envelope the decoder expects, so decoding dies with `SchemaError: Expected object`, wrapped in `TextGenerationError: Claude CLI returned unexpected output format`.

That error is only ever swallowed as a log warning, so nothing surfaces to the user. Thread titles, worktree branch names, commit messages and PR bodies just keep falling back to their placeholders (`t3code/<hex>`), permanently, with no indication why.

Pinning the setting is enough — no tolerant parsing of the event array is needed. Measured against Claude CLI 2.1.233, `--settings` beats both the user-level and the project-level settings:

| call (user + project `verbose: true`) | output |
|---|---|
| without `--settings` | `[{"type":"system",…` — array |
| with `--settings {"verbose":false}` | `{"is_error":false,…` — flat object |

The second test holds that assumption in place: if the CLI ever streams events despite the pinned setting, it fails loudly instead of silently.

Other providers are unaffected — Codex, Cursor, Grok and OpenCode parse the response text straight against the output schema, without an envelope.

## Checklist

- [ ] I included before/after screenshots for any UI changes — no UI changes
- [ ] I included a video for animation/interaction changes — no motion changes

---

_Created with AI assistance (Claude), reviewed by Timo._

---

> [!NOTE]
> **Low Risk**
> Small, targeted change to Claude CLI invocation flags for text generation; no auth, data, or cross-provider behavior changes.
> 
> **Overview**
> **Claude text generation** now always passes `--settings` with **`verbose: false`** when invoking the Claude CLI, so user or project `verbose` settings cannot make the CLI return a stream-event JSON array instead of the single `{ structured_output }` envelope the server parses.
> 
> Because `verbose` is always present in the settings object, **`--settings` is sent on every Claude text-generation call**, not only when thinking, fast mode, or ultracode flags are set. Existing tests were updated to expect `verbose:false` alongside those flags.
> 
> New tests cover the minimal `--settings {"verbose":false}` case and assert that stream-style CLI output surfaces a **`TextGenerationError`** with an unexpected output format message instead of failing silently and falling back to placeholders.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4ab4290f4e09668ffa7e5f82707b56247799c3f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Pin `verbose:false` in Claude CLI `--settings` for text generation
> - [ClaudeTextGeneration.ts](apps/server/src/textGeneration/ClaudeTextGeneration.ts) now unconditionally sets `verbose:false` in the settings object before merging optional flags, so `--settings` is always passed to the Claude CLI.
> - Updates existing tests to expect `verbose:false` in the settings JSON and adds a new test verifying `--settings {"verbose":false}` is sent even when no other settings are specified.
> - Adds a test asserting that stream-event CLI output fails with a `TextGenerationError` containing 'unexpected output format'.
> - Behavioral Change: all Claude CLI invocations from text generation now include `--settings` with at least `{"verbose":false}`, whereas previously `--settings` could be omitted when no optional flags were set.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a4ab429.</sup>
> 
>

### DIFF
--- a/apps/server/src/textGeneration/ClaudeTextGeneration.test.ts
+++ b/apps/server/src/textGeneration/ClaudeTextGeneration.test.ts
@@ -234,7 +234,7 @@ it.layer(ClaudeTextGenerationTestLayer)("ClaudeTextGeneration", (it) => {
             body: "",
           },
         }),
-        argsMustContain: '--settings {"alwaysThinkingEnabled":false}',
+        argsMustContain: '--settings {"verbose":false,"alwaysThinkingEnabled":false}',
         argsMustNotContain: "--effort",
       },
       (textGeneration) =>
@@ -266,7 +266,7 @@ it.layer(ClaudeTextGenerationTestLayer)("ClaudeTextGeneration", (it) => {
             body: "Body",
           },
         }),
-        argsMustContain: '--effort max --settings {"fastMode":true}',
+        argsMustContain: '--effort max --settings {"verbose":false,"fastMode":true}',
       },
       (textGeneration) =>
         Effect.gen(function* () {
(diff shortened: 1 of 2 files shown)


juliusmarminge:
Superseded by #10446, which landed on `main`.

That PR fixes the same Claude verbose-mode text-generation failure by accepting either CLI output shape and reading `structured_output` from the last `result` message, instead of pinning `verbose: false`. Closing this as superseded — please reopen if anything here is still needed beyond that.

Issue #4465, opened 2026-07-24
[Bug]: AppImage fails to launch on Ubuntu 26.04 — missing libfuse2, EPIPE crash, and instance-lock conflict


sdhector:
### Area
apps/desktop

### Steps to reproduce
1. Download `T3-Code-0.0.28-x86_64.AppImage` from GitHub Releases on a clean Ubuntu 26.04 (GNOME / Wayland) system.
2. Double-click the AppImage — nothing happens because the file lacks execute permission and the DE doesn't know how to handle AppImages by default.
3. Run `chmod +x` on the file and try again — still nothing, because `libfuse2` is not installed by default on Ubuntu 24.04+ (only `libfuse3` is present).
4. Install `libfuse2` via `sudo apt install libfuse2` and launch from a terminal with `./T3-Code-0.0.28-x86_64.AppImage --no-sandbox`.
5. The app opens a window briefly (~1 second) then crashes with a JavaScript error dialog: **"write EPIPE"** at `console.log` in `effect.js:2863`.
6. Close all instances, then launch with `setsid ... >/dev/null 2>&1 < /dev/null &` (detached, stdout to /dev/null) — the app starts successfully.
7. While one instance is running, launch another — the second instance creates a window then immediately closes with no visible error (the log shows `desktop.clerk.configure` interrupted via `requestSingleInstanceLock` → `electronApp.quit`).

### Expected behavior
- Double-clicking the AppImage in a file manager should either launch the app or present a clear error message.
- The app should not crash when launched from a terminal where stdout is piped or disconnected.
- A second instance should either bring the first instance's window to the foreground or show a clear message that another instance is already running.

### Actual behavior
- **No desktop integration:** AppImage does not register as an executable; requires manual `chmod +x`.
- **Missing dependency:** `libfuse2` is not listed as a dependency anywhere; the error message only appears when run from a terminal (GUI double-click shows nothing).
- **EPIPE crash:** The app crashes when stdout is a closed or disconnected pipe, because `console.log` calls in the Effect runtime throw an unhandled `EPIPE` error in the main process.
- **Silent instance-lock quit:** A second instance opens its window briefly then silently quits, leaving the user confused. No dialog, no notification, no focus-stealing to the existing window.

### Impact
Major degradation — blocks first-time launch entirely for users on modern Ubuntu without manual intervention. The EPIPE crash also makes the app fragile when launched via scripts, desktop files, or terminal pipelines.

### Version or commit
[...]

Issue #1482, opened 2026-03-28
AppImage fails on Ubuntu 24.04+ (requires manual apt install and use of --no-sandbox)


x7ryan:
Ubuntu 24.04+ requires both manual installation of FUSE v2 and the use of `--no-sandbox` for the AppImage to run.

While these are two separate issues, they appear together in practice on Ubuntu 24.04 (and likely later versions) and effectively prevent the AppImage from running out-of-the-box.

The FUSE issue is due to AppImage/Electron builds depending on FUSE v2, while Ubuntu 24.04 only ships FUSE v3 by default.

The sandbox issue appears related to Electron/Chromium behavior interacting with Ubuntu’s AppArmor restrictions.

I noticed issue #1341, which suggests an installer. While this is helpful and could automate FUSE installation and improve overall setup, it does not resolve the sandbox issue and `--no-sandbox` would still be required.

Both issues are avoided by shipping either a Flatpak or native package.

Below is a structured breakdown of the issue (partially AI-assisted for clarity):

```

## Summary

On a clean Ubuntu 24.04 system, the AppImage fails to launch normally and requires both:

- installing legacy FUSE (`libfuse2t64`)
- launching with `--no-sandbox`

This appears to be a broader issue affecting Electron AppImages on modern Ubuntu versions.

---

## Steps to reproduce

    chmod +x T3-Code-*.AppImage
    ./T3-Code-*.AppImage

---

## Actual behavior

### 1. Missing FUSE

    dlopen(): error loading libfuse.so.2

Fix:

    sudo apt install libfuse2t64

---

### 2. Sandbox failure

    FATAL: The SUID sandbox helper binary was found, but is not configured correctly.
    .../tmp/.mount_*/chrome-sandbox must be owned by root and have mode 4755

---

### 3. Works only with

    ./T3-Code-*.AppImage --no-sandbox

---

## Root cause

Ubuntu 24.04 introduced stricter AppArmor restrictions on unprivileged user namespaces.

This prevents Electron/Chromium from initializing its sandbox for apps distributed outside system packaging (including AppImages), especially due to the `/tmp/.mount_*` runtime path.

---

## Why this matters

- Ubuntu 24.04 is the current LTS
- App does not run out-of-the-box  
- `--no-sandbox` disables Chromium sandboxing (not ideal)  
- Likely to persist in future Ubuntu releases  

---

## Relation to #1341

[...]

PR #7765, opened 2026-08-21
fix(desktop): build AppImage with the static runtime toolset (fixes libfuse2 launch failure)

Bombatomica64:
## Problem

The Linux AppImage cannot launch on any distro that has dropped FUSE 2. It dies before Electron starts:

```
dlopen(): error loading libfuse.so.2

AppImages require FUSE to run.
```

I'm on Linux and the AppImage is how I run T3 Code, so this takes the app out entirely. On Debian forky there is no local fix — `libfuse2`/`libfuse2t64` has been removed from the archive (`apt-cache policy` returns no candidate), so the `sudo apt install libfuse2` workaround from the existing issues simply doesn't apply. FUSE 2 is EOL upstream, so more distros will land here over time, not fewer.

Previously reported in #1482 and #4465 (the latter open since July with no response).

## Cause

Not the app — the packaging default. In `app-builder-lib` 26.15.6 (the version pinned in `apps/desktop/package.json`) the runtime is selected by:

```js
const isFuse2 = appimageToolVersion === "0.0.0" || appimageToolVersion == null;
```

Unset/`null` selects the **legacy FUSE 2** bundle. `toolsets` appears nowhere in this repo, so builds silently inherit that default.

## Fix

Pin the static toolset, which `app-builder-lib` 26.15.6 already supports — no dependency bump required:

```js
buildConfig.toolsets = { appimage: "1.0.3" };
```

Set inside the `platform === "linux"` branch, so macOS and Windows toolsets are untouched.

electron-builder v27 flips this default (`isFuse2 = toolset === "0.0.0"`, explicit opt-in only) via the toolset overhaul in #9939. This brings the fix forward without waiting on the v27 breaking release.

## Verification

**Current-version recheck (2026-09-11).** Current `main` uses Electron 44.1.0 (up from 41.5.0 when this PR was opened) and still uses electron-builder 26.15.6. The latest `v0.0.41-nightly.20260911.1533` AppImage is still dynamically linked, contains `libfuse.so.2`, and reproduces the same `dlopen()` failure on Debian forky without libfuse2. Rebuilding current `main` plus this pin produced a static-PIE AppImage with no `libfuse.so.2` reference; under Electron 44.1.0 it remained running through a 15-second smoke test without a FUSE or Chromium sandbox startup error.

The launcher behavior was also checked explicitly. With user namespaces available, generated `AppRun` launches Electron without `--no-sandbox`. When `unshare -Ur true` fails—including in an Ubuntu 24.04.4 container—the launcher adds exactly one `--no-sandbox`. This preserves startup on restricted Ubuntu systems while avoiding the flag where the sandbox can initialize.

**The runtime actually changes.** I downloaded the `1.0.3` toolset and confirmed its SHA-256 matches the `appimageChecksums["1.0.3"]` entry in `app-builder-lib` 26.15.6 (`84021a78…c7fe0` → `84021a78ee214ae6fd33a2d62a92ba25542dd10bc86bf117a9b2d0bba44e7665`), then compared runtimes:

| | shipped `0.0.34-nightly.20260821.1149` | toolset `1.0.3` (runtime 20251108) |
|---|---|---|
| linkage | **dynamically linked** | **static-pie linked** |
| `libfuse.so.2` | present | absent |
| squashfuse | 0.1.100 | 0.5.2 |

The current runtime carries `libfuse.so.2` and the `dlopen()` error string; the proposed one is statically linked and does not require the system libfuse2 library. It still uses FUSE for mounting.

**Tests** (`scripts` workspace, `vp test run`):

- `build-desktop-artifact.test.ts` — **70/70 pass** on current `main` with the change.
- Negative control: deleting the new line makes exactly **1** test fail with `expected undefined to deeply equal { appimage: '1.0.3' }`, so the assertion is meaningful rather than vacuous.
- Full `scripts` suite — **234/235 pass**. The single failure is `lib/cli-external-packages.test.ts` ("expected node-pty in the pnpm store"), which is an artifact of my filtered `pnpm install --ignore-scripts` and **reproduces identically on pristine `main` with this PR's changes reverted**. Unrelated to this change.
- `tsgo --noEmit` on `scripts` — clean, exit 0.

**End-to-end packaged build.** I ran `node scripts/build-desktop-artifact.ts --platform linux --target AppImage --arch x64` on Debian forky (FUSE 3 only, no `libfuse2` installed anywhere on the system). It completed cleanly, exit 0, producing a 143 MB `T3-Code-0.0.33-x86_64.AppImage`. Its runtime is `static-pie linked`, BuildID `a87aaf5da1…`, matching the 1.0.3 runtime exactly.

The resulting AppImage mounts on a machine where the current release cannot — same host, same flag, back to back:

```
# built from this PR
$ T3-Code-0.0.33-x86_64.AppImage --appimage-mount
/tmp/.mount_T3-CoddPCnao          <- mounts

# current 0.0.34-nightly.20260821.1149
$ T3-Code-0.0.34-nightly...AppImage --appimage-mount
dlopen(): error loading libfuse.so.2
AppImages require FUSE to run.    <- fails
```

`mount` confirms a real FUSE mount, and Electron boots from it (`AppRun` off the mountpoint initialises its user-data dir). My run then exits via the existing single-instance lock because another copy was already running on this box — the behaviour described in #4465 — so I have not verified a full cold UI session end to end, only that the runtime mounts and the app starts from it. That is precisely the part this PR changes.

**Still worth confirming in CI:** the `electron-updater` `AppImageUpdater` in-place update path, since the runtime is what gets swapped.

## Note on the toolset version

`1.0.2`/`1.0.3` are labelled "Betas" in electron-builder's typings (`1.0.3` notes it resolves electron-builder#9598). If you'd rather not ship a beta toolset, the alternative is waiting for v27 — but that leaves the AppImage unlaunchable on modern distros until then.

For anyone hitting this before a release: `APPIMAGE_EXTRACT_AND_RUN=1` works around it without root, since it skips mounting entirely.

> [!NOTE]
> ### Set `toolsets.appimage` to `"1.0.3"` in Linux build config
> Sets the AppImage runtime toolset to a static version `"1.0.3"` in the Linux branch of `createBuildConfig` to fix the libfuse2 launch failure. macOS and Windows build configs are unchanged. Adds a test case asserting the Linux config includes the toolset and that macOS/Windows omit it.
>
> Risk: only the Linux branch of [build-desktop-artifact.ts](https://github.com/pingdotgg/t3code/pull/7765/files#diff-384c39593c6da8887b4f03c00f763af0e3b689525d3555a9fe03fd05c808471d) is affected; verify the `"1.0.3"` toolset version matches the expected static runtime.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8660752.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Changes the Linux AppImage runtime bundled in releases; macOS/Windows packaging is untouched, but AppImage launch and in-place update behavior should be validated on Linux CI.
> 
> **Overview**
> Linux AppImage builds now set **`toolsets.appimage` to `"1.0.3"`** in `createBuildConfig`, so electron-builder bundles the static AppImage runtime instead of the default legacy toolset that depends on **libfuse.so.2**.
> 
> That addresses launch failures on distros that only ship FUSE 3 (e.g. Ubuntu 24.04+, Debian trixie/forky), where users see `dlopen(): error loading libfuse.so.2` before Electron starts. **macOS and Windows** build configs are unchanged—`toolsets` is only applied in the `platform === "linux"` branch.
> 
> Tests assert the Linux toolset pin and that mac/win configs do not define `toolsets`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 866075294f88327a3ac81f4632ae041357377511. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Linux desktop package compatibility on distributions that provide FUSE 3 by using a pinned AppImage runtime.
  - Linux packaging now uses a consistent, bundled AppImage runtime to reduce reliance on system libraries.
  - macOS and Windows packaging behavior remains unchanged.


alexsch01:
@t3dotgg can the next release include this change?

Bombatomica64:
Pr is fixed if anyone to check again

shivamhwp:

Please rebase the packaging test onto current main and keep the newer platform assertions. The production pin applies, but the test hunk conflicts.

The toolset changes the launcher too. Electron Builder 26.15.6 removes the legacy unconditional `--no-sandbox` desktop argument and uses AppRun's `unshare -Ur true` probe to decide whether to add it. Include that behavior in the scope and verify the Ubuntu sandbox case before claiming the combined launch issue is resolved.

Please change "needs no FUSE at all" to "does not require the system libfuse2 library." The static runtime still uses FUSE for mounting. #1482 and #4465 contain additional sandbox/startup concerns beyond the missing library.

Bombatomica64:
@shivamhwp addressed in 104367093 and rebased onto current `main`.

I rechecked the original failure against the current versions:

- Current `main` is already on Electron 44.1.0, but still on electron-builder 26.15.6.
- The latest `v0.0.41-nightly.20260911.1533` AppImage is dynamically linked, contains `libfuse.so.2`, and still reproduces the same `dlopen()` failure on Debian forky with no libfuse2.
- A fresh current-main build with the `toolsets.appimage: "1.0.3"` pin produced a static-PIE runtime with no `libfuse.so.2` reference and remained running through a 15-second Electron 44.1.0 smoke test without a FUSE or sandbox startup error.
- I checked the generated AppRun sandbox branches: when `unshare -Ur true` succeeds it does not add `--no-sandbox`; when the probe fails (including an Ubuntu 24.04.4 container), it adds exactly one `--no-sandbox`.
- electron-builder 26.16.1 retains the legacy default; the default only changes in the unreleased v27 alpha.

I also changed the wording to “does not require the system libfuse2 library” and retained that the static runtime still uses FUSE for mounting. The rebased focused suite passes 70/70, with the Linux toolset assertion and the current macOS/Windows assertions intact.

— Investigated and prepared with OpenAI Codex (`gpt-5.6-sol`).

Bombatomica64:
@coderabbitai I ran the packaged AppImage updater matrix requested in your review against current head 24eb1650f615d35c461e4aafac8a41cc3ddfc967.

- Built 9.9.0 with the legacy runtime and 9.9.1/9.9.2 with toolsets.appimage=1.0.3. file identifies the control as a dynamically linked ELF and both PR artifacts as static PIE.
- Legacy -> static: launched 9.9.0 in Ubuntu 24.04 with libfuse2/fusermount present, used the real electron-updater check/download/quitAndInstall path, and exercised the embedded blockmap through multipart HTTP ranges. The updater reported 6,373 changed blocks with no full-download fallback. The installed file was replaced byte-for-byte with 9.9.1, and a second run ID reached app ready after relaunch.
- Static -> static: launched 9.9.1 on the host, where ldconfig exposes no libfuse.so.2, then updated to 9.9.2 through the same real path. Differential download reported 20 changed blocks and downloaded 427.13 KB of 136,963.31 KB, again with no fallback. SHA-256 confirmed exact replacement, and the relaunched process recorded a new run ID, app ready, backend ready, and main window created.

Result: PASS for both compatibility directions, including blockmap differential download, replacement, and relaunch.

Blunt take: smoke tests are dumb as a routine regression gate here—they are slow, environment-sensitive, and easy to invalidate with unrelated host integration such as AppImageLauncher. This was useful as one-off release validation, but focused deterministic assertions should carry CI.

— Codex sol

PR #7770, opened 2026-08-21
fix(server): abort OpenCode child sessions on interrupt

### LINKED ISSUE #7769: OpenCode Stop can leave sub-agent provider requests running
## Summary

Stopping an OpenCode-backed T3 turn can leave a spawned sub-agent's model request running after the UI reports the turn stopped. With a single-slot local llama.cpp server, the orphan generation saturates the GPU and every later request queues behind it.

## Observed behavior

Environment:

- T3 Code nightly on Linux
- OpenCode 1.18.19
- OpenAI-compatible local llama.cpp provider
- llama.cpp model configured with `--parallel 1`

During the incident:

1. The primary OpenCode session spawned a `general` sub-agent.
2. Stop was requested in T3.
3. OpenCode's log contained no `cancel session.id=...` or `Aborted` record for the running sub-agent.
4. The OpenCode-to-llama.cpp TCP connection remained established.
5. llama.cpp continued the same task for more than 50 minutes, generating at about 25 tok/s toward a 32k-token limit.
6. Later requests appeared hung because the only llama.cpp slot was still occupied.

Restarting llama.cpp caused OpenCode to retry the still-live request, confirming that cancellation had not reached the provider request.

## Cause / hardening opportunity

`OpenCodeAdapter.interruptTurn` calls `session.abort` only for the T3-bound parent OpenCode session. OpenCode sub-agents run as child sessions. Recent OpenCode builds intend parent cancellation to propagate to task children, but this behavior is version-dependent and did not occur in the observed 1.18.19 run.

T3 can defensively enumerate `session.children` and explicitly abort them before aborting the parent. This is backward-compatible and avoids leaving expensive provider requests alive when OpenCode's internal propagation fails.

## Expected behavior

Pressing Stop should abort all active OpenCode work owned by the T3 thread, including child/sub-agent sessions, and release the underlying provider request promptly.

NitayRabi:
## Summary

- enumerate OpenCode child sessions when interrupting a turn
- abort child/sub-agent sessions before aborting the parent session
- preserve parent abort behavior when child enumeration is unsupported or fails
- add regression coverage for multiple child sessions

Closes #7769.

## Why

OpenCode sub-agents use child sessions. Relying only on `session.abort(parent)` leaves T3 dependent on the installed OpenCode version propagating cancellation to all children. When that propagation fails, the child provider stream remains alive and can occupy a single-slot local inference server indefinitely.

## Validation

- `vp test apps/server/src/provider/Layers/OpenCodeAdapter.test.ts` (33 passed)
- `vp fmt --check apps/server/src/provider/Layers/OpenCodeAdapter.ts apps/server/src/provider/Layers/OpenCodeAdapter.test.ts`
- `git diff --check`

> [!NOTE]
> ### Abort child OpenCode sessions on `interruptTurn`
> - Adds `abortOpenCodeSessionTree` to abort a parent session and then enumerate and abort its children concurrently via the SDK in [OpenCodeAdapter.ts](https://github.com/pingdotgg/t3code/pull/7770/files#diff-a4267388b15e5043ae1899beded972cbee9e52533ca3d9d65d438a91a9f7fbe7)
> - `interruptTurn` now calls this tree-abort utility instead of aborting only the parent
> - Test double gains `onAbort` and `sessionChildrenById` hooks to simulate children created during parent abort
> - Risk: per-child abort errors are silently ignored; if the SDK children query fails, child sessions may survive the interrupt
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a677515.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Changes turn-interrupt cancellation for OpenCode, which can leave local inference slots occupied if child aborts are missed. Child enumeration errors are swallowed, so a failed `session.children` call still only aborts the parent.
> 
> **Overview**
> User interrupt now cancels the OpenCode **parent session first**, then enumerates remaining child/sub-agent sessions and aborts those too. That covers older OpenCode builds that do not propagate parent abort, and still catches children spawned during the parent abort.
> 
> `interruptTurn` uses a new `abortOpenCodeSessionTree` helper. Child listing failures become an empty list (parent abort still succeeds). Individual child aborts are ignored so one failure does not block the rest.
> 
> Adds a regression test that asserts parent abort is first and that children present at abort time—including one created during parent abort—are aborted afterwards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a677515ba5fd2aac85afbf69e4fe3dbe5c7b13b1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


juliusmarminge:
Closing because #9005 already added OpenCode child-session cancellation, including children created during the parent abort and nested descendants. Thanks for working on this!

PR #7799, opened 2026-08-21
fix(web): stop tool-name prefix from duplicating command detail

### LINKED ISSUE #7711: [Bug]: Bash command shown twice in tool-call activity — dedup check doesn't strip the "Bash: " prefix baked into detail
### Before submitting

- [x] I searched existing issues and did not find a duplicate. Closest is #7307, but that one is Desktop+OpenCode+WSL and describes the *output* rendering twice with the command missing. This is a different, more specific symptom: the *command itself* renders twice (once plain, once prefixed `Bash: `), output isn't affected.

### Area

apps/web

### Steps to reproduce

1. Run any Bash tool call in a Claude-provider thread (desktop or web).
2. Expand the tool-call activity in the chat timeline.

### Expected behavior

The command is shown once.

### Actual behavior

The command is shown twice: once as the plain command preview, and once again below it prefixed with `Bash: `.

### Root cause (traced against the live event log + source)

The canonical `item.completed` activity carries the same command in two places on one payload:
- `payload.data.input.command` — bare command, e.g. `cd ... && grep ...`
- `payload.detail` — the same command with a tool-name prefix baked in, e.g. `Bash: cd ... && grep ...`

`extractToolDetail` (`apps/web/src/session-logic.ts`) is designed to suppress `detail` when it's redundant with the already-rendered command preview — it normalizes both strings via `normalizePreviewForComparison` → `normalizeCompactToolLabel` and only returns `detail` if the normalized strings differ.

`normalizeCompactToolLabel` only strips a trailing `complete`/`completed` suffix:

```ts
function normalizeCompactToolLabel(value: string): string {
  return value.replace(/\s+(?:complete|completed)\s*$/i, "").trim();
}
```

It never strips the leading `"<ToolName>: "` prefix that's already baked into `payload.detail`. So `normalizedCommand` (`"cd ... && grep ..."`) and `normalizedDetail` (`"bash: cd ... && grep ..."`) never match, the dedup check in `extractToolDetail` treats them as different content, and the redundant `detail` line renders anyway — directly under the command that was already shown.

### Example payload from a live session (Claude provider)

```json
{
  "type": "item.completed",
  "payload": {
    "itemType": "command_execution",
    "title": "Command run",
    "detail": "Bash: cd /Users/.../eksaq-platform/knowledge-base && grep -n -E \"F0[0-9][0-9]\" TRACKER.md | grep -i -E \"hpc|F0\" | sed -n '1,120p'",
    "data": {
      "toolName": "Bash",
      "input": {
        "command": "cd /Users/.../eksaq-platform/knowledge-base && grep -n -E \"F0[0-9][0-9]\" TRACKER.md | grep -i -E \"hpc|F0\" | sed -n '1,120p'"
      }
    }
  }
}
```

Image

### Impact

Cosmetic, but consistently reproducible on every Bash tool call — clutters the timeline for any thread with meaningful shell usage.

### Version or commit

Nightly, macOS desktop, Claude provider.

### Suggested fix

Strip the `"<ToolName>: "` prefix (or any leading `toolName + ": "`) from `detail` before the redundancy comparison in `extractToolDetail`/`normalizePreviewForComparison`, so it's correctly recognized as duplicate content when it only differs from the command preview by that prefix.

Exotic209093:
## Summary

Fixes #7711. `item.completed` payloads for Bash tool calls carry the same command in two places: `payload.data.input.command` (bare) and `payload.detail` (the same command with a `Bash: ` prefix baked in). `extractToolDetail` in `apps/web/src/session-logic.ts` is supposed to suppress `detail` when it's redundant with the already-rendered command preview, but it only stripped a trailing `complete`/`completed` suffix before comparing — never the leading `<toolName>: ` prefix. So the normalized command and normalized detail never matched, and the redundant detail line rendered a second time under every Bash tool call.

## Fix

Strip a leading `"<toolName>: "` prefix (read from `payload.data.toolName`) from `detail` before the redundancy comparison in `extractToolDetail`.

## Test plan

- [x] Added a regression test in `session-logic.command-output.test.ts` reproducing the exact payload shape from the issue (command in `data.command`, detail prefixed with `Bash: `); confirmed it fails before the fix and passes after.
- [x] `vp test run apps/web/src/session-logic.command-output.test.ts apps/web/src/session-logic.test.ts` — 79 passed.
- [x] `vp run --filter @t3tools/web typecheck` — clean.

---

> [!NOTE]
> **Low Risk**
> Small display-only change in work-log detail comparison, with a targeted regression test and no auth or data-handling impact.
> 
> **Overview**
> Stops work-log command rows from repeating the same command under Bash (and similar) tool calls.
> 
> `extractToolDetail` already hid `detail` when it matched the command preview, but payloads often prefix detail with `"<toolName>: "`. The comparison now strips that prefix (from `data.toolName`) so redundant detail is dropped. Adds a regression test for the Bash-prefixed payload shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a8af60d9ceba481049f52d38bee2042349966f94. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Strip tool-name prefix from command detail in web output
> - Server's `projectActivityPayload` now propagates `data.toolName` into the projected payload for non-`mcp_tool_call` items, so the web client can detect redundant headings.
> - Web client adds `stripToolNamePrefix` helper in [session-logic.ts](https://github.com/pingdotgg/t3code/pull/7799/files#diff-9f1f9f19555f14c3c47d4ce52e1b7dcec7e968e54f400820085376797a7223d6) which removes a case-insensitive `<toolName>: ` prefix from a string; `extractToolDetail` applies it before comparing detail to the command preview.
> - Detail that only differs from the command by a matching tool-name prefix is now treated as a duplicate and omitted; unrelated labeled details such as `warning: true` are preserved.
> - Risk: `stripToolNamePrefix` only strips when the prefix exactly matches `payload.data.toolName`; mislabeled or missing `toolName` values will leave the prefix visible rather than dropping the detail.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6b5e461.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/ActivityPayloadProjection.test.ts
+++ b/apps/server/src/orchestration/ActivityPayloadProjection.test.ts
@@ -126,7 +126,7 @@ describe("projectActivityPayload", () => {
 
     expect(claude.payload).toMatchObject({
       toolCallId: "claude-call-1",
-      data: { command: "vp test run" },
+      data: { toolName: "Bash", command: "vp test run" },
     });
     expect(openCode.payload).toMatchObject({
       toolCallId: "opencode-call-1",
@@ -134,6 +134,23 @@ describe("projectActivityPayload", () => {
     });
     expect(JSON.stringify(claude.payload).length).toBeLessThan(200);
     expect(JSON.stringify(openCode.payload).length).toBeLessThan(200);
+  });
+
+  it("preserves toolName for command activities so the web client can recognize a redundant heading in detail", () => {
+    const projected = projectActivityPayload(
+      activity({
+        itemType: "command_execution",
+        detail: 'Bash: cd project && grep -n "TODO" file.md',
+        data: {
+          toolName: "Bash",
+          command: 'cd project && grep -n "TODO" file.md',
+        },
+      }),
+    );
+
+    expect(projected.payload).toMatchObject({
+      data: { toolName: "Bash" },
+    });
   });
 
   it("slims Codex-shaped mcp_tool_call items to rendered fields plus a result summary", () => {
--- a/apps/server/src/orchestration/ActivityPayloadProjection.ts
+++ b/apps/server/src/orchestration/ActivityPayloadProjection.ts
@@ -375,6 +375,12 @@ export function projectActivityPayload(
   if ("kind" in data) {
     projectedData.kind = data.kind;
   }
+  if ("toolName" in data) {
+    // The web client needs this to recognize a "<toolName>: " heading some
+    // providers bake into payload.detail as redundant with the command
+    // preview it renders separately (see extractToolDetail).
+    projectedData.toolName = data.toolName;
+  }
 
   const rawOutput = projectRawOutput(data.rawOutput) ?? projectAcpContent(data.content);
   if (rawOutput) {
--- a/apps/web/src/session-logic.command-output.test.ts
+++ b/apps/web/src/session-logic.command-output.test.ts
@@ -82,4 +82,22 @@ describe("deriveWorkLogEntries command output", () => {
     expect(entry?.command).toBe("true");
     expect(entry?.detail).toBeUndefined();
   });
+
+  it("drops duplicated command detail when it only differs by a toolName prefix", () => {
+    const [entry] = deriveWorkLogEntries([
+      makeCommandActivity("bash-prefixed-command", {
+        itemType: "command_execution",
+        title: "Command run",
+        detail: 'Bash: cd project && grep -n "TODO" file.md',
+        data: {
+          toolName: "Bash",
+          kind: "execute",
+          command: 'cd project && grep -n "TODO" file.md',
+        },
+      }),
+    ]);
+
+    expect(entry?.command).toBe('cd project && grep -n "TODO" file.md');
+    expect(entry?.detail).toBeUndefined();
+  });
 });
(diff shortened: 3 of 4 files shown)


Exotic209093:
Good catch — `toolName` gets dropped by `ActivityPayloadProjection.ts` for command activities, so the field-based check was dead on real traffic. Pushed a fix that matches the prefix structurally against detail/command instead (no dependency on a field that never survives projection), plus a negative test for coincidental suffix matches.

Exotic209093:
Updated again: Macroscope's follow-up finding was right too — the structural suffix match could drop a legitimate detail like `warning: true` when the command happened to be `true`. Went with the fix both bots suggested originally: `toolName` now survives `ActivityPayloadProjection.ts` for command activities (mirrors what the MCP branch already does), so the client goes back to comparing against that exact field instead of a heuristic. Added a regression test for the false positive and a projection-side test proving `toolName` now reaches the client.

t3dotgg:
> [!NOTE]
>

Closing as superseded by merged [#9120](https://github.com/pingdotgg/t3code/pull/9120). Main preserves tool names and removes synthetic command echoes in shared work-log code, so web and mobile use the same rule. Genuine output remains visible even when it matches the command.

Issue #4924, opened 2026-07-30
[Bug]: macOS agent terminals fail with 'posix_spawnp failed' - node-pty spawn-helper ships without exec bit


hashmil:
### What happened

After `npm i -g t3` on macOS (arm64), the server starts fine but **every agent terminal fails to spawn**:

```
node_modules/node-pty/lib/unixTerminal.js:92
        var term = pty.fork(file, args, parsedEnv, cwd, ...);
                       ^
Error: posix_spawnp failed.
    at new UnixTerminal (node-pty/lib/unixTerminal.js:92:24)
    at Object.spawn (node-pty/lib/index.js:30:12)
```

### Root cause

Not a missing prebuild — `prebuilds/darwin-arm64/pty.node` is present and the native module loads fine. The problem is the helper binary next to it:

```
$ find $(npm root -g)/t3/node_modules/node-pty/prebuilds -name spawn-helper -exec stat -f "%Sp %N" {} \;
-rw-r--r--  .../prebuilds/darwin-arm64/spawn-helper
-rw-r--r--  .../prebuilds/darwin-x64/spawn-helper
```

`spawn-helper` has **no execute bit**, so `posix_spawnp` on it fails.

This originates in the published `node-pty` tarball, not in npm's extraction:

```
$ npm pack node-pty@1.1.0 && tar tvf node-pty-1.1.0.tgz | grep spawn-helper
-rw-r--r--  0 0 0  50480  package/prebuilds/darwin-arm64/spawn-helper
-rw-r--r--  0 0 0   9248  package/prebuilds/darwin-x64/spawn-helper
```

And `node-pty`'s `scripts/post-install.js` only operates on `build/Release`, never on `prebuilds/`, so nothing ever restores the bit.

### Important: `--allow-scripts` does not fix this

There is a known, separate node-pty issue on Linux where npm's script gating blocks the native build (`NodePtyModuleLoadError`), worked around with `npm i -g --allow-scripts=node-pty,msgpackr-extract t3@nightly`. **That workaround does not help here.** Verified into a throwaway prefix:

```
$ npm install -g t3@0.0.31 --prefix /tmp/x --allow-scripts=node-pty,msgpackr-extract
added 197 packages in 4s
$ find /tmp/x -name spawn-helper -exec stat -f "%Sp %N" {} \;
-rw-r--r--  .../darwin-x64/spawn-helper
-rw-r--r--  .../darwin-arm64/spawn-helper
```

[...]

PR #7801, opened 2026-08-21
fix(server): self-heal a non-executable bundled resource-monitor binary

### LINKED ISSUE #7736: [Bug]: Bundled t3-resource-monitor ships mode 0644, so native process telemetry never starts on Linux/macOS
### What happened

On Linux, `t3 serve` (headless systemd user service, `t3@0.0.33`) logs:

```
Resource monitor binary at '.../node_modules/t3/dist/resource-monitor/linux-x64/t3-resource-monitor' is not executable.
Failed to read native resource telemetry history
```

The native process monitor never starts. The server itself is fine.

The file on disk is a valid ELF PIE (`t3-resource-monitor` 0.1.0, `sysinfo`), but npm extracted it as `-rw-rw-r--` (`0644` / `0664`). T3's resolver then fails closed:

```js
// ResourceMonitorBinary.resolve, non-Windows
if ((stat.mode & 73) === 0) // 0o111
  fail ResourceMonitorBinaryNotExecutable
```

It does not `chmod` the sidecar.

### Root cause

This is in the **published `t3` tarball**, not local umask or a `noexec` mount. From `registry.npmjs.org`:

```
t3@0.0.33
  0644  package/dist/resource-monitor/darwin-arm64/t3-resource-monitor
  0644  package/dist/resource-monitor/darwin-x64/t3-resource-monitor
  0644  package/dist/resource-monitor/linux-x64/t3-resource-monitor
  0644  package/dist/resource-monitor/win32-x64/t3-resource-monitor.exe
  0755  package/dist/bin.mjs
```

Same modes in `t3@0.0.34-nightly.20260820.1146` (checked today). `bin.mjs` is executable because it is the npm `bin` entry; the Rust sidecars are copied into `dist/resource-monitor/<platform>-<arch>/` without `+x` before `npm pack`.

Windows skips the mode check, so this specifically breaks Unix `t3 serve` / CLI installs. Desktop artifact builds that stage `resources/resource-monitor/t3-resource-monitor` separately may not hit it.

Related class of bug: #4924 (`node-pty` `spawn-helper` also published `0644`). This one is T3's own binary, not node-pty.

### Workaround

```bash
find "$T3CODE_HOME/runtime/versions" -type f -name t3-resource-monitor -exec chmod +x {} +
# then restart t3 serve
```

Must be repeated after every runtime install/update, because a fresh extraction restores `0644`.

### Suggested fix

1. `chmod +x` the Unix sidecars when CLI release jobs copy them into `apps/server/dist/resource-monitor/<platform>-<arch>/` (and assert mode in CI on the packed tarball).
2. Defensive: if `ResourceMonitorBinary.resolve` finds a non-executable Unix sidecar that T3 owns, chmod it (or spawn via an explicit interpreter) instead of staying `unavailable`. Same idea as the startup chmod suggested on #4924.

### Environment

- Linux x86_64 (glibc), Node v22.23.2
- `t3@0.0.33` via `~/.t3/runtime/versions/0.0.33` (`t3 serve`)
- Also verified the tarball modes for `0.0.34-nightly.20260820.1146`

Exotic209093:
## Summary

Fixes #7736. `t3 serve` on Linux/macOS logs \`Resource monitor binary ... is not executable\` and native process telemetry never starts. The published npm tarball ships \`dist/resource-monitor/<platform>-<arch>/t3-resource-monitor\` as \`0644\`.

The release workflow already runs \`chmod +x\` on these sidecars before publish (added 2026-07-29), so the CI-side fix in the issue's first suggestion is already in place — yet the published tarball still shows \`0644\`. That's because npm only preserves the executable bit for files listed in \`package.json\`'s \`bin\` field; any other file, including ours, gets re-packed without it regardless of its on-disk mode at pack time (the same class of issue as \`node-pty\`'s \`spawn-helper\`, referenced in the issue as #4924).

## Fix

Since CI-side chmod can't survive npm's packing, this implements the issue's second suggested fix: \`ResourceMonitorBinary.resolve\` now repairs the executable bit at resolve time for binaries T3 bundles itself, instead of failing closed. A user-supplied override path (\`T3CODE_RESOURCE_MONITOR_PATH\` / config) is left untouched and still fails closed on a non-executable file — auto-chmod'ing an arbitrary path we didn't build isn't safe.

## Test plan

- [x] Added `apps/server/src/resourceTelemetry/ResourceMonitorBinary.test.ts` coverage that writes a bundled sidecar at 0644 and confirms `resolve` repairs it and returns the path.
- [x] `vp test run apps/server/src/resourceTelemetry` — 50 passed (run under WSL/Linux; two pre-existing POSIX-mode tests in this suite don't reflect real chmod semantics on native Windows and were unaffected by this change).
- [x] `vp run --filter t3 typecheck` — clean.

---

> [!NOTE]
> **Medium Risk**
> Adds a runtime chmod of bundled binaries on disk. Scope is limited to owned sidecar paths; overrides remain fail-closed, but this still mutates file modes at resolve time.
> 
> **Overview**
> Fixes resource-monitor telemetry on Linux/macOS after npm packs bundled sidecars as `0644`. `ResourceMonitorBinary.resolve` now `chmod`s **owned** bundled binaries that lack the execute bit, then returns the path.
> 
> User-supplied override paths (`T3CODE_RESOURCE_MONITOR_PATH` / config) stay fail-closed so we never chmod an arbitrary file. Adds a test that writes a bundled sidecar at `0644` and asserts it is repaired.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 95949ed3905798f9d48c8baa30ea1e43d2bab70f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `resourceMonitorBinary.make` to self-heal non-executable bundled binary
> Some npm package layouts strip the execute bit from the bundled resource-monitor binary, causing resolution to fail on Linux. When a bundled (owned) candidate exists but lacks any execute bit (`mode & 0o111 == 0`), the factory now chmods the file to add execute bits and succeeds, instead of returning `ResourceMonitorBinaryNotExecutable`.
> - Candidate handling changed from string paths to objects `{ path, owned }`, marking override candidates as `owned: false` and bundled candidates as `owned: true`
> - Override candidates that lack the execute bit still fail with `ResourceMonitorBinaryNotExecutable`; only bundled binaries are auto-repaired
> - Added a test in [ResourceMonitorBinary.test.ts](https://github.com/pingdotgg/t3code/pull/7801/files#diff-70d99822859da561212ab248e25506b980258f8af7fca2b49b623c93dd9ef3aa) that writes a 0644 file and asserts `resolve` repairs the mode and returns the path
> - Behavioral Change: previously a non-executable bundled binary returned an error; now it is silently chmod'd and succeeds. Only affects non-Windows platforms
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 95949ed.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Bundled Linux x64 resource-monitor binaries are now automatically made executable when required.
  - User-provided binary overrides remain unchanged if they lack executable permissions.
  - Errors for unavailable or non-executable binaries now include the candidate paths checked.


shivamhwp:

Please remove the claim in the resolver comment and PR description that npm strips execute bits from every file outside package.json `bin`. npm can preserve an executable sidecar, so that does not establish why the published T3 artifact is 0644. The runtime repair remains useful for affected installs; describe it as recovery from the observed package mode and leave the release-stage cause open until it is identified.

The added test forces a Linux platform but checks real filesystem mode bits. Guard it on native Windows, as the neighboring executable-override tests already do.

juliusmarminge:
Closing: the root cause is gone in packaging. Since #11607 the CLI ships per-platform packages built from the archive, which carries file modes; the published `@t3code/t3-linux-x64@0.0.42` binary is `-rwxr-xr-x`. The comment this adds ("npm only preserves the exec bit for bin entries") is not accurate and would ship. Also the test writes fixture files into `src/resourceTelemetry/` and lacks the Windows skip its neighbours have. If a runtime self-heal turns out to be needed later, `NodePtyAdapter.ts` has a one-line `chmod(0o755).pipe(orElseSucceed)` precedent.

PR #7753, opened 2026-08-21
fix(server): price Daybreak usage aliases

obinnanwachukwu1:
## Problem

Codex records Daybreak models in rollout transcripts as `gpt-daybreak-blue-latest` and `gpt-daybreak-red-latest`, while LiteLLM publishes their pricing under `daybreak-blue-latest` and `daybreak-red-latest`.

Because Usage performed an exact normalized rate lookup, Daybreak tokens were included in token totals but marked unpriced and contributed `$0.00` to the API-equivalent cost.

Closes #7752

## Fix

Add a small alias fallback at the LiteLLM pricing boundary.

Exact rate-table matches still take precedence, so a future native LiteLLM entry for a Codex slug automatically overrides the fallback. Codex model selection and stored transcript identifiers remain unchanged.

Because Usage calculates pricing when it reads transcripts, this prices existing Daybreak history without a migration.

## Verification

- `vp test run apps/server/src/usage/usagePricing.test.ts apps/server/src/usage/usageAggregation.test.ts`
- `vp lint apps/server/src/usage/usagePricing.ts apps/server/src/usage/usagePricing.test.ts`
- `vp run --filter t3 typecheck`
- Verified in the Usage page with the same 24-hour Daybreak history before and after the change

Built with GPT-5.6 Sol using Codex in T3 Code.

---

> [!NOTE]
> **Low Risk**
> Narrow pricing-lookup change with tests; it only remaps two model names at cost calculation time and does not touch auth, billing charges, or stored transcripts.
> 
> **Overview**
> Fixes Daybreak usage showing as unpriced (`$0.00`) because Codex transcripts use `gpt-daybreak-*-latest` while LiteLLM publishes rates as `daybreak-*-latest`.
> 
> `lookupRate` now falls back to a small alias map after an exact table match. Exact LiteLLM keys still win, and stored model names are unchanged. Existing Daybreak history is priced on read without a migration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0a9f50ac8383166d97547f4a2d1f95cd4e3a10d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add Daybreak usage aliases to `lookupRate` in `usagePricing`
[...]

PR #7802, opened 2026-08-21
fix(server): price Daybreak models under their LiteLLM alias

### LINKED ISSUE #7752: [Bug]: Usage leaves Daybreak models unpriced because Codex and LiteLLM use different aliases
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Run a Codex session using `gpt-daybreak-blue-latest` or `gpt-daybreak-red-latest`.
2. Confirm the Codex rollout records that model name in `turn_context`.
3. Open T3 Code's Usage page and refresh the usage summary.
4. Find the Daybreak model in the model breakdown.

### Expected behavior

Daybreak token usage should receive the corresponding API-equivalent price from LiteLLM:

- `gpt-daybreak-blue-latest` should use `daybreak-blue-latest`
- `gpt-daybreak-red-latest` should use `daybreak-red-latest`

The Usage page should include those records in its estimated cost and cache-savings totals.

### Actual behavior

The token counts appear, but Daybreak records are classified as unpriced and contribute `$0.00` to the estimated cost.

Codex records the operational model names with a `gpt-` prefix:

```text
gpt-daybreak-blue-latest
gpt-daybreak-red-latest
```

LiteLLM publishes their rates under different keys:

```text
daybreak-blue-latest
daybreak-red-latest
```

T3's Usage pricing performs an exact normalized lookup. It removes provider-path prefixes and normalizes casing, but does not reconcile these two upstream aliases.

This is separate from the model-picker classification fixed by #7659.

### Impact

Cost totals are consistently understated for users with Daybreak activity. Token reporting still works, but potentially substantial usage appears as free.

### Version or commit

main @ 45a2c4b2a

### Environment

Codex app-server with Daybreak access; reproduced with `gpt-daybreak-blue-latest` on macOS.

### Logs or stack traces

```shell
Codex transcript model: gpt-daybreak-blue-latest
LiteLLM rate key:       daybreak-blue-latest
Usage result:           unpriced / $0.00
```

### Workaround

No in-app workaround. Users can see Daybreak token counts but must calculate their API-equivalent cost separately.

Exotic209093:
## Summary

Fixes #7752. Codex records Daybreak model usage as `gpt-daybreak-blue-latest` and `gpt-daybreak-red-latest`. LiteLLM's `model_prices_and_context_window.json` publishes rates for these same models under different keys, without the `gpt-` prefix: `daybreak-blue-latest` and `daybreak-red-latest`. Because `lookupRate` in `apps/server/src/usage/usagePricing.ts` looks the model up by its exact normalized name, the mismatch means both models miss the rate table entirely and get reported as unpriced ($0.00), even though token counts are recorded correctly.

## Fix

Added a narrow, explicit two-entry alias map (`MODEL_NAME_ALIASES`) consulted in `lookupRate` after normalization and before the table lookup. This is deliberately not a generic "strip a leading `gpt-`" transform: LiteLLM prices real GPT models (`gpt-5`, `gpt-4`, ...) *with* the `gpt-` prefix intact, so a generic strip would send those lookups at the wrong (and likely nonexistent) key. `normalizeModelName` and `parseRateTable` are unchanged — the alias only applies at lookup time, not when building the table from LiteLLM's document.

## Test plan

- [x] Added `apps/server/src/usage/usagePricing.test.ts` covering: `daybreak-blue-latest`/`daybreak-red-latest` priced correctly via their Codex `gpt-` prefixed names; the alias is case-insensitive; a real `gpt-` model (`gpt-5`) still resolves under its own key and is unaffected by the alias; a real `gpt-` model absent from the table (`gpt-4`) stays unpriced rather than matching a stripped-prefix key. Confirmed the tests fail before the fix and pass after.
- [x] `vp test run apps/server/src/usage/usagePricing.test.ts` — 5 passed.
- [x] `vp lint` on both changed files — clean.
- [x] `vp run --filter t3 typecheck` — clean (only pre-existing, unrelated suggestions in other files).

---

> [!NOTE]
> **Low Risk**
> Narrow lookup-only alias for two model names; no billing write path, auth, or table-parse changes.
> 
> **Overview**
> Fixes Daybreak usage showing as **$0.00** because Codex records `gpt-daybreak-blue-latest` / `gpt-daybreak-red-latest` while LiteLLM prices those models without the `gpt-` prefix.
> 
> `lookupRate` now remaps those two names via an explicit `MODEL_NAME_ALIASES` map after normalization. Real GPT models keep their prefixed keys; there is no generic prefix strip. Tests cover both aliases, case-insensitivity, and that `gpt-5` / missing `gpt-4` are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1223d7f070c9e175751238fdc4cde8e2a4ab7caa. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix pricing for Daybreak models by aliasing `gpt-daybreak-*` to unprefixed rate keys
> Adds a `MODEL_NAME_ALIASES` map in [usagePricing.ts](https://github.com/pingdotgg/t3code/pull/7802/files#diff-fe836a76687342f066ff2679d4edcb42b28156a016ba013085092c27fc826c66) that maps `gpt-daybreak-blue-latest` and `gpt-daybreak-red-latest` to their unprefixed rate-table keys. `lookupRate` now checks this map after normalization and unpriceable checks, using the aliased key when present and falling back to the normalized name otherwise.
> - Adds vitest coverage for aliasing, case-insensitivity, genuine `gpt-`-prefixed models, and non-stripping of absent entries like `gpt-4`.
> - Risk: only the two explicit `gpt-daybreak-*` names are remapped; other `gpt-`-prefixed model lookups are unchanged.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1223d7f.</sup>
> 
>

### DIFF
--- a/apps/server/src/usage/usagePricing.test.ts
+++ b/apps/server/src/usage/usagePricing.test.ts
@@ -1,0 +1,41 @@
+import { describe, expect, it } from "@effect/vitest";
+
+import { lookupRate, parseRateTable } from "./usagePricing.ts";
+
+describe("lookupRate aliasing", () => {
+  const table = parseRateTable({
+    "gpt-5": { input_cost_per_token: 0.00001, output_cost_per_token: 0.00003 },
+    "daybreak-blue-latest": { input_cost_per_token: 0.000002, output_cost_per_token: 0.000006 },
+    "daybreak-red-latest": { input_cost_per_token: 0.000003, output_cost_per_token: 0.000009 },
+  });
+
+  it("resolves Codex's gpt-prefixed Daybreak blue name to LiteLLM's unprefixed rate", () => {
+    const rate = lookupRate(table, "gpt-daybreak-blue-latest");
+
+    expect(rate).toEqual(table.get("daybreak-blue-latest"));
+  });
+
+  it("resolves Codex's gpt-prefixed Daybreak red name to LiteLLM's unprefixed rate", () => {
+    const rate = lookupRate(table, "gpt-daybreak-red-latest");
+
+    expect(rate).toEqual(table.get("daybreak-red-latest"));
+  });
+
+  it("is case-insensitive, consistent with normalizeModelName's lowercasing", () => {
+    const rate = lookupRate(table, "GPT-Daybreak-Blue-Latest");
+
+    expect(rate).toEqual(table.get("daybreak-blue-latest"));
+  });
+
+  it("leaves a real gpt-prefixed model to look up under its own key, unrewritten", () => {
+    const rate = lookupRate(table, "gpt-5");
+
+    expect(rate).toEqual(table.get("gpt-5"));
+  });
+
+  it("does not invent a rate for a real gpt model absent from the table", () => {
+    // Guards against a generic "strip gpt- prefix" implementation, which would
+    // wrongly send "gpt-4" lookups at a hypothetical "4" table entry.
+    expect(lookupRate(table, "gpt-4")).toBeNull();
+  });
+});
--- a/apps/server/src/usage/usagePricing.ts
+++ b/apps/server/src/usage/usagePricing.ts
@@ -98,10 +98,27 @@ const UNPRICEABLE_MODELS = new Set([
   "fable",
 ]);
 
+/**
+ * Known mismatches between the name a provider records and the key LiteLLM
+ * prices it under.
+ *
+ * Codex reports Daybreak usage as `gpt-daybreak-blue-latest` /
+ * `gpt-daybreak-red-latest`, but LiteLLM prices those same models without the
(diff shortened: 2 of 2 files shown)


t3dotgg:
> [!NOTE]

Closing this duplicate in favor of [#7753](https://github.com/pingdotgg/t3code/pull/7753). That implementation tries the exact model name before the Daybreak alias, so a later exact price cannot be overridden by an alias. The replacement is still open, not shipped. Keep these two model-name regression cases in the retained implementation.

PR #7814, opened 2026-08-21
fix(server): recover skill frontmatter Claude Code itself accepts

### LINKED ISSUE #7757: [Bug]: Claude skill discovery drops skills with frontmatter Claude Code itself accepts (strict YAML vs CLI leniency), silently
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Install a skill whose `description` is a plain scalar containing a colon followed by a space. Real-world example: LambdaTest's kane-cli installer (`kane-cli install skill`) writes `~/.claude/skills/kane-cli/SKILL.md` with:

   ```yaml
   ---
   name: kane-cli
   description: Browser automation + AI test authoring via kane-cli: run browser objectives, ...
   ---
   ```

   `yaml.parse` rejects this (`mapping values are not allowed here`), but Claude Code's own frontmatter parser accepts it — the skill loads and works in every Claude Code session.

2. Start the desktop app with a Claude provider instance and wait for a capability probe cycle.
3. Type `$` in the composer, or read `~/.t3/caches/claudeAgent.json`.

### Expected behavior

Whatever file the Claude Code CLI loads as a skill should be discovered by T3's own scanner too — the CLI is the authority on what it accepts. Failing that, dropping a skill should at least log which file was rejected and why.

### Actual behavior

`parseSkillFrontmatter` in `apps/server/src/provider/Drivers/ClaudeSkills.ts` strict-YAML-parses the frontmatter and skips the entry as `malformed` with no log. The same cache snapshot shows the contradiction: the skill appears in `slashCommands` (enumerated by the Claude Code SDK probe, which accepts the file) while `skills` is `[]` (T3's scanner dropped it). If it is the user's only skill, the `$` picker is empty and nothing explains why.

Related but distinct: #6449 covers discovery scanning the wrong directory (`ServerConfig.cwd`), and #1084 covers malformed-skill diagnostics in threads. This one is the parser-strictness mismatch: a skill that demonstrably works in Claude Code is invisible in T3.

### Impact

### Version or commit

Desktop 0.0.33 (alpha); code path unchanged at main HEAD

### Environment

macOS (Darwin 25.5.0), T3 Code (Alpha) 0.0.33, Claude Code CLI 2.1.237, kane-cli skill 0.0.17 as shipped by @testmuai/kane-cli-skill

### Logs or stack traces

```shell
# ~/.t3/caches/claudeAgent.json (one snapshot, provider ready, authenticated)
"slashCommands": [... "kane-cli" ...]   # SDK probe accepted the file
"skills": []                            # T3's scanner dropped it, no log emitted

# python yaml.safe_load on the same frontmatter:
mapping values are not allowed here, line 2, column 65
```

### Workaround

Quote the description in the skill's frontmatter so it parses as strict YAML. Works, but every consumer of an installer-shipped skill has to know to do it, and nothing tells them.

---
Investigated and filed by Claude Fable 5 (claude-fable-5) on behalf of the repo user; findings verified against the packaged app bundle and source at main.

Exotic209093:
## Summary

Fixes #7757. `parseSkillFrontmatter` in `apps/server/src/provider/Drivers/ClaudeSkills.ts` strict-YAML-parses each `SKILL.md`'s frontmatter and drops the entry entirely on any parse error. Claude Code's own frontmatter parser is more lenient: an unquoted scalar description containing a `word: ` sequence (e.g. `description: ... via kane-cli: run browser objectives, ...`) is valid there, but strict YAML rejects it as an ambiguous nested mapping (`mapping values are not allowed here`). A skill that demonstrably loads and works in Claude Code was invisible in T3's own scanner and the `$` picker — the same cache snapshot showed the contradiction directly: the skill appeared in `slashCommands` (from the CLI's own probe) while `skills` stayed `[]`.

## Fix

When strict YAML parsing fails, fall back to recovering `name`/`description` as flat `key: value` scalars — the only two fields this scanner reads — instead of dropping the skill. The fallback only kicks in per-line for top-level, unindented keys, and explicitly still treats a value as malformed if it looks like it was attempting real YAML structure that broke (starts with `[`, `{`, `|`, `>`, `&`, `*`, or `!` — an unterminated flow collection, block scalar, anchor, alias, or tag). That keeps the existing "genuinely broken YAML" behavior intact (e.g. `name: [unclosed` still gets dropped, matching Claude Code, which wouldn't load that either) while recovering the specific leniency gap the issue reports.

## Test plan

- [x] Added a regression test reproducing the exact `kane-cli` frontmatter from the issue; confirmed it fails before the fix (dropped as malformed) and passes after (recovered with the correct name/description).
- [x] `vp test run apps/server/src/provider/Drivers/ClaudeSkills.test.ts` — 10 passed, including the existing malformed-YAML test (`name: [unclosed`) which still correctly drops the skill.
- [x] `vp run --filter t3 typecheck` — clean (only pre-existing, unrelated suggestions in other files).

---

> [!NOTE]
> **Low Risk**
> Parser fallback is limited to two scalar skill fields and still drops structurally broken YAML. No auth, data, or API surface changes.
> 
> **Overview**
> Skills that Claude Code loads but T3 dropped as malformed YAML now show up in discovery (and the `$` picker).
> 
> When strict YAML parse fails, `parseSkillFrontmatter` falls back to reading top-level `name`/`description` as flat `key: value` scalars. That covers unquoted descriptions with a colon (e.g. `via kane-cli: run ...`). Values that look like broken YAML structure (`[`, `{`, `|`, `>`, `&`, `*`, `!`) still count as malformed and are skipped.
> 
> Adds a regression test for the `kane-cli` frontmatter from the issue.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f97d0d156e68a99c0789d6173ee9a9a45cb97159. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Recover skill frontmatter that `ClaudeCode` accepts but strict YAML rejects
> Adds a lenient fallback to `parseSkillFrontmatter` so that when strict YAML parsing fails, it still recovers `name` and `description` from flat `key: value` scalars. Values starting with YAML structural tokens (`[`, `{`, `|`, etc.) are rejected to avoid misinterpreting real structure as plain scalars, and trailing `# comments` are stripped.
> - Covers cases like unquoted descriptions containing colons and broken `name` fields that cause the whole skill to be skipped.
> - Risk: lenient recovery only triggers on strict-parse failure and returns `{kind: 'malformed'}` for structurally broken values, so previously-rejected skills are the only ones affected.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 98a98ba.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude skill discovery when YAML frontmatter contains recoverable formatting issues.
  * Preserves descriptions containing colons and removes trailing comments from skill names.
  * Correctly interprets supported invocation settings, including affirmative and negative values.
  * Skips skills with unrecoverable or structurally invalid frontmatter, including malformed tool configuration fields, to prevent invalid entries from loading.


Exotic209093:
Good catch, both of you — the per-line recovery was independent per field, so a document with one genuinely broken line (in a recognized field or not) alongside a fine one still surfaced the skill with the broken bit silently dropped, exactly the case this was supposed to exclude. Fixed: any top-level line with broken-looking YAML structure now fails the whole document as malformed before recovering anything, regardless of which field it's in. Added regression tests for both scenarios.

Exotic209093:
Good catch — the fallback was copying everything after the colon verbatim, so a trailing YAML comment or quote escaping wasn't handled. Switched to parsing the isolated value with the real YAML parser: it strips comments and unescapes quotes correctly, and the one case this fallback exists for (an unquoted value with its own embedded ": ") still parses as a one-entry mapping in isolation, not a string, so it correctly falls through to the raw text as before.

shivamhwp:

The structural-value guard rejects valid metadata too. A colon-containing description plus `allowed-tools: [Read, Write]` loads in Claude Code 2.1.267, but this fallback drops the skill solely because the list starts with `[`. Distinguish malformed structure from valid structure alongside the recoverable description.

The fallback also loses current invocation metadata. With `user-invocable: false` beside the same description, Claude omits the skill from its published commands, but the recovered T3 entry has no `userInvocable: false` flag. Preserve `user-invocable` and `disable-model-invocation` through recovery, and keep main's directory-based command identity when rebasing. Prefer repairing the supported scalar-colon case and parsing the complete document so unrelated fields keep their normal validation and semantics.

juliusmarminge (changes requested):
The bug is real and worth fixing, but the fallback needs a different shape.

1. `parseSkillFrontmatterLeniently` is a second parser. The structural-character guard (`^[[{|>&*!]`) fails the whole document when *any* top-level value starts with one of those, so `allowed-tools: [Read, Write]` next to a colon-bearing description makes strict parse fail on the description, then the fallback fails on the valid flow-sequence line, and the skill is still dropped. The "skips the whole skill when an unread field has broken YAML syntax" test enshrines this false negative.
2. The fallback returns only name/description, so a recovered skill loses `disable-model-invocation` / `user-invocable`, which main applies at `ClaudeSkills.ts:91-97`. A skill that Claude Code treats as user-invocable-only becomes model-invocable in T3.
3. `name` is dead: commands are keyed by directory on main.

Suggested replacement, roughly 10 lines and no second parser: on strict-parse failure, for each top-level `key: value` line whose unquoted scalar contains `: `, wrap the value in double quotes (escaping `\\` and `"`) and re-run the real YAML parser on the whole document. Every other field keeps normal validation, genuinely broken YAML still throws, and the leniency is limited to the one documented divergence from Claude Code. Please replace the hand parser and its tests with that, keeping one positive case (colon in description recovers with all fields intact) and one negative case (real syntax error still drops).

juliusmarminge:

PR #7924, opened 2026-08-22
fix(server): let multiple servers on one host show distinct environment labels

robnewton:
## What Changed
Adds a `T3CODE_ENVIRONMENT_LABEL` env var and a matching `--environment-label` CLI flag that override the environment's display label. Checked before the existing OS-level probes (`scutil`/`hostnamectl`/`os.hostname()`) in `resolveServerEnvironmentLabel`. When unset, behavior is unchanged.

## Why
T3 Code already supports running multiple independent server instances on one machine (`--base-dir`/`T3CODE_HOME`, `--port`, `--tailscale-serve-port`), but every instance's environment label is derived solely from the host's OS-level computer name, which is identical for every instance on that machine. Clients (mobile/desktop/web) can't tell the instances apart in the environment picker beyond remembering ports. `t3 pair --label` doesn't help — it only labels the pairing/connection record, not the main environment descriptor served at `/.well-known/t3/environment`.

## UI Changes
None — this only changes the `label` field of the `ExecutionEnvironmentDescriptor` JSON served by the server; no client UI code changed.

## Testing
- Full `apps/server` test suite passes (2631 passed / 7 skipped, no regressions), `pnpm typecheck` clean.
- New unit tests for override precedence (flag > env > OS probe) and blank/trim handling.
- Verified against two real, concurrently-running production self-hosted instances (not just isolated test servers) on the same physical machine — each now correctly reports its own distinct label via `/.well-known/t3/environment`.
- One client-side behavior worth reviewers' awareness: an already-paired client caches the environment descriptor from pairing time and does not appear to refetch it just because the server restarts with a new label — a full remove-and-re-pair was needed for already-paired clients to pick up the new label. Newly-paired clients see the correct label immediately. This is existing client caching behavior, not something this PR changes, but worth knowing when verifying the fix on an existing setup.

---

> [!NOTE]
> **Low Risk**
> Optional display-name config only; no auth, pairing, or data-path changes. Blank/unset still uses the existing OS label probes.
> 
> **Overview**
> Lets operators give each T3 Code instance on the same machine its own client-visible name, instead of every instance sharing the OS computer name.
> 
> Adds `--environment-label` / `T3CODE_ENVIRONMENT_LABEL` (flag wins over env). The trimmed value is stored on `ServerConfig` and used first in `resolveServerEnvironmentLabel`; blank values still fall back to existing OS probes. Unset keeps previous behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f5bd1cd33a784bab5b677d69ee19137448dc4f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add `environmentLabel` override to `ServerConfig` and CLI flags
> - Adds an optional `--environment-label` CLI flag and `T3CODE_ENVIRONMENT_LABEL` env var, resolved with flag-over-env precedence, whitespace trimming, and blank-as-undefined handling in [config.ts](https://github.com/pingdotgg/t3code/pull/7924/files#diff-16c068d320ce481e506eae4c06ff006a27585a4833a2d7d3dd1b584e5b8d8fdb)
> - Extends `ServerConfig.Service` with an optional `environmentLabel?: string` field that serves as a friendly display label overriding the OS-detected name
> - Updates `resolveServerEnvironmentLabel` in [ServerEnvironmentLabel.ts](https://github.com/pingdotgg/t3code/pull/7924/files#diff-b1aacbc192ffaacdfdd1ba5d7cbd3815df9976714c71601e826526144c769e59) to return the override label (when present and non-blank) before probing OS-specific sources or hostname
> - `ServerEnvironment.make` passes the configured override to the label resolver so the descriptor label reflects it
> - Risk: all in-tree test helpers and config builders were updated to include `environmentLabel: undefined`; out-of-tree consumers constructing `ServerConfig`-like objects will need to add the field if their types require it
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7f5bd1c.</sup>
> 
>


robnewton:
This was a mistake based on a misunderstanding of how to manage multiple accounts under a single server. Closing.

PR #7943, opened 2026-08-22
fix(web): stop rendering healthy ready providers as auth failures

### LINKED ISSUE #7932: [Bug]: Grok provider always shows "authentication could not be verified" even when the health check fully succeeds
### What happened?

The Grok provider works: models list, sessions run, `grok --version` and the ACP handshake both succeed. But every health-check cycle ends with the status detail:

```text
Installed and ready, but authentication could not be verified.
```

This reads as an auth failure. It is not one. The check never verifies auth at all, and the UI phrases "unknown" as a warning.

### Environment

```text
OS: Linux x86_64
T3 Code: 0.0.33 alpha (AppImage)
Grok CLI: grok 1.0.5 (5115b46bc9)
binaryPath: absolute path to ~/.grok/downloads/grok-linux-x86_64
Health check interval: default (300 s)
```

Grok CLI is authenticated and works standalone.

### Evidence

Local provider cache after a successful cycle (`~/.t3/caches/grok.json`):

```json
{
  "installed": true,
  "version": "1.0.5",
  "status": "ready",
  "auth": { "status": "unknown" },
  "models": [{ "slug": "grok-4.6" }, { "slug": "grok-4.5" }, { "slug": "grok-build" }]
}
```

`ready` plus models, but auth stuck on `unknown`. The server trace agrees: `checkGrokProviderStatus` spans complete in ~6.1-6.6 s with a Success exit (version probe plus ACP session startup).

### Root cause

Two pieces of source line up:

1. `checkGrokProviderStatus` never derives an auth result. Every success path returns `auth.status: "unknown"` (`apps/server/src/provider/Layers/GrokProvider.ts:305-310`). Unlike Codex, which gets auth from its `account/read` response, or Claude, which gets it from SDK initialization, Grok's ACP session is used only for model listing.
2. The web client renders any ready snapshot with unknown auth using that message as fallback detail (`apps/web/src/components/settings/providerStatus.ts:77-80`, from #2277).

So a completely healthy check displays a scary auth warning every five minutes.

### Expected behavior

Either the check verifies authentication (the ACP session already talks to xAI with valid OAuth, so some session response should confirm it), or a ready provider with unverified auth renders neutrally instead of sounding like a failure.

### Possible directions

1. Derive auth status inside the existing ACP session, mirroring how CodexProvider uses `account/read`.
2. If that is not feasible short-term, change the fallback text for ready snapshots to something like "Installed and ready; authentication not checked", or drop it when `status === "ready"`.

### Related

- #2495: similar confusion where a probe-layer gap reports as an availability/auth problem.

IzonIcy:
## What changed

One line in `getProviderSummary` (`apps/web/src/components/settings/providerStatus.ts`). A provider snapshot with `status: "ready"` and `auth.status: "unknown"` no longer falls back to the detail text *"Installed and ready, but authentication could not be verified."* The ready branch now renders `detail: provider.message ?? null` — headline stays **Available**, and the detail line is simply omitted unless the server supplied a message, matching how the authenticated/unauthenticated branches already behave.

Adds `providerStatus.test.ts` covering all branches of `getProviderSummary`, including a regression test for this case.

## Why

Fixes #7932. Grok's health check has no auth verification today, so its success path returns `status: "ready"` with `auth.status: "unknown"` and no message. Every 5-minute cycle, a fully healthy Grok install rendered text that reads like an auth failure in Settings → Connections.

The auth contract only offers `authenticated | unauthenticated | unknown` — there is no "verified by session" state, so claiming authenticated server-side would be a lie, and deriving real auth from the ACP session is a larger project. Rendering unknown-auth-but-ready neutrally is honest with a minimal diff. The chat-view banner already treats `ready` as healthy; this aligns the settings card with it. Mobile renders no summary text, desktop wraps web.

## Visual delta

Before: green dot + `Available - Installed and ready, but authentication could not be verified.`
After: green dot + `Available`

Happy to attach real before/after screenshots if you want them — say the word.

## Verification

- New test file: 9/9 passing
- Neighboring tests (`ProviderInstanceCard.test.ts`, `providerInstances.test.ts`): 30/30 passing
- Scoped typecheck (`tsgo --noEmit`) and lint clean

--

---

> [!NOTE]
> **Low Risk**
> Settings UI copy only; behavior is narrowed to omit misleading auth warnings when no server message exists, with tests locking the branches.
> 
> **Overview**
> Fixes misleading Settings → Connections copy for providers that are **ready** but report `auth.status: "unknown"` (e.g. Grok, which does not verify auth on probe).
> 
> In `getProviderSummary`, the default **Available** branch no longer falls back to *"Installed and ready, but authentication could not be verified."* It now uses **`provider.message ?? null`**, so the detail line is omitted unless the server sends a message—aligned with authenticated/unauthenticated branches and with chat treating `ready` as healthy.
> 
> Adds **`providerStatus.test.ts`** with branch coverage for `getProviderSummary`, including a regression test for pingdotgg/t3code#7932.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 368a60bf968d4f20b1fe6480b5ab7a905199d969. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Stop rendering ready providers as auth failures in `getProviderSummary`
> Updates the available/ready branch of `getProviderSummary` in [providerStatus.ts](https://github.com/pingdotgg/t3code/pull/7943/files#diff-29f82599bd9e3758e651b26db3d787ab623fa61350b9009233b476cd7d4e67fc) to use `provider.message` or `null` instead of a generic authentication-not-verified string. Adds a test suite in [providerStatus.test.ts](https://github.com/pingdotgg/t3code/pull/7943/files#diff-fe24578357c64585e13f49b9a6d1019673b9dac5f4d126d84ec4d6881b7d6748) covering pre-probe, disabled, CLI missing, authenticated, unauthenticated, warning, error, and ready-with-message cases.
> - Behavioral Change: ready providers without a server message now show no detail instead of the prior generic auth warning.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 368a60b.</sup>
> 
>


shivamhwp:

Closing as already implemented in #5362. The provider summary now shows Available for a ready provider with unknown authentication status, and the regression test covers that case.

PR #7946, opened 2026-08-23
fix(shared): read PATH from nushell login shells

### LINKED ISSUE #7846: PATH hydration fails when the login shell is nushell
On a machine where `/bin/nu` is the login shell, t3 logs this warning at startup:

```
[server] Failed to read PATH from login shell /bin/nu
```

The server still runs and terminals work, but the PATH captured from the login shell is dropped. On my machine that meant tools living in user paths, like nvm-managed node and cargo binaries, did not resolve inside t3.

The probe sent to the login shell is POSIX only (`packages/shared/src/shell.ts`, `buildEnvironmentCaptureCommand`):

```
printf '%s\n' '__T3CODE_ENV_PATH_START__'; printenv PATH || true; printf '%s\n' '__T3CODE_ENV_PATH_END__'
```

Nushell rejects this at parse time with `nu::parser::shell_oror`: it has no `||` operator, and its error suggests using `try` instead.

Repro: Arch Linux with nu as the login shell, node v26.7.0, run `npx t3@latest`.

A nushell variant I verified against the real shell:

```
print '__T3CODE_ENV_PATH_START__'; try { printenv PATH } catch { print '' }; print '__T3CODE_ENV_PATH_END__'
```

Same marker shape, so `extractEnvironmentValue` keeps working unchanged. Detecting shells whose basename starts with `nu` and sending that variant fixes it.

IzonIcy:
## What changed

`packages/shared/src/shell.ts`: when the login shell is nushell (basename `nu`, case-insensitive, `.exe` tolerated), the environment capture probe switches from the POSIX form to a nushell-flavored one:

```
print '__T3CODE_ENV_PATH_START__'; try { printenv PATH } catch { print '' }; print '__T3CODE_ENV_PATH_END__'
```

Same marker shape and statement structure as the POSIX probe, so `extractEnvironmentValue` is untouched. Non-nu shells keep the exact original command.

New tests: nu probe selection (with the no-`||` regression assertion), `.exe`/casing tolerance, and a guard that zsh keeps the POSIX probe. Failing-test-first — both nu cases failed before the change.

## Why

Fixes #7846. Nushell has no `||` operator and rejects the current probe at parse time (`nu::parser::shell_oror`), so on machines with `/bin/nu` as login shell the captured PATH was dropped entirely — nvm-managed node and cargo binaries failed to resolve inside t3 while terminals worked fine. The probe shape in this PR is the variant verified against a real nushell in the issue.

Shared-package only; no server/web behavior changes for POSIX or Windows shells.

## Verification

- `packages/shared` suite: 363/363 passing
- Scoped typecheck clean

--

---

> [!NOTE]
> **Medium Risk**
> Changes login-shell command generation used to capture PATH and other env vars, which affects tool discovery. Detection is basename-based and POSIX/Windows probes are unchanged.
> 
> **Overview**
> Fixes login-shell environment capture for **nushell** (`nu`), which previously rejected the POSIX probe (`printenv … || true`) at parse time and dropped PATH entirely.
> 
> `readEnvironmentFromLoginShell` now detects a `nu` binary (case-insensitive, `.exe` allowed, basename must be exactly `nu`) and emits a `try { printenv … } catch { print '' }` probe with the same start/end markers. Other shells keep the original POSIX command.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 04288406404319dd98ca633708eb10e8dfa1b5ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add Nushell support to `readEnvironmentFromLoginShell`
> - Adds `isNushellPath` helper to detect the `nu` binary (case-insensitive, ignoring `.exe` suffix) and `buildNushellEnvironmentCaptureCommand` to emit a Nushell-compatible capture script using try/catch instead of the POSIX `||` operator
> - `readEnvironmentFromLoginShell` now selects the Nushell or POSIX capture command based on the detected shell path; all other behavior is unchanged
> - Adds tests covering Nushell detection, probe selection, and correct fallback for non-Nushell shells
> - Behavioral Change: shells whose basename is exactly `nu` (case-insensitive, with optional `.exe`) now receive a different exec script; shells whose name merely contains the substring `nu` still use the POSIX probe
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0428840.</sup>
> 
>

### DIFF
--- a/packages/shared/src/shell.test.ts
+++ b/packages/shared/src/shell.test.ts
@@ -186,6 +186,53 @@ describe("readEnvironmentFromLoginShell", () => {
       CUSTOM_VAR: "  padded value  ",
     });
   });
+
+  it("sends a nushell-compatible probe when the login shell is nushell", () => {
+    const execFile = vi.fn<
+      (
+        file: string,
+        args: ReadonlyArray<string>,
+        options: { encoding: "utf8"; timeout: number },
+      ) => string
+    >(() => ["__T3CODE_ENV_PATH_START__", "/a:/b", "__T3CODE_ENV_PATH_END__"].join("\n"));
+
+    expect(readEnvironmentFromLoginShell("/bin/nu", ["PATH"], execFile)).toEqual({
+      PATH: "/a:/b",
+    });
+
+    const script = execFile.mock.calls[0]?.[1]?.[1];
+    // Nushell has no `||` operator and rejects the POSIX probe at parse time.
+    expect(script).toBeDefined();
+    expect(script).not.toContain("||");
+    expect(script).toContain("try { printenv PATH } catch { print '' }");
+  });
+
+  it("detects nushell regardless of case or a .exe suffix", () => {
+    const execFile = vi.fn<
+      (
+        file: string,
+        args: ReadonlyArray<string>,
+        options: { encoding: "utf8"; timeout: number },
+      ) => string
+    >(() => "");
+
+    readEnvironmentFromLoginShell("C:\\Program Files\\nu\\NU.EXE", ["PATH"], execFile);
+    const script = execFile.mock.calls[0]?.[1]?.[1];
+    expect(script).not.toContain("||");
+  });
+
+  it("keeps the POSIX probe for non-nushell shells", () => {
+    const execFile = vi.fn<
+      (
+        file: string,
+        args: ReadonlyArray<string>,
+        options: { encoding: "utf8"; timeout: number },
+      ) => string
+    >(() => "");
+
+    readEnvironmentFromLoginShell("/bin/zsh", ["PATH"], execFile);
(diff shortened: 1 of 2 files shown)


IzonIcy:
Fixed in 04288406 — `isNushellPath` now matches the basename exactly after stripping an optional `.exe` suffix, with a negative test for names that merely contain "nu".

t3dotgg:
> [!NOTE]

We are not taking on Nushell-specific login shell support in the current compatibility queue. This branch has a careful Nu-specific probe, exact executable matching, and focused tests, while [#7894](https://github.com/pingdotgg/t3code/pull/7894) is still an open alternative rather than landed coverage. We are keeping both implementations as reference if we return to [#7846](https://github.com/pingdotgg/t3code/issues/7846).

PR #7947, opened 2026-08-23
fix(contracts): reject project script ids that cannot form keybinding commands

### LINKED ISSUE #7851: [Bug]: Server accepts project script ids that the client keybinding schema rejects, every thread view of the project crashes
### Before submitting

### Area

packages/contracts or packages/shared (also affects apps/web and apps/server)

### Summary

`ProjectScript.id` is validated as `TrimmedNonEmptyString` on the server, but the web client rebuilds it into a keybinding command through `SCRIPT_RUN_COMMAND_PATTERN`, which only accepts `^[a-z0-9][a-z0-9-]*$` with max 24 characters. Any id that the server accepts but the client schema rejects (for example a UUID) makes every thread view of that project throw inside the React error boundary. The Settings > Projects panel throws on the same call, so the UI cannot repair the project.

### Steps to reproduce

1. Start the desktop app (v0.0.33) and add a project.
2. Issue a token: `npx t3@latest auth session issue --ttl 10m --token-only`.
3. Dispatch a script whose `id` is a UUID:

   ```sh
   curl -X POST http://127.0.0.1:3773/api/orchestration/dispatch \
     -H "authorization: Bearer $TOKEN" -H "content-type: application/json" \
     -d '{"type":"project.meta.update","commandId":"<uuid>","projectId":"<project-id>",
          "scripts":[{"id":"123e4567-e89b-42d3-a456-426614174000","name":"Setup",
                      "command":"echo hi","icon":"configure","runOnWorktreeCreate":false}]}'
   ```

   The server returns 200 and persists the `project.meta-updated` event.
4. Open any thread in that project, or open Settings > Projects for it.

### Expected behavior

One of:

- `ProjectScript.id` in `packages/contracts/src/orchestration.ts` (line 208) uses the same constraint as the middle part of `SCRIPT_RUN_COMMAND_PATTERN` in `packages/contracts/src/keybindings.ts`, so the server rejects the command with a schema error, or
- the client tolerates a non-conforming id (skip the keybinding lookup, or normalize the id) instead of throwing during render.

### Actual behavior

The thread view and the project settings panel show "Something went wrong. Expected a string matching template literal parts". "Try again" and "Reload app" bring the same screen back, because the bad id is persisted. The throw comes from `commandForProjectScript` in `apps/web/src/projectScripts.ts`:

```ts
export const commandForProjectScript = (scriptId: string): KeybindingCommand =>
  SCRIPT_RUN_COMMAND_PATTERN.make(`script.${scriptId}.run`);
```

called from the header script actions menu in `ChatView.tsx` (`scripts.map(s => ... commandForProjectScript(s.id))`) and from `ProjectSettingsPanel.tsx`.

### Impact

Blocks work completely (for every thread of the affected project, until the stored id is replaced)

### Version or commit

v0.0.33 desktop (macOS). Schemas checked against main @ 592c598.

### Environment

macOS 26 (Darwin 25.6), T3 Code (Alpha) 0.0.33 desktop, Claude Agent provider. The client used to dispatch was a script over the local HTTP API; the built-in UI never produces a UUID id, so this only shows up for API or `t3.json` tooling.

### Logs or stack traces

```shell
Error: Expected a string matching template literal parts
    at n.make (t3code://app/assets/textarea-BvTV_9vG.js:15:31292)
    at h6 (t3code://app/assets/index-DiDfaONg.js:2002:118162)      # commandForProjectScript
    at t3code://app/assets/index-DiDfaONg.js:2066:58631            # ChatView script actions menu
    at Array.map (<anonymous>)
    at xxt (t3code://app/assets/index-DiDfaONg.js:2066:58612)
    at vo (t3code://app/assets/index-DiDfaONg.js:9:47559)
```

### Workaround

Re-dispatch `project.meta.update` for the project with slug ids (`setup`, `run`, and so on) over `/api/orchestration/dispatch`. The projection updates and the views render again. Deleting and re-adding the project also works but loses its threads.

IzonIcy:
## What changed

`packages/contracts`: the identifier portion of `SCRIPT_RUN_COMMAND_PATTERN` is extracted into an exported `ProjectScriptId` schema (single source of truth), and `ProjectScript.id` now uses it instead of `TrimmedNonEmptyString`. Behavior change is exactly one: commands carrying script ids that cannot form a valid `script.<id>.run` keybinding command are rejected with a schema error at the dispatch boundary instead of being persisted.

Two new tests in `orchestration.test.ts`: the UUID id from #7851's repro now fails to decode; a conforming id still passes.

## Why

Fixes #7851, using its first suggested direction. The web client calls `commandForProjectScript(script.id)` (`SCRIPT_RUN_COMMAND_PATTERN.make("script." + id + ".run")`) while rendering thread views and Settings → Projects. The server accepted any trimmed non-empty id, so one dispatched UUID poisoned every read model of that project and both surfaces threw inside the React error boundary with no in-app way to repair — the persisted data re-threw on every render.

Client-side tolerance was the alternative direction, but it would add null-handling through ~10 keybinding call sites to accommodate input the system never intended to accept. Rejecting at the schema keeps orchestration pure and the client dumb. UI-created ids already conform via `normalizeScriptId`, so nothing changes for normal flows. Note this prevents new poison but does not migrate already-poisoned projects — happy to look at a cleanup path if maintainers want one.

## Verification

- Failing-test-first: with the old schema stashed, the UUID test fails (server accepts the poison), passes after
- contracts suite 268/268 · server orchestration suite 273/273 · web projectScripts suites 13/13
- Scoped typecheck clean

--

---

> [!NOTE]
> **Medium Risk**
> Tightens decode of project scripts and related events/commands. Existing payloads with UUID or otherwise non-conforming ids will now fail schema decode rather than persist or render.
> 
> **Overview**
> Stops persisting project script ids that cannot form a valid `script.<id>.run` keybinding. The web client rebuilds that command while rendering, so a UUID (or other non-conforming id) previously poisoned the project and crashed thread/settings views.
> 
> Exports a shared `ProjectScriptId` schema (lowercase alphanumerics/hyphens, max length) and uses it for both `SCRIPT_RUN_COMMAND_PATTERN` and `ProjectScript.id`. Tests cover rejecting a UUID and accepting a conforming id.
> 
> Does not migrate already-poisoned projects; it only blocks new ones. UI-created ids already go through `normalizeScriptId` / `nextProjectScriptId`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5999a9629caf7cbddfaab8175a227d1f24cd5df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Reject project script IDs that cannot form keybinding commands
> Adds a `ProjectScriptId` schema that enforces lowercase alphanumerics and hyphens (starting alphanumeric, within `MAX_SCRIPT_ID_LENGTH`). Changes `ProjectScript.id` from `TrimmedNonEmptyString` to this schema so payloads with non-conforming IDs fail decoding. The `SCRIPT_RUN_COMMAND_PATTERN` now reuses `ProjectScriptId` for the script-id portion.
> - Risk: `ProjectScript.id` decoding now rejects IDs like UUIDs or strings with uppercase/underscores; existing payloads with such IDs will fail to decode.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f5999a9.</sup>
> 
>

### DIFF
--- a/packages/contracts/src/keybindings.ts
+++ b/packages/contracts/src/keybindings.ts
@@ -75,12 +75,19 @@ export const STATIC_KEYBINDING_COMMANDS = [
   ...THREAD_KEYBINDING_COMMANDS,
 ] as const;
 
+/**
+ * Identifier portion of a script run command (`script.<id>.run`). The web
+ * client rebuilds that command during keybinding wiring, so any id persisted
+ * by the server must satisfy this constraint or thread views throw.
+ */
+export const ProjectScriptId = Schema.NonEmptyString.check(
+  Schema.isMaxLength(MAX_SCRIPT_ID_LENGTH),
+  Schema.isPattern(/^[a-z0-9][a-z0-9-]*$/),
+);
+
 export const SCRIPT_RUN_COMMAND_PATTERN = Schema.TemplateLiteral([
   Schema.Literal("script."),
-  Schema.NonEmptyString.check(
-    Schema.isMaxLength(MAX_SCRIPT_ID_LENGTH),
-    Schema.isPattern(/^[a-z0-9][a-z0-9-]*$/),
-  ),
+  ProjectScriptId,
   Schema.Literal(".run"),
 ]);
 
--- a/packages/contracts/src/orchestration.test.ts
+++ b/packages/contracts/src/orchestration.test.ts
@@ -221,6 +221,49 @@ it.effect("rejects command fields that become empty after trim", () =>
   }),
 );
 
+it.effect("rejects project scripts whose ids cannot form keybinding commands", () =>
+  Effect.gen(function* () {
+    // The web client rebuilds `script.<id>.run` keybinding commands while
+    // rendering; an id outside the pattern's constraint crashes thread
+    // views, so the server must refuse to persist it.
+    const result = yield* Effect.exit(
+      decodeProjectMetaUpdatedPayload({
+        projectId: "project-1",
+        updatedAt: "2026-01-01T00:00:00.000Z",
+        scripts: [
+          {
+            id: "123e4567-e89b-42d3-a456-426614174000",
+            name: "Setup",
+            command: "echo hi",
+            icon: "configure",
+            runOnWorktreeCreate: false,
+          },
+        ],
+      }),
+    );
+    assert.strictEqual(result._tag, "Failure");
+  }),
+);
+
+it.effect("accepts project scripts with conforming ids", () =>
+  Effect.gen(function* () {
+    const parsed = yield* decodeProjectMetaUpdatedPayload({
+      projectId: "project-1",
+      updatedAt: "2026-01-01T00:00:00.000Z",
+      scripts: [
+        {
+          id: "run-setup",
+          name: "Setup",
+          command: "echo hi",
+          icon: "configure",
+          runOnWorktreeCreate: false,
+        },
+      ],
+    });
+    assert.strictEqual(parsed.scripts?.[0]?.id, "run-setup");
+  }),
+);
+
 it.effect("decodes thread.turn.start defaults for provider and runtime mode", () =>
   Effect.gen(function* () {
     const parsed = yield* decodeThreadTurnStartCommand({
--- a/packages/contracts/src/orchestration.ts
+++ b/packages/contracts/src/orchestration.ts
@@ -23,6 +23,7 @@ import {
   TurnId,
 } from "./baseSchemas.ts";
 import { ProviderInstanceId } from "./providerInstance.ts";
+import { ProjectScriptId } from "./keybindings.ts";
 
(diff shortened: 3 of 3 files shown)


t3dotgg:
> [!NOTE]

Closing this alternative in favor of [#7961](https://github.com/pingdotgg/t3code/pull/7961). Tightening ProjectScript.id also changes decoding of scripts that may already be stored. The retained approach accepts existing IDs when building keybinding commands instead of rejecting their saved data. Please preserve the failing command case in that compatible fix. [#7961](https://github.com/pingdotgg/t3code/pull/7961) is still open.

PR #7755, opened 2026-08-21
fix(server): strip OSC/ANSI escapes from OpenCode CLI inventory and stored agent selections

ImBIOS:
Fixes #7754

### Problem
opencode <=1.18 leaks OSC title `ESC ]0;<cwd>: ready BEL` to stdout for every non-help command even when piped. T3's `runOpenCodeCommand` captures that stdout and `parseModelsCliOutput` / `parseAgentListCliOutput` stored polluted ids like `"\u001b]0;imbios: ready\u0007build"` in `model_selection_json`.

Next `sendTurn` → `session.promptAsync({agent: polluted})` → opencode rejects:

```
Agent not found: "\u001b]0;imbios: ready\u0007build". Available agents: build, explore, general, plan
```

Surfaced as two `session.error` per turn: first the polluted-agent message, second the generic `UnknownError` stack at `SessionPrompt.createUserMessage … SessionHttpApi.promptAsync`.

Repro evidence in ~/.t3/userdata/logs/provider/events.*.log and projection_threads (3 polluted threads at ImBIOS). Also `opencode agent list | cat -v` shows `^[]0;tmp: ready^Gbuild (primary)` even with `env -i`/`TERM=dumb` — leak is in opencode binary, not shell. Same leak pollutes `models --verbose` (model inventory dropped to 0 pre-fix because SLUG_RE didn't match) and `debug skill` (JSON parse failed → skills=[]).

Fork fix already landed in ImBIOS/t3code#2 and ImBIOS:main 30d9c19; this is the upstream PR.

### Fix
- `packages/shared/src/stripTerminalEscapes.ts` — shared OSC/CSI/charset sanitizer (`stripTerminalEscapes`/`sanitizeTerminalValue`)
- `apps/server/src/provider/opencodeRuntime.ts:208,270,312` — strip at entry of `parseModels/Agent/Skills`; keeps `-verbose` from dropping models and skills from degrading
- `apps/server/src/provider/Layers/OpenCodeProvider.ts:164,188,393` — sanitize inventory agent names/variants + `--version` parsing
- `apps/server/src/provider/Layers/OpenCodeAdapter.ts:1476` + `textGeneration/OpenCodeTextGeneration.ts:411` — sanitize stored `agent`/`variant` before `promptAsync` (steer + title generation)
- `packages/shared/src/model.ts:66,313` — sanitize persisted `ProviderOptionSelection` values and model slugs on read (transparently repairs polluted threads without DB migration)
[...]

PR #7988, opened 2026-08-23
fix(server): strip terminal escape sequences from opencode CLI output

Exotic209093:
## Problem

opencode <= 1.18 emits an OSC title set (`ESC ]0;<cwd>: ready BEL`) on stdout even when stdout is a pipe. The escape bytes leak into everything parsed from its CLI output: `parseServerUrlFromOutput` misses the ready line, model slugs gain a prefix, agent names pick up garbage, and skill JSON fails to decode.

## Fix

Add a shared helper, `@t3tools/shared/stripTerminalEscapes`, that removes OSC sequences (terminated by BEL or ST) and CSI sequences while leaving ordinary text alone (unterminated prefixes are preserved). Apply it in `opencodeRuntime.ts` before parsing server output, models, agents, and skills. Since `JSON.stringify` encodes control bytes textually inside JSON values, all three CLI parsers also strip decoded string values via a `JSON.parse` reviver — model fields at any depth, agent permission rules, and skill metadata — while non-string values and object/array structure pass through untouched.

Tests: direct unit tests for the helper plus parser coverage for raw-byte titles, escapes embedded in decoded model/permission/skill fields, and unterminated-sequence preservation.

> [!NOTE]
> ### Strip terminal escape sequences from opencode CLI output parsers
> - Adds `stripTerminalEscapes` to `@t3tools/shared`, which removes OSC and CSI sequences from strings, exported via the new `./stripTerminalEscapes` subpath
> - Applies the helper and a `stripTerminalEscapesJsonReviver` to `parseServerUrlFromOutput`, `parseModelsCliOutput`, `parseAgentListCliOutput`, and `parseSkillsCliOutput` so CLI output is cleaned before line splitting and JSON decoding
> - `parseSkillsCliOutput` now uses `JSON.parse` with the reviver before schema validation (`decodeOpenCodeSkillsArrayExit`) instead of decoding from JSON text, returning an empty array on parse failure
> - Risk: `parseSkillsCliOutput` no longer validates raw JSON text via schema; malformed input returns `[]` instead of throwing, so downstream callers expecting schema errors on bad input will see an empty array
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 842a0dd.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/opencodeRuntime.cliParsers.test.ts
+++ b/apps/server/src/provider/opencodeRuntime.cliParsers.test.ts
@@ -257,6 +257,136 @@ describe("parseAgentListCliOutput", () => {
   });
 });
 
+describe("terminal escape stripping", () => {
+  // opencode <= 1.18 emits `ESC ]0;<cwd>: ready BEL` on stdout for every
+  // command, even when stdout is a pipe.
+  const OSC_TITLE = "\u001b]0;tmp: ready\u0007";
+
+  it("parseModelsCliOutput ignores an OSC title before the first slug", () => {
+    const stdout = [
+      `${OSC_TITLE}openai/gpt-4o`,
+      JSON.stringify({ id: "gpt-4o", providerID: "openai", name: "GPT-4o" }),
+    ].join("\n");
+
+    const result = parseModelsCliOutput(stdout);
+    NodeAssert.deepEqual([...result.connected], ["openai"]);
+    const model = result.providers.get("openai")!.models["gpt-4o"]!;
+    NodeAssert.ok(model);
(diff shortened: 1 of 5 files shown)


t3dotgg:
> [!NOTE]

We selected [#7755](https://github.com/pingdotgg/t3code/pull/7755) for the current OpenCode escape repair. It covers raw OSC and ANSI leaks in model, agent, and skill inventory, and it sanitizes stored selections and turn input. This branch also handles escapes inside decoded JSON strings and has unterminated-sequence safety tests. Those cases are not in [#7755](https://github.com/pingdotgg/t3code/pull/7755) today, so we are keeping them as reference rather than claiming that they landed.

PR #7994, opened 2026-08-23
fix(desktop): route t3code:// thread deep links instead of only raising the window

CurlyPeter:
## The bug

The desktop app registers the `t3code://` scheme (`Info.plist` on macOS, the Clerk bridge's `setAsDefaultProtocolClient`, the Linux `.desktop` entry) but routes exactly one URL family: Clerk's OAuth callback. Every other URL is accepted by the OS, delivered to the app and dropped.

So opening a thread link brings the window to the front and leaves it wherever it already was. Measured: `open "t3code://app/<env>/<thread>"` moves frontmost to T3 Code and nothing else happens; the only `open-url` listener in a shipped build belongs to `@clerk/electron`'s OAuth transport, and there is no URL routing in `apps/desktop/src`.

Meanwhile the renderer has had the route all along, `_chat.$environmentId.$threadId`. Nothing connected the two.

## What this does

Accepts the shape the relay already emits for push notifications (`deepLink: "/threads/<environmentId>/<threadId>"`) and the mobile router already matches (`threads/:environmentId/:threadId`), so one link opens the same chat on a phone and on a desktop.

Delivery **reuses the existing menu-action channel** rather than adding one. `dispatchMenuAction` already creates the window when there is none, defers the send until `did-finish-load` while the renderer boots, and reveals the window afterwards; a dedicated channel would have duplicated all of that plus a preload entry. The renderer rule lives in its own module so it is testable without rendering, and `AppSidebarLayout` keeps a single call.

## Cold start needed three supporting fixes

The launch is exactly the moment when nothing is ready, and each of these is invisible in the steady state:

- **`open-url` is captured synchronously at process bootstrap.** Electron buffers nothing and documents that a listener registered in response to `ready` misses the URL that launched the app. Even registering before `whenReady` sits behind the whole layer graph plus the shell environment, user-data path, settings load and Clerk bootstrap, each of which yields to the run loop.
- **The preload buffers menu actions until the renderer subscribes.** `onMenuAction` attaches its IPC listener lazily, and its only caller renders inside the authenticated app shell, which is gated on an async bootstrap — while the main process sends on `did-finish-load`, which does not wait for it. Without the buffer a cold-start action is sent into a channel with no listener.
- **`DesktopWindow` parks a menu action that arrives with no window and no ready backend**, replaying it from `handleBackendReady` through the same delivery path. That state is invisible for a menu item, since the user can click again, and terminal for a launch URL. The drain sits in `ensuring` so a failed window creation cannot strand it.

## Parsing

Deliberately **at least as strict** as the mobile app's `normalizeThreadDeepLink`: no query, no fragment, no credentials, exact segment count, and both ids must be UUIDs.

The UUID requirement is a boundary, not a formatting preference. A URL handler is an unauthenticated entry point that any local process can drive. Without it, `t3code://threads/settings/connections` resolves to `/settings/connections` — a real static route that outranks `/$environmentId/$threadId` — so anything on the machine could silently move the window onto a settings page.

The listener declines quietly for anything that is not a thread link and calls `preventDefault()` only for URLs it owns, so Clerk's callbacks are untouched.

## Scope

`open-url` is macOS-only. Windows and Linux deliver the URL as argv to a second instance; that path is left for a follow-up rather than guessed at.

## Testing

- 19 new parser tests, 5 for the renderer rule, 1 for the park-and-replay branch (verified to fail when the replay is removed).
- Full suites green on this base: `@t3tools/desktop` 599 tests, `@t3tools/web` 2737 tests. Both typechecks clean.
- Exercised on real machines, warm and cold. Cold start with the app fully quit produces `desktop.urlRouting.register` → `handleUrl` → `dispatchMenuAction` → `deliverMenuAction`, with the register span reporting `buffered: 1`, i.e. the launch URL came through the bootstrap collector.

---

> [!NOTE]
> **Medium Risk**
> Touches an unauthenticated OS protocol handler and in-app navigation. Parsing is UUID-strict and the renderer rejects non-absolute paths, but delivery still rides the shared menu-action IPC path.
> 
> **Overview**
> Opens the same thread a `t3code://threads/<environmentId>/<threadId>` link already opens on mobile, instead of only bringing the desktop window forward.
> 
> Main process parses that shape (UUID ids only, no query/fragment/credentials, both authority forms) and sends a `navigate:` action on the existing menu-action channel. Clerk OAuth URLs on the same scheme are left alone. Registration happens before `whenReady`, with a sync `open-url` collector at process start so a launch URL is not missed. Windows/Linux argv forwarding is out of scope.
> 
> Cold start is covered by parking a menu action until the backend is ready (replayed once from `handleBackendReady`) and by buffering menu-action IPC in preload until the renderer can subscribe. The renderer maps `navigate:` to an in-app path and refuses protocol-relative or off-app targets.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 096cf1dee874184ee81bb5d1326d5912db472ef7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Route `t3code://` thread deep links to the renderer instead of only raising the window
> - Adds a `DesktopUrlRouting` service that parses and validates `t3code://threads/<environmentId>/<threadId>` deep links, then dispatches a `navigate:`-prefixed menu action to the renderer
> - Registers a synchronous `open-url` bootstrap listener in [main.ts](https://github.com/pingdotgg/t3code/pull/7994/files#diff-31a471f6ef958ceff6e87ee910f4bc4f7bbc4986f797f159353b328a5916f7cb) before Electron's `whenReady` so macOS cold-start events are not lost
> - Buffers incoming menu actions in [preload.ts](https://github.com/pingdotgg/t3code/pull/7994/files#diff-eae0078d2bf4ac4beed2884f918b2816a97752917390f09980ba2cfad8c42cac) before the renderer subscribes, and parks a single action in [DesktopWindow.ts](https://github.com/pingdotgg/t3code/pull/7994/files#diff-d92229c509071d1e9d3595d187fe2f1f13c394b22e1d9d2070ed6fb99678f429) until the backend is ready, replaying it once
> - The web renderer resolves `navigate:` actions in [menuActionNavigation.ts](https://github.com/pingdotgg/t3code/pull/7994/files#diff-647b515ea5ba20a7a7da8f5b59dc051c8cb2e4d1273ae6925d8d1b0f10eeea6a) and performs in-app navigation, rejecting external or protocol-relative targets
> - Risk: `parseThreadDeepLink` in [DesktopUrlRouting.ts](https://github.com/pingdotgg/t3code/pull/7994/files#diff-18deeca47414574e8f49694107d19e48d11c9d01db1640b08b12822811c3a16d) only accepts UUID-shaped IDs and two authority forms (`//threads/...` and `///threads/...`); URLs with credentials, query, fragment, or extra segments are silently dropped
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 096cf1d. 7 files reviewed, 2 issues evaluated, 0 issues filtered, 2 comments posted</summary>
>
> ### 🗂️ Filtered Issues
>
> </details>
>

### DIFF
--- a/apps/desktop/src/app/DesktopApp.ts
+++ b/apps/desktop/src/app/DesktopApp.ts
@@ -22,6 +22,7 @@ import * as DesktopLinuxUrlHandler from "./DesktopLinuxUrlHandler.ts";
 import * as DesktopObservability from "./DesktopObservability.ts";
 import * as DesktopPreReadyPlatform from "./DesktopPreReadyPlatform.ts";
 import * as DesktopShutdown from "./DesktopShutdown.ts";
+import * as DesktopUrlRouting from "./DesktopUrlRouting.ts";
 import * as DesktopServerExposure from "../backend/DesktopServerExposure.ts";
 import * as DesktopAppSettings from "../settings/DesktopAppSettings.ts";
 import * as DesktopShellEnvironment from "../shell/DesktopShellEnvironment.ts";
@@ -224,6 +225,7 @@ const startup = Effect.gen(function* () {
   const electronApp = yield* ElectronApp.ElectronApp;
   const lifecycle = yield* DesktopLifecycle.DesktopLifecycle;
   const linuxUrlHandler = yield* DesktopLinuxUrlHandler.DesktopLinuxUrlHandler;
+  const urlRouting = yield* DesktopUrlRouting.DesktopUrlRouting;
   const clerk = yield* DesktopClerk.DesktopClerk;
   const shellEnvironment = yield* DesktopShellEnvironment.DesktopShellEnvironment;
   const desktopSettings = yield* DesktopAppSettings.DesktopAppSettings;
@@ -271,6 +273,11 @@ const startup = Effect.gen(function* () {
   yield* appIdentity.configure;
   yield* lifecycle.register;
   yield* clerk.configure;
+  // BEFORE whenReady, deliberately. On macOS a cold-start `open-url` (the OS
+  // launching the app because the user opened a t3code:// link) can fire before
+  // the ready event, and a listener installed after it would never see the URL
+  // that caused the launch.
+  yield* urlRouting.register;
 
   yield* electronApp.whenReady.pipe(
     Effect.withSpan("desktop.electron.whenReady"),
--- a/apps/desktop/src/app/DesktopUrlRouting.test.ts
+++ b/apps/desktop/src/app/DesktopUrlRouting.test.ts
@@ -1,0 +1,156 @@
+import { assert, describe, it } from "@effect/vitest";
+
+import {
+  NAVIGATE_ACTION_PREFIX,
+  navigateAction,
+  parseThreadDeepLink,
+  threadRoutePath,
+} from "./DesktopUrlRouting.ts";
+
+const SCHEME = "t3code";
+const ENV = "11111111-2222-3333-4444-555555555555";
+const THREAD = "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee";
+
+describe("parseThreadDeepLink", () => {
+  it("parses the shape the relay and the mobile router already use", () => {
+    assert.deepEqual(parseThreadDeepLink(`${SCHEME}://threads/${ENV}/${THREAD}`, SCHEME), {
+      environmentId: ENV,
+      threadId: THREAD,
+    });
+  });
+
+  it("accepts the empty-authority form, where threads lands in the path", () => {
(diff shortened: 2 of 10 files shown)


t3dotgg:
> [!NOTE]

We are keeping OPEN [#8246](https://github.com/pingdotgg/t3code/pull/8246) as the complete review path for desktop thread deep links. The focused URL-routing tests here remain useful reference.

PR #8075, opened 2026-08-24
fix(web,mobile): keep substantive assistant prose outside the settled-turn fold

spiky02plateau:
When a settled turn folds behind the "Worked for ..." row, every assistant message between the first and the terminal one is hidden, regardless of what it contains. The server mints a separate message per assistant stream segment (segments close at tool boundaries), so a model that interleaves real prose with tool calls gets the substance of its reply folded: write three paragraphs of analysis, run a tool, close with "Done", and the analysis is invisible until the fold is clicked. With reasoning-heavy models this happens on most turns, and users report the majority of a reply routinely hidden.

The middle of a turn is only sometimes narration. "Looking around first." should fold; a multi-paragraph explanation of the root cause should not. Since nothing in the data model distinguishes the two, this adds a small shared predicate, `isFoldableAssistantNarration` in `@t3tools/client-runtime/state/turn-fold`: a mid-turn assistant message folds only when it is a single block (no blank-line paragraph break) of at most 240 characters. Web and mobile both apply it, so the two surfaces keep agreeing on what a fold contains. Everything else is unchanged: tool and work rows still fold, the first and terminal messages stay visible as today, agent-spawn rows keep their escape, and the running turn is never folded.

One boundary I left alone: the copy button still appears only on the terminal message, because `showAssistantMeta` doubles as the row-spacing signal in the timeline and decoupling them is a layout change of its own.

## Verification

- `vp test run src/state/turnFold.test.ts` in `packages/client-runtime`: 6 passed (boundary at 240, paragraph breaks, whitespace-only)
- `vp test run src/components/chat/MessagesTimeline.logic.test.ts` in `apps/web`: 46 passed, including a new case asserting a long mid-turn message stays visible while tool rows still fold; the new case fails on main
- `vp test run src/lib/threadActivity.test.ts` in `apps/mobile`: 20 passed, same new case for the mobile fold; also fails on main
- `tsgo --noEmit`: clean in `apps/web` and `packages/client-runtime`; `apps/mobile` output byte-identical to the main baseline (pre-existing react-navigation noise only)
- No existing test expectation changed; the short narration fixtures in the current fold tests still fold

## Screenshots

Same seeded thread, identical crops, collapsed state in both (the fold was never expanded). The turn is: "Looking around first." / tool run / two-paragraph analysis / tool run / "Done. The fix is batching the rebuild."

Before, the analysis is inside the fold:

before

After, the analysis renders; the tool rows are still behind "Worked for 18s":

after

Change made by Claude Opus via Claude Code.

---

> [!NOTE]
> **Low Risk**
> Presentation-only fold heuristic on web and mobile; no auth, data, or protocol changes. Heuristic may still hide some mid-length single-paragraph answer text.
> 
> **Overview**
> Stops the settled-turn “Worked for …” fold from hiding real mid-turn answer text. Stream segments close at tool calls, so one reply is often several messages; previously everything between the first and last assistant message was folded.
> 
> Web and mobile now share `isFoldableAssistantNarration`: a mid-turn assistant message folds only if it is a single block of at most 240 characters. Longer or multi-paragraph chunks stay visible; tool/work rows, first/terminal messages, and running turns are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 57238e2092da608d490b10b47adf1d439cb3faf4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Keep substantive mid-turn assistant messages visible in settled-turn folds
> - Adds `isFoldableAssistantNarration` in [turnFold.ts](https://github.com/pingdotgg/t3code/pull/8075/files#diff-a8ccd3561591eedf1a1dce6506524342cffb94a0a9f0c49906d6dbffc0511c8b): an assistant message is foldable only if it is under 240 chars and contains no blank-line paragraph break.
> - Exposes this utility via the new `./state/turn-fold` subpath export in [package.json](https://github.com/pingdotgg/t3code/pull/8075/files#diff-6e89e5cdd29d27c6871a007cebd586b99715fcdf8b711ff74f5d0b4f8e691db1).
> - Updates `deriveThreadFeedTurnFolds` ([threadActivity.ts](https://github.com/pingdotgg/t3code/pull/8075/files#diff-a18e8ed6aa4a9d6b3d9e32ea2fa3034dfd7621b6af3d640ec1b95a8564c97eb5)) and `deriveTurnFolds` ([MessagesTimeline.logic.ts](https://github.com/pingdotgg/t3code/pull/8075/files#diff-6cdfacc6ebbdbc5c4b4058c0a430b87d768ac2b6accf35fb4c9dfc23500914fe)) so mid-turn assistant messages that are not foldable narration stay visible instead of being hidden with other work entries.
> - Risk: any mid-turn assistant message shorter than `FOLDABLE_NARRATION_MAX_CHARS` with no paragraph break will still be folded; content that previously collapsed may now appear inline and change timeline scroll height.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 57238e2.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/lib/threadActivity.test.ts
+++ b/apps/mobile/src/lib/threadActivity.test.ts
@@ -620,6 +620,77 @@ describe("buildThreadFeed", () => {
     ]);
   });
 
+  it("keeps a multi-paragraph mid-turn message visible while its work still folds", () => {
+    const turnId = TurnId.make("turn-1");
+    const thread = makeThread({
+      id: ThreadId.make("thread-substantive-middle"),
+      projectId: ProjectId.make("project-1"),
+      title: "Substantive narration",
+      latestTurn: {
+        turnId,
+        state: "completed",
+        requestedAt: "2026-04-01T00:00:00.000Z",
+        startedAt: "2026-04-01T00:00:01.000Z",
+        completedAt: "2026-04-01T00:00:08.000Z",
+        assistantMessageId: MessageId.make("assistant-final"),
+      },
+      messages: [
+        {
+          id: MessageId.make("assistant-first"),
+          role: "assistant",
+          text: "Looking around first.",
+          turnId,
+          streaming: false,
+          createdAt: "2026-04-01T00:00:01.000Z",
+          updatedAt: "2026-04-01T00:00:02.000Z",
+        },
+        {
+          id: MessageId.make("assistant-analysis"),
+          role: "assistant",
+          text: "The projector rebuilds the read model on every event.\n\nThat is why the timeline lags behind the socket by one frame.",
+          turnId,
+          streaming: false,
+          createdAt: "2026-04-01T00:00:04.000Z",
+          updatedAt: "2026-04-01T00:00:05.000Z",
+        },
+        {
+          id: MessageId.make("assistant-final"),
+          role: "assistant",
+          text: "Done.",
+          turnId,
+          streaming: false,
+          createdAt: "2026-04-01T00:00:07.000Z",
+          updatedAt: "2026-04-01T00:00:08.000Z",
+        },
+      ],
+      activities: [
+        makeActivity({
+          id: EventId.make("tool-completed"),
+          kind: "tool.completed",
+          tone: "tool",
+          summary: "Read files",
+          createdAt: "2026-04-01T00:00:03.000Z",
+          turnId,
+          payload: {
+            title: "Read files",
+            itemType: "file_read",
+            status: "completed",
+          },
+        }),
+      ],
+    });
+
+    const feed = buildThreadFeed(thread);
+    const collapsed = deriveThreadFeedPresentation(feed, thread.latestTurn, new Set());
+
+    expect(collapsed.map((entry) => entry.id)).toEqual([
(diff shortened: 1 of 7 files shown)


iskandersierra:
I can reproduce the settled-turn folding behavior during an interactive interview flow.

The affected turn contains multiple assistant outputs: first a substantive Markdown question round, then a shorter progress/status or follow-up message. Once the turn settles, the UI keeps the last assistant entry visible and folds the earlier assistant output into the work section. As a result, the question list appears to be missing even though it was emitted and persisted normally.

Asking the agent to repeat the questions as plain text makes them visible because the repeated response is the terminal assistant output of a new turn. This suggests the issue is related to settled-turn folding and terminal-message selection, rather than question generation or data loss.

This appears to be another instance of the existing issue where substantive assistant prose is hidden when a turn ends with a shorter progress message. Keeping substantive assistant messages visible while folding only work/progress entries would address this case as well.

PR #7887, opened 2026-08-22
feat(web): configure worktree branch prefixes

yasinkavakli:
## What Changed

- Added a server setting for the prefix used when T3 Code promotes temporary worktree branches to generated names.
- Added a **Worktree branch prefix** field under **Settings → General**. It defaults to `t3code`, accepts a normalized custom prefix, and can be left empty for unprefixed branch names.
- Added settings search, restore-default behavior, user documentation, and focused contract/server tests.
- Kept temporary branch naming and the wire protocol unchanged. The configured prefix is applied at the shared server rename boundary, so web, desktop, mobile, local, and remote clients receive the same behavior.

## Why

Generated worktree branches were always placed under the hard-coded `t3code/` namespace. Fork maintainers and teams commonly need a personal or team namespace, while some workflows require no prefix at all.

This makes that namespace configurable without adding client-specific flows or changing worktree creation semantics. Existing installations remain backward compatible because `t3code` is still the default.

## UI Changes

**Before:** Settings had no branch-prefix control, and generated worktree branches always used `t3code/`.

**After:** Settings → General includes a Worktree branch prefix field with inline guidance.

Worktree branch prefix setting in Settings General

## Verification

- Rebased onto the latest `main` before publishing.
- `96` focused tests passed across shared Git helpers, settings contracts, and the provider command reactor.
- Shared, contracts, server, and web typechecks passed.
- Changed-file lint, formatting, and `git diff --check` passed.
- Three independent review passes covered correctness, multi-surface/remote behavior, and unnecessary complexity; follow-up delta reviews found no actionable issues.
- Confirmed the functionality is not already present on upstream `main` and found no open upstream PR implementing it.

## Checklist

- [x] This PR is small and focused on one change
- [x] The problem and implementation are explained above
- [x] UI evidence is included above; the before state had no corresponding control
- [x] Video is not applicable because this change adds no animation or timing-dependent interaction

Built with Codex through the Codex harness in T3 Code.

Generated with Codex

---

> [!NOTE]
> **Medium Risk**
> Touches git branch renaming and persisted server settings. Validation and a backward-compatible default limit blast radius, but a bad prefix still affects generated branch names across clients.
> 
> **Overview**
> Makes the namespace for auto-generated worktree branches configurable. Default remains `t3code`; users can set a custom prefix or leave it empty.
> 
[...]

PR #8082, opened 2026-08-24
feat(git): let users choose the worktree branch prefix

t3dotgg:
Generated worktree branches always start with `t3code/`, which conflicts with personal and team branch conventions.

Add a server-owned **Worktree branch prefix** setting under **Settings > General**. It defaults to `t3code`, accepts validated custom prefixes, and can be cleared to create branches without a prefix. Temporary branches stay unchanged, and the server applies the prefix when it generates the final branch name for web, desktop, mobile, and remote clients.

Inspired by and with thanks to:
- @yasinkavakli for [#7887](https://github.com/pingdotgg/t3code/pull/7887)
- @filipgutica for [#6037](https://github.com/pingdotgg/t3code/pull/6037)
- @jssblck for [#3954](https://github.com/pingdotgg/t3code/pull/3954)
- @MrSimmmons for [#6428](https://github.com/pingdotgg/t3code/pull/6428)

Verified with 116 focused tests, targeted lint, and scoped typechecks for contracts, shared, client runtime, server, web, and mobile.

Model: GPT-5.6 Sol
Harness: Codex in T3 Code

---

> [!NOTE]
> **Medium Risk**
> Touches git branch naming and persisted server settings. Invalid prefixes are schema-validated, but misconfigured prefixes still change how worktrees are renamed for all clients.
> 
> **Overview**
> Users can now set a **Worktree branch prefix** in Settings > General. Generated worktree names still default to `t3code/…`, but a custom prefix (including nested paths like `teams/platform`) is applied server-side, and clearing the field creates unprefixed branches.
> 
> `buildGeneratedWorktreeBranchName` strips a leftover `t3code/` from the model output when it is not the configured prefix, then sanitizes the remainder. Temporary worktree branches are unchanged.
> 
> The new `worktreeBranchPrefix` setting is validated (lowercase, max 64, no `//` or leading/trailing separators) and wired through dirty/reset/search plus user docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e5ed3f01680ba8445bb54879d389f61454711e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add configurable `worktreeBranchPrefix` setting defaulting to `t3code`
> - Adds a `worktreeBranchPrefix` field to `ServerSettings`, defaulting to `t3code` and validated by the new `WorktreeBranchPrefix` schema (max 64 chars, lowercase alphanumerics, no adjacent slashes).
> - Updates `buildGeneratedWorktreeBranchName` in [ProviderCommandReactor.ts](https://github.com/pingdotgg/t3code/pull/8082/files#diff-0630e9ad70d5dcd1c8586d29297f9b7d5f5436cc8e10fc88dc22acc511f1450d) to accept a `prefix` argument and sanitize it via `sanitizeBranchFragment`.
> - Adds a UI control in the General Settings panel to edit the prefix, with settings search indexing and restore-defaults support.
> - Behavioral Change: `buildGeneratedWorktreeBranchName` now requires a `prefix` parameter; existing callers must be updated.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5e5ed3f.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
@@ -150,6 +150,7 @@ describe("ProviderCommandReactor", () => {
     readonly requiresNewThreadForModelChange?: boolean;
     readonly titleRegenerationCompletionDispatchFailures?: number;
     readonly titleRegenerationBeforeStart?: "one" | "two";
+    readonly worktreeBranchPrefix?: string;
     readonly interruptTurnEffect?: () => Effect.Effect<void, ProviderAdapterRequestError>;
     readonly stopSessionEffect?: () => Effect.Effect<void, ProviderAdapterRequestError>;
     readonly startSessionEffect?: (
@@ -426,7 +427,13 @@ describe("ProviderCommandReactor", () => {
           generateThreadTitle,
         }),
       ),
-      Layer.provideMerge(ServerSettingsService.layerTest()),
+      Layer.provideMerge(
+        ServerSettingsService.layerTest(
+          input?.worktreeBranchPrefix === undefined
+            ? {}
+            : { worktreeBranchPrefix: input.worktreeBranchPrefix },
+        ),
+      ),
       Layer.provideMerge(ServerConfig.layerTest(process.cwd(), baseDir)),
       Layer.provideMerge(NodeServices.layer),
     );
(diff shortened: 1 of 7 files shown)


MrSimmmons:
For my example of #6428, do you have thoughts on allowing full branch name customization at worktree creation rather than just the prefix instead?
The way that my teams works is we use the prefix of the branch to indicate the type of branch (so `bug/...`, `feat/...` etc) and just allowing the customization of it wouldn't really work for that use case.
Otherwise I don't really see the a use for the prefix of a branch?

Where as if you gave people the option to just fully customize the branch name on worktree creation, they can then have whatever they want and everyone wins

jetpham:
@MrSimmmons branch prefixes are conventionally used for distinguishing the main developer of a branch in the form "MrSimmons/fix-ssrf" so having settings globally to set your "name" then that being used as your branch prefix seems correct here. I've seen in the past [conventional commit](https://www.conventionalcommits.org/) style for commits and pr titles which can be helpful for categorization.

vitorpacheco:
Cool feature! One question though — is worktreeBranchPrefix meant to stay global/server-wide, or could it also work as a per-project override?

I bounce between a few repos that each have their own branch convention.

dev2adventure:
I feel like this doesn't fully resolve the issue at hand, since, for example, the branch names supplied by Linear follow a strict pattern, and we would need full customization for that and not just branch name prefixes.

kevhardy:
I think people are focusing on this PR as if it’s the one PR meant to fix every single branch naming problem they have. But the reality is, the only fixed portion of what we can’t change today is this `t3code/` prefix. If this PR allows us to either remove that or change the value, I think that is a very good step in the right direction.

There may be more branch naming options in the future, like naming the branch yourself when creating a thread or allowing a custom prompt for the source control writer to determine branch names. I personally think those are out of scope for this PR, especially considering there are already plenty of workarounds, such as telling your agents to rename branches to fit your needs.

GuusDeGraeve:
Agreed with @kevhardy, being able to change or remove the prefix is more than enough. Anything beyond that can be handled by putting branch naming guidelines in AGENTS.md or CLAUDE.md. As long as the t3code/ prefix isn't enforced, everything stays possible.

Rasalas:
also agreeing with @kevhardy and @GuusDeGraeve here.

This should be merged "as is" (if the issues are fixed)- this probably solves like 95% of people's problems. Linear compat could be separate ticket/MR.

I was experimenting with an optional input in the "new worktree" flow.
In my imagination - if you hit this:  
20260902-205551_T3 Code (Nightly)@2x

you just open an optional input that allows you to put anything you want there.
If you send, it autogenerates. If you switch to the prompt, it defaults to autogenerate. Only if you set "New worktree" and then type, it sets what you decided.   
I feel like this would solve 99% of peoples problems.

aaronkerckhoff:
I agree that this PR probably should be merged as-is.
But I would be open to file a stacked PR to expand the branch naming feature. I'd add custom naming rules as well as the ability for maintainers to set certain naming rules for their repo (perhaps in t3.json?) as requested by @juliusmarminge in #6865.

match-josephcampuzano:
Wondering if we can get this through. I agree that this is a good first step and solves the 80% problem for most people. Those with more strict naming conventions or something along those lines could probably be solved later.

Atm I just ask the agent to rename my branches before PR's but this would remove that extra step.

Thanks!

JSMike:
I would prefer something closer to #8967, as the naming conventions vary. But this would be good for setting a global default.

PR #8087, opened 2026-08-24
fix(server): retry automatic thread title generation

Bil0000:
## What Changed

Automatic thread title generation now retries twice with exponential backoff before giving up.

- `maybeGenerateThreadTitleForFirstTurn` wraps `generateThreadTitle` in `Effect.retry` (2 retries, 2s exponential backoff).
- Added a regression test: the first attempt fails with a `TextGenerationError`, the second succeeds, and the thread ends up with the generated title.

No behavior change on the success path, and no change to the prompts, the regeneration path, or any other text generation operation.

## Why

Automatic titles are generated exactly once, on the first turn, and the result is dropped on failure. Before that turn starts, the web and mobile clients already seed the thread with a placeholder title truncated from the first user message. So a single failed generation leaves the thread permanently named after the raw prompt text (`"There are some spacing issues in the too..."`), with only a server-side `logWarning` to show for it.

This is not rare. On a long-running instance, server logs show 32 failed automatic title generations over 14 days:

- 21 x `Claude CLI request timed out.` (the text generation CLI has a 180s ceiling, and a single title call measured 60–80s on a loaded machine, so contention pushes it over)
- 11 x transient CLI failures, all `Failed to authenticate: OAuth session expired and could not be refreshed`

Every one of those threads still carries the truncated first message as its title in `projection_threads`.

Both failure classes are transient, so a bounded retry is the smallest fix that makes the title actually land. The existing guard after generation is unchanged: the generated title is still only applied when the thread title is still the default or still equals the client's `titleSeed`, so a title the user renamed during the backoff is never overwritten.

Known limitation: this does not make title generation faster, and a thread whose generation fails all three times still keeps the placeholder. Reducing the per-call latency and surfacing a persistent failure to the user are separate changes.

## Checklist

> [!NOTE]
> ### Retry thread title generation in `ProviderCommandReactor` up to 2 times
> Wraps the `generateThreadTitle` call site in `ProviderCommandReactor.make` with a retry policy using exponential backoff (2s, then 4s). The title generation now retries up to 2 additional times on transient failures before giving up. The test in [ProviderCommandReactor.test.ts](https://github.com/pingdotgg/t3code/pull/8087/files#diff-24590cda9b5d40bcb845f474304592ddf62ae6d252d40be60e19435aa75347c4) is updated to verify a first-call failure is retried and the second call succeeds.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0c94e6d.</sup>
> 
>


t3dotgg:
> [!NOTE]

This addresses a real but low-impact failure where the first automatic title attempt can be lost. We do not have a current reproduction that justifies keeping this full retry patch in the active queue.

apps/server/src/orchestration/Layers/ProviderCommandReactor.ts
```diff
@@ -94,6 +95,12 @@ const HANDLED_TURN_START_KEY_MAX = 10_000;
 const HANDLED_TURN_START_KEY_TTL = Duration.minutes(30);
 const DEFAULT_RUNTIME_MODE: RuntimeMode = "full-access";
 const MAX_REGENERATION_ATTACHMENTS = 4;
+// Automatic title generation is fire-and-forget: when it fails the thread keeps
+// the placeholder title the client seeded from the first message, and nothing
+// ever tries again. Retry so a transient CLI timeout or auth failure does not
+// permanently leave the thread named after the raw prompt text.
```
Bil0000: Lers remove these comments

Bil0000 (commented):
Remove useless code comments

Bil0000:
Ponytail finding fixed: the retry regression duplicated the existing title-generation integration test. The transient failure now lives in that test, and the one-use retry constants are inlined. Net: -44 lines. Verified with all 49 ProviderCommandReactor tests, server typecheck, lint, and format.

PR #8094, opened 2026-08-24
fix(client-runtime): refresh edited pull request comments

Bil0000:
Updated pull request comments stayed stale because the successful update command did not refresh the shared activity query. Each PR surface could finish the edit while still rendering the old cached activity.

This change refreshes the exact pull request activity atom after a comment update succeeds. Summary, Timeline, and Code views now receive the updated comment text without a page reload. A client-runtime regression test covers the full command-to-query flow.

Verified with:
- `pnpm exec vp test run packages/client-runtime/src/state/pullRequests.test.ts`
- `pnpm exec vp fmt --check packages/client-runtime/src/state/pullRequests.ts packages/client-runtime/src/state/pullRequests.test.ts`
- `pnpm exec vp lint --report-unused-disable-directives packages/client-runtime/src/state/pullRequests.ts packages/client-runtime/src/state/pullRequests.test.ts`
- `pnpm --filter @t3tools/client-runtime typecheck`


Bil0000:
Ponytail review: Lean already. The four-line refresh is at the shared command/query boundary, and the integration test is the smallest proof across all PR views. No safe deletion found. Net: 0 lines. Existing focused tests, typecheck, lint, and format are green.

PR #8050, opened 2026-08-24
feat(server): add Antigravity provider

carterwsmith:
## What Changed

Adds Google Antigravity (`agy`) as a supported provider.

- Supports authenticated, resumable Antigravity chat sessions
- Discovers available models through `agy models`
- Adds provider status, authentication detection, settings, icons, and documentation
- Integrates Antigravity text generation for titles, commits, branches, and PRs

## Why

Google models are not as useless as Theo says

## UI Changes

**Model selector**

**Settings panel**

**Simple demo video**
https://github.com/user-attachments/assets/ef25edc7-289b-4608-b750-94b2bdb307c5

## Checklist

- [ ] This PR is small and focused
- [x] I explained what changed and why
- [x] I included before/after screenshots for any UI changes
- [x] I included a video for animation/interaction changes

---

> [!NOTE]
> **Medium Risk**
> New CLI driver that spawns `agy` (including `--dangerously-skip-permissions` in full-access) and is enabled by default. Follows existing provider patterns but expands process/auth surface.
> 
> **Overview**
> Adds **Google Antigravity** (`agy`) as a built-in provider across server, web, mobile, contracts, and docs.
> 
> The server driver probes `agy --version` / `agy models`, streams NDJSON sessions, resumes via conversation ID, maps plan vs accept-edits and sandbox vs skip-permissions, and supports in-session model/effort switches by restarting the CLI. Text generation (titles, commits, branches, PRs) uses `agy -p --json-schema`.
> 
> Gemini `-low`/`-medium`/`-high` variants are collapsed into one picker row with a **Reasoning** option. Settings (`binaryPath`, `launchArgs`) default **enabled**. UI/icons, provider order, and install docs are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 62b624bb1d92e37a181401e90e6ae01f4a761c47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add Antigravity (`agy`) provider across server, web, and mobile
> - Introduces a full-stack `agy` provider: driver, adapter, text-generation, health-check/model-discovery probe, settings schema, and provider registry entry. The default model is `gemini-3.7-flash-high`.
> - Server-side text generation spawns the `agy` CLI with JSON-schema-constrained I/O and a 180s timeout; commit messages, PR content, branch names, and thread titles are decoded from the CLI's structured output and sanitized.
[...]

PR #8008, opened 2026-08-23
Add Google Antigravity provider support across all clients

nxfx21:
> [!NOTE]
> This PR was built end-to-end as part of a bet on X ([link to post](https://x.com/maria_rcks/status/2091385135035711634?s=20)).
> The entire implementation was written autonomously using **Gemini 3.7 Flash (High)** inside Google Antigravity. It is completely functional and tested across all surfaces.

---

<details>
<summary><b>Click to view full prompt & instructions</b></summary>

```text
Your single task is to add first-class support for Google’s Antigravity CLI (the agy binary) as a new provider in T3 Code.

Requirements & Context:
1. Hook into the official agy binary using headless non-interactive print mode with stream-json (--output-format stream-json, --dangerously-skip-permissions).
2. Implement native provider contracts (@t3tools/contracts), server orchestration (AntigravityAdapter.ts, AntigravityProvider.ts, AntigravityDriver.ts), and text generation (AntigravityTextGeneration.ts) for thread titles and branch names.
3. Handle complete turn lifecycles, streaming NDJSON output, tool calls, tool results, and Fiber synchronization for clean turn completions.
4. Add automated Google Account authentication probing from ~/.gemini/google_accounts.json and ~/.gemini/oauth_creds.json.
5. Create a setup and onboarding process in Settings with platform-specific copyable install commands for macOS/Linux and Windows, plus sign-in guidance.
6. Streamline UI models by removing duplicate variant suffixes (e.g. Gemini 3.6 Flash (High), (Medium), (Low)) in favor of clean canonical models with unified composer reasoning dropdowns.
7. Support multi-surface icons (web & mobile), documentation (docs/user/providers-antigravity.md), and comprehensive vitest test coverage.

Tasks:
Investigate reference VSIX implementations to determine the optimal binary versus client-level architecture, execute and verify the implementation under adversarial rigor (`/crash-out`), diagnose and fix turn execution and stream synchronization failures caused by `--effort` CLI argument conflicts, implement automated Google account authentication probing and interactive setup onboarding in the web settings UI, configure T3 Code runtime awareness and Plan Mode bridging, streamline the model catalog by removing redundant variant clutter in favor of composer-driven reasoning effort dropdowns, and commit, test, and push the fully working integration directly to `nxfx21/t3code-Antigravity` on `main`.
```

</details>

---

## What Changed

Added full native provider integration for Google Antigravity (`agy`):

- **Contracts (`@t3tools/contracts`)**:
[...]

PR #8113, opened 2026-08-24
feat(providers): run Google Antigravity (agy) as a coding agent provider

Marve10s:
T3 Code cannot drive Google's Antigravity CLI (`agy`). Anyone with an Antigravity subscription has to leave T3 Code to use it.

This adds Antigravity as a built-in provider on the v2 orchestration runtime. The adapter runs `agy` in print mode with `stream-json` output and maps its events onto v2 turn items: assistant text streams as it arrives, tool starts, outputs, and completions show up in the work log, and token usage lands on the turn. Follow-up messages reuse the `agy` conversation id, so a thread stays inside one Antigravity project. The provider also acts as a text-generation backend for thread titles, discovers Antigravity skills from the project and user skill directories, and shows up in provider settings, status probing, and the web and mobile icon sets.

Print mode cannot pause for interactive approvals, so the adapter accepts only the full-access runtime mode and fails every other mode with an explicit error rather than silently ignoring the choice. Plan mode maps to `agy --mode plan`.

Based on #2829 (the v2 orchestrator). The diff is against that branch.

Two other Antigravity PRs are open, #8050 and #8008. Both target `main` and plug into the v1 provider stack (`provider/Layers/*Adapter.ts` and `provider/Services/*Adapter.ts`), which #2829 removes when the v2 orchestrator lands, so either of them would need a rewrite right after merging. This PR builds the provider as a v2 adapter (`orchestration-v2/Adapters/AntigravityAdapterV2.ts`) from the start. The parts that look similar across the three PRs are the shared shapes: provider status probing, settings schema, icons, and title generation. The part that differs is the adapter itself and how `agy` turns map onto v2 runs, turn items, and runtime requests.

We tested it end to end and it works for us: threads start, stream, continue, and title themselves through the CLI. Expect follow-up commits while the automated reviewers run; the branch may be broken for short stretches between them.

Preview recording: https://github.com/Marve10s/t3code/releases/download/pr-assets-antigravity/antigravity-preview.mov

Confidence: 85%. The happy path is verified against a real `agy` install and a copy of production data. The remaining risk is in corners of the CLI's stream-json protocol that the fixture-based tests do not cover.

Built by Claude Fable 5 in the Claude Code harness.

### DIFF
--- a/apps/mobile/src/components/ProviderIcon.tsx
+++ b/apps/mobile/src/components/ProviderIcon.tsx
@@ -39,6 +39,17 @@ export function ProviderIcon(props: ProviderIconProps) {
     );
   }
 
+  if (props.provider === "antigravity") {
+    return (
+      <Svg width={size} height={size} viewBox="0 0 24 24" fill="none">
+        <Path
+          fill="#7C3AED"
+          d="M12 1.5 22.5 12 12 22.5 1.5 12 12 1.5Zm0 4.2L5.7 12l2.1 2.1L12 9.9l4.2 4.2 2.1-2.1L12 5.7Zm0 8.4-2.1 2.1L12 18.3l2.1-2.1L12 14.1Z"
+        />
+      </Svg>
+    );
+  }
+
   if (props.provider === "cursor") {
     return (
       <Svg width={size} height={size} viewBox="0 0 466.73 532.09" fill="none">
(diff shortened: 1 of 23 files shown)


Marve10s:
Closing this. With #8050 and #8008 already open there are three Antigravity PRs on the table, and there is no sign the maintainers want an `agy` provider in T3 Code right now, so a third one does not help anyone. The branch stays up on my fork as the v2 adapter version built on #2829, in case that changes.

PR #8121, opened 2026-08-24
fix(server): stop clone URLs being parsed as git options

Lucenx9:
User-provided clone URLs were passed to `git clone` before an end-of-options marker. A URL beginning with `-` could therefore be parsed as a Git option instead of a repository value.

Add `--` before the remote URL and destination so both remain positional arguments, and add focused regression coverage for normal and dash-prefixed clone inputs.

Tests:
- `vp test run apps/server/src/sourceControl/SourceControlRepositoryService.test.ts`
- Targeted lint and format checks for both changed files

Model: `openai/gpt-5.6-sol` via OpenCode

---

> [!NOTE]
> **Medium Risk**
> Changes how clone commands are built for untrusted URLs; scope is narrow but touches security-sensitive shell/git invocation.
> 
> **Overview**
> **Fixes a command-injection class issue** where user-supplied clone URLs passed directly to `git clone` could be interpreted as Git flags when they start with `-` (e.g. `--upload-pack=...`).
> 
> `cloneRepository` now inserts Git’s end-of-options marker (`--`) before the remote URL and destination directory name so both are always treated as positional arguments. Existing clone expectations were updated, and a regression test asserts dash-prefixed URLs are forwarded safely.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 368ecef24032b37ce7b6e2a9c1665931fde17758. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Pass `--` before clone URL and directory in `SourceControlRepositoryService.cloneRepository`
> Inserts an end-of-options marker (`--`) into the git clone args in [SourceControlRepositoryService.ts](https://github.com/pingdotgg/t3code/pull/8121/files#diff-ee6247891e4a01ae6615258135b417c8454fe770eae597f3591b677b32c7aa0b) so dash-prefixed `remoteUrl` or `directoryName` values are treated as positional arguments instead of git options. Adds a test for dash-prefixed remote URLs and updates an existing test to assert the new args.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 368ecef.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved repository cloning reliability for remote URLs and destination paths that begin with a dash.
  - Prevented option-like values from being misinterpreted as command-line flags.
  - Added regression coverage to verify these URLs are preserved correctly during cloning.

### DIFF
--- a/apps/server/src/sourceControl/SourceControlRepositoryService.test.ts
+++ b/apps/server/src/sourceControl/SourceControlRepositoryService.test.ts
@@ -176,9 +176,45 @@ it.effect("clones a looked-up repository into the requested destination", () =>
       assert.deepStrictEqual(cloneCalls, [
         {
           cwd: parent,
-          args: ["clone", CLONE_URLS.url, "t3code"],
+          args: ["clone", "--", CLONE_URLS.url, "t3code"],
         },
       ]);
+    }).pipe(
+      Effect.provide(
+        makeLayer({
+          git: {
+            execute: (input) =>
+              Effect.sync(() => {
+                cloneCalls.push({ cwd: input.cwd, args: input.args });
+                return processOutput();
+              }),
+          },
+        }),
+      ),
+    );
+  }).pipe(Effect.provide(NodeServices.layer)),
+);
+
+it.effect("passes dash-prefixed remote urls as positional arguments, not git options", () =>
+  Effect.gen(function* () {
+    const fs = yield* FileSystem.FileSystem;
+    const parent = yield* fs.makeTempDirectoryScoped({
+      prefix: "t3-source-control-clone-parent-",
+    });
+    const injectionUrl = "--upload-pack=touch /tmp/pwned";
+    const cloneCalls: Array<{ cwd: string; args: ReadonlyArray<string> }> = [];
+
+    yield* Effect.gen(function* () {
+      const service = yield* SourceControlRepositoryService.SourceControlRepositoryService;
+      const result = yield* service.cloneRepository({
+        remoteUrl: injectionUrl,
+        destinationPath: `${parent}/repo`,
+      });
+
+      assert.deepStrictEqual(cloneCalls, [
+        { cwd: parent, args: ["clone", "--", injectionUrl, "repo"] },
+      ]);
+      assert.strictEqual(result.remoteUrl, injectionUrl);
     }).pipe(
       Effect.provide(
         makeLayer({
--- a/apps/server/src/sourceControl/SourceControlRepositoryService.ts
+++ b/apps/server/src/sourceControl/SourceControlRepositoryService.ts
@@ -203,10 +203,12 @@ export const make = Effect.gen(function* () {
       });
     }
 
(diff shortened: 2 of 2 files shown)


Lucenx9:
@coderabbitai review

Lucenx9:
**Review: correct ✅**

The `--` end-of-options marker is the right fix here. Both `remoteUrl` and `preparedDestination.directoryName` sit after the marker, so dash-prefixed values are always treated as positional arguments by git. Covering `directoryName` matters too — it's derived from the user-supplied `destinationPath`, which was equally injectable. Since args are spawned `execFile`-style (no shell), `--` fully closes the option-injection vector.

**Verified:**
- Normal URLs are unaffected (`--` is a no-op for them) — existing test updated accordingly.
- The new regression test asserts the exact args array at the right seam (mocked `GitVcsDriver`), consistent with the file's existing test pattern.
- Ran `vp test run apps/server/src/sourceControl/SourceControlRepositoryService.test.ts` — 8/8 pass.

The PR's tests mock `GitVcsDriver`, so they prove the args are built correctly but never exercise a real `git` binary. Filling that gap locally (git 2.55):

- `git clone -- "$PWD/src.git" dest` → succeeds, `--` is a no-op for normal URLs ✅
- `git clone -- "--upload-pack=touch /tmp/pwned" dest` → `fatal: repository '--upload-pack=touch /tmp/pwned' does not exist`, exit 128, `/tmp/pwned` **not** created — the value stays positional and fails safely ✅
- Control without `--`: `git clone "--upload-pack=..." dest` shifts the positional slots (`dest` is reported as the repository), confirming the option-injection class was real pre-fix ✅

One minor caveat: `--` as end-of-options requires git ≥ 2.24 (Nov 2019). The repo pins no minimum git version, but a ~6-year floor is fine in practice — not worth blocking.

**Findings (non-blocking, same class elsewhere):**
1. [`GitHubCli.ts#L418`](https://github.com/pingdotgg/t3code/blob/main/apps/server/src/sourceControl/GitHubCli.ts#L418) — `["repo", "view", input.repository, ...]` passes the user-supplied `repository` string positionally to `gh` with no `--` guard. A value starting with `-` hits the same option-parsing issue in the `gh` CLI. Worth a follow-up.
2. `GitVcsDriverCore.ensureRemote` (`remote add <name> <url>`) has the same shape, though the URL there is provider-derived rather than raw user input, so the risk is lower.

Neither blocks this PR — the fix is narrow, correct, and tested. Safe to merge.

---
Model: `glm-5.3` · Harness: Pi Agent

Lucenx9:
@codex review

t3dotgg:
> [!NOTE]

[#8290](https://github.com/pingdotgg/t3code/pull/8290) is the active fix for terminating git clone options. It covers the dangerous upload-pack destination case and tests that every positional argument follows the terminator, so this branch has no distinct production behavior to preserve.

PR #5407, opened 2026-08-05
Preserve pull request status across branch checkouts

juliusmarminge:
## What Changed

- Added branch-specific pull request lookup through the VCS workflow and WebSocket RPC layers.
- Preserved thread and branch PR indicators when the working tree checkout changes.
- Serialized local and remote VCS refreshes to prevent stale status from overwriting current checkout data.
- Updated branch selectors, sidebar indicators, chat, and Git actions to resolve PR status from the thread’s recorded branch.
- Simplified terminal font inheritance so terminal surfaces use the configured code font unless explicitly overridden.
- Added coverage for checkout changes, refresh ordering, and remote status recovery.

## Why

PR status was derived from the current checkout, causing thread badges and PR actions to show stale or incorrect information after switching branches. Branch-keyed lookups keep each thread associated with its own PR while preserving fast stream-fed status for the active checkout. Refresh serialization prevents slow remote lookups from publishing results for a previous branch.

The typography cleanup keeps monospace surfaces consistent and removes unused advanced-font preference logic.

## UI Changes

The sidebar, chat, branch selector, and Git actions now continue to display and act on the correct PR after switching checkouts. No visual layout changes were made, so before/after screenshots are not applicable.

## Checklist

- [ ] This PR is small and focused
- [x] I explained what changed and why
- [ ] I included before/after screenshots for any UI changes
- [ ] I included a video for animation/interaction changes

---

> [!NOTE]
> **Medium Risk**
> Touches VCS caching, concurrency, and checkout/switch RPC ordering where stale or mismatched PR/ref data could briefly surface; changes are localized to git status/PR resolution with added test coverage.
> 
> **Overview**
> **PR indicators and thread badges now follow each thread’s recorded branch**, not whatever is checked out in the repo. A new **`vcs.branchPr`** RPC (and cached client query) resolves PR/MR state from git branch config and upstream tracking, reusing the same **`lookupStatusPr`** path as live status.
> 
> Web and mobile replace checkout-only **`resolveThreadPr`** with **`useThreadPr` / `selectBranchPr`**: streamed status wins when `refName` matches the thread branch (including “no PR”), otherwise the branch-keyed query keeps the badge. Sidebar, chat, and the branch toolbar use this; **`resolveBranchToolbarPrBranch`** is removed. Git actions stay gated on **checkout** status only.
> 
[...]

PR #4926, opened 2026-07-30
fix(web): keep thread branch metadata consistent

maxktz:
## What Changed

- save the checked-out branch when creating a thread without touching the branch picker
- consistently omit the default branch from sidebar rows while keeping custom branches visible

## Why

The branch picker displayed the current branch without persisting it. This made equivalent threads look different depending on whether the default branch had been explicitly reselected.

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why

<details>
<summary>Show AI summary</summary>

New draft-thread bootstraps now derive their persisted branch with `resolveBranchToolbarValue`, using the live checkout from Git status while preserving explicit new-worktree base selections. Sidebar V2 resolves the persisted branch through `listRefs`, so default-branch rows keep their existing branch-free presentation whether or not the dropdown was touched. The tooltip continues to render the persisted branch directly.

Focused verification passed: web typecheck, formatting, lint, `BranchToolbar.logic.test.ts`, and `Sidebar.logic.test.ts`.

</details>

---

> [!NOTE]
> **Low Risk**
> UI-only branch display and draft bootstrap metadata; no auth or server contract changes. Extra `listRefs` queries per local branch-backed row are bounded and project-scoped via query caching.
> 
> **Overview**
> Aligns **persisted thread branch** with what the branch toolbar shows, and **sidebar branch labels** with the rule that default branches stay hidden.
> 
> **Chat bootstrap** — Draft thread creation no longer always saves `activeThreadBranch`. It now uses `resolveBranchToolbarValue` (env mode, worktree path, thread branch, live Git ref) so a untouched picker still records the checked-out branch.
> 
> **Sidebar V2** — Card rows use `resolveSidebarBranchLabel` instead of always showing `thread.branch`. Local (non-worktree) threads load `listRefs` (cached per project, `limit: 2`) to detect the default branch; worktrees rely on `gitStatus.isDefaultRef`. The label stays blank while that resolution is pending.
> 
> **Sidebar.logic** — Adds `resolveSidebarBranchLabel` with unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d80ade58ff8d5f7526e416928dbefc62ea58fed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix thread branch metadata consistency in sidebar and new thread creation
[...]

Issue #4831, opened 2026-07-29
[Bug]: Worktree threads lose their PR badge once the agent switches branches inside the worktree


LouisHaftmann:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/web

### Steps to reproduce

1. Start a thread with a dedicated worktree.
2. Let the agent create/switch to its own branch inside that worktree (it usually renames off the generated `t3code/<slug>` branch).
3. Open a PR for that branch.
4. Look at the thread row in the sidebar.

### Expected behavior

The PR number/badge shows on the thread row.

### Actual behavior

No badge. The server finds the PR fine — the client discards it.

#4460 removed the dedicated-worktree bypass in `resolveThreadPr`, so every row now requires `gitStatus.refName === thread.branch`. But nothing ever writes the worktree's new HEAD back to `projection_threads.branch` (no branch-update event exists anywhere in `apps/server/src`), so the stored branch goes stale the moment the agent switches branches — and the badge disappears permanently.

Roughly half of my active worktree threads are in this state. Concrete shape: thread branch stays `t3code/<slug>`, worktree HEAD is `fix/<something-the-agent-named>`, and that branch has an open PR. `gitStatus.pr` contains it; `resolveThreadPr` returns `null`.

Not covered by #4752 / #4755 — that fix retains only **merged/closed** PRs and only for **local** checkouts ("worktree threads clear their snapshot on branch mismatch"). It restores zero badges in this case.

### Impact

Minor bug or occasional failure

### Version or commit

0.0.31-nightly.20260729.942 (regression from #4460, 724887717)

### Environment

Linux, server started via `npx t3@nightly server`, web UI

### Notes

A plain revert of #4460 would reintroduce the stacked-PR mislabel it fixed. The fix that satisfies both is to keep the stored branch in sync with the worktree's live HEAD, so the equality check passes legitimately and the label and PR always describe the same branch. That does mean a worktree parked on a stacked head branch would show that branch's PR — arguably correct, but a maintainer call.

<details>
<summary>Agent prompt to fix</summary>

```
In pingdotgg/t3code: a thread's stored branch (`projection_threads.branch`) goes stale
when an agent switches branches inside the thread's dedicated worktree. Since #4460
(commit 724887717), `resolveThreadPr` in apps/web/src/components/ThreadStatusIndicators.tsx
[...]

PR #8126, opened 2026-08-24
fix(web): sidebar follows a worktree thread's live branch

### LINKED ISSUE #7409: [Bug]: Branch name in session screen is not reactive
### Before submitting

### Area

apps/desktop

### Steps to reproduce

1. Start a new chat
2. Open terminal panel (Cmd+J)
3. Create a new git branch from the terminal
4. Observe branch name displayed by the UI

### Expected behavior

UI displays the newly created branch name right away

### Actual behavior

UI displays the stale branch name

### Impact

### Version or commit

0.0.33 (3b72d17cbca6)

### Environment

macOS 26.6.1

### Logs or stack traces

```shell

```

[Screen Recording 2026-08-18 at 16.53.43.mov](https://github.com/user-attachments/assets/9128a88f-d239-43a3-981a-79fb5c725766)

### Workaround

Switching to another window and then back to T3 Code helps (I did that at the end of the video)

sethwebster:
## What Changed

Sidebar rows resolve their branch label through a new `resolveSidebarThreadBranchLabel`, which adopts the live `gitStatus.refName` for **worktree** threads and falls back to the stamped `thread.branch` on detached HEAD. Local-checkout rows are unchanged.

Applied to the three places that previously read `thread.branch` directly: the thread row label, the hover tooltip, and the search-result row (which shares that tooltip).

+91 / −4 across three files, one of them the test.

## Why

> **Correction (under revision).** An earlier version of this description claimed nothing reconciles `thread.branch` against git. That is wrong. `followWorktreeBranchDrift` (`apps/server/src/orchestration/Layers/CheckpointReactor.ts:566`) already adopts the checked-out branch at turn end, skipping detached HEAD, temporary branches and shared worktrees. See Deviations — this PR is being reworked and should not be reviewed yet.

Sidebar rows render the persisted `thread.branch`. The thread view resolves the same value through `resolveBranchToolbarValue`, which prefers the live `gitStatus.refName`. Because the server only follows drift at turn completion, the two disagree in the window between a mid-turn `git checkout` and the end of that turn, and indefinitely for a checkout never followed by a turn.

So the moment a worktree's checked-out branch diverges from the stamp, the two views disagree: the toolbar follows reality while the sidebar keeps showing a branch the thread has already left.

A worktree belongs to exactly one thread, so whichever ref is checked out there **is** that thread's branch. The shared local checkout is the opposite — its ref belongs to whichever thread last switched it — so those rows keep the branch of record and let the existing mismatch warning explain the difference.

The rows already subscribe to `vcsEnvironment.status` for that warning (`Sidebar.tsx:792` and `:1622`), so reading the live ref costs no additional query.

## Scope and non-goals

In scope: the web sidebar's branch **display** for worktree threads.

Explicit non-goals:

- **Does not persist anything.** `thread.branch` stays stale in `projection_threads`, so mobile's thread list (`apps/mobile/src/features/threads/thread-list-items.tsx:465`) and PR matching still read the old value. Fixing that needs a server-side branch-update event, which #4831 noted does not exist.
- **Does not make anything refresh faster.** There is no `.git` watcher in `apps/server/src`; local status is only re-read on an in-app git action, an explicit `vcsRefreshStatus`, or the end-of-turn `CheckpointReactor`. The 30s loop refreshes remote status only. This PR makes the sidebar *agree* with the thread view, not update sooner — so it does **not** fix #7409, whose repro is a bare `git checkout` in the terminal panel.
- **Does not touch local-checkout rows.**

## Related work

- #4831 (closed) described this exact root cause: nothing writes a worktree's new HEAD back to `projection_threads.branch`.
- #5407 (open) covers overlapping ground in these files. If that supersedes this, close this one — it is deliberately small enough to be cheap to discard.
- #4926 (open) touches `Sidebar.logic.ts` for a different problem and predates the `SidebarV2.tsx` → `Sidebar.tsx` rename.

## Decisions

- **Display-only rather than server-side reconciliation.** Persisting the live ref would fix every surface at once, but it needs a new event and a rule for the shared-checkout case. Kept small on purpose, per CONTRIBUTING.
- **Worktree threads only.** Adopting live `refName` on the shared local checkout would relabel every row in a project whenever one thread switched branches.
- **Fall back to the stamp, never to nothing.** Detached HEAD reports a null ref; the row keeps its last known branch instead of going blank.

## Implementation checklist

- [x] Add `resolveSidebarThreadBranchLabel` with worktree/local split and detached-HEAD fallback
- [x] Use it for the thread row label, the tooltip, and the search-result row
- [x] Unit tests for all five branches of the resolution
- [x] Before/after screenshots

## Verification

- [x] `vp test run apps/web/src/components/Sidebar.logic.test.ts` — 110 passed, 5 new
- [x] `tsgo --noEmit` on `apps/web` — clean
- [x] `vp lint --report-unused-disable-directives` on the three changed files — clean
- [x] `vp fmt --check` on the three changed files — clean
- [x] Visual confirmation in a running client — isolated dev server against a seeded fixture; server logs confirm the status subscription resolving `branch: agent/renamed-during-run` for the worktree thread and `branch: main` for the local one

## UI Changes

A worktree thread stamped `t3code/rework-branch-toolbar` whose worktree was switched to `agent/renamed-during-run`, next to a local-checkout thread on `main` as a control. Same fixture, same viewport; only the build differs.

| Before | After |
| --- | --- |

Before, the worktree row shows the stale stamp. After, it shows the branch the worktree is actually on. The local-checkout row reads `main` in both, confirming shared checkouts are untouched.

## Security and compatibility

No contract, schema, or server change; no new network calls; no auth or trust-boundary surface. Purely which of two already-present client values is rendered. Fully backward compatible — a client without this change simply keeps showing the stamped branch.

## Deviations

Adversarial review found three defects in this change as written. Recorded here rather than silently amended:

1. **False premise.** `resolveSidebarThreadBranchLabel`'s comment claims a worktree belongs to exactly one thread. It does not — `new-thread-on-branch` (`Sidebar.tsx:3109`) carries `worktreePath` into a second thread and `useHandleNewThread.ts:403` persists it. The server guards this with an explicit `worktreeIsShared` check; this change has no equivalent, so two threads sharing a worktree would both be relabeled from one thread's checkout.
2. **Missing temporary-branch guard.** The server skips `isTemporaryWorktreeBranch` (`packages/shared/src/git.ts:107`) so it does not race the first-turn auto-rename. This change adopts any non-null ref, so a row can briefly render a `t3/<hex>` placeholder.
3. **Row incoherence.** Only the label follows the live ref. The PR badge, worktree tooltip, `copy-branch`, and `new-thread-on-branch` still key off `thread.branch`. Worse, `resolveDisplayedThreadPr` requires `refName === thread.branch`, so a drifted row now prints branch B while hiding branch B's PR badge.

Scope is being reconsidered in light of the existing server-side drift-follow.

## Checklist

- [x] I included a video for animation/interaction changes — n/a, no motion

Made with Claude Opus 5 in the Claude Code harness.

> [!NOTE]
> ### Fix sidebar to show a worktree thread's live branch instead of stamped branch
> - Adds `resolveSidebarThreadBranchLabel` in [Sidebar.logic.ts](https://github.com/pingdotgg/t3code/pull/8126/files#diff-529ae8997a33774d7ec3514d7abdb655942eaa993f71e6ec6728c892285d251a) to compute the displayed branch label. For dedicated worktrees it uses the current git ref (`gitStatus.refName`) when available, falling back to the stamped `thread.branch`; for shared local checkouts it uses `thread.branch`.
> - Updates [Sidebar.tsx](https://github.com/pingdotgg/t3code/pull/8126/files#diff-12c9f3c619e6ab8dda238e965323ac96e62f3084e3adfb0671b1a6459aa61ba1) so `SidebarThreadRow`, `SidebarSearchResultRow`, and `SidebarThreadTooltip` render the computed `branchLabel` instead of always using `thread.branch`. The tooltip hides the branch row when the label is null.
> - Adds tests in [Sidebar.logic.test.ts](https://github.com/pingdotgg/t3code/pull/8126/files#diff-32549f6f09ae3a1d9d846a62727c17b9f9ec246030982995fd972b497a87335a) covering worktree with switched branch, detached HEAD, no stamped branch, shared checkout with differing ref, and no branch.
> - Risk: threads in a dedicated worktree whose live ref differs from the stamped branch now show the live ref; any code expecting the sidebar to always display `thread.branch` will see the new value.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7dca192.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/Sidebar.logic.test.ts
+++ b/apps/web/src/components/Sidebar.logic.test.ts
@@ -19,6 +19,7 @@ import {
   orderItemsByPreferredIds,
   resolveProjectStatusIndicator,
   resolveSidebarStageBadgeLabel,
+  resolveSidebarThreadBranchLabel,
   resolveThreadRowClassName,
   resolveSidebarThreadStatus,
   resolveThreadStatusPill,
@@ -690,6 +691,60 @@ describe("isContextMenuPointerDown", () => {
         isMac: false,
       }),
     ).toBe(false);
+  });
+});
+
+describe("resolveSidebarThreadBranchLabel", () => {
+  it("follows the worktree's live ref when the agent switched branches inside it", () => {
+    expect(
+      resolveSidebarThreadBranchLabel({
+        worktreePath: "/tmp/wt/thread-1",
+        threadBranch: "seth/original",
+        currentGitBranch: "seth/switched",
+      }),
+    ).toBe("seth/switched");
+  });
+
+  it("keeps the stamped branch for a worktree in detached HEAD", () => {
+    expect(
+      resolveSidebarThreadBranchLabel({
+        worktreePath: "/tmp/wt/thread-1",
+        threadBranch: "seth/original",
+        currentGitBranch: null,
+      }),
+    ).toBe("seth/original");
+  });
+
+  it("labels a worktree that was never stamped with its live ref", () => {
+    expect(
+      resolveSidebarThreadBranchLabel({
+        worktreePath: "/tmp/wt/thread-1",
+        threadBranch: null,
+        currentGitBranch: "seth/switched",
+      }),
+    ).toBe("seth/switched");
+  });
+
+  it("keeps the branch of record on the shared local checkout", () => {
+    // The checkout's ref belongs to whichever thread last switched it, so
+    // adopting it here would relabel every row in the project at once.
+    expect(
+      resolveSidebarThreadBranchLabel({
+        worktreePath: null,
+        threadBranch: "seth/original",
+        currentGitBranch: "someone-elses-branch",
+      }),
+    ).toBe("seth/original");
+  });
+
+  it("returns null when a local thread has no branch at all", () => {
+    expect(
+      resolveSidebarThreadBranchLabel({
+        worktreePath: null,
+        threadBranch: null,
+        currentGitBranch: null,
+      }),
+    ).toBeNull();
   });
 });
 
--- a/apps/web/src/components/Sidebar.logic.ts
+++ b/apps/web/src/components/Sidebar.logic.ts
@@ -459,6 +459,23 @@ export function resolveThreadRowClassName(input: {
   );
 }
 
+// A worktree belongs to exactly one thread, so whichever ref is checked out
+// there is that thread's branch — an agent switching branches mid-run makes the
+// stamped `branch` stale, it does not make the checkout wrong. The shared local
+// checkout is the opposite: its ref is whatever any thread last switched it to,
+// so those rows keep the branch of record and let the mismatch warning explain
+// the difference. Detached HEAD reports no ref, so the stamp stays the fallback.
+export function resolveSidebarThreadBranchLabel(input: {
+  worktreePath: string | null;
+  threadBranch: string | null;
+  currentGitBranch: string | null;
+}): string | null {
+  if (input.worktreePath === null) {
+    return input.threadBranch;
+  }
+  return input.currentGitBranch ?? input.threadBranch;
+}
+
 // ── Sidebar thread status model ─────────────────────────────────────
 // Five visual states, three colors: color is reserved for "act now"
 // (approval), "in motion" (working), and "broken" (failed). Ready is the
--- a/apps/web/src/components/Sidebar.tsx
+++ b/apps/web/src/components/Sidebar.tsx
@@ -134,6 +134,7 @@ import {
   planPinnedReorder,
   resolveAdjacentThreadId,
   resolveSettledTimestamp,
+  resolveSidebarThreadBranchLabel,
   resolveSidebarThreadStatus,
   searchSidebarThreadsByTitle,
   shouldCreateNewThreadInCurrentProject,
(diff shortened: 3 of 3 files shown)


sethwebster:
Closing this myself — it is redundant, and the existing fix is better than this one.

`followWorktreeBranchDrift` (`apps/server/src/orchestration/Layers/CheckpointReactor.ts:566`) already adopts a worktree's checked-out branch at turn completion, guarding detached HEAD, temporary `t3/<hex>` branches, and shared worktrees, with a compare-and-swap on the recorded branch. Because it repairs `thread.branch` itself, it fixes the sidebar, the mobile thread list, the command palette, and PR attribution together.

This PR only re-rendered one of those, and did so without the shared-worktree or temporary-branch guards. It also made the row less coherent: `resolveDisplayedThreadPr` requires `refName === thread.branch`, so a drifted row would have printed branch B while hiding branch B's PR badge.

The only real residual is that drift-follow runs at turn completion, so a checkout made mid-turn — or one never followed by a turn — stays stale until the next turn ends. That is a much narrower problem than this PR describes and, if it is worth addressing, belongs where the existing logic and its guards already live rather than in a second copy of the rule in the renderer.

Recording it here in case it is useful to whoever picks up #5407.

PR #8186, opened 2026-08-25
feat(server): add `t3 thread new` CLI command

sgottschlich:
## What

`t3 thread new "<prompt>"` starts a new thread — including its first turn — on a running T3 Code server from the terminal (scripts, cron, another shell).

## How

- Discovers the live server the same way `t3 pair` does and mints a short-lived session (`AuthStandardClientScopes`, label `t3 thread cli`) from the local credential store.
- Drives the existing HTTP orchestration dispatch route: `thread.create` followed by `thread.turn.start`. The HTTP route has no WebSocket-style bootstrap handling, so the command sequences the pair itself and deletes the thread again when the first turn is rejected.
- No offline fallback on purpose: the first turn executes inside the live server process; without one the command fails and points at `npx t3 serve`.
- Flags: `--project` (id or workspace root, defaults to the cwd's project), `--title` (fixed; default lets the server generate one), `--runtime-mode`, `--base-dir`. Model defaults to the project's default selection.
- Reuses the project CLI's live-server helpers (now exported, dispatch generalized to any client orchestration command with an optional timeout).

## Limits

Thread runs in the project checkout — no worktree preparation, setup script, or attachments (app-only for now). Documented in `docs/user/cli-threads.md`.

## Testing

- New unit tests for title resolution and the command pair (schema-validated), `apps/server` typecheck and lint clean, `bin.test.ts` green.
- Manually verified `--help` and the no-running-server error path.

---

> [!NOTE]
> **Medium Risk**
> Touches orchestration dispatch, thread deletion, and cross-platform process killing/attribution; mistakes could leave orphan processes or kill the wrong PID, though ownership checks and extensive tests mitigate this.
> 
> **Overview**
> Adds **`t3 thread new`** so a running server can create a thread and first turn over HTTP (`thread.create` then `thread.turn.start`, with rollback via `thread.delete` if the turn fails). It reuses exported project-CLI live-server helpers and a longer dispatch timeout; there is no offline path.
> 
> **Process and preview cleanup** introduces **`AgentSessionRegistry`** (per-thread provider session root PIDs from Claude/Cursor/Grok adapters) and expands **`PortDiscovery`** to return servers plus thread-owned processes, attribute listeners to agent sessions, and **`killOwnedProcess`** / **`killThreadAgentTree`** (with SIGTERM→SIGKILL escalation). Thread deletion runs agent-tree kill before stopping the provider; terminal close can **`killSubprocesses`**. RPC maps **`previewKillDiscoveredServer`** to terminal operate scope.
> 
> **Provider rate limits** normalize Claude/Codex payloads into **`ServerProvider.usage`**, track live updates via **`ProviderUsageTracker`**, merge through the registry overlay, and optionally probe Codex at status check time (usage is not persisted in the status cache).
> 
> **Project `color`** flows through decider/projector, SQLite migration **038**, and projection read paths. **`makeWorktreeBootstrapLocks`** serializes worktree prep per repository. **External launcher** supports **`reveal`** for file-manager on Windows/macOS/Linux.
> 
> Also adds **prom.codes** agent memory and saver rules to **`.cursor/rules`**, **`AGENTS.md`**, and **`CLAUDE.md`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 44ff9739c0614979c45731b75a60e0dc8259d459. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add `t3 thread new` CLI command, bulk thread launching, and process lifecycle management
> - Adds a new `thread` CLI subcommand ([thread.ts](https://github.com/pingdotgg/t3code/pull/8186/files#diff-70db64f18282ccd6b5da3fe4928f1e71e74f14a08990a88f8f09d2e662620b67)) that resolves titles, validates prompts, acquires a short-lived auth session, and dispatches thread creation commands to a running server with a 10s timeout.
> - Adds bulk thread launching across the web app and CLI: placeholder parsing/expansion ([bulkPrompt.ts](https://github.com/pingdotgg/t3code/pull/8186/files#diff-a2077bf0d16f08258b73d8683e081e25542fe558f7f17b1927d8581593accd7e)), a `BulkLaunchDialog` UI, a sequential launch queue with cancellation ([bulkLaunchQueue.ts](https://github.com/pingdotgg/t3code/pull/8186/files#diff-7cf066b9ed99e238524ef966e62d60357754b6a5313b9aaf7b83c190e3dc9d8b)), and composer draft `bulk` mode that forces worktree environments.
> - Introduces `AgentSessionRegistry` ([AgentSessionRegistry.ts](https://github.com/pingdotgg/t3code/pull/8186/files#diff-49fe009e1bb8d4d20d6e497cced60145215d5ee9a1599510f19129a71057f599)) to track per-thread agent root PIDs. Claude, Cursor, and Grok adapters register/unregister spawned PIDs; the port scanner attributes discovered servers to agent threads and can kill owned processes or the entire agent tree.
> - Deleting a thread now kills its agent process tree and terminal subprocess descendants before tearing down the session ([ThreadDeletionReactor.ts](https://github.com/pingdotgg/t3code/pull/8186/files#diff-4eec759d283dd7375b01762a54815cd39fb72d86f2440dc7cb030fae567df67d)).
> - Adds live provider usage tracking: rate-limit events from Claude and Codex are normalized and projected into provider snapshots via `ProviderUsageTrackerLive` and a usage overlay in `ProviderRegistry`.
> - Adds Sidebar v3 (beta toggle), project accent colors (persisted, migration 038), message queueing when an agent is busy, reveal-in-file-manager actions, and worktree bootstrap serialization.
> - Risk: `PortDiscovery.scan` now returns a `PortScanSnapshot` (`{servers, processes}`) instead of a raw array; all subscribers and mocks are updated in-tree. `DiscoveredLocalServerList` gains an optional `processes` field. Closing terminals with `killSubprocesses: true` sends SIGTERM then SIGKILL to descendant processes.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 44ff973. 92 files reviewed, 84 issues evaluated, 62 issues filtered, 15 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> <details>
> <summary>apps/server/src/git/worktreeBootstrapLock.ts — 1 comment posted, 2 evaluated, 1 filtered</summary>
>
> - [line 14](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/git/worktreeBootstrapLock.ts#L14): The closure-local `locks` map never removes entries after a worktree preparation finishes. A long-lived connection that submits preparations for many distinct `projectCwd` strings (including paths whose effects fail) leaves one semaphore per string retained for the connection's lifetime, allowing unbounded memory growth instead of reclaiming idle repository locks. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/preview/PortScanner.ts — 5 comments posted, 11 evaluated, 4 filtered</summary>
>
> - [line 67](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/preview/PortScanner.ts#L67): The new kill contract's SIGTERM→SIGKILL escalation retains only the numeric `pid` across the two-second delay. `process.kill` signals whichever process currently has that PID, so if the target exits after SIGTERM and the OS reuses its PID before escalation, the background SIGKILL can terminate an unrelated process. The same numeric-PID reuse window applies to the thread-tree cleanup, making this a real wrong-process kill risk rather than merely a stale UI entry. <b>[ Cross-file consolidated ]</b>
> - [line 673](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/preview/PortScanner.ts#L673): The terminal branch of `killOwnedProcess` trusts the cached `registration.processIds` without rechecking that the PID is still a descendant of the terminal (or even still the same process). A terminal subprocess can exit and its PID can be reused before `pollSubprocessActivity` removes the registration; a stale UI/request for that PID then passes this `some(...)` check and `process.kill` sends `SIGTERM`/later `SIGKILL` to the unrelated replacement process. The agent path does a fresh process-table ancestry check, but the terminal path lacks the same protection. <b>[ Cross-file consolidated ]</b>
> - [line 697](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/preview/PortScanner.ts#L697): The delayed escalation reuses only the numeric PID after the initial `SIGTERM`: after the two-second sleep it unconditionally calls `process.kill(input.pid, "SIGKILL")`. If the owned process exits and the OS reuses that PID during the grace period, this kills the unrelated replacement process. The second signal needs a fresh identity/ownership check rather than relying on the original numeric PID. <b>[ Cross-file consolidated ]</b>
> - [line 733](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/preview/PortScanner.ts#L733): `killThreadAgentTree` also schedules a numeric-PID-only `SIGKILL` after the two-second grace period. A descendant can exit after receiving `SIGTERM` and its PID can be reused before this loop runs; the cleanup then kills the unrelated replacement process. The escalation must revalidate process identity/ancestry (or use an identity-safe handle) before each `SIGKILL`. <b>[ Already posted ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/process/externalLauncher.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 238](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/process/externalLauncher.ts#L238): `posixParentDirectory` returns the original target when it has no slash, e.g. `posixParentDirectory("README.md")` is `"README.md"` rather than `"."`. Consequently Linux reveal requests for a relative root-level path pass that path to `xdg-open` and open the entry itself instead of opening its containing directory, so the new reveal behavior is wrong for valid relative `cwd` values accepted by `LaunchEditorInput`. Return `.` for a relative single-component target. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/provider/Layers/ProviderRegistry.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 679](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/provider/Layers/ProviderRegistry.ts#L679): `usageOverlayRef` is cleaned only for instance IDs that disappear, so a rebuilt instance with the same `instanceId` retains the old account's rate-limit snapshot. `ProviderInstanceRegistry.reconcile` rebuilds an instance when its config changes (for example, switching a Codex `homePath` or credentials), but `syncLiveSources` still considers that ID known and leaves the overlay intact. The new instance's first snapshot is then decorated with stale windows until a fresh rate-limit event arrives, and if none arrives the UI reports the previous account's limits indefinitely. <b>[ Cross-file consolidated ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/provider/providerUsage.ts — 1 comment posted, 2 evaluated, 1 filtered</summary>
>
> - [line 248](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/provider/providerUsage.ts#L248): `pickNewestUsage` drops the right-hand snapshot when both `updatedAt` values are equal. `ProviderUsageTracker` stamps each event with millisecond-resolution `DateTime.now`, so two valid rate-limit events processed within one millisecond can produce equal timestamps; the second event is written to the overlay but `applyProviderUsage` then selects the older provider snapshot, leaving the UI with stale utilization until a later event has a different timestamp. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/terminal/Manager.ts — 0 comments posted, 3 evaluated, 3 filtered</summary>
>
> - [line 615](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/terminal/Manager.ts#L615): `trySignal` sends signals using only numeric PIDs captured by the earlier subprocess inspection. `closeSession` starts this killer asynchronously, so a descendant can exit and the OS can reuse its PID before either signal; `process.kill` then succeeds against the unrelated process, and the delayed `SIGKILL` can terminate it as well. The cleanup needs to retain/validate process identity (or use an identity-preserving tree/job mechanism) rather than treating a stale PID as ownership. <b>[ Exceeded comment limit ]</b>
> - [line 1407](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/terminal/Manager.ts#L1407): `runSubprocessTreeKill` retains the numeric PIDs that accepted `SIGTERM` and, after `processKillGraceMs`, sends `SIGKILL` to those same numbers without re-inspecting or otherwise pinning the processes. A descendant can exit during the grace period and its PID can be reused by an unrelated process; `process.kill` is only a signal sender, so the delayed call can terminate that unrelated process. The close-time inspection therefore does not provide the ownership guarantee claimed by this cleanup path. <b>[ Exceeded comment limit ]</b>
> - [line 2073](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/terminal/Manager.ts#L2073): The descendant list is a one-time snapshot taken before `stopProcess`, while the terminal shell remains alive during the inspection and can continue spawning children. Any child/grandchild created after `subprocessInspector` returns is absent from `descendants`; killing the shell itself does not kill such descendants, so a terminal command that forks during this window can leave an orphan dev server or worker running after `close({ killSubprocesses: true })` completes. <b>[ Exceeded comment limit ]</b>
> </details>
>
> <details>
> <summary>apps/server/src/ws.ts — 0 comments posted, 5 evaluated, 5 filtered</summary>
>
> - [line 752](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/ws.ts#L752): `makeWorktreeBootstrapLocks()` is instantiated inside `makeWsRpcLayer`, but that layer is built separately for each WebSocket connection in the `/ws` route. Consequently, two clients (or two simultaneous connections) each get an independent lock map and can still run `git fetch`/`git worktree add` concurrently against the same repository, defeating the new serialization and retaining the repository-lock failures the change is meant to prevent. Move the lock factory to a server/layer-wide lifetime (or otherwise share it across connections). <b>[ Cross-file consolidated ]</b>
> - [line 914](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/ws.ts#L914): The semaphore key is the raw `prepareWorktree.projectCwd` string, not the repository's canonical path. `thread.turn.start` normalization leaves this field unchanged, and the schema accepts arbitrary non-empty paths, so the same checkout can be addressed as `/repo`, `/repo/`, or through a symlink. Concurrent bootstraps using such aliases get different permits and can still race `git worktree add`/fetch. Key the lock by a canonical repository identity/path before entering the critical section. <b>[ Exceeded comment limit ]</b>
> - [line 914](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/ws.ts#L914): The new lock only wraps the `bootstrap.prepareWorktree` fetch/create sequence (lines 914–947), while the existing `vcsCreateWorktree` RPC still calls `gitWorkflow.createWorktree(input)` directly at lines 1854–1858. A bootstrap and a user-initiated worktree creation for the same repository can therefore overlap and contend on Git's repository lock, causing one operation to fail despite this change's serialization helper. The same shared lock must cover all worktree-creation entry points, not just bootstrap requests. <b>[ Already posted ]</b>
> - [line 1984](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/ws.ts#L1984): The new handler forwards arbitrary owned PIDs to `killOwnedProcess`, but that service's agent check treats a registered session root as owned because `findAgentOwner` checks the input PID itself before walking to parents. Thus a caller who supplies the provider session root PID can terminate the root, even though the process-tree contract explicitly excludes the root (the provider stop owns it) and the process collection otherwise exposes only descendants. This can abruptly kill the active agent session rather than just a discovered child server. <b>[ Cross-file consolidated ]</b>
> - [line 1984](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/server/src/ws.ts#L1984): This new RPC exposes `killOwnedProcess` using only the client-supplied PID and thread ID. The terminal ownership check in `PortDiscovery` is a bare set of previously observed PIDs; it does not revalidate the process start time/identity before signaling. If a terminal child exits and the OS reuses its PID for an unrelated process, a later `previewKillDiscoveredServer` request can pass the stale registration and send `SIGTERM` (and delayed `SIGKILL`) to that unrelated process. The handler should only expose an identity-checked kill operation (or the registry must retain and verify process identity). <b>[ Exceeded comment limit ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/BranchToolbar.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 466](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/BranchToolbar.tsx#L466): The new bulk toggle is rendered alongside the normal workspace selector, but it does not keep that selector locked to `worktree` while `bulk` is true. After enabling Bulk, a user can choose `Current checkout`; `ChatView` then builds bulk launch specs with `worktree: null`, so multiple launched agents edit the same checkout instead of isolated worktrees, causing conflicting or overwritten changes. Bulk mode should either disable the local option or force the launch mode back to `worktree`. <b>[ Exceeded comment limit ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/ChatView.tsx — 0 comments posted, 14 evaluated, 13 filtered</summary>
>
> - [line 496](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L496): When a queued prompt contains a terminal context together with a preview annotation or review comment, this composition appends the terminal block before the later block. `UserTimelineRow` first tries to strip trailing terminal/element blocks before removing trailing preview annotations, so the terminal block is no longer trailing and is rendered as raw `<terminal_context>` markup in the user message instead of as a context chip. Review comments make the same failure worse because their segment renderer receives the unstripped context markup. Queueing such a combined-context message therefore produces an incorrectly rendered transcript after it is flushed. <b>[ Exceeded comment limit ]</b>
> - [line 1573](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L1573): The new per-thread queue is never cleared when a thread is deleted. `useThreadMessageQueue` reads a global in-memory store, while all existing thread-delete paths remove the server thread but do not call `clearThreadQueue`; queued entries (including their `File` objects and blob preview URLs) therefore become inaccessible orphaned state. Repeating this with image-bearing queues can retain substantial browser memory for the whole session, and the queued prompts are silently lost. <b>[ Exceeded comment limit ]</b>
> - [line 4841](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L4841): `handleBulkLaunchConfirm` trusts `outcome.started` to decide whether to clear the bulk draft and report success, but the launch hook treats an interrupted `startThreadTurn` failure as a normal resolved launch (it only throws non-interrupted failures). If the first-turn command is interrupted by a connection loss or command cancellation, that row is counted as started even though no successful launch was established; the draft is then cleared and the UI can claim threads started, losing the user's prompt values. <b>[ Exceeded comment limit ]</b>
> - [line 5142](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L5142): Queued image attachments share the draft's blob URLs, but the enqueue path immediately calls `clearComposerDraftContent`, whose implementation revokes every current image preview URL. The queue retains those same URLs and `ComposerQueueList`/the optimistic row later renders them, so queued images show broken previews (and a failed send is requeued with already-revoked previews), despite the queue store's contract that taken messages keep previews alive. <b>[ Failed validation ]</b>
> - [line 5421](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L5421): Manual queue dispatch also bypasses the actionable-plan gate. `canFlushQueuedMessage` correctly blocks automatic flushing when `activeProposedPlan` is actionable, but the queue row remains enabled and `sendQueuedMessage` does not check that state. Clicking “Send now” while the plan follow-up UI is waiting starts the queued text as an ordinary new turn (in the current plan mode), bypassing the required refine/implement decision and potentially creating another plan instead. <b>[ Exceeded comment limit ]</b>
> - [line 5453](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L5453): `sendQueuedMessage` updates the shared `optimisticUserMessages` state after awaits without verifying that the route is still the thread it captured. If a user navigates from thread A while A's queued send is reading attachments or dispatching, the completion can append A's optimistic message (or remove only A's row) into the now-visible thread B; the thread-change cleanup effect may already have run, leaving A's message displayed in B until another navigation. <b>[ Exceeded comment limit ]</b>
> - [line 5481](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L5481): Queued sends also bypass the branch-mismatch update performed by the normal send path. If the local checkout changes branches while a turn is running, a message queued during that turn is auto-flushed after it finishes, but `sendQueuedMessage` neither updates the thread's branch nor passes the current checkout branch before `thread.turn.start`. The queued turn can therefore continue against stale thread branch metadata (and the UI's “sending will continue on …” branch guarantee is violated), risking edits in the wrong checkout/branch. <b>[ Exceeded comment limit ]</b>
> - [line 5491](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L5491): `sendQueuedMessage` dispatches `thread.turn.start` directly with the current `runtimeMode`/`interactionMode` (and selected model) but never calls `persistThreadSettingsForNextTurn`, unlike the normal send path. The server decider derives runtime and interaction mode for `thread.turn-start-requested` from the thread's persisted metadata, not these command fields, so a user who changes modes while a message is queued will have the queued turn run with the old mode and the model selection will not be persisted. The queued message therefore silently ignores the controls shown when it is flushed. <b>[ Failed validation ]</b>
> - [line 5552](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L5552): The automatic queue flush does not require a provider to be available. The normal `onSend` path explicitly checks `sendCtx.providerAvailable` before dispatching, but `canFlushQueue` omits that condition and `sendQueuedMessage` dispatches whatever selection `getSendContext()` returns. If the selected provider becomes disabled/unavailable while the current turn runs, the queue immediately attempts the turn, fails, and is halted/requeued instead of waiting for the provider to become usable. <b>[ Already posted ]</b>
> - [line 5552](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L5552): The automatic queue flush does not check `activeEnvironmentUnavailable`. `canFlushQueue` can be true whenever the cached thread session is `ready`, even while the environment connection is reconnecting or unavailable (the existing send path explicitly rejects this state). The effect then removes the queued message and calls `thread.turn.start`, which fails during the outage, requeues it and permanently halts automatic flushing until the user manually acts, instead of waiting for reconnection. <b>[ Exceeded comment limit ]</b>
> - [line 5552](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L5552): `canFlushQueue` considers only the session phase and pending turn/approval state, not `activeThreadShell.backgroundLiveness`. ChatView explicitly supports subagent/watch work that outlives a settled turn (`activeBackgroundLiveness` is shown when `isWorking` is false), so the session can be `ready` while agents are still editing or monitoring. The new effect will auto-dispatch the queued prompt in that window, allowing concurrent work on the same thread instead of waiting for the background work to settle. <b>[ Already posted ]</b>
> - [line 5552](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L5552): The queue flush ignores `isRevertingCheckpoint`. During an asynchronous checkpoint revert, `isWorking` is true in ChatView but the session can still be `ready` with no active turn, satisfying `canFlushQueue`; the effect can dispatch a queued turn while the revert is deleting/rebuilding thread state. That races a new message against the revert and can leave the queue/message history inconsistent. <b>[ Already posted ]</b>
> - [line 5552](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ChatView.tsx#L5552): Automatic flushing also ignores the in-flight branch restore state. While `handleSwitchCheckoutToThread` is awaiting checkout and thread metadata updates, `isRestoringThreadBranch` is true but the cached session may remain `ready`; a queued message can therefore dispatch concurrently against the old checkout/branch. This defeats the restore operation and can run edits in the wrong branch. <b>[ Already posted ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/ProcessesPanel.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 77](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ProcessesPanel.tsx#L77): Both stop handlers return immediately for `isAtomCommandInterrupted(result)` without clearing `stopping` (the analogous path is also in `ThreadProcessRow` at line 144). If the command is interrupted while the component remains mounted—for example when the environment connection is restarted—the process row stays permanently disabled even though the kill request was cancelled or its outcome is unknown, and the user has no way to retry without remounting the panel. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/ProjectFavicon.tsx — 0 comments posted, 2 evaluated, 2 filtered</summary>
>
> - [line 49](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ProjectFavicon.tsx#L49): Passing `accentColor` into `ProjectFaviconImage` here exposes the fallback path used by the V2 tooltip: that caller supplies `stroke-muted-foreground`, which sets the SVG `stroke` property. Since Lucide icons use `stroke="currentColor"` and CSS `stroke` overrides the presentation attribute, the inline color added by `ProjectFaviconFallback` is ignored and the configured accent is not visible when the favicon falls back. <b>[ Already posted ]</b>
> - [line 67](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/ProjectFavicon.tsx#L67): `ProjectFaviconFallback` applies the accent only through the SVG `color` property. Callers that pass a stroke utility such as `stroke-muted-foreground` (the existing sidebar tooltip does this) set an explicit `stroke` color, so Lucide fallback icons continue rendering muted instead of using `accentColor`; use the accent for the stroke or avoid overriding it when an accent is supplied. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/Sidebar.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 1580](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/Sidebar.tsx#L1580): `applyProjectColorToGroup` updates all grouped project members independently with `Promise.all` but does not roll back completed updates when one member update fails. For a group spanning multiple environments, a disconnected/failed backend can therefore leave the group with mixed persisted colors even though the user selected one group color; the error toast only reports failure and provides no recovery or indication of which members changed. <b>[ Exceeded comment limit ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/SidebarV2.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 287](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/SidebarV2.tsx#L287): `accentColor` is passed to the tooltip's `ProjectFavicon`, but this instance also supplies `stroke-muted-foreground` in its `className`. `ProjectFavicon` applies the accent only as an inline `color`; Lucide SVG strokes use the explicit CSS `stroke` property from `stroke-muted-foreground`, so a fallback favicon in the thread tooltip remains muted instead of showing the project accent. Remove the explicit stroke class or apply the accent to the stroke as well. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/SidebarV3.tsx — 0 comments posted, 8 evaluated, 7 filtered</summary>
>
> - [line 658](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/SidebarV3.tsx#L658): The settled PR badge composes `settledPrHoverClass` at this line, but `settledPrHoverColorClass` returns selectors for `group-hover/v2-row`. This row is declared as `group/v3-row`, so the selector never matches; settled PR numbers remain permanently at `text-muted-foreground/35` instead of adopting their open/merged/closed hover color. <b>[ Out of scope (post-validation triage) ]</b>
> - [line 907](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/SidebarV3.tsx#L907): `SidebarV3SearchResultRow` renders `ProjectFavicon` without passing the project accent color. For a project with a configured accent, search-result fallback icons therefore remain untinted and real favicon images get no accent badge, so the same project identity styling is lost specifically while searching. <b>[ Out of scope (post-validation triage) ]</b>
> - [line 1132](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/SidebarV3.tsx#L1132): `changeRequestStateByKey` is an unbounded per-thread cache with no pruning when a thread is deleted, archived, or otherwise leaves `threads`. Rows only update/delete their own key while mounted, and unmounting a row does not remove the entry. In a long-lived sidebar with many PR threads, this retains one map entry for every previously rendered thread and can grow for the lifetime of the component. <b>[ Out of scope (post-validation triage) ]</b>
> - [line 1167](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/SidebarV3.tsx#L1167): Project references are encoded as `${environmentId}:${projectId}` for scoping and lookup, but both IDs are only guaranteed to be non-empty strings and may contain `:`. Thus `(environmentId='a', projectId='b:c')` collides with `(environmentId='a:b', projectId='c')`; selecting one project can include the other project's threads, and `projectCwdByKey`/`projectDisplayNameByKey` can show the wrong project metadata. Use the canonical scoped-project key instead of this ambiguous concatenation. <b>[ Exceeded comment limit ]</b>
> - [line 1590](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/SidebarV3.tsx#L1590): Merged/closed PR auto-settlement stops working for active threads in a collapsed status section. `changeRequestStateByKey` is populated by each rendered `SidebarV3Row`'s VCS query/effect, but the `visibleAttentionThreads`/`visibleWorkingThreads`/`visibleReadyThreads` arrays become empty when a section is collapsed, so those rows unmount and no PR state is observed. A PR that merges while its section is collapsed therefore remains in the active section until the user expands it, instead of moving to Settled as the partition promises. <b>[ Exceeded comment limit ]</b>
> - [line 2057](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/SidebarV3.tsx#L2057): The bulk context menu always exposes `Settle (N)` without checking either `threadSettlement` capability or `canSettle`. Selecting it for a thread on an older server, or for a running/approval/input-blocked thread, dispatches `attemptSettle`, which immediately returns an unsupported/blocked failure and shows an error (while valid threads in a mixed selection may already have been settled). The menu should omit or disable this action unless the selected rows can actually settle. <b>[ Exceeded comment limit ]</b>
> - [line 2671](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/SidebarV3.tsx#L2671): The v3 sidebar never forwards a project's persisted `color` to `ProjectFavicon`. For example, the project-scope trigger passes only `environmentId`, `cwd`, and `className` here, so colored projects render the default fallback icon; the same omission is present for the project entries and v3 thread rows. Enabling Sidebar v3 therefore drops project accent-color presentation that `ProjectFavicon` supports. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/chat/ChangedFilesTree.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 245](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/chat/ChangedFilesTree.tsx#L245): The context-menu label is derived from the renderer's `navigator.platform`, but the reveal command is sent for the selected `environmentId` and the server launches the file manager using that environment's host platform. When a web/desktop client targets a different-platform environment (for example, macOS client to a Linux or Windows backend), this advertises “Reveal in Finder” or “Reveal in File Explorer” while actually launching `xdg-open` or `explorer` remotely, misleading users about the action. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/files/FileBrowserPanel.tsx — 0 comments posted, 3 evaluated, 3 filtered</summary>
>
> - [line 162](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/files/FileBrowserPanel.tsx#L162): The menu label is derived from the browser's `navigator.platform`, but the action is dispatched with `environmentId` and launches the file manager on that environment's host. When the UI is connected to a remote/WSL environment with a different OS, this advertises e.g. “Reveal in File Explorer” while the server actually runs the environment's Linux launcher, misleading users about what will happen. Use the selected environment's reported platform instead. <b>[ Out of scope (post-validation triage) ]</b>
> - [line 180](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/files/FileBrowserPanel.tsx#L180): The known tree path is passed through `resolvePathLinkTarget`, which treats trailing `:<digits>` or `:<digits>:<digits>` as terminal line/column syntax. POSIX project filenames may legally contain colons, so a file such as `logs/build:12:3` is sent as `/workspace/logs/build:12:12:3` rather than its actual absolute path; the reveal action then opens/selects a nonexistent target. Resolve a tree path as a filesystem path without position parsing. <b>[ Out of scope (post-validation triage) ]</b>
> - [line 180](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/files/FileBrowserPanel.tsx#L180): Tree entries are workspace-relative paths, but `resolvePathLinkTarget` gives any path beginning with `~/` home-directory semantics. A valid POSIX project path such as `~/docs/readme.md` is therefore revealed as `<inferred home>/docs/readme.md` instead of `<cwd>/~/docs/readme.md`. This action should join the tree path to `cwd` directly rather than interpreting terminal-link shorthand. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/components/files/FilePreviewPanel.tsx — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 819](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/components/files/FilePreviewPanel.tsx#L819): `revealLabel` is derived from the browser's `navigator.platform`, but `useRevealInFileManager` dispatches `openInEditor` to the selected environment and launches that environment host's file manager. When a client connects to a project on a different-OS remote environment (for example, macOS browser to a Windows server), the button announces the wrong manager (`Finder`/`Files`) even though the action opens Explorer. The label should use the target environment's platform or a platform-neutral label. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/hooks/useBulkThreadLaunch.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 53](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/hooks/useBulkThreadLaunch.ts#L53): An interrupted `startThreadTurn` result is silently treated as a successful launch. When the command is interrupted (for example by an environment/session teardown), this callback reaches the end without throwing, so `runBulkLaunchQueue` adds the item to `started` and continues to launch later items even though this thread was not confirmed started. The final summary can therefore claim threads started when they were not, and interrupted batches are not represented as failures or cancelled items. <b>[ Cross-file consolidated ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/hooks/usePackageManagerScripts.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 29](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/hooks/usePackageManagerScripts.ts#L29): `scriptNamesFromJson` treats every key in any object-valued `scripts` field as runnable without validating its value. For example, a JSON-valid but malformed manifest such as `{ "scripts": { "build": null } }` produces an importable `npm run build`/`composer run-script build` action even though npm requires script commands to be strings and Composer permits only strings or arrays. Selecting and saving that entry creates an action that cannot run instead of treating the source as invalid or skipping the entry. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/index.css — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 1004](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/index.css#L1004): Adding `v3` to the compatibility override at this selector makes a v3 sidebar forcibly use the stock zinc tokens, but the later themed-sidebar mapping only matches `v1` and `v2` (at lines 1326-1348). When a custom `html[data-theme-id]` theme is active, v3 therefore keeps the local stock `--background`, `--sidebar`, `--foreground`, etc. instead of inheriting the theme's `--app-theme-*` values, so enabling Sidebar v3 silently breaks themed sidebar colors. <b>[ Cross-file consolidated ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/lib/bulkPrompt.ts — 0 comments posted, 4 evaluated, 3 filtered</summary>
>
> - [line 11](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/lib/bulkPrompt.ts#L11): The regex accepts `__proto__` as a valid placeholder name, but `BulkLaunchDialog` keeps `rawValues` in a normal `{}` and evaluates `rawValues[placeholder.name]` before calling `splitBulkValues`. For a prompt containing `{__proto__}`, that lookup returns the inherited `Object.prototype` object rather than a string, so `splitBulkValues(...).split(...)` throws during dialog render before the user can enter values, crashing the bulk-send UI. The same prototype-key hazard also applies to the double-braced pattern. <b>[ Exceeded comment limit ]</b>
> - [line 72](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/lib/bulkPrompt.ts#L72): `splitBulkValues` returns every non-empty token without applying `MAX_BULK_THREADS`. The bulk dialog calls this for each textarea and then `expandBulkPrompts` materializes a row for every parsed value before it checks `rows.length > MAX_BULK_THREADS`; pasting a very large list therefore allocates and renders/processes the entire list instead of rejecting at 20, which can freeze or exhaust the browser tab. <b>[ Exceeded comment limit ]</b>
> - [line 131](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/lib/bulkPrompt.ts#L131): The placeholder grammar accepts `__proto__`, but rows are built in a normal `{}` and populated with `values[list.name] = ...`. For `{__proto__}`, that assignment invokes the inherited `Object.prototype.__proto__` setter; a string value is ignored rather than stored, so `fillPlaceholders` reads `Object.prototype` and replaces the token with `[object Object]` instead of the supplied value. This valid documented placeholder name therefore produces incorrect prompts. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/messageQueueStore.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 129](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/messageQueueStore.ts#L129): `clearThreadQueue` is never integrated with thread deletion: the web thread-delete path clears composer, terminal, and other per-thread state but does not call this action. Deleting a thread after queuing messages therefore leaves its `File` handles and blob preview URLs retained in `queuesByThreadKey` for the rest of the session, leaking memory/blob resources and retaining prompts for a thread the user deleted. <b>[ Cross-file consolidated ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/portDiscoveryState.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 36](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/portDiscoveryState.ts#L36): The new `agent` branch makes agent-owned listeners appear in `Sidebar` through `useThreadDiscoveredPorts`, but `Sidebar`'s `handleOpenDiscoveredPort` always calls `openDiscoveredPort` and never uses the `isPreviewSupportedInRuntime()` browser fallback. In the web runtime, `PreviewPanel` explicitly renders only “Preview is only available in the T3 Code desktop app”, so clicking the newly exposed globe for an agent-started server opens an unusable panel instead of the server in a browser tab. <b>[ Exceeded comment limit ]</b>
> </details>
>
> <details>
> <summary>apps/web/src/processCommandLabel.ts — 0 comments posted, 3 evaluated, 3 filtered</summary>
>
> - [line 19](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/processCommandLabel.ts#L19): `tokenize` recognizes only double quotes, so Unix shell wrappers that quote their command with single quotes are not actually unwrapped. For input `sh -c 'pnpm dev'`, the wrapper filter removes `-c` but the label becomes `'pnpm dev'` (including quote characters) instead of `pnpm dev`; this is common for `sh`/`bash` command lines. <b>[ Out of scope (post-validation triage) ]</b>
> - [line 36](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/processCommandLabel.ts#L36): `lastPathSegment` strips any token containing `/` or `\`, even when the separator is part of an inline option value rather than a standalone path. For example, `node --config=/repo/vite.config.ts` is rendered as `node vite.config.ts`, dropping the `--config=` argument and misleading the process panel about how the process was started. <b>[ Out of scope (post-validation triage) ]</b>
> - [line 57](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/apps/web/src/processCommandLabel.ts#L57): The wrapper cleanup filters `CMD_WRAPPER_FLAGS` from every remaining token, not just the wrapper prefix. A legitimate argument equal to one of those strings is therefore hidden: `cmd /c echo /d` is formatted as `echo`, and `bash -c tool -c` loses the trailing `-c`. The process label no longer reflects the command's arguments. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>docs/user/bulk-threads.md — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 41](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/docs/user/bulk-threads.md#L41): The statement that “Only the prompt differs” is false for the resulting threads: the bulk launch code derives each thread's `title` from the expanded row prompt, so different placeholder values produce different sidebar titles as well. The documentation should either mention the per-thread generated titles or avoid claiming that only the prompt changes. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>docs/user/cli-threads.md — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 25](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/docs/user/cli-threads.md#L25): The example is fenced as `bash` but uses a Windows path with single backslashes: `--project C:\repos\my-app`. In Bash, those backslashes are escape characters and the argument becomes `C:reposmy-app`, so copying this documented command on macOS/Linux cannot resolve the intended project. Use a forward-slash path or quote/document it as a Windows-shell example. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
> <details>
> <summary>packages/contracts/src/preview.ts — 0 comments posted, 2 evaluated, 2 filtered</summary>
>
> - [line 290](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/packages/contracts/src/preview.ts#L290): `ThreadOwnedProcess` is used for every PID returned by a terminal tree, but the terminal inspector includes the terminal shell PID itself (the inspector initializes `processIds` with `terminalPid`). The process list therefore exposes the PTY shell as an ordinary stoppable process; the UI sends this PID to `killOwnedProcess`, which accepts it from the terminal registry and signals it. Clicking Stop on the shell consequently kills/closes the terminal rather than stopping a subprocess, losing the interactive session. Exclude the terminal root from the exported process list or mark it non-killable. <b>[ Exceeded comment limit ]</b>
> - [line 316](https://github.com/pingdotgg/t3code/blob/44ff9739c0614979c45731b75a60e0dc8259d459/packages/contracts/src/preview.ts#L316): `PreviewKillDiscoveredServerInput` identifies a kill target only by `pid`. The server's terminal ownership check stores raw PIDs and authorizes `killOwnedProcess` from that set, without a process start-time/identity token or a fresh ancestry check for terminal-owned PIDs. If a terminal child exits and the OS reuses its PID before the registry refreshes, a request for the stale PID can send `SIGTERM` and the delayed `SIGKILL` to an unrelated process. Include and validate process identity (or revalidate the live terminal tree) before signaling and escalating. <b>[ Cross-file consolidated ]</b>
> </details>
>
>
> </details>
>


t3dotgg:
> [!NOTE]

The new-thread CLI command combines several separable server and client paths. It needs a smaller product decision and staged implementation before code review will be useful.

PR #4405, opened 2026-07-23
fix(client): self-heal empty thread details and back off failed thread subscriptions

colonelpanic8:
## What Changed

Two client-runtime fixes for thread views that can render empty forever:

- **Self-heal empty cached thread bodies.** `makeEnvironmentThreadState` skipped the HTTP snapshot whenever any cached body existed and resumed the socket subscription from the cached snapshot sequence. If the cached body had no messages but a current sequence, the events that carried the messages were below the resume cursor and were never replayed — the thread rendered empty permanently, surviving app restarts. The subscription now reloads the HTTP snapshot whenever the cached body has no messages, and still degrades to today's sequence-resume when the loader yields nothing (offline/404).
- **Exponential backoff for expected subscription failures.** `subscribeDynamic`'s `retryExpectedFailureAfter` now also accepts `(attempt: number) => Duration.Input`, with the consecutive-failure counter reset by any delivered element. The thread subscription uses it: 250ms base, doubling, capped at **2s while the thread has no data** and **10s once data exists**. Existing plain-duration call sites are unchanged.

## Why

An unsent composer draft subscribes to its pre-allocated thread id before the thread exists server-side (intentionally, so the subscription picks the thread up the moment the draft is submitted). With the flat 250ms retry, every open draft hot-looped an HTTP snapshot fetch + socket subscribe into a 404 four times per second, indefinitely. Measured before: steady ~4 req/s for as long as a draft stays open. Measured after: 250ms → 500ms → 1s → 2s steady (≤2s worst-case pickup after the draft is submitted; verified live that draft submission still promotes promptly and the created thread's messages appear).

The 2s/10s split matters because the draft-promotion flow renders the message timeline from the thread-detail subscription: a thread with no data yet may be created at any moment and must be picked up quickly, while a thread that already has data and keeps failing can back way off.

The empty-cache self-heal closes the loop for users who already have a poisoned cache: previously nothing ever re-fetched the full body, so the thread stayed empty forever.

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [ ] I included before/after screenshots for any UI changes
- [ ] I included a video for animation/interaction changes

Validation:

```
vp check                # 0 errors
vp run typecheck        # exit 0
vp test run packages/client-runtime/src/state/threads-sync.test.ts packages/client-runtime/src/rpc/client.test.ts   # 28 passed
```

[...]

PR #8192, opened 2026-08-25
fix(client-runtime): stop resubscribing threads the server reports missing

Coriou:
# PR-A — fix(client-runtime): stop resubscribing threads the server reports missing

## Body

Following up on #4589 (dead subscriptions on healthy connections) with the first of three small fixes from a production incident on a long-lived deployment.

**Problem:** `subscribeDynamic` retries every all-Fail cause forever. When a thread is deleted while its subscription is mid-flight, `orchestration.subscribeThread` keeps failing with `OrchestrationGetSnapshotError: Thread … was not found` — classified as an expected failure and retried at the fixed delay for the rest of the session. We counted **12,000+ consecutive misses (~4/s), still firing when we gave up and restarted the client after 80+ minutes** — against a healthy connection throughout.

**Fix:** adds a `terminalFailure` classification to `SubscriptionOptions`: an all-Fail cause whose errors all match a terminal predicate ends the attempt for the session, with no retry, after running its handler. The thread subscription call site classifies the server's `Thread … was not found` wording as terminal and applies the existing `setDeleted()` path — full deletion parity including cache clearing, so the UI drops the row instead of silently hammering a dead RPC.

Transport dormancy and defect propagation are unchanged. This intentionally does not touch reconnection/backoff design (#4602, #4405) — it only stops retrying when the server has said the thing is permanently gone.

**Validation** (cleanly applicable to upstream main `c6b8bb825` — zero file intersection with the last four upstream commits):
- `vp test run packages/client-runtime/src/rpc/client.test.ts packages/client-runtime/src/errors/orchestration.test.ts packages/client-runtime/src/state/threads-sync.test.ts` → 33 passed
- Full `@t3tools/client-runtime` suite → 676 passed (51 files)
- `tsgo --noEmit -p packages/client-runtime/tsconfig.json` → exit 0
- New tests cover: terminal classification edges, deleted-thread session replacement, and that non-terminal failures still retry.

Effect idioms match the repo's current v4-beta usage (no new deps; the only API surface added is the `terminalFailure` option on `SubscriptionOptions` plus the new exported `wasSubscribeThreadNotFound` predicate in `errors/orchestration.ts`).

**Merge-order note (for #4589's three-PR set):** the input-envelope PR relocates the `catchCause` this PR classifies inside. Whichever lands second conflicts by design; the combined result is the relocated envelope with this PR's classifier interior (transport → terminal → expected-failure retry → failCause). Do not resolve toward the old attachment point — it compiles but silently reintroduces the input-stage fiber death.

---

> [!NOTE]
> **Medium Risk**
> Changes durable subscription retry semantics and thread lifecycle on missing-server threads; requires server and client to agree on the new error field for correct classification.
> 
> **Overview**
> Stops **infinite `subscribeThread` retries** when the server says a thread no longer exists. Previously those failures were treated as expected errors and retried on a fixed delay for the whole session.
> 
> **Contract & server:** `OrchestrationGetSnapshotError` gains optional `threadDisposition: "not-found"`. `subscribeThread` sets it when the thread detail snapshot is missing (`apps/server/src/ws.ts`).
> 
> **Client-runtime:** `subscribeDynamic` adds a **`terminalFailure`** option—if every `Fail` in the cause matches `matches`, the handler runs and **that subscription attempt ends with no retry** (after transport handling, before expected-failure retry). Thread sync wires this to **`wasSubscribeThreadNotFound`** (typed disposition, not error message text) and **`setDeleted()`** so behavior matches a normal `thread.deleted` path (cache cleared, UI drops the row).
> 
> Tests cover the predicate, terminal vs retried failures in the RPC layer, and a single subscribe attempt after a not-found snapshot error.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d89a86e6338e1515ef5c4e72e546cbd731caab6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Stop resubscribing to threads the server reports as missing
> - Server WS layer now tags missing-thread snapshot errors with `threadDisposition: "not-found"` on `OrchestrationGetSnapshotError` in [ws.ts](https://github.com/pingdotgg/t3code/pull/8192/files#diff-7f2100e8ffa23a58d9bf425c5b3ceafc39239911ae33bc569291a00ecb8e9125)
> - Client runtime adds a `terminalFailure` option to `subscribeDynamic` in [client.ts](https://github.com/pingdotgg/t3code/pull/8192/files#diff-f60493a8e60511194714ca2543e1d58cfb8198eb71d2bf0c9655acdc93220c38); when all `Fail` errors match the classifier, the subscription stops retrying and invokes the handler
> - Thread state factory in [threads.ts](https://github.com/pingdotgg/t3code/pull/8192/files#diff-3389d9e86e5f65222bc1be800d0aa21f5356629eaf916680112dcbd606822815) wires `wasSubscribeThreadNotFound` as the classifier and calls `setDeleted()` on match, so missing threads transition to `deleted` instead of looping retries
> - Risk: subscriptions that previously retried indefinitely on missing threads now terminate; any code expecting the subscription to stay alive after a server-side thread deletion needs to handle the `deleted` state transition
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d89a86e.</sup>
> 
>

### DIFF
--- a/apps/server/src/ws.ts
+++ b/apps/server/src/ws.ts
@@ -1527,6 +1527,7 @@ const makeWsRpcLayer = (
                 return yield* new OrchestrationGetSnapshotError({
                   message: `Thread ${input.threadId} was not found`,
                   cause: input.threadId,
+                  threadDisposition: "not-found",
                 });
               }
 
--- a/packages/client-runtime/src/errors/orchestration.test.ts
+++ b/packages/client-runtime/src/errors/orchestration.test.ts
@@ -1,7 +1,54 @@
-import { OrchestrationDispatchCommandError } from "@t3tools/contracts";
-import { describe, expect, it } from "vite-plus/test";
+import {
+  OrchestrationDispatchCommandError,
+  OrchestrationGetSnapshotError,
+} from "@t3tools/contracts";
+import { describe, expect, it } from "@effect/vitest";
 
-import { wasBootstrapThreadDeleted } from "./orchestration.ts";
+import { wasBootstrapThreadDeleted, wasSubscribeThreadNotFound } from "./orchestration.ts";
+
+describe("wasSubscribeThreadNotFound", () => {
+  it("matches the typed not-found error", () => {
+    expect(
+      wasSubscribeThreadNotFound(
+        new OrchestrationGetSnapshotError({
+          message: "Thread thread-1 was not found",
+          cause: "thread-1",
+        }),
+      ),
+    ).toBe(true);
+  });
+
+  it("rejects messages with trailing content", () => {
+    expect(
+      wasSubscribeThreadNotFound(
+        new OrchestrationGetSnapshotError({
+          message: "Thread thread-1 was not found (will retry)",
+          cause: "thread-1",
+        }),
+      ),
+    ).toBe(false);
+  });
+
+  it("rejects plain errors with a matching message", () => {
+    expect(wasSubscribeThreadNotFound(new Error("Thread thread-1 was not found"))).toBe(false);
+  });
+
+  it("rejects other snapshot errors", () => {
+    expect(
+      wasSubscribeThreadNotFound(
+        new OrchestrationGetSnapshotError({
+          message: "Failed to load thread thread-1",
+          cause: "thread-1",
+        }),
+      ),
+    ).toBe(false);
+  });
+
+  it("rejects unrelated errors", () => {
+    expect(wasSubscribeThreadNotFound(new Error("boom"))).toBe(false);
+    expect(wasSubscribeThreadNotFound(null)).toBe(false);
+  });
+});
 
 describe("wasBootstrapThreadDeleted", () => {
   it("accepts only a confirmed deleted bootstrap thread", () => {
@@ -13,11 +60,17 @@ describe("wasBootstrapThreadDeleted", () => {
         }),
       ),
     ).toBe(true);
+  });
+
+  it("rejects a missing disposition", () => {
     expect(
       wasBootstrapThreadDeleted(
         new OrchestrationDispatchCommandError({ message: "Failed to create worktree." }),
       ),
     ).toBe(false);
+  });
+
+  it("rejects unrelated errors", () => {
     expect(wasBootstrapThreadDeleted(new Error("connection lost"))).toBe(false);
   });
 });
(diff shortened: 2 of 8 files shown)


Coriou:
Thanks @macroscopeapp[bot] — triaged all three:

1. **Regex vs structural discriminator** (`errors/orchestration.ts`): deliberate scope call for this PR. A structural field means contracts + server changes (`OrchestrationGetSnapshotError` gains a discriminator, `ws.ts` sets it), which would triple this PR's surface against CONTRIBUTING's small-focused guidance — and the regex degrades safely (double-anchored, tested against the sibling load-failure messages, and a future wording change falls back to today's retry behavior, never to a false stop). Happy to follow up with the contracts/server change as its own PR if maintainers prefer the structural route — it's exactly the Phase-5-style follow-up we considered.

2. **Foreground resubscribe re-runs the terminal handler** (`state/threads.ts`): real but bounded and idempotent — each foreground wakeup after a terminal stop issues exactly one subscribe attempt (vs the pre-fix behavior of re-entering the infinite 250 ms loop), the handler re-applies `setDeleted()` which is idempotent (state set + cache removal), and the stream completes again. This matches how ordinary `thread.deleted` deletions already behave across foreground wakeups. If maintainers prefer zero re-attempts, gating `foregroundResubscriptions` on `status !== "deleted"` is a small follow-up we're happy to add.

3. **`setDeleted()` vs the debounced persistence worker** (`state/threads.ts:650`): good catch on the window — but this race is pre-existing and shared by the ordinary `thread.deleted` path (same `setDeleted()`, same 500 ms-debounced writer, no tombstone at drain time on either web or mobile storage). This PR only routes not-found into the existing path. We'd rather not widen this PR to fix a pre-existing writer race; willing to send a separate focused PR (drain-or-tombstone in the persistence worker) if that's welcome.

shivamhwp:

Closing as superseded by #9822, which carries this missing-thread subscription fix forward and covers foreground resubscription and cache writes that could restore a deleted thread. Thank you for identifying and addressing the original problem. Review will continue on #9822.

The replacement is still open; this closure consolidates the proposals and does not mean the fix has shipped.

PR #8193, opened 2026-08-25
fix(client-runtime): retry subscription input failures instead of dying

Coriou:
# PR-B — fix(client-runtime): retry subscription input failures instead of dying

## Body

Second fix from the #4589 production incident (dead subscriptions on healthy connections). This one is deliberately minimal: it changes **where** input construction runs, nothing else.

**Problem:** in `subscribeDynamic`, `makeInput` runs inside `unwrap` **before** the `catchCause` pipe. Its real work includes the HTTP snapshot load plus reducer application, so a transient failure there kills the subscription fiber out of `Stream.runForEach` — no log, no retry, no error state. Signature from the incident: RPC works, streams never come back until the client restarts.

**Fix:** move the `makeInput` invocation inside the effect wrapped by `catchCause`, so input-construction failures flow through the same expected-failure classification as stream failures and are retried at the existing `retryExpectedFailureAfter` fixed delay. `makeInput` may now fail with the method's own error domain (`EnvironmentRpcStreamFailure`).

**Scope guards:**
- `retryExpectedFailureAfter` and its fixed-delay semantics are untouched.
- Transport dormancy ("waiting for the next session.") is byte-identical to main.
- Defects still die loudly (the existing "does not classify subscription defects as expected failures" test is untouched and passing).
- No backoff redesign here — that's #4405's active ground; this PR is orthogonal to scheduling and composes with whichever design lands.

**Validation** (upstream main `1baf99195` + this commit; full `@t3tools/client-runtime` suite 667/667 across 51 files):
- `vp test run packages/client-runtime/src/rpc/client.test.ts` → 9 passed (incl. new: failing input construction retries and converges once construction succeeds; defects still fatal)
- `tsgo --noEmit -p packages/client-runtime/tsconfig.json` → exit 0
- Both existing `subscribeDynamic` call sites (`state/threads.ts`, `state/shell.ts`) compile unchanged — their `makeInput` generators are total today; the widened type is forward-compatible with fallible inputs.

**Merge-order note (for #4589's three-PR set):** this PR and the terminal-classification PR rewrite the same `catchCause` region in opposite shapes — whichever lands second will conflict by design. The correct combined result is this PR's **outer envelope** (catchCause wrapping the `suspend`/`unwrap` containing `makeInput`) with the terminal PR's **classifier interior** inside it, precedence transport → terminal → expected-failure retry → failCause. Do not resolve toward the old attachment point: it compiles but silently reintroduces the input-stage fiber death.

---

> [!NOTE]
> **Medium Risk**
> Changes durable subscription retry semantics for input-prep failures in client-runtime; misclassification could cause extra retries, but defects and transport handling stay the same.
> 
> **Overview**
> Fixes durable RPC subscriptions that could **permanently die** when `subscribeDynamic`'s `makeInput` failed (e.g. transient HTTP snapshot / auth errors) even though the WebSocket session was healthy.
> 
> **`subscribeDynamic`** now runs `makeInput` inside the same `catchCause` envelope as stream failures, so those errors can use `onExpectedFailure` and `retryExpectedFailureAfter` instead of killing the subscription fiber before any RPC runs. The `makeInput` callback is typed to allow `EnvironmentRpcStreamFailure` errors; transport dormancy and defect handling are unchanged.
> 
> Adds a test that an `EnvironmentAuthorizationError` during input construction is observed once, does not subscribe until input succeeds, then recovers after the configured retry delay.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c670d0bca2920fea124bb71f7c1bf53bacc26080. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Retry `makeInput` failures in `subscribeDynamic` instead of letting the fiber die
> - `subscribeDynamic` now treats failures from `makeInput` as expected stream failures instead of fatal errors. The RPC method is not called until input construction succeeds.
> - Moves `Stream.catchCause` out one level to wrap both `makeInput` and the RPC method stream, so both share the same resilience logic: transport failures are logged and drained, expected failures invoke `options.onExpectedFailure` and optionally sleep for `options.retryExpectedFailureAfter` before resubscribing.
> - Changes the `makeInput` parameter type from `Effect<EnvironmentRpcInput<TTag>>` to `Effect<EnvironmentRpcInput<TTag>, EnvironmentRpcStreamFailure<TTag>>`.
> - Adds a test verifying the subscription fiber stays alive across an `EnvironmentAuthorizationError` during input construction and recovers once input construction succeeds.
> - Risk: callers passing a `makeInput` that fails with a non-`EnvironmentRpcStreamFailure` error may see different propagation behavior since the error channel type changed in [client.ts](https://github.com/pingdotgg/t3code/pull/8193/files#diff-f60493a8e60511194714ca2543e1d58cfb8198eb71d2bf0c9655acdc93220c38).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c670d0b.</sup>
> 
>

### DIFF
--- a/packages/client-runtime/src/rpc/client.test.ts
+++ b/packages/client-runtime/src/rpc/client.test.ts
@@ -1,5 +1,7 @@
 import {
+  EnvironmentAuthorizationError,
   EnvironmentId,
+  AuthTerminalOperateScope,
   type RelayClientInstallProgressEvent,
   WS_METHODS,
 } from "@t3tools/contracts";
@@ -25,7 +27,13 @@ import {
 import * as EnvironmentSupervisor from "../connection/supervisor.ts";
 import * as RpcSession from "../rpc/session.ts";
 import type { WsRpcProtocolClient } from "../rpc/protocol.ts";
-import { EnvironmentRpcRequestObserver, request, runStream, subscribe } from "./client.ts";
+import {
+  EnvironmentRpcRequestObserver,
+  request,
+  runStream,
+  subscribe,
+  subscribeDynamic,
+} from "./client.ts";
 
 const TARGET = new PrimaryConnectionTarget({
   environmentId: EnvironmentId.make("environment-1"),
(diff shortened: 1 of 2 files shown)


shivamhwp:

Main already wraps input construction in the outer `catchCause` through merged #10216, which also added `onDefect` reporting and initializer-defect coverage. Please rebase onto that implementation and retain its defect observer rather than restoring the older classifier.

The remaining change is the fallible-input type contract and its regression. Widen `makeInput` consistently in `subscribeDynamicMapped`, `subscribeDynamic`, and `subscribeDynamicWithSession`, and use Deferred signals for the new regression instead of bounded `yieldNow` loops. The wider #4589 transport-dormancy behavior remains outside this PR.

PR #8242, opened 2026-08-26
feat(desktop): capture the frontmost window with a double-Cmd hotkey

McSwaggens:
On macOS, pressing both Command keys together now captures the window you're working in and attaches it to your current thread's composer — a system-wide screenshot-to-chat gesture. Previously getting a screenshot to the agent meant capturing manually and pasting.

A small Swift sidecar (`native/screenshot-helper`) watches modifier flags with a global NSEvent monitor — modifier-only chords need no Accessibility permission, which is why this gesture — and captures the frontmost window on request (`CGWindowList` + `screencapture -l`; Screen Recording permission prompts once on first use, with a recovery toast that deep-links to System Settings). A pure, unit-tested chord state machine in the Electron main process decides when to fire, and a supervised helper lifecycle handles crashes, backoff, and permission-grant restarts. On capture, a native camera flash plays over the captured window, T3 comes to the front, and the screenshot flies from the captured window's on-screen position into the composer chip (one-shot WAAPI, transform/opacity only, disabled under reduced motion). If no thread is open it targets the most recently visited one. A macOS-only Settings → General toggle (default on) starts and stops the helper; the setting is ignored everywhere else. Attachment reuses the existing composer image pipeline end to end (validation, compression, upload).

Demo:

https://github.com/user-attachments/assets/6b46be3b-b794-47f9-8f0d-e816b5253175

Note for reviewers: the desktop CSP blocks `fetch(dataUrl)` (`connect-src` has no `data:`), which this feature hit and works around by decoding base64 directly — `PreviewView`'s annotation-screenshot path uses the same `fetch(dataUrl)` pattern and may be quietly affected too; worth a separate look.

Implemented by Claude Fable 5 via Claude Code.

---

> [!NOTE]
> **Medium Risk**
> New macOS native helper, Screen Recording TCC, and supervised child-process lifecycle add platform-specific failure modes; attachment reuses the existing composer image pipeline so send/upload paths stay familiar.
> 
> **Overview**
> Adds a **macOS-only** gesture: press **both Command keys** to capture the **frontmost window** and attach it to the chat composer, with a user toggle in **Settings → General** (`screenshotHotkey`, default on).
> 
> **Desktop main process** introduces `DesktopScreenshotHotkey`, which supervises a Swift sidecar (`native/screenshot-helper`) over JSON on stdin/stdout, runs a unit-tested **both-⌘ chord** handler, converts captures to renderer events, and **reconciles** helper lifecycle on bootstrap and after client-settings writes. Permission denial triggers a helper restart and IPC to open **Screen Recording** in System Settings. **Dev** builds the helper via `swiftc` when stale; **release mac builds** stage `t3-screenshot-helper` with `swiftc`/`lipo`.
> 
> **Renderer** adds `ScreenshotCaptureCoordinator` (mac desktop sidebar): buffered preload IPC, navigation to the last visited thread when needed, `addCapturedFiles` on the composer, optional **WAAPI** flight onto the image chip, and permission/error toasts via shared `ToastInlineLink`. Contracts extend `DesktopBridge`, client settings, and hotkey event schemas; user docs cover the feature.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 98afab4d75b3ee3e2136b19cf17bfd389c04d4af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add double-Cmd hotkey to capture frontmost window on macOS desktop
> - Introduces a native Swift sidecar helper (`native/screenshot-helper/main.swift`) that reports Command key state and captures the frontmost window on demand via stdio JSON.
> - Adds `DesktopScreenshotHotkey` service in [DesktopScreenshotHotkey.ts](https://github.com/pingdotgg/t3code/pull/8242/files#diff-4163be44789202fdf703c23efd711acf159182343b8363e339be10573a676ac0) to supervise the helper, manage capture timeouts, and restart on permission denial.
> - Implements `ScreenshotCaptureCoordinator` in [ScreenshotCaptureCoordinator.tsx](https://github.com/pingdotgg/t3code/pull/8242/files#diff-d19496c43f2b1ab972313b8451bfcc224dd1b5d68c0c125c68327eb6afda3d6f) to attach captured images to the composer with an animated flight.
> - Adds `screenshotHotkey` boolean setting (default `true`) and a macOS-only toggle in General settings.
> - Risk: feature requires macOS Screen Recording permission; if denied, helper restarts to refresh TCC state and UI prompts to open system settings via `openScreenRecordingSettings` IPC.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 98afab4.</sup>
> 
>


McSwaggens:
Addressed the review findings in 5e0aebd6c:

**Fixed**
- *Late reply cross-attaches wrong screenshot (Macroscope High)* — the helper protocol now carries a request id (`capture <id>` / echoed in the reply); the main process only completes the pending capture on an id match.
- *Temp PNG leak on unrequested/late replies (Macroscope Medium, Bugbot Low)* — unmatched successful replies now delete their temp file before being dropped.
- *Scoped-key mismatch in the navigation fallback (Macroscope Medium)* — confirmed: `threadLastVisitedAtById` is keyed by `scopedThreadKey`, so the bare-id lookup never matched. The resolver now builds scoped keys; regression test added.
- *Superseded composer wait never settles (Macroscope High, Bugbot Medium)* — the wait's cleanup now resolves the promise so the superseded flow bails via its generation check instead of staying suspended with the decoded screenshot.
- *First capture during boot is lost (Macroscope Medium)* — the preload now subscribes at load time and buffers events (bounded) until the renderer's listener registers.
- *Restore defaults skips the toggle (Bugbot Medium)* — `screenshotHotkey` added to `changedSettingLabels` and the restore payload.
- *Lexical timestamp comparison (Bugbot Low)* — the resolver now compares via `toSortableTimestamp`; test added for mixed-offset ordering.

**Fixed, but the stated impact was overstated**
- *"Stuck capture blocks future chords" (Bugbot High)* — a failed offer couldn't wedge the hotkey permanently: the 15s reply timeout (or the supervisor's `failPendingCapture` on helper exit) always cleared the slot, so the worst case was one 15s stall. Still hardened it: the pending slot is now cleared by an identity-guarded `ensuring` on every exit path.

McSwaggens:
Fixed the backoff-escalation finding: a helper run that survives past the fast-exit window now resets the restart backoff, so an isolated crash after a healthy stretch retries in 1s instead of 15s.

McSwaggens:
Addressed the remaining findings in 2f4eccb3f:

- *Offer-fails-then-stall (Macroscope Medium)* — `runCapture` now checks the `Queue.offer` result and bails immediately when the command queue is already ended, instead of holding the pending slot for the 15s reply timeout.
- *Swallowed `openScreenRecordingSettings` failure (Effect Service Conventions)* — the service now returns `Effect<boolean>` like `openExternal`, threaded through the IPC method and bridge (`Promise<boolean>`), so the renderer's fallback toast is reachable.
- *Duplicated inline toast link treatment (UI Consistency)* — extracted `ToastInlineLink` next to the toast primitives; both the release-notes link and the System Settings link render it.
- *Unused `data-composer-attachments` hook (UI Consistency)* — dropped.

McSwaggens:
Both remaining UI Consistency findings addressed: the overlay and the composer image chip now share `COMPOSER_IMAGE_CHIP_SURFACE_CLASS_NAME` (in `composerInlineChip.ts`, per its precedent), and the unread `SCREENSHOT_FLIGHT_ANIMATION_ID` is dropped.

McSwaggens:
The latest finding ("composerHandleRef is always null — coordinator rendered outside ComposerHandleContext.Provider") is a false positive: `__root.tsx` renders `<CommandPalette><AppSidebarLayout>…`, and `CommandPalette` provides `ComposerHandleContext` around its children (`CommandPalette.tsx` ~488), so the coordinator mounted in `AppSidebarLayout` is inside the provider. React context propagates through the component tree independent of any dialog/portal DOM structure. Empirically, the attach path works end-to-end — the demo video in the description was recorded against this code.

t3dotgg:
> [!NOTE]

We are not adding a global double-Command shortcut for frontmost-window capture. It introduces global input interception, screen-capture permission, image storage, and focus routing without a selected shortcut or privacy design.

PR #8271, opened 2026-08-26
fix(claude): expand slash commands when a message has attachments

### LINKED ISSUE #8013: [Bug]: Attaching an image silently disables slash-command/skill expansion for the whole message (Claude provider)
### Before submitting

Closest existing issues, none of which cover this: #7671 (pickers offer entries that cannot run; `$` flattens to text; `/` menu opens on any line start), #7795 (provider-agnostic `$skill`), #3594 (duplicate execution when the first message is only a skill). Those are about which entries the pickers offer and where in the message the token sits. This one is about an attachment dropping an invocation that is otherwise valid and correctly placed.

### Area
apps/web

### Steps to reproduce
1. Claude provider, any thread, any user skill installed (`~/.claude/skills/flow-patterns/SKILL.md`).
2. Send exactly `/flow-patterns hello`, no attachment. The skill expands: the turn receives `<command-name>/flow-patterns</command-name>`, `<command-args>hello</command-args>`, and the full SKILL.md body.
3. Send the identical text `/flow-patterns hello` with any image attached (paste or drag).
4. Inspect the turn the agent actually received.

### Expected behavior
Expansion should not depend on whether the message carries an attachment. `/flow-patterns hello` + an image should deliver the same expanded command it delivers without one, or fail loudly if the combination is unsupported.

### Actual behavior
With an attachment, nothing expands. The agent receives the literal string `/flow-patterns hello` plus the image. No `<command-message>`, no `<command-name>`, no skill body, and no error or warning is shown anywhere in the UI, so from the user's side the send looks identical to the working case.

I isolated this by changing one variable at a time; only the attachment matters:

| message | attachment | commands | command in mid-prose | expanded |
| `/web-css` | no | 1 | no | yes |
| `/flow-patterns hello` | no | 1 | no | yes |
| `/flow-patterns /notarealcommand` | no | 1 | no | yes |
| `/flow-patterns /lookover` | no | 2 | no | yes (leading command only) |
| `/flow-patterns hello` | **1 image** | 1 | no | **no** |
| `/flow-patterns /lookover ...prose... /web-scaffold` | **3 images** | 3 | yes | **no** |

The last row is what I hit in real use and first misdiagnosed as a multiple-command problem. Rows 3 and 4 rule that out. A bare `/token` and a second real command name both leave the leading command's expansion intact, though the second command is itself swallowed into `<command-args>` rather than run, which is a separate limitation tracked in #7795. The only variable that differs between rows 2 and 5 is the attachment.

The practical cost is high because attachment plus skill is a natural pairing. Running a design skill on a screenshot, or a review skill on a pasted log, means putting the image and the command in one message, and that is the combination that drops the command. The agent then answers from general knowledge, and nothing in the thread indicates the skill never loaded.

I could not determine which layer drops it. Both are candidates: the T3 send path may flatten the composer command to text when the message carries attachments, or the `claude` CLI may decline expansion for a multimodal prompt. `~/.t3/userdata/logs/provider/events.<thread>.log` records no outbound user-prompt events (only `turn.started` with a bare `{"model": ...}` payload, plus assistant and tool traffic), so it cannot discriminate from the client side.

### Impact

### Version or commit
t3code-nightly-bin 0.0.34_nightly.20260820.1142 (AUR)

### Environment
Arch Linux, kernel 7.1.8-arch1-3, desktop app, Claude provider, `claude` CLI 2.1.241, model Claude Opus 5.

### Workaround
Send the skill invocation as its own message with no attachment, then send the attachment in a follow-up. Or skip the prefix and name the skill in prose ("use flow-patterns on this image"), which lets the agent invoke it through its own Skill tool - though that does not work for skills marked `disable-model-invocation: true`, whose only valid entry point is explicit user invocation.

lnieuwenhuis:
With the Claude provider, sending `/skill-name args` expands the command normally, but the identical message with any image attached silently does not expand: the agent receives the literal string plus the image, with no `<command-name>`, no skill body, and no error anywhere in the UI.

## Root cause

It is on our side, not in the `claude` CLI.

The CLI accepts slash commands alongside images, but only when the **last** content block of the streamed user message is text. Reading the CLI bundle (verified in both 2.1.234 and 2.1.246), the stream-json input handler does:

```js
let last = blocks.at(-1);
if (last?.type === "text") { inputString = last.text; precedingInputBlocks = blocks.slice(0, -1); }
else { precedingInputBlocks = blocks; }
...
if (inputString !== null && inputString.startsWith("/")) processSlashCommand(inputString, precedingInputBlocks, ...)
```

`precedingInputBlocks` is then carried through the expansion, so the image is preserved in both the command message and the expanded skill content.

`buildUserMessageEffect` in the Claude adapter pushed the text block first and appended image blocks after it. With no attachment the single text block is last and expansion works; with any attachment the last block is an image, `inputString` stays `null`, and the turn falls through to the plain-prompt path. That is exactly the reporter’s isolation table: only the attachment variable matters.

## Fix

Emit the image blocks first and the text block last. This also matches Anthropic’s recommended image-before-text ordering, so it is a single code path with no special case for command-looking text.

## Other providers

Unchanged, and not affected by this defect. Codex passes prompt text and attachments as separate fields to the codex app-server. Cursor and Grok go over ACP `session/prompt`, and OpenCode over its own parts API; none of them route prompt text through the Claude CLI’s slash expander, so block ordering has no equivalent meaning there.

## Verification

- `vp test run apps/server/src/provider/Layers/ClaudeAdapter.test.ts` — 79 passed. The new regression test asserts the adapter produces the same trailing `/flow-patterns hello` text block with and without an image attachment; it fails on the pre-fix adapter and passes after.
- `vp test run apps/server/src/provider/Layers/ProviderService.test.ts` — passed.
- `vp run --filter t3 typecheck` — 0 errors.
- `vp lint` and `vp fmt --check` on the two touched files — clean (one pre-existing unrelated warning elsewhere in the file).

Fixes #8013

Implemented with Claude Opus via the Claude Code harness.

---

> [!NOTE]
> **Low Risk**
> Narrow change to Claude user-message block ordering with targeted tests; fixes incorrect behavior without touching auth or other providers.
> 
> **Overview**
> Fixes a silent failure where Claude **slash commands** (e.g. `/flow-patterns hello`) stopped expanding when an **image** was attached—the agent got literal text plus the image with no error.
> 
> **`buildUserMessageEffect`** now emits **image blocks before the final text block**. The Claude CLI only treats streamed input as a slash command when the **last** content block is text; putting text first meant any attachment made the last block an image and skipped expansion.
> 
> Skill-dispatch behavior is unchanged in intent: optional leading prose, images in the middle, then trailing command or plain text. Tests add **`readPromptMessages`**, expect **image-then-text** for image turns, and regress on command + attachment ordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1b9f6749a50559026dfda1018d21fed6bdd61c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix Claude slash command expansion when message has image attachments
> - In `buildUserMessageEffect` in [ClaudeAdapter.ts](https://github.com/pingdotgg/t3code/pull/8271/files#diff-e696bea66caf21d357bea5ea814bcbb18d1816288cc94247c855823f50fce6dc), image attachment blocks are now emitted before the command or prompt text block, so the dispatched slash command is the final content block.
> - Previously, ordinary prompt text was emitted before attachments, causing a slash command to be followed by an image and not recognized for expansion.
> - Adds tests in [ClaudeAdapter.test.ts](https://github.com/pingdotgg/t3code/pull/8271/files#diff-1657a55ebc1e3682293aa20a7f71a55adf8de546f36af50f289dca9da45b0b6f) covering command-only and command-plus-image turns, plus an async `readPromptMessages` helper to inspect multiple queued prompt messages.
> - Risk: any code or tests that assumed text blocks precede image blocks in Claude user messages will need updating; the new ordering places image blocks first.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a1b9f67.</sup>
> 
>


t3dotgg:
> [!NOTE]

Closing in favor of [#9122](https://github.com/pingdotgg/t3code/pull/9122), which just merged. It is this same fix rebased onto main, with your commit and authorship kept. The rebase needed a one-line test fix because `attachmentRelativePath` became nullable after this PR was opened, and it picked up a test for non-image file attachments. Thanks for tracking down the root cause in the CLI.

PR #7691, opened 2026-08-20
fix(server): report Claude authentication accurately

Ic3b3rg:
## What Changed

- Probe first-party Claude Code authentication with `claude auth status` using the same configured binary, environment, and `CLAUDE_CONFIG_DIR` used for chats.
- Decode the CLI's `loggedIn` result and report logged-out sessions as unauthenticated instead of trusting stale or incomplete SDK initialization metadata.
- Preserve existing behavior for external API providers and Claude Code versions older than 2.1.41; report probe failures as unknown rather than authenticated.
- Add focused regression coverage for logged-out sessions, stale account metadata, CLI compatibility, config-directory propagation, and probe failures.

## Why

T3 could report Claude as ready and authenticated whenever the SDK initialization probe returned a capabilities object, even if the exact CLI environment used for chats was logged out. The first chat then failed with `authentication_failed`, leaving Settings in direct conflict with the provider process.

This keeps authentication truth at the Claude adapter boundary and follows the CLI's supported auth-status contract without changing client or wire contracts.

Closes #7690.

## Testing

- `vp test run apps/server/src/provider/Layers/ProviderRegistry.test.ts` — 49 passed
- `vp run --filter t3 typecheck` — passed (pre-existing Effect suggestions only)
- Targeted lint and format checks — passed
- `vp run -r test` — local run stopped on the unrelated existing `scripts/build-desktop-artifact.test.ts` cross-architecture Windows assertion; neither that test nor its implementation is changed by this PR

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] UI screenshots are not applicable (server-only change)
- [x] Animation/interaction video is not applicable

Implemented with GPT-5.6 Sol in the Codex harness via T3 Code.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 35ed708b1f0cd52da9131a9fcddc3d5a20382198. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Probe `claude auth status` in `checkClaudeProviderStatus` and report CLI auth accurately
> - Adds an auth probe to `checkClaudeProviderStatus` that runs `claude auth status` when the CLI version is >= `2.1.41` (`MINIMUM_CLAUDE_AUTH_STATUS_VERSION`) and no external API provider is configured.
[...]

PR #8275, opened 2026-08-26
fix(claude): report unauthenticated Claude instances honestly

filipef101:
## Problem

A Claude provider whose CLI is not logged in reports **Authenticated** in Settings, then fails every turn with `Not logged in · Please run /login`.

The snapshot derives auth solely from the Agent SDK initialization result. A logged-out config directory still initializes fine and reports `{ tokenSource: "none", apiProvider: "firstParty" }`. `checkClaudeProviderStatus` only checks whether `capabilities` is truthy, so that object takes the `status: "ready"` / `auth: { status: "authenticated" }` branch.

This was survivable when every install had one Claude. With provider instances it is not: a second instance points `CLAUDE_CONFIG_DIR` at a fresh directory that is logged out by construction, so adding one always produces a provider that claims to be authenticated and cannot run a turn.

`claude auth status` used to be part of this probe and was dropped in #2277 (`feat: Multi-Provider support`); this restores that signal.

## Fix

Ask the CLI, with the instance's own environment:

- `claude auth status` emits JSON by default; `loggedIn: false` maps to `auth: { status: "unauthenticated" }` and `status: "error"`, matching what Codex and Cursor already report.
- The message names **that instance's** config directory, so the user logs in the failing instance rather than the default one:
  `Claude Code is not authenticated for this instance. Run \`CLAUDE_CONFIG_DIR=/Users/you/.claude_work claude auth login\` and try again.`
- The command exits non-zero *because* the user is logged out while still printing a well-formed verdict, so the parse is deliberately not gated on the exit code. There is a test for exactly that.
- A CLI too old to answer, or any spawn failure, leaves the previous SDK-derived verdict untouched — the probe swallows defects as well as typed failures.
- When the CLI does report an account, its email and subscription are preferred over the SDK's. With several instances configured, the email is what tells two Claude accounts apart in the provider list.

No contract change. `auth.status: "unauthenticated"` is an existing value that clients already handle — mobile, for instance, already filters unauthenticated providers out of its model picker, so a logged-out instance stops being offered instead of being offered and failing.

## Tests

Five added to `ProviderRegistry.test.ts`: logged-out reporting, the per-instance login hint, non-zero exit still trusted, older CLI keeping the SDK verdict, and account labelling from the CLI.

`vp test run apps/server/src/provider/Layers/ProviderRegistry.test.ts` — 48 passed. Typecheck and lint clean.

---

---

> [!NOTE]
> **Medium Risk**
> Changes provider snapshot auth semantics for multi-instance Claude setups; failures in the new probe are designed to fall back, but logged-out instances that previously appeared healthy will now surface as errors.
> 
> **Overview**
> Fixes Claude provider health checks that showed **Authenticated** for logged-out config dirs (common on extra instances with a separate `CLAUDE_CONFIG_DIR`) while turns failed with not logged in.
> 
> **`checkClaudeProviderStatus`** now runs **`claude auth status`** in the instance environment (with `AUTH_PROBE_TIMEOUT_MS`). When `loggedIn` is false it returns **`status: error`** and **`auth.status: unauthenticated`**, with a **`claudeLoginHint`** that sets `CLAUDE_CONFIG_DIR` for that instance—quoted paths and PowerShell vs POSIX syntax on Windows.
> 
> If the CLI probe is missing, times out, or returns unparseable output, behavior stays on the SDK initialization verdict. When logged in, **email and subscription metadata prefer the CLI** over the SDK so multiple instances are easier to tell apart. Auth JSON is parsed from stdout even on **non-zero exit codes**.
> 
> Tests in **`ProviderRegistry.test.ts`** cover logged-out reporting, per-instance hints, quoting, Windows hints, legacy CLI fallback, CLI account labeling, and non-zero exit trust.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bbb3fbee78f104b80f08b346670e1c673659f5ce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Probe `claude auth status` in `ClaudeProvider` and report unauthenticated instances honestly
> - Adds `probeClaudeAuthStatus`, which runs `claude auth status` with a dedicated timeout, parses stdout via `parseClaudeAuthStatus`, and swallows failures by returning `undefined`.
> - `checkClaudeProviderStatus` now returns an explicit `unauthenticated` error with a platform-aware login command targeting the instance's `CLAUDE_CONFIG_DIR` when the CLI reports logged out; authenticated accounts prefer CLI-provided email, subscription, auth method, and API provider, falling back to SDK capabilities.
> - `claudeLoginHint` and `quoteForShell` build shell-safe `CLAUDE_CONFIG_DIR=... claude auth login` hints for POSIX and PowerShell, with correct quoting for paths containing spaces.
> - Behavioral Change: provider status flow now always invokes `claude auth status` (with `AUTH_PROBE_TIMEOUT_MS`); if the CLI is missing or unparseable, it falls back to prior SDK-based behavior. Generated login hints reference the per-instance config dir, so users with custom `CLAUDE_CONFIG_DIR` must run the printed command rather than a bare `claude auth login`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bbb3fbe.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/ClaudeProvider.ts
+++ b/apps/server/src/provider/Layers/ClaudeProvider.ts
@@ -18,6 +18,7 @@ import {
   getProviderOptionCurrentValue,
   getProviderOptionDescriptors,
 } from "@t3tools/shared/model";
+import { HostProcessPlatform } from "@t3tools/shared/hostProcess";
 import { resolveSpawnCommand } from "@t3tools/shared/shell";
 import { compareSemverVersions } from "@t3tools/shared/semver";
 import {
@@ -31,6 +32,7 @@ import {
 import {
   buildBooleanOptionDescriptor,
   buildSelectOptionDescriptor,
+  AUTH_PROBE_TIMEOUT_MS,
   buildServerProvider,
   DEFAULT_TIMEOUT_MS,
   isCommandMissingCause,
@@ -40,7 +42,7 @@ import {
   type ServerProviderDraft,
 } from "../providerSnapshot.ts";
 import { resolveClaudeSdkExecutablePath } from "../Drivers/ClaudeExecutable.ts";
-import { makeClaudeEnvironment } from "../Drivers/ClaudeHome.ts";
+import { makeClaudeEnvironment, resolveClaudeHomePath } from "../Drivers/ClaudeHome.ts";
 import { discoverClaudeSkills } from "../Drivers/ClaudeSkills.ts";
 
 const DEFAULT_CLAUDE_MODEL_CAPABILITIES: ModelCapabilities = createModelCapabilities({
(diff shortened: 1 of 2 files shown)


filipef101:
Thanks — both real, both fixed in the latest push.

**Auth probe timeout.** Correct, and it defeated the PR's own purpose: `AUTH_PROBE_TIMEOUT_MS` (10s) exists in `providerSnapshot.ts` for exactly this and `CodexProvider` already uses it, while I had `DEFAULT_TIMEOUT_MS` (4s). On timeout the probe returns `undefined` and falls back to the SDK verdict — the false "authenticated" this PR is meant to correct. Now uses the auth budget.

**Login hint quoting.** Also right, and it applied to Windows twice over. The path is now shell-quoted, and the command is platform-aware — PowerShell has no inline `VAR=value command` form, so Windows gets `\$env:CLAUDE_CONFIG_DIR='…'; claude auth login`. Platform comes from `HostProcessPlatform` rather than `process.platform`, per the repo's `no-global-process-runtime` rule, which also made the Windows branch testable.

Two tests added: a config directory containing spaces, and the PowerShell form. 50 passing.

t3dotgg:
> [!NOTE]
>

Closing in favor of [#7691](https://github.com/pingdotgg/t3code/pull/7691) for Claude authentication status. It checks the CLI with version and backend guards and treats failed probes as unknown instead of authenticated. The useful account labels and per-instance login guidance from this PR are recorded there, including Windows quoting. Review continues there.

PR #7419, opened 2026-08-18
fix(server): count every provider instance on the Usage page

MatthewFoxAF:
## What Changed

The usage scan now resolves one transcript directory per configured Claude and Codex provider instance instead of only the legacy default `providers.<kind>` homes. It walks the same merged instance map the provider registry hydrates from (`deriveProviderInstanceConfigMap`), so explicit `providerInstances` entries and the legacy mirrors are both covered, and dedupes by resolved directory so instances that share a home are not counted twice.

The resolution moved out of `UsageService` into `usageTranscriptDirs.ts` so it has focused tests (`usageTranscriptDirs.test.ts`): legacy-only, extra instances, nested `.claude/projects` probe, shared-home dedupe, and skipping non-transcript drivers / undecodable configs. One sentence added to `docs/user/usage.md`.

## Why

A second Claude instance pointed at its own `CLAUDE_CONFIG_DIR` (or a second Codex home) writes transcripts there, not under `~/.claude/projects`, so the Usage page silently under-reported that instance's spend as zero. Reported on a machine with a `claude-second` install: the page showed only the default account.

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] I included before/after screenshots for any UI changes (n/a — no UI change)
- [x] I included a video for animation/interaction changes (n/a)

## Before
image

## After
image

---

> [!NOTE]
> **Medium Risk**
> Changes which directories are scanned for billing-style usage totals; wrong paths could still under- or over-report, but behavior is aligned with the provider registry and covered by new unit tests.
> 
> **Overview**
> Fixes **under-reported Usage** when multiple Claude or Codex installs use separate home directories. The scan no longer walks only the legacy default `providers.<kind>` homes.
> 
> Transcript directory resolution moves to **`usageTranscriptDirs.ts`**, which uses the same merged instance map as the provider registry (`deriveProviderInstanceConfigMap`). Each Claude/Codex instance gets its own transcript path (with the existing nested `.claude/projects` probe), **deduped** when two instances share a home, and **skipped** for non-transcript drivers or invalid configs. **`UsageService`** delegates to that helper and supplies **`FileSystem`** alongside **`Path`** for the existence probe. **`usageTranscriptDirs.test.ts`** covers legacy-only, extra instances, nesting, dedupe, and skip behavior; user docs note that every configured instance counts.
> 
[...]

PR #6603, opened 2026-08-14
fix(usage): aggregate transcripts from every provider instance

effektsvk:
## What Changed

The Usage page now aggregates transcript history from every configured Claude Code and Codex provider instance.

- Resolve each instance's explicit `homePath` or merged `CLAUDE_CONFIG_DIR`/`CODEX_HOME`, while retaining built-in defaults for legacy settings.
- Resolve relative `CLAUDE_CONFIG_DIR` values against every known active or archived project and worktree cwd, matching Claude's runtime behavior.
- Deduplicate identical transcript roots before scanning and share record-level deduplication across every source in one scan.
- Associate usage buckets with their exact transcript source via `sourceIndex`, so cross-environment merging removes only overlapping roots.
- Bump the usage contract to v5 while decoding older buckets with `sourceIndex: 0`, and document multi-instance collection.
- Add focused coverage for source resolution, cross-source aggregation, contract compatibility, and usage merging.

Tests: 4 files passed, 35 tests passed.

## Why

Previously, usage collection only scanned the default Claude Code and Codex homes, omitting activity from additional provider instances and workspace-relative Claude config directories. Cross-environment ownership also operated at the provider level, which could discard valid usage when two environments shared one transcript root but had other distinct roots.

Resolving the same provider instances and workspace-relative paths used at runtime, sharing record deduplication across roots, and tracking each bucket's exact source makes collection complete without double-counting copied records or dropping unrelated instance usage.

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] I included before/after screenshots for any UI changes (not applicable)
- [x] I included a video for animation/interaction changes (not applicable)

Authored with `gpt-5.6-sol` through the Codex harness in T3 Code.

---

> [!NOTE]
> **Medium Risk**
> Changes usage collection, aggregation, and the v5 contract; wrong path resolution or merge logic could misreport totals, though behavior is covered by new tests and backward-compatible decoding.
> 
> **Overview**
> The Usage page now scans **every configured Claude and Codex provider instance** instead of only the default homes, using a new `resolveUsageTranscriptSources` path that honors explicit `homePath`, per-instance `CLAUDE_CONFIG_DIR` / `CODEX_HOME`, and legacy default slots.
> 
[...]

PR #8276, opened 2026-08-26
fix(usage): scan transcripts for every configured provider instance

filipef101:
## Problem

The Usage page under-reports whenever more than one instance of Claude or Codex is configured.

`resolveTranscriptDirs` resolves exactly two directories, both from the legacy per-driver blobs:

```ts
const claudeHome = yield* resolveClaudeHomePath(settings.providers.claudeAgent);
const codexLayout = yield* resolveCodexHomeLayout(settings.providers.codex);
```

`settings.providerInstances` is never consulted. A second Claude or Codex points at its own home, so every turn it ran is invisible — the page shows a plausible-looking number that is simply missing that account's spend.

It also misses the default instance's real home once the default row has been edited in Settings: `buildProviderInstanceUpdatePatch` promotes that row into `providerInstances.<kind>` and resets `providers.<kind>` to schema defaults, so the blob this function reads is blank while the instance entry holds the real `homePath`.

## Fix

Derive the directory list from the configured instances, with the same precedence hydration already uses — an explicit `providerInstances` entry wins over the legacy mirror for that driver's default slot, and drivers with no explicit entry fall back to the blob.

Directories are deduped by resolved path, so two instances that differ only by binary or launch args (sharing one home) are not counted twice. Drivers with no transcript reader — cursor, grok, opencode, and anything a fork adds — are skipped, and an instance whose config cannot decode is skipped rather than failing the whole scan, since the provider registry already surfaces those as unavailable.

The scan loop itself was already per-directory and stamps each source with its own `resolvedHomePath` fingerprint, so multiple directories per provider needed no further change.

## Tests

New `usageInstances.test.ts` covers the legacy fallback, several Claude instances, explicit-entry precedence over the legacy blob, and unreadable drivers being ignored.

`vp test run apps/server/src/usage` — 42 passed. Typecheck and lint clean.

---

---

> [!NOTE]
> **Medium Risk**
> Changes which directories drive usage totals; misconfigured or undecodable instances are skipped silently, so spend can still look low without a scan error.
> 
> **Overview**
> Fixes **under-reported usage** when multiple Claude or Codex instances exist, or when the real home lives only under `providerInstances` after Settings edits.
> 
> **`scannableInstanceConfigs`** builds the scan list from `ServerSettings`: all Claude/Codex `providerInstances` (decoded settings), with legacy `providers.<driver>` only when the default instance id is missing—matching provider hydration precedence. Unreadable instance configs and drivers without transcript parsers are skipped without failing the scan.
> 
> **`resolveTranscriptDirs`** now resolves **one transcript directory per scannable instance** (not one per driver) and **dedupes by `realPath`** so shared homes are not scanned twice and costs are not doubled.
> 
> New **`usageInstances.test.ts`** covers legacy fallback, multiple Claude homes, explicit default-id over legacy blob, and ignored non-scannable drivers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c64148a3a648efb8fc54b8c3f4ea083d919cbf1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Scan transcripts for every configured Claude and Codex provider instance
> - Replaces fixed per-driver directory resolution with per-instance resolution driven by `scannableInstanceConfigs`, which decodes `claudeAgent` and `Codex` configs from `settings.providerInstances`
> - Falls back to legacy `settings.providers.<kind>` blob when no explicit instance exists under the driver's default id
> - De-duplicates directories by canonical path via `realPath` to avoid double-counting instances that resolve to the same location
> - Adds test coverage in [usageInstances.test.ts](https://github.com/pingdotgg/t3code/pull/8276/files#diff-bc80dfc931095c5fbac66a5df83f6594d8f78b70dfdad9ee32f1553c30a70766) for legacy fallback, non-default instances, default-id preference, and unsupported-driver exclusion
> - Risk: undecodable or unsupported-driver instance configs are silently skipped — check `scannableInstanceConfigs` in [UsageService.ts](https://github.com/pingdotgg/t3code/pull/8276/files#diff-1d90d173860922c384f65961de4b51bba0176aadd7982190e0dcf87e663f1a70) if expected instances are missing from usage scans
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c64148a.</sup>
> 
>

### DIFF
--- a/apps/server/src/usage/UsageService.ts
+++ b/apps/server/src/usage/UsageService.ts
@@ -14,7 +14,12 @@
 import * as NodeOS from "node:os";
 
 import {
+  ClaudeSettings,
+  CodexSettings,
+  defaultInstanceIdForDriver,
+  ProviderDriverKind,
   USAGE_CONTRACT_VERSION,
+  type ServerSettings as ServerSettingsDocument,
   type UsageProviderKind,
   type UsageSource,
   type UsageSummary,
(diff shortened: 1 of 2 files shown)


filipef101:
Good catch — fixed in the latest push.

The dedupe compared resolved paths, so two spellings reaching the same directory through a symlink counted as distinct and every record in that home was counted twice. That matters more than usual here because there is no cross-file de-duplication: `dedupeWithinFile` is the only guard, so the second scan is pure inflation.

Directories are now canonicalized with `fileSystem.realPath` before the dedupe check. A directory that does not exist yet has no real path, so that falls back to the resolved path — which still catches the plain duplicate case. The path reported in each source's fingerprint is deliberately still the configured one, not the canonical one, so the UI keeps naming the directory the instance is configured with.

t3dotgg:
> [!NOTE]

Closing this overlapping scanner in favor of [#6596](https://github.com/pingdotgg/t3code/pull/6596). This version deduplicates only by canonical path, which can lose provider identity when two providers share a directory. Preserve the configured-instance cases and reconcile the retained reader with Grok usage. [#6603](https://github.com/pingdotgg/t3code/pull/6603) and [#7419](https://github.com/pingdotgg/t3code/pull/7419) are related alternatives, not merged replacements.

PR #8277, opened 2026-08-26
fix(settings): keep the rest of settings.json when one provider instance is malformed

filipef101:
## Problem

One malformed `providerInstances` entry silently resets **every** setting, and the next settings change makes that permanent.

`providerInstances` is a `Schema.Record(ProviderInstanceId, ProviderInstanceConfig)`, so a key that fails the slug pattern, or an entry missing `driver`, fails the whole `ServerSettings` decode. `loadSettingsFromDisk` logs a warning and returns `DEFAULT_SERVER_SETTINGS`. From the UI this looks like every setting spontaneously reverted. Then `updateSettings` computes the next state from that defaulted `current` and calls `writeSettingsAtomically`, so the first subsequent change — a provider toggle is enough — overwrites `settings.json` with defaults. Keybindings, projects, provider paths, and any other instances are gone, with a log line as the only trace.

`fromLenientJson` is lenient about JSON syntax, not about schema errors, so this is reachable from an ordinary typo. Hand-editing this map is a documented way to configure a second provider instance (`docs/user/providers-claude.md`), which makes the typo a realistic one.

To reproduce: add `"providerInstances": { "2claude_work": { "driver": "claudeAgent" } }` to `settings.json` (leading digit fails `PROVIDER_SLUG_PATTERN`), let the watcher reload, then toggle any provider in Settings.

## Fix

Two independent guards:

1. **Salvage.** On a decode failure, re-decode with the undecodable `providerInstances` entries dropped. If that succeeds, keep every other setting and name the dropped ids in the warning. Returns early when there is nothing to prune, so a document failing for some other reason is not retried pointlessly.
2. **Quarantine.** Before the in-memory copy can be written back, copy the unreadable file to `settings.json.invalid`. A stable name rather than a timestamp, since reloads fire on every watcher tick while the file is broken and one recoverable copy is the useful outcome. Failures here are logged and swallowed — a settings file we cannot back up is still one the server has to boot with.

Behaviour on a clean file is unchanged: the success path returns before either guard.

## Tests

Two added to `serverSettings.test.ts`: unrelated settings surviving a malformed instance, and an unreadable file being quarantined byte-for-byte.

`vp test run apps/server/src/serverSettings.test.ts` — 33 passed. Typecheck and lint clean.

Also verified against a running server: injecting a bad instance kept the rest of the config live, wrote `settings.json.invalid` with the bad entry intact, and a subsequent provider toggle persisted without losing anything.

---

---

> [!NOTE]
> **Medium Risk**
> Changes critical settings load/persist behavior on decode failure; quarantine write failures are logged and swallowed, so users could still lose recoverability on restricted filesystems.
> 
> **Overview**
> When `settings.json` fails schema decode, the server no longer throws away the whole file and falls back to defaults (which the next `updateSettings` could persist over the user’s config).
> 
> **Salvage:** On decode failure, it prunes only `providerInstances` entries whose keys or values fail `ProviderInstanceId` / `ProviderInstanceConfig` decode (using schema decode so trimmed keys like `"  claudeAgent_work  "` stay valid), then re-decodes the rest. Unrelated fields such as `providers.*` are kept; dropped instance ids are logged.
> 
> **Quarantine:** Before applying defaults or salvaged state, the raw file is copied once per “broken file” episode to `${settingsPath}.invalid` so a later settings write cannot erase the last on-disk config. The flag resets when the file decodes successfully again.
> 
> Tests cover malformed instance + preserved provider paths, full-file quarantine on type errors, and whitespace-padded instance keys.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d6a9721153cd033a34e5b8a841f369ec4f7946e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Keep valid settings when one provider instance in settings.json is malformed
> - On decode failure in `make`, the server now backs up the raw file to `<settingsPath>.invalid` via `quarantineUnreadableSettings` before attempting recovery
> - `salvageSettingsWithoutMalformedInstances` leniently parses the JSON and drops only the `providerInstances` entries that fail to decode as `ProviderInstanceId` or `ProviderInstanceConfig`, preserving the rest of the document
> - If salvage still fails, the server logs a warning with the dropped keys and quarantine path, then falls back to in-memory defaults
> - Behavioral Change: a module-local `quarantinedThisEpisode` flag suppresses repeated `.invalid` overwrites until the next successful decode; the flag is reset on success in [serverSettings.ts](https://github.com/pingdotgg/t3code/pull/8277/files#diff-ec910be99e19e00c0e09b15484059b34164d71de8d0026f4ba98471bbbb3680e)
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6d6a972.</sup>
> 
>

### DIFF
--- a/apps/server/src/serverSettings.test.ts
+++ b/apps/server/src/serverSettings.test.ts
@@ -985,6 +985,68 @@ it.layer(NodeServices.layer)("server settings", (it) => {
     }).pipe(Effect.provide(makeServerSettingsLayer())),
   );
 
+  // `providerInstances` is a Record schema, so one bad key fails the whole
+  // settings decode. Reverting every unrelated setting because of a typo in a
+  // hand-added second provider is not an acceptable blast radius.
+  it.effect("keeps unrelated settings when one provider instance is malformed", () =>
+    Effect.gen(function* () {
+      const serverConfig = yield* ServerConfig.ServerConfig;
+      const fileSystem = yield* FileSystem.FileSystem;
+      const serverSettings = yield* ServerSettingsModule.ServerSettingsService;
+
+      yield* fileSystem.writeFileString(
+        serverConfig.settingsPath,
+        // Leading digit on the second key fails the instance-id slug pattern.
+        '{"providers":{"claudeAgent":{"binaryPath":"/tmp/claude"}},' +
+          '"providerInstances":{"claudeAgent_work":{"driver":"claudeAgent",' +
+          '"config":{"homePath":"/tmp/work"}},"2claude_broken":{"driver":"claudeAgent"}}}',
+      );
+
+      const settings = yield* serverSettings.getSettings;
+
+      assert.strictEqual(settings.providers.claudeAgent.binaryPath, "/tmp/claude");
+      assert.deepEqual(Object.keys(settings.providerInstances), ["claudeAgent_work"]);
+    }).pipe(Effect.provide(makeServerSettingsLayer())),
+  );
+
+  // The in-memory settings are what the next write persists, so a file we could
+  // not read has to survive somewhere before that write lands.
+  it.effect("quarantines a settings file it cannot read", () =>
+    Effect.gen(function* () {
+      const serverConfig = yield* ServerConfig.ServerConfig;
+      const fileSystem = yield* FileSystem.FileSystem;
+      const serverSettings = yield* ServerSettingsModule.ServerSettingsService;
+      const original = '{"providers":{"claudeAgent":{"binaryPath":42}}}';
+
+      yield* fileSystem.writeFileString(serverConfig.settingsPath, original);
+
+      const settings = yield* serverSettings.getSettings;
+      assert.strictEqual(settings.providers.claudeAgent.binaryPath, "claude");
+
+      const quarantined = yield* fileSystem.readFileString(`${serverConfig.settingsPath}.invalid`);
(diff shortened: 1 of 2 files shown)


filipef101:
Both findings were real — fixed in the latest push.

**Whitespace instance keys.** Correct, and it inverted the point of the change: `TrimmedString` trims on decode, so `"  codex_work  "` is accepted by the `Record` but rejected by `Schema.is` on the raw key. Salvage would then drop a valid instance — the loss this function exists to prevent. Keys are now validated by decoding them, the same way the values already were. Test added.

**Quarantine overwrite.** Also right. It now keeps the *first* copy of each episode: a flag is set on the first failed load and cleared when the file decodes again, so a half-saved buffer landing on the next watcher tick cannot replace the last good configuration, while a later unrelated breakage still gets its own copy.

I did not add a test for the quarantine guard. A second load in one process only happens via the file watcher, and the test I wrote asserted through `updateSettings`, which reads cached state rather than re-reading from disk — so it passed without exercising the guard at all. I removed it rather than leave a test that proves nothing; the first-copy behaviour is covered by the flag and its comment. Happy to add a watcher-driven test if you would rather have one.

PR #8337, opened 2026-08-27
fix(web): hide deleted providers with prototype keys

Lucenx9:
## What Changed

- Check provider instance settings maps for own properties before applying them.
- Treat removed custom and fork-provider snapshots as disabled while keeping valid prototype-named instances configurable.
- Add regression coverage for inherited keys and the built-in legacy fallback.

## Why

Plain object lookups can resolve inherited values such as `Object.prototype.constructor`. A deleted provider with that instance ID was therefore treated as configured and could remain visible until the next provider snapshot.

Own-property checks match the settings invariant without restricting valid provider slugs.

## UI Changes

No visual design changes. This only removes stale deleted provider entries during settings reconciliation.

## Checklist

- [x] I included before/after screenshots for any UI changes (not applicable: no visual changes)

Built with GPT-5.6 Sol in the Codex harness through T3 Code.

---

> [!NOTE]
> **Low Risk**
> Narrow change to client-side settings overlay for provider picker visibility; no auth or server persistence changes.
> 
> **Overview**
> Fixes stale provider entries staying visible in the picker when a deleted instance ID collides with inherited object keys (e.g. `constructor`, `toString`).
> 
> **`applyProviderInstanceSettings`** now uses **`Object.hasOwn`** for both `providerInstances` and legacy `providers` before reading settings. Removed custom instances (including fork defaults with no legacy row) are forced **disabled** when absent from settings; legacy **`providers.*.enabled`** applies only to built-in defaults that actually have an own legacy entry. Explicitly configured instances with prototype-like names still honor their settings.
> 
> Regression tests cover inherited-key deletion, valid prototype-named instances, fork-driver defaults, and built-in legacy fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bdeb3b5705fa0e032bb09381c93dc2e68981b3da. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix deleted providers with prototype keys in `applyProviderInstanceSettings`
> - `applyProviderInstanceSettings` in [providerInstances.ts](https://github.com/pingdotgg/t3code/pull/8337/files#diff-366ee4ff966b3b9fefcc872cdc0d10484ac52c174b5ed0e8ed59e84e97f58d5b) now uses `Object.hasOwn` before reading `providerInstances` and `legacyProviders`, so prototype members like `constructor` or `toString` are no longer treated as configured instances or legacy provider entries.
> - Default instances without an explicit config or a real legacy entry now default to disabled instead of preserving the streamed enabled state.
> - Risk: default instances absent from both `providerInstances` and `legacyProviders` will now be disabled rather than kept enabled; reviewers should check `providerInstances.ts` defaulting logic for any callers relying on the old pass-through behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bdeb3b5.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected provider instance enablement when custom instances use reserved names such as `constructor` or `toString`.
  * Removed custom and forked default instances are now correctly treated as disabled while preserving default status.
  * Legacy provider settings that disable built-in instances are now honored consistently.
  * Default instances without legacy configuration no longer retain an unintended enabled state.

### DIFF
--- a/apps/web/src/providerInstances.test.ts
+++ b/apps/web/src/providerInstances.test.ts
@@ -119,6 +119,77 @@ describe("applyProviderInstanceSettings", () => {
 
     expect(entry?.enabled).toBe(false);
   });
+
+  it.each(["constructor", "toString"])(
+    "treats a removed custom instance named %s as disabled",
+    (instanceId) => {
+      const entries = deriveProviderInstanceEntries([
+        provider({
+          provider: ProviderDriverKind.make("claudeAgent"),
+          instanceId,
+        }),
+      ]);
+      const [entry] = applyProviderInstanceSettings(entries, {
+        providerInstances: {},
+        providers: {} as never,
+      });
+
+      expect(entry?.enabled).toBe(false);
+    },
+  );
+
+  it("uses settings for a configured custom instance named constructor", () => {
+    const instanceId = ProviderInstanceId.make("constructor");
+    const entries = deriveProviderInstanceEntries([
+      provider({
+        provider: ProviderDriverKind.make("claudeAgent"),
+        instanceId,
+      }),
+    ]);
+    const [entry] = applyProviderInstanceSettings(entries, {
+      providerInstances: {
+        [instanceId]: {
+          driver: ProviderDriverKind.make("claudeAgent"),
+          enabled: false,
+        },
+      },
+      providers: {} as never,
+    });
+
+    expect(entry?.enabled).toBe(false);
+  });
+
+  it("treats a removed default instance for a fork driver as disabled", () => {
+    const driver = ProviderDriverKind.make("constructor");
+    const entries = deriveProviderInstanceEntries([
+      provider({
+        provider: driver,
+        instanceId: "constructor",
+      }),
+    ]);
+    const [entry] = applyProviderInstanceSettings(entries, {
+      providerInstances: {},
+      providers: {} as never,
+    });
+
+    expect(entry?.isDefault).toBe(true);
+    expect(entry?.enabled).toBe(false);
+  });
+
+  it("uses legacy settings for a built-in default instance", () => {
+    const entries = deriveProviderInstanceEntries([
+      provider({
+        provider: ProviderDriverKind.make("codex"),
+        instanceId: "codex",
+      }),
+    ]);
+    const [entry] = applyProviderInstanceSettings(entries, {
+      providerInstances: {},
+      providers: { codex: { enabled: false } } as never,
+    });
+
+    expect(entry?.enabled).toBe(false);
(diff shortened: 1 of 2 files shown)


Lucenx9:
@coderabbitai review

Lucenx9 (commented):
## Review: verified correct ✅

Reviewed the diff locally at `bdeb3b5` and traced every production call path. Approving the fix logic; one non-blocking note below.

### The bug is real
`settings.providerInstances?.[entry.instanceId]` resolved inherited keys like `Object.prototype.constructor`, so a deleted custom instance named after a prototype member stayed visible until the next provider snapshot reconciled it. Same failure class existed on the legacy `providers` lookup for fork-driver default instances (`legacyProviders[entry.driverKind]` where `driverKind === "constructor"`).

### The fix is right
- `Object.hasOwn` correctly guards both maps without restricting valid instance slugs — a custom instance genuinely named `constructor` still resolves its own settings row.
- The tightened default branch (`entry.isDefault && legacyProvider ? ... : false`) matches the documented invariant: only built-in defaults own a legacy `providers.*` entry, so absence there means a stale snapshot and "disabled" is correct, including for fork drivers that exist solely through `providerInstances`.
- No regression risk for built-ins: their legacy keys always exist as own properties after schema decode (`withDecodingDefault({})`), so the new `hasOwn` check always passes for them.

### Call-site safety checked
All four production callers pass the full decoded settings object, never a hand-built partial:
- `ChatView.tsx:2788`
- `ChatComposer.tsx:824`
- `SettingsPanels.tsx:1878`
- `ProjectSettingsPanel.tsx:423`

The atom fallback (`primaryServerConfigAtom?.settings ?? DEFAULT_SERVER_SETTINGS`, `state/server.ts:74`) guarantees both maps are present at runtime, so `Object.hasOwn(undefined, ...)` cannot throw.

### Non-blocking note
The old `?.` tolerated caller-built partials missing the settings fields entirely; the new code would throw on those. The types already forbid this shape and no caller constructs one, so no change needed — just worth knowing if a future caller passes an ad-hoc object.

### Verification
- `vitest run apps/web/src/providerInstances.test.ts` → 34/34 pass
- `tsc --noEmit -p apps/web` → clean
- Web-only module (`apps/web/src`); mobile surface untouched, correctly
- CI green: Check, Test ×3, Macroscope, Bugbot

---
Reviewed with GLM 5.3 Flash (Z.ai) in the opencode harness.

PR #8342, opened 2026-08-27
fix(server): Codex threads survive protocol drift on resume

### LINKED ISSUE #8322: [Bug]: Regression of #3742: Codex thread/resume rejects persisted subagent activity kind
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Open an existing Codex-backed thread in T3 Code Alpha.
2. Restart T3 Code.
3. Reopen the same thread and send another message.
4. Retry after the first failure.

The affected thread is:
2d62072c-8127-4986-b3e6-d1f3b3b77608

### Expected behavior

T3 Code resumes the persisted Codex thread and starts the next turn. If an older persisted item is incompatible, T3 Code should migrate it or provide a recovery path that preserves the conversation.

### Actual behavior

Every attempt to continue this thread fails before the turn begins. Restarting T3 Code does not repair the thread. A newly created thread works, but the affected conversation cannot be continued. This appears to be a regression or narrower variant of #3742. That issue rejected the entire subAgentActivity item and was closed after PR #3036 regenerated the Codex schema. This failure reaches the nested kind field and rejects its persisted value.

### Impact

Blocks work completely

### Version or commit

T3 Code Alpha: `0.0.34` (bundle version `0.0.34`)

### Environment

macOS 26.5.1 (25F80) T3 Code Alpha 0.0.34 Codex CLI 0.150.0 Provider: Codex Runtime mode: full access

### Logs or stack traces

```shell
ProviderAdapterProcessError: Provider adapter process error (codex) for thread 2d62072c-8127-4986-b3e6-d1f3b3b77608: Invalid payload for method 'thread/resume' during 'decode-payload'

Caused by:

SchemaError: Expected "started" | "interacted" | "interrupted"
  at ["thread"]["turns"][0]["items"][18]["kind"]

Three failed startSession attempts used the persisted resume cursor:

2026-08-26 15:32:05 MDT — trace 33495ffcf98222ea04b186ea3fef3fec
2026-08-26 15:32:28 MDT — trace ef3d3615247f735b116ab1abc0381775
2026-08-26 15:35:14 MDT — trace a30a45e45edd5fa509aa4475312551c6

provider.resume_cursor.source=persisted
provider.resume_cursor.present=true
```

[Screenshot 2026-08-26 at 15.49.25.png](https://github.com/user-attachments/assets/2442c720-ce84-4f98-ad7e-20ba52a5c686)
[Screenshot 2026-08-26 at 15.49.33.png](https://github.com/user-attachments/assets/95cad0e0-4a9a-49e5-8872-c51f58d5ccea)

### Workaround

Create a new thread in the same project and manually provide a handoff summary. Restarting the application alone does not recover the affected thread.

D3OXY:
## Problem

Codex CLI 0.150.0 added a fourth `SubAgentActivityKind` — `"completed"` — alongside `started`, `interacted`, and `interrupted`. Our generated app-server bindings are pinned to an openai/codex revision from 2026-07-19 that knows only three, so decoding a `thread/resume` response whose history contained that kind failed:

```
SchemaError: Expected "started" | "interacted" | "interrupted"
  at ["thread"]["turns"][0]["items"][18]["kind"]
```

`isRecoverableThreadResumeError` only matches "thread not found"-style messages, so a schema error was fatal. Every reopen of an affected thread failed the same way and the conversation could never be continued — the reporter's words: "Blocks work completely."

The same drift hit `item/started` / `item/completed` notifications, where decode failures were swallowed silently, so subagent lifecycle events were dropped with no trace.

## Fix

`thread/start` and `thread/resume` now decode only the three fields the session runtime actually consumes (`thread.id`, `cwd`, `model`). Codex replays an entire thread history in those responses; session state is rebuilt from notifications anyway, so history this build cannot name no longer fails the request.

`client.request` takes an optional response schema for that purpose — params are still encoded with the generated schema, only the response decode is narrowed.

Two smaller changes complete the recovery path:

- A response that still cannot be decoded now counts as a recoverable resume error, so the thread falls back to a fresh Codex session instead of becoming permanently unopenable.
- Notifications that cannot be decoded are logged as a warning instead of vanishing.

## Why not just regenerate the bindings

Regenerating to 0.150.0 moves the breakage rather than removing it: upstream also made `Thread.projectId` **required** in that release, so bindings pinned forward would reject responses from every CLI at 0.149 or older. Bindings should still be refreshed on their own cadence — this change is what absorbs the gap between refreshes. `docs/internals/providers.md` records the constraint.

## Verification

`packages/effect-codex-app-server/src/client.test.ts` drives the real client against the mock peer replaying a resume payload with `kind: "completed"`, and asserts the generated schema rejects it with `operation: "decode-payload"` while the narrow schema resumes cleanly. `CodexSessionRuntime.test.ts` covers the same shape through `openCodexThread`, plus the decode-error fallback. Both new tests fail against the previous decode path.

```
apps/server        CodexSessionRuntime/CodexAdapter/CodexCollabWire  80 passed
effect-codex-app-server                                              21 passed
typecheck + lint   clean on touched packages
```

No UI surface changes.

Fixes #8322

---
Model: Claude Opus 5. Harness: T3 Code.

> [!NOTE]
> ### Make `openCodexThread` survive protocol drift on resume by decoding minimal fields
> - `openCodexThread` now decodes only `thread.id`, `cwd`, and `model` from `thread/start` and `thread/resume` responses using a narrow `CodexThreadOpenResponse` schema, instead of requiring the full generated payload to validate.
> - `isRecoverableThreadResumeError` now treats `CodexAppServerRequestError` with operation `decode-payload` as recoverable, so decode failures during resume trigger a warning and fall back to `thread/start`.
> - `CodexAppServerClient.Service.request` accepts an optional caller-supplied response schema; undecodable notifications are dropped with a single warning per method instead of being silently ignored.
> - Added guidance in [providers.md](https://github.com/pingdotgg/t3code/pull/8342/files#diff-3d909027abc31a863f0c0e5ecef5aa50ac285620ad827db723eb0a4ac46d3f85) on handling generated Codex bindings drift via narrow schemas and recoverable decode failures.
> - Risk: the narrowed `CodexThreadOpenResponse` schema in [CodexSessionRuntime.ts](https://github.com/pingdotgg/t3code/pull/8342/files#diff-e29bf409c8e737a2d4e655f7e9e7def9c78c008d98a0a23575baa5e714c4e12c) only extracts three fields; if the runtime later needs additional response fields, the schema must be widened or decode will silently drop them.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d352756.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/CodexSessionRuntime.test.ts
+++ b/apps/server/src/provider/Layers/CodexSessionRuntime.test.ts
@@ -6,7 +6,6 @@ import * as Schema from "effect/Schema";
 import { describe } from "vite-plus/test";
 import { DEFAULT_MODEL, ThreadId } from "@t3tools/contracts";
 import * as CodexErrors from "effect-codex-app-server/errors";
-import * as CodexRpc from "effect-codex-app-server/rpc";
 import * as EffectCodexSchema from "effect-codex-app-server/schema";
 
 import {
@@ -43,9 +42,12 @@ describe("CodexSessionRuntimeIdentifierGenerationError", () => {
   });
 });
 
-function makeThreadOpenResponse(
-  threadId: string,
-): CodexRpc.ClientRequestResponsesByMethod["thread/start"] {
+/**
+ * Raw `thread/start` / `thread/resume` payload as Codex puts it on the wire.
+ * `items` accepts arbitrary history entries so tests can replay shapes newer
+ * than the generated bindings.
+ */
+function makeThreadOpenResponse(threadId: string, items: ReadonlyArray<unknown> = []): unknown {
   return {
     cwd: "/tmp/project",
     model: "gpt-5.3-codex",
@@ -57,13 +59,48 @@ function makeThreadOpenResponse(
       id: threadId,
       createdAt: "2026-04-18T00:00:00.000Z",
       source: { session: "cli" },
-      turns: [],
+      turns: items.length === 0 ? [] : [{ id: "turn-1", status: "completed", items }],
       status: {
         state: "idle",
         activeFlags: [],
       },
     },
-  } as unknown as CodexRpc.ClientRequestResponsesByMethod["thread/start"];
+  };
+}
+
+/**
+ * Mirrors the real client: params are ignored, and the raw payload is decoded
+ * with whichever response schema the caller supplied, surfacing failures the
+ * same way (`operation: "decode-payload"`).
+ */
+function makeThreadOpenClient(
+  respond: (
+    method: "thread/start" | "thread/resume",
+  ) => Effect.Effect<unknown, CodexErrors.CodexAppServerError>,
+) {
+  return {
+    request: <A, I>(
+      method: "thread/start" | "thread/resume",
+      _payload: unknown,
+      responseSchema: Schema.Codec<A, I>,
+    ) =>
+      respond(method).pipe(
+        Effect.flatMap((raw) =>
+          Schema.decodeUnknownEffect(responseSchema)(raw).pipe(
+            Effect.mapError(
+              (cause) =>
+                new CodexErrors.CodexAppServerRequestError({
+                  code: -32603,
+                  errorMessage: `Invalid payload for method '${method}' during 'decode-payload'`,
+                  method,
+                  operation: "decode-payload",
+                  cause,
+                }),
+            ),
+          ),
+        ),
+      ),
+  };
 }
 
 describe("buildTurnStartParams", () => {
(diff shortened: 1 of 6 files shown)


D3OXY:
Follow-up commit addressing review findings:

- `cwd` and `model` are now **required** in the narrow thread-open schema. Defaulting them to the requested values could record a model Codex never actually honored; both fields are required in every Codex release in range, so reading them honestly is right.
- Undecodable notifications now warn **once per method** rather than once per message. Drift in a high-frequency stream such as `item/agentMessage/delta` would otherwise emit a warning per token.
- The thread-open test double is faithful to the real client: generated param types, and the same `invalidPayload` error the client raises.
- Added a test covering the drop-and-warn path via the mock peer.

**Known gap, deliberately not in this PR:** the same enum drift affects `item/started` / `item/completed`, so a `subAgentActivity` item with `kind: "completed"` is still dropped during a live turn. Subagent registration also happens via `thread/started`, so this costs a terminal activity item rather than hanging a turn — but it is real. Fixing it needs either a bindings refresh or per-notification narrowing, which is a separate concern from the fatal resume path.

D3OXY:
Rebased onto main now that #8346 has landed. No file overlap, so it merged clean — but one assertion here went stale and is fixed in `0577e31`.

`client.test.ts` asserted that the generated bindings **reject** `kind: "completed"`. #8346 taught them that value, so the assertion tested nothing (and failed outright, since the request now succeeds where the test expected a decode error). The fixtures now use a kind no build names, which keeps exercising drift after the bindings catch up with any given release.

That is the honest division between the two PRs: #8346 fixes the variants we know about today, this one keeps the ones we do not know about yet from being fatal. Also recorded the generator's definition overrides next to the two drift rules in `docs/internals/providers.md`, so the doc describes both halves.

Codex tests green post-rebase: 80 in `apps/server`, 23 in `effect-codex-app-server`.

PR #8427, opened 2026-08-27
fix(server): back off relay client restarts

Adolanium:
## What Changed

The Cloudflare relay supervisor now waits before it respawns `cloudflared`.

If the child exits and that config is still desired, the next start waits 500ms, then 1s, 2s, up to 10s. The wait does not hold the reconcile lock, so a new `applyConfig` still starts right away and resets the backoff.

## Why

A bad tunnel token or a child that dies on start was a tight spawn loop. CPU plus process churn, with `TUNNEL_TOKEN` in the child env the whole time.

Native telemetry already backs off the same way. The relay client did not.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes connector supervision and timing for cloud tunnel availability; mis-tuned backoff or lock handling could delay recovery or allow brief spawn races, but scope is limited to managed relay restarts.
> 
> **Overview**
> Adds **exponential backoff** when the Cloudflare relay (`cloudflared`) supervisor respawns a child that exited while the same tunnel config is still desired. Delays start at **500ms**, double per failure, and cap at **10s** via exported `relayClientRestartDelay`.
> 
> The supervisor **releases the reconcile lock during the sleep**, then re-acquires it to call reconcile only if nothing else is active and the desired config still matches. **`applyConfig` resets the attempt counter to 0**, so an explicit config change can restart immediately instead of waiting through backoff.
> 
> Tests cover the delay curve and use **TestClock** so the post-exit restart test asserts no second spawn until ~500ms has passed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90f9ea0eb73ed4af9de091156298a0f1c68827af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add exponential backoff to relay client restarts in `ManagedEndpointRuntime`
> Relay client restarts now delay with exponential backoff starting at 500 ms, doubling per attempt, and capped at 10 seconds via `relayClientRestartDelay`. A new `restartAttemptRef` tracks the attempt count; `applyConfig` resets it to 0 whenever a new config is applied. After the backoff sleep, `superviseConnector` re-checks state before restarting so it only restarts if the connector is still desired and not already active. Behavioral Change: restarts are no longer immediate — there is now a delay of up to 10 seconds between process exit and respawn; existing callers expecting instant restarts will see delayed restarts.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 90f9ea0.</sup>
> 
>

### DIFF
--- a/apps/server/src/cloud/ManagedEndpointRuntime.test.ts
+++ b/apps/server/src/cloud/ManagedEndpointRuntime.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from "@effect/vitest";
 import { vi } from "vite-plus/test";
 import * as Deferred from "effect/Deferred";
+import * as Duration from "effect/Duration";
 import * as Effect from "effect/Effect";
 import * as Fiber from "effect/Fiber";
 import * as Layer from "effect/Layer";
@@ -8,6 +9,7 @@ import * as Option from "effect/Option";
 import * as PlatformError from "effect/PlatformError";
 import * as Sink from "effect/Sink";
 import * as Stream from "effect/Stream";
+import * as TestClock from "effect/testing/TestClock";
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process";
 import * as RelayClient from "@t3tools/shared/relayClient";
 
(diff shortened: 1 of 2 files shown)


MrTrigger:
I have the same issue, relay client stuck in a tight restart loop (every ~10-15ms) that never establishes a tunnel connection, and eventually OOM-kills the backend if left running. Confirms the root cause described here. Restarting/rebooting didn't fix it; toggling T3 Connect off/on repeatedly eventually got it to connect.
I'm on archlinux

t3dotgg:
> [!NOTE]

We are keeping the relay restart fix in the open [PR #8788](https://github.com/pingdotgg/t3code/pull/8788) instead of maintaining two implementations. It addresses the same rapid-exit loop with capped exponential delays, resets after stable uptime or a configuration change, and keeps configuration changes responsive while it waits. The fix is not on `main` yet.

PR #8438, opened 2026-08-27
fix(desktop): resolve WSL distro IP from the default route, not hostname -I order

### LINKED ISSUE #5211: [Bug]: Desktop stuck on "Connecting to WSL" — getDistroIp picks unreachable Docker bridge IP instead of eth0

Sanjay-doppalapudi:
## What Changed

`getDistroIpImpl` in `apps/desktop/src/wsl/DesktopWslEnvironment.ts` no longer trusts `hostname -I` ordering. It now gathers IPv4 candidates from two probes — the `src` field of `ip -4 route get 1.1.1.1` (a pure routing-table lookup, no packet is sent), then the `hostname -I` list — and validates them against the Windows-side interfaces in ranked passes, strongest signal first: a candidate inside the subnet of a WSL-named vEthernet adapter (NAT mode — checked first so a Docker bridge address that coincides with some other Windows adapter cannot pose as mirrored mode), then a candidate equal to a Windows interface address (mirrored networking, which has no NAT adapter to match; the existing `isLocalHostIpv4` detection then swaps the renderer URL to loopback), then a candidate inside any other Windows interface's subnet (renamed/custom Hyper-V switches). The ranking means a Windows-side VPN whose 10.x/172.x space overlaps an in-distro tunnel or Docker bridge cannot capture the probe while the real WSL adapter has a match. Candidates are validated as strict dotted quads (four octets, each 0–255). When no candidate matches a Windows interface, a lone candidate is still returned, but several candidates without evidence yield no address so the caller falls back to loopback instead of guessing an unreachable one. The probe logs the full candidate list plus the chosen address.

The parsing/selection is extracted into exported `parseDistroIpCandidates` / `pickDistroIp` / `windowsIpv4Interfaces` pure helpers with unit tests, following the file's existing pattern for the other probe parsers. The Windows interfaces are read through the existing `DesktopNetworkInterfaces` service (acquired in the layer, passed into the probe) rather than a direct `node:os` call, so the dependency is visible in the layer's requirements and substitutable in tests.

## Why

Fixes #5211.

The desktop app hangs on "Connecting to WSL..." forever whenever Docker has created bridge networks inside the distro: `hostname -I` lists the bridge addresses (172.17.x.x, ...) before eth0, and the code took the first IPv4 as the renderer-visible host — an address Windows can never reach. `docker network prune` only rotates which unreachable bridge IP gets picked.

Validating against the Windows-side interfaces (rather than preferring the Internet route's `src` outright) also keeps distros with an in-distro VPN/tunnel/VRF working: there the route to 1.1.1.1 reports the tunnel's address, which Windows cannot reach, while eth0 remains the correct choice.

This additionally repairs mirrored-mode detection in the Docker scenario — with a bridge IP the `isLocalHostIpv4` host-interface comparison always failed, so the loopback fallback (which the issue reporter confirmed reachable) was never taken.

Verified with `vp test run apps/desktop/src/wsl/DesktopWslEnvironment.test.ts` (44 passed, including 16 new cases covering the Docker-bridge fixture from the issue in both NAT and mirrored mode, a bridge address coinciding with another Windows adapter, the in-distro VPN case, the overlapping Windows-VPN case, custom switch names, octet validation, the lone-candidate/no-evidence rules, and the interface flattening), plus desktop typecheck, lint, and `vp fmt --check` on the touched files.

## Checklist

- [ ] I included before/after screenshots for any UI changes (N/A — no UI change)

> [!NOTE]
> ### Resolve WSL distro IP from default route instead of `hostname -I` order
> - Changes the WSL IP probe in `getDistroIpImpl` to run a combined route-lookup and hostname command, then parse and rank candidates against Windows IPv4 interfaces instead of returning the first hostname token
> - Adds `parseDistroIpCandidates` (ordered, deduplicated IPv4 extraction), `pickDistroIp` (subnet-match ranking: WSL adapter first, exact interface match second, other subnets third), `windowsIpv4Interfaces` (adapter flattener), and `ipv4ToInt`/`isValidIpv4` (strict dotted-quad validation rejecting octets > 255)
> - Wires `DesktopNetworkInterfaces` as a layer dependency so host interface data is available to every distro-IP probe
> - Adds test suites for all new helpers covering route ordering, deduplication, subnet precedence, mirrored-mode equality, and ambiguous-candidate null results
> - Updates `DesktopBackendConfiguration` test runtime to provide an empty `DesktopNetworkInterfaces` implementation
> - Behavioral Change: `getDistroIpImpl` now returns an empty option (instead of a best-effort first token) when no candidate matches ranking rules and multiple unmatched candidates exist; callers of `getDistroIp` should handle the empty case
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5675129.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Changes Windows↔WSL connectivity URL selection for all WSL backends; logic is heavily unit-tested but wrong IP choice would still block connection until loopback fallback.
> 
> **Overview**
> Fixes **#5211** by changing how the desktop app picks the WSL address the renderer uses instead of taking the first IPv4 from `hostname -I`.
> 
> **WSL distro IP** — `getDistroIp` now runs a shell probe that combines `ip -4 route get 1.1.1.1` (route `src`) with `hostname -I`, parses candidates via **`parseDistroIpCandidates`**, and chooses one with **`pickDistroIp`** against Windows IPv4 adapters from **`DesktopNetworkInterfaces`** (WSL vEthernet subnet first, then exact Windows IP for mirrored mode, then other subnets). Invalid octets are rejected; a single unmatched candidate is kept, but several unmatched candidates return **none** so configuration can fall back to loopback rather than a Docker bridge or in-distro VPN address. The WSL layer now depends on **`DesktopNetworkInterfaces`**; backend configuration tests stub that service for the sync `runSync` label probe.
> 
> The diff also includes **format-only** edits (arrow/`Effect.map` wrapping, generated schema unions, minor spacing) across mobile, server, and packages with no behavior change there.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5675129e82b26df2a7e71cf6dc0282afd28936dc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


t3dotgg:
> [!NOTE]
>

Carryover from [#5889](https://github.com/pingdotgg/t3code/pull/5889) at [34a6876d09](https://github.com/pingdotgg/t3code/commit/34a6876d09ae742561038feca2419ef9b8a6d018): retain strict IPv4 octet validation, accept only an exact `src` token rather than `prefsrc`, and return no address when `hostname -I` offers several candidates without enough evidence to choose one. Its route query and fallback share one timeout and must not start the fallback after that budget expires. Keep these checks alongside Windows-interface/VPN matching. The existing one-address NAT test does not verify the reported mirrored-mode Docker case.

juliusmarminge:

**Reason:** WSL default-route IP fix drowned in a 35-file contaminated tree; no realistic merge path.

PR #8482, opened 2026-08-28
fix(web): let empty-provider composer open settings

AmAzing129:
## What Changed

When no provider is available, the composer footer used to render a disabled **No provider available** chip. That chip is now a link to **Settings → Providers**. Hovering it shows "Open provider settings". Send stays disabled until a provider is enabled.

Web and desktop share this composer. Mobile does not: providers are configured on the host.

## Why

The empty-provider composer already told people to enable a provider in Settings, but the only control in that slot was disabled. There was no way through from the screen they were stuck on.

This keeps the same empty state and makes the obvious control do the thing the copy already promised.

## UI Changes

**Before:** disabled status chip in the model-picker slot. Placeholder says to enable a provider in Settings, but nothing in the composer navigates there.

**After:** same chip, now a link. Click opens `/settings/providers`. Tooltip: "Open provider settings".

Verified on desktop with every provider disabled: the chip appears, click lands on provider settings, enabling a provider restores the model picker.

Please drop before/after screenshots into this description (the empty composer chip, then the same chip after hover/click). I could not attach image files from the CLI.

## Checklist


juliusmarminge:

**Reason:** Empty-provider composer already opens settings on main.
**Related:** #9348

PR #8488, opened 2026-08-28
feat(pull-requests): link a PR to its thread from the PR action menu

mfattakhov:
A thread can be linked to a pull request, but only from the thread side. Standing in the PR detail panel — the place you land from the PR list, from the sidebar PR row, or from a review notification — there is no way to say "this PR belongs to that thread". After unlinking there is no way back at all, because the panel's link affordance was gated on the handoff attach target: once you unlink, the thread stops matching the PR by branch, the panel falls back to page context, and the entry point disappears with it.

This adds link and unlink entries to the PR action menu, driven by the open thread rather than by the attach target, so the entry stays put across link/unlink cycles. It reuses `thread.meta.update` through a small `useThreadPullRequestLinkActions` hook — no new command, no contract change.

Why it is worth having: any workflow where the branch a thread worked on is not the branch the PR ends up on — stacked branches, GitButler virtual branches, squash-and-rename flows — makes branch matching miss, so the thread and its PR drift apart. Linking from the PR side is the one place the association is obvious, and it is also what makes the thread settle when the PR merges. Today the only fix is to go find the thread, and a single accidental unlink removes the affordance entirely, which is a one-way door.

Model: Claude Opus 5 (1M context), harness: Claude Code.

---

> [!NOTE]
> **Low Risk**
> UI-only, capability-gated metadata updates through an existing command; no contract or auth changes.
> 
> **Overview**
> Adds **link / unlink pull request** actions to the PR detail panel’s overflow menu when a real thread is open beside the panel and the server advertises `threadPullRequestLinking`.
> 
> Linking keys off the **open thread** (`composerDraftTarget`), not the handoff attach target, so the menu stays available after unlink when branch-based matching would drop the panel back to page context.
> 
> A new `useThreadPullRequestLinkActions` hook is the dedicated path to set or clear `linkedPullRequest` via `thread.meta.update`. The panel serializes in-flight link/unlink with pending state (stable labels, disabled item), gates on server capability, and shows success/error toasts (suppressing errors for interrupted commands).
> 
> `isThreadLinkedToPullRequest` in `pullRequestDetail.logic` decides Link vs Unlink using `matchesLinkedPullRequestUrl` plus an exact URL/number fallback for hosts the parser cannot match; unit tests cover subpages, wrong PR numbers, and self-hosted URLs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aec1a54fb680a0790238bef501c54fb3ad06cc7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add link/unlink thread actions to `PullRequestDetailPanel` menu
> - Adds a conditional menu item to [PullRequestDetailPanel.tsx](https://github.com/pingdotgg/t3code/pull/8488/files#diff-3c76d43f3539848054e21911e658895287b4fe9c387c3267c21684290baebd3a) that shows "Link PR to this thread" or "Unlink PR from this thread" when the panel renders beside a concrete thread and the server reports the `threadPullRequestLinking` capability
> - Introduces `isThreadLinkedToPullRequest` in [pullRequestDetail.logic.ts](https://github.com/pingdotgg/t3code/pull/8488/files#diff-2c5f917dba9decc3bd48ec3231f7b0db073e8b642379cf000265960a03718c18) to detect the linked state via tolerant URL matching, with strict-equality fallback for unparseable hosts
> - Adds the `useThreadPullRequestLinkActions` hook in [useThreadPullRequestLink.ts](https://github.com/pingdotgg/t3code/pull/8488/files#diff-71ae110c17b3ecea6ac52d3753970f0dce298a349f1d5e31a4b9ec97996391b5) which updates `linkedPullRequest` in thread metadata through `updateMetadata`, suppressing error toasts for interrupted commands and blocking concurrent operations with pending state
> - Behavioral Change: menu items only appear when `threadPullRequestLinking` server config is enabled; existing panels without a concrete thread context show no new items
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized aec1a54.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/pullRequest/PullRequestDetailPanel.tsx
+++ b/apps/web/src/components/pullRequest/PullRequestDetailPanel.tsx
@@ -1,5 +1,8 @@
 import { scopedThreadKey, scopeProjectRef } from "@t3tools/client-runtime/environment";
-import { squashAtomCommandFailure } from "@t3tools/client-runtime/state/runtime";
+import {
+  isAtomCommandInterrupted,
+  squashAtomCommandFailure,
+} from "@t3tools/client-runtime/state/runtime";
 import type {
   EnvironmentId,
   PullRequestAction,
@@ -8,6 +11,7 @@ import type {
   PullRequestRef,
   PullRequestState,
   ScopedThreadRef,
+  ThreadLinkedPullRequest,
 } from "@t3tools/contracts";
 import {
   ArrowDownUpIcon,
@@ -26,6 +30,7 @@ import {
   GitPullRequestIcon,
   HammerIcon,
   LayersIcon,
+  Link2Icon,
   MessageCircleQuestionIcon,
   MessageSquareIcon,
   LinkIcon,
@@ -35,6 +40,7 @@ import {
   RefreshCwIcon,
   ServerIcon,
   TriangleAlertIcon,
+  UnlinkIcon,
 } from "lucide-react";
 import {
   lazy,
@@ -51,12 +57,17 @@ import {
 import { type DraftId, useComposerDraftStore } from "~/composerDraftStore";
 import { useNewThreadHandler } from "~/hooks/useHandleNewThread";
 import { useCopyToClipboard, writeTextToClipboard } from "~/hooks/useCopyToClipboard";
-import { changeRequestRepositoryUrl, gitHubPullRequestBrowserUrl } from "~/lib/openPullRequestLink";
+import { useThreadPullRequestLinkActions } from "~/hooks/useThreadPullRequestLink";
+import {
+  changeRequestRepositoryUrl,
+  gitHubPullRequestBrowserUrl,
+  matchesLinkedPullRequestUrl,
+} from "~/lib/openPullRequestLink";
 import { usePreparePullRequestThreadAction } from "~/lib/sourceControlActions";
 import { cn } from "~/lib/utils";
 import { readLocalApi } from "~/localApi";
 import type { ReviewCommentContext } from "~/reviewCommentContext";
-import { useProjects } from "~/state/entities";
+import { useProjects, useServerConfigs, useThreadShell } from "~/state/entities";
 import { useEnvironments } from "~/state/environments";
 import { useEnvironmentQuery } from "~/state/query";
 import { useLiveRefresh } from "~/hooks/useLiveRefresh";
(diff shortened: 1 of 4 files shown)


mfattakhov:
Both findings were real and are fixed in `2c0f115`.

**Capability gate.** The link/unlink items now render only when the thread's environment reports `threadPullRequestLinking`. The read is reactive (`useServerConfigs()` keyed by `attachThreadRef.environmentId`), because this popup is React-rendered and capabilities arrive after connect. `readEnvironmentSupportsPullRequestLink` in `state/entities.ts` is gone from this PR — the non-reactive reader had no caller here, so it was dead code rather than a gate.

**Match too strict.** `isLinkedToThisPullRequest` now uses `matchesLinkedPullRequestUrl`, the same host + repository + number match the thread side uses, so a link stored from an agent-written `/pull/42/files` href resolves to the canonical PR and the menu offers Unlink rather than a second Link.

mfattakhov:
Two more findings from the second pass, both fixed in `efde976`.

- **Concurrent link/unlink.** `thread.meta.update` is last-write-wins, so two of these in flight could land reversed and leave the thread on the earlier choice. The menu item now holds an in-flight lock: one request at a time, and the label reads "Linking..." / "Unlinking..." while it runs.
- **Unparseable hosts never offering Unlink.** This menu stores `detail.url` verbatim, so a host `parseChangeRequestUrl` does not recognise would link once and never show the way back. `isLinkedToThisPullRequest` now falls back to exact number + URL equality when the shared match cannot parse both sides.

The "Link menu skips capability gate" comment re-anchored onto the new commit but its cited locations still point at `3f5b329`; the gate landed in `2c0f115` (`supportsPullRequestLink`, read reactively from the thread environment's `threadPullRequestLinking`) and is still in place.

mfattakhov:
All bot checks are green on this branch. The remaining approvability note was about coverage, so `0aea50a` adds it: the menu's linked/not-linked decision moved into `pullRequestDetail.logic.ts` as `isThreadLinkedToPullRequest`, with tests for a link stored from a subpage, a different PR in the same repository, a self-hosted host the change-request parser cannot read (the exact-URL fallback), and the missing-either-side cases.

mfattakhov:
Fixed in `aec1a54fb`: the pending state now records *which* operation is running (`"link" | "unlink" | null`) rather than a bare boolean. The thread shell picks up the new link before the command settles, so a label read off the live state flipped to the opposite verb mid-flight. The running operation now outranks live state until it settles, and the item keeps the verb the reader pressed.

juliusmarminge:
Superseded by #10839, which landed link/unlink in the PR More menu (and multi-PR linking). Closing this earlier single-link menu attempt as superseded.

PR #8498, opened 2026-08-28
fix(server): roll back failed service installs

### LINKED ISSUE #8476: Failed service install leaves a dead unit that status reports as healthy
## What happened

`t3 service install` failed at `enable-linger` on a host with no polkit authority. It left a unit that was enabled but never started, and nothing cleaned up.

`t3 service status` then reported the service as installed and current. Re-running `t3 service install` printed "already installed" and did nothing. The only way out was `systemctl --user restart t3code.service`, which is not mentioned anywhere in the error.

Until then the server ran unmanaged, so it advertised no `serverSelfUpdate` capability and clients offered "Copy update command" instead of a one-click update.

## Diagnosis

Two changes in [d3037064e](https://github.com/pingdotgg/t3code/commit/d3037064e61a9f059eafbd4f9869679779bd2a7c) (#5181) combine to produce this.

**The install rollback was removed.** v0.0.31 wrapped activation in `Effect.tapError(() => rollbackFailedInstall(previousUnit))`, which on a fresh install ran `systemctl --user disable --now`, removed the unit file and reloaded systemd. Its comment stated the reason:

> If any activation step fails, remove the unit again: a leftover file would make service status report it as installed even though it was never enabled or lingered.

Current code has no equivalent. The only handler is at [bootService.ts:668-670](https://github.com/pingdotgg/t3code/blob/b0ae3f3a8/apps/server/src/cloud/bootService.ts#L668-L670):

    Effect.tapError(() =>
      installed ? runSteps(manager.restart).pipe(Effect.ignore) : Effect.void,
    )

`installed` means a unit file existed before the launcher, state and unit writes. Repairing an existing install gets a best-effort restart. A first install gets `Effect.void`, so nothing happens.

**`enable-linger` moved before the start.** v0.0.31 ordered activation daemon-reload, enable, restart, enable-linger, so a linger failure came after a successful start. Current order is daemon-reload, enable, enable-linger, restart ([bootService.ts:225-243](https://github.com/pingdotgg/t3code/blob/b0ae3f3a8/apps/server/src/cloud/bootService.ts#L225-L243)), which puts a step that can fail on an unrelated precondition directly between "enabled" and "started".

**`status` cannot see the problem.** It reports `installed` from the unit file existing, and computes `current` from file contents, the launcher, the pinned runtime entry, the sentinel and the state file ([bootService.ts:698-723](https://github.com/pingdotgg/t3code/blob/b0ae3f3a8/apps/server/src/cloud/bootService.ts#L698-L723)). It never runs `is-enabled` or `is-active`. Every one of those files was written correctly before activation failed, so status reports installed and current.

`reconcileService` then short-circuits on exactly that ([cli/service.ts:35-38](https://github.com/pingdotgg/t3code/blob/b0ae3f3a8/apps/server/src/cli/service.ts#L35-L38)), so `t3 service install` and `t3 service update` both refuse to do anything. The CLI insists the service is fine while it is not running.

This is the failure the removed rollback comment predicted, word for word.

## Steps to reproduce

1. On a Linux host, make `enable-linger` fail. Running with no polkit authority does it, and so does removing `loginctl` from PATH.
2. Run `t3 service install` with no service previously installed.
3. Run `t3 service status`, and observe it reports the service as installed and current.
4. Run `t3 service install` again, and observe it reports the service as already installed.
5. Check `systemctl --user is-active t3code.service`.

## Version

0.0.36-nightly.20260827.1207

## Environment

Arch Linux x86_64, systemd 260, Node 26.7.0 via mise. No polkit authority running.

## Evidence

    $ t3 service install
    ERROR: BootServiceCommandError: Background setup failed while enabling lingering for this user (exit code 1).

    $ systemctl --user is-enabled t3code.service
    enabled
    $ systemctl --user is-active t3code.service
    inactive

    $ ls ~/.t3/runtime
    service-launcher.mjs  service-state.json  versions

    $ cat ~/.t3/runtime/service-state.json
    { "protocol": 2, "activeVersion": "0.0.36-nightly.20260827.1207" }

The unit file, launcher, pinned version and state file were all present. Only the start had not run.

## Related issues

#4894 is the failure that exposed this. Its report of a clean rollback was accurate for t3@0.0.31, before #5181.

## Fix applied or workaround

    systemctl --user restart t3code.service

The service started at once and the server came up launcher-managed, advertising `serverSelfUpdate: "boot-service"`. Re-running the installer does not work, because status already considers the install complete.

## Filed by

Claude Opus 5 in Claude Code, from a `t3 triage` run. Codex (gpt-5.6-sol) analysed the source and fact-checked the write-up.

Gigioxx:
## Problem

A fresh systemd service install could fail during activation after writing and enabling the unit. The leftover unit made `t3 service status` report a current install, so retries short-circuited while the service remained inactive.

## Fix

Roll back failed fresh systemd activations through the existing service-manager cleanup steps: disable the unit, remove its file, and reload systemd. Existing-install repair recovery and documented launchd behavior stay unchanged.

## Verification

- `vp test run apps/server/src/cloud/bootService.test.ts` (21 tests)
- changed-file `vp lint`
- changed-file `vp fmt --check`
- `git diff --check`
- Codex review: no findings

Fixes #8476

---

> [!NOTE]
> **Medium Risk**
> Touches boot-service install failure recovery on Linux; incorrect rollback could leave systemd in a bad state, but changes are scoped to fresh installs and mirror existing uninstall steps.
> 
> **Overview**
> **Failed fresh systemd installs** no longer leave a half-enabled unit on disk. When `install` errors during activation and the service was not already present, the error handler now runs the same **deactivate → remove unit file → finalize** path as uninstall, instead of doing nothing.
> 
> **Repair behavior is unchanged**: if activation fails on an **existing** install, the flow still attempts **restart** only. **launchd** fresh-install failures still do not roll back (documented behavior).
> 
> A new test simulates activation failing at `loginctl enable-linger` and asserts `installed` is false and the expected `systemctl` rollback sequence runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b168f578ec81e1b8be8cf3347f5558582588970. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Roll back failed systemd service installs in `bootService`
> When a fresh `systemd` install fails during activation, the error handler in the `install` effect now runs the full rollback sequence: deactivate steps, remove the unit file at `unitPath`, and finalize steps. All rollback steps ignore their own errors. Previously the handler did nothing for uninstalled services. Adds a test verifying the expected `systemctl` command sequence on rollback.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7b168f5.</sup>
> 
>

### DIFF
--- a/apps/server/src/cloud/bootService.test.ts
+++ b/apps/server/src/cloud/bootService.test.ts
@@ -223,6 +223,22 @@ it.layer(NodeServices.layer)("boot service install", (it) => {
     }),
   );
 
+  it.effect("rolls back a fresh install when activation fails", () =>
+    Effect.gen(function* () {
+      const { service, commands, control } = yield* makeHarness();
+      control.failCommand = "loginctl enable-linger";
+
+      expect((yield* service.install.pipe(Effect.flip))._tag).toBe("BootServiceCommandError");
+      expect((yield* service.status).installed).toBe(false);
+      expect(commands.filter((command) => command.startsWith("systemctl "))).toEqual([
+        "systemctl --user daemon-reload",
+        "systemctl --user enable t3code.service",
+        "systemctl --user disable --now t3code.service",
+        "systemctl --user daemon-reload",
+      ]);
+    }),
+  );
+
   it.effect("restarts an installed service when repair fails", () =>
     Effect.gen(function* () {
       const { service, commands, control } = yield* makeHarness();
--- a/apps/server/src/cloud/bootService.ts
+++ b/apps/server/src/cloud/bootService.ts
@@ -666,7 +666,15 @@ export const make = Effect.fn("cloud.boot_service.make")(function* (input: {
       yield* runSteps(manager.activate);
     }).pipe(
       Effect.tapError(() =>
-        installed ? runSteps(manager.restart).pipe(Effect.ignore) : Effect.void,
+        installed
+          ? runSteps(manager.restart).pipe(Effect.ignore)
+          : manager.kind === "systemd"
+            ? Effect.gen(function* () {
+                yield* runSteps(manager.deactivate).pipe(Effect.ignore);
+                yield* fs.remove(unitPath).pipe(Effect.ignore);
+                yield* runSteps(manager.finalize).pipe(Effect.ignore);
+              })
+            : Effect.void,
       ),
     );
     return plan;


juliusmarminge:
Superseded by #9602, which landed the incomplete headless Connect setup diagnostics and closed #8476 via prerequisite checks and repairable status instead of rolling back a failed fresh install.

PR #8503, opened 2026-08-28
fix(web): stop the resize cursor sticking when the panel changes mid-drag

alundgren:
## Problem

The app can get stuck showing the `col-resize` cursor everywhere in the main content area, with text selection disabled alongside it. It persists for the rest of the session. Moving the pointer outside the window restores the normal cursor, and coming back shows the resize cursor again.

`useResizableWidth()` sets `cursor: col-resize` and `user-select: none` on `document.body` when a drag starts, and only clears them from `onPointerUp` and `onPointerCancel`. `PreviewPanelShell` owns the hook but renders the drag handle only while the panel is inline and not maximized. Maximizing the panel (`rightPanel.toggleMaximized`), switching its mode, or closing it therefore removes the handle mid-drag. After that no pointer handler can ever run, and both global styles stay behind.

## Fix

Clean up from a ref callback on the drag handle, reusing the existing `releasePointer()`. React runs it when the element leaves the DOM, so it covers both the handle being removed while the hook stays mounted and the panel unmounting outright, and it cannot drift out of sync with the condition that renders the handle. `releasePointer()` already no-ops without an active drag and tolerates a double release, so normal pointer-up and pointer-cancel are unaffected and the final width still persists.

## Tests

`useResizableWidth` had no coverage. Added `useResizableWidth.test.tsx`, using the DOM-free React harness this suite already uses in `components/preview/PreviewView.test.tsx`. The harness mirrors `PreviewPanelShell`: the hook lives in the parent and the handle is rendered conditionally beneath it.

Three cases: a normal pointer-down/pointer-up drag, the handle being removed while the hook stays mounted, and the whole panel unmounting mid-drag. The last two fail on `main` with the body still holding `col-resize`.

- `vp test run apps/web/src/hooks/useResizableWidth.test.tsx`
- `vp test run apps/web/src/components/preview/PreviewPanelShell.test.ts apps/web/src/components/preview/PreviewView.test.tsx`
- targeted lint for the changed files

## Visual evidence

None. The symptom is the shape of the OS pointer, which a screenshot does not capture.

## Not covered

Whether pointer capture can also be lost without the handle being removed, for example on window blur or a native menu opening mid-drag.

Made with Claude Opus 5 in Claude Code.

> [!NOTE]
> ### Fix resize cursor sticking when panel closes mid-drag in `useResizableWidth`
> - The `useResizableWidth` hook now exposes a `ref` callback on its returned handlers object. Attaching it to the drag handle element ensures that an active drag is terminated and global `cursor`/`user-select` styles are cleared if the handle unmounts mid-drag.
> - The cleanup function returned by the ref calls `releasePointer(state.pointerId)` when the element is removed while dragging.
> - New tests in [useResizableWidth.test.tsx](https://github.com/pingdotgg/t3code/pull/8503/files#diff-45feddf54052b842123ac8f1ed51e7a48fdfa04fc31732d2feeda689216bb92c) verify style cleanup on pointer up, handle unmount mid-drag, and full panel unmount mid-drag.
> - Behavioral Change: `ResizableWidthHandlers` now requires consumers to attach the new `ref` to the drag handle; existing consumers that do not wire up the ref will not get the mid-drag unmount cleanup.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b8702f4.</sup>
> 
>

---

> [!NOTE]
> **Low Risk**
> Localized UI/pointer-drag cleanup with idempotent release logic; no auth, data, or API changes.
> 
> **Overview**
> Fixes a session-long **col-resize** cursor and disabled text selection when the preview panel is maximized, closed, or otherwise removes its drag handle while a resize drag is still active. Previously, global `document.body` styles were only cleared on pointer up/cancel, so unmounting the handle left no handler to run.
> 
> **`useResizableWidth`** now returns a **`ref` callback** on `handlers` that, on handle unmount, calls the existing **`releasePointer`** path to drop pointer capture and remove body `cursor` / `user-select`. Consumers that spread handlers onto the handle (e.g. `RightPanelResizeHandle`) pick this up without extra wiring.
> 
> Adds **`useResizableWidth.test.tsx`** with a DOM-free React harness (aligned with preview tests): normal pointer-up cleanup, handle removed mid-drag, and full unmount mid-drag.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b8702f403e04f9bd013b8cf2124ad2e791fdeed9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


juliusmarminge:
Superseded by #10461, which landed on `main`.

That PR clears the stuck `col-resize` cursor / text-selection lock from `useResizableWidth` when a panel resize is interrupted (unmount, lost pointer capture, or window blur), which is the same mid-drag stuck-cursor bug this PR targets. Closing this as superseded — please reopen if anything here is still needed beyond that.

PR #5198, opened 2026-08-01
fix(client-runtime): tolerate transient foreground probe timeouts

Zeus-Deus:
## Problem

When the host is temporarily resource constrained, a foreground health probe can time out even though the existing session is still valid. Immediately replacing that session can start a reconnect while the same host is still stalled, leaving the client unusable until it recovers.

## Fix

Keep the existing desktop/web session after the first foreground probe timeout. A successful probe resets the count, while a second consecutive timeout still triggers normal recovery. Definite probe failures and mobile resume behavior are unchanged.

## Testing

- Focused connection supervisor suite: 35 tests passed
- Client-runtime typecheck
- Targeted lint and formatting checks

Model: GPT-5; harness: Codex.

---

> [!NOTE]
> **Medium Risk**
> Changes live connection monitoring and reconnect timing in client-runtime; behavior is well-covered by supervisor tests but affects when sessions are replaced under load.
> 
> **Overview**
> **Foreground health probes** on `application-active` no longer tear down the session on a single **15s** timeout when the host is briefly stalled. The supervisor keeps the existing lease, logs a warning, and only triggers normal reconnect recovery after **two consecutive** timeouts (`FOREGROUND_PROBE_TIMEOUTS_BEFORE_RECONNECT`).
> 
> A successful probe resets the counter. **Definite probe failures** and **`application-active-probe`** (mobile, **3s** timeout) behavior are unchanged.
> 
> Tests cover: one timeout stays connected; two timeouts → backoff/release; success between timeouts resets the count.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3fdbd155b93f90d9b9e24ca89ed010b3ff097887. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Tolerate transient foreground liveness probe timeouts in `EnvironmentSupervisor`
> - During `application-active` wakeups, a single probe timeout no longer triggers a reconnect; the supervisor only reconnects after 2 consecutive timeouts (`FOREGROUND_PROBE_TIMEOUTS_BEFORE_RECONNECT = 2`).
> - A successful probe resets the timeout counter to 0, so isolated timeouts are always tolerated.
> - `application-active-probe` signals retain the existing fail-fast behavior on timeout.
> - Two new tests cover consecutive timeout reconnection and counter reset after a successful probe.
>
> 
>
[...]

PR #8522, opened 2026-08-28
fix(client-runtime): tolerate transient foreground probe timeouts

### LINKED ISSUE #7231: [Bug]: One slow foreground health probe tears down a live session and disables the composer
### Before submitting

### Area

### Steps to reproduce

1. Open the desktop app (or the web client) against a local environment.
2. Give the backend real work: a fleet of subagents, a large repository, or any thread busy enough to keep the server's event loop saturated for a few seconds.
3. Switch to another window and come back, which raises an `application-active` wakeup.
4. Watch the environment banner and the composer.

There is no deterministic repro. It needs the backend to be slow enough that one health probe misses its deadline, which on a loaded Windows host happens often.

### Expected behavior

A backend that is busy but alive should stay connected. The WebSocket is still open and no close event arrived, so the client has positive evidence the transport is fine. One slow answer is not a disconnect.

### Actual behavior

The environment drops to "Failed to connect. Reconnecting..." with the reason "<label> did not respond to a connection health check", and the composer is disabled until it recovers. The backend was never gone.

The path is in `packages/client-runtime/src/connection/supervisor.ts`:

- `supervisor.ts:34` — `CONNECTION_PROBE_TIMEOUT = "15 seconds"`.
- `supervisor.ts:421` — on an `application-active` wakeup the supervisor runs `lease.session.probe`.
- `supervisor.ts:428` — on timeout it fails with `ConnectionTransientError({ reason: "timeout" })`.
- `supervisor.ts:451` — that exit is propagated, `monitorConnectedLease` returns, and the lease is replaced.

A single timeout is enough. There is no second look, and the code does not distinguish a deadline from a definite failure even though it constructs the two differently. The socket state is never consulted.

The user-visible damage is mostly the composer. Losing a few seconds of connection would be tolerable; losing the ability to type is not, and it lands exactly when someone has come back to the app to send a message.

### Impact

### Version or commit

main @ 4c1d99d7f

### Environment

Windows 11, T3 Code Nightly desktop, local environment. The supervisor is shared by web, desktop, and mobile.

### Logs or stack traces

```shell
Banner text:

  Failed to connect. Reconnecting...
  Reason: <environment label> did not respond to a connection health check.
```

_No response_

### Workaround

Wait. It reconnects on its own, usually within seconds. There is nothing to do in the meantime because the composer is disabled.

## Related

#3553 reported this and was closed by #4137, which made the probe itself lightweight (`serverProbe` instead of `serverGetConfig`). That was the right fix for probe cost, and it helps, but it does not address the teardown rule: one timeout still replaces a live session. On a host that is starved rather than merely doing extra work, a cheap probe misses its deadline too.

#5198 proposed tolerating a transient timeout. It has been sitting with merge conflicts since 1 Aug.

Backend starvation is tracked separately in #4773; this issue is about the client's reaction to it, which is worth fixing on its own because the client cannot assume the server will ever be fast.

Radu028:
## What Changed

The connection supervisor no longer replaces a live desktop or web session after a single foreground health probe timeout. It reconnects only after two consecutive timeouts, and a successful probe resets the count. When it does reconnect, it still skips the first backoff rung, the same as a definite probe failure. Definite probe failures and the mobile resume probe are unchanged.

## Why

Under temporary host load, one slow probe (15s) tore down an otherwise healthy WebSocket, showed "Failed to connect. Reconnecting..." and disabled the composer while the same host was still stalled.

Fixes #7231. Same approach as #5198, which no longer applies to `main`.

## Verification

- `vp test run packages/client-runtime/src/connection/supervisor.test.ts` (36 passed; covers single-timeout tolerance, reconnect on the second timeout, and reset after a successful probe)
- client-runtime typecheck, targeted lint and fmt

## Checklist

- [x] No UI changes

Built with Claude Fable 5.1 in Claude Code.

---

> [!NOTE]
> **Medium Risk**
> Changes connection recovery timing for desktop foreground probes (up to ~30s before reconnect on hung probes) in a core client-runtime path, though definite failures and mobile probes are unchanged.
> 
> **Overview**
> **Desktop/web foreground health checks** no longer tear down a live session on a single 15s probe timeout. `EnvironmentSupervisor` now requires **two consecutive** stalled `application-active` probes before treating the transport as dead and reconnecting (with the same immediate retry behavior as a definite probe failure). A probe that completes successfully **resets** the timeout counter, and the first timeout only emits a warning while keeping the existing session.
> 
> **Mobile `application-active-probe`** behavior is unchanged: timeouts still fail fast on the shorter 3s window. Tests were expanded to cover single-timeout tolerance, reconnect on the second timeout, and counter reset after a successful probe.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6ace61a6a1ca605e9bd5855fd0927903df3bae9e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Tolerate transient foreground probe timeouts in `EnvironmentSupervisor`
> Adds `FOREGROUND_PROBE_TIMEOUTS_BEFORE_RECONNECT` (set to 2) so the desktop foreground probe tolerates a single timeout before forcing a reconnect. A successful probe resets the timeout counter. Mobile explicit foreground probes (`application-active-probe`) still fail fast on timeout.
> - Risk: desktop reconnects now trigger only after 2 consecutive timeouts instead of 1, so a probe that permanently hangs takes up to 30s to recover; check `foregroundProbeTimeoutCount` logic in [supervisor.ts](https://github.com/pingdotgg/t3code/pull/8522/files#diff-9cea0e0ed5756232bad1eb5801220443f2c1a6cbabaedc22ad335b39a79c7de9).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 97442ad.</sup>
> 
>

### DIFF
--- a/packages/client-runtime/src/connection/supervisor.test.ts
+++ b/packages/client-runtime/src/connection/supervisor.test.ts
@@ -994,10 +994,19 @@ describe("EnvironmentSupervisor", () => {
     }).pipe(Effect.provide(TestClock.layer())),
   );
 
-  it.effect("uses the full tolerance window for a stalled desktop foreground probe", () =>
+  it.effect("reconnects only after consecutive desktop foreground probe timeouts", () =>
     Effect.gen(function* () {
+      const firstProbeStarted = yield* Deferred.make<void>();
+      const secondProbeStarted = yield* Deferred.make<void>();
+      const probeCount = yield* Ref.make(0);
       const harness = yield* makeHarness({
-        probe: (attempt) => (attempt === 1 ? Effect.never : Effect.void),
+        probe: () =>
+          Ref.updateAndGet(probeCount, (count) => count + 1).pipe(
+            Effect.tap((count) =>
+              Deferred.succeed(count === 1 ? firstProbeStarted : secondProbeStarted, undefined),
+            ),
+            Effect.andThen(Effect.never),
+          ),
       });
       const supervisor = yield* EnvironmentSupervisor.make(TARGET_ENTRY, {
         initiallyDesired: true,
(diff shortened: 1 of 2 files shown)


t3dotgg:
> [!NOTE]
>

Closing in favor of [#7233](https://github.com/pingdotgg/t3code/pull/7233), which retries a timed-out foreground probe automatically within a bounded window. This version waits for another visibility change before it probes again, so a stalled RPC session can remain marked connected indefinitely. The replacement preserves the separate mobile resume behavior. Review continues there.

PR #8281, opened 2026-08-26
fix(server): symlinked directories show up in the path picker

PengLx:
## What Changed

`WorkspaceEntries.browse` — the single server-side implementation behind the web command palette / add-project picker and mobile's `AddProjectScreen` — filtered `readdir` entries with `dirent.isDirectory()`. `readdir` reports a symlink by its own type and never its target's, so every symlink pointing at a directory was silently dropped from the listing.

Candidates now include symlinks, and each link gets one follow-the-link `stat` before it is offered as browsable. Broken links, links to files, and unreadable targets still drop out, so the visible set is unchanged apart from the symlinked directories that were missing. Plain directories pay no extra syscall, and the `stat` calls only happen for symlinks that already passed the name-prefix and hidden-file filters (concurrency capped at 16).

No contract change — `FilesystemBrowseEntry` is still `{ name, fullPath }`, and the client-side `filterFilesystemBrowseEntries` filters by name only.

## Why

Reported on Linux: picking a project path that goes through a symlinked directory is impossible in the UI. The link never appears in the dropdown, so you have to type the path past it by hand before autocomplete works again. Symlinked directories are everywhere on Linux — dotfile managers, `/mnt` mount layouts, pnpm store paths, worktree setups.

Repro:

```
mkdir -p /tmp/real-project
ln -s /tmp/real-project ~/linked-project
```

Then browse `~/` in the path picker: `linked-project` is absent before this change, present after.

## UI Changes

No visual change — the same dropdown now lists entries it was incorrectly omitting.

## Testing

Added `follows symlinks that point at directories and skips the rest` to `WorkspaceEntries.test.ts`, covering a symlinked directory (listed), a symlink to a file (skipped), and a broken symlink (skipped). Verified it fails against the old filter.

- `vp test run apps/server/src/workspace/WorkspaceEntries.test.ts` — 31 passed
- `tsgo --noEmit` in `apps/server` — no errors
- `vp lint apps/server/src/workspace/` — clean

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] I included before/after screenshots for any UI changes (n/a — no visual change)
- [x] I included a video for animation/interaction changes (n/a)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Scoped to filesystem browse listing with the same API contract; extra `stat` calls only on symlink candidates that already passed filters.
> 
> **Overview**
[...]

PR #8539, opened 2026-08-28
fix(server): include symlink directories in project browse

### LINKED ISSUE #8408: Add Project browser and path autocomplete omit symlinked directories
## What happened

The Add Project file browser and the path autocomplete do not show symlinks to
directories, so I could not browse into my projects at all.

On the server machine, `~/Projects` is a symlink to `/work/projects` (a mounted
drive). When I open Add Project and browse my home directory, the `Projects`
symlink never appears as a folder, so none of my projects are reachable through
the UI. The only ways in are typing the fully resolved real path by hand, or
adding the project on the server machine directly.

Surface: desktop app 0.0.35 on macOS, connected via the SSH-launch flow to a
Linux server running `t3@0.0.35`.

## Diagnosis

Single root cause in `apps/server/src/workspace/WorkspaceEntries.ts:222`, inside
`browse()`:

```ts
for (const dirent of dirents) {
  if (
    dirent.isDirectory() &&
    dirent.name.toLowerCase().startsWith(lowerPrefix) &&
    (showHidden || !dirent.name.startsWith("."))
  ) {
```

`readdir(..., { withFileTypes: true })` returns dirents with `lstat` semantics:
for a symlink, `dirent.isDirectory()` describes the link itself, not its target,
and is therefore `false`. A symlink pointing at a directory is classified as
neither a file nor a directory and is filtered out entirely.

The Add Project directory picker and the path autocomplete are the same code
path — both are served by the `filesystem.browse` RPC
(`packages/contracts/src/rpc.ts:224`, `WsFilesystemBrowseRpc` at line 674), and
`browse()` above is its only implementation. The client
(`packages/client-runtime/src/state/filesystem.ts`,
`filterFilesystemBrowseEntries`) filters only on the name prefix, so all
directory classification happens in that one server-side loop. That is why both
surfaces fail identically.

The call succeeds — it does not error. Entries are silently dropped, so nothing
surfaces in the logs and the directory just looks empty of that folder.

Suggested fix, matching the smallest-scope proposal in the earlier report: keep
dirents where `isDirectory()` is true, and additionally keep
`isSymbolicLink()` dirents whose `stat()` target is a directory (guarding
`ELOOP`/`ENOENT` for broken or cyclic links, which should be excluded). One
extra `stat()` per symlink candidate, on an interactive-only path. No contract
change is required — `FilesystemBrowseEntry` is `{ name, fullPath }` and
`fullPath` should stay the symlink path the user typed, not the resolved target.

This is still present on `main` as of 2026-08-27: line 222 is unchanged.

## Steps to reproduce

On the machine running the server:

```
mkdir -p /tmp/real
ln -s /tmp/real ~/link-to-real
```

1. Open Add Project.
2. Browse to the home directory (`~/`).
3. `link-to-real` does not appear in the listing. `/tmp/real` browsed directly
   does appear.
4. Same in the path autocomplete: typing `~/link` produces no completion.

Reproducible every time. Platform-independent — nothing macOS-specific about it.

## Version

0.0.35 (server `t3@0.0.35`, launched as `npx t3`); desktop app 0.0.35 on macOS.

## Environment

Server: Linux x64 (kernel 7.1.9), Node v26.7.0, server running on
127.0.0.1:3773. Client: desktop app 0.0.35 on macOS, connected over the
SSH-launch flow.

## Evidence

```shell
# Real symlinks on the server machine:
~/Projects  -> /work/projects
~/Worktrees -> /work/worktrees

# Node dirent semantics that browse() relies on:
link-to-real   dirent.isDirectory=false  dirent.isSymbolicLink=true  | stat(target).isDirectory=true
plaindir       dirent.isDirectory=true   dirent.isSymbolicLink=false | stat(target).isDirectory=true

# Applying the v0.0.35 filter vs. a symlink-aware filter to the home directory:
v0.0.35 browse("~/") returns:  Documents, Downloads, Music, Pictures, Videos
symlink-aware filter returns:  Documents, Downloads, Music, Pictures, Projects, Videos, Worktrees
MISSING:                       Projects, Worktrees

# server.trace.ndjson — the RPC succeeds; entries are dropped silently:
{"type":"effect-span","name":"WorkspaceEntries.browse","durationMs":0.457326,"exit":{"_tag":"Success"}}
{"type":"effect-span","name":"ws.rpc.filesystem.browse","attributes":{"rpc.method":"filesystem.browse",
 "rpc.aggregate":"workspace"},"durationMs":0.549747,"exit":{"_tag":"Success"}}
# (2026-08-27T14:33:04.409Z and 14:33:29.650Z)
```

Nothing appears in `server.log` for these calls — there is no failure to log.

## Related issues

- #2135 — same root cause and same line, filed 2026-04-17 as a feature request.
  Its PR #2136 was closed unmerged on 2026-07-17, and the issue was converted to
  a discussion and locked on 2026-08-15, so the report is no longer in the
  tracker and cannot be commented on. Refiling as a bug: this is not a missing
  enhancement but a directory that cannot be browsed at all. Scope here is
  deliberately narrower than #2135 — symlinks only, no macOS Finder aliases, no
  contract or icon changes.
- #6023 — external drives missing from the directory picker; also converted to a
  discussion on 2026-08-15. Different cause (mount points, not symlinks), same
  user-facing shape of "a valid directory is unreachable through the picker".
- #8234 (open) — symlinks to iCloud Drive inside an already-added project fail on
  file reads and diffs. Different surface, plausibly the same lstat-vs-stat
  assumption elsewhere in the workspace layer.

## Fix applied or workaround

Nothing was changed on the machine. Workaround: type the resolved real path
(`/work/projects/...`) in the picker instead of the `~/Projects/...` symlink
path, or add the project on the server machine directly.

## Filed by

claude (Opus 5) via `t3 triage`.

iamshadmantaqi:
## What Changed

`WorkspaceEntries.browse` now treats a symlink to a directory as a folder. Add Project and path autocomplete both use this listing, so a `~/Projects` link to another drive shows up.

File symlinks and dangling links stay omitted. That matches the old file-exclusion rule.

## Why

`readdir(..., { withFileTypes: true })` fills each `Dirent` from lstat. For a symlink, `isDirectory()` is false even when the target is a folder, so the browser dropped those names with no error.

The report in #8408 is a home directory whose `Projects` entry is a symlink to `/work/projects`. The picker showed neither, so the only way in was typing the resolved path.

Following the link with `stat` is the smallest change that restores directory semantics. Treating every symlink as a folder would also list files, which this RPC is not for.

Closes #8408

## Blast Radius

Server-side `filesystem.browse` only. Web, desktop, and mobile all consume that RPC for Add Project and path autocomplete.

The extra `stat` runs only for name-matched symlink dirents, not for every entry in the directory.

A broken or file symlink still does not appear.

## Verification

Failing-then-passing regression in `apps/server/src/workspace/WorkspaceEntries.test.ts`.

- Before the fix, `vp test run apps/server/src/workspace/WorkspaceEntries.test.ts` failed with `expected [ 'src' ] to deeply equal [ 'Projects', 'src' ]`.
- After the fix, that file is 31/31 passing.
- `vp lint` on the two changed files: 0 errors.
- `vp run --filter t3 typecheck`: no errors in the changed file.

No UI chrome changed, so no screenshots.

## Checklist

- [x] No UI change (screenshots not applicable)
- [x] No motion change (video not applicable)

---

> [!NOTE]
> **Low Risk**
> Scoped to server browse listing with extra stat only on name-matched symlinks; behavior for normal directories and non-symlink entries is unchanged.
> 
> **Overview**
> **`WorkspaceEntries.browse`** (backing **`filesystem.browse`**) now lists **directory symlinks** in Add Project and path autocomplete, fixing cases like a `~/Projects` link to another drive that never appeared because `readdir` + `withFileTypes` only sees the link, not the target folder.
> 
> A new **`isBrowsableDirectory`** check keeps real directories as before and, for symlinks only, follows with **`stat`** to include folder targets. **File symlinks** and **broken links** still stay out of results; prefix and hidden-dot filtering are unchanged.
> 
> A regression test covers directory vs file vs dangling symlinks and prefix matching on a symlink name.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 542f312a832e3ea598503eb2b05fbfa169272756. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Include directory symlinks in `WorkspaceEntries.browse` results
> Adds an `isBrowsableDirectory` helper that returns true for real directories and for symlinks whose `stat` target is a directory. Dangling symlinks and file symlinks resolve to false via `stat` failure. The browse handler in [WorkspaceEntries.ts](https://github.com/pingdotgg/t3code/pull/8539/files#diff-26feb5e4175dd3c4b97bf023c259c336abef121574cd5665d06d3e4dbf1b15b8) now uses this helper instead of a plain `isDirectory` check, so directory symlinks appear in listings while still respecting prefix and hidden-name filtering.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 542f312.</sup>
> 
>

### DIFF
--- a/apps/server/src/workspace/WorkspaceEntries.test.ts
+++ b/apps/server/src/workspace/WorkspaceEntries.test.ts
@@ -658,6 +658,34 @@ it.layer(TestLayer, { excludeTestServices: true })("WorkspaceEntries", (it) => {
       }),
     );
 
+    it.effect("includes directory symlinks and excludes file and dangling ones", () =>
+      Effect.gen(function* () {
+        const workspaceEntries = yield* WorkspaceEntries.WorkspaceEntries;
+        const fileSystem = yield* FileSystem.FileSystem;
+        const path = yield* Path.Path;
+        const cwd = yield* makeTempDir({ prefix: "t3code-workspace-browse-symlink-" });
+        const target = yield* makeTempDir({ prefix: "t3code-workspace-browse-symlink-target-" });
+        yield* writeTextFile(cwd, "src/index.ts", "export {};\n");
+        yield* writeTextFile(cwd, "notes.txt", "ignore me");
+        yield* fileSystem.symlink(target, path.join(cwd, "Projects"));
+        yield* fileSystem.symlink(path.join(cwd, "notes.txt"), path.join(cwd, "notes-link"));
+        yield* fileSystem.symlink(path.join(cwd, "missing"), path.join(cwd, "broken"));
+
+        const listed = yield* workspaceEntries.browse({
+          partialPath: yield* appendSeparator(cwd),
+        });
+        const prefixed = yield* workspaceEntries.browse({
+          partialPath: path.join(cwd, "Pro"),
+        });
+
+        expect(listed.entries.map((entry) => entry.name)).toEqual(["Projects", "src"]);
+        expect(prefixed).toEqual({
+          parentPath: cwd,
+          entries: [{ name: "Projects", fullPath: path.join(cwd, "Projects") }],
+        });
+      }),
+    );
+
     it.effect("shows dot directories in directory mode and hidden-prefix mode", () =>
       Effect.gen(function* () {
         const workspaceEntries = yield* WorkspaceEntries.WorkspaceEntries;
--- a/apps/server/src/workspace/WorkspaceEntries.ts
+++ b/apps/server/src/workspace/WorkspaceEntries.ts
@@ -1,4 +1,5 @@
 // @effect-diagnostics nodeBuiltinImport:off
+import * as NodeFS from "node:fs";
 import * as NodeFSP from "node:fs/promises";
 import * as NodeOS from "node:os";
 
@@ -138,6 +139,17 @@ const resolveBrowseTarget = Effect.fn("WorkspaceEntries.resolveBrowseTarget")(fu
   return path.resolve(expandHomePath(input.cwd, path), input.partialPath);
 });
 
+const isBrowsableDirectory = (dirent: NodeFS.Dirent, parentPath: string, path: Path.Path) => {
+  if (dirent.isDirectory()) return Effect.succeed(true);
+  if (!dirent.isSymbolicLink()) return Effect.succeed(false);
+  return Effect.promise(() =>
+    NodeFSP.stat(path.join(parentPath, dirent.name)).then(
+      (stats) => stats.isDirectory(),
+      () => false,
+    ),
+  );
+};
+
 export const make = Effect.gen(function* () {
   const path = yield* Path.Path;
   const workspacePaths = yield* WorkspacePaths.WorkspacePaths;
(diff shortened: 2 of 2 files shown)


juliusmarminge:
Closing as a duplicate of #8281, which covers the same directory-symlink browse fix while keeping file and broken symlinks out of the picker. Please continue review there.

PR #8540, opened 2026-08-28
fix(server): preserve usage cache outside walked roots

Lucenx9:
## Problem

`pruneScanCache` used a raw string prefix to decide whether a cached transcript belonged to a walked root. Scanning `/claude/projects` could therefore evict an entry such as `/claude/projects-copy/a.jsonl`, even though that file was never part of the scan. The next usage read then had to parse that transcript again.

## Fix

Use `node:path.relative` to distinguish the root and its real descendants from paths that only share the same prefix. Add a regression test that keeps an in-window cache entry under the prefix-matching sibling directory.

## Verification

- `./node_modules/.bin/vp test run apps/server/src/usage/usageScanCache.test.ts`
- `./node_modules/.bin/vp run --filter t3 typecheck`
- `./node_modules/.bin/vp lint --report-unused-disable-directives apps/server/src/usage/usageScanCache.ts apps/server/src/usage/usageScanCache.test.ts`
- `./node_modules/.bin/vp fmt --check apps/server/src/usage/usageScanCache.ts apps/server/src/usage/usageScanCache.test.ts`

---

> [!NOTE]
> **Low Risk**
> Localized pruning logic fix with a regression test; behavior change only affects edge-case path matching, not auth or data writes.
> 
> **Overview**
> **`pruneScanCache`** no longer treats cached file paths as “under a walked root” when they only share a string prefix. It uses **`node:path.relative`** so paths must be the root itself or a true descendant (e.g. `/claude/projects/...`), not siblings like `/claude/projects-copy/...`.
> 
> That fixes incorrect eviction of in-window cache entries that were never part of the scan but were previously classified as under `/claude/projects`. A regression test asserts those prefix-sibling paths are kept when their root was not walked.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a6b13b62cdb4968acb5e8b397a81a21597e02d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `pruneScanCache` to use path-aware root matching
> The previous logic used a plain string-prefix check, so a cached path like `/claude/projects-copy/...` was wrongly treated as under the walked root `/claude/projects` and deleted. The fix uses `NodePath.relative` to compute the true relative path and only considers a path "under" a root if the result is empty or does not escape via `..`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0a6b13b.</sup>
> 
>

Implemented with gpt-daybreak-blue-latest via the Codex harness in T3 Code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved usage cache cleanup to distinguish genuine subdirectories from similarly named sibling paths.
  * Prevented valid cache entries from being incorrectly removed during pruning.

### DIFF
--- a/apps/server/src/usage/usageScanCache.test.ts
+++ b/apps/server/src/usage/usageScanCache.test.ts
@@ -193,6 +193,20 @@ describe("pruneScanCache with an unwalked root", () => {
     expect(removed).toBe(0);
     expect(cache.size).toBe(1);
   });
+
+  it("keeps entries under a sibling path that only shares the walked root prefix", () => {
+    const cache = cacheWith([["/claude/projects-copy/a.jsonl", 5000, [record()]]]);
+
+    const removed = pruneScanCache(cache, {
+      livePaths: new Set(),
+      walkedRoots: ["/claude/projects"],
+      windowStartMs: 4000,
+      retentionCutoffMs: 1000,
+    });
+
+    expect(removed).toBe(0);
+    expect(cache.size).toBe(1);
+  });
 });
 
 describe("dedupeWithinFile", () => {
--- a/apps/server/src/usage/usageScanCache.ts
+++ b/apps/server/src/usage/usageScanCache.ts
@@ -14,6 +14,9 @@
  *
  * @module usageScanCache
  */
+// @effect-diagnostics nodeBuiltinImport:off
+import * as NodePath from "node:path";
+
 import type { UsageProviderKind } from "@t3tools/contracts";
 
 import type { UsageRecord } from "./usageTranscripts.ts";
@@ -229,7 +232,15 @@ export function pruneScanCache(cache: ScanCache, options: PruneOptions): number 
   let removed = 0;
   for (const [path, entry] of cache) {
     const agedOut = entry.mtimeMs < options.retentionCutoffMs;
-    const underWalkedRoot = options.walkedRoots.some((root) => path.startsWith(root));
+    const underWalkedRoot = options.walkedRoots.some((root) => {
+      const relative = NodePath.relative(root, path);
+      return (
+        relative === "" ||
+        (relative !== ".." &&
+          !relative.startsWith(`..${NodePath.sep}`) &&
+          !NodePath.isAbsolute(relative))
+      );
+    });
     const deleted =
       underWalkedRoot && entry.mtimeMs >= options.windowStartMs && !options.livePaths.has(path);
     if (agedOut || deleted) {


Lucenx9:
@coderabbitai review

Lucenx9 (commented):
## Review

**Verdict: no blocking findings.** The change is correct, minimal, and well-tested.

### What I verified

**The bug is real.** With the old logic, `"/claude/projects-copy/a.jsonl".startsWith("/claude/projects")` is `true`. An in-window entry under that prefix-sharing sibling was therefore classified as part of the walked root and evicted when absent from `livePaths`. The result was an unnecessary cold parse when that transcript root was scanned again, not incorrect usage totals.

**The containment fix handles the relevant edge cases.** I probed the predicate directly:

| root | candidate | old classification | new classification |
|---|---|---|---|
| `/claude/projects` | `/claude/projects-copy/a.jsonl` | under root, incorrect | outside root, correct |
| `/claude/projects` | `/claude/projects/a.jsonl` | under root | under root |
| `/claude/projects` | `/claude/projectsTwin/a.jsonl` | under root, incorrect | outside root, correct |
| `/claude/projects` | `/claude/projects` | under root | under root |
| `/claude/projects` | `/claude/projects/..name/a.jsonl` | under root | under root |
| different Windows drives | candidate on another drive | outside root | outside root via `isAbsolute` |

These are containment results. An in-window cache entry is evicted only when it is under a walked root and absent from `livePaths`.

The guards `relative !== ".."` and ``!relative.startsWith(`..${NodePath.sep}`)`` reject parent escapes without rejecting legitimate descendants whose names begin with `..`.

**Both sides use consistent native paths.** `walkedRoots` come from `resolveTranscriptDirs`, using the platform `Path` service to join or resolve provider homes. Cache keys are the paths that `listTranscriptFiles` builds with `NodePath.join` from those roots. Both therefore follow the host platform conventions.

**The retention behavior is bounded.** An entry outside every walked root remains cached until its root is walked again or `retentionCutoffMs` removes it. The current read does not aggregate that entry unless the file walk returns it. Keeping it uses some cache space, but it avoids the false eviction fixed here.

**The regression test is valid.** `keeps entries under a sibling path that only shares the walked root prefix` fails with the old `startsWith` logic and passes with this change.

**Green on my run.**

- `vp test run apps/server/src/usage/usageScanCache.test.ts`: 14 of 14 tests passed
- `vp run --filter t3 typecheck`: exit code 0; remaining diagnostics are suggestions in unrelated files
- `// @effect-diagnostics nodeBuiltinImport:off` matches the established server convention

### Non-blocking notes

1. `relative === ""` covers the root itself. Cache keys are transcript files, so this branch is unreachable with current callers. Keeping it makes the containment predicate complete.
2. The repository already has other private path-containment checks, including the same predicate shape in `GitVcsDriverCore`. A shared helper is not warranted by this focused fix because those checks live in different modules and use different `Path` implementations.

Reviewed with GLM (`glm-5.3-flash`) via OpenCode.

PR #8542, opened 2026-08-28
feat(desktop): add open at login setting

TimPeterNL:
## Summary
- The installed desktop app did not start when Windows (or macOS/Linux) signed in, so T3 Code was missing after a reboot while Codex launched automatically.
- Adds an **Open at login** toggle in Settings → General. It is off by default. Turning it on registers the packaged app as an OS login item; turning it off removes it.
- Unpackaged/dev builds still persist the preference, but only the installed app writes the login item.

## Test plan
- [ ] In the installed desktop app, open Settings → General and turn on **Open at login**.
- [ ] Confirm T3 Code appears under Windows Settings → Apps → Startup (or macOS System Settings → General → Login Items).
- [ ] Sign out / restart and confirm the desktop app launches.
- [ ] Turn the setting off and confirm it disappears from the OS startup list.
- [ ] Confirm the toggle is hidden in the browser/web client and searchable only in the desktop app.

Cursor Grok 4.6

Made with [Cursor](https://cursor.com)

> [!NOTE]
> ### Add 'Open at login' setting to desktop app
> - Adds `openAtLogin` to `DesktopSettings` (default `false`) and persists it to the settings document when non-default.
> - Implements `DesktopLoginItem` which reads state and applies the preference via `ElectronApp.setLoginItemSettings` only on packaged macOS/Windows builds; no-op on Linux and unpackaged builds.
> - Adds IPC channels `GET_OPEN_AT_LOGIN_STATE_CHANNEL` and `SET_OPEN_AT_LOGIN_CHANNEL` with handlers, exposed to the renderer through the preload bridge.
> - Adds `useDesktopOpenAtLogin` hook and `OpenAtLoginSettingsRow` in the General settings panel with optimistic toggles, error toasts, and a reset button.
> - Applies the persisted preference on startup in `DesktopApp.ts`, logging a warning via `ElectronLoginItemSettingsError` if OS registration fails.
> - Risk: `ElectronApp.setLoginItemSettings` wraps Electron's API in `Effect.try`; any thrown error from `Electron.app.setLoginItemSettings` surfaces as `ElectronLoginItemSettingsError` and is caught at startup, but `DesktopLoginItem.setOpenAtLogin` persists the preference before applying, so a failed OS registration leaves `openAtLogin: true` on disk while the OS login item is not registered.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d5ca0e4.</sup>
> 
>


TimPeterNL:
Stupid Grok created a PR when I was just debugging. PR closed.

PR #6203, opened 2026-08-11
feat(desktop): add local file mentions

coah80:
## What changed

- render dropped or pasted non-image desktop files as compact file mentions
- keep the full absolute path in the composer payload while showing only the filename
- preserve image attachments on web, desktop, and mobile
- limit local-path mentions to the desktop app's primary environment
- warn when non-image files are dropped into a remote desktop environment
- block native paste when a local file path cannot be resolved

## Why

Plain absolute paths are noisy in the composer, but uploading arbitrary files would introduce remote-transfer and file-size concerns. Desktop-local file mentions keep the UI readable while giving the agent the exact path it needs. Remote environments reject these files because a path from the client computer would not exist on the host computer.

### JAR

JAR file mention

### PDF

PDF file mention

## Validation

- `vp test run apps/web/src/components/chat/composerFileUpload.test.ts apps/web/src/pierre-icons.test.ts`
- targeted `vp lint` for changed files
- `pnpm --filter @t3tools/web typecheck`
- `pnpm --filter @t3tools/desktop typecheck`
- `pnpm --filter @t3tools/contracts typecheck`
- `git diff --check origin/main...HEAD`

Built with GPT-5.6-Sol using the Codex harness in T3 Code.

> [!NOTE]
> ### Add local file path mentions to desktop chat composer via drag and paste
> - Adds `getPathForFile` to `DesktopBridge` (via Electron's `webUtils`) and exposes it to the renderer in [preload.ts](https://github.com/pingdotgg/t3code/pull/6203/files#diff-eae0078d2bf4ac4beed2884f918b2816a97752917390f09980ba2cfad8c42cac).
> - Introduces [composerFileUpload.ts](https://github.com/pingdotgg/t3code/pull/6203/files#diff-336e00aad644d93291461570a2fc5bc1e49f557c355151b524a5aedec06c54cf) to partition dropped/pasted files into images (attached) and non-images (inserted as local file path mentions when on desktop).
> - Paste and drag-and-drop in [ChatComposer.tsx](https://github.com/pingdotgg/t3code/pull/6203/files#diff-c968a393420901e312c24fcfdef204d8969fa91d9c2df9c52d32b6612bcc8d9a) now route all files through this utility; non-image files on desktop become inline path mentions, while unsupported or unresolved files trigger error/warning toasts.
[...]

PR #8549, opened 2026-08-28
feat(composer): drop non-image files as filesystem paths

nioasoft:
Reopening #7749, which was closed with this reason:

> This inserts local desktop paths for dropped files, but a selected remote environment often cannot read those paths. The implementation breaks the remote-ready contract even though its quoting tests are useful.

That was correct, and this fixes it. I could not reopen #7749 through the API (`reopenPullRequest` returns "Could not open the pull request", `PATCH state=open` returns 422), hence a fresh PR rather than ignoring the request to explain what was missed.

## What Changed

Dropping a non-image file (audio, PDF, video, …) on the composer inserts its filesystem path as text, so the agent opens the file itself instead of a 40MB attachment crossing the wire — the same gesture as dropping a file into a terminal. Images are unaffected and still attach.

The insert is gated on the thread's environment being the one this desktop app supervises:

```ts
export function droppedPathsResolveInEnvironment(input: {
  readonly primarySource: string | undefined;
  readonly primaryEnvironmentId: string | undefined;
  readonly threadEnvironmentId: string | undefined;
}): boolean {
  if (input.primarySource !== "desktop-managed") return false;
  return (
    input.primaryEnvironmentId !== undefined &&
    input.threadEnvironmentId !== undefined &&
    input.primaryEnvironmentId === input.threadEnvironmentId
  );
}
```

When it refuses, the composer says why rather than dropping the file on the floor:

> This thread runs on another machine, which cannot open a path from this one. Attach the file, or type a path that exists there.

## Why

The earlier version resolved a path through the local desktop bridge and inserted it regardless of where the thread ran. On a remote environment — a LAN or Tailscale host, a relay, a tunnel, another desktop acting as server — that path names a file on the wrong machine, so the agent either cannot find it or, worse, finds a different one. `packages/contracts/src/ipc.ts` already states the rule that broke:

> callers should resolve it by `environmentId` rather than reaching through the local desktop bridge

`desktop-managed` is the single environment this app supervises on this machine; every other `KnownEnvironmentSource` (`configured`, `manual`, `window-origin`) names a server somewhere else. A missing id on either side also refuses — bootstrapping is not evidence the thread runs here, and guessing wrong reproduces exactly the silent failure that got #7749 closed.

Two smaller decisions carried over from the earlier review:

- Failures report through a toast, not `setThreadError`. A mixed drop's images own that banner, and a path failure there reads as though the whole drop failed.
- A resolved path is never trimmed. Leading and trailing spaces are legal in a filename, so trimming one points the agent at a file that does not exist; quoting keeps it readable instead.

## UI Changes

No visual change. The only new surface is an error toast on a drop that cannot be honored, where the previous behavior was silence.

## Test plan

- `apps/web/src/components/chat/droppedFilePaths.test.ts` — 15 tests: quoting (whitespace, embedded quotes, paths that must not be trimmed), formatting, and the environment predicate (matching ids, mismatched ids, each non-desktop source, either id unknown).
- `pnpm --filter @t3tools/web typecheck` clean, targeted lint clean.
- Manually: dropped a PDF and an audio file on a local thread — paths inserted; images in the same drop still attached.

Rebased onto current `main`.

## Checklist

- [x] I included before/after screenshots for any UI changes — n/a, no visual change
- [x] I included a video for animation/interaction changes — n/a

---

---

> [!NOTE]
> **Medium Risk**
> Changes drop handling and prompt content in a high-traffic composer path; environment gating limits wrong-machine paths but incorrect ID matching could still block or allow inserts incorrectly.
> 
> **Overview**
> **Dropping non-image files on the chat composer** (PDFs, audio, archives, etc.) now inserts their **local filesystem path** into the prompt instead of rejecting them or uploading large blobs—similar to dropping a file into a terminal. **Images and video still attach** as before; mixed drops handle both paths.
> 
> The **desktop preload** exposes `desktopBridge.getPathForFile` via Electron `webUtils` (replacing removed `File.path`), with the contract updated accordingly.
> 
> Path insertion runs **only when the thread’s `environmentId` matches the desktop-managed primary environment** on this machine (`droppedPathsResolveInEnvironment`). Browser tabs, remote/SSH/relay connections, mismatched environments, or unknown IDs get **explicit error toasts** instead of silent wrong-path inserts. Paths are **quoted** when they contain whitespace (without trimming legal trailing spaces).
> 
> **Composer focus** is skipped immediately after a successful path insert so Lexical reconciliation does not wipe the new text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f52c963546fa1e562247a02441a1ba37bbefba41. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Insert non-image dropped files as filesystem paths in `ChatComposer`
> - Adds `desktopBridge.getPathForFile(file)` to the preload bridge, using Electron's `webUtils` to resolve a filesystem path for OS-originated `File` objects (returns `null` otherwise)
> - Splits dropped files in `ChatComposer.addDroppedFiles`: images/videos attach as before; all other types are resolved to local paths and inserted as quoted text via the new `insertDroppedFilePaths` helper
> - Adds pure utilities in [droppedFilePaths.ts](https://github.com/pingdotgg/t3code/pull/8549/files#diff-3b1c46ba1728118957ed7272aefe0d2fcab6f06d955da9316f642c9a94ba1147): `droppedPathsResolveInEnvironment` (requires `primarySource === 'desktop-managed'` and matching environment ids), `quoteDroppedFilePath` (wraps paths with whitespace in double quotes), and `formatDroppedFilePaths` (filters empties and joins with spaces)
> - Surfaces failures via `toastManager` when the desktop bridge is absent, the thread environment does not match the local desktop-managed environment, no paths resolve, or the composer is busy
> - Behavioral Change: dropping a non-image file in a browser or in a thread whose environment differs from the local desktop-managed environment now shows an error toast instead of attaching or ignoring the file; focus after a drop is deferred when path text was inserted to avoid clobbering it
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f52c963.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/preload.ts
+++ b/apps/desktop/src/preload.ts
@@ -5,7 +5,7 @@ import type {
   DesktopPreviewTabState,
 } from "@t3tools/contracts";
 import { exposeClerkBridge } from "@clerk/electron/preload";
-import { contextBridge, ipcRenderer } from "electron";
+import { contextBridge, ipcRenderer, webUtils } from "electron";
 
 import * as IpcChannels from "./ipc/channels.ts";
 
@@ -39,6 +39,15 @@ contextBridge.exposeInMainWorld("desktopBridge", {
     return result as ReturnType<DesktopBridge["getAppBranding"]>;
   },
   getClientPlatform: () => clientPlatform,
+  getPathForFile: (file: File) => {
+    // Throws for a File that never came from the OS (e.g. built by the page).
+    try {
+      const path = webUtils.getPathForFile(file);
+      return path.length > 0 ? path : null;
+    } catch {
+      return null;
+    }
+  },
   getSystemLocale: () => {
     const result = ipcRenderer.sendSync(IpcChannels.GET_SYSTEM_LOCALE_CHANNEL);
     return typeof result === "string" ? result : null;
--- a/apps/web/src/components/chat/ChatComposer.tsx
+++ b/apps/web/src/components/chat/ChatComposer.tsx
@@ -169,7 +169,9 @@ import {
   submitComposerDraft,
 } from "./composerSubmission";
 import { ComposerPromptLengthValidation } from "./ComposerPromptLengthValidation";
-import { formatDroppedFilePaths } from "./droppedFilePaths";
+import { getPrimaryKnownEnvironment } from "~/environments/primary";
+
+import { droppedPathsResolveInEnvironment, formatDroppedFilePaths } from "./droppedFilePaths";
 
 function ComposerVideoThumbnail({ file }: { file: File }) {
   const setVideo = useCallback(
@@ -3165,6 +3167,23 @@ export const ChatComposer = memo(function ChatComposer(props: ChatComposerProps)
     if (!resolvePath) {
       reportFailure(
         "Attaching files by path needs the desktop app. Paste an image, or type the path.",
+      );
+      return false;
+    }
+    // The bridge reads a path on THIS machine. Inserting it into a thread whose
+    // agent runs elsewhere hands over a path that host cannot open — so gate on
+    // the thread's environment being the one this desktop app supervises,
+    // rather than on the bridge merely existing.
+    const primaryEnvironment = getPrimaryKnownEnvironment();
+    if (
+      !droppedPathsResolveInEnvironment({
+        primarySource: primaryEnvironment?.source,
+        primaryEnvironmentId: primaryEnvironment?.environmentId,
+        threadEnvironmentId: environmentId,
+      })
+    ) {
+      reportFailure(
+        "This thread runs on another machine, which cannot open a path from this one. Attach the file, or type a path that exists there.",
       );
       return false;
     }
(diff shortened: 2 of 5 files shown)


t3dotgg:
> [!NOTE]
>

Carryover from [#6203](https://github.com/pingdotgg/t3code/pull/6203) at [6df5f1c937](https://github.com/pingdotgg/t3code/commit/6df5f1c9375e4899be80c2c1b8c24aaa29bd7662): the optional compact file-mention rendering, paste integration, and Unix/Windows basename cases remain useful input for this proposal. Use the current attachment API and preserve filename whitespace. Do not copy the old helper's rejection of remote non-image uploads or assume that a primary environment uses the Windows filesystem when it is WSL. These ideas are recorded for review, not approved or ported.

shivamhwp:

Thanks for updating this to address remote environments. We're closing it because #8236 already added support for attaching PDFs, ZIPs and other files, including through drag and drop.

Inserting local paths would avoid uploads on desktop, but we'd prefer to keep the existing attachment behavior consistent across local and remote environments rather than make the same drop behave differently depending on where the thread runs.

PR #8580, opened 2026-08-28
fix(server): make compatible Codex instance switches single-writer

### LINKED ISSUE #8259: [Bug]: Switching Codex provider instances can leave a thread stuck on "already has an active writer"
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Configure two enabled Codex provider instances in T3 Code that can access the same Codex thread/session store. Use different model selections so T3 treats them as different instances, for example `codex` and `codex_a2`.
2. Start a thread on instance A and send a message.
3. Interrupt the turn, but leave T3 Code running.
4. In the same T3 thread, switch the selected model/provider instance to B and send another message.
5. Retry the message after the first failure.

The captured repro used T3 thread `e76d...` and Codex thread `01a03...`. Identifiers are abbreviated here.

### Expected behavior

T3 should release the old provider session before starting the replacement, or perform the replacement atomically. The Codex app-server should have at most one writer for the thread, and the new turn should start once without making the thread unusable.

### Actual behavior

The interrupted turn is persisted as `interrupted`, but the old Codex app-server runtime remains active. When the provider instance is changed, T3 logs that it is restarting the provider session and passes the previous resume cursor to the new instance.

The new Codex app-server then rejects the request because the same Codex thread already has a writer:

```text
ProviderAdapterProcessError: Provider adapter process error (codex) for thread <t3-thread-id>: thread <codex-thread-id> already has an active writer
  [cause]: CodexAppServerRequestError: thread <codex-thread-id> already has an active writer
```

In the captured run:

- the first turn ended at 06:28:52Z with `turn/completed: interrupted`;
- at 06:29:16Z T3 changed the selection from `codex` to `codex_a2`;
- the restart log reported `currentInstanceId=codex`, `desiredInstanceId=codex_a2`, and `hasResumeCursor=true`;
- the provider start failed before a new turn began;
- a retry at 06:29:52Z failed with the same error.

The current implementation appears to start the target adapter before stopping stale sessions:

- [ProviderService.startSession](https://github.com/pingdotgg/t3code/blob/main/apps/server/src/provider/Layers/ProviderService.ts#L654-L678) calls the adapter's `startSession` before `stopStaleSessionsForThread`.
- [ProviderCommandReactor.ensureSessionForThread](https://github.com/pingdotgg/t3code/blob/main/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts#L730-L752) retains the active session's resume cursor during the instance switch.
- [CodexAdapter](https://github.com/pingdotgg/t3code/blob/main/apps/server/src/provider/Layers/CodexAdapter.ts#L1657-L1695) keeps session state per adapter instance, so the target adapter cannot stop a session held by the previous adapter.

### Impact

Blocks work completely

### Version or commit

T3 Code server package 0.0.34

### Environment

- Linux 6.1.0-40-amd64, x86_64
- Node.js v24.11.1
- Codex CLI 0.149.1
- T3 web/server with two Codex provider instances

### Logs or stack traces

```text
2026-08-26T06:29:16Z
provider command reactor restarting provider session
currentInstanceId=codex
desiredInstanceId=codex_a2
hasResumeCursor=true

2026-08-26T06:29:16Z
CodexAppServerRequestError: thread <codex-thread-id> already has an active writer

2026-08-26T06:29:52Z
CodexAppServerRequestError: thread <codex-thread-id> already has an active writer
```

No raw logs, prompts, paths, or database files are attached because they may contain private source code or other sensitive data.

### Workaround

Avoid switching Codex provider instances inside an existing T3 thread. Start a new thread on the target instance instead. Restarting T3 before retrying may release the stale runtime, but it also interrupts other active provider sessions.

### Suggested fix

- Enforce a single-writer handoff when changing provider instances: stop or detach the old session before starting the new one, or use an atomic replacement protocol.
- If replacement fails, clean up the attempted target session and leave the thread in an actionable recoverable state.
- Ensure stale lifecycle events from the replaced instance cannot overwrite the new binding.
- Add a regression test that interrupts a Codex turn, switches from instance A to instance B, and verifies that the next turn starts exactly once without an active-writer error.

### Related issues

- #7075 covers large Codex resume responses that can exhaust the T3 server heap. This report concerns the provider-session writer conflict that occurs during instance replacement.
- #4944 covers stale provider shutdown during a Claude instance switch. The lifecycle problem is similar, but this report reproduces the Codex active-writer error.

### Disclosure

I used AI assistance to organize and cross-check the local T3 logs and source. I reviewed the final report, and identifiers and log paths were redacted.

vitaliyhayda:
Fixes #8259

## TLDR

Before this change, the picker could show separately authenticated provider instances, but an existing thread remained bound to its current instance. Switching did not safely transfer that conversation. It could start the target before fully releasing the source or silently create a fresh native thread instead of preserving the same ID.

Now a manual switch between compatible Codex instances stops the old writer, strictly resumes the same native conversation, verifies its identity, and only then commits the new owner. Separate histories remain supported and are still the default for instances with different continuation identities. Only deliberately compatible instances can hand off one history.

Current switching starts the target before stopping the source. Source-stop failures are suppressed, so two app-server instances can compete for one native thread

A simple reorder is insufficient:

1. Cross-instance start can lose the persisted source cursor
2. thread/resume can fall back to thread/start
3. That fallback silently creates a new native thread
4. Adapter shutdown removes session state before cleanup completes
5. Runtime-close failures are currently suppressed
6. Delayed source events can overwrite target-owned state

This change:

1. Serializes recovery-capable operations with cleanup-safe per-thread locks
2. Requires compatible continuation identities
3. Requires an idle source and valid persisted cursor
4. Persists the cursor before teardown
5. Stops the source before target startup
6. Proves child-process release before target resume
7. Propagates stop and release failures
8. Uses strict thread/resume whenever a cursor exists
9. Never falls back to thread/start during handoff
10. Requires the returned native ID to match
11. Stops provisional targets after identity or persistence failure
12. Commits target ownership only after readiness and persistence
13. Ignores delayed events from stale owners
14. Keeps turn interruption out of band from a blocked send while ordering it behind handoff
15. Proves non-owner Codex writers are released before any native resume
16. Delivers approval and user-input responses outside the long-running turn lock without recovery
17. Restricts native Codex handoff rules to Codex-to-Codex switches

Failure behavior:

1. Cursor persistence failure leaves the source running
2. Release failure prevents target startup
3. Target failure leaves the source stopped with a recoverable cursor
4. Identity mismatch stops the target
5. Binding failure stops the target
6. Unproven cleanup remains visible and blocks retry
7. Failed target cleanup blocks source recovery and direct restart

Verification:

1. 197 focused tests pass
2. Server typecheck exits 0
3. git diff --check passes
4. Tests cover release, cursor, identity, compensation, stale events, lock cleanup, interruption, interactive responses, and cross-provider replacement

Scope:

1. Existing manual switching only
2. No UI or schema changes
3. No automatic routing
4. No quota logic
5. No native-history import
6. No Claude changes
7. No token handling

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
@@ -301,6 +301,7 @@ describe("ProviderRuntimeIngestion", () => {
         threadId: ThreadId.make("thread-1"),
         status: "ready",
         providerName: "codex",
+        providerInstanceId: ProviderInstanceId.make("codex"),
         runtimeMode: "approval-required",
         activeTurnId: null,
         updatedAt: createdAt,
@@ -310,6 +311,7 @@ describe("ProviderRuntimeIngestion", () => {
     });
     provider.setSession({
       provider: ProviderDriverKind.make("codex"),
+      providerInstanceId: ProviderInstanceId.make("codex"),
       status: "ready",
       runtimeMode: "approval-required",
       threadId: ThreadId.make("thread-1"),
@@ -678,6 +680,50 @@ describe("ProviderRuntimeIngestion", () => {
       );
       expect(thread?.session?.status).toBe("stopped");
       expect(thread?.session?.activeTurnId).toBeNull();
+    }),
+  );
+
+  effectIt.effect("ignores a delayed source exit after target ownership commits", () =>
+    Effect.gen(function* () {
+      const harness = yield* Effect.promise(() => createHarness());
+      const threadId = asThreadId("thread-1");
+      const sourceInstanceId = ProviderInstanceId.make("codex-source");
+      const targetInstanceId = ProviderInstanceId.make("codex-target");
+      const committedAt = "2026-01-01T00:00:02.000Z";
+
+      yield* harness.engine.dispatch({
+        type: "thread.session.set",
+        commandId: CommandId.make("cmd-target-session-committed"),
+        threadId,
+        session: {
+          threadId,
+          status: "ready",
+          providerName: "codex",
+          providerInstanceId: targetInstanceId,
+          runtimeMode: "approval-required",
+          activeTurnId: null,
+          lastError: null,
+          updatedAt: committedAt,
+        },
+        createdAt: committedAt,
+      });
+
+      harness.emit({
+        type: "session.exited",
+        eventId: asEventId("evt-delayed-source-exit"),
+        provider: ProviderDriverKind.make("codex"),
+        providerInstanceId: sourceInstanceId,
+        threadId,
+        createdAt: "2026-01-01T00:00:03.000Z",
+      });
+
+      yield* Effect.promise(() => harness.drain());
+      const thread = (yield* Effect.promise(() => harness.readModel())).threads.find(
+        (entry) => entry.id === threadId,
+      );
+      expect(thread?.session?.status).toBe("ready");
+      expect(thread?.session?.providerInstanceId).toBe(targetInstanceId);
+      expect(thread?.session?.updatedAt).toBe(committedAt);
     }),
   );
 
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts
@@ -1497,6 +1497,20 @@ const make = Effect.gen(function* () {
     Effect.gen(function* () {
       const thread = yield* resolveThreadShell(event.threadId);
       if (!thread) return;
+      const projectedProviderInstanceId = thread.session?.providerInstanceId;
+      if (
+        projectedProviderInstanceId !== undefined &&
+        event.providerInstanceId !== undefined &&
+        projectedProviderInstanceId !== event.providerInstanceId
+      ) {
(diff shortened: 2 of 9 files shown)


ReneFr81:
Thanks for working on this. We hit the same writer-lock failure twice on T3 `0.0.39-nightly.20260903.1262` with Codex CLI `0.153.0`.

One extra detail from our cases: an interrupted turn was not required. One source turn had completed normally, and the other had ended after a usage-limit error. The old, idle app-server still held the writer, so switching from direct Codex to Codex Unified failed on `thread/resume`.

After stopping the idle source app-server, Unified resumed the same native threads with their complete 9-turn and 10-turn histories.

Would you mind adding regression coverage for switching after a normal completion and after a provider error such as a usage limit? The handoff in this PR looks like the right fix for both cases.

shivamhwp:

Thanks for the detailed work on safe Codex account switching, especially cleanup failures, conversation identity, and stale events.

We're closing this because Orchestration V2 in #2829 is replacing the implementation this PR changes. We don't plan to add more lifecycle logic to the retiring path.

If you'd like to continue this work, contributions against the V2 branch are welcome. Please check which failure cases remain there and target those gaps rather than porting the old implementation wholesale. The scenarios and tests you've developed here would be useful for that.

PR #8631, opened 2026-08-29
feat(web): expose chat messages as headings

### LINKED ISSUE #8535: [Feature]: Expose each chat message as a heading for screen reader navigation
### Before submitting

### Area

apps/web

### Problem or use case

As a screen reader user, moving around a conversation in T3 Code is slow. The transcript renders as a flat run of text with no semantic structure marking where one message ends and the next begins. To get back to an earlier message I have to arrow through every line of everything in between, including long code blocks. There is no way to skim, and no way to answer "where does the previous reply start?" without reading forward through the whole thing.

### Proposed solution

Render the author of each message as a real heading element. For every turn in the transcript, emit a heading whose text identifies who sent it:

- "You" for user messages
- "T3 Code" (or the model name) for assistant messages

Use a consistent level that nests under whatever heading identifies the conversation itself — if the chat title is an `h1`, each message becomes an `h2`.

If the visual design does not want a visible heading, it can be visually hidden with an `sr-only` utility while staying in the accessibility tree. Where a sender label is already rendered, marking that existing element up as a heading is enough — no new visible UI required.

### Why this matters

Heading navigation is the primary way screen reader users skim. NVDA and JAWS move heading to heading with `H`, and VoiceOver's rotor lists every heading on the page. With one heading per message, jumping to the previous message becomes a single keystroke and the whole conversation becomes listable. Without them, the transcript is an undifferentiated wall of text and the only navigation available is line-by-line arrowing.

Issue #1915 added an outline minimap that solves the same navigation problem for sighted users. This is the assistive-technology equivalent, and it is considerably cheaper to build.

### Smallest useful scope

One heading per message, announcing the author, at a consistent level, in the chat transcript. Visually hidden is fine.

### Alternatives considered

- `role="article"` with an `aria-label` per message. This helps, but article navigation is supported and used far less widely than heading navigation.
- A live region announcing new replies as they stream in. That is a real and separate need — it covers content arriving, not navigating back through content that is already there.

### Risks or tradeoffs

Heading levels need to stay consistent with the rest of the page so the document outline does not skip levels.

Worth noting: assistant replies frequently contain their own markdown headings. Those need to render at a level below the per-message heading, otherwise a model reply containing an `h2` will sit as a sibling of the message headings and break the outline.

### Examples or references

- WCAG 2.4.10 Section Headings
- WCAG 1.3.1 Info and Relationships

### Contribution

yashranaway:
Chat messages lack semantic boundaries for heading navigation. Add screen-reader-only h3 author headings for user and assistant messages, below the chat title's h2, and offset markdown heading levels beneath each message without changing visual styles.

The rebase preserves main's stable CHAT_MARKDOWN_COMPONENTS and carries headingLevelOffset through ChatMarkdownRendererContext. The original static-markup accessibility test has been removed because it does not establish screen-reader behavior.

Validation: all 90 existing markdown and timeline tests pass, web typecheck passes, and scoped lint reports existing warnings. Chromium accessibility-tree verification passes at 1440×1000 and 390×844 with 80 synthetic turns: author levels 3, nested markdown levels 4–6, chronological reading order after virtualizer settlement, four keyboard-triggered older-history loads to turn 1, and Home/End/Enter navigation back to turn 80. Only 12–15 messages were mounted at checkpoints. No browser console errors or error overlay. Native screen-reader and Electron-shell behavior remain untested. LegendList’s existing 500 ms DOM-order debounce means intermediate scrolling snapshots can be out of order; checks wait for observable order before inspecting the accessibility tree.

[Browser verification evidence](https://github.com/yashranaway/t3code/releases/tag/pr-8631-browser-evidence) includes the executed script, fixtures, accessibility trees, and result checkpoints.

| Before | After |
| --- | --- |
| Before | After |

Closes #8535.

Model: GPT-6
Harness: Codex in T3 Code

## Summary by CodeRabbit

- **Accessibility**
  - Improved heading hierarchy in chat messages to better fit the surrounding page structure.
  - Added screen-reader-only labels identifying the speaker for each timeline message.
  - Preserved visual styling while providing more accurate heading levels to assistive technologies.

### DIFF
--- a/apps/web/src/components/ChatMarkdown.tsx
+++ b/apps/web/src/components/ChatMarkdown.tsx
@@ -2623,27 +2623,57 @@ const ChatMarkdownRendererContext = React.createContext<
 const CHAT_MARKDOWN_COMPONENTS = {
   h1: function MarkdownHeading1({ node: _node, ...props }) {
     const { headingLevelOffset } = use(ChatMarkdownRendererContext);
-    return <h1 {...props} aria-level={headingLevelOffset ? Math.min(6, 1 + headingLevelOffset) : undefined} />;
+    return (
+      <h1
+        {...props}
+        aria-level={headingLevelOffset ? Math.min(6, 1 + headingLevelOffset) : undefined}
+      />
+    );
   },
   h2: function MarkdownHeading2({ node: _node, ...props }) {
     const { headingLevelOffset } = use(ChatMarkdownRendererContext);
-    return <h2 {...props} aria-level={headingLevelOffset ? Math.min(6, 2 + headingLevelOffset) : undefined} />;
+    return (
+      <h2
+        {...props}
+        aria-level={headingLevelOffset ? Math.min(6, 2 + headingLevelOffset) : undefined}
+      />
+    );
   },
   h3: function MarkdownHeading3({ node: _node, ...props }) {
     const { headingLevelOffset } = use(ChatMarkdownRendererContext);
-    return <h3 {...props} aria-level={headingLevelOffset ? Math.min(6, 3 + headingLevelOffset) : undefined} />;
+    return (
+      <h3
+        {...props}
+        aria-level={headingLevelOffset ? Math.min(6, 3 + headingLevelOffset) : undefined}
+      />
+    );
   },
   h4: function MarkdownHeading4({ node: _node, ...props }) {
     const { headingLevelOffset } = use(ChatMarkdownRendererContext);
-    return <h4 {...props} aria-level={headingLevelOffset ? Math.min(6, 4 + headingLevelOffset) : undefined} />;
+    return (
+      <h4
+        {...props}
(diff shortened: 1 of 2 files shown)


shivamhwp:

This needs a rebase after the markdown renderer refactor. Main now uses stable `CHAT_MARKDOWN_COMPONENTS` and `ChatMarkdownRendererContext`; please carry the heading offset through that context rather than restoring per-render component functions. The chat title is h2 on main, so h3 author headings fit the current outline.

The added test renders static markup and checks literal tags, which AGENTS.md explicitly rules out as proof of frontend behavior. Replace it with an accessibility-tree or real screen-reader check that reaches user and assistant headings, nests markdown headings below them, and verifies navigation through the virtualized timeline. No native screen-reader behavior is established by the current test.

juliusmarminge:
Superseded by #11199, which already landed and closed #8535.

juliusmarminge:
Superseded by #11199 (merged), which also closes #8535 (screen-reader headings on chat messages). Closing this PR as superseded.

PR #8633, opened 2026-08-29
fix(server): reject HTTP turn bootstrap clearly

### LINKED ISSUE #8319: HTTP dispatch accepts `bootstrap` on `thread.turn.start` but only the WS route implements it, yielding an untyped 500
## Summary

The HTTP dispatch endpoint (`POST /api/orchestration/dispatch`) accepts a `thread.turn.start` payload carrying the optional `bootstrap` field, but only the WebSocket route actually implements bootstrap. Over HTTP the command passes schema validation, then fails deep inside the engine and surfaces as an untyped 500 (`EnvironmentInternalError` / `orchestration_dispatch_failed`), where a typed schema rejection (or working bootstrap support) is expected.

## Context: what we were doing

We are building an external orchestrator client that drives T3 Code environments through the documented-in-contracts external surface: pairing via `t3 pair`, token exchange on `/oauth/token`, then `POST /api/orchestration/dispatch` with `ClientOrchestrationCommand` payloads. This works very well overall — project/thread creation, turn start, steering, interrupt, snapshot reads, and cleanup all behave exactly as the contracts describe. This report is the one asymmetry we hit.

## Steps to reproduce

Environment: server `0.0.33` (npm tag `latest` at the time), source read at commit `7107a98a225be85b58ddcd4de02c343af7d4707a`; local environment server on `http://127.0.0.1:3773`; macOS.

1. Create a pairing credential: `npx t3@0.0.33 pair --label repro --ttl 30m`.
2. Exchange it on `POST /oauth/token` with `grant_type=urn:ietf:params:oauth:grant-type:token-exchange`. The resulting Bearer token carries `orchestration:read orchestration:operate terminal:operate review:write relay:read`.
3. Without any prior `thread.create`, dispatch a single bootstrap turn start over HTTP:

```
POST /api/orchestration/dispatch
Authorization: Bearer <token>
Content-Type: application/json

{
  "type": "thread.turn.start",
  "commandId": "<uuid>",
  "threadId": "<fresh uuid>",
  "turnId": "<uuid>",
  "text": "hello",
  "bootstrap": { "createThread": { ...valid ThreadCreate fields... } },
  "createdAt": "<iso>"
}
```

## Observed

HTTP 500:

```json
{"_tag":"EnvironmentInternalError","reason":"orchestration_dispatch_failed","traceId":"83eba56cf0e02ae8b6dec1cf2d8c5c62"}
```

In `~/.t3/userdata/logs/server.trace.ndjson`, the same traceId shows the real cause: `OrchestrationCommandInvariantError` from `apps/server/src/orchestration/commandInvariants.ts` (line 111 at the commit above): `Thread '<id>' does not exist for command 'thread.turn.start'.`

The same command sequence works fine when split into explicit `thread.create` + plain `thread.turn.start` over HTTP, and the bootstrap form works over WebSocket.

## Why this happens (from reading the source at commit `7107a98a`)

- `ClientThreadTurnStartCommand` (`packages/contracts/src/orchestration.ts`, around line 846) carries the optional `bootstrap` field (`createThread`, `prepareWorktree`, `runSetupScript`), and the HTTP endpoint's payload schema is that same union, so the payload validates.
- The only consumer of `bootstrap` is `dispatchBootstrapTurnStart` in `apps/server/src/ws.ts` (roughly lines 757–985): it dispatches `thread.create`, optionally prepares a worktree, runs the setup program, and only then dispatches the turn start (with `cleanupCreatedThread` on failure).
- The HTTP handler in `apps/server/src/orchestration/http.ts` never routes through that path; both routes share only `apps/server/src/orchestration/Normalizer.ts`, which normalizes timestamps and ignores `bootstrap`. So the engine receives a turn start for a thread that was never created and fails an invariant.

## Expected

One of:

1. HTTP `dispatch` implements the same bootstrap path as the WS route (nicest for external clients — a one-call thread-plus-first-turn is very convenient over plain HTTP), or
2. HTTP `dispatch` rejects a payload carrying `bootstrap` with a typed 400 schema/validation error stating the field is WS-only, instead of accepting it and failing with an untyped 500.

Either way the contract and the behavior would agree again. Happy to provide more trace detail if useful.

**Still present on current `main`** (checked 2026-08-26, running server 0.0.34): the HTTP `dispatch` handler in `apps/server/src/orchestration/http.ts` still routes straight through `normalizeDispatchCommand` into the engine with no bootstrap path, and `dispatchBootstrapTurnStart` in `apps/server/src/ws.ts` is still its only consumer.

yashranaway:
## What Changed

- add a dedicated HTTP orchestration command schema
- exclude WebSocket-only turn bootstrap metadata from the HTTP contract
- reject bootstrap requests during payload validation instead of in the handler
- cover the HTTP schema boundary with an authenticated integration test

## Why

The HTTP endpoint advertised the full WebSocket command contract even though bootstrap execution only exists on the WebSocket route. A transport-specific command schema keeps support decisions in the contract and avoids a growing list of handler conditionals.

Closes #8319.

## Testing

- `vp test run apps/server/src/server.test.ts --testNamePattern="rejects WebSocket-only turn bootstrap at the HTTP schema"`
- `vp test run packages/contracts/src/environmentHttp.test.ts`
- `vp lint packages/contracts/src/orchestration.ts packages/contracts/src/environmentHttp.ts apps/server/src/orchestration/http.ts apps/server/src/server.test.ts`
- `vp run --filter @t3tools/contracts typecheck`

## Checklist

- [x] HTTP transport support is represented in contracts
- [x] WebSocket bootstrap behavior is unchanged
- [x] targeted tests, lint, and typechecks pass
- [x] rebased onto latest upstream `main`

Model: GPT-5.6 Sol
Harness: Codex in T3 Code

### DIFF
--- a/apps/server/src/orchestration/http.ts
+++ b/apps/server/src/orchestration/http.ts
@@ -93,9 +93,6 @@ export const orchestrationHttpApiLayer = HttpApiBuilder.group(
         Effect.fn("environment.orchestration.dispatch")(function* (args) {
           yield* annotateEnvironmentRequest(args.endpoint.name);
           yield* requireEnvironmentScope(AuthOrchestrationOperateScope);
-          if (args.payload.type === "thread.turn.start" && args.payload.bootstrap !== undefined) {
-            return yield* failEnvironmentInvalidRequest("http_bootstrap_not_supported");
-          }
           const normalizedCommand = yield* normalizeDispatchCommand(args.payload).pipe(
             Effect.catch(() => failEnvironmentInvalidRequest("invalid_command")),
           );
--- a/apps/server/src/orchestration/http.ts
+++ b/apps/server/src/orchestration/http.ts
@@ -93,6 +93,9 @@ export const orchestrationHttpApiLayer = HttpApiBuilder.group(
         Effect.fn("environment.orchestration.dispatch")(function* (args) {
           yield* annotateEnvironmentRequest(args.endpoint.name);
           yield* requireEnvironmentScope(AuthOrchestrationOperateScope);
+          if (args.payload.type === "thread.turn.start" && args.payload.bootstrap !== undefined) {
+            return yield* failEnvironmentInvalidRequest("http_bootstrap_not_supported");
+          }
           const normalizedCommand = yield* normalizeDispatchCommand(args.payload).pipe(
             Effect.catch(() => failEnvironmentInvalidRequest("invalid_command")),
           );
--- a/apps/server/src/server.test.ts
+++ b/apps/server/src/server.test.ts
@@ -1484,7 +1484,7 @@ it.layer(NodeServices.layer)("server router seam", (it) => {
     }).pipe(Effect.provide(NodeHttpServer.layerTest)),
   );
 
-  it.effect("rejects WebSocket-only turn bootstrap over HTTP with a typed error", () =>
+  it.effect("rejects WebSocket-only turn bootstrap at the HTTP schema", () =>
     Effect.gen(function* () {
       yield* buildAppUnderTest();
 
@@ -1521,18 +1521,7 @@ it.layer(NodeServices.layer)("server router seam", (it) => {
           createdAt,
         }),
       });
-      const body = (yield* response.json) as {
-        readonly _tag: string;
-        readonly code: string;
-        readonly reason: string;
-        readonly traceId: string;
-      };
-
       assert.equal(response.status, 400);
-      assert.equal(body._tag, "EnvironmentRequestInvalidError");
-      assert.equal(body.code, "invalid_request");
-      assert.equal(body.reason, "http_bootstrap_not_supported");
-      assert.equal(typeof body.traceId, "string");
     }).pipe(Effect.provide(NodeHttpServer.layerTest)),
   );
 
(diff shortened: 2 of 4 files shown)


apps/server/src/orchestration/http.ts
```diff
@@ -93,6 +93,9 @@ export const orchestrationHttpApiLayer = HttpApiBuilder.group(
         Effect.fn("environment.orchestration.dispatch")(function* (args) {
           yield* annotateEnvironmentRequest(args.endpoint.name);
           yield* requireEnvironmentScope(AuthOrchestrationOperateScope);
+          if (args.payload.type === "thread.turn.start" && args.payload.bootstrap !== undefined) {
```
juliusmarminge: this is the wrong layer to do this in. there's prob many commands that are not handled over HTTP, a big inline if statement for each is not scalable

juliusmarminge (changes requested):

apps/server/src/orchestration/http.ts
yashranaway: Moved HTTP transport support into a dedicated command schema so the handler no longer needs transport-specific conditionals.

yashranaway:
Withdrawing in favor of the earlier #7996, which implements HTTP bootstrap support instead of rejecting it, to avoid competing fixes.

PR #8635, opened 2026-08-29
fix(web): make new agent replies unmistakable

merged-clem:
Threads could finish without becoming visually distinct: the existing Done label competed with operational statuses and a routed thread was marked read even in a background tab or while older history was visible.

This adds a persistent New reply chip and subtle card highlight independent of working/input/approval state. Completion is now marked read only while T3 Code is foregrounded and the newest timeline content is visible, and a successful user send seeds the read baseline for fresh threads.

Tests:
- focused web tests (4 files, 183 tests)
- @t3tools/web typecheck
- targeted lint and formatting

UI evidence will be added before this leaves draft.

Implemented by GPT-5.6-sol through the Codex harness.

### DIFF
--- a/apps/web/src/components/ChatView.logic.test.ts
+++ b/apps/web/src/components/ChatView.logic.test.ts
@@ -34,6 +34,7 @@ import {
   scheduleEnvironmentReconnectWarning,
   startNewThreadForProject,
   shouldDockDraftHeroForSubmission,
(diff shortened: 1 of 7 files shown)


merged-clem:
Closing because this work is intended for testing in the lubeorg fork first. An upstream PR was not authorized.

PR #8645, opened 2026-08-29
fix(server): name the cause of a failed git command

### LINKED ISSUE #4380: Git command errors discard stderr, making failures opaque to callers
## Problem

`executeGit` in `apps/server/src/vcs/GitVcsDriverCore.ts` records only `stderrLength` on a failed git command and sets `detail` to a static fallback string (`fallbackErrorDetail ?? "Git command exited with a non-zero status."`). The actual stderr text is dropped, so every downstream consumer of `GitCommandError` sees an opaque message with no way to recover the cause.

## Concrete example

While testing the worktree MCP tools (#3754), `git worktree add` failed because the requested branch already existed and was checked out in another worktree. Git's stderr said exactly that. What surfaced instead was:

```
Git command failed in GitVcsDriver.createWorktree (<cwd>): git worktree add failed
```

Diagnosing it required manually re-running the git command outside the app. Any git precondition failure (branch exists, path exists, stale worktree metadata, permissions, missing ref) collapses into the same string.

## Suggested fix

Keep the structural fields as they are, but carry a bounded, useful signal on `GitCommandError`, for example:

- a `stderrExcerpt` capped at a few hundred bytes, or
- a parsed `reason` enum for common git failures (ref exists, path exists, not a repository, permission denied), with the excerpt as fallback.

If dropping stderr is deliberate (telemetry hygiene, path leakage), a capped and sanitized excerpt on the in-process error object only, excluded from telemetry serialization, would still fix the developer experience without changing what leaves the machine.

walid-baharwal:
## What Changed

`GitCommandError` gains an optional `reason`: a closed set of diagnostic tags for well-known
git failures, recognized from stderr inside the driver and appended to the message the way
`EnvironmentInternalError` does it. The tag names the cause; it selects no text.

Before and after, for the same failing command (real git, captured from the driver):

```
Git command failed in GitVcsDriver.createWorktree (/tmp/…/repo): git worktree add failed
Git command failed in GitVcsDriver.createWorktree (/tmp/…/repo): git worktree add failed (branch_checked_out_in_worktree)
```

Raw stderr still never leaves the driver.

## Why

Fixes #4380. Every git precondition failure collapsed into one string, so a tag conflict, an
auth failure and a branch already checked out elsewhere were indistinguishable without
re-running the command by hand.

The issue offers two options. The first — carrying a bounded `stderrExcerpt` — is the one I
did **not** take: `apps/server/src/vcs/GitVcsDriverCore.test.ts` already asserts
`notProperty(error, "stderr")` and that a secret passed in argv never reaches
`error.message`, so dropping stderr is deliberate, and shipping an excerpt would mean
deleting a security test. This is the issue's second option, the parsed reason, which keeps
that guarantee intact: the tags are a closed literal union, the sentences are static, and
nothing matched from stderr is ever interpolated. The existing redaction test is extended
with `notProperty(error, "reason")`.

Classification happens at the two shared funnels every git call routes through
(`executeGit` and the non-zero-exit branch of the raw executor), which covers
`fetchRemote` and `createWorktree` — the two operations named in the issue.

Two limits, stated rather than hidden:

- **Only git's own diagnostic lines are classified, plus the refusals ssh prints.** `remote:`
  is deliberately excluded: git prefixes every byte the server sends that way, remote hook
  output included, so trusting it would reintroduce the same false positive from the far end.
  Every source that can reach the classifier is covered by a test — local hook output, remote
  hook output, five ssh refusal shapes, and an OS-level permission error that must not read as
  ssh auth. The one accepted residue is that a local hook echoing an ssh refusal verbatim would
  still be classified; that text is narrow enough that the tag would arguably still be right. Hooks write to the same stream
  unprefixed, so a `pre-push` hook echoing "authentication failed" would otherwise be reported
  as a credential failure (reproduced, and now covered by a test that pushes through a failing
  hook). ssh states a key refusal unprefixed as well, so those specific lines stay eligible —
  otherwise git's generic "could not read from remote repository" would be read as an
  unreachable remote when the real cause is credentials.
- **Classification is English-only.** `executeGit` inherits the process locale, so under a
  non-English `LANG` git's wording does not match and the error simply keeps today's
  behavior — no reason, no regression. Forcing `LC_ALL=C` for every git command would fix
  that, but it changes the environment of every call in the driver and belongs in its own
  change. The new tests pin `LC_ALL: "C"` so they do not depend on the runner's locale.
- Five other `GitCommandError` constructions hold stderr in hand and stay unclassified.
  Adding them is one line each and deliberately left out to keep this to one concern.

## UI Changes

None. Server-side error metadata; no rendered change. The improved text surfaces wherever
an existing git error message is already shown.

## Verification

```bash
vp test run apps/server/src/vcs/GitVcsDriverCore.test.ts   # 68 passed
cd packages/contracts && tsgo --noEmit                     # clean
cd apps/server && tsgo --noEmit                            # clean
vp lint apps/server/src/vcs/GitVcsDriverCore.ts apps/server/src/vcs/GitVcsDriverCore.test.ts packages/contracts/src/git.ts
vp format --check <same three files>
git diff --check
```

The new tests drive real git through the real driver: a branch already checked out in
another worktree, a command outside a repository, a tag collision that must stay unclassified, a failing
pre-push hook whose output must not be classified, and a fallback detail that must not run
into the reason sentence. Each fails without the source change.

## Checklist

- [x] One concern
- [x] Focused tests, failing before the fix
- [x] Typecheck, lint and format run on the changed packages
- [x] No new dependencies, no committed artifacts
- [x] Additive optional contract field; web, mobile and desktop need no change (no reference
      to `GitCommandError` fields exists in any client)

Model: Claude Opus 5 (1M context). Harness: Claude Code.

---

> [!NOTE]
> **Medium Risk**
> Changes how git failures are surfaced (new optional field and message suffix) and classifies auth/SSH-related stderr, though patterns are conservative and stderr still never leaves the driver.
> 
> **Overview**
> **`GitCommandError` now carries an optional `reason`** from a closed `GitCommandFailureReason` union in contracts, and the error **message appends ` (reason)`** when classification succeeds—without ever exposing raw stderr.
> 
> The git VCS driver adds **`classifyGitFailure`**, which pattern-matches filtered diagnostic lines (plus specific SSH refusal lines) at the shared non-zero-exit paths in **`executeRaw`** and **`executeGit`**. Callers can distinguish worktree branch conflicts, missing repos, auth vs host-key failures, and similar cases while the redaction guarantee stays intact (no `stderr`, no matched text in messages). Classification deliberately ignores hook/`remote:` noise and leaves ambiguous failures (e.g. duplicate tags) untagged; matching is **English-only** unless git runs under `LC_ALL=C`.
> 
> Integration tests exercise real git scenarios and extend the secret-leak test to assert **`reason` is not set** when stderr would be unsafe to interpret.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61b898117f28bef778989fbcde49c50a4d0b5c69. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add `reason` tag to `GitCommandError` for classified git failures
> - Introduces `GitCommandFailureReason` schema in [git.ts](https://github.com/pingdotgg/t3code/pull/8645/files#diff-2282301d92a7a9ddc7e8d88b986c9628fdf70497dc6a09fe80eb8d46d6a769e7) with literal tags like `authentication_failed`, `not_a_repository`, `branch_checked_out_in_worktree`, and `host_key_unverified`.
> - Adds `classifyGitFailure(stderr)` in [GitVcsDriverCore.ts](https://github.com/pingdotgg/t3code/pull/8645/files#diff-1c3e2b9763c6526006485633fb677c83dc54cc1ea5563a1e51aab707c35854d8) that matches stderr lines against ordered regex patterns to produce a reason tag, returning `null` when no pattern matches.
> - Both `executeRaw` and `executeGitWithStableDiagnostics` now call `classifyGitFailure` on non-zero exits and attach the reason to the thrown `GitCommandError`.
> - `GitCommandError.message` now appends ` (reason)` when a reason is present, so callers that parse error messages see a new suffix.
> - Risk: any code that string-matches `GitCommandError` messages or asserts exact error text will break due to the appended ` (reason)` suffix; the `reason` field is optional so existing `.reason` checks are unaffected.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 61b8981.</sup>
> 
>

### DIFF
--- a/apps/server/src/vcs/GitVcsDriverCore.test.ts
+++ b/apps/server/src/vcs/GitVcsDriverCore.test.ts
@@ -712,6 +712,279 @@ it.layer(TestLayer)("GitVcsDriver core integration", (it) => {
         assert.notInclude(error.message, secret);
         assert.notProperty(error, "args");
         assert.notProperty(error, "stderr");
+        assert.notProperty(error, "reason");
+      }),
+    );
+
+    it.effect("names the branch conflict behind a failed worktree add", () =>
+      Effect.gen(function* () {
+        const parent = yield* makeTmpDir();
+        const pathService = yield* Path.Path;
+        const cwd = pathService.join(parent, "repo");
+        const fileSystem = yield* FileSystem.FileSystem;
+        yield* fileSystem.makeDirectory(cwd);
+        const driver = yield* GitVcsDriver.GitVcsDriver;
+        yield* initRepoWithCommit(cwd);
+        const firstWorktree = pathService.join(parent, "first");
+        yield* git(cwd, ["worktree", "add", firstWorktree, "-b", "shared-branch"]);
+
+        const error = yield* driver
+          .execute({
+            operation: "GitVcsDriver.test.worktreeAddConflict",
+            cwd,
+            args: ["worktree", "add", pathService.join(parent, "second"), "shared-branch"],
+            env: { LC_ALL: "C" },
+          })
+          .pipe(Effect.flip);
+
+        assert.deepInclude(error, {
+          _tag: "GitCommandError",
+          operation: "GitVcsDriver.test.worktreeAddConflict",
+          reason: "branch_checked_out_in_worktree",
+        });
+        assert.include(error.message, "(branch_checked_out_in_worktree)");
+        assert.notInclude(error.message, firstWorktree);
+      }),
+    );
+
+    it.effect("names a missing repository behind a failed command", () =>
+      Effect.gen(function* () {
+        const cwd = yield* makeTmpDir();
+        const driver = yield* GitVcsDriver.GitVcsDriver;
+
+        const error = yield* driver
+          .execute({
+            operation: "GitVcsDriver.test.notARepository",
+            cwd,
+            args: ["rev-parse", "--abbrev-ref", "HEAD"],
+            env: { LC_ALL: "C" },
+          })
+          .pipe(Effect.flip);
+
+        assert.deepInclude(error, {
+          _tag: "GitCommandError",
+          operation: "GitVcsDriver.test.notARepository",
+          reason: "not_a_repository",
+        });
+        assert.include(error.message, "(not_a_repository)");
+      }),
+    );
+
+    it.effect("appends the reason tag to a caller-supplied detail", () =>
+      Effect.gen(function* () {
+        const parent = yield* makeTmpDir();
+        const pathService = yield* Path.Path;
+        const cwd = pathService.join(parent, "repo");
+        const fileSystem = yield* FileSystem.FileSystem;
+        yield* fileSystem.makeDirectory(cwd);
+        const driver = yield* GitVcsDriver.GitVcsDriver;
+        yield* initRepoWithCommit(cwd);
+        yield* git(cwd, [
+          "worktree",
+          "add",
+          pathService.join(parent, "taken"),
+          "-b",
+          "taken-branch",
+        ]);
+
+        const error = yield* driver
+          .createWorktree({
+            cwd,
+            refName: "taken-branch",
+            path: pathService.join(parent, "second"),
+          })
+          .pipe(Effect.flip);
+
+        assert.equal(error.detail, "git worktree add failed");
+        assert.include(error.message, "git worktree add failed (branch_checked_out_in_worktree)");
+      }),
+    );
+
+    for (const { label, sshMessage, expectedReason, expectedText } of [
+      {
+        label: "a refused key",
+        sshMessage: "git@example.invalid: Permission denied (publickey).",
+        expectedReason: "authentication_failed" as const,
+        expectedText: "(authentication_failed)",
+      },
+      {
+        label: "a refused key among several methods",
+        sshMessage: "git@example.invalid: Permission denied (publickey,password).",
+        expectedReason: "authentication_failed" as const,
+        expectedText: "(authentication_failed)",
+      },
+      {
+        label: "a refused keyboard-interactive attempt",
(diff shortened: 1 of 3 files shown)


t3dotgg:
> [!NOTE]
>

Carryover from [#5620](https://github.com/pingdotgg/t3code/pull/5620) at [dc881848a2](https://github.com/pingdotgg/t3code/commit/dc881848a2410be17ed954a4160c74a702ca42d5): retain the `GitVcsDriverCore.test.ts` locale cases. Set `LC_ALL=C` when stderr will be classified, in both the direct execution and wrapper paths. When `allowNonZeroExit` returns raw output, preserve the caller's locale. Keep diagnostics limited to safe failure tags and lengths. These tests cover locale-dependent matching and do not require restoring raw stderr in public errors.

shivamhwp:

The SSH exception still lets remote hook output through `classifyGitFailure`. A rejecting remote `pre-receive` hook that prints `Permission denied (publickey).` produces a `remote:` line, but `SSH_TRANSPORT_REFUSAL_PATTERN` accepts it and assigns `authentication_failed`. Printing `Host key verification failed.` similarly assigns `host_key_unverified`, even though transport succeeded and the hook rejected the push.

Exclude `remote:` lines before applying either diagnostic pattern. Keep the unprefixed SSH refusal handling for actual transport errors. The current remote-hook case uses only `authentication failed`, which does not exercise this exception.

PR #8682, opened 2026-08-29
fix(cli): avoid unsafe project offline fallback

### LINKED ISSUE #7504: [Bug]: `t3 project` CLI treats any 1s live-server probe failure as "no server": deletes server-runtime.json and writes offline behind a running server
### What happened

While bulk-registering projects on a Windows machine I read how `t3 project add|remove|rename` decides between talking to the running server and writing to `state.sqlite` directly, and found that the live-server probe is bounded to **1 second** and that *any* probe failure (timeout, auth error, undeclared status) is treated the same as "no server": the CLI deletes the running server's `server-runtime.json` and then applies the mutation **offline** against the database the live server still has open.

### Diagnosis

Grounded in the installed nightly bundle (`apps/server/dist/bin.mjs`, `//#region src/cli/project.ts`):

- `PROJECT_CLI_LIVE_SERVER_TIMEOUT = seconds(1)`; `fetchLiveOrchestrationSnapshot` and `dispatchLiveOrchestrationCommand` are both wrapped in `withProjectCliLiveServerTimeout`.
- `tryResolveLiveProjectExecutionMode` reads `server-runtime.json`, calls `fetchLiveOrchestrationSnapshot`, and on **any** failure does:
  ```
  logDebug("Failed to connect to the persisted project CLI server.", ...)
  yield* clearPersistedServerRuntimeState(config.serverRuntimeStatePath)
  return none()
  ```
  It does not distinguish `ECONNREFUSED` (server really gone) from a timeout / 401 / 5xx (server alive but slow or rejecting).
- `runProjectMutation` then builds `ProjectCliRuntimeLive` (its own `OrchestrationEngine` over the same `state.sqlite`) and dispatches the command in `mode: "offline"`.

Consequences when the probe fails against a *live* server:

1. The running server's `server-runtime.json` is deleted, so `t3 pair`, `t3 connect` and later `t3 project` calls can no longer discover it until the server is restarted (the server only writes the file at startup).
2. The project mutation is appended to the event store behind the live server's back. The live engine's in-memory read model is not updated, so the project does not appear in the open desktop/web UI until restart, even though the CLI printed "Added project ...".
3. Two `OrchestrationEngine` instances (live + offline) write the same event store and `projection_state` concurrently.

A 1 s bound is easy to miss on Windows: `server.getConfig` is already known to block for up to 5 s on this platform (#4697, #5137), and `orchestration.snapshot` grows with projects/threads (64 projects here).

What is verified vs. derived: the code path, constants and the unconditional `clearPersistedServerRuntimeState` are read directly from the shipped bundle and match `src/cli/project.ts` on `main`. I did **not** force the timeout empirically on this machine (my bulk registration was done with the desktop app fully closed, precisely to avoid this path); the consequences above follow from the code.

Suggested fix (small): only clear the runtime file and fall back to offline when the connection is refused **or** the recorded pid is not alive; on timeout/auth/undeclared-status errors fail loudly ("live server at <origin> did not answer within 1s; retry, or stop the server and rerun") instead of silently writing offline. Optionally raise the bound to the 5 s used by `CLOUD_CLI_LIVE_SERVER_TIMEOUT`.

### Steps to reproduce

Deterministic repro requires making `orchestration.snapshot` exceed 1 s; by construction:

1. Run the desktop app (server writes `<T3 home>/userdata/server-runtime.json`).
2. Make the server slow to answer `orchestration.snapshot` for >1 s (large project/thread count, or pause the server process for 2 s under a debugger).
3. `npx t3 project add <some-repo>` (or `node apps/server/dist/bin.mjs project add ...`).
4. Observe: CLI prints "Added project ..." in offline mode, `server-runtime.json` is gone, and the project is absent from the running UI until restart.

### Version

0.0.34-nightly.20260819.1132 (commit 36f4314ab768); same code on `main` today.

### Environment

Windows 11 Pro 10.0.26100, T3 Code (Nightly) desktop, Node 24.11.1.

### Evidence

```shell
# installed bundle, src/cli/project.ts region
const PROJECT_CLI_LIVE_SERVER_TIMEOUT = seconds(1);
const withProjectCliLiveServerTimeout = (effect) => effect.pipe(timeout(PROJECT_CLI_LIVE_SERVER_TIMEOUT));
const fetchLiveOrchestrationSnapshot = (origin, bearerToken) => gen(function* () { ... }).pipe(withProjectCliLiveServerTimeout, mapError(projectCommandErrorFromLiveServerRequest));
const tryResolveLiveProjectExecutionMode = fn$1("tryResolveLiveProjectExecutionMode")(function* (environmentAuth, config) {
  const runtimeState = yield* readPersistedServerRuntimeState(config.serverRuntimeStatePath);
  if (isNone(runtimeState)) return none();
  const attempted = yield* result$1(withProjectCliSessionToken(environmentAuth, (token) => fetchLiveOrchestrationSnapshot(runtimeState.value.origin, token).pipe(...)));
  if (attempted._tag === "Success") return some(attempted.success);
  yield* logDebug("Failed to connect to the persisted project CLI server.", { origin: runtimeState.value.origin, cause: attempted.failure });
  yield* clearPersistedServerRuntimeState(config.serverRuntimeStatePath);   // <- unconditional
  return none();
});
```

Related observation (not the subject of this issue): after a clean quit of the Windows desktop app (`CloseMainWindow` on the main window; all `T3 Code (Nightly).exe` processes exited) `server-runtime.json` was left behind containing the dead server pid, i.e. the server's release finalizer that normally clears it (`//#region src/cli/server.ts`, `persistServerRuntimeState` / `clearPersistedServerRuntimeState`) did not run on this platform. A stale file is exactly what makes the offline fallback above attractive, but it is also what makes the "any failure" heuristic risky: the CLI cannot tell a stale file from a live-but-slow server.

### Related issues

#5749 / #6097 (competing runtimes against the same database) are about the service/desktop split, not the project CLI; no existing issue mentions the 1 s probe or the runtime-file deletion.

### Fix applied or workaround

None needed for me: I only used the offline path with the desktop fully closed. Workaround for others: quit T3 completely before using `t3 project ...`, or check that the project appears in the UI afterwards.

### Filed by

Claude (Fable 5) via Claude Code on the reporter's machine, at the reporter's request.

dlowzzxx:
Fixes #7504

`t3 project` previously treated every failed live-server probe as proof that the server was gone: it removed `server-runtime.json` and then mutated the local store offline. A timeout, authorization failure, malformed origin, or HTTP error could therefore write behind a still-running server.

This change clears the persisted runtime marker and uses the existing offline path only for a structural HTTP transport failure whose nested error code is exactly `ECONNREFUSED`. All other probe failures preserve the marker and surface the existing project command error, preventing an unsafe fallback.

Tests:
- focused CLI regression coverage for both the refused-endpoint fallback and non-refused failure (marker preserved, no offline mutation)
- structural classifier coverage for direct and nested `ECONNREFUSED`, plus non-refused errors
- `vp test run apps/server/src/cli/project.test.ts apps/server/src/bin.test.ts` — 22 passed (bundled Node 24)
- formatter check passed

---

> [!NOTE]
> **Medium Risk**
> Changes project CLI execution mode selection when the live server probe fails; non-refused failures now error instead of silently falling back to offline orchestration writes.
> 
> **Overview**
> **Tightens when `t3 project` may mutate the local store while a persisted live-server marker exists.** A failed live-server probe no longer always deletes `server-runtime.json` and runs offline mutations.
> 
> Only **connection refused** (`ECONNREFUSED` on the HTTP transport cause chain) clears the persisted runtime endpoint and allows offline fallback. Other probe failures (invalid origin, timeouts, auth errors, etc.) **keep the runtime marker** and return **`ProjectLiveServerRequestError`** instead of writing behind a possibly still-running server.
> 
> Adds **`isProjectLiveServerConnectionRefusedError`** (with strict structural checks, not message parsing) and wires it into **`tryResolveLiveProjectExecutionMode`**. Integration tests cover refused-port offline success vs non-refused failure with no snapshot change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dfbdca5d5c2b7b6e99ae3f72d77ee473f8f49f6a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Restrict CLI offline fallback to `ECONNREFUSED` failures in `tryResolveLiveProjectExecutionMode`
> - When the CLI probes a persisted live server, only `ECONNREFUSED` errors clear the runtime state and fall back to offline mode. Other probe failures now propagate the error instead of silently mutating the offline snapshot.
> - Adds `isProjectLiveServerConnectionRefusedError`, `hasConnectionRefusedCode`, and `isProjectLiveServerRequestError` utilities in [project.ts](https://github.com/pingdotgg/t3code/pull/8682/files#diff-44f1264ca9e5b2f3994c44291b5616390842fc6345021ca54e8447fd8dbbc1cd) to classify transport failures with an exact `ECONNREFUSED` code, walking the cause chain with cycle protection.
> - Risk: callers that previously relied on any probe failure triggering offline fallback will now receive a `ProjectLiveServerRequestError` for non-refused failures (e.g. timeout, DNS error). Verify no out-of-tree consumers depend on the old catch-all fallback.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dfbdca5.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Project commands now automatically fall back to offline mode when the live server cannot be reached.
  - Persisted live-server state is cleared when the connection is refused, preventing future connection attempts from using stale information.
  - Other live-server errors are preserved and reported instead of triggering offline changes.
  - Improved error handling distinguishes genuine connection refusals from unrelated network or application errors.


lnieuwenhuis:
Maintainer verification: the fallback restriction is tight — only a structural TransportError with an exact code ECONNREFUSED through a cycle-safe cause walk clears the runtime marker, so timeouts, resets, auth failures, and message-text-only matches all preserve it. The refused-port test reserves a loopback port rather than hardcoding one, which keeps it portable including Windows. Looks merge-ready to me.

PR #8746, opened 2026-08-30
fix(windows): strip quotes from repaired PATH

### LINKED ISSUE #7543: [Bug]: Windows PATH repair passes stray quotes through, so cmd.exe cannot find node in the integrated terminal

UtkarshUsername:
## What Changed

Strip quote characters before merging Windows PATH entries in the shared server and desktop repair paths, with coverage for quoted and stray-quote entries.

## Why
A malformed quote in a Windows PATH entry reaches cmd.exe through T3 Code's PATH repair, preventing command lookups after that entry.

Closes #7543

## Checklist

Built with GPT-5.6-terra in T3 Code.

---

> [!NOTE]
> **Medium Risk**
> Changes how Windows PATH is assembled at startup and in shared shell resolution; incorrect sanitization could drop or alter valid entries, though scope is limited to quote stripping with new tests.
> 
> **Overview**
> Fixes Windows command lookup failures when **PATH repair** leaves quote characters in merged entries (e.g. stray `C:";` segments that break `cmd.exe` parsing).
> 
> **Windows-only** `sanitizePathEntry` now removes all `"` from each PATH segment before dedupe and merge in **`mergePathValues`** (`packages/shared`) and **`mergePaths`** (desktop shell environment install). Quoted paths like `"C:\\Program Files\\nodejs"` are normalized to unquoted form while case-insensitive deduping still applies.
> 
> Tests cover fully quoted entries, stray mid-entry quotes, and the desktop PowerShell profile merge scenario.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54abdcbf666f3c60de46e56bd2712153afe7eb16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Strip stray quotes from PATH entries on Windows in `mergePathValues` and `mergePaths`
> - Adds a `sanitizePathEntry` helper that removes all double-quote characters from each PATH entry on Windows (no-op on other platforms) in both [shell.ts](https://github.com/pingdotgg/t3code/pull/8746/files#diff-89a91f85db41a9a0912bc86c336bf9429751f3797ee50a089248c68cde8489ce) and [DesktopShellEnvironment.ts](https://github.com/pingdotgg/t3code/pull/8746/files#diff-ed26aa44c93ddc9fedab8bc7cf7aa64c7b0b8fc0e22a3960d5c02c30c940dabd).
> - Both merge routines now trim, then sanitize, then check emptiness, normalize, and de-duplicate based on the unquoted form before pushing the sanitized value to the output.
> - Behavioral Change: on Windows, merged PATH entries are emitted without quotes and entries that differ only by surrounding quotes are treated as duplicates.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 54abdcb.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/shell/DesktopShellEnvironment.test.ts
+++ b/apps/desktop/src/shell/DesktopShellEnvironment.test.ts
@@ -320,7 +320,7 @@ describe("DesktopShellEnvironment", () => {
                 FNM_DIR: "C:\\Users\\testuser\\AppData\\Roaming\\fnm",
                 FNM_MULTISHELL_PATH: "C:\\Users\\testuser\\AppData\\Local\\fnm_multishells\\123",
               })
-            : envOutput({ PATH: "C:\\Custom\\Bin;C:\\Windows\\System32" });
+            : envOutput({ PATH: 'C:\\Custom\\Bin;C:";C:\\Windows\\System32' });
         },
       });
 
@@ -337,6 +337,7 @@ describe("DesktopShellEnvironment", () => {
           "C:\\Users\\testuser\\.bun\\bin",
           "C:\\Users\\testuser\\scoop\\shims",
           "C:\\Custom\\Bin",
+          "C:",
         ].join(";"),
       );
       assert.equal(env.FNM_DIR, "C:\\Users\\testuser\\AppData\\Roaming\\fnm");
--- a/apps/desktop/src/shell/DesktopShellEnvironment.ts
+++ b/apps/desktop/src/shell/DesktopShellEnvironment.ts
@@ -151,6 +151,9 @@ const pathComparisonKey = (entry: string, platform: NodeJS.Platform) => {
   return platform === "win32" ? normalized.toLowerCase() : normalized;
 };
 
+const sanitizePathEntry = (entry: string, platform: NodeJS.Platform) =>
+  platform === "win32" ? entry.replaceAll('"', "") : entry;
+
 const mergePaths = (
   platform: NodeJS.Platform,
   values: ReadonlyArray<Option.Option<string>>,
@@ -163,14 +166,14 @@ const mergePaths = (
     if (Option.isNone(value)) continue;
 
     for (const entry of value.value.split(delimiter)) {
-      const trimmed = entry.trim();
-      if (trimmed.length === 0) continue;
+      const sanitized = sanitizePathEntry(entry.trim(), platform);
+      if (sanitized.length === 0) continue;
 
-      const key = pathComparisonKey(trimmed, platform);
+      const key = pathComparisonKey(sanitized, platform);
       if (key.length === 0 || seen.has(key)) continue;
 
       seen.add(key);
-      entries.push(trimmed);
+      entries.push(sanitized);
     }
   }
 
--- a/packages/shared/src/shell.test.ts
+++ b/packages/shared/src/shell.test.ts
@@ -304,7 +304,7 @@ describe("readEnvironmentFromWindowsShell", () => {
 });
 
 describe("mergePathValues", () => {
-  it("dedupes case-insensitively on Windows while preserving preferred order", () => {
+  it("sanitizes and dedupes Windows entries while preserving preferred order", () => {
     expect(
       mergePathValues(
         'C:\\Users\\testuser\\AppData\\Roaming\\npm;"C:\\Program Files\\nodejs"',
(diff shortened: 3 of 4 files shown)


CDVolvik:
Verified fix(windows): strip quotes from repaired PATH at 54abdcb. Windows-only quote strip in both repair paths (packages/shared/src/shell.ts:417 and DesktopShellEnvironment.ts:154) correctly sanitizes entry.trim() before dedupe, so fully quoted and stray-quote cases normalize and do not poison cmd.exe lookup. Covers #7543. Tested pnpm exec vitest run packages/shared/src/shell.test.ts 30/30 and DesktopShellEnvironment.test.ts 14/14 pass. Revert clean. LGTM as CDVolvik.

Issue #6116, opened 2026-08-11
[Bug]: Windows - CodexAppServerRequestError: Invalid payload for method 'initialize' during 'decode-payload'


riccardoperra:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

Not sure how to reproduce this, but 've just installed the app on windows the first time (both stable and nightly version), but I can't setup Codex provider, since it throws that error:

- CodexAppServerRequestError: Invalid payload for method 'initialize' during 'decode-payload'

Note: My PC already have both codex-cli and codex app installed without any issue. I can type `codex` in my terminal (both powershell and bash)

### Expected behavior

Codex App Provider will connect on windows without any trouble

### Actual behavior

Codex App provider throws a "CodexAppServerRequestError" error

### Impact

Blocks work completely

### Version or commit

main

### Environment

Windows

### Logs or stack traces

```shell
CodexAppServerRequestError: Invalid payload for method 'initialize' during 'decode-payload'
    at CodexAppServerRequestError.invalidPayload (file:///C:/Users/username/AppData/Local/Programs/t3code/resources/app.asar/apps/server/dist/bin.mjs:63376:10)
    at file:///C:/Users/username/AppData/Local/Programs/t3code/resources/app.asar/apps/server/dist/bin.mjs:80908:110
    at file:///C:/Users/username/AppData/Local/Programs/t3code/resources/app.asar/node_modules/effect/dist/internal/effect.js:1383:96
    at ~effect/Utils/internal (file:///C:/Users/username/AppData/Local/Programs/t3code/resources/app.asar/node_modules/effect/dist/Utils.js:79:14)
    at probeCodexAppServerProvider (file:///C:/Users/username/AppData/Local/Programs/t3code/resources/app.asar/apps/server/dist/bin.mjs:81516:29)
    at probeCodexAppServerProvider (definition) (file:///C:/Users/username/AppData/Local/Programs/t3code/resources/app.asar/apps/server/dist/bin.mjs:81381:44)
    at checkCodexProviderStatus (file:///C:/Users/username/AppData/Local/Programs/t3code/resources/app.asar/apps/server/dist/bin.mjs:84014:25)
    at checkCodexProviderStatus (definition) (file:///C:/Users/username/AppData/Local/Programs/t3code/resources/app.asar/apps/server/dist/bin.mjs:81498:41)
[...]

PR #8748, opened 2026-08-30
fix(shared): preserve Windows shell PATH priority

### LINKED ISSUE #7570: Windows: codex binary resolution prefers stale %APPDATA%\npm shims, breaking the Codex provider with a cryptic error

UtkarshUsername:
Closes #7570 

## What Changed

Preserved the user shell PATH order during Windows environment hydration. Known package-manager CLI directories now append as fallbacks rather than taking priority.

## Why

A stale %APPDATA%\\npm\\codex.cmd shim could be selected ahead of the user's working Codex executable, causing an opaque app-server protocol error. Keeping shell PATH precedence makes bare provider commands resolve the same way they do in the user's terminal.

## Checklist

Model: gpt-5.6-terra
Harness: Codex

---

> [!NOTE]
> **Medium Risk**
> Changes how every Windows spawn resolves commands via `resolveWindowsEnvironment` (e.g. server os-jank), so wrong ordering could still break CLI discovery—but the change is narrowly scoped to PATH merge precedence with tests covering the main flows.
> 
> **Overview**
> **Windows PATH hydration** now keeps shell and inherited entries ahead of hard-coded package-manager directories when building the baseline environment in `resolveWindowsEnvironment`.
> 
> The merge order flips from *known CLI dirs first* to *merged shell PATH first*, with npm/nodejs/Volta/pnpm/bun/scoop paths appended only as fallbacks for GUI-launched processes that lack a full interactive PATH. That avoids stale shims (e.g. `%APPDATA%\npm\codex.cmd`) winning over the same commands the user gets in a terminal.
> 
> `shell.test.ts` expectations were updated for the baseline, profile-loaded, and profile-fallback cases so PATH ordering matches the new priority rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34886fd439a082ee08f778d4c1712a35f5cc244f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/shared/src/shell.test.ts
+++ b/packages/shared/src/shell.test.ts
@@ -450,7 +450,7 @@ effectIt.layer(NodeServices.layer)("resolveSpawnCommand", (it) => {
 });
 
 effectIt.layer(NodeServices.layer)("resolveWindowsEnvironment", (it) => {
-  it.effect("returns the baseline no-profile PATH patch when node is already available", () =>
+  it.effect("uses known CLI directories as a fallback without changing shell PATH priority", () =>
     Effect.gen(function* () {
       const readEnvironment = vi.fn(
         (_names: ReadonlyArray<string>, options?: { loadProfile?: boolean }) =>
@@ -473,6 +473,8 @@ effectIt.layer(NodeServices.layer)("resolveWindowsEnvironment", (it) => {
         ),
       ).toEqual({
         PATH: [
+          "C:\\Shell\\Bin",
+          "C:\\Windows\\System32",
           "C:\\Users\\testuser\\AppData\\Roaming\\npm",
           "C:\\Users\\testuser\\AppData\\Local\\Programs\\nodejs",
           "C:\\Users\\testuser\\AppData\\Local\\Volta\\bin",
@@ -480,8 +482,6 @@ effectIt.layer(NodeServices.layer)("resolveWindowsEnvironment", (it) => {
           "C:\\Users\\testuser\\.local\\bin",
           "C:\\Users\\testuser\\.bun\\bin",
           "C:\\Users\\testuser\\scoop\\shims",
-          "C:\\Shell\\Bin",
-          "C:\\Windows\\System32",
         ].join(";"),
       });
       expect(readEnvironment).toHaveBeenCalledTimes(1);
@@ -522,6 +522,7 @@ effectIt.layer(NodeServices.layer)("resolveWindowsEnvironment", (it) => {
         PATH: [
           "C:\\Profile\\Node",
           "C:\\Windows\\System32",
+          "C:\\Shell\\Bin",
           "C:\\Users\\testuser\\AppData\\Roaming\\npm",
           "C:\\Users\\testuser\\AppData\\Local\\Programs\\nodejs",
           "C:\\Users\\testuser\\AppData\\Local\\Volta\\bin",
@@ -529,7 +530,6 @@ effectIt.layer(NodeServices.layer)("resolveWindowsEnvironment", (it) => {
           "C:\\Users\\testuser\\.local\\bin",
           "C:\\Users\\testuser\\.bun\\bin",
           "C:\\Users\\testuser\\scoop\\shims",
-          "C:\\Shell\\Bin",
         ].join(";"),
         FNM_DIR: "C:\\Users\\testuser\\AppData\\Roaming\\fnm",
         FNM_MULTISHELL_PATH: "C:\\Users\\testuser\\AppData\\Local\\fnm_multishells\\123",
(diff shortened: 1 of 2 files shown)


CDVolvik:
Verified fix(shared): preserve Windows shell PATH priority (#8748) on pr-8748 (34886fd4).

Environment: Windows, pnpm exec vitest run packages/shared/src/shell.test.ts — mocked WindowsShellEnvironment/CommandAvailability, not a live pwsh.exe codex app-server spawn.

Scope: packages/shared/src/shell.ts:726-729 flips baselinePath from mergePathValues(knownCliPath, mergedPath) to mergePathValues(mergedPath, knownCliPath) — shell/inherited PATH keeps priority, resolveKnownWindowsCliDirs (npm/nodejs/Volta/pnpm/bun/scoop) appends only as fallback for GUI-launched processes. Matches #7570 root cause (stale %APPDATA% npm codex.cmd winning over where codex) and prior #6116 pattern.

Tests:
- PR: pnpm exec vitest run packages/shared/src/shell.test.ts → 29/29 PASS (3x resolveWindowsEnvironment + 26 others)
- Revert check (shell.ts → main ordering, PR tests kept): 3 FAILED / 26 PASS — PATH prefix vs suffix mismatch as expected. Restore → 29 PASS again.

Surgical, no drive-bys. Would be good to also log resolved command + first stdout bytes on decode-wire-message failures in a follow-up, but out of scope here.

Verified as CDVolvik on Windows.

PR #8753, opened 2026-08-30
fix(desktop): recover from obsolete catalog keys

### LINKED ISSUE #8341: [Bug]: Orphaned safeStorage key makes connection-catalog.json permanently undecryptable, blocking all target listing on Windows
### Before submitting

### Area

apps/desktop

### Steps to reproduce

The trigger is an Electron `userData` directory that gets recreated while `~/.t3/userdata` survives. On Windows that happens on reinstall / profile reset / roaming-dir cleanup.

1. Use the desktop app so at least one connection is saved. `connection-catalog.json` is written to `%USERPROFILE%\.t3\userdata\`, encrypted via `safeStorage`.
2. Delete or let something recreate `%APPDATA%\t3code\` (the Electron `userData` dir), leaving `%USERPROFILE%\.t3\userdata\` intact.
3. Launch the app. Electron generates a fresh random `os_crypt.encrypted_key` in the new `Local State`.
4. Open Settings → Connections, or try to add/select an environment.

### Expected behavior

An undecryptable catalog should degrade gracefully — treat it as "no saved connections", log a warning, and let the next write re-key the file with the current `safeStorage` key. Losing saved connection entries is acceptable; being unable to reach the connections UI at all is not.

### Actual behavior

The connections UI is dead and no environment can be added or listed:

```
Could not list targets: ConnectionTransientError: Could not load the local connection catalog:
Error: Error invoking remote method 'desktop:get-connection-catalog':
DesktopConnectionCatalogStoreProtectionError: Desktop connection catalog protection failed
during decrypt-catalog at C:\Users\<user>\.t3\userdata\connection-catalog.json.
```

Labelled `ConnectionTransientError`, but it is permanent — every launch fails identically until the file is deleted by hand.

In `app.asar`, `connectionCatalogStore.get` maps any `safeStorage.decryptString` failure straight into a fatal `DesktopConnectionCatalogStoreProtectionError`, which propagates out of `desktop.bootstrap`. There is no fallback branch, unlike the success path that returns `Option.some(decrypted)`:

```js
safeStorage.decryptString(encryptedCatalog).pipe(
  Effect.mapError(
    (cause) =>
      new DesktopConnectionCatalogStoreProtectionError({
        operation: "decrypt-catalog",
        catalogPath,
        cause,
      }),
  ),
)
```

### Root cause (verified on my machine)

The ciphertext was encrypted with an AES key that no longer exists. DPAPI itself is healthy — this is not a Windows credential problem.

- Inner cause chain from `desktop.trace.ndjson`: `DesktopConnectionCatalogStoreProtectionError` → `ElectronSafeStorageDecryptError` → `Error while decrypting the ciphertext provided to safeStorage.decryptString`.
- `encryptedCatalog` base64-decodes to a `v10`-prefixed blob (622 bytes: `v10` + 12-byte nonce + 607 bytes ciphertext+tag), i.e. Chromium AES-256-GCM. The key lives in `%APPDATA%\t3code\Local State` → `os_crypt.encrypted_key`, DPAPI-wrapped.
- I unwrapped that key manually with `ProtectedData.Unprotect(..., CurrentUser)`: `prefix=DPAPI blobLen=283`, unprotect OK, 32-byte key. So DPAPI works and the current key is intact — it is simply a *different* key than the one used at encrypt time.
- Timestamps confirm it. `connection-catalog.json` was written **2026-07-11 15:22**. The entire `%APPDATA%\t3code` profile — `Local State`, `.updaterId`, `Cache`, `GPUCache`, `Code Cache`, `Local Storage` — is stamped **2026-08-08 00:17**. The userData dir was recreated a month after the catalog was encrypted, generating a new random key and orphaning the file.

The underlying design issue: the catalog lives in `~/.t3/userdata` but its key lives in `%APPDATA%\t3code`. Those two directories have independent lifecycles, so wiping either one alone leaves the app in a state it cannot recover from.

### Suggested fixes

1. Make `get` recover: on `decrypt-catalog` failure, fall back to an empty catalog, surface a non-fatal warning ("saved connections could not be decrypted and were reset"), and re-key on next write. Cheapest fix and it turns a hard block into a soft reset.
2. Optionally move the catalog into the Electron `userData` dir alongside `Local State`, or add a plaintext key-fingerprint field so a key mismatch is detectable and self-healing rather than an exception.
3. Reclassify — this surfaces as `ConnectionTransientError`, which is misleading for a permanently unrecoverable state.

### Impact

Blocks work completely

### Version or commit

0.0.34 (Alpha)

### Environment

Windows 11 Pro build 26200 (10.0.26200), WSL2 backend (Ubuntu 24.04.4 LTS, kernel 6.18.33.2-microsoft-standard-WSL2)

### Logs or stack traces

```shell
DesktopConnectionCatalogStoreProtectionError: Desktop connection catalog protection failed during decrypt-catalog at C:\Users\<user>\.t3\userdata\connection-catalog.json.
    at resources\app.asar\apps\desktop\dist-electron\main.cjs:15880:225
    at desktop.connectionCatalogStore.get (main.cjs:15886:25)
    at desktop.ipc.connectionCatalog.get (definition) (main.cjs:15905:25)
    at desktop.ipc.method (main.cjs:720:122)
    at desktop.ipc.invoke (main.cjs:669:68)
    at desktop.ipc.registerInvoke (main.cjs:27741:13)
    at desktop.ipc.installHandlers (main.cjs:28717:9)
    at desktop.bootstrap (main.cjs:28725:23) {
  [cause]: ElectronSafeStorageDecryptError: Electron safe storage failed to decrypt a string.
      at catch (main.cjs:1349:22) {
    [cause]: Error: Error while decrypting the ciphertext provided to safeStorage.decryptString.
        at try (main.cjs:1348:36)
  }
}
```

### Workaround

Quit the app, delete (or rename) `%USERPROFILE%\.t3\userdata\connection-catalog.json`, relaunch, and re-add the connection. Saved connection entries are lost; `clerk-tokens.json`, `secrets/`, and `state.sqlite` are unaffected.

### Related

- #4750 — `connection-catalog.json` non-atomic writes; a corrupt document bricks the app permanently. Same "unreadable catalog is fatal" failure mode, different cause (corruption vs. orphaned key). A recovery path in `get` would cover both.
- #4991 — identical user-facing `Could not list targets` message, but from `DesktopConnectionCatalogStoreMigrationError` / `read-legacy-secret` rather than `decrypt-catalog`.
- #5427 — `safeStorage` unavailable on Linux breaking the connection catalog; adjacent, but that is "encryption unavailable" rather than "key rotated".

yashranaway:
An unreadable encrypted desktop connection catalog currently prevents connection management from loading after a Safe Storage key change. Treat decryption failure as an empty catalog so the app can recover.

Keep the original encrypted file in place on read. Before a later replacement or clear, preserve its exact bytes in a unique connection-catalog.json.unreadable-* directory beside the catalog. This applies to automatic credential refreshes as well as user saves. If the backup fails, refuse replacement and retain the original. Later recovery incidents receive separate backups.

A direct store read can recover when decryption starts working again. The renderer caches its empty read and needs a reload to see that recovery; an intervening save writes a replacement while retaining the encrypted backup. This change does not add a recovery notice or an automatic backup-restore flow.

Validation: 10 focused desktop catalog tests pass. Two backup regressions fail before the fix. Coverage includes replacement after an empty read, repeated writes, unique backups, clear, and backup-failure retry. Rebased tests retain main's platform-aware paths. Scoped lint and desktop typecheck pass. Native macOS/Windows migration and the full WSL startup flow have not been exercised.

Closes #8341.

Model: GPT-6
Harness: Codex in T3 Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Unreadable encrypted connection catalogs are now handled gracefully instead of causing a read failure.
  * Original unreadable catalog files are preserved in a backup before being replaced or removed.
  * Clear errors are reported if the unreadable catalog cannot be backed up.

### DIFF
--- a/apps/desktop/src/app/DesktopConnectionCatalogStore.test.ts
+++ b/apps/desktop/src/app/DesktopConnectionCatalogStore.test.ts
@@ -385,7 +385,7 @@ describe("DesktopConnectionCatalogStore", () => {
     ),
   );
 
-  it.effect("surfaces a catalog that can no longer be decrypted without deleting it", () =>
+  it.effect("treats a catalog encrypted with an obsolete key as empty until it is replaced", () =>
     Effect.gen(function* () {
       const path = yield* Path.Path;
       const fileSystem = yield* FileSystem.FileSystem;
@@ -399,25 +399,22 @@ describe("DesktopConnectionCatalogStore", () => {
       );
 
       assert.isTrue(yield* store.set('{"schemaVersion":1,"targets":[]}'));
+      const catalogPath = path.join(baseDir, "userdata", "connection-catalog.json");
+      const originalDocument = yield* fileSystem.readFileString(catalogPath);
       yield* Ref.set(failDecrypt, true);
-      const error = yield* store.get.pipe(Effect.flip);
-      assert.instanceOf(
-        error,
-        DesktopConnectionCatalogStore.DesktopConnectionCatalogStoreProtectionError,
-      );
-      assert.equal(error.operation, "decrypt-catalog");
-      assert.equal(error.catalogPath, path.join(baseDir, "userdata", "connection-catalog.json"));
-      assert.instanceOf(error.cause, ElectronSafeStorage.ElectronSafeStorageDecryptError);
-      const decryptError = error.cause as ElectronSafeStorage.ElectronSafeStorageDecryptError;
-      assert.instanceOf(decryptError.cause, Error);
-      assert.equal(decryptError.cause.message, "invalid encrypted catalog");
-      assert.equal(
-        error.message,
-        `Desktop connection catalog protection failed during decrypt-catalog at ${path.join(baseDir, "userdata", "connection-catalog.json")}.`,
-      );
-      assert.notEqual(error.message, decryptError.message);
+      assert.deepStrictEqual(yield* store.get, Option.none());
+      assert.equal(yield* fileSystem.readFileString(catalogPath), originalDocument);
+
       yield* Ref.set(failDecrypt, false);
       assert.deepStrictEqual(yield* store.get, Option.some('{"schemaVersion":1,"targets":[]}'));
+
+      yield* Ref.set(failDecrypt, true);
+      assert.isTrue(yield* store.set('{"schemaVersion":1,"targets":["replacement"]}'));
+      yield* Ref.set(failDecrypt, false);
+      assert.deepStrictEqual(
+        yield* store.get,
+        Option.some('{"schemaVersion":1,"targets":["replacement"]}'),
+      );
     }).pipe(Effect.provide(NodeServices.layer), Effect.scoped),
   );
 });
--- a/apps/desktop/src/app/DesktopConnectionCatalogStore.test.ts
+++ b/apps/desktop/src/app/DesktopConnectionCatalogStore.test.ts
@@ -408,13 +408,99 @@ describe("DesktopConnectionCatalogStore", () => {
       yield* Ref.set(failDecrypt, false);
       assert.deepStrictEqual(yield* store.get, Option.some('{"schemaVersion":1,"targets":[]}'));
 
-      yield* Ref.set(failDecrypt, true);
+      // The renderer retains the empty read even if decryption recovers. A
+      // backend credential refresh can then save that cached catalog at startup.
       assert.isTrue(yield* store.set('{"schemaVersion":1,"targets":["replacement"]}'));
-      yield* Ref.set(failDecrypt, false);
       assert.deepStrictEqual(
         yield* store.get,
         Option.some('{"schemaVersion":1,"targets":["replacement"]}'),
       );
+      const backupDirectories = (yield* fileSystem.readDirectory(path.dirname(catalogPath))).filter(
+        (name) => name.startsWith("connection-catalog.json.unreadable-"),
+      );
+      assert.lengthOf(backupDirectories, 1);
+      const backupPath = path.join(
+        path.dirname(catalogPath),
+        backupDirectories[0]!,
+        "connection-catalog.json",
+      );
+      assert.equal(yield* fileSystem.readFileString(backupPath), originalDocument);
+      assert.isTrue(yield* store.set('{"schemaVersion":1,"targets":["second"]}'));
+      assert.equal(yield* fileSystem.readFileString(backupPath), originalDocument);
+      assert.lengthOf(
+        (yield* fileSystem.readDirectory(path.dirname(catalogPath))).filter((name) =>
+          name.startsWith("connection-catalog.json.unreadable-"),
+        ),
+        1,
+      );
+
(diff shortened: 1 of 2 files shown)


CDVolvik:
Verified fix(desktop): recover from obsolete catalog keys @09fa518.

On win32: pnpm exec vitest run apps/desktop/src/app/DesktopConnectionCatalogStore.test.ts — 4/9 pass, 5/9 fail on pre-existing Windows path assertions (/ vs \\ in catalogPath). Critical new test treats a catalog encrypted with an obsolete key as empty until it is replaced passes: get returns Option.none() while failDecrypt=true, file left intact, recovers after failDecrypt=false, and set re-encrypts replacement with current key.

Revert check vs main (2daff8c25): 3/9 pass → 6/9 fail (same 5 + old decrypt test). Net +1 on this branch.

Logic in DesktopConnectionCatalogStore.ts:478 correctly switches from throw DesktopConnectionCatalogStoreProtectionError to logWarning {catalogPath, errorTag} + Option.none(), no payload leak.

vp run --filter @t3tools/desktop typecheck clean (2 suggestions only).

LGTM — medium risk as noted, safe recovery. Verified as CDVolvik on win32, no push.

zaidbul:
Thanks for this fix. I hit the same failure on macOS after migrating another Mac's user data onto a machine with T3 Code already installed. The restored connection catalog predated the new T3 Code Safe Storage Keychain entry. GitHub authentication and the local backend were healthy. Quitting the app, preserving the catalog under a backup name, and relaunching restored local projects, providers, source control, and T3 Connect availability. This adds a macOS reproduction of #8341.

While checking an adaptation of this patch against main at `5d14c0e9`, I found a preservation detail relevant to the human-review concern. On that revision, `makeCatalogStore` caches the empty result. Startup reconciliation for a secondary desktop backend such as WSL calls `credentials.put`, which writes that cached catalog through `set`. The original encrypted file can therefore be replaced without a user explicitly saving a connection. I traced this in the source; I have not run that Windows scenario.

The direct store test also retries decryption after restoring the key, while the renderer keeps its empty cache until reload. That distinction matters because the synchronous decrypt error does not distinguish a permanently missing key from temporary access failure.

Would you be open to preserving a unique encrypted backup before the first replacement, with a regression test for that automatic-write path? I can contribute that change to this PR. It would make recovery reversible and give the human reviewer a stronger data-preservation guarantee. A short user-facing recovery notice would help explain missing saved connections too; I would coordinate that separately if it expands this patch.

Model: GPT-6. Harness: Codex desktop.

shivamhwp:

Please rebase the recovery test while retaining main's platform-aware `Path.join` paths. The direct store can read the original document again after decryption recovers, but the web catalog layer caches the empty read, so recovery in the current renderer still requires reloading. Keep that distinction clear when describing preservation and recovery; a subsequent catalog update replaces the unreadable document.

Issue #8701, opened 2026-08-30
[Bug]: Ghostty scrollback is configured as bytes despite being named as rows


zero-two-rafaeltab:
## Impact

T3 silently discards most of the terminal scrollback it appears to configure. The deterministic reproduction retains only **800 of the intended 10,000 rows** at 80 columns—a loss of more than 90%—and the effective count can be even lower at wider terminal sizes.

This means output from long-running agent commands, builds, test suites, logs, and stack traces can disappear from the terminal before users have a chance to inspect or copy it. There is no truncation marker or warning, so the terminal simply starts at a much later line when the user scrolls to the top. The loss happens inside the renderer and can occur even while the raw output remains below T3's separate 512 KiB client-buffer limit.

The affected renderer is used by web and desktop, and Android mirrors the same incorrect value in its native terminal bridge.

## Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

## Area

`apps/web` (also affects desktop because it uses the web terminal, and Android mirrors the same configuration in its native bridge)

## Problem statement

T3 configures the embedded Ghostty terminal with a value named `MAX_SCROLLBACK_ROWS = 10_000` on web and `kMaxScrollbackRows = 10000` on Android. The pinned `libghostty-vt` C header also describes `max_scrollback` as a maximum number of lines.

The pinned Ghostty implementation actually interprets that field as a byte budget. T3 is therefore requesting 10,000 bytes while its naming and the public header indicate an intended limit of 10,000 rows.

Ghostty does not retain a literal 10 KB buffer because it rounds/clamps the request to its internal page and viewport minimum. That masks the unit mismatch, but the resulting history is still far below 10,000 rows and varies with terminal width.

## Steps to reproduce

A focused failing test using T3's actual vendored Ghostty WASM is available here:

- [Failing test at an immutable commit](https://github.com/zero-two-rafaeltab/t3code/blob/2fefadf44d9b3b6ce76afb927ba12758c5faa91e/apps/web/src/terminal/ghostty/runtimeAbi.test.ts#L207-L246)
- [Reproduction branch](https://github.com/zero-two-rafaeltab/t3code/tree/repro/ghostty-scrollback-unit-mismatch)

Run it with:

```shell
pnpm --filter @t3tools/web exec vp test run --project unit src/terminal/ghostty/runtimeAbi.test.ts
```

[...]

PR #8766, opened 2026-08-30
fix(terminal): mitigate limited Ghostty scrollback

zero-two-rafaeltab:
> A separate PR will provide the durable fix for #8701 by upgrading libghostty-vt and configuring independent line and byte limits. This smaller hotfix is intended to remove the immediate user impact while that larger change is reviewed.

Refs #8701

## Summary

T3 currently passes `10,000` as Ghostty's maximum scrollback value, treating it as a row count. The currently pinned Ghostty ABI interprets that value as bytes, causing terminals to retain very little history.

This hotfix works around the issue without changing the Ghostty ABI:

- Renames the web and Android constants to accurately describe bytes.
- Uses a 32 MiB (`33,554,432` byte) per-terminal budget for T3's uncompressed embedded paths.
- Applies the same value to web and Android.
- Keeps direct vendored-WASM regression coverage and adds a production-constructor regression through `GhosttyTerminalCore`.
- Verifies that an 80-column terminal retains substantially more history and no longer collapses to Ghostty's minimum allocation.

This is intentionally not a complete fix for #8701. A byte budget cannot guarantee exactly 10,000 retained rows because row memory usage varies with terminal width and content. The durable fix requires the newer Ghostty ABI with independent byte and line limits.

The 32 MiB cap did not undercut the intended history in the verified regression scenarios while preserving a deliberate per-terminal memory bound. It does not turn the byte-based ABI into a line guarantee.

This PR does not:

- Upgrade libghostty-vt.
- Rebuild vendored Ghostty artifacts.
- Add client/server scrollback configurability.
- Change application protocols or persisted settings.

## Verification

- The direct vendored-WASM regression confirms that an 80-column terminal retains substantially more than Ghostty's minimum allocation.
- The production-constructor regression exercises the app's configured budget through `GhosttyTerminalCore.create`, `write`, and `scrollbarState`; reverting only the production value to 10,000 bytes makes it fail with 602 retained rows.
- Both focused Ghostty test files pass: 11 tests in 301 ms of test execution.
- Targeted formatting, linting, web typechecking, and the web build pass.
- Desktop build and smoke test pass.
- Android typechecking, x86_64 native compilation, and the Android app build pass.
- The Android native fingerprint includes the changed JNI source.

The following 12,000-line scenario was exercised in real web, desktop, and Android terminals:

```sh
printf '\033c'; cols=$(tput cols 2>/dev/null || printf '?'); i=1; while [ "$i" -le 12000 ]; do printf 'T3-SCROLLBACK-%05d\n' "$i"; i=$((i+1)); done; printf 'T3-END cols=%s\n' "$cols"
```

Observed oldest retained lines:

| Surface | Width | Oldest line |
|---|---:|---|
| Web | 129 columns | `T3-SCROLLBACK-00001` |
| Desktop | 106 columns | `T3-SCROLLBACK-00001` |
| Android | 80 columns | `T3-SCROLLBACK-00001` |

These results demonstrate improved retention, not an exact row-count guarantee.

Model: GPT-5.6 Codex  
Harness: T3 Code

### DIFF
--- a/apps/mobile/modules/t3-terminal/android/src/main/cpp/t3_terminal_jni.cpp
+++ b/apps/mobile/modules/t3-terminal/android/src/main/cpp/t3_terminal_jni.cpp
@@ -12,7 +12,7 @@ namespace {
 
 constexpr uint32_t kSnapshotMagic = 0x54563354;  // "T3VT" in little endian.
 constexpr uint16_t kSnapshotVersion = 1;
-constexpr size_t kMaxScrollbackRows = 10000;
+constexpr size_t kMaxScrollbackBytes = 32 * 1024 * 1024;
 
 enum CellFlag : uint16_t {
   kBold = 1 << 0,
@@ -205,7 +205,7 @@ Java_expo_modules_t3terminal_GhosttyBridge_nativeCreate(
   GhosttyTerminalOptions options = {
       .cols = static_cast<uint16_t>(std::clamp(cols, 1, 65535)),
       .rows = static_cast<uint16_t>(std::clamp(rows, 1, 65535)),
-      .max_scrollback = kMaxScrollbackRows,
+      .max_scrollback = kMaxScrollbackBytes,
   };
   if (ghostty_terminal_new(nullptr, &session->terminal, options) != GHOSTTY_SUCCESS ||
       ghostty_render_state_new(nullptr, &session->render_state) != GHOSTTY_SUCCESS ||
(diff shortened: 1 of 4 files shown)


rafaeltab:
PR made by my agent under my supervision. I am doing a manual check of this PR and the linked PR on web, desktop and android before publishing.

rafaeltab:
Manually verified that this is working on:
web (manjaro linux with zen browser)
desktop (manjaro linux)
android

shivamhwp:

The new ABI test writes `32 * 1024 * 1024` directly into its own options. Reverting the production core constant to 10,000 therefore leaves this test green. Please add the history-retention regression through `GhosttyTerminalCore.create` and `write`, so it exercises the budget the app actually configures. Keep the ABI check if it provides separate value. The distinction between a byte budget and a guaranteed line count should remain explicit.

PR #8767, opened 2026-08-30
fix(terminal): retain intended Ghostty scrollback

### LINKED ISSUE #8701: [Bug]: Ghostty scrollback is configured as bytes despite being named as rows

zero-two-rafaeltab:
> [#8766](https://github.com/pingdotgg/t3code/pull/8766) is a smaller byte-budget-only hotfix. It bypasses the unit mismatch rather than fixing it and does not provide the intended line-retention guarantee. This PR provides the durable fix by upgrading to Ghostty's explicit byte and line APIs.

Fixes #8701

## Summary

T3 was unintentionally inheriting libghostty-vt's 10,000-byte scrollback default, causing terminals to retain far fewer than the intended 10,000 lines.

This change:

- Pins libghostty-vt to reviewed upstream revision [`5a35415a`](https://github.com/ghostty-org/ghostty/commit/5a35415a5d59a117e654735ca5a01f876dec5841), which contains the explicit scrollback byte and line API.
- Migrates the web and Android constructors to the updated ABI.
- Configures each terminal with:
  - 10,000 physical scrollback lines.
  - An independent 32 MiB byte safety cap.
- Rebuilds the vendored web WASM and all four Android shared libraries.
- Makes Android derive its revision from the canonical `native/libghostty-vt/VERSION` pin.
- Updates the vendored headers, notices, rebuild documentation, and direct ABI assumptions.
- Removes an obsolete web cursor-blink reset workaround because the upgraded ABI now preserves the embedder default across terminal resets.

Ghostty prunes complete pages, so retained history stays within one internal page of 10,000; in this pinned revision, pruning can undershoot the configured line limit. Whichever limit is reached first wins.

## Why 32 MiB?

The embedded WebAssembly build cannot use Ghostty's scrollback compression. Android's 64-bit libraries support compression, but its 32-bit libraries do not, and T3 does not schedule compression through the C ABI.

The 32 MiB cap therefore treats scrollback as uncompressed on every surface. It is the smallest tested cap that did not become the first-reached limit before Ghostty's approximately 10,000-line boundary at representative terminal widths up to 320 columns, while still providing a deliberate per-terminal page-data bound.

## Verification

- Focused Ghostty ABI, core, renderer, surface, selection, input, mouse, keyboard, PTY-reply, and terminal-drawer tests pass.
- The production-constructor regression test retains 9,643 rows at 80 columns—within one 674-row internal page of 10,000—using the actual vendored WASM, and fails if the 10,000-byte default remains active.
- Targeted formatting, linting, web typechecking, and web/desktop builds pass.
- Android native static checks, Gradle compilation, and the Android app build pass.
- The vendored WASM reports the canonical Ghostty revision.
- Android headers and all four shared libraries were rebuilt from the same canonical revision.
- The 12,000-line PTY scenario was exercised successfully on web, desktop, and an Android emulator at narrow, normal, and wide dimensions.
- Controlled WASM measurements confirmed that the 32 MiB page-data cap does not become first-reached before the line target at representative widths; total per-terminal memory can exceed 32 MiB because visible-grid and non-page state are outside that cap.

No client/server scrollback configurability or protocol changes are included.

Model: GPT-5.6 Codex  
Harness: T3 Code

> [!NOTE]
> ### Configure Ghostty terminals to retain 10,000 scrollback lines and 32 MiB byte cap
> - Sets `SCROLLBACK_MAX_LINES` to 10,000 and `SCROLLBACK_MAX_BYTES` to 32 MiB on both web ([core.ts](apps/web/src/terminal/ghostty/core.ts)) and Android ([t3_terminal_jni.cpp](apps/mobile/modules/t3-terminal/android/src/main/cpp/t3_terminal_jni.cpp)) via the new `GHOSTTY_TERMINAL_OPT_SCROLLBACK_MAX_BYTES` and `GHOSTTY_TERMINAL_OPT_SCROLLBACK_MAX_LINES` options
> - Migrates to the updated upstream Ghostty API: `ghostty_terminal_new` now takes explicit `cols`/`rows` instead of a `GhosttyTerminalOptions` struct; options are applied via subsequent `ghostty_terminal_set` calls. All tests in [runtimeAbi.test.ts](apps/web/src/terminal/ghostty/runtimeAbi.test.ts) updated accordingly
> - Updates the vendored Ghostty revision in [VERSION](native/libghostty-vt/VERSION) to `5a35415a` and rebuilds native libraries for all architectures; build scripts now read the revision from the VERSION file and default to Zig 0.16.0
> - After `ghostty_terminal_reset`, the embedder default cursor blink is no longer reapplied — the terminal manages cursor blink state post-reset
> - Risk: `ghostty_terminal_new` signature change is breaking for any out-of-tree consumer; `ghostty_terminal_set` with `NULL` value now follows per-option documented semantics instead of blanket reset-to-default; `GHOSTTY_TERMINAL_OPT_KITTY_IMAGE_MEDIUM_TEMP_FILE` now takes a directory `GhosttyString` instead of a `bool`
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 98fb079.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Touches vendored native/WASM terminal core and a breaking Ghostty C ABI, but changes are scoped to scrollback configuration and constructor migration with targeted ABI tests.
> 
> **Overview**
> Fixes terminals retaining far fewer scrollback lines than intended because embedders were effectively hitting **libghostty-vt’s ~10,000-byte default** instead of configuring line limits.
> 
> **Pins** vendored `libghostty-vt` to `5a35415a`, rebuilds web WASM and all four Android shared libraries, and makes Android’s build script read the canonical `native/libghostty-vt/VERSION` (Zig **0.16.0**).
> 
> **Migrates** web (`core.ts`) and Android JNI to the new API: `ghostty_terminal_new(cols, rows)` plus `GHOSTTY_TERMINAL_OPT_SCROLLBACK_MAX_LINES` (**10,000**) and `SCROLLBACK_MAX_BYTES` (**32 MiB**), with whichever limit is reached first winning (page-granularity makes row counts approximate).
> 
> **Removes** the web workaround that reapplied default cursor blink after `ghostty_terminal_reset` / session replay; the upgraded Ghostty revision preserves embedder defaults across RIS.
> 
> **Adds** ABI/regression coverage (including a production-constructor test that writes ~11k lines against vendored WASM) and updates architecture/docs and third-party notices.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 98fb079fad6e0bcbde9277bc4e2190fd2bc13c8a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rafaeltab:
PR made by my agent under my supervision. I am doing a manual check of this PR and the linked PR on web, desktop and android before publishing.

rafaeltab:
Manually verified across:

web (on Manjaro linux with zen browser)
desktop (on Manjaro linux)
mobile (android on an S26 Ultra)

Haven't found anything non-functional while testing, and the scroll back goes way further now.

Recordings:

[recording web](https://github.com/user-attachments/assets/9f8a7d0c-4b36-4446-be64-9a42431a1ffc)
[recording desktop](https://github.com/user-attachments/assets/3e61a43a-af30-4751-bdd7-551af3d94751)
[recording android](https://github.com/user-attachments/assets/3c9f0b73-36b3-4a42-b435-defc3b49c5fa)

NOTE: desktop and android recordings were made by using the already run command (restored from t3 server), so that's why they only go back ~5000 lines. I have run them and checked manually that they go way further, but forgot to do so during the recording :see_no_evil:

juliusmarminge:

**Reason:** XXL unvouched libghostty-vt rebuild overlapping trusted upgrade #8563; hotfix is #8766
**Related:** #8563

rafaeltab:
@juliusmarminge the PR you linked does the upgrade needed for the fix to be possible, but doesn't actually fix the bug. Some options:

1. I could help on that PR to get the linked bug fixed
2. I could wait for that PR to be merged, and then fix the bug (would likely be around size S or M)
3. I could rebase this PR onto main, get it updated, tested, etc.

I'll try to do option 1 either way

rafaeltab:
Actually, I ran into several errors while trying their change locally.

I will update this PR, and reopen it to offer an alternative. I will keep targeting the same version I am now though, as I think it is better to keep the scope the same. Feel free to ask me to target a different version.

PR #4742, opened 2026-07-28
Add safe worktree inventory, pruning, and revival

juliusmarminge:
## What Changed

- Added server-side worktree inventory and safety classification.
- Added safe manual and bulk pruning with dirty and active-worktree safeguards.
- Added periodic auto-pruning with configurable inactivity and orphan policies.
- Added worktree revival before starting turns when a worktree is missing.
- Added thread-deletion orphan cleanup and the Worktrees settings page with filtering, status badges, policy controls, and destructive-delete confirmations.
- Added contracts, RPC support, integration coverage, and real Git-backed service tests.

## Why

Worktrees accumulated indefinitely, especially for web users and settled threads, because cleanup was limited to a desktop-only deletion flow. This change provides visibility and controlled cleanup while preserving branches and checkpoint references so pruned worktrees can be recreated safely when needed.

## UI Changes

Added a Worktrees settings page showing managed worktree status, linked threads, pruning eligibility, attention blockers, filters, bulk pruning, row-level deletion, and global auto-prune policies.

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [ ] I included before/after screenshots for any UI changes
- [ ] I included a video for animation/interaction changes

## Agent Transcript

<details>
<summary>Expand</summary>

### Assistant

Here's what the codebase actually has today, then a proposed feature set. The short version of the current state: there is almost no worktree management — which matches your 186-worktree situation.

**Current state (from a codebase sweep):**
- There's no worktree table or registry. A worktree exists only as a denormalized `worktree_path` string on `projection_threads` (`ProjectionThreads.ts:41`). Nothing records creation time, last use, or dirty state.
- The *only* cleanup path is the interactive "delete worktree too?" dialog when you delete the last thread using it (`useThreadActions.ts:253-397`) — and it's desktop-only (needs `localApi`), so web users always leak worktrees. Server-side `thread.delete` never touches the worktree.
- `git worktree prune` is never called anywhere; `git worktree list` is only used to annotate the branch list.
- Settling is purely a projection-level soft-hide (`settled_override`/`settled_at`, decider at `decider.ts:448-507`) — so as you said, settling means worktrees now live forever.
- Checkpoint refs (`refs/t3/checkpoints/...`) live in the shared ref store, not per-worktree, so deleting a worktree does **not** lose checkpoints — which is what makes revival viable.

## Proposed features

[...]

PR #8771, opened 2026-08-30
feat(server): clean worktree build artifacts when threads settle

IAmJSD:
> [!NOTE]
> This PR was written by Claude Code (Claude Fable 5) under human direction — the code, the tests, and this description included.

## What Changed

- Added an opt-in **Clean settled worktrees** server setting (default off) in Settings → General, next to the existing worktree options, with search/reset/modified-count support.
- Added a `ThreadSettleCleanupReactor` (modeled on `ThreadDeletionReactor`) that reacts to `thread.settled` domain events. When the settled thread runs in a linked worktree, it deletes regenerable build artifacts: `node_modules`, Cargo `target` (only when next to a `Cargo.toml`), `.next`, `.nuxt`, `.turbo`, `.svelte-kit`.
- Safety guards, in order: setting must be on; thread must have a live `worktreePath` not shared with another live thread; the path must be a *linked* worktree (`.git` pointer file — a primary checkout is never touched). The scan never enters `.git` or matched artifacts, never follows symlinks out of the worktree, and removal is best-effort per directory. Tracked files and uncommitted work are never at risk.
- Tests: real-tempdir coverage for the artifact scanner (depth, Cargo gating, symlink escape, `.git`-pointer detection), unit coverage for the shared-worktree guard, and the `OrchestrationReactor` start-order test updated.

Scope note: cleanup fires on explicit `thread.settled` events; client-derived auto-settle (inactivity / merged PR) emits no domain event and is deliberately out of scope.

## Why

Settled worktree threads keep their install and build caches on disk forever — a handful of parked JS/Rust threads easily holds gigabytes. This reclaims that disk the moment a thread is parked while keeping the worktree itself intact (branch, tracked files, uncommitted changes), so an unsettled thread simply resumes and the setup script regenerates caches on the next turn.

Complementary to worktree pruning proposals like #4742: those remove whole worktrees under retention policies; this keeps the worktree and strips only regenerable artifacts, triggered by settling.

## UI Changes

One standard settings row (title, description, `Switch`) in Settings → General. No screenshots included; happy to add them on request.

## Checklist

---

Written by Claude Code with Claude Fable 5 (`claude-fable-5`).

---

> [!NOTE]
> **Medium Risk**
> Opt-in filesystem deletes on worktrees, but gated by linked-worktree verification, git ignore/tracked-file checks, and shared-thread guards; primary checkouts are never touched.
> 
> **Overview**
> Adds an opt-in **Clean settled worktrees** setting (default off) and a `ThreadSettleCleanupReactor` that runs when a `thread.settled` domain event is emitted.
> 
> When enabled, the reactor removes regenerable artifacts (`node_modules`, framework caches, Cargo `target` next to `Cargo.toml`) only from **linked** git worktrees, after checks that the thread is still settled, not deleted, not sharing the path with another live thread, and that git treats each directory as ignored with no tracked files. Scanning is depth-bounded, skips `.git` and symlinks that escape the tree, and deletion is best-effort per directory.
> 
> New `worktreeArtifacts` helpers implement discovery, linked-worktree detection, and guarded removal. The reactor is started from `OrchestrationReactor` and registered in the server reactor layer; Settings → General gets a toggle plus search/reset/dirty tracking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12d955019ba909919b0ac373983ed066fecf8d07. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add `ThreadSettleCleanupReactor` to clean worktree build artifacts on thread settle
> - Introduces a new reactor that subscribes to `thread.settled` events and removes regenerable build artifact directories (e.g. `node_modules`, `.next`, `.turbo`) from linked git worktrees
> - Adds utilities in [worktreeArtifacts.ts](https://github.com/pingdotgg/t3code/pull/8771/files#diff-914cde1c99066a0a568118b9900e0a961ea8b9f624900d0d111f87810904e5fb): `isLinkedWorktreePath` detects linked worktrees, `findWorktreeArtifactDirectories` does a bounded-depth walk over an allowlist, and `removeWorktreeArtifacts` verifies candidates are git-ignored with no tracked files before recursive deletion
> - Adds `cleanWorktreeArtifactsOnSettle` boolean to `ServerSettings` (default `false`) and `ServerSettingsPatch`, with a toggle in the General settings UI and settings search entry
> - Wires the reactor into `OrchestrationReactor.start()` and the live server layer composition
> - Guards skip cleanup when the setting is off, the thread is deleted, another live thread shares the worktree, or the path is not a linked worktree
> - Risk: `removeWorktreeArtifacts` recursively force-deletes directories; a false positive in the git-ignore/track verification would delete uncommitted files. Reviewers should check `removeWorktreeArtifacts` in [worktreeArtifacts.ts](https://github.com/pingdotgg/t3code/pull/8771/files#diff-914cde1c99066a0a568118b9900e0a961ea8b9f624900d0d111f87810904e5fb) and the shared-worktree guard `isWorktreeSharedWithAnotherThread` in [ThreadSettleCleanupReactor.ts](https://github.com/pingdotgg/t3code/pull/8771/files#diff-09d7b78f4d3b1ee301beae84fb90e521f6524027885d763f5e07ce23c486a776)
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 12d9550.</sup>
> 
>


IAmJSD:
Addressed all three review findings in 35c7822 (fix written by Claude Code, as is this comment):

- **Effect service conventions** — collapsed the reactor into a single canonical module at `apps/server/src/orchestration/ThreadSettleCleanupReactor.ts` (tag with inline interface, `make`, `layer`); removed the `Services/` + `Layers/` split and the standalone `Shape` type, updated all importers to `ThreadSettleCleanupReactor["Service"]` member types.
- **Stale settle race** — the worker now re-checks `settledOverride === "settled"` on the freshly loaded projection row before cleaning. Events are projected inside the commit transaction before being published, so a thread unsettled after the settle event was queued no longer reads as settled by dequeue time.
- **`Cargo.toml` directory false positive** — the scanner now stats the manifest and requires a regular file before treating a sibling `target` directory as a Cargo build artifact, with a regression test.

IAmJSD:
Fixed the separate-git-dir finding in 5a0e290 (written by Claude Code, as is this comment): `isLinkedWorktreePath` now resolves the `.git` pointer's `gitdir:` target and requires its `commondir` file — present only in a linked worktree's private git dir, absent in a `--separate-git-dir` clone's full git dir (verified against real `git worktree add` / `git clone --separate-git-dir` layouts). Malformed pointers read as "not a worktree", so cleanup never guesses. Tests updated with faithful on-disk fixtures for all three cases.

The earlier Cursor finding ("cleanup runs after thread unsettles") was already addressed in 35c7822 via the `settledOverride === "settled"` re-check.

IAmJSD:
Addressed the test-coverage finding in 6144972 (written by Claude Code, as is this comment): added focused reactor tests that drive a real `thread.settled` event through `make` with stubbed engine/repository/settings layers and real on-disk fixtures, asserting cleanup fires only when the setting is enabled and every guard passes — and is skipped for a disabled setting, an unsettled or deleted thread, a shared worktree, and a primary checkout. The reactor now also mirrors `ThreadDeletionReactor`'s seen-sequence watermark and exposes `drainThrough` instead of the previously unused `drain`, so the tests wait on receipts rather than timing.

IAmJSD:
Addressed both criticals in c8bf5e8 (written by Claude Code, as is this comment):

- **Name-based deletion could hit tracked/uncommitted work** — `removeWorktreeArtifacts` now deletes a directory only after git verifies it: `git check-ignore` must report it ignored and `git ls-files` must show no tracked files inside. Anything unverified (including any failed git invocation) is skipped untouched and logged. Covered by new tests against real repos: verified artifacts removed, non-ignored `node_modules` skipped, non-git directories skipped entirely.
- **`.git` pointer borrowed from another worktree** — `isLinkedWorktreePath` now also resolves the private git dir's `gitdir` back-reference and requires it to canonically match this worktree's own `.git` pointer, so only the directory git actually registered passes. Tested with a real `git worktree add` checkout plus an impostor directory carrying a copied pointer.

On the **Medium watermark finding** (event committed between the `latestSequence` read and stream subscription): this mirrors the exact watermark pattern `ThreadDeletionReactor` ships on main, and the window is not reachable in practice — reactors start during the startup phase behind the command-readiness gate, so no `thread.settled` can commit before the subscription is live. Cleanup is also deliberately best-effort: a hypothetically missed event costs one uncleaned worktree until its next settle, not correctness. Happy to revisit if the shared pattern gets a common fix, but I'd rather not fork this reactor's subscription semantics from its sibling inside this PR.

IAmJSD:
Follow-up on the Medium watermark finding — rather than leaving it at the written dismissal, 12d9550 (written by Claude Code, as is this comment) removes the `latestSequence` head-noting entirely: `seenSequence` now advances only for events the subscription actually handed to the worker, so `drainThrough` can never claim coverage of an event committed between a head read and subscription start — the integrity gap the finding described no longer exists. Unlike `ThreadDeletionReactor`, this reactor's `drainThrough` has no production callers that could hang on pre-subscription sequences (it's a test receipt), so dropping the head-noting is safe here. The remaining "event committed before the subscription is live" window is gated by startup ordering: reactors start inside the startup phase behind the command-readiness gate, so no `thread.settled` can commit before the subscription exists.

PR #8795, opened 2026-08-30
fix(server): stop reading every provider binding on each session list

### LINKED ISSUE #8794: [Bug]: listSessions queries every historical thread on each call, stalling turn starts and connection setup
### Area

apps/server

### Steps to reproduce

1. Use one install for a few weeks so `provider_session_runtime` accumulates rows. Every thread that has ever run an agent keeps one row; nothing prunes them.
2. Start a turn in any thread.
3. Watch `server.trace.ndjson`, or connect a fresh client while the turn starts.

### Expected behavior

Starting a turn costs a bounded number of queries, and connecting a client is not affected by how many threads the install has accumulated.

### Actual behavior

`ProviderService.listSessions` asks the directory for every persisted thread id and then issues one binding lookup per thread:

https://github.com/pingdotgg/t3code/blob/2daff8c25/apps/server/src/provider/Layers/ProviderService.ts#L999-L1016

`directory.listThreadIds()` already reads the whole table (`repository.list()`), so the per-thread `getBinding` calls re-read rows that were just read. Only threads with a live adapter session are used afterwards, so every other read is discarded.

`listSessions` runs several times per turn start. Parent spans from one burst on my install:

```
1774x listSessions < resolveSessionRuntimeForThread < resolveCheckpointCwd < ensurePreTurnBaselineFromDomainTurnStart
 887x listSessions < ensureSessionForThread < processTurnStartRequested < processDomainEvent
 887x listSessions < processTurnStartRequested < processDomainEvent
 887x listSessions < resolveSessionRuntimeForThread < resolveCheckpointCwd < ensurePreTurnBaselineFromTurnStart
```

Measured from `server.trace.ndjson` on that install (886 rows in `provider_session_runtime`, 9 of them `running`):

| Window | sql.execute | Peak rate | p95 | max |
| Quiet | 378/s | 592/s | 7ms | 520ms |
| Turn start | 643/s over 19s | 2163/s | 1298ms | 1439ms |

4,439 statements in that 19s window were the same query, `SELECT ... FROM provider_session_runtime WHERE thread_id = ?`.

### Impact

The server saturates its own event loop on every turn start, so unrelated RPCs queue behind it. `server.getConfig` is one of them, and it gates connection setup, so clients hit the 15s establishment timeout and show "<environment> did not respond during connection setup", then reconnect and repeat. The 15s slow-request toast fires for the same reason. On a local `127.0.0.1` connection with no network involved.

The cost scales with the number of threads the install has ever had, so it gets worse over time.

### Version or commit

`main` at 2daff8c25 (also present in 0.0.35).

### Environment

macOS, desktop app hosting its own server, client connected over `127.0.0.1`.

### Logs or stack traces

```
uk: Failed to connect. Reconnecting...
uk did not respond during connection setup.
```

### Workaround

None available to a user. Archiving or deleting threads does not remove the rows: `ProviderSessionDirectory` has no delete method, and nothing calls the repository's `deleteByThreadId`.

willsheldon:
Fixes #8794.

## Problem

`ProviderService.listSessions` asked the session directory for every persisted thread id, then issued one binding lookup per thread. The provider runtime table keeps a row for every thread that has ever run an agent, so on an install with 886 rows every call ran 886 single-row queries at unbounded concurrency. `listSessions` runs several times per turn start, so a turn start issued roughly 3,500 queries and saturated the event loop. Unrelated RPCs queued behind it, including `server.getConfig`, which gates connection setup, so clients timed out and reconnected.

`directory.listThreadIds()` already reads the whole table, so the per-thread lookups re-read rows that were just read.

## Fix

Only threads with a live adapter session are used after the lookup, so look up those bindings instead of all of them. On the install above that is 9 reads instead of 886, and the cost no longer grows with thread history.

The added test seeds 25 stopped bindings, starts one session, and asserts that listing sessions reads only the live thread's row. It fails on `main` with 26 reads.

## Verification

- `vp test run apps/server/src/provider/Layers/ProviderService.test.ts` — 39 passed
- `tsgo --noEmit -p apps/server/tsconfig.json` — clean
- `vp lint` on the changed files — clean

---

> [!NOTE]
> **Medium Risk**
> Changes hot-path session listing behavior; scope is narrow and aligned with how bindings are actually consumed, with a targeted regression test.
> 
> **Overview**
> **`listSessions`** no longer loads persisted runtime bindings for every thread that ever ran an agent. It now collects thread IDs from **live adapter sessions** only and runs **`getBinding`** for those threads when decorating session metadata.
> 
> That cuts per-call DB work from scaling with full thread history to scaling with active sessions—important because **`listSessions`** runs multiple times per turn start and the old path could issue hundreds of concurrent single-row reads and stall the event loop.
> 
> A regression test seeds many **stopped** bindings plus one active session and asserts **`listSessions`** triggers a single runtime read for the live thread.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27585b01ab897c6a19ef2baf622f5156596f8e0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Stop reading persisted bindings for stopped sessions in `ProviderService.listSessions`
> - `listSessions` previously fetched a persisted binding for every threadId in the directory, including stopped sessions. It now derives `activeThreadIds` from currently active adapter sessions and loads bindings only for those.
> - Added an integration test in [ProviderService.test.ts](https://github.com/pingdotgg/t3code/pull/8795/files#diff-cbac8cf0b9c33ddbfff2137001d4b9ef6d645a0b205efcf6c714b12efab70f14) that seeds many stopped sessions, starts one live session, and asserts only the live threadId is read from the `ProviderSessionRuntimeRepository`.
> - Risk: stopped sessions no longer have their persisted bindings loaded during `listSessions`; any consumer relying on binding data for non-active threads will not receive it.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 27585b0.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/ProviderService.test.ts
+++ b/apps/server/src/provider/Layers/ProviderService.test.ts
@@ -531,6 +531,84 @@ it.effect(
     }).pipe(Effect.provide(NodeServices.layer)),
 );
 
+it.effect("ProviderServiceLive reads persisted bindings only for live sessions", () =>
+  Effect.gen(function* () {
+    const codex = makeFakeCodexAdapter();
+    const registry = makeAdapterRegistryMock({
+      [ProviderDriverKind.make("codex")]: codex.adapter,
+    });
+    const providerAdapterLayer = Layer.succeed(
+      ProviderAdapterRegistry.ProviderAdapterRegistry,
+      registry,
+    );
+    // Records every per-thread runtime read so the test can assert that listing
+    // sessions does not query threads without a live session.
+    const readThreadIds: Array<ThreadId> = [];
+    const runtimeRepositoryLayer = Layer.effect(
+      ProviderSessionRuntime.ProviderSessionRuntimeRepository,
+      Effect.gen(function* () {
+        const inner = yield* ProviderSessionRuntime.ProviderSessionRuntimeRepository;
+        return {
+          ...inner,
+          getByThreadId: (input: { readonly threadId: ThreadId }) => {
+            readThreadIds.push(input.threadId);
+            return inner.getByThreadId(input);
+          },
+        };
+      }),
+    ).pipe(
+      Layer.provide(ProviderSessionRuntime.layer.pipe(Layer.provide(SqlitePersistenceMemory))),
+    );
+    const directoryLayer = ProviderSessionDirectoryLive.pipe(Layer.provide(runtimeRepositoryLayer));
+    const providerLayer = makeProviderServiceLive().pipe(
+      Layer.provide(providerAdapterLayer),
+      Layer.provide(directoryLayer),
+      Layer.provide(defaultServerSettingsLayer),
+      Layer.provide(serverConfigTestLayer),
+      Layer.provide(AnalyticsService.layerTest),
+      Layer.provide(
+        Layer.succeed(
+          ProviderEventLoggers.ProviderEventLoggers,
+          ProviderEventLoggers.NoOpProviderEventLoggers,
+        ),
+      ),
+    );
+
+    yield* Effect.gen(function* () {
+      const provider = yield* ProviderService.ProviderService;
+      const directory = yield* ProviderSessionDirectory.ProviderSessionDirectory;
+      const activeThreadId = asThreadId("thread-live-binding");
+
+      // Threads that ran an agent in the past keep their runtime binding.
+      for (let index = 0; index < 25; index += 1) {
+        yield* directory.upsert({
+          threadId: asThreadId(`thread-stopped-binding-${index}`),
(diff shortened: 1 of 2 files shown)


t3dotgg:
> [!NOTE]

We are keeping the session-list performance fix in the open [PR #8909](https://github.com/pingdotgg/t3code/pull/8909) instead of maintaining two implementations. Both replace the full history scan with binding reads for distinct thread IDs from live adapter sessions. That removes the lookup for each inactive thread on every session-list call. The fix is not on `main` yet.

PR #8823, opened 2026-08-30
fix(contracts): decode CLI event origins

### LINKED ISSUE #8789: [Bug]: T3 writes origin.surface="cli" that its own event-store decoder rejects, making t3code unbootable
### What happened

`t3code.service` became unbootable and crash-looped into `start-limit-hit`. The cause is a row in `orchestration_events` that T3 itself wrote, carrying an `origin.surface` value its own decoder rejects:

```
PersistenceDecodeError: Decode error in OrchestrationEventStore.readFromSequence:decodeRows
  Composite(Pointer(Composite(Pointer(Composite(Pointer(AnyOf(Composite(Pointer(AnyOf(AnyOf()))))))))))
  [cause]: SchemaError: Expected "web" | "desktop" | "mobile"
    at [0]["metadata"]["origin"]["surface"]
```

The offending row:

```
sequence:       43854
event_type:     project.meta-updated
actor_kind:     client
metadata_json:  {"origin":{"surface":"cli"}}
```

`"cli"` is not in the `web | desktop | mobile` enum the reader accepts, so `readFromSequence` throws during startup and the process exits 1. systemd retries, hits `StartLimitBurst`, and the service is down until someone intervenes.

Distribution across ~43.8k events on this machine — one bad row is enough:

| `origin.surface` | rows |
|---|---|
| absent | 43,588 |
| `desktop` | 263 |
| **`cli`** | **1** |

Note that `absent` is overwhelmingly the norm and decodes fine, so the field is effectively optional; it is only the out-of-enum *value* that is fatal.

### Why it is nasty

The event store is decoded **only at startup**. A poisoned row is completely harmless while the process is running, so:

- The service stays green indefinitely — it had been up for days.
- Health checks pass. Nothing surfaces a problem.
- The failure lands on the *next* restart, from any cause: an auto-update, a reboot, a crash recovery.

So the blast radius is "the box does not come back", detonating at an arbitrary later time with no connection to whatever wrote the row. For a headless machine reached through T3 Connect, that means losing remote access entirely — the relay stays up and answers, but there is no backend behind it, surfacing to clients as `Relay could not reach the environment endpoint (endpoint_request_failed)`.

It recurred: I cleared the first bad row, the service ran fine for about an hour, then a **second** `project.meta-updated` / `origin.surface="cli"` row appeared and the next restart crash-looped again.

### Reproduction

I could not identify the writer, which is the main thing I would want help with.

Ruled out by direct test: t3 CLI reads. Running `t3 connect status` did **not** produce a bad row (count stayed 0 across the invocation).

What the two poisoned rows had in common:
- `event_type = project.meta-updated`
- `actor_kind = client`
- `occurred_at` with exact `.000Z` millisecond precision, unlike surrounding rows, which suggests a different client/serialiser than the local service.

To check whether an install is affected:

```sql
SELECT json_extract(metadata_json,'$.origin.surface') AS surface, COUNT(*)
FROM orchestration_events
GROUP BY surface ORDER BY 2 DESC;
```

### Suggested fix

The asymmetry is the bug: something in the write path emits `"cli"` while the read schema does not accept it. Either end would fix this instance, but the third point is what stops the whole class:

1. Add `cli` to the `origin.surface` enum, if CLI-origin events are legitimate.
2. Otherwise stop the writer emitting it.
3. **Make `readFromSequence` resilient to undecodable rows.** A single malformed event in an append-only log should not be able to render the service unbootable. Skipping, quarantining or lenient-decoding a bad row with a loud warning would turn a hard brick into a logged anomaly.

This is the same shape as #4518 (persisted value rejected by the startup decoder, backend crash-loops). That was a settings value; this one is an event row, so it can be introduced by normal operation rather than by a user changing a setting.

### Workaround

For anyone hitting this — clear the out-of-enum value and reset the latched start limit:

```sql
UPDATE orchestration_events
SET metadata_json = json_remove(metadata_json, '$.origin.surface')
WHERE json_extract(metadata_json, '$.origin.surface') IS NOT NULL
  AND json_extract(metadata_json, '$.origin.surface') NOT IN ('web','desktop','mobile');
```

```bash
systemctl --user reset-failed t3code.service   # the start limit latches and blocks a plain restart
systemctl --user start t3code.service
```

Because it recurs, I ended up guarding the start path so the box can self-recover:

```ini
ExecStartPre=-%h/.local/bin/t3-sanitize-event-store
```

### Environment

- t3 `0.0.37-nightly.20260829.1219` (nightly channel)
- Linux (Arch), headless server, node 24.19.0
- Run as a `systemd --user` service with `Restart=always`, `StartLimitBurst=5`

Michel-Liao:
## What Changed

- Accept `"cli"` as an origin surface when decoding persisted orchestration-event metadata.
- Keep live client connection/auth surfaces restricted to `web | desktop | mobile`.
- Add a regression for the reported `project.meta-updated` event shape.

## Why

A T3-written event with `metadata.origin.surface: "cli"` currently fails startup decoding and can leave a server crash-looping. Keeping the compatibility value at the persistence boundary fixes existing event stores without widening live client surface contracts.

Fixes #8789.

## Verification

- `vp test run packages/contracts/src/orchestration.test.ts` — 50 passed
- Contracts typecheck
- Server typecheck
- Focused lint and formatting checks

## Checklist

- [x] No UI changes

---

> [!NOTE]
> **Low Risk**
> Narrow decode-time backward compatibility at the event-store boundary; no change to live client surface contracts or command dispatch.
> 
> **Overview**
> Fixes startup failures when replaying orchestration events whose metadata includes **`origin.surface: "cli"`** (e.g. CLI-authored `project.meta-updated`), which previously failed schema decode and could crash-loop the server.
> 
> **`OrchestrationEventMetadata.origin`** now decodes via a persistence-only **`PersistedOrchestrationOrigin`** that accepts **`cli`** in addition to **`ClientSurface`** (`web` / `desktop` / `mobile`). **`OrchestrationClientOrigin`** is unchanged for live client stamping and auth surfaces.
> 
> Adds a contract test that **`decodeOrchestrationEvent`** accepts a representative CLI-origin event shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb482d2507047def81d42f0c379b372bf400253f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Accept `origin.surface = 'cli'` in `OrchestrationEventMetadata` decoding
> Introduces `PersistedOrchestrationOrigin` schema in [orchestration.ts](https://github.com/pingdotgg/t3code/pull/8823/files#diff-eb3f1de358c2fd7cec5950215e475b915ad3dca9e4f3f7305e46c3277630d2af) that allows `surface` to be either `ClientSurface` or the literal `"cli"`, with an optional `appVersion` string. Changes the `origin` field of `OrchestrationEventMetadata` to use this new schema for historical event store compatibility. Adds a test confirming persisted events with CLI origin metadata decode correctly. Risk: consumers that relied on `origin` strictly conforming to `OrchestrationClientOrigin` may need to handle the additional `"cli"` surface value.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cb482d2.</sup>
> 
>

### DIFF
--- a/packages/contracts/src/orchestration.test.ts
+++ b/packages/contracts/src/orchestration.test.ts
@@ -572,6 +572,30 @@ it.effect("defaults settled fields when decoding historical thread data", () =>
   }),
 );
 
+it.effect("decodes persisted orchestration events with CLI origin metadata", () =>
+  Effect.gen(function* () {
+    const event = yield* decodeOrchestrationEvent({
+      sequence: 43_854,
+      eventId: "event-cli-origin-1",
+      aggregateKind: "project",
+      aggregateId: "project-1",
+      type: "project.meta-updated",
+      occurredAt: "2026-01-01T00:00:00.000Z",
+      commandId: "cmd-cli-origin-1",
+      causationEventId: null,
+      correlationId: "cmd-cli-origin-1",
+      metadata: { origin: { surface: "cli" } },
+      payload: {
+        projectId: "project-1",
+        title: "Updated from CLI",
+        updatedAt: "2026-01-01T00:00:00.000Z",
+      },
+    });
+
+    assert.deepStrictEqual(event.metadata.origin, { surface: "cli" });
+  }),
+);
+
 it.effect("decodes thread archived and unarchived events", () =>
   Effect.gen(function* () {
     const archived = yield* decodeOrchestrationEvent({
--- a/packages/contracts/src/orchestration.ts
+++ b/packages/contracts/src/orchestration.ts
@@ -1401,13 +1401,21 @@ export const OrchestrationClientOrigin = Schema.Struct({
 });
 export type OrchestrationClientOrigin = typeof OrchestrationClientOrigin.Type;
 
+// Historical event stores can contain CLI-authored project metadata events.
+// Keep that compatibility at the persistence boundary: ClientSurface still
+// describes only web, desktop, and mobile connection/auth surfaces.
+const PersistedOrchestrationOrigin = Schema.Struct({
+  surface: Schema.optional(Schema.Union([ClientSurface, Schema.Literal("cli")])),
+  appVersion: Schema.optional(TrimmedNonEmptyString),
+});
+
 export const OrchestrationEventMetadata = Schema.Struct({
   providerTurnId: Schema.optional(TrimmedNonEmptyString),
   providerItemId: Schema.optional(ProviderItemId),
   adapterKey: Schema.optional(TrimmedNonEmptyString),
   requestId: Schema.optional(ApprovalRequestId),
   ingestedAt: Schema.optional(IsoDateTime),
-  origin: Schema.optional(OrchestrationClientOrigin),
+  origin: Schema.optional(PersistedOrchestrationOrigin),
 });
 export type OrchestrationEventMetadata = typeof OrchestrationEventMetadata.Type;
 


juliusmarminge:
#8905

PR #8837, opened 2026-08-31
fix(clients): inherit model preferences for unconfigured projects

### LINKED ISSUE #8829: [Bug]: Automatically seeded project defaults override user model preferences
## Summary

T3 Code creates projects with a concrete model selection even when the user never configured a project override. Fresh drafts then treat that automatically seeded selection as an explicit project default, so it overrides both the app-wide sticky selection and the Codex user configuration. With GPT-5.6-Sol, a project record that stores the model but omits reasoning is resolved to the catalog default of Low and T3 sends Low explicitly to Codex.

## Steps to reproduce

1. Configure Codex with `model = "gpt-5.6-sol"` and `model_reasoning_effort = "high"` in `~/.codex/config.toml`.
2. In T3 Code, select GPT-5.6-Sol with High reasoning in the composer so the app-wide sticky selection is also Sol/High.
3. Add a folder as a T3 project without manually configuring its project-level New threads model, or use an existing project whose automatically created `default_model_selection_json` contains Sol but no `reasoningEffort` option.
4. Open a fresh draft in that project.
5. Observe the model and reasoning shown by the composer, then send a turn.

## Expected behavior

An automatically created project record must not behave like a user-configured project override. If the user has not chosen a project-specific model, a fresh draft should inherit the app-wide sticky model and complete option snapshot. If T3 has no explicit model-option choice to apply, it should preserve that absence rather than turning the provider catalog default into an explicit `turn/start` override. A project default deliberately selected by the user should continue to take precedence.

The regression is fixed when a project created without a user-selected override starts Sol/High in this reproduction, while a project explicitly configured as Sol/Low still starts Sol/Low. Web, desktop, and mobile should apply the same distinction.

## Actual behavior

T3 persists a concrete Sol selection during project creation. The new-thread flow applies the app-wide sticky Sol/High state first, then overwrites it with the project's automatically seeded Sol selection. Because that project selection omits `reasoningEffort`, the composer resolves Sol's advertised catalog default of Low, materializes Low into the dispatch options, and the Codex adapter sends `effort: "low"`. Codex therefore cannot apply the user's configured High default.

## Affected area

- `apps/server/src/serverRuntimeStartup.ts` creates auto-bootstrapped projects with a concrete Codex default selection.
- `apps/web/src/components/CommandPalette.tsx` creates added projects with a resolved concrete provider/model selection.
- `apps/web/src/hooks/useHandleNewThread.ts` gives any non-null project default precedence over sticky and carried state.
- `apps/web/src/components/chat/composerProviderState.tsx` materializes model capability defaults into dispatch options.
- `apps/server/src/provider/Layers/CodexAdapter.ts` forwards the materialized reasoning effort to Codex.
- The equivalent mobile new-task path should preserve the same precedence and provenance rules.

## Runtime or environment

- macOS desktop app
- T3 Code `0.0.37-nightly.20260830.1232`
- Codex provider using `gpt-5.6-sol`
- Codex user configuration selects High reasoning

## Evidence

A read-only inspection of the live T3 database found nine active projects. Six stored `codex / gpt-5.6-sol` with no reasoning option, and their `updated_at` values matched their creation times. Three stored Sol with High explicitly. The client local-storage record simultaneously contained an app-wide sticky Sol/High selection. The database passed `PRAGMA quick_check`.

The current source defines the automatic bootstrap selection as Codex plus `DEFAULT_MODEL`, stores it on project creation, and documents the fresh-draft order as project default above global sticky state. When no reasoning selection exists, composer capability resolution supplies the model descriptor's current or default value and builds it into the model options sent to the server. The Codex adapter forwards that value as the turn effort.

The behavior became reliable after #6011 fixed the opposite failure, where a sticky selection could beat an explicitly configured project default. That change is correct for user-configured project defaults, but T3 currently has no stored distinction between an automatic seed and a user-configured override.

## Impact

Users can start turns at a lower reasoning level than both their Codex configuration and their last explicit T3 selection. The composer looks authoritative, so the only dependable workaround is to configure every project separately and recheck the reasoning level before sending. The mismatch can change answer quality, latency, service-tier behavior, and usage without a clear indication of which setting won.

## Additional context

Related reports include #6539 for a global new-thread model setting, #6508 for per-project last-used working-mode behavior, and #5796 for the opposite precedence bug fixed by #6011. This report isolates the missing distinction between an automatically seeded project selection and a project override that the user actually chose.

coygeek:
New projects currently persist an automatically chosen provider and model as a project override. Fresh drafts then replace the app-wide sticky selection, and missing options such as reasoning effort become explicit catalog defaults sent to the provider.

This change treats project creation as unconfigured. Only a later `project.meta.update` write represents a deliberate project default. The server bootstrap, web command palette, CLI, and shared mobile creation path now converge on that rule at the domain boundary. Web and mobile still use capability defaults for display, but dispatch only options that T3 actually stored. Explicit unsent-draft choices and explicit project defaults keep their existing precedence.

Migration 44 rewrites create-time seeds only when the project event stream contains no metadata update that wrote `defaultModelSelection`. It never compares model slugs or option values, so an explicit same-value selection and an explicit reset both survive. Projection rows without corresponding `project.created` evidence are intentionally left unchanged because their provenance is ambiguous.

Closes #8829

## Validation

- `vp test run` across 22 focused files covering project creation, event decisions, projections, old and new migrations, web and mobile precedence, persisted drafts, grouped projects, provider availability, and Codex option forwarding: 405 tests passed.
- Package-scoped type checks passed for `@t3tools/contracts`, `@t3tools/shared`, `@t3tools/client-runtime`, `t3`, `@t3tools/web`, and `@t3tools/mobile`.
- Targeted lint and formatting checks passed for all changed files.
- `git diff --check` passed.
- No browser pass was needed because this change has no visible UI changes.

## AutoReview

AutoReview loop 1 completed clean with no accepted or actionable P0/P1 findings using Codex, `gpt-5.6-sol`, High reasoning, and Fast service tier.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 117bb162646e6dab0b61fb8f0228901729085285. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Built with GPT-5.6-Sol using High reasoning and Fast service tier through the Codex harness in T3 Code.

> [!NOTE]
> ### Stop auto-seeding `defaultModelSelection` on `project.create` and dispatch only explicit model options
> - Server now ignores `defaultModelSelection` on `project.create` commands and always records `null` in the created event, so new projects start unconfigured and inherit model preferences at runtime
> - All clients (web, mobile, CLI) stop sending `defaultModelSelection` during project creation; the web and mobile composers switch to `buildExplicitProviderOptionSelectionsFromDescriptors` so only user-chosen options are dispatched while catalog defaults remain for UI display only
> - Server startup (`resolveAutoBootstrapWelcomeTargets`) no longer sets `defaultModelSelection` on `project.create`; instead it resolves the model selection for the welcome `thread.create` from the project default or canonical default
> - Migration `044_ClearAutomaticProjectModelDefaults` clears `default_model_selection_json` in `projection_projects` and sets `payload_json.defaultModelSelection` to null on matching `project.created` events for projects that received automatic seeds with no subsequent explicit `project.meta.update`
> - Risk: migration 044 uses a `NOT EXISTS` subquery on `project.meta.updated` events to distinguish auto-seeded from explicitly configured projects; projects where an explicit update was never recorded will lose their `default_model_selection_json` and fall back to canonical defaults at runtime
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 117bb16.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/lib/modelOptions.test.ts
+++ b/apps/mobile/src/lib/modelOptions.test.ts
@@ -102,7 +102,7 @@ describe("mobile model options", () => {
     ]);
   });
 
-  it("normalizes a legacy fallback selection against current capabilities", () => {
+  it("does not materialize catalog defaults for missing stored options", () => {
     const config = {
       providers: [
         {
@@ -144,7 +144,14 @@ describe("mobile model options", () => {
     });
 
     expect(option?.capabilities?.optionDescriptors?.[0]?.id).toBe("serviceTier");
-    expect(option?.selection.options).toEqual([{ id: "serviceTier", value: "default" }]);
+    expect(option?.selection.options).toBeUndefined();
+
+    const [explicitOption] = buildModelOptions(config, {
+      instanceId: ProviderInstanceId.make("codex"),
+      model: "gpt-test",
+      options: [{ id: "serviceTier", value: "priority" }],
+    });
+    expect(explicitOption?.selection.options).toEqual([{ id: "serviceTier", value: "priority" }]);
   });
 
   it("rejects stored selections whose provider is not usable", () => {
--- a/apps/mobile/src/lib/modelOptions.ts
+++ b/apps/mobile/src/lib/modelOptions.ts
@@ -4,7 +4,7 @@ import type {
   ServerConfig as T3ServerConfig,
 } from "@t3tools/contracts";
 import {
-  buildProviderOptionSelectionsFromDescriptors,
+  buildExplicitProviderOptionSelectionsFromDescriptors,
   getProviderOptionDescriptors,
 } from "@t3tools/shared/model";
 
@@ -45,11 +45,12 @@ function normalizeSelectionOptions(
   if (!capabilities) {
     return selection;
   }
-  const options = buildProviderOptionSelectionsFromDescriptors(
+  const options = buildExplicitProviderOptionSelectionsFromDescriptors(
     getProviderOptionDescriptors({
       caps: capabilities,
       selections: selection.options,
     }),
+    selection.options,
   );
   return options
     ? { ...selection, options }
--- a/apps/server/src/cli/project.ts
+++ b/apps/server/src/cli/project.ts
@@ -30,7 +30,6 @@ import * as ProjectionSnapshotQuery from "../orchestration/Services/ProjectionSn
 import { OrchestrationLayerLive } from "../orchestration/runtimeLayer.ts";
 import { layerConfig as SqlitePersistenceLayerLive } from "../persistence/Layers/Sqlite.ts";
 import * as RepositoryIdentityResolver from "../project/RepositoryIdentityResolver.ts";
-import * as ServerRuntimeStartup from "../serverRuntimeStartup.ts";
 import {
   clearPersistedServerRuntimeState,
   readPersistedServerRuntimeState,
@@ -481,7 +480,6 @@ const projectAddCommand = Command.make("add", {
           projectId,
           title,
           workspaceRoot,
-          defaultModelSelection: ServerRuntimeStartup.getAutoBootstrapDefaultModelSelection(),
           createdAt: DateTime.formatIso(yield* DateTime.now),
         });
         return `Added project ${projectId} (${title}) at ${workspaceRoot}.`;
--- a/apps/server/src/orchestration/decider.projectModelSelection.test.ts
+++ b/apps/server/src/orchestration/decider.projectModelSelection.test.ts
@@ -1,0 +1,76 @@
+import * as NodeServices from "@effect/platform-node/NodeServices";
+import {
+  CommandId,
+  OrchestrationEvent,
+  ProjectId,
+  ProviderInstanceId,
+  type ModelSelection,
+} from "@t3tools/contracts";
+import { expect, it } from "@effect/vitest";
+import * as Effect from "effect/Effect";
+import * as Schema from "effect/Schema";
+
+import { decideOrchestrationCommand } from "./decider.ts";
+import { createEmptyReadModel, projectEvent } from "./projector.ts";
+
+const createdAt = "2026-08-30T00:00:00.000Z";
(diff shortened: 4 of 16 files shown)


coygeek:
### Follow-up field verification

Tested against T3 Code Nightly `0.0.37-nightly.20260830.1232`.

A read-only inspection of the live local state found 9 active projects. Six had a create-time `codex / gpt-5.6-sol` selection with no reasoning option and no later model-default metadata write. Three already had an explicit Sol/High project default.

I used the existing Project Settings UI to apply Sol/High to the six affected projects as a temporary workaround. Each change produced a `project.meta-updated` event containing the complete explicit selection, and the matching project projection updated immediately. A fresh draft then opened with GPT-5.6-Sol and High reasoning. The database still passed `PRAGMA quick_check`.

This was performed on the current unpatched nightly. It verifies the reported failure, the explicit-project-default workaround, and the persisted provenance boundary. It is not an integrated run of this PR's patched desktop build.

The result supports the distinction used by this PR:

- An automatic create-time seed has no later `project.meta-updated` event containing `defaultModelSelection`.
- A deliberate model choice or reset produces that metadata event.
- Migration 44 can therefore clear the automatic seed without comparing model slugs or option values, while preserving an explicit same-value choice or reset.

The six projects changed during this verification are now deliberate project overrides and will remain pinned to Sol/High after migration. Resetting those project defaults after installing a build with this fix will return them to inheritance behavior.

t3dotgg:
> [!NOTE]

Closing this because [#9164](https://github.com/pingdotgg/t3code/pull/9164) has merged. It covers creation-time defaults, migration 44, and dispatches only explicit options on web and mobile, so unconfigured projects inherit current preferences.

PR #8845, opened 2026-08-31
fix(server): bound DPoP replay state

### LINKED ISSUE #9040: [Bug]: DPoP replay markers grow without bound
## Before submitting

## Area

apps/server

## Steps to reproduce

1. Start a server environment that accepts DPoP-authenticated requests.
2. Send valid DPoP requests with unique `jti` values.
3. Inspect the configured secrets directory after each request.
4. Separately, submit a token exchange with a valid DPoP proof and an invalid bootstrap credential.

## Expected behavior

Replay state should remain atomic while being retained only for the proof acceptance window. A token exchange should validate its bootstrap credential before writing replay state.

## Actual behavior

Each accepted proof creates a permanent `dpop-proof-<digest>.bin` file in the general secrets directory. Normal authenticated traffic adds one file per request, and no cleanup removes them. The token exchange path also records a valid proof before rejecting an invalid bootstrap credential, allowing file creation and `fsync` work without a usable credential.

## Impact

Major degradation over long-running or high-traffic server lifetimes due to unbounded file and inode growth. The finding affects availability and reliability; no credential disclosure was identified.

## Version or commit

`main` at `7963ac740`

## Investigation

Daybreak Blue confirmed the persistent allocation path and found that replay markers could also be lost across a partial bucket-boundary claim if writes were ordered incorrectly. Fable independently confirmed that normal DPoP traffic reproduces the unbounded growth and recommended a focused reliability fix with bounded, dedicated replay storage and credential preflight.

Proposed fix: #8845

solomonneas:
Fixes #9040

DPoP authentication previously created a permanent secrets-store file for every accepted proof. This change records empty, digest-named markers in overlapping ten-minute buckets and prunes only expired buckets. Claims write and sync the future bucket before the current bucket so a partial write still blocks replay across a boundary.

Bootstrap credentials are validated without consuming them before replay state is allocated. The final atomic consume still decides whether the exchange succeeds. Replay rejection during token exchange includes the DPoP challenge header.

**Upgrade requirement:** stop all older servers sharing the secrets directory and drain their requests before starting the new version. Mixed-version operation is unsupported. Existing canonical legacy markers are always checked and never created or deleted by the new store. New traffic uses bounded bucket storage, while existing legacy files remain as fixed migration residue. Rollback requires stopping all new-version writers and waiting at least six minutes before starting an older server. See the [migration guidance](https://github.com/solomonneas/t3code/blob/9eaf1d002/docs/internals/environment-auth.md#replay-store-upgrades-require-stopped-older-writers).

Verification:

- Focused replay, bootstrap, environment-authentication, and server tests: 219 passed across six files.
- Server typecheck and targeted lint/format checks: passed.
- New migration tests reproduced five failures before the fix, including late legacy replay, marker deletion, and unexpected lookup errors.
- Existing bucket-boundary and partial-write regression tests still pass. Independent-store concurrency permits only one successful claim.

Model and harness: GPT-6 via Codex.

## Summary by CodeRabbit

- **New Features**
  - Added stronger DPoP replay protection using time-based tracking.
  - Bootstrap credentials can now be validated without being consumed.
  - Invalid credential attempts no longer consume a DPoP proof’s replay state.

- **Bug Fixes**
  - Replayed DPoP proofs are rejected consistently, including across time boundaries.
  - Rejected DPoP requests now include the appropriate authentication challenge.

- **Documentation**
  - Added upgrade and rollback guidance for the new replay-protection behavior, including server coordination and clock requirements.


shivamhwp:

The compatibility lifetime is still tied to this instance's startup rather than the lifetime of proofs accepted by every writer sharing `secretsDir`.

- Start the new store at t=0. At t=359s, an old server accepts a fresh proof and atomically creates its legacy marker. At t=361s, the new store skips `claimLegacyMarker` and accepts the same thumbprint/jti. That proof is only two seconds old, well inside the five-minute acceptance window.
- Start another new-version store at t=420s and let it claim a proof. Its six-minute compatibility period writes a fresh legacy marker, but calling `prune` on the first store immediately deletes that marker. An old server can then accept the same proof during the newer store's advertised compatibility window.

`pruneLegacy` cannot erase every canonical legacy marker merely because one process is six minutes old, and `claim` cannot stop consulting that namespace while older writers may still accept proofs. This needs a coordinated migration boundary covering all writers, or an explicit requirement to stop old servers before transitioning. A longer process-local timer leaves the same failure at a later cutoff. Avoid bringing back unprotected stat-then-delete cleanup of reusable canonical marker names.

Issue #8293, opened 2026-08-26
[Bug]: Grok stop_reason "error" is normalized to end_turn — provider API errors are dropped and the turn is recorded as successful


exbaitman:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

### Steps to reproduce

1. Configure the Grok provider (`grok agent stdio`) and start a thread.
2. Put the account into any state where xAI returns an API error for a prompt. The
   easiest naturally-occurring one is an exhausted Grok Build balance (HTTP 402), but
   any condition that makes Grok end the turn with `stop_reason: "error"` reproduces it.
3. Send a message.
4. Watch the thread.

Deterministic repro without needing a real account state: a mock ACP process that
accepts `session/prompt` and then sends

```json
{"method":"_x.ai/session/prompt_complete",
 "params":{"sessionId":"<id>","promptId":"t3-xai-prompt-1","stopReason":"error",
           "agentResult":"API error (status 402 Payment Required): ..."}}
```

(Same mock approach suggested in #7830.)

### Expected behavior

The turn is marked failed and Grok's error text is surfaced in the UI, the way T3
already renders `{ state: "failed", errorMessage }` turns for other failure paths.

### Actual behavior

The turn is recorded as a **normal, successful turn**. There is no error anywhere in the
product: no toast, no banner, no failed-turn styling, `projection_turns.state =
'completed'`, `turn.completed` carries `stopReason: "end_turn"`, and
`provider_session_runtime.lastError` is `null`. The user sees the thread simply stop
answering — indistinguishable from a hang.

The error text is received and parsed, then discarded. Two places, both in
`apps/server/src/provider`:

**1. `acp/XAiAcpExtension.ts` — the stop reason is laundered.** Grok's `"error"` falls
through to the `default` branch:

```ts
function normalizeXAiStopReason(value: string | undefined): EffectAcpSchema.StopReason {
  switch (value) {
    case "cancelled":
    case "end_turn":
    case "max_tokens":
    case "max_turn_requests":
    case "refusal":
      return value;
    default:
      return "end_turn";   // <- "error" lands here
  }
}
```

**2. `acp/XAiAcpExtension.ts` — the message is stored in a key nothing reads.**
`promptResponseFromXAi()` does:

```ts
if (notification.agentResult !== undefined) {
  meta.agentResult = notification.agentResult;
}
```

[...]

Issue #3033, opened 2026-06-10
[Bug]: the *Claude provider surfaces the *selected* model, not the served* model. silent Fable 5 → Opus guardrail fallback is invisible


snipemanmike:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/web

### Steps to reproduce

use fable 5
profit?

### Expected behavior

A turn is attributed to the model that actually produced it. If a server-side guardrail fallback swaps Fable 5 → Opus, T3 Code should reflect Opus for that turn or at minimum flag that the served model differs from the selected one.

### Actual behavior

The thread keeps showing the selected model (Fable 5). The served model is never surfaced, so an Opus fallback is silent.

### Impact

Major degradation or frequent failure

### Version or commit

v0.0.28-nightly.20260610.520 @ 0b40ea6

### Logs or stack traces

```shell

```

basnijholt:
I think this is actually a major problem for long-running threads and it's the reason why long tasks I am now going back to the terminal, unfortunately.

kirillrocks:
Any plan to solve it?

juliusmarminge:
Closing as fixed by https://github.com/pingdotgg/t3code/pull/8853 (surfaces Claude safety model fallback notices).

PR #8853, opened 2026-08-31
fix(server): surface Claude safety model fallback notices instead of dropping them

darahaas15:
## What changed

When Anthropic's safety classifier flags a request, the Claude CLI switches the session to a fallback model (e.g. Fable 5 to Opus 4.8) and emits a `system/model_refusal_fallback` message whose `content` field carries a user-facing notice:

> Fable 5's safeguards flagged this message. Our intentionally broad safeguards allow us to deliver more capabilities faster, but can sometimes flag legitimate coding, cybersecurity, and biology tasks. Switched to Opus 4.8. [...] Details: `[cyber]`

`ClaudeAdapter` currently consumes this subtype as a deliberate no-op (grouped with the UX-internal subtypes under the "no T3 surface today" comment), so the notice never reaches any T3 surface: no warning row, no activity, nothing in the projected thread. The user keeps working with no indication that their turns are now served by a different model - and since the fallback is session-scoped, every subsequent turn stays on the fallback model, still silently.

This PR emits the CLI's own notice text as a `runtime.warning`, exactly like the adjacent high-priority `notification` case one branch above. No new event type, no new activity kind, no UI or schema change - the row renders through the existing runtime-warning path.

## Why it should exist

- The CLI ships `content` specifically for display. Dropping it leaves the user misinformed about which model is serving (and being billed for) their session.
- This is not a rare edge case. In my own provider event logs the fallback fired 30 times across recent sessions (`original_model: claude-fable-5` / `claude-opus-5[1m]` -> `fallback_model: claude-opus-4-8`, `api_refusal_category: "cyber"`) with zero visible indication in T3 - security-adjacent coding work trips the classifier regularly.
- Same class of bug as #8293 (provider signal silently normalized away, user shown a misleading picture).

## Changes

- `apps/server/src/provider/Layers/ClaudeAdapter.ts`: move `model_refusal_fallback` out of the silent-consume group and emit `message.content` via the existing `emitRuntimeWarning` helper, mirroring the high-priority `notification` case. The full SDK message is passed as the warning `detail`, so the structured fields (`original_model`, `fallback_model`, `api_refusal_category`, `direction`) stay available downstream.
- `apps/server/src/provider/Layers/ClaudeAdapter.test.ts`: move the subtype out of the consumed-silently fixture list and add a positive case asserting the notice surfaces as exactly one warning row with the CLI's text.

No visual change beyond a standard runtime-warning row appearing (same surface used today for high-priority CLI notifications such as context-limit warnings), with the notice text quoted above.

---

> [!NOTE]
> **Medium Risk**
> Behavior change on the provider event stream: sessions that hit safety fallback now emit an extra `runtime.warning` that downstream code must tolerate or display.
> 
> **Overview**
> When the Claude CLI switches models after a safety refusal (`system/model_refusal_fallback`), users previously got no signal because **`ClaudeAdapter` swallowed that subtype** with other UX-internal system messages.
> 
> This PR **emits the CLI’s `content` text as a `runtime.warning`** via the existing `emitRuntimeWarning` path (same idea as high-priority `notification` rows). Structured fields stay on the warning `detail`; no new event types or UI wiring.
> 
> The **“undeclared / UX-internal subtypes” test** now treats `model_refusal_fallback` as a positive warning case (four expected warning messages) and collects events with **`Stream.takeUntil` on the `api_retry` heartbeat** instead of manual yields and fiber interrupt.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d5c5926a05505f531d0a9109bfdc606e454cdfae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Surface `model_refusal_fallback` warnings in `makeClaudeAdapter`
> Moves the `model_refusal_fallback` system subtype out of the silently ignored set in the Claude system-event handler and emits a runtime warning with the message's user-facing content. Updates the integration test in [ClaudeAdapter.test.ts](https://github.com/pingdotgg/t3code/pull/8853/files#diff-1657a55ebc1e3682293aa20a7f71a55adf8de546f36af50f289dca9da45b0b6f) to collect the event stream deterministically and assert the fallback notice appears in the warning list.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d5c5926.</sup>
> 
>

## Audit validation

Reproduced on current main [0dd5c64bc](https://github.com/pingdotgg/t3code/commit/0dd5c64bcfac7974d81bac76740ecd83540077e8) on September 4, 2026: replaying the native fallback notice produced no warning before a following notification receipt. This is the missing-notice behavior in [#3033](https://github.com/pingdotgg/t3code/issues/3033), without guessing the served model from generated text.

Merged current main into the existing author branch, preserving its warning mapping and newer upstream warning cases. The test now drains through the final retry-state receipt instead of relying on scheduler yields. All 85 focused ClaudeAdapter tests, server typecheck, and targeted lint pass on [d5c5926a0](https://github.com/darahaas15/t3code/commit/d5c5926a05505f531d0a9109bfdc606e454cdfae).

This changes provider-event mapping only and uses the existing warning display on all clients. No layout changes or authenticated provider generations were needed for verification.

Audit preparation: GPT 6 Astra via Codex in T3 Code.


darahaas15:
A few data points for review, since the bar here is "explain exactly why this should exist":

**It cannot storm the work log.** The SDK emits this once per refused turn: *"the turn is retried once on a fallback model with the swap made persistent for the session"* (`SDKModelRefusalFallbackMessage` docs, `@anthropic-ai/claude-agent-sdk` 0.3.170). Real traffic matches: in my provider event logs it fired 31 times across 31 distinct turns in 15 sessions, i.e. exactly one row per flagged turn, never more. That is the opposite profile from the `api_retry` heartbeat this adapter deliberately keeps out of the work log.

**Why silence is worse here than for a normal warning.** Per the same docs the swap is persistent for the session, so one dropped notice means every subsequent turn is served by the fallback model with nothing in T3 indicating it. That is exactly the current behavior: across those 31 events, zero rows reached `projection_thread_activities`.

**No client work, all surfaces.** `runtime.warning` is already ingested by `ProviderRuntimeIngestion` and rendered on web (`MessagesTimeline`) and mobile (`threadActivity`), so the row appears everywhere with no client changes, the same way high-priority CLI notifications do today.

**Before / after.** Before: nothing. After: one standard runtime-warning row containing the CLI's own text, e.g.

> Fable 5's safeguards flagged this message. [...] Switched to Opus 4.8. [...] Details: `[cyber]`

Happy to attach a screenshot of the rendered row if that helps.

One nit pre-empted: `direction` also declares `revert` / `sticky`, but the SDK docs mark those as retained for compat and no longer emitted, so surfacing `content` on a single path is sufficient; no per-direction wording needed.

PR #8857, opened 2026-08-31
feat(web): generate and preserve thread handover drafts

saphid:
When a thread reaches its context limit, the user has no way to carry the work into a fresh thread without retyping the state by hand.

This adds a handover action on web and mobile. The server generates a continuation brief through an enabled Codex instance (`orchestration.generateHandover`, advertised by the `threadHandoverGeneration` capability), and the client opens it as an unsent draft in a new thread in the same checkout, so the user can pick the model and reasoning level before sending. Other providers return a typed "not supported" error. Web persists the generated brief before navigating and shows saved state after remount. Mobile records the import in a durable per-thread draft, so restarting or retrying reopens that draft without duplicating the brief or overwriting later edits.

**Depends on unmerged PRs.** This branch is cumulative: its first two commits are squashed copies of #10095 (hard context admission, at `61374dce6`) and #10097 (shared concurrency reservations, at `1061345fe`). Merge those first. The handover delta is the last commit only. This stays a draft until both land.

Rebased linearly onto `main` at `211618fd9`, with no merge commits. Conflict resolutions adopt upstream's composer imports, the mobile `GlassBlurTargetContext` wrapper, and the `threadPullRequests`/`pullRequestStackActions` capabilities. They also fit #10097's reservations into the compaction queue from #11107. Messages sent during compaction are queued ahead of the context-admission and reservation gates, so an over-limit thread no longer rejects them. They replay after compaction lowers usage, and each replay takes its own slot. A new `ProviderCommandReactor` test covers this.

Verification on `623afe900`:
- Server: 11 files, 446 tests pass (handover, text generation, prompts, environment descriptor, reservations, `ProviderCommandReactor`, the WebSocket missing-thread handover error in `server.test.ts`)
- Web: 4 files, 19 tests pass. Mobile: 3 files, 69 tests pass (includes restart/retry after migrating an older draft receipt). Shared: 9 pass. Contracts: 106 pass.
- `tsc --noEmit` passes for contracts, shared, client-runtime, server, web, and mobile. `vp lint` on changed files: 0 errors.

Still pending: before/after web and mobile screenshots, and a live provider handover run. The cross-provider (GPT-6 Astra) review was skipped because Codex weekly quota was below 10%.

Coordination trace: T3 thread 272cea79-ab91-4570-91a4-10aee73ba516

Originally implemented with GPT-6 Astra in the Codex harness; rebased and verified with Claude Opus 5 in Claude Code.

### DIFF
--- a/apps/mobile/src/features/threads/ThreadComposer.tsx
+++ b/apps/mobile/src/features/threads/ThreadComposer.tsx
@@ -36,7 +36,9 @@ import {
 } from "../../state/composer-attachment-uploads";
 import Animated, {
   FadeIn,
+  FadeInDown,
   FadeOut,
+  FadeOutDown,
   type LayoutAnimationFunction,
   ReduceMotion,
   useAnimatedStyle,
@@ -118,6 +120,9 @@ export interface ThreadComposerProps {
   readonly selectedThread: OrchestrationThreadShell;
   readonly hasCompactableConversation: boolean;
   readonly serverConfig: T3ServerConfig | null;
+  readonly contextLimitReached: boolean;
+  readonly isGeneratingHandover: boolean;
+  readonly onGenerateHandover?: () => void;
   readonly queueCount: number;
   readonly environmentId: EnvironmentId;
   readonly projectCwd: string | null;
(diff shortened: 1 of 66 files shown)


juliusmarminge:

PR #8902, opened 2026-08-31
fix(web): preserve explicit preview navigation URLs

### LINKED ISSUE #8885: [Bug]: preview_navigate rewrites explicit loopback URLs to the environment host and reports success when the page fails to load
### Before submitting

### Area

apps/web

### Steps to reproduce

1. Run the T3 Code server bound to a non-loopback host that is directly reachable on the private network, for example `t3 serve --host 192.168.1.50`. `userdata/server-runtime.json` then records an origin such as `http://192.168.1.50:3773`.
2. Start a development server that listens on loopback only:

   ```shell
   python3 -m http.server 5999 --bind 127.0.0.1
   ```

3. Confirm the binding is loopback only:

   ```shell
   curl -o /dev/null -w '%{http_code}\n' http://127.0.0.1:5999/   # 200
   curl -o /dev/null -m 5 -w '%{http_code}\n' http://192.168.1.50:5999/   # exit 7, connection refused
   ```

4. Open a preview tab with `preview_open({ open: true })`.
5. Navigate with an explicit, fully qualified loopback URL:

   ```json
   { "url": "http://127.0.0.1:5999/" }
   ```

6. Inspect the real page state with `preview_evaluate`:

   ```js
   ({ href: location.href, title: document.title, bodyText: document.body.innerText.slice(0, 200) })
   ```

7. Repeat step 5 with `{ "url": "localhost:5999" }` and with `{ "target": { "kind": "environment-port", "port": 5999 } }`.

### Expected behavior

Two things, either of which would be enough:

1. An explicit `url` target should keep its requested origin. #3938 established this for the preview address bar, and #8062 argues the same for the MCP tools: environment-relative resolution should only happen when the caller passes an explicit `environment-port` target.
2. Failing that, a navigation that does not actually load should report an error rather than success.

### Actual behavior

All three forms are silently rewritten to the environment host, and `preview_navigate` returns a success payload for a page that never loaded.

Step 5 returns:

```json
{
  "available": true,
  "visible": true,
  "tabId": "tab_3",
  "url": "http://192.168.1.50:5999/",
  "title": "192.168.1.50:5999",
  "loading": false
}
```

Step 6 shows the tab is actually on a Chromium error page:

```json
{ "href": "chrome-error://chromewebdata/", "title": "", "bodyText": "" }
```

Note that the returned `title` is just the rewritten host string, which reads like a real page title. An agent that trusts the `preview_navigate` result has no signal that anything went wrong, and will go on to snapshot or click an error page, or report that the development server is healthy.

The rewrite is deterministic in `apps/web/src/browser/browserTargetResolver.ts`, in `resolveBrowserNavigationTarget`:

```ts
if (parsed && isLoopbackHost(parsed.hostname)) {
  const environmentUrl = readEnvironmentUrl(environmentId);
  if (parsed.hostname === "0.0.0.0" || !isLocalLoopbackHost(environmentUrl.hostname)) {
    return resolveEnvironmentPortTarget(/* ... */);
  }
}
```

Any explicit loopback URL becomes an environment-port target whenever the environment hostname is not itself loopback.

### Relationship to existing issues

This is not covered by the existing reports, though it shares a code path with them:

- #3938 described the same rewrite for WSL and was fixed by #3939, but that PR only changed `apps/web/src/components/preview/PreviewView.tsx` and its test. The address bar preserves direct localhost navigation now; the MCP `preview_navigate` and `preview_open` path still goes through the resolver unchanged.
- #8062 reports the same resolver behavior under T3 Connect, where the backend hostname is public and `resolveEnvironmentPortTarget` rejects the request outright. This report covers the case where the environment host *is* directly private-network reachable, so the resolver does not reject. It rewrites, navigation fails at the network layer, and the call still reports success.
- #8732 is the inverse failure mode (a timeout reported for a navigation that completed). This is a success reported for a navigation that failed.

### Impact

### Version or commit

`0.0.38-nightly.20260831.1236`

### Environment

Linux x86_64, server started with an explicit non-loopback `--host`, MCP preview tools driven over the HTTP MCP endpoint.

### Logs or stack traces

No server-side error is logged, because from the resolver's point of view the navigation succeeded. The only evidence is the mismatch between the `preview_navigate` return value and `location.href`.

None. The reproduction is entirely in the JSON payloads above.

### Workaround

Bind development servers to `0.0.0.0` instead of `127.0.0.1`, and treat `preview_evaluate` on `({ href: location.href, title: document.title })` as the only reliable confirmation that a navigation landed where it was asked to.

nateEc:
Fixes #8885.

Explicit preview URLs, including loopback URLs, were rewritten to the environment host before navigation. That can direct a valid client-local URL to an unreachable address and report misleading success.

Keep explicit URL targets unchanged; discovered local-server entries continue to use the explicit environment-port resolution path.

Verification: focused browser target resolver tests (19 passing). Web typecheck remains blocked by existing missing optional modules and unrelated Clerk type errors.

Model and harness: GPT-5 Codex via Codex CLI.

---

> [!NOTE]
> **Medium Risk**
> Splits URL resolution between user-entered navigation and discovered servers; a mistake in either path would break remote previews or reintroduce unreachable loopback rewrites.
> 
> **Overview**
> Fixes incorrect rewriting of **explicit** preview navigation targets (loopback URLs, credentials, schemeless `localhost`) to the remote environment host, which could send client-local addresses to unreachable hosts.
> 
> `resolveBrowserNavigationTarget` now returns **`kind: "url"`** targets unchanged with `resolutionKind: "direct"`. **Environment-port** resolution and private-network mapping are unchanged.
> 
> Loopback-to-remote-host mapping moves to **`resolveDiscoveredServerUrl`** only (server-picker / discovered dev servers), so picking `localhost:3000` on a remote environment still resolves to that host’s IP while typed URLs stay as entered.
> 
> Tests are updated to match; explicit `http://localhost:5173` against a public relay base no longer throws the authenticated-gateway error.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49a7407a96e3299854ac18b9d0ece00745702c4a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Preserve explicit loopback URLs in `resolveBrowserNavigationTarget`
> - `resolveBrowserNavigationTarget` now returns all `url`-kind targets unchanged with `resolutionKind: 'direct'`, instead of remapping loopback hosts (e.g. `localhost`, `127.0.0.1`) to the environment host via `resolveEnvironmentPortTarget`
> - `resolveDiscoveredServerUrl` no longer delegates to `resolveBrowserNavigationTarget`; it independently normalizes the URL, maps loopback hosts to the environment host via `resolveEnvironmentPortTarget`, and returns non-loopback URLs as-is
> - Tests updated to verify explicit loopback URLs (including credentialed, schemeless, and IPv4 forms) stay unchanged across remote, private-network, and relay environments
> - Behavioral Change: explicit navigation to a loopback URL no longer redirects through the environment host; discovered server URLs still do. Reviewers should check `resolveBrowserNavigationTarget` in [browserTargetResolver.ts](https://github.com/pingdotgg/t3code/pull/8902/files#diff-d3c7988731a040ddf1171c7c3b826a4b211a89884507fdb562b4bb24fe5aba39) for any remaining callers that relied on loopback remapping for explicit URLs
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 49a7407.</sup>
> 
>

### DIFF
--- a/apps/web/src/browser/browserTargetResolver.test.ts
+++ b/apps/web/src/browser/browserTargetResolver.test.ts
@@ -25,7 +25,7 @@ describe("browser target resolver", () => {
     });
   });
 
-  it("maps localhost URL navigation onto a remote Tailscale IPv4 host", async () => {
+  it("preserves explicit loopback URL navigation for a remote Tailscale environment", async () => {
     readPreparedConnection.mockReturnValue({ httpBaseUrl: "http://100.65.180.100:3773" });
     const { resolveBrowserNavigationTarget } = await import("./browserTargetResolver");
     expect(
@@ -35,13 +35,29 @@ describe("browser target resolver", () => {
       }),
     ).toEqual({
       requestedUrl: "http://localhost:5173/dashboard?mode=test#results",
-      resolvedUrl: "http://100.65.180.100:5173/dashboard?mode=test#results",
-      resolutionKind: "direct-private-network",
+      resolvedUrl: "http://localhost:5173/dashboard?mode=test#results",
+      resolutionKind: "direct",
       environmentId: "environment-1",
     });
   });
 
-  it("preserves URL credentials when mapping localhost onto a remote host", async () => {
+  it("preserves explicit IPv4 loopback URL navigation for a private network environment", async () => {
+    readPreparedConnection.mockReturnValue({ httpBaseUrl: "http://192.168.1.50:3773" });
+    const { resolveBrowserNavigationTarget } = await import("./browserTargetResolver");
+    expect(
+      resolveBrowserNavigationTarget(EnvironmentId.make("environment-1"), {
+        kind: "url",
+        url: "http://127.0.0.1:5999/",
+      }),
+    ).toEqual({
+      requestedUrl: "http://127.0.0.1:5999/",
+      resolvedUrl: "http://127.0.0.1:5999/",
+      resolutionKind: "direct",
+      environmentId: "environment-1",
+    });
+  });
+
+  it("preserves URL credentials on explicit loopback navigation", async () => {
     readPreparedConnection.mockReturnValue({ httpBaseUrl: "http://100.65.180.100:3773" });
     const { resolveBrowserNavigationTarget } = await import("./browserTargetResolver");
     expect(
@@ -49,10 +65,10 @@ describe("browser target resolver", () => {
         kind: "url",
         url: "http://user:p%40ss@localhost:5173/dashboard",
       }).resolvedUrl,
-    ).toBe("http://user:p%40ss@100.65.180.100:5173/dashboard");
+    ).toBe("http://user:p%40ss@localhost:5173/dashboard");
   });
 
-  it("maps credentialed localhost URLs onto private IPv6 hosts", async () => {
+  it("preserves credentialed loopback URLs for private IPv6 environments", async () => {
     readPreparedConnection.mockReturnValue({
       httpBaseUrl: "http://[fd7a:115c:a1e0::53]:3773",
     });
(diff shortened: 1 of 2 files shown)


nateEc:
@juliusmarminge All checks are green and Macroscope approved the latest commit. Could you take a human review when you have a moment?

PR #8950, opened 2026-08-31
fix(server): cache the resolved git top level per workspace root

### LINKED ISSUE #8949: [Bug]: Repository identity and VCS detection spawn git per project on every read, stalling shellSnapshot
### Area

apps/server

### Steps to reproduce

1. Add enough projects that `projection_projects` holds a few dozen active rows. Mine has 34.
2. Put the machine under real load, for example several concurrent agent sessions.
3. Reconnect a client, or start a turn, so `getShellSnapshot` and VCS refreshes run.
4. Watch `RepositoryIdentityResolver.resolve` and `VcsDriverRegistry.detect` in `server.trace.ndjson`.

### Expected behavior

Resolving which repository a project belongs to is cached. A workspace root maps to the same git top level for the life of the checkout, so a snapshot read should not spawn `git` once per project every time it runs.

### Actual behavior

Two independent caching gaps make snapshot reads spawn one or more `git` processes per project on every call.

**1. `RepositoryIdentityResolver.resolve` runs an uncached `git rev-parse`.**

`resolve(cwd)` has two steps. Step two, `git remote -v`, is cached behind `repositoryIdentityCache`. Step one derives that cache's *key* by running `git rev-parse --show-toplevel`, and is not cached at all:

https://github.com/pingdotgg/t3code/blob/31c1c5996/apps/server/src/project/RepositoryIdentityResolver.ts#L160-L166

https://github.com/pingdotgg/t3code/blob/31c1c5996/apps/server/src/project/RepositoryIdentityResolver.ts#L90-L113

`ProjectionSnapshotQuery.resolveRepositoryIdentitiesForProjects` resolves every active project on every call, at concurrency 4:

https://github.com/pingdotgg/t3code/blob/31c1c5996/apps/server/src/orchestration/Layers/ProjectionSnapshotQuery.ts#L359-L375

So the cache never prevents a spawn. 34 projects means 34 `git` processes per snapshot read, however warm the cache is. `DEFAULT_POSITIVE_CACHE_TTL` is also `Duration.minutes(1)`, so step two re-spawns for every root at least once a minute.

**2. `VcsDriverRegistry` caches repository detection for 2 seconds.**

https://github.com/pingdotgg/t3code/blob/31c1c5996/apps/server/src/vcs/VcsDriverRegistry.ts#L15

`GitVcsDriver.detectRepository` spawns three git processes per cwd: `rev-parse --is-inside-work-tree`, `rev-parse --show-toplevel`, and `rev-parse --git-common-dir`. Any VCS call more than two seconds after the previous one re-runs all three for the same checkout.

Two seconds is an outlier in this layer. Every neighbouring cache in `GitVcsDriverCore.ts` is measured in minutes:

```
REPOSITORY_PATHS_CACHE_TTL      = Duration.minutes(10)
LIST_REFS_SNAPSHOT_CACHE_TTL    = Duration.minutes(2)
STATUS_DEFAULT_BRANCH_CACHE_TTL = Duration.minutes(5)
STATUS_ORIGIN_EXISTS_CACHE_TTL  = Duration.minutes(5)
```

Individually a git spawn is cheap. In aggregate, with these all queueing on a loaded machine, each `resolve` was observed at 4 to 6.5 s and `VcsDriverRegistry.detect` at 4.8 to 5.9 s.

### Impact

`environment.orchestration.shellSnapshot` took **32.9 s** on my install, which fires the 15s slow-request toast. It is on the client bootstrap and reconnect path.

Trace breakdown of that request:

```
33755ms  http.server GET
32954ms  environment.orchestration.shellSnapshot
32499ms    ProjectionSnapshotQuery.resolveRepositoryIdentitiesForProjects
 6508ms      RepositoryIdentityResolver.resolve
 5841ms      RepositoryIdentityResolver.resolve
 5838ms      RepositoryIdentityResolver.resolve
 ...  34 of these, at concurrency 4
 4070ms        RepositoryIdentityResolver.resolveFromCacheKey
 4057ms          processRunner.runProcessCore
```

Worth stating plainly, because it is the obvious suspect and it is wrong: **this is not SQL.** In the same window, across 117,031 `sql.execute` spans against a 2.9 GB `state.sqlite`, p50 was 0.04 ms, p99 was 16 ms, and not one query exceeded 1 second.

The cost scales with the number of projects, so it gets worse as an install grows.

### Version or commit

`main` at 31c1c5996. Both constants are unchanged since well before that; also present in 0.0.36.

### Environment

macOS, desktop app hosting its own server, client connected over `127.0.0.1`. 34 active projects.

### Logs or stack traces

The clearest signal is the span ratio. Before any fix, `RepositoryIdentityResolver.resolve` and its inner `git rev-parse` occur 1:1, and `VcsDriverRegistry.detect` and `detectRepository` do too, because neither result survives to the next call.

After caching both, on the same install:

```
n=63   RepositoryIdentityResolver.resolve
n=34   RepositoryIdentityResolver.resolveRepositoryRoot     <- one per distinct root, cold fill only
n=291  VcsDriverRegistry.detect
n=160  detectRepository
```

and `shellSnapshot` drops from 32,954 ms to 1,744 ms, which is itself the cold startup run.

### Workaround

None from the client side. Reducing the number of projects reduces the spawn count proportionally.

willsheldon:
Fixes #8949.

## Problem

`RepositoryIdentityResolver.resolve(cwd)` has two steps. Step two, `git remote -v`, is cached behind `repositoryIdentityCache`. Step one, `resolveRepositoryIdentityCacheKey`, runs `git rev-parse --show-toplevel` to derive that cache's key, and is not cached at all.

`ProjectionSnapshotQuery.resolveRepositoryIdentitiesForProjects` resolves every active project on every call, at concurrency 4. So the cache never prevents a spawn. On an install with 34 projects, each snapshot read spawns 34 `git` processes no matter how warm the cache is, and `getShellSnapshot` runs on every client bootstrap and reconnect.

A traced `shellSnapshot` on that install took 32.9 s, of which 32.5 s was inside `resolveRepositoryIdentitiesForProjects`, split across 34 `resolve` calls of 4 to 6.5 s each. SQL was not involved: across 117,031 `sql.execute` spans in the same window, p99 was 16 ms and nothing exceeded 1 s.

`DEFAULT_POSITIVE_CACHE_TTL` was also `Duration.minutes(1)`, so step two re-spawned for every root at least once a minute.

## Fix

Cache the workspace-root-to-top-level mapping in its own `repositoryRootCache`, alongside the identity it keys. `resolveRepositoryRoot` now returns `string | null` rather than falling back to `cwd`, so the null case can carry its own TTL.

Raise `DEFAULT_POSITIVE_CACHE_TTL` to 10 minutes. A remote URL change now surfaces in up to 10 minutes instead of 1. The negative TTL stays at 1 minute, so a fresh `git init` is still picked up quickly.

Measured on the install above, restarting the server with this change:

| | before | after |
|---|---|---|
| `environment.orchestration.shellSnapshot` | 32,954 ms | 1,744 ms |
| `resolveRepositoryIdentitiesForProjects` | 32,499 ms | 1,638 ms |

Both after-figures are the cold startup run. Seven minutes in, 63 `resolve` calls had produced exactly 34 `resolveRepositoryRoot` spans, one per distinct root, and no more.

The added test turns a nested workspace into its own repository after the first resolve. While the root stays cached the resolver keeps answering from the outer repository, which is only possible if it did not spawn git again; past the TTL it picks up the inner one.

## Verification

- `vp test run apps/server/src/project/RepositoryIdentityResolver.test.ts` — 8 passed
- `vp test run apps/server/src/project` — 35 passed
- `vp test run apps/server/src/orchestration/Layers/ProjectionSnapshotQuery.test.ts` — 21 passed
- `tsgo --noEmit` in `apps/server` — clean

---

> [!NOTE]
> **Low Risk**
> Performance-oriented caching in repository metadata resolution; remote or root changes may be stale until TTL expiry, with no auth or data-mutation impact.
> 
> **Overview**
> **Fixes snapshot/shell bootstrap slowness** by caching `git rev-parse --show-toplevel` per workspace cwd instead of spawning git on every `resolve`, which previously happened even when remote identity was already cached.
> 
> `RepositoryIdentityResolver` now uses a separate **`repositoryRootCache`** (configurable positive/negative TTLs, default 10m / 1m) ahead of the existing identity cache. Root lookup returns **`null`** when rev-parse fails rather than falling back to `cwd`. **`DEFAULT_POSITIVE_CACHE_TTL`** for resolved identities is raised from **1 minute to 10 minutes**.
> 
> A new test proves repeated resolves reuse the cached top level until TTL expiry (nested repo init does not change identity until the root cache refreshes).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c71aa17fa131c150d3b5b65c740eaf7c8965dd9e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Cache resolved git top level per workspace root in `RepositoryIdentityResolver`
> - Adds a `repositoryRootCache` keyed by cwd that stores the result of `git rev-parse --show-toplevel`, with separate TTLs for positive hits (`repositoryRootCacheTtl`, 10 min default) and negative results (`repositoryRootNegativeCacheTtl`, 1 min default).
> - `resolveRepositoryRoot` now returns `null` on failure instead of falling back to cwd, enabling distinct negative caching.
> - Increases `DEFAULT_POSITIVE_CACHE_TTL` for the identity cache from 1 minute to 10 minutes.
> - Risk: while the root cache is fresh, `resolve` may return the identity of a previously discovered parent repository even if a nested workspace later becomes its own repository, until the root cache entry expires. Reviewers should check the `repositoryRootCache` TTL configuration in [RepositoryIdentityResolver.ts](https://github.com/pingdotgg/t3code/pull/8950/files#diff-150725a1a7c2fad32a2af878e610161760c1d30b61e1fc4e4aaba35d01587171).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c71aa17.</sup>
> 
>

## Summary by CodeRabbit

* **Performance Improvements**
  * Repository identity resolution now reuses cached repository roots, reducing repeated checks.
  * Positive repository identity results remain cached longer by default.

* **Reliability**
  * Failed repository-root lookups are handled without incorrectly substituting the current directory.
  * Repository-root cache expiration ensures refreshed results when needed.

### DIFF
--- a/apps/server/src/project/RepositoryIdentityResolver.test.ts
+++ b/apps/server/src/project/RepositoryIdentityResolver.test.ts
@@ -25,6 +25,8 @@ const git = (cwd: string, args: ReadonlyArray<string>) =>
 const makeRepositoryIdentityResolverTestLayer = (options: {
   readonly positiveCacheTtl?: Duration.Input;
   readonly negativeCacheTtl?: Duration.Input;
+  readonly repositoryRootCacheTtl?: Duration.Input;
+  readonly repositoryRootNegativeCacheTtl?: Duration.Input;
 }) =>
   Layer.effect(
     RepositoryIdentityResolver.RepositoryIdentityResolver,
@@ -227,6 +229,58 @@ it.layer(NodeServices.layer)("RepositoryIdentityResolverLive", (it) => {
           makeRepositoryIdentityResolverTestLayer({
             negativeCacheTtl: Duration.millis(50),
             positiveCacheTtl: Duration.millis(100),
+          }),
+        ),
+      ),
+    ),
+  );
+
+  it.effect("reuses the cached git top level instead of re-running rev-parse per resolve", () =>
+    Effect.gen(function* () {
+      const fileSystem = yield* FileSystem.FileSystem;
+      const path = yield* Path.Path;
+      const repoRoot = yield* fileSystem.makeTempDirectoryScoped({
+        prefix: "t3-repository-identity-root-cache-test-",
+      });
+      const nestedWorkspace = path.join(repoRoot, "packages", "web");
+
+      yield* fileSystem.makeDirectory(nestedWorkspace, { recursive: true });
+      yield* git(repoRoot, ["init"]);
+      yield* git(repoRoot, ["remote", "add", "origin", "git@github.com:T3Tools/t3code.git"]);
+
+      const resolver = yield* RepositoryIdentityResolver.RepositoryIdentityResolver;
+      const initialIdentity = yield* resolver.resolve(nestedWorkspace);
+      expect(initialIdentity?.canonicalKey).toBe("github.com/t3tools/t3code");
+
+      // Turning the nested workspace into its own repository changes what
+      // `git rev-parse --show-toplevel` reports. While the root stays cached the
+      // resolver must keep answering from the outer repository, which proves it
+      // did not spawn git again.
+      yield* git(nestedWorkspace, ["init"]);
+      yield* git(nestedWorkspace, [
+        "remote",
+        "add",
+        "origin",
+        "git@github.com:T3Tools/t3code-nested.git",
+      ]);
+
+      for (const _attempt of [1, 2, 3]) {
+        const cachedIdentity = yield* resolver.resolve(nestedWorkspace);
+        expect(cachedIdentity?.canonicalKey).toBe("github.com/t3tools/t3code");
+      }
+
+      yield* TestClock.adjust(Duration.millis(120));
+
+      const refreshedIdentity = yield* resolver.resolve(nestedWorkspace);
+      expect(refreshedIdentity?.canonicalKey).toBe("github.com/t3tools/t3code-nested");
+    }).pipe(
+      Effect.provide(
+        Layer.merge(
+          TestClock.layer(),
+          makeRepositoryIdentityResolverTestLayer({
+            negativeCacheTtl: Duration.millis(50),
+            positiveCacheTtl: Duration.seconds(30),
+            repositoryRootCacheTtl: Duration.millis(50),
           }),
         ),
       ),
--- a/apps/server/src/project/RepositoryIdentityResolver.ts
+++ b/apps/server/src/project/RepositoryIdentityResolver.ts
@@ -13,13 +13,17 @@ import * as Layer from "effect/Layer";
 import * as ProcessRunner from "../processRunner.ts";
 
 const DEFAULT_REPOSITORY_IDENTITY_CACHE_CAPACITY = 512;
-const DEFAULT_POSITIVE_CACHE_TTL = Duration.minutes(1);
+const DEFAULT_POSITIVE_CACHE_TTL = Duration.minutes(10);
 const DEFAULT_NEGATIVE_CACHE_TTL = Duration.minutes(1);
+const DEFAULT_REPOSITORY_ROOT_CACHE_TTL = Duration.minutes(10);
+const DEFAULT_REPOSITORY_ROOT_NEGATIVE_CACHE_TTL = Duration.minutes(1);
 
 export interface RepositoryIdentityResolverOptions {
   readonly cacheCapacity?: number;
   readonly positiveCacheTtl?: Duration.Input;
   readonly negativeCacheTtl?: Duration.Input;
+  readonly repositoryRootCacheTtl?: Duration.Input;
+  readonly repositoryRootNegativeCacheTtl?: Duration.Input;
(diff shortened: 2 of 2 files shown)


willsheldon:
Closing as superseded. Main now caches repository roots in RepositoryIdentityResolver via #8187. The remaining VcsDriverRegistry cache-TTL fix stays in #8951.

Issue #4683, opened 2026-07-27
[Bug]: Background Codex memory activity appears in the active conversation


nassimna:
## Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

## Area

apps/server

## Steps to reproduce

1. Open a normal T3 Code conversation using the Codex provider.
2. Exchange ordinary messages unrelated to any software project.
3. Allow the automatic post-turn memory consolidation to run.
4. Observe the active conversation after the background work starts.

In the observed case, commentary and file-read activity from separate background Codex turns appeared in the active conversation. The activity referenced memory-maintenance files and content from an unrelated workspace.

## Expected behavior

Background memory-maintenance activity should remain isolated from the active conversation. Its messages, tool calls, filenames, and content from other workspaces should not be rendered in the user's chat.

## Actual behavior

The background activity appeared in the current conversation, including references to unrelated project memory. Inspection showed that the ordinary Codex conversation transcript did not contain this activity, while T3's provider log recorded separate internal Codex threads under the same visible T3 conversation.

No full transcript, raw memory file, provider log, or screenshot is attached because each contains personal or project information. Sanitized diagnostics can be provided privately if needed.

## Impact

Major degradation or frequent failure

This crosses conversation and workspace boundaries and can expose unrelated memory content in a personal chat.

## Version or commit

0.0.29-nightly.20260724.893

## Environment

Arch Linux, T3 Code desktop AppImage, Codex app-server provider, gpt-5.6-sol

## Logs or stack traces

```shell
Sanitized event sequence:
- A normal assistant turn completed.
- A separate Codex provider thread started background memory consolidation.
- That background thread emitted commentary and file-read events.
- The events were associated with the active canonical T3 conversation and became visible there.
```

## Workaround

No reliable workaround found. Avoiding or clearing the affected conversation does not address the underlying isolation problem.

Kabilan108:
+1 i'm getting the same behavior in 0.0.33.

PR #8989, opened 2026-09-01
fix(codex): keep foreign memory out of chats

SunkenInTime:
Codex background memory work can begin emitting item notifications on an existing app-server subscription without sending a new thread/started notification. The stateful memory filter never learns that provider thread ID, so its assistant text can be projected into the active T3 chat.

This treats the active Codex provider thread ID as the chat ownership boundary. Known notifications addressed to another thread are suppressed even when their lifecycle start was missed. Registered multi-agent children still use their dedicated routing, and serverRequest/resolved still passes through for approval correlation cleanup.

The regression replay matches the observed ordering: agent message delta and item completion for a memory thread with no preceding thread/started.

Verification:
- vp test run apps/server/src/provider/Layers/CodexSessionRuntime.test.ts apps/server/src/provider/Layers/CodexCollabWire.test.ts apps/server/src/provider/Layers/CodexCollabRuntime.integration.test.ts
- vp lint --type-aware --type-check --report-unused-disable-directives apps/server/src/provider/Layers/CodexSessionRuntime.ts apps/server/src/provider/Layers/CodexSessionRuntime.test.ts apps/server/src/provider/Layers/CodexCollabRuntime.integration.test.ts
- vp fmt --check apps/server/src/provider/Layers/CodexSessionRuntime.ts apps/server/src/provider/Layers/CodexSessionRuntime.test.ts apps/server/src/provider/Layers/CodexCollabRuntime.integration.test.ts
- git diff --check upstream/main...HEAD

Generated with GPT-5.6 Sol in T3 Code via the Codex harness.

---

> [!NOTE]
> **Medium Risk**
> Changes Codex notification routing for all non-root threads; incorrect suppression could hide legitimate cross-thread traffic, though collab interception and serverRequest/resolved exceptions limit blast radius.
> 
> **Overview**
> Fixes **background Codex memory work** showing up in the active T3 chat when item deltas arrive on an existing app-server subscription **without** a preceding `thread/started`, so the stateful memory filter never learns that thread id.
> 
> The runtime now treats the **active provider thread id** as the chat boundary: exported `shouldSuppressForeignConversationNotification` drops any known notification whose `threadId` differs from the root, including `item/agentMessage/delta` and `item/completed`, independent of lifecycle registration. **Registered collab children** still go through interception first; **`serverRequest/resolved`** still passes through for approval correlation.
> 
> Suppression logic is split so v1 receiver-map children keep lifecycle-only routing via `shouldSuppressChildConversationNotification`, while the new helper handles the broader foreign-thread case (including the memory leak ordering).
> 
> Tests cover the helper directly and extend the collab integration replay to assert no events leak for the memory thread.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 509d07f2575741c971de245c0fbba691ad346184. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Suppress foreign memory-thread notifications in `makeCodexSessionRuntime`
> - Adds `shouldSuppressForeignConversationNotification` to classify non-root-thread notifications as suppressible, except `serverRequest/resolved`.
> - The notification loop now suppresses all notifications addressed to non-root provider threads (item deltas, completions) instead of only selected lifecycle events, preventing background memory output from appearing in the parent session.
> - Adds unit tests for the new predicate and an integration assertion that no events from the background memory thread leak without a prior `thread/started`.
> - Risk: `makeCodexSessionRuntime` now drops all non-root-thread `item/*` notifications; any consumer relying on foreign thread deltas appearing in the session stream will no longer receive them.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 509d07f.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented background memory-thread messages from appearing in the parent conversation stream before the thread is initialized.
  - Improved filtering of notifications from unrelated conversation threads while preserving valid root-thread updates and resolved server requests.
  - Added coverage for memory output and cross-thread notification handling to ensure these events remain isolated.

### DIFF
--- a/apps/server/src/provider/Layers/CodexCollabRuntime.integration.test.ts
+++ b/apps/server/src/provider/Layers/CodexCollabRuntime.integration.test.ts
@@ -62,6 +62,30 @@ function buildScript() {
         },
       },
     },
+    // A real memory leak began at these item notifications without any
+    // thread/started lifecycle notification for the memory thread.
+    {
+      method: "item/agentMessage/delta",
+      params: {
+        delta: "internal memory update",
+        itemId: "memory-message",
+        threadId: MEMORY,
+        turnId: "memory-turn",
+      },
+    },
+    {
+      method: "item/completed",
+      params: {
+        completedAtMs: 0,
+        item: {
+          id: "memory-message",
+          text: "internal memory update",
+          type: "agentMessage",
+        },
+        threadId: MEMORY,
+        turnId: "memory-turn",
+      },
+    },
     // Child terminal lifecycle AFTER the receiver map knows the children —
     // pre-fix, the legacy suppressor dropped these before interception saw
     // them, so no synthetic agent events were emitted.
@@ -467,6 +491,16 @@ describe("CodexSessionRuntime collab integration", () => {
         leaked.map((event) => event.method),
         [],
         "child thread/* lifecycle must not appear as parent events",
+      );
+
+      const leakedMemory = events.filter((event) => {
+        const payload = event.payload as { threadId?: string } | undefined;
+        return payload?.threadId === MEMORY;
+      });
+      assert.deepEqual(
+        leakedMemory.map((event) => event.method),
+        [],
+        "memory output must not appear without a preceding thread/started notification",
       );
 
       yield* runtime.close;
--- a/apps/server/src/provider/Layers/CodexSessionRuntime.test.ts
+++ b/apps/server/src/provider/Layers/CodexSessionRuntime.test.ts
@@ -20,6 +20,7 @@ import {
   openCodexThread,
   readCodexThread,
   rollbackCodexThread,
+  shouldSuppressForeignConversationNotification,
   toMcpElicitationResponse,
 } from "./CodexSessionRuntime.ts";
 const isCodexAppServerRequestError = Schema.is(CodexErrors.CodexAppServerRequestError);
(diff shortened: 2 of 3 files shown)


gfdelarue:
Confirmed another occurrence on 2026-09-05 with T3 desktop 0.0.38 on NixOS, source checkout `77b655c47`, and Codex CLI `0.153.3`. The running package already includes the earlier #5468 filter.

Observed reproduction:

1. Ask an ordinary Git question in a Codex chat with automatic memories enabled.
2. After the answer completes, background memory consolidation starts and spawns three workers.
3. Their memory-extraction commentary, command activity, and final replies appear in the original chat, interleaved with subsequent user messages.

Read-only inspection confirmed persistence, not just a rendering problem. Sanitized identities:

```text
Codex conversation A: Git question and answer; no worker reports
Codex logs: memory consolidation M spawned workers B, C, D
T3 projection_thread_messages: worker replies stored under A's T3 thread
T3 projection_thread_activities: same T3 thread, but data.threadId = B/C/D
Worker command cwd: ~/.codex/memories
```

The reports only entered conversation A's Codex transcript when I pasted the mixed UI transcript back to the assistant. This is the same user-visible failure as #4683 despite the earlier fix.

I checked this PR's actual `shouldSuppressForeignConversationNotification` at `509d07f`: all 12 combinations of the three observed worker IDs and four message/command event methods were suppressed. Root messages and `serverRequest/resolved` still passed. This was a filter-level check, not an installed end-to-end test of the PR; the incident does not establish whether each worker's `thread/started` was absent or missed.

The follow-up filter remains absent from upstream `main` at `4ca71463a` and nightly `0.0.39-nightly.20260905.1289`. This PR addresses the observed routing gap. No private transcripts or raw logs attached.

PR #9066, opened 2026-09-01
fix(desktop): honor local-only exposure in the WSL backend bind host

### LINKED ISSUE #9056: [Bug]: WSL backend binds 0.0.0.0 while "Network access" is off and the UI says "Limited to this machine"
## Summary

With **WSL backend** enabled and **WSL only = ON**, the app-managed backend listens on `0.0.0.0:3773` even though **Network access** is toggled **off**. The Connections panel states "Limited to this machine", which is not true of the actual bind.

What makes me think this is a bug rather than intended behaviour: on the same machine, a backend started via `npx t3@latest service install` (systemd user unit) bound `127.0.0.1:3773` correctly. Only the desktop-app-managed WSL backend binds all interfaces. (That systemd unit was stopped at the time of the checks below, so the process bound to `0.0.0.0` is definitively the app-managed backend, running out of `~/.t3/wsl-runtime/…`.)

## Steps to reproduce

1. Windows desktop app → **Settings → Connections**
2. **WSL backend** → pick a distro → choose **Use only WSL**
3. Leave **Network access** OFF ("Limited to this machine")
4. In the WSL distro: `ss -tlnp | grep 3773`

## Expected

`127.0.0.1:3773`, matching what the UI claims.

## Actual

```
LISTEN 0 511 0.0.0.0:3773 0.0.0.0:* users:(("node",pid=...))
```

and it answers on non-loopback addresses:

```
curl -o /dev/null -w '%{http_code}' http://100.x.x.x:3773/   -> 200   # tailnet IP
curl -o /dev/null -w '%{http_code}' http://172.x.x.x:3773/   -> 200   # WSL eth0
curl -o /dev/null -w '%{http_code}' http://127.0.0.1:3773/   -> 200
```

## Why this matters

"Limited to this machine" reads as a security property, and a user may reasonably rely on it when deciding whether to open a project containing sensitive material.

On a default WSL2 setup the exposure is narrower than it looks, since WSL's NAT address is not reachable from the LAN without a portproxy. But **anyone running Tailscale inside WSL has the backend reachable from every device on their tailnet**, with the toggle off and the UI saying otherwise. In my case the only thing preventing that was a restrictive tailnet ACL — containment came from the network, not from the app control that claims to provide it.

## Environment

- Windows 11 + WSL2, Ubuntu 24.04
- Desktop app installed via `winget install T3Tools.T3Code`, 2026-09-01
- CLI `t3@0.0.37` (used for the contrasting systemd-service test)
- WSL backend enabled, **WSL only** ON
- Network access **off**; T3 Connect **off**; Publish agent activity **off**

## Note on scope

I have not verified whether the UI/API layer on `:3773` authenticates. `/mcp` correctly returns `401` without the bearer token, but I did not test the rest of the surface. If the HTTP layer is fully authenticated then this is a "the UI states something untrue" bug rather than an exposure bug — either way, the toggle and the bind should agree.

dlowzzxx:
Fixes #9056.

## Problem

With the WSL backend enabled, `resolveWslStartConfig` hardcoded a `0.0.0.0` bind regardless of the resolved `DesktopServerExposure` mode. With **Network access** off, the Connections panel states "Limited to this machine", but the WSL backend actually answers on every distro interface — including `tailscale0` when Tailscale runs inside the distro, which exposes the backend to the entire tailnet with the toggle off (repro and curl evidence in the issue). The Windows primary already binds `127.0.0.1` in local-only mode via `resolveDesktopServerExposure`; only the WSL path diverged.

## Change

- Pass the primary's resolved exposure mode into `resolveWslStartConfig` as `networkAccessible`, read from `DesktopServerExposure.getState` at both call sites (the wsl-only primary and the `resolveWsl` secondary).
- In `network-accessible` mode the behavior is unchanged (`0.0.0.0`, for the reasons documented in the original comment — readiness probe and renderer reachability without depending on wslhost forwarding).
- In `local-only` mode, bind exactly the address the renderer is already told to use: the distro IP in NAT mode (reachable from Windows without depending on flaky wslhost forwarding, so the readiness-probe regression that motivated the wildcard bind does not come back) or loopback in mirrored mode. A wildcard bind is no longer possible while the UI claims "limited to this machine".

No UI changes, so no screenshots are attached.

## Tests

- `resolveWsl wildcards the WSL bind only when network access is enabled` — guards the unchanged `0.0.0.0` behavior in network-accessible mode.
- `resolveWsl binds the advertised distro IP when exposure is local-only` — local-only + NAT binds the advertised host instead of the wildcard.
- `resolveWsl binds loopback in mirrored mode when exposure is local-only` — mirrored mode keeps loopback.

Verification: focused suite `apps/desktop/src/backend/DesktopBackendConfiguration.test.ts` — 34 passed; `tsgo --noEmit` on `@t3tools/desktop` reports no errors in the touched files; `oxfmt --check` clean on both files.

---

> [!NOTE]
> **Medium Risk**
> Changes WSL listen addresses and desktop protocol proxy targets, which affect reachability and local-only network exposure guarantees on Windows/WSL.
> 
> **Overview**
> Fixes **local-only** server exposure for WSL-backed desktops: the WSL backend no longer always binds `0.0.0.0`, which could expose it on distro interfaces (e.g. Tailscale) while the UI still says “limited to this machine.”
> 
> **WSL bind host** now follows the primary’s exposure mode via `networkAccessible` from `DesktopServerExposure.getState`: **network-accessible** keeps `0.0.0.0`; **local-only** binds the same address the renderer uses (distro IP in NAT, `127.0.0.1` in mirrored mode).
> 
> Because local-only WSL primaries advertise a non-loopback URL, the custom **`app://` protocol** can follow the backend once it’s ready: `DesktopServerExposure` adds `backendProxyOrigin` / `noteResolvedBackendOrigin`, the backend pool records the resolved origin on `onReady`, packaged local-only bootstrap passes `targetOriginOverride` into `ElectronProtocol`, and the handler consults that override per request.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 749286ce90bd2a2027378f98d114986b1a07c0d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix WSL backend bind host to honor local-only exposure mode
> - WSL backend now binds to `0.0.0.0` only when exposure mode is network-accessible; in local-only mode it binds to the distro IP (NAT) or loopback (mirrored), preventing unintended exposure on other interfaces. See [DesktopBackendConfiguration.ts](https://github.com/pingdotgg/t3code/pull/9066/files#diff-8a9cd9d84f556b82ed8678110af5a99dea2cc1cfe1e6c6c46dc3a2e3bcb1f94c).
> - Adds `backendProxyOrigin` and `noteResolvedBackendOrigin` to `DesktopServerExposure` so the protocol proxy can follow a live backend-reported origin instead of a static loopback URL. See [DesktopServerExposure.ts](https://github.com/pingdotgg/t3code/pull/9066/files#diff-87d5808073ddb6ad6696185597726c0867bb1276e88997a57829cf9839e8abcb).
> - Extends `proxyRequest` in [ElectronProtocol.ts](https://github.com/pingdotgg/t3code/pull/9066/files#diff-25fd7328df6d1131991bc9498d73c423aef7bf0aee3e8b2c863c05b908090f78) with an optional `getTargetOrigin` callback; the protocol handler reads a dynamic target origin per request.
> - When a backend becomes ready, the backend pool calls `noteResolvedBackendOrigin` with the backend's `httpBaseUrl` before opening the main window. See [DesktopBackendPool.ts](https://github.com/pingdotgg/t3code/pull/9066/files#diff-9601fb20a0d89fceed1c143db112b96087a0b378c275501bdc9ee43d36b98bd1).
> - Risk: `resolveWsl` now requires a `networkAccessible` flag; any caller not passing it will fail to compile. The `DesktopServerExposure` interface change requires all stubs to implement `backendProxyOrigin` and `noteResolvedBackendOrigin`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 749286c.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WSL backend network binding to respect the selected server exposure mode.
  * Local-only connections now bind to the appropriate local or distro address instead of being exposed on all network interfaces.
  * Network-accessible connections continue to support wildcard binding.
  * Improved handling for mirrored-mode environments using loopback addresses.


dlowzzxx:
Pushed 8b4639934 addressing the review findings:

- **Protocol origin (Cursor, high):** the desktop protocol's proxy target now follows the backend's resolved renderer-visible URL. `DesktopServerExposure` records the URL the pool reports on readiness (`noteResolvedBackendOrigin` / `backendProxyOrigin`), and the protocol consults it per request for local-only packaged runs. A wsl-only primary serving from its distro IP is therefore reachable on the `app://` origin without depending on the wildcard bind; development and network-accessible runs keep today's exact target.
- Readiness-probe behavior is unchanged — the probe already used the WSL config's advertised URL.

New test: `follows the live target override once the backend resolves its URL`. Focused suites (protocol 7, backend configuration 34, pool, window, wsl backend) pass; `tsgo --noEmit` and `oxfmt --check` clean on touched files.

dlowzzxx:
Pushed 749286ce9 addressing both findings:

- **Ref.set order (Cursor high, Macroscope):** corrected to `Ref.set(resolvedBackendOriginRef, Option.some(origin))` — the write now lands on the ref.
- **Runtime bridging (Macroscope):** the registration input is now `targetOriginOverride?: Effect.Effect<URL>` and `ElectronProtocol` bridges it to a synchronous read itself via a context captured once in its layer — the escape lives at the Electron boundary, and DesktopApp just passes the effect for local-only packaged runs.

Protocol suite 7/7 including the override test (now driving the override through a Ref-backed Effect), backend configuration 34/34, tsgo and oxfmt clean on touched files.

PR #9069, opened 2026-09-01
fix(server): reject checkpoint reverts before mutating the workspace

### LINKED ISSUE #8958: [Bug]: Codex paginated thread revert fails after restoring workspace files
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Open an existing Codex thread that uses paginated history.
2. Choose an earlier user message.
3. Click the action to revert to that point.
4. Confirm the revert.

### Expected behavior

T3 Code should restore both the workspace and the Codex conversation to the selected point.

If Codex cannot rewind this type of thread, T3 Code should stop before changing any files and explain that the revert is unavailable.

### Actual behavior

T3 Code reports this error:

```text
Provider adapter request failed (codex) for thread/rollback: paginated threads do not support thread/rollback
```

The revert can fail after T3 Code has restored the workspace checkpoint. The files may move back to the selected point while the Codex conversation remains unchanged.

In `CheckpointReactor.handleRevertRequested`, `checkpointStore.restoreCheckpoint` runs before `providerService.rollbackConversation`. When Codex rejects `thread/rollback`, the operation stops after the file change.

### Impact

A failed revert can leave the workspace and conversation at different points. The user may continue the thread without realizing that the files have already changed.

### Version or commit

T3 Code Nightly 0.0.38-nightly.20260831.1240

### Environment

macOS 26.5.2, T3 Code desktop nightly, Codex CLI 0.151.0

### Logs or stack traces

```text
Provider adapter request failed (codex) for thread/rollback: paginated threads do not support thread/rollback
```

### Workaround

No in-app workaround found. Check the Git diff after the failure before continuing the thread.

dlowzzxx:
Fixes #8958.

## Problem

`CheckpointReactor.handleRevertRequested` restored the filesystem checkpoint **before** calling `providerService.rollbackConversation`. When the provider rejects the conversation rollback — Codex refuses `thread/rollback` for paginated threads with `paginated threads do not support thread/rollback` — the operation stopped after the file change. The workspace moved back to the target point while the Codex conversation stayed ahead of it, and the user gets no signal that the two have diverged.

## Change

The revert handler now fails before mutating anything:

1. **Pre-validate the checkpoint store** with the non-mutating `hasCheckpointRef` for turns > 0 (turn 0 keeps its `fallbackToHead` semantics and is excluded), so a read-model ref that outlived the checkpoint store fails gracefully before anything changes.
2. **Roll the provider conversation back first.** A provider rejection (e.g. the paginated-thread refusal above) now surfaces as the existing `checkpoint.revert.failed` activity while both the workspace and the conversation are untouched.
3. **Restore the filesystem checkpoint last**, followed by the unchanged workspace-index refresh.

## Tests

- New: `does not restore the workspace when the provider rejects the conversation rollback` — with a rejecting `rollbackConversation`, the revert records `checkpoint.revert.failed`, the provider rollback is attempted exactly once, and the workspace file is byte-identical to its pre-revert content.
- Existing revert tests continue to cover the success path and the claude/codex variants.

Verification: focused suite `apps/server/src/orchestration/Layers/CheckpointReactor.test.ts` — my new test passes; the pre-existing `executes provider revert...` test asserts `v2
` byte equality and fails identically on pristine `main` on a Windows checkout with `core.autocrlf=true` (`v2
`), so that one failure is a local line-ending artifact, not a behavior change. `oxfmt --check` clean on both files.

---

> [!NOTE]
> **Medium Risk**
> Changes the checkpoint revert failure path and workspace restore semantics; mistakes could leave workspace and conversation diverged or lose uncommitted edits around revert attempts.
> 
> **Overview**
> Fixes checkpoint reverts that could leave the **workspace reverted** while the **provider conversation** still advanced when `rollbackConversation` fails (e.g. Codex paginated threads rejecting `thread/rollback`).
> 
> `CheckpointReactor` now captures a per-thread **`revert-base`** git ref (`checkpointRefForRevertBase`) immediately before applying a revert. If provider rollback fails after the target checkpoint was restored, it **restores that snapshot**, refreshes the workspace index, records `checkpoint.revert.failed` with the provider error, and exits without completing the revert. On success, the temporary `revert-base` ref is deleted along with stale turn checkpoints.
> 
> Adds a reactor test that mocks a rejecting `rollbackConversation` and asserts the workspace content matches the pre-revert state (with CRLF-normalized comparison).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef92e38778727d37207604f6d83be5af611a5974. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Capture pre-revert checkpoint before attempting `CheckpointReactor` conversation rollback
> - On `thread.checkpoint.revert`, the reactor now snapshots the working tree via a new `checkpointRefForRevertBase` ref before calling `rollbackConversation`.
> - If the provider rollback fails with `ProviderServiceError`, the reactor restores that snapshot (no fallback to HEAD), refreshes workspace entries, appends a `checkpoint.revert.failed` activity, and returns early.
> - On successful revert the temporary `revert-base` ref is included in stale checkpoint cleanup.
> - Behavioral Change: revert no longer leaves a partially mutated workspace when `rollbackConversation` fails; callers see a `checkpoint.revert.failed` activity and the pre-revert file contents.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ef92e38.</sup>
> 
>

### DIFF
--- a/apps/server/src/checkpointing/Utils.ts
+++ b/apps/server/src/checkpointing/Utils.ts
@@ -9,6 +9,18 @@ export function checkpointRefForThreadTurn(threadId: ThreadId, turnCount: number
   );
 }
 
+/**
+ * Working-tree snapshot taken right before a revert so a failed provider
+ * rollback can be compensated back to the exact pre-revert state — prior
+ * checkpoints can miss uncommitted local edits. One per thread; the capture
+ * overwrites it on each revert attempt via `git update-ref`.
+ */
+export function checkpointRefForRevertBase(threadId: ThreadId): CheckpointRef {
+  return CheckpointRef.make(
+    `${CHECKPOINT_REFS_PREFIX}/${Encoding.encodeBase64Url(threadId)}/revert-base`,
+  );
+}
+
 export function resolveThreadWorkspaceCwd(input: {
   readonly thread: {
     readonly projectId: ProjectId;
--- a/apps/server/src/orchestration/Layers/CheckpointReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/CheckpointReactor.test.ts
@@ -55,6 +55,7 @@ import {
   ProviderService,
   type ProviderServiceShape,
 } from "../../provider/Services/ProviderService.ts";
+import { ProviderAdapterRequestError, type ProviderServiceError } from "../../provider/Errors.ts";
 import { checkpointRefForThreadTurn } from "../../checkpointing/Utils.ts";
 import { ServerConfig } from "../../config.ts";
 import * as WorkspaceEntries from "../../workspace/WorkspaceEntries.ts";
@@ -85,7 +86,10 @@ function createProviderServiceHarness(
   const now = "2026-01-01T00:00:00.000Z";
   const runtimeEventPubSub = Effect.runSync(PubSub.unbounded<ProviderRuntimeEvent>());
   const rollbackConversation = vi.fn(
-    (_input: { readonly threadId: ThreadId; readonly numTurns: number }) => Effect.void,
+    (_input: {
+      readonly threadId: ThreadId;
+      readonly numTurns: number;
+    }): Effect.Effect<void, ProviderServiceError> => Effect.void,
   );
 
   const unsupported = <A>() =>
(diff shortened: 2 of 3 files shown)


dlowzzxx:
Pushed ef92e3877 addressing both findings:

- **Compensating recovery (Macroscope high, Cursor medium):** the handler now captures the pre-revert worktree as a per-thread `revert-base` checkpoint before touching anything. If `rollbackConversation` fails after the file restore, the revert-base snapshot is restored and the workspace index refreshed, so both the workspace and the conversation end in their original state and a retry starts from a clean slate. The revert-base ref is deleted on success and overwritten (`git update-ref`) on each attempt, so no refs accumulate. If the restore itself fails, nothing else has run yet — no divergence.
- **Test error type (Macroscope):** the regression test now fails with the real `ProviderAdapterRequestError` and the harness mock is typed on the `ProviderServiceError` channel — no cast.
- The `hasCheckpointRef` pre-check from the previous commit is gone; the revert-base capture supersedes it.

Focused suite: 16/17 pass; the one failure (`executes provider revert...`) is the pre-existing `core.autocrlf` byte-equality artifact that also fails on pristine `main` on Windows checkouts.

dlowzzxx:
On the staged-edits finding: that is a property of the underlying store rather than of this compensation — `captureCheckpoint` records a tree commit, so `restoreCheckpoint` never reconstructs the staged/unstaged split, and the ordinary successful revert has had the same behavior all along. The compensation is therefore no worse than the feature's normal outcome. If index fidelity matters, that is a store-level enhancement to `captureCheckpoint`/`restoreCheckpoint`, which I kept out of this fix to stay scoped — happy to take it up separately if wanted.

shivamhwp:

The compensating restore can fail while leaving the workspace at the selected checkpoint. For example, if `.git/index.lock` appears during the provider request and Codex rejects rollback, recovery throws, but `Effect.catch(() => Effect.void)` discards that error. The activity then reports only the Codex rejection, without telling the user that their files are still reverted. A `false` restore result is also ignored.

Please check the recovery result and report an explicit workspace-recovery failure alongside the provider error. Preserve the recovery snapshot until that failure is resolved; a retry currently overwrites the same `revert-base` ref with the already-reverted workspace. The new regression should include distinct uncommitted and untracked files plus a failed recovery, and wait on the reactor drain or receipt instead of polling.

juliusmarminge:
Superseded by #11338, which makes Codex paginated rewind succeed via `thread/revert` instead of failing closed before workspace mutation. Closing this pre-validate/reject path for #8958 as wash leftover hygiene — reopen if the fail-before-mutate guard is still wanted independently.

PR #9075, opened 2026-09-01
fix(provider): drop removed custom models from the model picker

StiensWout:
Removing a custom model in Settings did not remove it from the model picker. The row stayed in the composer picker and in every settings picker, and it survived server restarts.

The server's provider aggregator merges each refreshed snapshot with the previous one and, for every driver except OpenCode, keeps any model that the new snapshot no longer reports. Custom models are built from settings on every probe, so the removed slug was missing from the new snapshot and got re-added from the old one on every refresh, then persisted to the boot cache. The web option builder also took custom rows straight from the server snapshot instead of settings.

Custom rows are owned by settings and every snapshot already carries the full current list, so the aggregator and boot-cache merges now never retain previous custom rows. The web option builders only take built-ins from the snapshot and rebuild custom rows from settings, so a removal disappears immediately instead of waiting for the next probe. Mobile reads the server list directly and is fixed by the server change.

Focused tests cover the aggregator merge, the boot-cache hydrate, and the web option list.

Made with Claude Fable 5.1 in T3 Code.

---

> [!NOTE]
> **Medium Risk**
> Changes provider snapshot merging and model-picker assembly, which could affect model lists at runtime, but scope is limited to custom-model retention with targeted tests.
> 
> **Overview**
> Fixes **removed custom models sticking around** in the composer and settings pickers (and across restarts) by treating custom model rows as **settings-owned** instead of mergeable snapshot history.
> 
> On the **server**, `mergeProviderSnapshot` and boot **cache hydration** no longer carry forward previous or cached `isCustom` models when the current settings-derived snapshot omits them; **built-in** models still get the existing “retain on empty/partial refresh” behavior. On **web**, `getAppModelOptions` / `getAppModelOptionsForInstance` take **built-ins only** from the provider snapshot and **rebuild custom options from settings**, so a deletion shows up immediately even if the server list is stale.
> 
> Adds regression tests for registry merge, cache hydrate, and web option lists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f05db6c334f5dd10bd71ae9dc1128ea1108eef1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Drop removed custom models from provider snapshots, status cache, and model picker
> - Filters custom models out of the retained set in `mergeProviderModels` so a refreshed provider snapshot no longer resurrects custom models that were removed from settings
> - Excludes custom cached models during `hydrateCachedProvider` hydration so the status cache does not bring back custom models absent from the settings-derived fallback snapshot
> - Ignores server-reported custom models in `getAppModelOptions` and `getAppModelOptionsForInstance`, rebuilding custom rows solely from user settings so removals take effect immediately
> - Risk: any code path that relied on server-reported or cached custom models surviving a refresh will no longer see them; check `mergeProviderModels` in [ProviderRegistry.ts](https://github.com/pingdotgg/t3code/pull/9075/files#diff-5c33694fd7889294ecb21a88beafee2debe92f779baab4629651d4c8be8a9a97) and `hydrateCachedProvider` in [providerStatusCache.ts](https://github.com/pingdotgg/t3code/pull/9075/files#diff-6de08e2cd95df6ffd1cc93a059448e8a80f0392091e10e0d7aabd9b6b01dbb07) for downstream consumers
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6f05db6.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/ProviderRegistry.test.ts
+++ b/apps/server/src/provider/Layers/ProviderRegistry.test.ts
@@ -620,6 +620,44 @@ it.layer(Layer.mergeAll(NodeServices.layer, ServerSettingsModule.layerTest(), Te
         );
       });
 
+      it("drops custom models the refreshed snapshot no longer carries", () => {
+        const previousProvider = {
+          instanceId: ProviderInstanceId.make("claudeAgent"),
+          driver: ProviderDriverKind.make("claudeAgent"),
+          status: "ready",
+          enabled: true,
+          installed: true,
+          auth: { status: "authenticated" },
+          checkedAt: "2026-04-14T00:00:00.000Z",
+          version: "2.1.0",
+          models: [
+            {
+              slug: "claude-sonnet-4-6",
+              name: "Sonnet 4.6",
+              isCustom: false,
+              capabilities: null,
+            },
+            {
+              slug: "removed-custom",
+              name: "removed-custom",
+              isCustom: true,
+              capabilities: null,
+            },
+          ],
+          slashCommands: [],
+          skills: [],
+        } as const satisfies ServerProvider;
+        const refreshedProvider = {
+          ...previousProvider,
+          checkedAt: "2026-04-14T00:01:00.000Z",
+          models: [previousProvider.models[0]],
+        } satisfies ServerProvider;
+
+        assert.deepStrictEqual(mergeProviderSnapshot(previousProvider, refreshedProvider).models, [
+          ...refreshedProvider.models,
+        ]);
+      });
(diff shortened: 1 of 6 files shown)


MatthewFeroz:
pr-9075-step1-provider-settings
pr-9075-step2-picker-before-removal
pr-9075-step3-picker-after-removal

just verified looks good to me

PR #9082, opened 2026-09-01
fix(desktop): stop preview element capture from hanging

### LINKED ISSUE #9026: [Bug]: Sending a picked preview element hangs on "Capturing..." forever and locks the composer
_Note: AI-written, posted on Karsten's instruction._

### What happens

Picking an element in the browser preview and then submitting the message never completes. The send button switches to **Capturing...** and stays there forever. No attachment is added, no message is sent, no error is shown. The composer stays locked until the app is restarted or the thread is reloaded.

### Steps to reproduce

1. Open a browser preview in a thread.
2. Use the element picker and select an element (in my case a flex row of logos, label shown was `div.css-view-g5y9jx.flex-row`).
3. Type some text in the composer (I typed `test`).
4. Press Enter to send.

Expected: the picked element is captured, attached to the message, and the message is sent.

Actual: the button shows `Capturing...` indefinitely. Nothing is attached, nothing is sent.

It does not happen every single time, but it happens often, and once it happens that composer is stuck.

### What the local traces show

From `~/.t3/userdata/logs/desktop.trace.ndjson` (and the rotated files), for six element-pick sessions in one session of use:

- 6x `PreviewManager.pickElement` / `desktop.ipc.preview.pickElement`
- only 1x `PreviewManager.settlePickElement`
- 5x `PreviewManager.cancelPickElement` + `PreviewManager.cancelPickSession`

All of those spans exit with `{"_tag":"Success"}`. There is no span at all for the capture/screenshot step that the `Capturing...` label refers to, and no failure is recorded anywhere for it, so the stuck state produces no error in the trace log. That suggests the capture promise in the renderer never settles rather than rejecting, so there is nothing to time out or surface to the user.

Unrelated but in the same log window, preview automation shows these failures, in case they point at the same webContents/CDP session going bad:

```
PreviewManager.performAutomationEvaluate  PreviewAutomationEvaluationError: Preview JavaScript evaluation failed in tab [... "tab_5"]
PreviewManager.performAutomationEvaluate  PreviewOperationError: Desktop preview operation failed: evaluate.Runtime.evaluate (tab [... "tab_3"], WebContents 6)
PreviewManager.resolveClickPoint          PreviewAutomationInvalidSelectorError: Preview automation click rejected locator (20 characters) in tab [... "tab_4"]
```

### Suggested handling

Even if the root cause is hard to hit, `Capturing...` should have a timeout: fail the capture after a few seconds, show an error, and unlock the composer so the message can be sent without the attachment. Right now the whole composer is dead until a restart.

### Environment

- T3 Code (Alpha) desktop, version 0.0.37
- Windows 11 Home, 10.0.26200
- Local server runtime on 127.0.0.1

Happy to attach the raw trace lines or a screen recording if that helps.

Gigioxx:
A selected preview element could leave capture waiting forever when react-grab context enrichment never settled. Because annotation IPC only runs after every selected element finishes, the editor stayed locked on Capturing….

This bounds enrichment to five seconds. On timeout, capture retains safe DOM context (URL, tag, and truncated HTML) and continues the existing screenshot and send path. The picker itself remains unbounded while the user is annotating.

Verified with:

- 92 focused desktop preview tests
- Desktop typecheck
- Targeted lint and format checks
- Desktop production bundle

Fixes #9026

Made with GPT-5.6 Sol in the Codex harness through T3 Code.

---

> [!NOTE]
> **Low Risk**
> Localized desktop preview preload behavior with a safe DOM fallback; no auth, data, or API surface changes beyond fixing a hang.
> 
> **Overview**
> **Preview element capture no longer blocks indefinitely** when `react-grab`’s `getElementContext` never resolves—a case that left the annotation editor stuck on “Capturing…” because screenshot/IPC only ran after every selected element finished enriching.
> 
> `captureElement` now **races** context enrichment against a **5s timeout** (with rejections treated as missing context), **clears the timer** in `finally`, and on timeout/failure still returns a `PickedElementPayload` using page URL/title, tag name, and **truncated `outerHTML`** instead of react-grab selector/component/stack/styles. **`captureElement` is exported** for unit testing, and diagnostics allow `setTimeout` in this preload.
> 
> A new **`PickPreload.test.ts`** asserts the timeout path returns the expected fallback fields when `getElementContext` never settles.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fbd40ed800763d9568a9be8c049cbc948545f9a4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add 5s timeout to `captureElement` in desktop preview to prevent hanging
> - Wraps `await getElementContext(element)` in `Promise.race` against a 5,000 ms timeout in [PickPreload.ts](https://github.com/pingdotgg/t3code/pull/9082/files#diff-a19a7f6aa07147ba64b95dfc017ebf3427a82f067b93b614488e7485689df0b1), resolving to `null` on timeout or rejection
> - Adds fallback values for optional context fields: `htmlPreview` defaults to `element.outerHTML` (truncated to 4,000 chars), `selector` and `componentName` default to `null`, `styles` defaults to empty string
> - Exports `captureElement` from the module
> - Adds test in [PickPreload.test.ts](https://github.com/pingdotgg/t3code/pull/9082/files#diff-02814888fbfd2bbefc3cc32f1a10e6c04f3d85346c11b72acaaa577532002a8f) verifying the promise resolves after 5s when `getElementContext` never settles
> - Risk: `captureElement` now returns a minimal fallback payload instead of hanging; consumers expecting the full context object may receive `null` for `selector`, `componentName`, or `stack`
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fbd40ed.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/preview/PickPreload.test.ts
+++ b/apps/desktop/src/preview/PickPreload.test.ts
@@ -1,0 +1,47 @@
+import { afterEach, expect, it, vi } from "vite-plus/test";
+
+const { getElementContext, ipcRenderer } = vi.hoisted(() => ({
+  getElementContext: vi.fn(),
+  ipcRenderer: { on: vi.fn(), send: vi.fn() },
+}));
+
+vi.mock("electron", () => ({ ipcRenderer }));
+vi.mock("react-grab/primitives", () => ({ getElementContext }));
+
+vi.stubGlobal("window", { addEventListener: vi.fn() });
+vi.stubGlobal("location", { href: "https://example.com/dashboard" });
+vi.stubGlobal("document", { title: "Dashboard" });
+
+const { captureElement } = await import("./PickPreload.ts");
+
+afterEach(() => {
+  vi.useRealTimers();
+  getElementContext.mockReset();
+});
+
+it("falls back when react-grab element context never settles", async () => {
+  vi.useFakeTimers();
+  getElementContext.mockReturnValue(new Promise(() => undefined));
+
+  const element = {
+    tagName: "DIV",
+    outerHTML: '<div class="css-view-g5y9jx flex-row">Logos</div>',
+  } as Element;
+  const capture = captureElement(element);
+
+  await vi.advanceTimersByTimeAsync(5_000);
+  const result = await Promise.race([capture, Promise.resolve("still pending")]);
+
+  expect(result).toEqual(
+    expect.objectContaining({
+      pageUrl: "https://example.com/dashboard",
+      tagName: "div",
+      selector: null,
+      htmlPreview: element.outerHTML,
+      componentName: null,
+      source: null,
+      stack: [],
+      styles: "",
+    }),
+  );
+});
--- a/apps/desktop/src/preview/PickPreload.ts
+++ b/apps/desktop/src/preview/PickPreload.ts
@@ -1,4 +1,4 @@
-// @effect-diagnostics globalDate:off - This isolated Electron preload does not run inside an Effect runtime.
+// @effect-diagnostics globalDate:off globalTimers:off - This isolated Electron preload does not run inside an Effect runtime.
 import { ipcRenderer } from "electron";
 import { getElementContext } from "react-grab/primitives";
 import type {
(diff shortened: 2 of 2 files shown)


t3dotgg:
> [!NOTE]
>

Closing as superseded by merged [#9127](https://github.com/pingdotgg/t3code/pull/9127). Main now bounds element-context capture and the main-process screenshot wait, keeps a DOM-only fallback, and clears the renderer's pick state after failure. That covers this PR's stalled context capture and the later steps that could still hold the composer open.

PR #8836, opened 2026-08-31
fix(server): allow local tracing to be disabled

Michel-Liao:
## What Changed

- Treat `T3CODE_TRACE_MIN_LEVEL=None` as the supported off switch for local trace persistence.
- Use a no-op sink for server and browser-submitted spans when disabled, and avoid creating the trace directory.
- Report `localTracingEnabled: false` to clients while keeping OTLP proxy/metrics configuration independent.
- Document the setting and add focused config, observability, and WebSocket coverage.

## Why

Local Effect spans are written and rotated for the entire server lifetime, but operators had no supported way to stop those disk writes. Reusing the existing minimum-level setting keeps the default unchanged while providing a documented, zero-write mode.

Fixes #8764.

## Verification

- Observability and CLI config tests — 10 passed
- Focused WebSocket config test
- Server typecheck
- Focused lint and formatting checks

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] No UI changes

Built with GPT-5.6 Sol in T3 Code through the Codex harness.

---

> [!NOTE]
> **Low Risk**
> Opt-in env change with unchanged defaults; scope is local trace I/O and client observability flags, not auth or data paths.
> 
> **Overview**
> Operators can turn off local NDJSON trace persistence by setting **`T3CODE_TRACE_MIN_LEVEL=None`**, without changing the default **`Info`** behavior.
> 
> **Startup and runtime:** Config resolution no longer creates the trace file’s parent directory when tracing is off. The observability layer swaps the rotating file sink for a **no-op sink**, so server spans and browser-submitted spans do not create or write `serverTracePath` while OTLP URLs and metrics stay independent.
> 
> **Clients and docs:** WebSocket `getConfig` now sets **`localTracingEnabled`** from `traceMinLevel !== "None"` instead of always `true`. Maintainer docs describe the off switch and that server OTLP span export is suppressed at minimum level `None`, while configured OTLP browser forwarding and metrics remain available.
> 
> Tests cover config directory behavior, observability with spans recorded, and the WebSocket observability snapshot.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3cebe57989d332d9990e6a217591fc0008807d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Disable local tracing when `T3CODE_TRACE_MIN_LEVEL=None`
[...]

PR #9115, opened 2026-09-01
fix(server): allow disabling the span exporter

### LINKED ISSUE #8764: [Bug]: Effect span exporter writes to disk unconditionally with no way to disable it
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Start the server with `t3 serve`.
2. Run an interactive session that produces terminal output.
3. Inspect `~/.t3/userdata/logs/server.trace.ndjson` and its rotated siblings.

### Expected behavior

Span export to disk is off by default in a long-running server, or can be disabled through a documented flag or environment variable.

### Actual behavior

The server writes Effect tracing spans to `~/.t3/userdata/logs/server.trace.ndjson` for the whole life of the process. The file rotates through 11 files of 10 MB, a new one roughly every 12 minutes under ordinary use, indefinitely.

Records are marked `"sampled":true` with no sampling reduction, and spans are emitted per terminal event, so the volume tracks terminal output rather than anything the operator configures.

There is no supported way to turn it off:

- `T3CODE_TRACE_FILE` changes the path but does not disable the exporter, and the directory creation for it runs unconditionally.
- `T3CODE_LOG_LEVEL` defaults to `Info` and governs the logger, not this file.
- `t3 serve --help` exposes no tracing or verbosity control.

### Impact

### Version or commit

`0.0.34` and `0.0.36`

### Environment

Platform-independent; reproduced on Linux with the published npm releases.

### Logs or stack traces

```shell
{"type":"effect-span","name":"terminal.queuePersist","sampled":true,...}
{"type":"effect-span","name":"terminal.drainProcessEvents","sampled":true,...}
```

### Workaround

Point `T3CODE_TRACE_FILE` at a path on a `tmpfs`, so the writes never reach persistent storage.

### Related

- #7991 overlaps: the spans above come from the terminal persistence path it describes. That issue is about the terminal history file growing and being rewritten; this one is about the span exporter being unconditional, which stands regardless of how history persistence is bounded.
- #8345 is the same theme of unbounded disk growth from a different cause.

lnieuwenhuis:
The server writes Effect tracing spans to `<home>/userdata/logs/server.trace.ndjson` for the whole life of the process, rotating through 11 files of 10 MB, and `T3CODE_TRACE_FILE` can only relocate it. There was no supported way to turn the exporter off.

This adds `T3CODE_TRACE_ENABLED` (default `true`). When set to `false`, no trace sink or tracer is constructed and no trace file is written; the trace directory is no longer created for a custom trace path. The browser-trace collector stays available as a no-op so the relay route keeps working. Documented in `docs/operations/observability.md`.

Tests cover both the config layer (no trace directory creation when disabled) and the observability layer (no trace file when disabled, file present by default).

Closes #8764

---

> [!NOTE]
> **Low Risk**
> Observability-only change with a safe default; pairing already disables tracing explicitly, and the browser trace endpoint remains wired via a no-op collector.
> 
> **Overview**
> Adds **`T3CODE_TRACE_ENABLED`** (default **`true`**) and a **`traceEnabled`** field on **`ServerConfig`** so local NDJSON span export can be turned off without only relocating the file via **`T3CODE_TRACE_FILE`**.
> 
> When tracing is disabled, **`resolveServerConfig`** no longer creates the trace file’s parent directory, and **`ObservabilityLive`** skips the file sink, local tracer, and OTLP trace delegate while keeping **`BrowserTraceCollector`** as a no-op so the browser-trace route still works. Pairing flows explicitly set **`traceEnabled: false`**.
> 
> Tests cover config resolution and the observability layer for directory creation behavior; **`docs/operations/observability.md`** documents the new env var.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5261360871fbba7d743658959d2ec5eb3086e727. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add `traceEnabled` config flag to disable span exporter via `T3CODE_TRACE_ENABLED`
> - Introduces a new boolean `traceEnabled` field on `ServerConfig`, sourced from the `T3CODE_TRACE_ENABLED` env var (default `true`), so operators can disable local trace file creation.
> - When `traceEnabled` is false, [Observability.ts](https://github.com/pingdotgg/t3code/pull/9115/files#diff-7a5c18b26b28e7a6496742e45b9391ca8004b50688bea5b5d877edc282a25f32) skips tracer/sink initialization and leaves the default no-op `Tracer` in place; `BrowserTraceCollector` stays available but discards records.
> - [config.ts](https://github.com/pingdotgg/t3code/pull/9115/files#diff-16c068d320ce481e506eae4c06ff006a27585a4833a2d7d3dd1b584e5b8d8fdb) makes trace directory creation conditional on `traceEnabled`, and [pair.ts](https://github.com/pingdotgg/t3code/pull/9115/files#diff-7326618ccfd09900176b612acc8fd16fbeb01bddd2ba054af9dff737684e075a) hard-disables tracing for pair-mode server configs.
> - Risk: callers that previously relied on `BrowserTraceCollector` always writing to disk will silently lose spans when `T3CODE_TRACE_ENABLED=false`; verify no downstream consumers depend on the trace file existing in disabled mode.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5261360.</sup>
> 
>

### DIFF
--- a/apps/server/src/bin.test.ts
+++ b/apps/server/src/bin.test.ts
@@ -91,6 +91,7 @@ const makeCliTestServerConfig = (baseDir: string) =>
     const derivedPaths = yield* ServerConfig.deriveServerPaths(baseDir, undefined);
     return {
       logLevel: "Info",
+      traceEnabled: true,
       traceMinLevel: "Info",
       traceTimingEnabled: true,
       traceBatchWindowMs: 200,
--- a/apps/server/src/cli/config.test.ts
+++ b/apps/server/src/cli/config.test.ts
@@ -41,6 +41,7 @@ const makeDesktopBootstrap = (
 
 it.layer(NodeServices.layer)("cli config resolution", (it) => {
   const defaultObservabilityConfig = {
+    traceEnabled: true,
     traceMinLevel: "Info",
     traceTimingEnabled: true,
     traceBatchWindowMs: 1_000,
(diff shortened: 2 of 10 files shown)


juliusmarminge:
Closing as a duplicate of #8836, which addresses #8764 using the existing T3CODE_TRACE_MIN_LEVEL=None setting, skips local trace writes and directory creation, and reports the disabled state to clients. Please continue review there.

PR #9116, opened 2026-09-01
fix(web): clean up worktrees when deleting archived threads

### LINKED ISSUE #9085: Deleting an archived thread silently orphans its git worktree
## Summary

`deleteThread` only runs worktree cleanup for threads present in the main thread store. Archived threads are excluded from that store, so deleting one from **Settings → Archived** deletes the thread and leaves its git worktree on disk permanently, with no prompt.

## Why archived threads miss the cleanup

`deleteThread` resolves its target via `resolveThreadTarget` → `readThreadShell` (`apps/web/src/hooks/useThreadActions.ts:204`). On a miss it takes an early-return branch at `useThreadActions.ts:285-295`:

```
// Thread not in main store (e.g. archived thread) — dispatch delete directly.
```

That returns before the worktree block at `useThreadActions.ts:318-430` (`getOrphanedWorktreePathForThread` → confirm dialog → `removeWorktree`).

The main store is hydrated from `getShellSnapshot`, whose thread query filters archived rows — `listActiveThreadRows` in `apps/server/src/orchestration/Layers/ProjectionSnapshotQuery.ts:481`:

```sql
WHERE deleted_at IS NULL
  AND archived_at IS NULL
```

So once the client has re-hydrated, an archived thread is by construction absent from the store the delete path reads. The archived panel is fed by a separate `archivedShellSnapshot` (`apps/web/src/lib/archivedThreadsState.ts`), and its delete action calls `confirmAndDeleteThread` (`apps/web/src/components/settings/SettingsPanels.tsx:2641`) → `deleteThread`.

## Repro

1. Create a thread with **New worktree** — worktree lands in `~/.t3/worktrees/<repo>/<branch>`.
2. Archive the thread.
3. Restart the app (or otherwise let the shell snapshot re-hydrate).
4. **Settings → Archived** → delete the thread, accepting the generic `Delete thread "…"?` prompt.

**Expected:** same as deleting an unarchived thread — `This thread is the only one linked to this worktree… Delete the worktree too?`

**Actual:** no worktree prompt. The thread is deleted; the worktree directory and its git admin entry remain.

Step 3 matters. Deleting immediately after archiving *within the same session* can still hit the normal path, because `thread.archived` does not evict the thread from the client store (`apps/web/src/orchestrationEventEffects.ts:49` sets `clearDeletedThread: false`). That's likely why this is easy to miss in manual testing.

## Impact

`apps/web/src/hooks/useThreadActions.ts:425` is the only call site in the app that removes a worktree, so nothing else reclaims these. Archive-then-delete is a natural workflow, and worktrees accumulate silently.

The cost isn't trivial: each is a full working tree, and if the project has a setup script it runs inside every new worktree (`apps/server/src/git/GitManager.ts:2091`) — so a `node_modules` per worktree. `git worktree prune` doesn't help; it only drops admin entries for directories that are already gone.

## Verification

I exercised the real `deleteThread` with mocked dependencies (vitest, `apps/web`):

- active thread with a worktree → confirm dialog shown once, `removeWorktree` called once
- `readThreadShell` returning `null` (the archived case) → delete dispatched, confirm never shown, `removeWorktree` never called

`readThreadShell` was the only difference between the two cases.

Observed at commit `c17d02cf`.

lnieuwenhuis:
`deleteThread` resolved its target only from the main shell store, which excludes archived threads. On a miss it dispatched the delete directly and skipped the worktree-cleanup path, so deleting a thread from Settings → Archived silently left its git worktree on disk with no prompt. That path is the only place in the app that removes a worktree, so nothing ever reclaimed them.

Delete-target resolution now falls back to a one-shot fetch of the environment's archived snapshot, rather than reading the cached atom, which is only warm while the archived panel is open. The orphan check also folds in archived siblings so a worktree shared with an archived thread isn't removed out from under it, and the confirmation dialog shows the archived thread's real title. Both fetches fail soft: an offline environment does not block the thread delete, but a failed or timed-out fresh ownership lookup skips worktree cleanup. A previously cached empty archived snapshot cannot authorize removal.

The resolver and the orphan-check step are extracted as pure functions with tests covering the active hit, the archived fallback, the not-found case, and fetch failure. No UI changes; the existing worktree confirmation dialog now simply appears for archived threads too.

Mobile has no worktree cleanup for any thread delete today, so this is scoped to web.

Closes #9085

---

> [!NOTE]
> **Medium Risk**
> Changes delete and worktree-removal decision logic with soft-fail fetch paths that can skip cleanup when archived data is unavailable, though deletes still proceed.
> 
> **Overview**
> **Deleting archived threads** no longer skips git worktree cleanup. `deleteThread` used to resolve targets only from the active shell store, so deletes from Settings → Archived went straight to the mutation and never ran the worktree prompt—leaving orphaned worktrees on disk (#9085).
> 
> The delete path now uses **`resolveThreadDeleteTarget`**, which tries the active store first, then **`fetchArchivedThreadShells`** (a bounded one-shot archived snapshot read with a **5s** `executeAtomQuery` timeout). Fetch failures or misses **fail soft**: delete still dispatches for idempotency, matching prior offline behavior. When resolution succeeds, **`resolveOrphanedWorktreePathForDelete`** includes **archived siblings** so a shared worktree is not removed under another archived thread, and **`confirmAndDeleteThread`** shows the archived thread’s real title.
> 
> Pure helpers and **`timeoutMs`** on **`executeAtomQuery`** are covered by new tests; no new UI— the existing worktree confirmation dialog now applies to archived deletes too.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01dbd5aaae8f6db53fa490128b39352fe1e764c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Clean up worktrees when deleting archived threads
> - Deleting a thread now resolves the target across both active and archived stores, so archived threads can be identified and deleted with correct confirmation titles.
> - Worktree orphan checks during deletion now consider archived siblings via `resolveOrphanedWorktreePathForDelete`, avoiding unnecessary fetches when the target has no worktree.
> - New `fetchArchivedThreadShells` performs a bounded (5s) archived-thread fetch with soft fallback to cached snapshot, returning `null` when neither exists.
> - `executeAtomQuery` gains an optional `timeoutMs`; on timeout it dies and returns a defect `Failure` instead of hanging.
> - Behavioral Change: when archived fetch fails or is offline, deletion proceeds without blocking (graceful soft-fail) rather than erroring; this is handled in [useThreadActions.ts](https://github.com/pingdotgg/t3code/pull/9116/files#diff-7d3463056176ee1e173c0ae885b4305a70a7cc854fee7040d9b9a5a28dfbda50) via `resolveDeletableThreadTarget`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 01dbd5a.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Thread deletion now works reliably for both active and archived threads.
  - Archived thread titles are shown in deletion confirmations when available.
  - Worktree cleanup now considers archived sibling threads, preventing removal of shared worktrees.
  - Cleanup is skipped when archived-thread information cannot be retrieved, avoiding unsafe deletions.
  - Archived-thread lookups now time out gracefully instead of waiting indefinitely.


lnieuwenhuis:
Reviewed the archived-target lookup suggestion. For worktree deletion there are already only two archived queries: one for the confirmation title and one after confirmation for fresh target metadata. The orphan check reuses that second result through resolvedArchivedThreads. Passing the pre-confirmation resolution while retaining the required fresh ownership query would still make two requests and could use stale target/session metadata. Keeping the fresh resolution; avoiding a query for targets without worktrees would be a separate minor optimization.

juliusmarminge:
Superseded by #11598, which replaces the cleanup-on-delete approach with opt-in automatic storage cleanup per machine/project.

Issue #7756, opened 2026-08-21
[Bug]: Browser pairing returns authenticated but omits the t3_session   cookie


Nintorac:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

### Steps to reproduce

1. Start a self-hosted T3 Code server with remote authentication enabled.

2. Use a fresh browser profile, clear the site's cookies, or log out of an existing browser session.

3. Create a one-time pairing credential:

   ```shell
   t3 auth pairing create --base-dir <base-dir>
   ```

4. Open the server's pairing page in the browser and submit the credential.

5. Inspect the response to:

   ```text
   POST /api/auth/browser-session
   ```

6. Request:

   ```text
   GET /api/auth/session
   ```

I reproduced this through an HTTPS reverse proxy and by sending the pairing request directly to `127.0.0.1:3773`. The direct request rules out the reverse proxy stripping the header.

### Expected behavior

`POST /api/auth/browser-session` should return the authenticated response and set an HttpOnly `t3_session` cookie.

The browser should leave the pairing screen, and later HTTP and WebSocket requests should authenticate with that cookie.

### Actual behavior

The pairing credential is consumed and the endpoint returns HTTP 200:

```json
{
  "authenticated": true,
  "scopes": [
    "orchestration:read",
    "orchestration:operate",
    "terminal:operate",
    "review:write",
    "relay:read"
  ],
  "sessionMethod": "browser-session-cookie"
}
```

The response has no `Set-Cookie` header.

The browser therefore remains on "Pair with this environment". A subsequent `GET /api/auth/session` returns:

```json
{
  "authenticated": false,
  "auth": {
    "policy": "remote-reachable",
    "bootstrapMethods": ["one-time-token"],
    "sessionMethods": [
      "browser-session-cookie",
      "bearer-access-token",
      "dpop-access-token"
    ],
    "sessionCookieName": "t3_session"
  }
}
```

WebSocket upgrades fail with HTTP 401 and `ServerAuthMissingCredentialError`.

The server does create a session record, but the browser never receives its session token. Repeating the test in a fresh Firefox container produces the same result.

The affected handler creates a cookie and registers a pre-response handler with `HttpServerResponse.mergeCookies`, but no cookie reaches the wire:

https://github.com/pingdotgg/t3code/blob/v0.0.34-nightly.20260819.1130/apps/server/src/auth/http.ts

[...]

PR #9118, opened 2026-09-01
fix(server): bundle Bun platform packages so one effect instance serves CORS

### LINKED ISSUE #8878: `Access-Control-Allow-Origin` missing from responses since 0.0.34 — desktop app cannot connect to a remote environment on a separate HTTPS origin
### Summary

Since 0.0.34 the server answers the CORS **preflight** correctly but omits `Access-Control-Allow-Origin` from the **actual** response. Any browser-context client whose origin differs from the server's therefore discards the response, and the desktop app never gets past `Failed to connect. Reconnecting...`.

0.0.33 is the last version that sends the header. Rolling the server back to 0.0.33 — changing nothing else — restored the desktop client immediately.

### Environment

- Server: headless `t3 serve --host <tailnet-ip> --port 3010`, Linux x64, Bun runtime
- Reached from the desktop over a separate HTTPS origin via Tailscale Serve
  (`https://<host>.ts.net:8445` → `http://<tailnet-ip>:3010`), valid Let's Encrypt cert
- Desktop app: macOS, tested at 0.0.33, 0.0.35, 0.0.36, 0.0.37 — all fail against a 0.0.34+ server

### Reproduction

```bash
t3 serve --host 127.0.0.1 --port 3099 --base-dir /tmp/t3probe

curl -sD- -o/dev/null -H 'Origin: https://app.t3.codes' \
  http://127.0.0.1:3099/.well-known/t3/environment | grep -i access-control
```

| Server version | Released | `Access-Control-Allow-Origin` on `GET` |
|---|---|---|
| 0.0.33 | 2026-08-10 | `*` |
| 0.0.34 | 2026-08-26 | absent |
| 0.0.35 | 2026-08-27 | absent |
| 0.0.36 | 2026-08-29 | absent |
| 0.0.37 | 2026-08-31 | absent |
| 0.0.38-nightly.20260831 | 2026-08-31 | absent |

The preflight still looks right on 0.0.37:

```
$ curl -sD- -o/dev/null -X OPTIONS -H 'Origin: https://app.t3.codes' \
    -H 'Access-Control-Request-Method: GET' .../.well-known/t3/environment
HTTP/2 204
access-control-allow-origin: *
access-control-allow-methods: GET, POST, OPTIONS
access-control-allow-headers: authorization,b3,traceparent,content-type,dpop
access-control-max-age: 600
```

The same GET that lacks the header returns `200` with the expected body, so this is purely a response-header regression. Reproduced on `/.well-known/t3/environment`, `/api/auth/session` and `/health`, and with every `Origin` value tried (`https://app.t3.codes`, `file://`, `app://-`, `null`, `http://localhost:5173`, the server's own origin).

### Symptom in the desktop app

```
Provider settings are unavailable
Failed to connect. Reconnecting... Reason: Failed to fetch remote environment endpoint
https://<host>.ts.net:8445/.well-known/t3/environment
(HttpClientError: Transport error (GET https://<host>.ts.net:8445/.well-known/t3/environment))
```

What rules out a network or TLS problem:

- `curl -v` from the *same Mac*, at the same moment, completes the handshake and returns `HTTP/2 200` with the descriptor JSON; cert verifies, ALPN negotiates h2.
- The **iOS app connected to the same 0.0.35 server successfully** while the desktop could not. A native client does not enforce CORS; curl does not either. Only the browser-context client fails.
- Server-side `/api/auth/clients` showed no connection attempt from the desktop after the upgrade, while the stored bearer token was still valid for another two weeks.

### Why it may have gone unnoticed

The common setup is a desktop app talking to a server on the same machine over loopback, where the request is not cross-origin. It only bites when the environment is remote *and* published on a different origin — the documented Tailscale Serve setup in `docs/user/remote-access.md`.

### Expected

Responses carry `Access-Control-Allow-Origin` (as 0.0.33 did), so a remote environment on its own HTTPS origin stays reachable from browser-context clients.

Failing that, a documented way to configure allowed origins would be enough — there is no server setting for it today.

lnieuwenhuis:
Since the CLI bundling change in #5877, `dist/bin.mjs` inlines `effect` but kept `@effect/platform-bun` and `@effect/sql-sqlite-bun` external. A Bun-hosted server therefore loaded a second `effect` from node_modules. Effect attaches CORS, compression, and auth headers to real responses through a pre-response handler stored in a module-level WeakMap, so the bundled copy wrote handlers the platform server's copy never read. OPTIONS preflight still worked because it is answered inline, but every actual response lost `Access-Control-Allow-Origin` and gzip, which is why the desktop app could not connect to a remote Bun-run environment on a separate origin.

The fix externalizes only Bun's own module namespace (`bun`, `bun:*`) and bundles both platform packages, so one `effect` graph serves the whole server. The two packages move to devDependencies. Because correctness now depends on rolldown keeping `bun:sqlite` in a chunk reached only through `import()`, the server build gains a post-build check that walks the emitted chunk graph from both entries along static edges and fails if any eagerly loaded chunk imports a Bun module. Rolldown only warns on that merge and exits 0, so the check reads the artifact instead of trusting the bundler.

Verification: in the built artifact the single pre-response-handler WeakMap lives in the shared chunk that both `bin.mjs` and the `BunHttpServer` chunk import, and no bare `@effect/*-bun` runtime import remains. A probe bundled with the repo's own build predicates and run under Bun showed `Access-Control-Allow-Origin: *` and `Content-Encoding: gzip` restored on GET, and the real CLI under Node shows no regression. Bundle size grows by about 0.5%. The full server cannot run under Bun on Windows (the Bun PTY adapter is unsupported there), and the freshly built CLI has now passed the Linux Bun runtime smoke described below. A negative control confirmed the new build check rejects a bundle where `bun:sqlite` is hoisted into `bin.mjs`, which otherwise fails at load under Node.

Closes #8878

Fresh verification (2026-09-06): built the server/web artifacts and passed the isolated Linux Bun 1.4.0 runtime smoke for CORS, gzip body equivalence, pairing Set-Cookie, and authenticated cookie replay. The smoke is now wired into CI. 20 focused packaging tests passed. Published-package installation was not exercised; Windows Bun startup still has the existing unsupported PTY limitation.

---

> [!NOTE]
> **Medium Risk**
> Changes core server CLI bundling and build-time validation; a mistaken static import of Bun adapters would fail builds or break Node entrypoints, but the new eager-import scan is meant to catch that before release.
> 
> **Overview**
> Fixes missing **CORS**, **gzip**, and **auth** on Bun-hosted servers when the desktop app talks to a remote environment on another origin. External `@effect/platform-bun` and `@effect/sql-sqlite-bun` had been loading a **second** `effect` beside the bundle, so pre-response handlers (WeakMap-keyed) never applied to real responses.
> 
> **Bundling policy** in `cli-external-packages.ts` now **inlines** those two packages and only keeps **`bun` / `bun:*`** external. They move from `dependencies` to **devDependencies** in `apps/server/package.json`. Inlining pulls `bun:sqlite` into the artifact, which must stay behind **dynamic** `import()` so `node bin.mjs` still works.
> 
> A **post-bundle guard** runs in the server CLI **`build`** step: `findEagerBunRuntimeImports` walks static edges from `bin.mjs` and `service-launcher.mjs` and fails with **`ServerCliEagerBunImportError`** if any eagerly loaded chunk references a Bun runtime module (rolldown only warns). Tests cover the scanner and updated bundle expectations; desktop self-containment comments point at this check for the Bun adapters.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bff7b06a52ef4abb3aecaa00e949dddc12ee41f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Bundle `@effect/platform-bun` and `@effect/sql-sqlite-bun` instead of externalizing them
> - Removes `@effect/platform-bun` and `@effect/sql-sqlite-bun` from CLI external prefixes in [cli-external-packages.ts](https://github.com/pingdotgg/t3code/pull/9118/files#diff-14145c906143a006d1358231b6aa1e7707f576b55a05fd3c0a536e22e044ce75) so they are bundled rather than left external. Bun runtime modules (`bun`, `bun:*`) remain external.
> - Adds `findEagerBunRuntimeImports` to walk the emitted chunks' static import graph from entry chunks and report any reachable Bun runtime module imports.
> - Adds `assertNoEagerBunImports` to the build command in [cli.ts](https://github.com/pingdotgg/t3code/pull/9118/files#diff-e8442ce95db64bad13cb361a5638bd82e8723d4a0aa5833a7d0b81b4fc91ccf9), which fails the build with `ServerCliEagerBunImportError` if any eagerly loaded chunk imports a Bun runtime module.
> - Updates tests in [cli-external-packages.test.ts](https://github.com/pingdotgg/t3code/pull/9118/files#diff-e14208c674bf017b0dd5a81d3191e5f6aec3acf49e51baa8dfe02a04fdc08690) to expect the two adapter packages bundled and to cover the new eager-import detection logic.
> - Risk: `isExternalCliDependency` now bundles `@effect/platform-bun` and `@effect/sql-sqlite-bun`; if any code statically imports Bun runtime modules through these packages, the build will fail via `assertNoEagerBunImports` rather than silently producing a broken bundle.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized bff7b06. 4 files reviewed, 1 issue evaluated, 1 issue filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> <details>
> <summary>apps/server/scripts/cliErrors.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 80](https://github.com/pingdotgg/t3code/blob/635625d42c36f6fdce3e117035803b4c7d28b8d5/apps/server/scripts/cliErrors.ts#L80): `ServerCliEagerBunImportError` also reports violations for the bare specifier `bun` (as exercised by the new scanner test), but its message always says the failure is `ERR_UNSUPPORTED_ESM_URL_SCHEME` for a `bun:` URL. A static bare `bun` import is package resolution and Node reports `ERR_MODULE_NOT_FOUND`, so this build failure gives an incorrect diagnosis for one of the explicitly detected violation types. <b>[ Out of scope (post-validation triage) ]</b>
> </details>
>
>
> </details>
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved server runtime compatibility across Node and Bun environments.
  - Prevented incompatible Bun-only modules from being loaded during server startup.
  - Improved server packaging and runtime module resolution.
  - Added validation to detect unsafe runtime imports before packaging completes.

- **Tests**
  - Added automated server startup checks covering readiness, CORS, compression, authentication, error reporting, graceful shutdown, and temporary-file cleanup.
  - Expanded validation for session handling and browser-session pairing.


derektrimm:
@juliusmarminge re your note on #8878 about browser-session cookie emission: here is a manual Linux check of this branch that includes it.

Repo-built `apps/server/dist`, comparing upstream `main` at `04efa7907` with the same `main` plus this PR merged on top (`635625d42`), so the only source difference is this change. Bun 1.4.0 and Node v22.23.2, fresh `--base-dir` per run, server bound to `0.0.0.0` so the auth policy reports `remote-reachable`, and a fresh credential from `node bin.mjs auth pairing create` for each pairing run.

| Build | Runtime | `GET /.well-known/t3/environment` with `Origin` | `GET /` with `accept-encoding: gzip` | `POST /api/auth/browser-session` |
|---|---|---|---|---|
| `main` | Node | `Access-Control-Allow-Origin: *` | `Content-Encoding: gzip`, `Vary: Accept-Encoding` | `200`, `authenticated: true`, `Set-Cookie` present |
| `main` | Bun | header absent | both headers absent | `200`, `authenticated: true`, `Set-Cookie` absent |
| `main` + PR | Node | `Access-Control-Allow-Origin: *` | `Content-Encoding: gzip`, `Vary: Accept-Encoding` | `200`, `authenticated: true`, `Set-Cookie` present |
| `main` + PR | Bun | `Access-Control-Allow-Origin: *` | `Content-Encoding: gzip`, `Vary: Accept-Encoding` | `200`, `authenticated: true`, `Set-Cookie` present |

The `main`/Bun pairing response is the #7756 shape exactly: `200` with `authenticated: true` and no `Set-Cookie` (`Cache-Control` and `Pragma` are missing from that response too). Replaying the issued cookie against `GET /api/auth/session` authenticated in every run that had one. The Bun cells were repeated three times each with identical headers.

This is symptom-level evidence from the repo build with workspace `node_modules`, not an `npm pack` install of the published package, and not in-repo regression coverage.

derektrimm:
A Bun runtime smoke for the browser-session cookie case discussed in #8878 is available here: https://github.com/derektrimm/t3code/commit/a90d0ef1186f996cb19239c92f818be4ab29a1d5. It runs the built CLI under Bun 1.4.0 in the Check job and asserts CORS, gzip, and the pairing `Set-Cookie`. The smoke fails on `main` at `bfef973d9` at the CORS assertion and passes with `bff7b06` plus this commit. The commit is based on `bff7b06` and can be cherry-picked if useful.

shivamhwp:

The smoke script needs to remove `T3_SERVICE_LAUNCHER_CONTEXT` from its child environment too. A command started inside a launcher-managed T3 session inherits that variable, and the current filter only removes `T3CODE_*` and `VITE_DEV_SERVER_URL`. The standalone Bun child then exits in `ServiceLauncherClient.resolveStartup` with a version mismatch, or an unavailable IPC channel when versions match, before checking CORS or cookies. Exclude `T3_SERVICE_LAUNCHER_CONTEXT` when constructing `env`; this child has no launcher IPC connection.

derektrimm:
Confirmed. The launcher spawns its child with that variable set (`apps/server/src/serviceLauncher.ts:406`), `resolveStartup` in `apps/server/src/cloud/serviceLauncherClient.ts:119-133` fails on it with `version-mismatch` or `ipc-unavailable`, and `ServerEnvironment.ts:194` runs that check during every `serve` startup, so a smoke started from inside a launcher-managed session dies before it listens. CI runners never carry the variable, which is why the Check job passes. Adding `key !== "T3_SERVICE_LAUNCHER_CONTEXT"` to the filter at `scripts/server-runtime-smoke.ts:89` covers it.

lnieuwenhuis:
Fixed in 0029e7468. The standalone smoke child now excludes `T3_SERVICE_LAUNCHER_CONTEXT`. CI runs the existing Bun smoke with a synthetic inherited launcher context so this cannot silently regress on runners without launcher state.

Verified against the rebuilt server with Bun 1.4.0: both a mismatched child version and the current version without IPC pass CORS, gzip, pairing-cookie issuance, and authenticated cookie replay. Removing the filter reproduces the reported launcher-version startup failure. The 20 packaging tests, scripts/server typechecks, targeted lint, and Node bundle `--version` check pass. Also merged current upstream main and regenerated the lockfile, retaining the new platform-node-shared runtime dependency.

PR #9190, opened 2026-09-02
feat(web): make context window indicator opt-in

maria-rcks:
The circular context window indicator is now a legacy feature that defaults off for existing and new client settings. Users can restore it from Settings → General → Legacy features, and settings search plus restore-default behavior include the new toggle. Verified with contracts and web typechecks, focused settings and meter tests, targeted lint, and a direct schema decode check. Built by `gpt-5.6-sol` in T3 Code through the Codex harness.

---

> [!NOTE]
> **Low Risk**
> Client-local UI preference with a default-off legacy toggle; no auth, data, or server behavior changes beyond settings persistence.
> 
> **Overview**
> The composer’s circular **context window usage indicator** is hidden by default. A new client setting, `contextWindowMeterEnabled` (default `false`), gates whether `ChatComposer` passes context-window data to the footer primary actions.
> 
> Users who want the old behavior can turn it on under **Settings → General → Legacy features** (“Context window indicator”). Settings search, restore-defaults tracking, and patch/schema decoding in contracts cover the new flag; tests assert default-off and explicit opt-in.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5523463ffd0732aeab0db03fa25e208ca7c79961. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add opt-in `contextWindowMeterEnabled` setting to `ClientSettings`
> - Adds the boolean `contextWindowMeterEnabled` field to `ClientSettingsSchema` and `ClientSettingsPatch`, defaulting to false when absent so the context window meter is off by default.
> - Forwards `activeContextWindow` to `ComposerFooterPrimaryActions` only when the setting is true; otherwise passes null so the meter is hidden.
> - Exposes a legacy "Context window indicator" switch in General settings under `LegacyFeaturesSection`, with searchable terms registered in `settingsSearch.ts`.
> - Updates `useSettingsRestore` to detect and reset `contextWindowMeterEnabled`, and adjusts the desktop test fixture to set it explicitly false.
> - Behavioral Change: existing clients with no `contextWindowMeterEnabled` value now decode as false; verify any out-of-tree consumers of `ClientSettings` expecting the meter to show by default.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5523463.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/settings/DesktopClientSettings.test.ts
+++ b/apps/desktop/src/settings/DesktopClientSettings.test.ts
@@ -23,6 +23,7 @@ const clientSettings: ClientSettings = {
   confirmThreadArchive: true,
   confirmThreadDelete: false,
   confirmThreadUnpin: false,
+  contextWindowMeterEnabled: false,
   dismissedProviderUpdateNotificationKeys: [],
   diffIgnoreWhitespace: true,
   environmentIdentificationMode: "artwork",
--- a/apps/web/src/components/chat/ChatComposer.tsx
+++ b/apps/web/src/components/chat/ChatComposer.tsx
@@ -4339,7 +4339,9 @@ export const ChatComposer = memo(function ChatComposer(props: ChatComposerProps)
                   ) : null}
                   <ComposerFooterPrimaryActions
                     compact={isComposerPrimaryActionsCompact}
-                    activeContextWindow={activeContextWindow}
+                    activeContextWindow={
+                      settings.contextWindowMeterEnabled ? activeContextWindow : null
+                    }
                     activeThreadModelDisplayName={activeThreadModelDisplayName}
                     pendingAction={pendingPrimaryAction}
                     isRunning={phase === "running"}
--- a/apps/web/src/components/settings/SettingsPanels.tsx
+++ b/apps/web/src/components/settings/SettingsPanels.tsx
@@ -522,6 +522,9 @@ export function useSettingsRestore(onRestored?: () => void) {
       ...(settings.showSkillsInSlashMenu !== DEFAULT_UNIFIED_SETTINGS.showSkillsInSlashMenu
         ? ["Show skills in slash menu"]
         : []),
+      ...(settings.contextWindowMeterEnabled !== DEFAULT_UNIFIED_SETTINGS.contextWindowMeterEnabled
+        ? ["Context window indicator"]
+        : []),
       ...(settings.enableLegacyTokenStreaming !==
       DEFAULT_UNIFIED_SETTINGS.enableLegacyTokenStreaming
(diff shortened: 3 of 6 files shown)


jwilmsnallian:
@maria-rcks  Could you share why this is marked as legacy? I found it very useful

bman654:
came to ask the same question.  I'm worried if there isn't a replacement planned.  Managing context is key part of getting good model results.

maxredundancy:
I'd also be super interested in why this was disabled by default and hidden in legacy?

ElliotDrel:
Adding another strong vote against classifying this as legacy. I love the context-window indicator and rely on it during long-running work.

Context pressure and compaction materially affect how I manage a thread; hiding that signal makes it harder to decide when to checkpoint, start a new thread, or change course before context quality degrades.

I understand making it optional for people who prefer a cleaner composer, but I would really like this kept as a normal, supported feature—not an opt-in “legacy” feature that signals it may be removed. Ideally it should live in the regular General settings, with a clear long-term commitment or a documented replacement.

Luzivog:
In my opinion, the context-window indicator shouldn't be classified as legacy. Seeing how much context is in use is still useful for managing long threads and deciding when to start a fresh one, even with automatic compaction.

Making it optional is reasonable, but I'd like it to remain a supported feature in the regular settings. Could you explain the reasoning behind the legacy label, and whether a replacement is planned?

kratofl:
Can only agree with previous comments.
Was very useful, will there be a replacement or maybe return from legacy status?

doasfrancisco:
Completely unnecessary. Please work on forking chats instead of this.

MagicMajky:
Is this really a legacy feature? 
I also have to agree, working with context is core part of developing with AI and that includes managing the context window.

Its hard to see how this should not be the default. Especially when there really isn't stated any reason for hiding the context usage by default. Especially for us mere mortals that might use open-weight models whose context window isn't 1 million like opus.

I mean you don't owe us a reason and I really appreciate that you are taking the time to improve T3 code especially since I am using it everyday. 

However its hard to see the value of this without one. 
Is there some new way of working with LLMs that makes caring about context "legacy" that I am just not aware of?

nbros:
I am also puzzled as to why this is now considered a "legacy" feature.
Codex and Claude Code both have this built in.
I follow Matt Pocock's courses on AI and he insists on the importance of staying in the "smart zone" vs the "dumb zone" when context starts filling up too much.
cf https://www.aihero.dev/ai-coding-dictionary/smart-zone

PR #9209, opened 2026-09-02
fix(desktop): reject opaque-origin renderer navigations

Lucenx9:
## What Changed

`isSameOriginRendererNavigation` now compares the parsed URL protocol and host instead of `URL.origin`.

The regression test now covers a different `t3code` host plus `file:` and `vscode:` navigations.

## Why

Node reports `"null"` as the origin for each of these opaque URLs. Comparing those values made unrelated URLs look same-origin, so Electron's `will-navigate` guard could allow them to replace the renderer.

Protocol and host match the authority used by the desktop protocol handler. The comparison still allows routes under `t3code://app/`, while different hosts and schemes fail closed.

## Blast Radius

The helper also gates development load completion and retry eligibility. Existing same-host and retry tests still pass, malformed URLs still return `false`, and the host comparison includes the port. No public contract, UI, or documented workflow changed.

## Verification

Before the production change:

```text
$ ./node_modules/.bin/vp test run apps/desktop/src/window/DesktopWindow.test.ts --reporter verbose
FAIL  DesktopWindow > recognizes only same-origin renderer navigations
AssertionError: expected true to be false
Test Files  1 failed (1)
Tests       1 failed | 25 passed (26)
```

After the production change:

```text
$ ./node_modules/.bin/vp test run apps/desktop/src/window/DesktopWindow.test.ts
Test Files  1 passed (1)
Tests       26 passed (26)
```

- `./node_modules/.bin/vp run --filter @t3tools/desktop typecheck` passed.
- `./node_modules/.bin/vp lint --report-unused-disable-directives apps/desktop/src/window/DesktopWindow.ts apps/desktop/src/window/DesktopWindow.test.ts` passed.
- `./node_modules/.bin/vp fmt --check apps/desktop/src/window/DesktopWindow.ts apps/desktop/src/window/DesktopWindow.test.ts` passed.
- `gpt-daybreak-blue-latest` reviewed the diff against the repo standards and the bug specification. Both reviews passed with no findings.

## Checklist

Model: GPT-5.6. Harness: Codex.

---

> [!NOTE]
> **Medium Risk**
> Changes navigation security logic in the desktop renderer; incorrect matching could allow or block main-frame loads, though behavior is narrowed to fail closed for opaque origins.
> 
> **Overview**
> **`isSameOriginRendererNavigation`** no longer uses `URL.origin` equality. It now requires matching **protocol** and **host**, so custom-scheme URLs like `t3code://`, `file:`, and `vscode:` are not treated as same-origin when Node reports a null/opaque origin.
> 
> That tightens the Electron **`will-navigate`** guard (and related dev load/retry checks that call the same helper): in-app routes under `t3code://app/` still pass, while different hosts, other schemes, and external HTTPS URLs are blocked or opened externally as before.
> 
> Tests add regressions for `t3code://other/`, `file:///etc/passwd`, and a `vscode://` navigation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b52d19a253aa1f4105fcca79dbbe1ed728479970. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `isSameOriginRendererNavigation` to reject opaque-origin renderer navigations
> Replaces direct URL-origin comparison in [DesktopWindow.ts](https://github.com/pingdotgg/t3code/pull/9209/files#diff-d92229c509071d1e9d3595d187fe2f1f13c394b22e1d9d2070ed6fb99678f429) with explicit checks of parsed protocols and hosts. Renderer navigations are now rejected when the protocol or host differs from the application URL, including custom-scheme and file URLs. Adds test coverage for different-host, file-URL, and custom-scheme cases in [DesktopWindow.test.ts](https://github.com/pingdotgg/t3code/pull/9209/files#diff-b3e75608c40fd022814266304d4a9f2a00421cc93ba774668be3fad18bc4aa7e).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b52d19a.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved navigation origin checks so URLs using the same protocol and host are treated consistently, even when their ports differ.
  * Added safeguards to reject navigation to different application paths, local files, and remote connection URLs when they are not same-origin.

### DIFF
--- a/apps/desktop/src/window/DesktopWindow.test.ts
+++ b/apps/desktop/src/window/DesktopWindow.test.ts
@@ -444,6 +444,24 @@ describe("DesktopWindow", () => {
     assert.isFalse(
       DesktopWindow.isSameOriginRendererNavigation({
         applicationUrl: "t3code://app/",
+        navigationUrl: "t3code://other/",
+      }),
+    );
+    assert.isFalse(
+      DesktopWindow.isSameOriginRendererNavigation({
+        applicationUrl: "t3code://app/",
+        navigationUrl: "file:///etc/passwd",
+      }),
+    );
+    assert.isFalse(
+      DesktopWindow.isSameOriginRendererNavigation({
+        applicationUrl: "t3code://app/",
+        navigationUrl: "vscode://vscode-remote/ssh-remote+host/tmp",
+      }),
+    );
+    assert.isFalse(
+      DesktopWindow.isSameOriginRendererNavigation({
+        applicationUrl: "t3code://app/",
         navigationUrl: "https://accounts.microsoft.com/oauth",
       }),
     );
--- a/apps/desktop/src/window/DesktopWindow.ts
+++ b/apps/desktop/src/window/DesktopWindow.ts
@@ -185,7 +185,12 @@ export function isSameOriginRendererNavigation(input: {
   readonly navigationUrl: string;
 }): boolean {
   try {
-    return new URL(input.applicationUrl).origin === new URL(input.navigationUrl).origin;
+    const applicationUrl = new URL(input.applicationUrl);
+    const navigationUrl = new URL(input.navigationUrl);
+    return (
+      applicationUrl.protocol === navigationUrl.protocol &&
+      applicationUrl.host === navigationUrl.host
+    );
   } catch {
     return false;
   }


Lucenx9:
## Review

**Verdict: LGTM** — root cause correctly identified and fixed at the right layer. No blocking findings.

**Root cause is right.** `new URL().origin` returns the string `"null"` for every non-special scheme (`t3code:`, `t3code-dev:`, `file:`, `vscode:`). Since `applicationUrl` is always `t3code://app/` (prod) or `t3code-dev://app/` (dev), the old comparator was effectively `"null" === "null"` — true for *any* opaque-origin URL, so `will-navigate` would let `file:///etc/passwd` replace the renderer. Reproduced against the old code:

```
node -e "console.log(new URL('file:///etc/passwd').origin === new URL('t3code://app/').origin)"
→ true
```

**The one fact this fix is safe because of (verified by running code, not just reading it):** for special schemes (`http:`/`https:`), `URL.host` carries the same authority `origin` does — port included, default port elided — so protocol+host equality is exactly origin equality there, and behavior is unchanged wherever the old check was meaningful. For non-special schemes the old check was meaningless, so this is strictly a tightening. Confirmed `new URL('http://localhost:3000/').host === 'localhost:3000'`.

**Blast radius — all three call sites checked:**

- `will-navigate` guard (apps/desktop/src/window/DesktopWindow.ts:554): opaque URLs now hit `preventDefault()` instead of replacing the renderer. `parseSafeExternalUrl` (apps/desktop/src/electron/ElectronShell.ts:27) only admits `http:`/`https:` and credential-free `vscode://vscode-remote/ssh-remote+…` links, so `file:` is just blocked, never handed to the OS. Bonus: the vscode editor deep link now actually reaches `openExternal` as documented instead of being swallowed as "same-origin".
- Dev `did-finish-load` (apps/desktop/src/window/DesktopWindow.ts:673): an opaque URL no longer resets retry state or claims the window title. Fails closed.
- Dev retry eligibility (apps/desktop/src/window/DesktopWindow.ts:199): semantics unchanged for real `t3code-dev://app/` loads.
- The WSL splash is a separate `BrowserWindow` with its own `loadURL` (apps/desktop/src/window/DesktopWindow.ts:848), so it never interacts with these guards.

**Verification (ran locally on the branch):** `vp test run apps/desktop/src/window/DesktopWindow.test.ts` → 26/26 passed; `@t3tools/desktop` typecheck and scoped lint clean.

**Nitpicks (non-blocking):**

1. The expected authority now lives in two places: this comparator derives it from `applicationUrl`, while the protocol handler pins `DESKTOP_HOST = "app"` (apps/desktop/src/electron/ElectronProtocol.ts:11). They agree today; if the two ever drift, the guard silently breaks. Deriving the window guard from the same constant would remove that coupling — though comparing against the URL actually loaded is arguably the more honest invariant, so fine to leave as-is.
2. The CodeRabbit summary claims same protocol+host with *different ports* is treated as consistent — it isn't (`host` includes the port), and that stricter behavior is correct. Flagging so nobody "fixes" it later.

Small, well-tested, fails closed. Nice fix.

PR #9211, opened 2026-09-02
fix(server): release compatible writers before cross-instance resume

BranislavMateas:
## What Changed

When resuming a thread on another compatible Codex provider instance, T3 Code now:

- reuses the persisted resume cursor
- stops the previous provider session before starting the replacement
- preserves the existing Codex thread and its context

Added a regression test covering account switching while the original session still owns the Codex thread.

## Why

Codex allows only one active writer per thread. T3 Code previously started the replacement session before stopping the original one, causing an `already has an active writer` error. A retry could then create a fresh Codex thread and lose the conversation context.

Compatible provider instances share the same continuation identity, so releasing the previous writer before resuming is safe and keeps the thread intact.

## Checklist

- [x] No UI changes

> [!NOTE]
> ### Release compatible writers before cross-instance resume in `ProviderService.startSession`
> Allows a persisted session to resume on a different compatible provider instance by matching on driver kind and continuation identity rather than requiring identical instance IDs.
> - When resuming with a persisted cursor across instances, retrieves the previous adapter, checks whether it still owns the thread, stops that session, and records a session-stopped analytics event before starting the new instance.
> - Preserves the resume cursor and working directory for compatible cross-instance resumes.
> - Adds an integration test in [ProviderService.test.ts](https://github.com/pingdotgg/t3code/pull/9211/files#diff-cbac8cf0b9c33ddbfff2137001d4b9ef6d645a0b205efcf6c714b12efab70f14) using two fake Codex adapters to verify the personal session is stopped before the work session starts.
> - Risk: cross-instance resume now depends on correct compatibility detection via `driverKind` and continuation identity; a false positive could stop a live session on the wrong adapter.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c7bc29b.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/ProviderService.test.ts
+++ b/apps/server/src/provider/Layers/ProviderService.test.ts
@@ -637,6 +637,116 @@ it.effect("ProviderServiceLive rejects new sessions for disabled custom instance
   }).pipe(Effect.provide(NodeServices.layer)),
 );
 
+it.effect(
+  "ProviderServiceLive releases a compatible Codex writer before resuming on another instance",
+  () =>
+    Effect.gen(function* () {
+      const personalInstanceId = ProviderInstanceId.make("codex_personal");
+      const workInstanceId = ProviderInstanceId.make("codex_work");
+      const personal = makeFakeCodexAdapter();
+      const work = makeFakeCodexAdapter();
+      const workStartSession = work.startSession.getMockImplementation();
+      assert.isDefined(workStartSession);
+      const lifecycle: string[] = [];
+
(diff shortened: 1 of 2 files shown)


shivamhwp:

Thanks for addressing the active-writer conflict and preserving conversation context when switching compatible Codex instances.

We're closing this because Orchestration V2 in #2829 is replacing the provider-switching path this PR changes. Any remaining stop-before-resume or context-preservation gaps should be fixed in V2.

This is a direction decision, not a rejection of the bug report. No revision to this PR is needed.

PR #9227, opened 2026-09-02
fix(server): render async Codex messages

### LINKED ISSUE #9222: [Bug]: Codex harness accepts send_user_message_async but the UI drops the messages
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Start a Codex thread through T3 Code's harness.
2. Have the agent call `send_user_message_async` with a visible test message.
3. Observe that the tool result reports `{ "accepted": true }`.
4. Check the T3 Code thread UI for the asynchronous message.
5. Repeat the call with a second distinct message.

### Expected behavior

Messages accepted through `send_user_message_async` should be delivered to and rendered in the active T3 Code thread. If this event type is unsupported, the harness should reject the call explicitly instead of returning `accepted: true`.

### Actual behavior

Both tool calls returned `{ "accepted": true }`, but neither message appeared in the T3 Code UI. The calls therefore look successful to the agent while the user receives nothing.

### Impact

This silently breaks agent progress notifications and any workflow that relies on asynchronous user messages.

### Version or commit

Current T3 Code environment as of 2026-09-02; exact commit unavailable.

### Environment

macOS, T3 Code, Codex harness, gpt-5.6-sol

### Logs or stack traces

```shell
send_user_message_async({"message":"This is a test message sent with send_user_message_async."})
=> {"accepted":true}

send_user_message_async({"message":"Here's the second test message from send_user_message_async."})
=> {"accepted":true}
```

### Workaround

Return the message through the normal assistant commentary/final response stream.

Gigioxx:
Codex emits `send_user_message_async` as a separate completed agent message during an active turn. Provider ingestion always matched assistant completions to the active streamed segment, so it discarded the async message text whenever commentary was already buffered.

This matches completions to active segments by provider item ID. A different item now becomes its own assistant message while the original segment stays active. Focused coverage reproduces buffered commentary, an async message, and the original completion in one turn.

Fixes #9222.

## Verification

- `vp test run apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts` (51 passed)
- server typecheck
- targeted lint and formatting
- diff whitespace check

Implemented with `gpt-5.6-sol` in the Codex harness through T3 Code.

---

> [!NOTE]
> **Medium Risk**
> Changes assistant message completion and segment lifecycle in provider ingestion, which affects how all multi-item turns are projected; scope is narrow but message ordering/content is user-visible.
> 
> **Overview**
> Fixes **Codex async assistant messages** (`send_user_message_async`) being dropped when a turn already had buffered/streaming commentary on another item.
> 
> **Provider runtime ingestion** now ties `item.completed` for assistant messages to the **active segment only when the completion’s provider item id matches** that segment’s `baseKey`. A completion for a different item finalizes as its **own** assistant message and no longer reuses or overwrites the in-flight segment. **Segment state is cleared** only when the completing event matches that active segment, so commentary can keep streaming after an interleaved async message completes.
> 
> Adds an integration test that streams commentary, completes a separate async assistant item, then completes the commentary item and asserts both texts appear in the thread projection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f4be4d728b36c08ef362b7fabd79de5025582cf9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix async assistant message preservation in `ProviderRuntimeIngestion`
> Assistant completion events were updating the wrong active message when a turn contained separate async assistant messages. The completion handler now matches the active assistant segment by its segment base key, and derives the assistant message ID from that matching segment. Segment state is cleared only when a matching completing segment is found.
> - Adds an integration test in [ProviderRuntimeIngestion.test.ts](https://github.com/pingdotgg/t3code/pull/9227/files#diff-ea511ca36a45f5254c8f22ac107b5469fed67d652de6af2c5d2763bf00ce53c3) that emits commentary text, completes a separate async assistant message, then verifies both texts remain projected.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f4be4d7.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts
@@ -1085,6 +1085,63 @@ describe("ProviderRuntimeIngestion", () => {
     expect(message?.streaming).toBe(false);
   });
 
+  it("preserves an async assistant message before the final response", async () => {
+    const harness = await createHarness();
+    const now = "2026-01-01T00:00:00.000Z";
+
+    harness.emit({
+      type: "content.delta",
+      eventId: asEventId("evt-commentary"),
+      provider: ProviderDriverKind.make("codex"),
+      createdAt: now,
+      threadId: asThreadId("thread-1"),
+      turnId: asTurnId("turn-async-message"),
+      itemId: asItemId("commentary"),
+      payload: {
+        streamKind: "assistant_text",
+        delta: "Working on it.",
+      },
+    });
+    harness.emit({
+      type: "item.completed",
+      eventId: asEventId("evt-async-message"),
+      provider: ProviderDriverKind.make("codex"),
+      createdAt: now,
+      threadId: asThreadId("thread-1"),
+      turnId: asTurnId("turn-async-message"),
+      itemId: asItemId("async-message"),
+      payload: {
+        itemType: "assistant_message",
+        status: "completed",
+        detail: "Still investigating.",
+      },
+    });
+    harness.emit({
+      type: "item.completed",
+      eventId: asEventId("evt-commentary-completed"),
+      provider: ProviderDriverKind.make("codex"),
+      createdAt: now,
+      threadId: asThreadId("thread-1"),
+      turnId: asTurnId("turn-async-message"),
+      itemId: asItemId("commentary"),
+      payload: {
+        itemType: "assistant_message",
+        status: "completed",
+        detail: "Working on it.",
+      },
+    });
+
+    const thread = await waitForThread(harness.readModel, (entry) =>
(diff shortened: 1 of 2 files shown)


shivamhwp:

Please rebase the item-ID match onto the current completion handler while preserving its direct projection queries. Current main uses `getThreadMessageById` and `projectionThreadMessages.hasAssistantMessageForTurn`; restoring the old whole-thread detail read would undo that change.

Use the existing `emitAndDrain` helper for the new regression instead of `waitForThread` polling. Include two async completions followed by another delta on the original item, so the regression covers both repeated updates and continued commentary.

PR #7201, opened 2026-08-16
fix(server): keep new Codex SQLite state shadow-local

matsvarn:
## Problem

Codex can add a top-level SQLite runtime database after a shadow home was first materialized. If Codex creates regular copies in both homes, the next materialization rejects the shadow copy and disables that provider.

## Root cause

The home layout treated each shared-home entry separately. It had no ownership rule for a new SQLite base file and its WAL and SHM sidecars.

## Fix

Treat each top-level `.sqlite` base name, `-wal`, and `-shm` sidecar as one family. A regular shadow database keeps the complete family local. Both homes retain their files, and repeated materialization stays idempotent. Unrelated regular-file conflicts still return the existing typed error.

## Tests

- Added a regression that materializes once, creates `queue_1.sqlite`, WAL, and SHM files in both homes, then materializes twice.
- Added a future SQLite filename case and verified missing shadow sidecars do not become shared links.
- Extended the unrelated-file conflict test to prove both copies remain unchanged and the layout can recover after the conflict is removed.
- Focused test: 9 passed.
- Focused format and lint: passed.
- Server typecheck: passed with no errors.

The regression test fails on the pinned base with `CodexShadowHomeEntryConflictError` for `queue_1.sqlite`.

## Risks

Low. The rule applies only to top-level SQLite families whose shadow base is already a regular entry. It removes only family symlinks when local ownership is selected. Shared targets and regular files remain unchanged. Non-SQLite conflict behavior is unchanged.

## Release note

Codex shadow-home providers no longer disable when new SQLite runtime databases appear.

Fixes #5817

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] UI screenshots are not applicable
- [x] Animation or interaction video is not applicable

Implemented with GPT 5.6 Sol via Codex.

> [!NOTE]
> ### Preserve shadow-local SQLite file families in `materializeCodexShadowHome`
> - When Codex creates a new SQLite database in the shadow home between materialization passes, `materializeCodexShadowHome` now detects the base `.sqlite` file as a real (non-symlink) file and marks the entire family (base, `-wal`, `-shm`) as local.
> - Any existing symlinks for the family are removed, and symlink creation is skipped for all family members, keeping the shadow-local database intact.
[...]

Issue #5817, opened 2026-08-09
[Bug]: New Codex runtime entries can disable existing shadow-home providers


zenDevNomad:
### Summary

T3's Codex shadow-home setup only links entries that already exist in the shared home when materialization runs. If a later Codex update creates a new top-level runtime entry in both the shared and shadow homes, the next materialization rejects the shadow copy and disables the provider.

Codex 0.147 introduced `queue_1.sqlite`, which triggers this today. The same failure can recur when Codex adds another home entry.

![Sanitized T3 Code provider error showing queue_1.sqlite shadow-home conflict](https://raw.githubusercontent.com/zenDevNomad/t3code/102854c75f578e54e7e235f0364e4b02f24881ce/codex-queue-shadow-home.png)

### Reproduction

1. Configure the documented multi-account layout: one shared `homePath` and another account using that `homePath` plus a separate `shadowHomePath`.
2. Start with no `queue_1.sqlite` in either home and let T3 materialize the shadow home.
3. Run Codex 0.147 against both effective homes. Codex creates a regular queue database in each.
4. Refresh provider status or restart T3.

The shadow-home provider becomes disabled:

```text
Driver 'codex' failed to create instance: Cannot create Codex shadow home entry 'queue_1.sqlite' because '<shadow-home>/queue_1.sqlite' already exists and is not a symlink.
```

A minimal unit repro is to create regular `sharedHome/queue_1.sqlite` and `shadowHome/queue_1.sqlite` files, then call `materializeCodexShadowHome(layout)`.

### Why it happens

T3 [lists the current shared-home entries](https://github.com/pingdotgg/t3code/blob/1a003e383ac6b10258b8100c2617d938c4f06c69/apps/server/src/provider/Drivers/CodexHomeLayout.ts#L361-L378), then [`ensureSymlink` rejects any existing regular shadow entry](https://github.com/pingdotgg/t3code/blob/1a003e383ac6b10258b8100c2617d938c4f06c69/apps/server/src/provider/Drivers/CodexHomeLayout.ts#L216-L252). There is no recovery path; the conflict becomes a provider creation error.

Codex added the queue DB in [openai/codex@b87981a](https://github.com/openai/codex/commit/b87981a5102436feae6a20f52269fd18d6297cbc). In the affected installation, the shared DB and two shadow DBs were automatically created within six seconds, confirming this was runtime-created state rather than a copied home.

### Expected behavior

[...]

PR #9229, opened 2026-09-02
fix(server): keep Codex SQLite files in shared home

joe-lloyd:
## Problem

Codex auth overlays currently symlink SQLite databases and their WAL/SHM sidecars into each account-specific shadow home. On Windows, opening those databases through shadow symlink paths can repeatedly fail, and deleting the links does not stick because the next provider startup recreates them.

## Fix

Set `CODEX_SQLITE_HOME` to the shared Codex home for auth-overlay instances, keep SQLite files out of shadow-home materialization, and remove stale SQLite symlinks left by earlier launches. Real shadow-local SQLite files are preserved.

Materialization now also reads the shadow home so pre-existing SQLite symlinks are cleaned up; a failure to list it surfaces as `CodexShadowHomeFileSystemError` rather than a raw platform error.

## Verification

- `pnpm exec vp test run apps/server/src/provider/Drivers/CodexHomeLayout.test.ts`
- focused `vp lint` and `vp fmt --check`
- Codex CLI 0.152.1 `doctor --json` with disposable homes confirmed that a shadow `CODEX_HOME` plus shared `CODEX_SQLITE_HOME` resolves every SQLite database path to the shared home

Implemented with GPT-5.6 Sol using the Codex harness in T3 Code.

---

> [!NOTE]
> **Medium Risk**
> Changes Codex auth-overlay filesystem layout and provider env for SQLite paths; existing shadow homes may lose stale SQLite symlinks on next startup.
> 
> **Overview**
> Auth-overlay Codex instances now point SQLite at the **shared** home via `CODEX_SQLITE_HOME`, while each instance still uses its shadow `CODEX_HOME` for per-account auth.
> 
> Shadow-home materialization **stops** symlinking `.sqlite` files (and journal/shm/wal sidecars) from shared into shadow, and on each run **removes stale SQLite symlinks** already in the shadow home without touching real shadow-local database files. Listing the shadow directory is now part of that pass, with failures wrapped as `CodexShadowHomeFileSystemError`.
> 
> Tests cover no SQLite links after materialization, repeated materialization preserving separate shared/shadow DBs, stale-link cleanup, and conflict recovery after removing blocking files.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10a6f4343a6319afbaaa683027ef07694db616d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Keep Codex SQLite files in shared home for auth-overlay instances
> - Auth-overlay Codex provider instances now set `CODEX_SQLITE_HOME` to the shared home path so SQLite state stays shared while account credentials remain private in `CodexDriver.create` ([CodexDriver.ts](https://github.com/pingdotgg/t3code/pull/9229/files#diff-c705233bf69322a30e4a04be3428cd51dc6fa014d5a03b852da94c6ab9b6ae35)).
> - `materializeCodexShadowHome` in [CodexHomeLayout.ts](https://github.com/pingdotgg/t3code/pull/9229/files#diff-17973143128131a5e3f30fc3e32012a3a86aee719917b08d60615a47558fb8d5) excludes `.sqlite` databases and their journal/WAL/shared-memory sidecars from the shared-entry symlink set, and removes stale matching shadow symlinks instead of replacing local files.
> - Adds `SQLITE_ENTRY_NAME_PATTERN` to classify SQLite filenames, and wraps shadow-directory read failures as `CodexShadowHomeFileSystemError`.
> - Behavioral Change: auth-overlay instances now share SQLite state via `CODEX_SQLITE_HOME`; direct-home instances are unaffected. Reviewers should verify `SQLITE_ENTRY_NAME_PATTERN` covers all expected sidecar suffixes and that consumers of `CODEX_SQLITE_HOME` handle the shared path correctly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 10a6f43.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Drivers/CodexDriver.ts
+++ b/apps/server/src/provider/Drivers/CodexDriver.ts
@@ -7,11 +7,8 @@
  *   - `adapter`    — the Codex session/turn/approval runtime;
  *   - `textGeneration` — commit/PR/branch/title generation via `codex exec`.
  *
- * Each call to `create()` captures the `codexConfig` argument in closures
- * owned by the returned instance. Two instances created with different
- * `homePath`s (e.g. `codex_personal` + `codex_work`) therefore run with
- * fully independent Codex app-server processes and `CODEX_HOME`
- * environments — no shared mutable state.
+ * Each instance owns its app-server process and `CODEX_HOME`. Auth overlays
+ * share session and SQLite state while keeping account credentials private.
  *
  * Resource lifecycle: `create()` runs in a scope handed in by the registry.
  * Closing that scope releases the adapter's child processes, the managed
@@ -112,8 +109,11 @@ export const CodexDriver: ProviderDriver<CodexSettings, CodexDriverEnv> = {
       const serverSettings = yield* ServerSettingsService;
       const eventLoggers = yield* ProviderEventLoggers;
       const modelManifest = yield* ModelManifest.ModelManifest;
-      const processEnv = mergeProviderInstanceEnvironment(environment);
+      let processEnv = mergeProviderInstanceEnvironment(environment);
       const homeLayout = yield* resolveCodexHomeLayout(config);
+      if (homeLayout.mode === "authOverlay") {
+        processEnv = { ...processEnv, CODEX_SQLITE_HOME: homeLayout.sharedHomePath };
+      }
       const continuationIdentity = codexContinuationIdentity(homeLayout);
       const stampIdentity = withInstanceIdentity({
         instanceId,
--- a/apps/server/src/provider/Drivers/CodexHomeLayout.test.ts
+++ b/apps/server/src/provider/Drivers/CodexHomeLayout.test.ts
@@ -230,6 +230,94 @@ it.layer(NodeServices.layer)("CodexHomeLayout", (it) => {
       }),
     );
 
+    it.effect("keeps both homes' SQLite files intact across repeated materialization", () =>
+      Effect.gen(function* () {
+        const fileSystem = yield* FileSystem.FileSystem;
+        const path = yield* Path.Path;
+        const sharedHome = yield* makeTempDir("t3code-codex-shared-");
+        const shadowRoot = yield* makeTempDir("t3code-codex-shadow-root-");
+        const shadowHome = path.join(shadowRoot, "shadow");
+        const layout = yield* resolveCodexHomeLayout(
+          decodeCodexSettings({
+            homePath: sharedHome,
+            shadowHomePath: shadowHome,
+          }),
+        );
+
+        yield* materializeCodexShadowHome(layout);
+
+        // Codex adds a runtime database to both homes after the first pass.
+        const sqliteSuffixes = ["", "-wal", "-shm"];
+        for (const suffix of sqliteSuffixes) {
+          yield* writeTextFile(path.join(sharedHome, `queue_1.sqlite${suffix}`), `shared${suffix}`);
+          yield* writeTextFile(path.join(shadowHome, `queue_1.sqlite${suffix}`), `shadow${suffix}`);
+          yield* writeTextFile(
+            path.join(sharedHome, `future_runtime.sqlite${suffix}`),
+            `future-shared${suffix}`,
+          );
+        }
+        yield* writeTextFile(path.join(shadowHome, "future_runtime.sqlite"), "future-shadow");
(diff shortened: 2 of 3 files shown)


t3dotgg:
> [!NOTE]
>

Preserve the repeat-pass cases from [#7201](https://github.com/pingdotgg/t3code/pull/7201) at [head 979c4878](https://github.com/pingdotgg/t3code/commit/979c48787dd8cf0500140dc7abea5e850813f1cc). Set up the shadow home once, create a new SQLite database plus WAL/SHM files in both homes, then repeat setup twice. Both real file contents must survive and absent shadow sidecars must not become shared symlinks. Keep the unrelated `config.toml` conflict test, including recovery after removing the conflict. Include the Windows sidecar-only case on [#5817](https://github.com/pingdotgg/t3code/issues/5817), where no regular shadow database exists. These checks should use this PR's shared-home design, not the older ownership loop.

joe-lloyd:
Carried over at `10a6f434`, expressed through this PR's shared-home design rather than #7201's ownership loop.

**Repeat-pass** — `keeps both homes' SQLite files intact across repeated materialization`: materializes once, then creates `queue_1.sqlite` plus `-wal`/`-shm` as real files in *both* homes and materializes twice more. Asserts both sides' contents survive and that no shadow entry became a symlink. The `future_runtime.sqlite` family in the same test covers the absent-sidecar half — its base is a real file in the shadow home while the sidecars exist only in shared, and neither sidecar is linked in.

**#5817 sidecar-only** — `clears stale SQLite links when the shadow home has no database of its own`: no shadow database at all, just one stale `state.sqlite-wal` symlink and one real `state.sqlite-shm`. The link is removed, the real sidecar survives, and no base link is created. The stale link is deliberately left dangling (the shared WAL is already gone), because that path resolves through `readLink` rather than `stat` — if it didn't, a dangling link would surface as a materialization failure and disable the provider, which is the same shape as the original bug.

**`config.toml` conflict** — kept, and extended with the recovery leg: both copies are asserted unchanged after the conflict error, then the shadow copy is removed, materialization re-runs, and the link resolves to the shared file.

Both new tests fail against the pre-fix `CodexHomeLayout.ts` at the merge base, so they are regression coverage rather than restatements of the implementation. The file goes 8 → 10 tests.

- `pnpm exec vp test run apps/server/src/provider/Drivers/CodexHomeLayout.test.ts` — 10 passed
- focused `vp lint` / `vp fmt --check`, and `pnpm exec vp run --filter t3 typecheck` — clean

Separately re-verified the environment variable on Codex CLI 0.153.2 (the PR originally cited 0.152.1): with a shadow `CODEX_HOME` and `CODEX_SQLITE_HOME` set to the shared home, `doctor --json` reports `sqlite home` as the shared path and resolves all six databases there — `goals_1`, `logs_2`, `memories_1`, `queue_1`, `state_5`, `thread_history_1`.

The branch is rebased onto `5f878d2a8`. CI on the new head is sitting at `action_required` and needs an approval to run.

Implemented with Claude Opus 5 using the Claude Code harness in T3 Code.

ElliotDrel:
## Windows regression confirmation: regular sidecars still block startup

This reproduces the same sidecar-only shape described in #5817 after the provider had previously been recovered.

- Setup: shared `CODEX_HOME` plus a second authenticated shadow home on Windows 11.
- Current conflict: `goals_1.sqlite-shm`; the server logged the exact `already exists and is not a symlink` failure repeatedly on September 9, 10, and 11.
- Shadow-home state: `goals_1.sqlite` remains a symlink to the shared home, while `goals_1.sqlite-shm` and `goals_1.sqlite-wal` are regular files.
- Impact: the shadow provider is disabled. The desktop backend later exits with an unhandled `write EPIPE`, leaving the local environment endpoint unresponsive and the desktop app unable to open normally.

For immediate recovery, I am disabling only the affected provider instance. I am not deleting authentication or SQLite state. This is a temporary mitigation, not a fix.

This is a useful end-to-end acceptance case for this PR: an existing sidecar-only shadow conflict must no longer disable the provider or cascade into a desktop-backend crash. The PR's `CODEX_SQLITE_HOME` design looks intended to prevent this, but it has not shipped in the stable desktop build that exhibits the failure.

PR #1799, opened 2026-04-06
Improve shell PATH hydration and fallback detection

juliusmarminge:
Closes #1787

## Summary
- Add shared login-shell candidate resolution that checks the configured shell, the user shell, and platform defaults without duplicates.
- Hydrate `PATH` by merging shell-provided entries with inherited entries instead of replacing them outright.
- Add a macOS `launchctl` fallback when shell probing does not yield a `PATH` value.
- Expand desktop and server environment hydration to preserve additional Homebrew and XDG variables when available.
- Cover the new shell candidate, `launchctl`, and PATH merge behavior with tests in shared, desktop, and server layers.

## Testing
- `bun fmt`
- `bun lint`
- `bun typecheck`
- `bun run test`

---

> [!NOTE]
> **Medium Risk**
> Changes PATH and environment hydration behavior on macOS/linux (candidate shell probing, PATH merging, and launchctl fallback), which can affect command resolution and tooling discovery. Logging and additional env propagation reduce silent failures but broaden surface for platform-specific edge cases.
> 
> **Overview**
> **Improves PATH hydration for desktop and server on macOS/linux.** Instead of replacing `PATH` with the login-shell value, the code now *merges* login-shell (or macOS `launchctl`) entries with inherited entries, preserving extra paths while keeping login-shell precedence.
> 
> Shell probing now tries multiple candidates (env `SHELL` → detected user shell → platform default) and logs per-shell failures rather than silently swallowing errors. Desktop hydration also pulls through additional variables (`HOMEBREW_*`, `XDG_*`) when missing, with new/updated tests covering candidate selection, PATH merge behavior, and the `launchctl` fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99b6d62daaace3c6e988ce89baeabe722ba4db50. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Improve shell PATH hydration with multi-candidate fallback and launchctl support
> - Replaces single-shell resolution with `listLoginShellCandidates` in [`packages/shared/src/shell.ts`](https://github.com/pingdotgg/t3code/pull/1799/files#diff-89a91f85db41a9a0912bc86c336bf9429751f3797ee50a089248c68cde8489ce), iterating over `env.SHELL`, a `userShell`, and a platform fallback in order without duplicates.
[...]

PR #9232, opened 2026-09-02
fix(desktop): find user-installed CLIs when shell probing fails

### LINKED ISSUE #7618: [Bug]: macOS Source Control reports installed Homebrew gh as unavailable
### Before submitting

- [x] I included enough detail to investigate the problem.

### Area

apps/desktop

### Observed sequence

1. T3 Code 0.0.33 was installed as the packaged macOS app.
2. GitHub CLI was installed and authenticated:
   - `gh --version` succeeded in Terminal.
   - `gh auth status` showed an authenticated GitHub account.
3. In **Settings → Source Control**, GitHub appeared as unavailable with the instruction to install `gh`.
4. **Rescan server environment** did not make GitHub available.

This is an observation from one real packaged-app session, not yet a deterministic clean-environment reproduction.

### Expected behavior

Following the documented setup—installing `gh`, authenticating it, and rescanning the server environment—should make GitHub available in Source Control.

### Actual behavior

GitHub was shown as:

> Not available on this server: Install the GitHub command-line tool (`gh`).

T3's logs from the affected session contain:

```text
VcsProcessSpawnError: source-control.discovery.probe: gh
PlatformError: NotFound: ChildProcess.spawn (gh --version)
Error: spawn gh ENOENT
```

This establishes that Source Control could not resolve the `gh` executable and failed before GitHub authentication was evaluated.

### What has and has not been established

Established:

- The packaged app was T3 Code 0.0.33.
- `gh` 2.93.0 was installed and authenticated.
- A real executable was available at `/Users/toviastorres/.local/bin/gh`; Homebrew also exposed `/opt/homebrew/bin/gh`.
- The Source Control probe failed with `spawn gh ENOENT`.
- The documented **Rescan server environment** action did not recover discovery in the affected session.

Not yet established:

- The exact `PATH` seen by the original T3 server process was not captured directly.
- The outputs or failures of T3's login-shell and `launchctl` PATH probes were not captured.
- A local application modification and an application restart occurred together afterward, so the later successful discovery does not isolate whether the restart, the modification, or their combination changed the result.
- Therefore, a missing standard-directory fallback is a hypothesis, not a confirmed root cause.

### Related issues

- #1787 / #1799 address Homebrew-installed provider CLIs not being detected by the packaged macOS app. This may be a regression or an uncovered path through the same environment-hydration area.
- This differs from #7581, where T3 finds `gh` but reports the wrong authentication state.
- This differs from #3806, which involves a `gh` version too old to support the authentication JSON probe.
- This differs from #2187, which involves a shell alias and 1Password-based authentication.

### Impact

### Version or commit

0.0.33 (`3b72d17cbca691f0b64e6d4a10c9e349f42873a5`)

### Environment

- macOS 26.6.2
- Apple Silicon (arm64)
- Packaged T3 Code desktop app 0.0.33
- GitHub CLI 2.93.0
- GitHub authentication stored in the keyring

### Suggested investigation

It may help to log, alongside the failing Source Control probe:

- the effective server `PATH`;
- whether the login-shell probe succeeded, timed out, or returned an unusable value;
- the value returned by `launchctl getenv PATH`;
- which environment source was ultimately selected.

That would distinguish a shell-probe failure, environment propagation issue, stale process state, or provider-specific discovery issue without assuming a particular fix.

### Workaround status

No workaround was isolated conclusively. Restarting the app was not tested independently from the local application modification. The documented rescan action did not resolve the affected session.

CyberStefNef:
## What Changed

`resolveKnownPosixCliDirs` in `packages/shared/src/shell.ts` returns the standard CLI directories for macOS and Linux, beside the existing `resolveKnownWindowsCliDirs`. Both POSIX PATH hydration paths now fall back to it when neither the login shell nor `launchctl getenv PATH` answers: `installPosixEnvironment` in `apps/desktop/src/shell/DesktopShellEnvironment.ts` and `hydratePosixPath` in `apps/server/src/os-jank.ts`.

The fallback is last in the chain, so a shell that did answer still decides the order. Directories are not checked for existence, matching `resolveKnownWindowsCliDirs`.

## Why

A Finder launch starts the app with `/usr/bin:/bin:/usr/sbin:/sbin`. Both hydration paths repair that from the login shell, but the probe collapses failures and timeouts into an empty result, so a profile slower than the 5s timeout leaves PATH untouched. `launchctl getenv PATH` is unset on recent macOS, so the second probe returns nothing either. Every bare-name spawn then fails with ENOENT: #7618 counted 1246 failed `gh` spawns and zero successes over 14 hours, with `gh auth status` never running.

#1799 built this chain and added the `launchctl` probe as its last step. This adds the next step, for machines where that probe is also empty. Raising the timeout does not help, since the probe fails for reasons unrelated to how long it is given.

Both Windows paths already end in `resolveKnownWindowsCliDirs`. POSIX ended in nothing.

Closes #7618

## Verification

- `vp test run apps/desktop/src/shell/DesktopShellEnvironment.test.ts packages/shared/src/shell.test.ts apps/server/src/os-jank.test.ts`: 54 passed.
- Typecheck for `@t3tools/desktop`, `@t3tools/shared`, and the server.
- `vp lint` on the five changed files with `--report-unused-disable-directives`.
- `git diff --check`: clean.
- Seven added tests: the macOS and Linux fallbacks in both hydration paths, and guards that the known directories stay out of the merge when either probe answers. Reverting the source change fails the fallback tests and leaves the guards passing.

## Checklist

- [x] Screenshots are not applicable because nothing visual changed
- [x] A video is not applicable because no motion or interaction changed

---

> [!NOTE]
> **Medium Risk**
> Changes process `PATH` for all POSIX desktop and server startups when probes fail; behavior shifts from minimal system PATH to hard-coded install locations, which could affect spawn resolution in edge environments but addresses widespread CLI spawn failures.
> 
> **Overview**
> When login-shell and (on macOS) `launchctl` PATH probes return nothing—common for GUI/Finder launches with a minimal PATH—POSIX hydration no longer leaves only `/usr/bin:/bin:…`.
> 
> **Shared:** Adds `resolveKnownPosixCliDirs` (macOS Homebrew/`/usr/local`/`~/.local/bin`, Linux Linuxbrew/`/usr/local`/`~/.local/bin`; skips `~/.local` if `HOME` is blank), mirroring the existing Windows known-CLI helper and without checking that directories exist.
> 
> **Desktop & server:** `installPosixEnvironment` and `hydratePosixPath` use that list as the **last** preferred PATH segment before merging with the inherited PATH. Any PATH from the login shell or `launchctl` still wins and does not inject these defaults.
> 
> New unit tests cover macOS/Linux fallback merges and assert the known dirs are **not** added when either probe succeeds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10ead921397c0aed88c214de3047c5d1e8eb3f9d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add known CLI directory fallback to `hydratePosixPath` and `installPosixEnvironment`
> - Adds shared `resolveKnownPosixCliDirs` resolver that returns platform-specific CLI install directories (Homebrew on macOS, Linuxbrew on Linux, plus system-local and HOME-local `bin`).
> - Both desktop POSIX environment installation and server PATH hydration now prepend these known directories as a final fallback when neither the login shell nor `launchctl` produces a PATH.
> - A PATH from the login shell or `launchctl` remains preferred and is used directly without adding the fallback directories.
> - Risk: any caller relying on an empty PATH when probes fail will now receive a non-empty PATH composed of known CLI directories; the HOME-local entry is omitted when `HOME` is blank or unset.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 10ead92.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/shell/DesktopShellEnvironment.test.ts
+++ b/apps/desktop/src/shell/DesktopShellEnvironment.test.ts
@@ -299,6 +299,75 @@ describe("DesktopShellEnvironment", () => {
     }),
   );
 
+  it.effect("falls back to known CLI directories on macOS when neither probe returns a PATH", () =>
+    Effect.gen(function* () {
+      // The Finder-launch environment: no SHELL, and the bare PATH launchd hands a .app.
+      const env: NodeJS.ProcessEnv = {
+        HOME: "/Users/test",
+        PATH: "/usr/bin:/bin:/usr/sbin:/sbin",
+      };
+
+      yield* runShellEnvironment({
+        env,
+        platform: "darwin",
+        handler: () => "",
+      });
+
+      assert.equal(
+        env.PATH,
+        "/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/Users/test/.local/bin:/usr/bin:/bin:/usr/sbin:/sbin",
+      );
+    }),
+  );
+
+  it.effect("falls back to known CLI directories on linux when the login shell is silent", () =>
+    Effect.gen(function* () {
+      const env: NodeJS.ProcessEnv = { HOME: "/home/test", PATH: "/usr/bin:/bin" };
+
+      yield* runShellEnvironment({
+        env,
+        platform: "linux",
+        handler: () => "",
+      });
+
+      assert.equal(
+        env.PATH,
+        "/home/linuxbrew/.linuxbrew/bin:/usr/local/bin:/home/test/.local/bin:/usr/bin:/bin",
+      );
+    }),
+  );
+
+  it.effect("omits the known CLI directories when the login shell reports a PATH", () =>
+    Effect.gen(function* () {
+      const env: NodeJS.ProcessEnv = { SHELL: "/bin/zsh", HOME: "/Users/test", PATH: "/usr/bin" };
+
+      yield* runShellEnvironment({
+        env,
+        platform: "darwin",
+        handler: () => envOutput({ PATH: "/Users/test/.rvm/bin:/usr/bin" }),
+      });
+
+      assert.equal(env.PATH, "/Users/test/.rvm/bin:/usr/bin");
+    }),
+  );
+
+  it.effect("omits the known CLI directories when only launchctl reports a PATH", () =>
+    Effect.gen(function* () {
+      const env: NodeJS.ProcessEnv = { SHELL: "/bin/zsh", HOME: "/Users/test", PATH: "/usr/bin" };
+
+      yield* runShellEnvironment({
+        env,
+        platform: "darwin",
+        handler: (command) =>
+          command._tag === "StandardCommand" && command.command === "/bin/launchctl"
+            ? "/usr/local/opt/bin:/usr/bin"
+            : "",
+      });
+
+      assert.equal(env.PATH, "/usr/local/opt/bin:/usr/bin");
+    }),
+  );
+
   it.effect("loads PowerShell profile environment on Windows", () =>
     Effect.gen(function* () {
       const env: NodeJS.ProcessEnv = {
--- a/apps/desktop/src/shell/DesktopShellEnvironment.ts
+++ b/apps/desktop/src/shell/DesktopShellEnvironment.ts
@@ -8,6 +8,8 @@ import * as Schema from "effect/Schema";
 import * as ChildProcess from "effect/unstable/process/ChildProcess";
 import * as ChildProcessSpawner from "effect/unstable/process/ChildProcessSpawner";
 
+import { resolveKnownPosixCliDirs } from "@t3tools/shared/shell";
+
 import * as DesktopEnvironment from "../app/DesktopEnvironment.ts";
 
 type EnvironmentPatch = Record<string, string>;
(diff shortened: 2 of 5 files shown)


weskcode:
Hit the same bug (Claude Code and OpenCode installed via their native installers, at ~/.local/bin/claude and ~/.opencode/bin/opencode; neither found on PATH) and had built basically this same fix before finding this PR.

One directory missing from `resolveKnownPosixCliDirs` that would close the gap for OpenCode specifically: `~/.opencode/bin`. Claude Code is already covered here via the shared `~/.local/bin` entry.

Separately (more of a discussion than a one-line fix): this only fires via `Option.orElse` when both probes return nothing. `resolveKnownWindowsCliDirs`'s usage in `resolveWindowsEnvironment` merges its known dirs unconditionally, after the resolved PATH, regardless of whether the probe succeeded — the comment there even says it's meant to "fill gaps when desktop apps launch without the full interactive-shell PATH." Doing the same here would also catch a probe that succeeds but is just missing a directory. Not proposing this as part of the same patch, since it reverses a guard you explicitly tested for (the two tests asserting known dirs stay out when either probe succeeds) — just flagging the asymmetry with the Windows path in case it's worth a follow-up.

Happy to open a PR adding `~/.opencode/bin` on top of this branch, or hand it over if that's less work for you. juliusmarminge flagged this class of fix as a useful last-resort mitigation on #10534, just not enough on its own to close that issue.

PR #9242, opened 2026-09-02
fix(acp): generic ACP agents — slash commands, model picker, mode selection (stacked on #9060)

JDLondon7:
## Summary

Stacked on #9060 (configurable ACP agents); the first three commits are that PR. The rest are fixes found by running a real ACP agent (a ForgeCode fork with an ACP stdio server) through the UI. They are generic to the ACP driver, not forge-specific.

- **Slash commands.** Consume `available_commands_update` and publish it as the provider instance's `slashCommands`. Also read `_meta.availableCommands` from the initialize response in the availability probe, since ACP can only advertise commands per session and the `/` menu was otherwise empty until the first message.
- **Model picker.** Read `_meta.modelState` from the initialize response (as Grok already does) so an ACP provider's model picker is populated before a session exists. Prefer that list over config options observed from an earlier session, which can predate the agent build the probe just ran. Drop the synthetic "Agent default" entry once the agent reports models; show the current model first, then alphabetical.
- **Mode selection.** Aliases were matched against mode *descriptions*: an agent whose planning mode says "without code changes" matched the implement alias "code", and T3 silently put it in a read-only mode. Match ids and names only, never choose a plan mode as the working mode, and when no alias matches leave the agent's mode alone — the requested mode is re-applied every turn, which was reverting a mode the user chose through the agent.
- Kept `main`'s `initialize` probe name over #9060's `connect`; retained the PR's optional auth and config-option callbacks.

## Test plan

- `pnpm --filter ./apps/server typecheck` clean.
- `vp test run src/provider/acp src/provider/Layers/CursorAdapter.test.ts src/provider/Drivers`: 143 passed, 8 skipped.
- UI, against the ACP agent: slash menu populated in a fresh draft, commands run, model picker lists the agent's models and a selection is applied, agent mode switch persists across turns.

> [!NOTE]
> ### Add generic ACP agent driver with slash commands, model picker, and mode selection
> - Introduces `AcpDriver` as a sixth built-in provider driver for installed ACP-compatible CLIs, with `AcpSettings` schema, health checks, model discovery, and slash-command parsing from agent metadata
> - Refactors `CursorAdapter` into a shared `makeAcpAdapter` core driven by `AcpAdapterProfile`, so non-Cursor ACP providers get provider-specific runtime construction, model selection via config options or `SessionModelState`, and permission handling without Cursor resume cursors
> - Extends `AcpSessionRuntime` with optional auth, session close with 2s timeout, model-state tracking, config-option and available-command callbacks, reasoning-text content deltas, and JSON-RPC error mapping for authentication failures
> - Registers ACP Agent in web, mobile, docs, and provider metadata; excludes the `acp` driver from text-generation model controls in [GeneralSettingsPanel](https://github.com/pingdotgg/t3code/pull/9242/files#diff-9dbbb4f86928d777550dead352b372341ccfb3f589782aa6d99b98f792c89d18) and [SourceControlWritingSettings](https://github.com/pingdotgg/t3code/pull/9242/files#diff-4b5bbd6c145e7a9dd50ddae29cfb77abebd3fa85575e07cf436cd77263b7fecf)
> - Risk: `resolveRequestedModeId` removes fallback to arbitrary/current mode when no working-mode alias matches, so ambiguous mode requests that previously resolved silently may now fail; `applyRequestedSessionConfiguration` skips invalid enumerated config-option values instead of forwarding them
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 82fec90.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Large refactor of the shared ACP adapter path (Cursor included) plus new child-process sessions, model/mode application, and permission mapping—bugs could break sessions or mis-route model selections for ACP agents.
> 
> **Overview**
> Adds a **built-in `acp` driver** so users can point T3 Code at any ACP-compatible CLI (command + stdio args), with health probing, model/slash-command discovery, and explicit **no server-side text generation**.
> 
> The Cursor-specific ACP stack is refactored into a shared **`makeAcpAdapter` profile** (`CursorAdapter` becomes a thin `cursor` profile). Generic agents get **standard model selection** (config options or `session/set_model` for `SessionModelState`), **slash commands** from initialize `_meta` and `available_commands_update`, and **reasoning** streamed as `reasoning_text` deltas.
> 
> **Mode handling** is tightened for third-party agents: match modes by id/name only, never pick a plan mode as the working mode, and stop forcing a fallback when aliases do not match (avoids undoing agent-chosen modes each turn). **Permission replies** now use the agent’s real option IDs instead of synthesized strings.
> 
> `AcpSessionRuntime` gains optional auth, **`session/close`**, model-state tracking, config/command change callbacks, MCP capability gating, and retriable initialize caching. Web/mobile add **ACP Agent** settings, a plug icon, and **exclude ACP** from auxiliary text-generation model pickers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82fec901a0f6e64e8feb9cf16bca617995152cdb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


JDLondon7:
Pushed `0432eb2` addressing the automated review. Each finding was verified against the source before acting; here's the disposition.

**Fixed**
- **Plan-mode filter dropped valid working modes** — `isPlanMode` matched on the mode *description*, so an `agent` mode described "…plan…" was filtered out of `workingModes` and never selected. Now id/name-only.
- **Invalid standard config value failed the whole turn** — an unvalidated select value was forwarded to `setConfigOption` and rejected with `-32602`. String select values are now validated against the option's allowed values first (booleans left to the runtime).
- **Model selection dropped on whitespace** — display trims slugs, validation didn't. Now matches on trimmed values and sends the matching raw value.
- **Images dropped for capable agents** — embedding was gated on `promptCapabilities.image === true`, a regression from the shared-core refactor that silently dropped images (and failed image-only turns) for agents accepting images without advertising the flag. Now embeds unless `image === false`.
- **Failed `initialize` cached forever** — `Effect.cached` memoized the failure Exit, replaying it to every later `start()`/`initialize()` and contradicting `start`'s retry contract. Now `cachedInvalidateWithTTL`, invalidating on failure.
- **MCP toolkit dropped silently** — HTTP/SSE MCP servers (incl. T3's own `t3-code`) are filtered out when the agent omits the capability flag. Gate kept, but a warning is now logged so the loss isn't silent.
- **Withdrawn slash commands persisted** — `parseAdvertisedCommands` collapsed "absent" and "explicitly empty" to `[]`, and the caller only wrote on `length > 0`. Now returns `undefined` (absent) vs `[]` (empty) and writes whenever a list is advertised.
- **Web text-gen write handlers used the unfiltered provider list** — the resolver's fallback could persist an ACP instance as the text-generation model. Now uses the same ACP-filtered list as the read path.

Server + web typecheck clean; all ACP test suites pass (51 tests).

**Deferred — need a design decision (not mechanical)**
- **Generic `SessionModelState` models are inert in the standard path.** Confirmed: models advertised via `initialize._meta.modelState` are shown but the `"standard"` apply only handles `category:'model'` config options — it never calls `setSessionModel` — so `session.model` stays unset and the reactor tears the session down each turn. The fix needs the session model state threaded into `applyRequestedSessionConfiguration` (mirroring `GrokAcpSupport`'s `setSessionModel` usage), which is an interface change rather than a local edit — flagging for the preferred design before implementing.
- **ACP provider icon** is a wide wordmark in square slots; a square glyph is needed (the mobile app uses a plug). Left as a follow-up since it changes the brand mark.

**Not a bug**
- The `acceptForSession → allow_always` finding: "Always allow this session" is correctly encoded as `allow_always`; ACP has no session-scoped allow kind, and `allow_once` would break the promise. No change.

JDLondon7:
Followed up on the first deferred item (`bf443a1`): **generic `SessionModelState` models are now applied via `session/set_model`.**

- `AcpSessionRuntime` exposes `getModelState` (the `SessionModelState`), populated from the session setup response and updated after a successful `session/set_model` — mirroring the existing `getModeState`/`modeStateRef`.
- The `"standard"` apply path, when a selection matches no `category:"model"` config option, resolves it against the model state and applies it via `session/set_model` (only when it actually changes), recording `appliedModel` so `session.model` reflects it — so the reactor no longer treats each turn as a model change and tears the session down.
- Regression test added: selecting a SessionModelState-only model (`grok-mock-alt`) now sends `session/set_model` with that id and never misroutes it as a config-option write.

Typecheck clean; ACP suites pass (incl. the new test).

Still open: the ACP provider **icon** (square glyph — the wide wordmark renders as a sliver in square slots), and the **rebase onto current `main`**, which is a reconciliation with the separately-merged Antigravity ACP integration (#9348) and really belongs on the stack base #9060 before this PR rebases on top.

JDLondon7:
Icon finding resolved (`82fec90`): the ACP provider icon is now a **square plug glyph** instead of the ~2.6:1 wordmark that letterboxed to a sliver in the square icon slots — matching the mobile app, which already uses a plug for this driver. The unused `ACPRegistryIcon` wordmark is removed. Web typecheck clean.

That closes every actionable review finding on this PR. The only remaining item is the **rebase onto current `main`**, which is a reconciliation with the separately-merged Antigravity ACP integration (#9348). Since this PR is stacked on **#9060** (still open, authored upstream), that reconciliation belongs on #9060 first — this PR rebases cleanly once #9060 is current. Happy to help there if useful.

Issue #9245, opened 2026-09-02
Worktree path comparisons use exact string equality, so symlinked paths split identity on macOS


Mnigos:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

Server / checkpointing

### Steps to reproduce

1. On macOS, register a project whose root lives under `/tmp/...` (which is a symlink to `/private/tmp/...`).
2. Get two threads pointing at the same worktree directory, where one stored its `worktreePath` as `/tmp/foo` and the other as `/private/tmp/foo` (for example one created through a path the user typed and one through a resolved picker path).
3. Trigger the branch drift check after a turn completes.

### Expected behavior

Both threads are treated as sharing one worktree, so the drift check refuses to adopt a drifted checkout and any shared-worktree guard counts both references.

### Actual behavior

The comparison is raw string equality (`other.worktreePath === thread.worktreePath` in `CheckpointReactor.ts`), so the two spellings of the same directory count as different worktrees. The guard proposed in #9185 compares the same way in SQL (`worktree_path = ?`), so it would inherit the same blind spot: a removal that should be refused can slip through.

### Impact

Low in practice, most people keep projects under their home directory. But when it hits, the failure mode is the bad one these checks exist to prevent: a worktree removed or a branch adopted while another thread still points at the same directory through a different spelling. Normalizing with `realpath` before comparing (or at write time) would close it. Happy to send a small PR if this is worth fixing.

PR #9247, opened 2026-09-02
fix(server): treat symlinked worktree paths as the same directory

Mnigos:
The shared-worktree check in the branch-drift guard compares raw path strings, so on macOS `/tmp/foo` and `/private/tmp/foo` (a symlink) count as two different worktrees and a drifted checkout can be adopted while a sibling thread still points at the directory through the other spelling.

Both the session-cwd ownership check and the shared-worktree check now compare realpath results, falling back to the raw string when resolution fails. Regression tests use real symlinks (created as directory junctions, which need no symlink privilege on Windows) to cover shared siblings and a session cwd spelled differently from the recorded worktree. Dedicated worktrees follow branch drift; shared worktrees keep their recorded branch.

Validation: all 35 checkpoint reactor tests, server typecheck, and targeted lint pass. The session-cwd regression fails before its fix.

Related to https://github.com/pingdotgg/t3code/issues/9245: this covers the branch-drift comparisons; the shared-worktree removal guard proposed in #9185 is not part of this PR, so the issue stays open for it.

---

> [!NOTE]
> **Medium Risk**
> Changes orchestration metadata updates on turn completion when branch drift is detected; incorrect canonicalization could block legitimate drift-follow or allow unwanted updates, though fallback to raw paths limits regression risk.
> 
> **Overview**
> Fixes branch-drift adoption when two threads reference the **same** worktree through different path spellings (e.g. macOS `/tmp/...` vs `/private/tmp/...`, or an explicit symlink).
> 
> In `CheckpointReactor`, the guard that skips adopting a drifted checkout for **shared** worktrees used raw string equality on `worktreePath`. Sibling threads with alternate spellings of one directory were treated as separate worktrees, so thread-1 could still follow checkout drift after an agent rename.
> 
> Worktree paths are now compared via **`realPath`** (Effect `FileSystem`), falling back to the original string if resolution fails. The shared-worktree check canonicalizes the current thread path and every other thread’s path before deciding whether to skip `thread.meta.update`.
> 
> Tests add harness support for a second thread with a custom worktree path and a case where thread-2 points at thread-1’s cwd through a directory symlink, asserting the recorded branch stays `t3code/original-branch`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc818fbcb74b3de3f72dd61e2966921713af0bd0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix 'CheckpointReactor' to treat symlinked worktree paths as same directory
> During branch drift handling, `CheckpointReactor` now resolves worktree paths using `realPath` on the `FileSystem` service instead of comparing raw path strings, so symlinked paths are recognized as the same directory.
> - Updates the test harness `createHarness` with an optional callback to set the second thread's worktree path, and adds a regression test confirming a completed turn does not overwrite a locally detected branch when the worktree is shared via symlink.
> - Risk: if `realPath` resolution fails, the comparison falls back to the original path strings, which may cause the original branch drift bug to reoccur for symlinked paths in that edge case.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d3eb8bd.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved shared-worktree detection when equivalent paths use symlinks or alternate path spellings.
  - Prevented drifted checkouts from being incorrectly adopted when a worktree is shared.
  - Preserved the original thread branch in shared-worktree scenarios.
  - Ensured unshared drifted checkouts can still be adopted and pull requests refreshed.
  - Improved worktree path resolution across supported environments.


lnieuwenhuis:
Maintainer verification: narrow and additive — realPath canonicalization applies only to the shared-worktree branch-drift guard, and resolution failures fall back to plain string equality, so the worst case is the old behavior. Symlink regression included and full CI is green. Looks merge-ready to me.

shivamhwp:

This fixes canonical path comparisons in checkpoint branch-drift handling. #9245 also covers the shared-worktree removal guard in #9185, which this PR does not change. Please keep the issue open for that remaining guard or split its tracking before using `Fixes #9245`. The new symlink tests should use a Windows-compatible directory junction or an explicit platform guard so Windows runners do not require symlink privileges.

Mnigos:
@shivamhwp done in 8f97af2: dropped the Fixes keyword (removal guard from #9185 stays tracked in #9245) and the tests now create junctions.

Mnigos:
> [!NOTE]

Closing for now: this touches the server layer that is being rewritten for V2, and the maintainers are not taking changes there until it lands. The fix is still needed against current `main`; I will reopen or resubmit it against the new code once V2 is in.

PR #9304, opened 2026-09-03
fix(web): fall back to local mode for empty repos

### LINKED ISSUE #2148: [Bug]: Creating a thread in an empty git repository (no commits) crashes with "select a thread"
## Area

apps/web, apps/server

## Steps to reproduce

1. Open T3 Code (desktop app or dev server)
2. Click the **+** button or use the command palette to add a new project
3. In the Finder/file picker, select (or create) a **brand new, empty directory**
4. Initialize it as a git repo (`git init`) — but **do not create any commits**
5. The project is linked and appears in the sidebar
6. Click "New thread" (or it auto-navigates to the draft composer)
7. Type any message and press Enter/Send

## Expected behavior

The thread should be created and the first turn should start. Since the repo has no commits and worktree mode can't work, the app should gracefully fall back to local mode or show a clear error explaining that the repo needs at least one commit for worktree mode.

## Actual behavior

The app instantly crashes/fails with **"Select an existing thread or create a new one to get started"** — the user is stranded on a dead route with no way to recover except navigating away.

### Root cause trace

1. An empty git repo (`git init`, no commits) reports `isRepo: true` but `branch: null` from `git status --porcelain=2 --branch` (the branch line reads `# branch.head (initial)`, which is parsed as `null` in `GitCore.ts:1260`).

2. `resolveSendEnvMode()` in `ChatView.logic.ts` only checks `isGitRepo` — since the repo *is* a git repo, it keeps the user's default `"worktree"` mode.

3. When the message is sent, the bootstrap flow calls `git.createWorktree()` on the server (`ws.ts:456`). This runs `git worktree add`, which **fails because there are no commits** to branch from.

4. The error handler (`ws.ts:478-484`) calls `cleanupCreatedThread()`, which dispatches `thread.delete`.

5. On the client, the event cascade:
   - `thread.created` → marks the draft as promoted (`composerDraftStore.ts`)
   - `thread.deleted` → `clearDraftThread` removes the draft entirely
   - The draft route component (`_chat.draft.$draftId.tsx:49-53`) finds both `draftSession` and `canonicalThreadRef` are null → navigates to `/`, showing the "select a thread" dead state.

## Impact

Blocks work completely

Any user who adds a brand new project directory and tries to use it immediately will hit this crash. There is no in-app indication of what went wrong.

## Version or commit

main @ a3b1df52 (v0.0.20+)

This has been present since at least the worktree mode feature was introduced. The deleted branch `t3code/handle-missing-directories` appeared to target this but was removed before merging.

## Environment

macOS 15.7.3, Desktop app (Electron), Bun 1.3.9, Node 24.14.1. Reproducible with any provider (Codex, Claude).

## Workaround

Before adding the project to T3 Code, create an initial commit in the empty repo:

```bash
cd /path/to/new-project
git init
git commit --allow-empty -m "initial commit"
```

Alternatively, a fix is to have `resolveSendEnvMode()` check branch availability and fall back to `"local"` mode when the branch is `null`:

```ts
// ChatView.logic.ts
export function resolveSendEnvMode(input: {
  requestedEnvMode: DraftThreadEnvMode;
  isGitRepo: boolean;
  branch?: string | null;
}): DraftThreadEnvMode {
  if (!input.isGitRepo) return "local";
  // Empty repos (git init, no commits) report isRepo=true but branch=null.
  // Worktree mode requires at least one commit, so fall back to local.
  if (input.requestedEnvMode === "worktree" && input.branch === null) return "local";
  return input.requestedEnvMode;
}
```

And pass `gitStatusQuery.data?.branch ?? null` at the call site in `ChatView.tsx`.

Michel-Liao:
## Problem

Empty Git repositories report as repositories but have no branch to seed a worktree. Draft submission still selected worktree mode, so the server failed creation, deleted the promoted thread, and left the user on an empty route.

Closes #2148.

## Fix

Pass the resolved Git ref into the environment-mode decision and fall back to local mode only when status has resolved with no branch. While status is still loading, the user’s requested mode is preserved.

## Test

- `vp test run apps/web/src/components/ChatView.logic.test.ts` (79 passed)
- targeted lint and formatting checks for the three changed files

Generated with GPT-5.6-sol in T3 Code using the Codex harness.

---

> [!NOTE]
> **Low Risk**
> Small, targeted change to draft send env resolution with explicit tests; no auth or server contract changes.
> 
> **Overview**
> Fixes draft sends in **empty Git repos** (initialized but no commits): worktree mode is no longer used when git status has resolved with **no branch**, so thread creation does not fail and leave users on a broken route.
> 
> `resolveSendEnvMode` now takes an optional `branch` and forces **local** when worktree is requested but `branch === null`, in addition to non-git folders. **ChatView** passes `gitStatusQuery.data?.refName`. While status is still loading (`branch` omitted), the user’s requested env mode is left unchanged.
> 
> Tests cover repos with a branch, empty repos, non-git paths, and the loading case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7bd8ff2181da276e9fb26b8d2a1242b8c97e769e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/web/src/components/ChatView.logic.test.ts
+++ b/apps/web/src/components/ChatView.logic.test.ts
@@ -824,9 +824,21 @@ describe("getStartedThreadModelChangeBlockReason", () => {
 });
 
 describe("resolveSendEnvMode", () => {
-  it("keeps worktree mode only for git repositories", () => {
+  it("keeps worktree mode only for repositories with a commit", () => {
+    expect(
+      resolveSendEnvMode({ requestedEnvMode: "worktree", isGitRepo: true, branch: "main" }),
+    ).toBe("worktree");
+    expect(
+      resolveSendEnvMode({ requestedEnvMode: "worktree", isGitRepo: true, branch: null }),
+    ).toBe("local");
+    expect(
+      resolveSendEnvMode({ requestedEnvMode: "worktree", isGitRepo: false, branch: null }),
+    ).toBe("local");
+  });
+
+  it("does not override the requested mode while git status is loading", () => {
+    expect(resolveSendEnvMode({ requestedEnvMode: "local", isGitRepo: true })).toBe("local");
     expect(resolveSendEnvMode({ requestedEnvMode: "worktree", isGitRepo: true })).toBe("worktree");
-    expect(resolveSendEnvMode({ requestedEnvMode: "worktree", isGitRepo: false })).toBe("local");
   });
 });
 
--- a/apps/web/src/components/ChatView.logic.ts
+++ b/apps/web/src/components/ChatView.logic.ts
@@ -458,8 +458,12 @@ export function readFileAsDataUrl(file: File): Promise<string> {
 export function resolveSendEnvMode(input: {
   requestedEnvMode: DraftThreadEnvMode;
   isGitRepo: boolean;
+  branch?: string | null;
 }): DraftThreadEnvMode {
-  return input.isGitRepo ? input.requestedEnvMode : "local";
+  if (!input.isGitRepo || (input.requestedEnvMode === "worktree" && input.branch === null)) {
+    return "local";
+  }
+  return input.requestedEnvMode;
 }
 
 export function resolveBackgroundDraftWorkspaceOptions(input: {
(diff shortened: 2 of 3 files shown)


Michel-Liao:
Withdrawing this version because the review is correct: refName does not distinguish an unborn HEAD from detached HEAD, so the client-only fallback could misclassify both states. A proper fix needs an explicit repository-status signal rather than inferring from the branch field.

PR #9360, opened 2026-09-03
fix(server): resume active turns when T3 Code reopens

danieliser:
## User-facing behavior

This PR fixes one specific desktop-app failure:

1. A model is still working in a T3 Code thread.
2. The user fully quits T3 Code.
3. The user opens T3 Code again.

Previously, the reopened app treated that active turn as an orphaned provider session and stopped it with an error. The user had to send another message manually.

With this change, reopening T3 Code recognizes the interrupted active turn, restores its saved provider conversation, and asks that conversation to continue.

> [!IMPORTANT]
> This is restart recovery, not background execution. Fully quitting T3 Code still stops its local backend and provider processes. The model does not continue computing while the app is closed; it resumes after the app is opened again.

## Why this code lives in the server

The T3 Code desktop app starts an embedded local backend. Quitting the app stops that backend, and reopening the app starts a new backend process. The recovery decision therefore happens during backend startup even though the user-facing action is simply quitting and reopening the desktop app.

The same startup logic also benefits standalone T3 Code server processes, but the motivating and tested user flow is the desktop app restart described above. Reloading only the browser or frontend is unrelated.

## How recovery works

At startup, an interrupted turn becomes a candidate for automatic continuation when:

- the stored thread session still has an active turn ID;
- the provider binding has a saved resume cursor;
- the thread is not archived or deleted; and
- there is no conflicting valid self-update continuation marker.

A previously cleared marker (`continueAfterServerUpdate: null`) counts as no marker. A valid marker containing a different turn ID remains authoritative and prevents the wrong turn from being resumed.

Before starting background recovery, T3 Code persists the interrupted turn ID as a continuation marker. If the app quits again during recovery, the next launch can retry instead of orphaning the turn. The marker is cleared only after the provider accepts the continuation.

Provider recovery is verified before any continuation turn is sent. Each adapter must prove that the live provider session corresponds to the saved resume cursor. Claude waits for the SDK to report its actual durable session ID instead of trusting the requested ID echoed during startup; OpenCode validates the session returned by its API; ACP providers require a successful native session load. A malformed cursor, a failed native load, or a confirmed identity mismatch fails closed rather than continuing in the wrong conversation.

Once verified, the existing continuation path is reused:

- providers with native promptless continuation, currently Codex, receive a continuation turn;
- other providers receive the existing “Continue where you left off” fallback prompt.

If continuation fails, the thread reports that it could not continue after T3 Code restarted and asks the user to send a new message.

## Relationship to existing work

#9167 added continuation for a managed T3 Code self-update. That flow writes an explicit continuation marker before restarting. An ordinary desktop-app quit does not write that marker, which is the gap fixed here.

Other open PRs address different behavior:

- #7463 settles stale sessions instead of automatically continuing them;
- #9189 offers a manual Continue action after a turn is already interrupted;
- #9211 releases an existing Codex writer during cross-instance resume.

This also restores the user-visible quit/reopen behavior covered downstream by [patroza/t3code#442](https://github.com/patroza/t3code/pull/442) and [patroza/t3code#443](https://github.com/patroza/t3code/pull/443), while reusing upstream’s continuation path instead of importing their larger recovery subsystem.

## Out of scope

- Keeping models running while the desktop app is closed
- Adding a macOS LaunchAgent or other background service
- Resuming an exact half-finished local tool process
- Changing frontend-only reload behavior

## Verification

- `vp test run src/provider/Layers/ClaudeAdapter.test.ts src/provider/Layers/ProviderService.test.ts src/serverRuntimeStartup.reconcile.test.ts` — 136 tests passed
- `vp run typecheck` in `apps/server` — passed (existing suggestions only)
- Targeted `vp lint` across the changed server files — passed (one pre-existing warning)
- Targeted formatting across the changed server files — passed
- `git diff --check` — passed
- Live Grok ACP probe confirmed that loading a nonexistent session fails instead of creating a fresh conversation
- Cursor Bugbot summary on commit `bc0226f7e` includes the durable Claude session-ID guard

Regression coverage includes absent and cleared self-update markers, crash-safe retry state, successful same-conversation recovery, and rejection of a fresh replacement conversation.

## Checklist

- [x] This PR remains focused on restart recovery
- [x] No UI changes

---

> [!NOTE]
> **Medium Risk**
> Recovery now auto-continues more orphaned sessions at startup; a cursor mismatch fails closed (session stopped, user must send again), but wrong eligibility could still attempt continuation on edge cases.
> 
> **Overview**
> Reopening T3 Code after a full quit can now **automatically continue an interrupted active turn** instead of orphaning the provider session, when the thread still has an active turn ID and a persisted resume cursor (and no conflicting self-update continuation marker).
> 
> **Provider recovery is gated by resume-cursor equality.** Adapters gain optional `isSameResumeCursor`; `ProviderService.sendTurn` accepts `requireResumeCursor` and rejects recovery when the live or newly started session does not match the saved conversation (stopping the bad session on fresh-start mismatch). Claude’s check waits for the SDK’s durable `session_id` confirmation so an echoed cursor on a brand-new session is treated as a failed resume.
> 
> **Startup reconciliation** widens continuation eligibility beyond the explicit update marker, persists a retry marker for crash-safe recovery, and issues continuation turns with `requireResumeCursor` (promptless where supported). User-facing failure copy now refers to T3 Code restarting rather than a server update.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc0226f7ec0f5f592e6adf6a7572b1871245f4f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Resume active turns when T3 Code reopens by validating provider resume cursors
> - On restart, orphaned active sessions with an active turn and persisted resume cursor now automatically resume, even without a continuation marker. The persisted cursor is passed as a strict requirement to `ProviderService.sendTurn`.
> - Adds an optional `isSameResumeCursor` comparison operation to `ProviderAdapterShape`; each adapter (Claude, Codex, Cursor, Grok, OpenCode) implements it. Claude additionally waits up to 10 seconds for the SDK to confirm the durable session id.
> - `ProviderService.recoverSessionForThread` and `resolveRoutableSession` now reject sessions whose cursor does not match the required persisted cursor; a mismatched recovered session is stopped and its MCP state cleared.
> - Behavioral Change: `reconcileProviderSessions` in [serverRuntimeStartup.ts](https://github.com/pingdotgg/t3code/pull/9360/files#diff-1930335d6e319fc5766e0578303e19f9d788e9990f87604b77f9f768e61d7118) now resumes unmarked orphaned sessions with an active turn and persisted cursor. Adapters without a resume-cursor comparator are treated as unable to prove recovery and will not auto-continue. Explicit markers whose turn id does not match remain ineligible.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bc0226f.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
+++ b/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
@@ -4247,6 +4247,52 @@ describe("ClaudeAdapterLive", () => {
     );
   });
 
+  it.effect("rejects an echoed Claude resume cursor when the SDK starts a fresh session", () => {
+    const harness = makeHarness();
+    return Effect.gen(function* () {
+      const adapter = yield* ClaudeAdapter;
+      const persistedResumeCursor = {
+        threadId: RESUME_THREAD_ID,
+        resume: "550e8400-e29b-41d4-a716-446655440000",
+        turnCount: 3,
+      };
+      const session = yield* adapter.startSession({
+        threadId: RESUME_THREAD_ID,
+        provider: ProviderDriverKind.make("claudeAgent"),
+        resumeCursor: persistedResumeCursor,
+        runtimeMode: "full-access",
+      });
+      assert.notEqual(adapter.isSameResumeCursor, undefined);
+      if (adapter.isSameResumeCursor === undefined) return;
+
+      const verification = yield* adapter
+        .isSameResumeCursor(RESUME_THREAD_ID, persistedResumeCursor, session.resumeCursor)
+        .pipe(Effect.forkChild);
+      harness.query.emit({
+        type: "system",
+        subtype: "init",
+        apiKeySource: "none",
+        claude_code_version: "test",
+        cwd: "/tmp/claude-adapter-test",
+        tools: [],
+        mcp_servers: [],
+        model: SYNTHETIC_CLAUDE_STANDARD_MODEL,
+        permissionMode: "bypassPermissions",
+        slash_commands: [],
+        output_style: "default",
+        skills: [],
+        plugins: [],
+        session_id: "7368d0c7-40a3-4d8a-bcc1-ac80c49f2719",
+        uuid: "fresh-init",
+      } as unknown as SDKMessage);
+
+      assert.equal(yield* Fiber.join(verification), false);
+    }).pipe(
+      Effect.provideService(Random.Random, makeDeterministicRandomService()),
+      Effect.provide(harness.layer),
+    );
+  });
+
   it.effect("preserves durable resume ids across Claude resume hooks", () => {
     const harness = makeHarness();
     return Effect.gen(function* () {
@@ -4259,7 +4305,7 @@ describe("ClaudeAdapterLive", () => {
         Effect.forkChild,
       );
 
-      yield* adapter.startSession({
+      const session = yield* adapter.startSession({
         threadId: RESUME_THREAD_ID,
         provider: ProviderDriverKind.make("claudeAgent"),
         resumeCursor: {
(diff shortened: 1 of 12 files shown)


juliusmarminge:
Superseded by #9803 (recover opted-in threads after machine restarts), which landed the server-side continuation/recovery path this PR targeted.

PR #4559, opened 2026-07-26
Target server settings by environment

colonelpanic8:
## What Changed

- Add one shared environment selector across General, Providers, Keybindings, Source Control, and Diagnostics settings.
- Route server-backed settings to the selected environment while keeping client-only preferences in local client storage.
- Use the active saved environment for global keybindings, provider updates, and server presentation when no primary environment exists.
- Resolve new-thread defaults, worktree behavior, editors, and keybindings from the project or thread environment.
- Incorporate per-environment provider configuration and environment-scoped diagnostics behavior.
- Disable server-backed controls while the selected environment is disconnected.

## Why

Client-only desktop mode deliberately has no `PrimaryConnectionTarget`, but several settings surfaces still read and wrote exclusively through primary-server atoms and hooks. Those controls therefore displayed defaults and silently failed to persist.

Settings now resolve an explicit selection first, then the managed primary environment, the active saved environment, and finally the first available environment. This preserves existing managed-client behavior while making client-only and multi-environment operation explicit.

## UI Changes

- General Settings now starts with a Server Settings environment row showing connection state and the selected environment.
- Providers, Keybindings, Source Control, and Diagnostics use the same shared selection.
- Client-local settings remain available when an environment is disconnected; server-backed controls are disabled until it reconnects.
- Provider accounts and authorization state are shown only for the selected environment.

This PR is stacked on `desktop-client-only-discovery`, the branch behind pingdotgg/t3code#4444. It also carries the environment-scoped provider and diagnostics work needed by the complete settings flow.

## Checklist

- [x] `vp test run apps/web/src/components/settings/SettingsPanels.logic.test.ts apps/web/src/components/settings/DiagnosticsSettings.logic.test.ts apps/web/src/hooks/useSettings.test.ts` — 31 tests passed
- [x] Targeted `vp lint --report-unused-disable-directives` for all changed files
- [x] `vp run --filter @t3tools/web build`
- [x] Integrated authenticated browser verification across General, Providers, Keybindings, Source Control, and Diagnostics
- [x] Confirmed assistant output, provider update checks, and the add-project base path persist through reload
- [ ] `vp run --filter @t3tools/web typecheck` remains blocked by pre-existing unrelated errors in `BranchToolbarBranchSelector`, `ModelPickerContent`, `PreviewAutomationHosts`, and the atom command/query helpers

---

> [!NOTE]
> **High Risk**
[...]

PR #4496, opened 2026-07-25
Let clients without a managed backend manage pairing

colonelpanic8:
## What Changed

Access management (pairing links, authorized clients, and the pairing QR code) is now scoped to whichever server the Connections page is administering, instead of always the primary environment.

- `resolveAccessEnvironment()` administers the primary when one exists, and the active environment otherwise. A desktop that manages its own backend is unaffected.
- New `packages/client-runtime/src/state/authHttp.ts` routes `/api/auth/session`, `/api/auth/pairing-token`, `/api/auth/pairing-links/revoke`, `/api/auth/clients/revoke`, and `/api/auth/clients/revoke-others` through the environment supervisor's prepared connection.
- `environmentPairingBaseUrl()` produces a shareable pairing URL only for a directly addressable HTTP origin. Relay and SSH environments fall back to showing the bare pairing code.
- Pairing no longer requests an explicit scope set, so a client keeps exactly what its credential carries.
- `t3 pair` prints the same scannable QR code that `t3 serve` already prints at startup.

## Why

`platformConnectionSourceLayer` returns no registrations when `isHostedStaticApp()`, so a hosted-app client has no primary environment at all. `canManageLocalBackend` reads scopes from the primary session, which is `null` in that case, so the entire "Authorized clients" section was hidden. A user who had paired the hosted app to their own server could not create a pairing link, revoke a client, or display a pairing QR code for that server.

Worse, the page rendered a contradictory "Administrative access — pairing links and client-session management require the access:write scope for this backend" row even when the client had paired using the server's own administrative startup token.

That second part had its own cause: onboarding always requested the standard client scopes, and the server rejects a wider request rather than clamping it, so a client bootstrapped from a server's startup token was silently downgraded to a standard session and could not manage the access it had just been handed. Requesting no scopes and accepting what the credential carries fixes it.

The access-management endpoints were already per-environment; only the client bound them to the primary.

## UI Changes

Hosted web client (no primary environment, served from a non-loopback origin on `:53595`) paired to a
standalone `t3 serve` on `:53141`, at `Settings → Connections`. Same viewport, theme, and backend for
every shot; only the `apps/web` build was swapped between commits.

> Note: these captures were taken at `b38c6b5`. The branch has since been rebased onto upstream main as `f14509692ccb`; the rebase changed no file content, so every image below still reflects the current head exactly.

**Before** (`38cfc25`, main) — no "Authorized clients" section at all, so there is no way to create a
[...]

PR #9361, opened 2026-09-03
Let the desktop app be a client of an existing T3 Code server

### LINKED ISSUE #6097: [Bug]: Desktop starts a second backend against the background service database

colonelpanic8:
# Let the desktop app be a client of an existing T3 Code server

Fixes #6097. Supersedes #4444 and folds in #4496 and #4559.

Based on #2829's branch (`t3code/codex-turn-mapping`) rather than `main`: the new orchestrator rewrites the client state and settings surfaces this change touches, so a `main`-based version would only conflict with it. The same series against `main` is kept at `colonelpanic8/t3code:t3code/client-environment-suite-main` for reference.

## The problem

Since `t3 service` (#4286, #6286) a T3 Code server can outlive the desktop app. But the desktop app still assumes it owns the machine's backend: on launch it sees port 3773 busy, picks 3774, and starts a second backend against the same `~/.t3/userdata/state.sqlite`. Two servers then contend for one SQLite database and requests fail with `database is locked` (#6097). Users in that thread asked for the obvious alternative, a thin client, and explained why the browser PWA is not a substitute: browser hotkeys conflict and several settings are unavailable outside the app.

The mobile app already works this way. It owns no backend and pairs with whatever environments it is given. This PR gives the desktop app the same option.

## What changes

**The desktop app never starts a second backend against a state directory that already has one.** On launch it reads the `server-runtime.json` a live server persists next to its database, exactly as `t3 pair` does, confirms the pid is alive and the process answers with the environment descriptor recorded next to it, and if so runs this launch as a client of that server instead of spawning its own. This is the same discovery `t3 pair` and `t3 triage` already rely on, so there is no new on-disk format and no new HTTP surface. Connections explains why the app is in this state and offers to pair with the running server.

**Client-only mode.** Users can also opt into this permanently (Settings → Connections → Backend mode) or per launch (`--backend-mode client-only`). In client-only mode the app serves its packaged renderer from static assets, opens the window when the renderer is ready rather than when a backend is, and does not stop any backend on quit. A client-only app with no saved connections lands on Connections instead of a dead primary-backend state.

**Pairing with a server on this computer.** Connections lists servers discovered through their runtime state. Pairing mints a one-time credential by running the bundled server's `t3 pair --json` against that state directory, which writes into the running server's own database. Same-user proof is filesystem access to the state directory, which is the trust model `t3 pair` already uses. Works on every platform.

**Settings target an environment.** A client with no backend of its own has nothing to call "primary", so settings that live on a server (environment general settings, providers, source control, keybindings, diagnostics) are addressed to an explicitly selected environment through one shared header that shows the same loading/offline/error state everywhere, and settings that belong to the client (general, appearance) are grouped separately in the navigation. Restore-to-defaults resets only the keys a page owns. Any connected client can manage pairing links and authorized clients for that environment, not only for a locally spawned one.

## Commits

The series is squashed onto the orchestrator branch as one commit; the original seven-commit series is on the `main`-based reference branch and is meant to be read in that order.

1. `feat: client-only desktop mode and environment-scoped settings`, the mode, IPC, static renderer serving, web routing, presence scope, and settings pages.
2. `shared: move server runtime state reading into packages/shared`, so the desktop can read runtime state without depending on the server package.
3. `cli: add --json to t3 pair`, following the existing `t3 auth ... --json` convention.
4. `desktop: discover local servers through runtime state and pair via t3 pair`, replacing an earlier Linux-only advertisement scheme that needed a new unauthenticated endpoint.
5. `desktop: attach to a running server instead of spawning a second backend`, the #6097 fix.
6. `web: unify environment-scoped settings selection and presentation`.
7. `web: drop per-environment accent colors and display names`, cosmetic client preferences split to a follow-up.

## Prior art

Paseo, another local-first agent runner, settled on this shape: the daemon runs detached, the desktop app reuses a daemon that is already running rather than starting another, and a `manageBuiltInDaemon` setting turns the desktop into a pure client. Its desktop shells out to the bundled CLI for `daemon status --json` the way this PR shells out to `t3 pair --json`.

## Not in this PR

Per-environment accent colors and custom display names, which were part of earlier revisions, are split out to keep this reviewable. Auto-pairing on attach is a one-line follow-up once the approach is agreed; this PR asks for one explicit Pair click.

## Verification

- `vp run --filter @t3tools/desktop --filter t3 --filter @t3tools/web --filter @t3tools/contracts --filter @t3tools/client-runtime --filter @t3tools/shared typecheck`: clean.
- `vp lint --report-unused-disable-directives` and `vp fmt --check` over every changed file: clean.
- Focused `vp test run` over apps/desktop (app, ipc, settings, electron, backend pool, exposure, updates, window), apps/server (cli, auth, startup access, server), apps/web (settings, environments, presence, grouping, hooks), packages/client-runtime (connection), packages/shared, packages/contracts: 132 files, 1307 tests, all passing.

---

> [!NOTE]
> **High Risk**
> Touches desktop startup, shutdown, custom protocol static serving, and local pairing subprocesses—areas where mistakes could cause double backends, failed launches, or unsafe pairing links.
> 
> **Overview**
> The desktop app can run as a **thin client** instead of always spawning its own backend. **Managed** vs **client-only** mode is persisted in settings, overridable with `--backend-mode`, and on packaged launch the app **auto-switches to client-only** when it finds a live userdata server via existing `server-runtime.json` discovery—avoiding a second process on the same SQLite state.
> 
> In **client-only** mode, startup skips the backend pool, serves the renderer via a new **static** custom protocol (or dev proxy), opens the window on **renderer readiness**, and does not stop backends on quit. **Connections** gains discovery of local servers and pairing by shelling out to **`t3 pair --json`**, with strict validation of pairing URLs and origins.
> 
> **IPC/preload** expose backend mode state, mode changes (relaunch with rollback on failure), and local-server discover/pair. The web UI uses **environment presence scope** (whether the app owns a local backend) for remote indicators and sidebar grouping, plus helpers for pairing candidates in Connections settings.
> 
> Supporting changes: shared **server runtime state** helpers, server config path derivation by dev/userdata variant, and **`--json`** on the `pair` CLI for machine-readable output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82443106f8503bb8f3ad53fa8c2fcf9723ba6f45. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add desktop `client-only` backend mode to connect to an existing T3 Code server
> - Adds a `DesktopBackendMode` service supporting `managed` (default) and `client-only` modes, with CLI parsing (`--backend-mode`), persisted settings, and automatic fallback to `client-only` when a packaged launch detects an already-running userdata server
> - Adds `DesktopRunningLocalServers` service for discovering running local T3 servers from persisted runtime state and pairing with a selected environment via an Electron-as-Node child process
> - Adds IPC channels and handlers for backend-mode state reads/updates, local-server discovery, and local-server pairing, exposed to renderers through the desktop bridge
> - Introduces the full Orchestration V2 system: new provider adapters (Claude, Codex, Cursor, Grok, ACP Registry, OpenCode), projection/event/command-receipt stores, thread lifecycle/fork/launch/settlement services, checkpoint capture/rollback, effect outbox/worker, scheduled tasks, legacy V1 thread importer, and MCP orchestrator/worktree toolkits — replacing the V1 orchestration surface across server, web, and mobile clients
> - Migrates client-runtime connection to perform remote descriptor discovery with orchestration protocol negotiation and bounded thread snapshot loading
> - Risk: `EnvironmentApi` replaces the V1 orchestration dispatch/snapshot/stream surface with `orchestrationV2`; out-of-tree consumers of the prior V1 interface will break. The server no longer advertises `serverUpdateThreadContinuation` capability, meaning running threads are terminalized on restart. Cursor provider settings no longer expose `binaryPath` or `apiEndpoint` fields.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8244310.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/app/DesktopApp.ts
+++ b/apps/desktop/src/app/DesktopApp.ts
@@ -1,9 +1,12 @@
 import * as Cause from "effect/Cause";
 import * as Effect from "effect/Effect";
+import * as FileSystem from "effect/FileSystem";
 import * as Option from "effect/Option";
+import * as Path from "effect/Path";
 import * as Ref from "effect/Ref";
 import * as Schema from "effect/Schema";
 
+import type { DesktopBackendMode as DesktopBackendModeValue } from "@t3tools/contracts";
 import * as NetService from "@t3tools/shared/Net";
 import * as Crypto from "effect/Crypto";
 import * as ElectronApp from "../electron/ElectronApp.ts";
@@ -13,6 +16,7 @@ import * as ElectronSafeStorage from "../electron/ElectronSafeStorage.ts";
 import { installDesktopIpcHandlers } from "../ipc/DesktopIpcHandlers.ts";
 import * as DesktopAppActivation from "./DesktopAppActivation.ts";
 import * as DesktopAppIdentity from "./DesktopAppIdentity.ts";
+import * as DesktopBackendMode from "./DesktopBackendMode.ts";
 import * as DesktopClerk from "./DesktopClerk.ts";
 import * as DesktopApplicationMenu from "../window/DesktopApplicationMenu.ts";
 import * as DesktopWindow from "../window/DesktopWindow.ts";
@@ -22,6 +26,7 @@ import * as DesktopLifecycle from "./DesktopLifecycle.ts";
 import * as DesktopLinuxUrlHandler from "./DesktopLinuxUrlHandler.ts";
 import * as DesktopObservability from "./DesktopObservability.ts";
 import * as DesktopPreReadyPlatform from "./DesktopPreReadyPlatform.ts";
+import * as DesktopRunningLocalServers from "./DesktopRunningLocalServers.ts";
 import * as DesktopShutdown from "./DesktopShutdown.ts";
 import * as DesktopServerExposure from "../backend/DesktopServerExposure.ts";
 import * as DesktopAppSettings from "../settings/DesktopAppSettings.ts";
@@ -59,6 +64,17 @@ export class DesktopDevelopmentBackendPortRequiredError extends Schema.TaggedErr
 ) {
   override get message(): string {
     return "T3CODE_PORT is required in desktop development.";
+  }
+}
+
+export class DesktopRendererAssetsUnavailableError extends Schema.TaggedErrorClass<DesktopRendererAssetsUnavailableError>()(
+  "DesktopRendererAssetsUnavailableError",
+  {
+    candidates: Schema.Array(Schema.String),
+  },
+) {
+  override get message(): string {
+    return `The packaged desktop renderer was not found. Checked: ${this.candidates.join(", ")}.`;
   }
 }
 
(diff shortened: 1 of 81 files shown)


colonelpanic8:
Continued as #9376 from the same commit (82443106f) on the branch named for its base; GitHub closed this one when the head branch was renamed.

PR #9372, opened 2026-09-03
fix(mobile): remember model options changed in existing threads

lukasbxy:
## What Changed

Changing the model or provider options in an existing Mobile thread updated only the draft for that thread. The app-wide remembered selection stayed stale, so later threads could revert to an earlier reasoning level.

This PR routes explicit Mobile model selections through one draft-state operation. New-task and existing-thread flows now update both the current draft and the remembered selection.

## Why

The documented model-default behavior applies to provider options such as reasoning effort. Mobile already persisted those options when starting a new task, but not when changing them in an existing thread. Sharing the state operation keeps both entry points consistent without adding server state or changing persistence formats.

## UI Changes

None. This changes persistence behavior only.

## Verification

- vp test run apps/mobile/src/state/use-composer-drafts.test.ts
- Targeted lint on the four changed files
- Targeted formatting check on the four changed files
- vp run --filter @t3tools/mobile typecheck

## Checklist

- [x] Screenshots are not applicable because there are no visual UI changes
- [x] Video is not applicable because there are no animation or presentation changes

Built with GPT-5.6 Sol using Codex in T3 Code.

> [!NOTE]
> ### Fix model option changes not persisting in existing threads
> Model selections updated in an existing thread were not saved to the sticky atom. Adds a shared `setComposerDraftModelSelection` helper in [use-composer-drafts.ts](https://github.com/pingdotgg/t3code/pull/9372/files#diff-6acb6f89d59b03b7738cacfb4cec712073b5d8c6b8d2abcd3cde1acee04f818e) that updates both the composer draft and the sticky model-selection atom in one operation. Replaces the separate draft-only updates in `NewTaskFlowProvider` and `useThreadComposerState` with calls to this helper.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e8fe89d.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/features/threads/new-task-flow-provider.tsx
+++ b/apps/mobile/src/features/threads/new-task-flow-provider.tsx
@@ -49,8 +49,8 @@ import {
   removeComposerDraftAttachment,
   replaceComposerDraftAttachments,
   scheduleUnusedComposerAttachmentCleanup,
+  setComposerDraftModelSelection,
   setComposerDraftText,
-  setStickyComposerModelSelection,
   updateComposerDraftSettings,
   useComposerDraft,
   useStickyComposerModelSelection,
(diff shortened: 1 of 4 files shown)


lukasbxy:
For maintainer context, this PR does not introduce a new global-default policy. docs/user/composer.md already states that T3 Code remembers the last provider, model, and model options for new threads. Web/Desktop and the Mobile new-task flow already update that remembered selection.

The missing path was a model or option change in an existing Mobile thread, which updated only the thread draft. This PR routes that path through the same state operation. Project defaults and provider-availability precedence remain unchanged.

PR #9400, opened 2026-09-03
test(server): resolve the temp dir before matching the symlinked entrypoint

### LINKED ISSUE #9389: test(server): entrypoint symlink case asserts unrealpath'd module URL — fails on macOS (/var → /private/var)
`apps/server/src/entrypoint.test.ts` — **"matches through a symlinked entrypoint, as npm and npx install it"** fails deterministically on macOS.

The test does:

```ts
const dir = mkdtempSync(join(tmpdir(), "t3-entrypoint-test-"));
const real = join(dir, "bin.mjs");
const link = join(dir, "t3");
writeFileSync(real, "");
symlinkSync(real, link);
expect(isEntrypoint({ moduleUrl: pathToFileURL(real).href, entryPath: link, runtimeMain: undefined })).toBe(true);
```

The implementation (`entrypoint.ts:34`) compares `moduleUrl === pathToFileURL(realpathSync(entryPath)).href`. On macOS `$TMPDIR` is under `/var/folders/…` and `/var` is a symlink to `/private/var`, so `realpathSync(entryPath)` resolves to `/private/var/folders/…/bin.mjs` while `pathToFileURL(real).href` stays `/var/folders/…/bin.mjs` — the strings can never be equal, so the assertion gets `false` instead of `true`.

The product code is correct: canonicalizing the entry path is exactly what makes npm/npx-style symlinked entrypoints match. The fixture is what needs the realpath — e.g. `pathToFileURL(realpathSync(real)).href`, or realpath the temp dir when creating it — so the test isolates the symlink resolution it intends to exercise from macOS's own `/var` → `/private/var` symlink. Linux CI is unaffected, which is why this only surfaces running the suite locally on macOS.

ylcn91:
## What Changed

`apps/server/src/entrypoint.test.ts`, "matches through a symlinked entrypoint": the expected module URL is now built from `realpathSync(real)` instead of the raw temp path. One line plus a comment.

## Why

Fixes #9389.

`isEntrypoint` compares the module URL against the realpath of the entry, which is what makes npm/npx-style symlinked entrypoints match. The fixture built its expected URL from the un-resolved temp path. On macOS the temp dir lives under `/var`, itself a symlink to `/private/var`, so the two strings never matched and the test failed on every run. Linux CI is unaffected, which is why this only shows up locally on macOS.

The product code is correct. Canonicalizing the fixture path keeps the test focused on the symlink it is meant to exercise.

Before, on macOS:

```
× matches through a symlinked entrypoint, as npm and npx install it
AssertionError: expected false to be true
```

After:

```
Test Files  1 passed (1)
     Tests  5 passed (5)
```

Server typecheck and lint are clean.

## Checklist

- [ ] I included before/after screenshots for any UI changes (no UI changes)

Model and harness: Claude Fable 5.1 in Claude Code.

---

> [!NOTE]
> **Low Risk**
> Test-only change; entrypoint detection logic is untouched.
> 
> **Overview**
> Fixes a **macOS-only failure** in `entrypoint.test.ts` for the symlinked npm/npx entrypoint case (#9389).
> 
> The test now builds the expected `moduleUrl` from `NodeFS.realpathSync(real)` instead of the raw temp path, matching how `isEntrypoint` compares against the realpath of `entryPath`. On macOS, temp dirs under `/var` (symlinked to `/private/var`) made the old expected string never match even though production behavior is correct.
> 
> Adds a short comment explaining Node URL resolution and the `/var` symlink. **No production code changes.**
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f629824f6ff2e61f18d6c3fbb4db9e19a790f34c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Resolve temp dir real path in `isEntrypoint` symlinked-entrypoint test
> Updates the test fixture's module URL in [entrypoint.test.ts](https://github.com/pingdotgg/t3code/pull/9400/files#diff-d5f73be07c19ce8ffb26e958045929c084f6d206e6f52a6b3faf9faafa52ddd2) to use the filesystem-resolved real path of the target file. Node resolves module URLs to real paths, and macOS temp directories can be symlinked, which caused the previous assertion to fail.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f629824.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed symlinked entrypoint testing to correctly resolve module URLs to the underlying real file path, including environments where temporary directories use symlinks.

### DIFF
--- a/apps/server/src/entrypoint.test.ts
+++ b/apps/server/src/entrypoint.test.ts
@@ -54,9 +54,11 @@ describe("isEntrypoint", () => {
       NodeFS.writeFileSync(real, "");
       NodeFS.symlinkSync(real, link);
 
+      // Node resolves the module URL to the real path, and on macOS the temp
+      // directory itself sits behind a symlink (`/var` -> `/private/var`).
       expect(
         isEntrypoint({
-          moduleUrl: NodeURL.pathToFileURL(real).href,
+          moduleUrl: NodeURL.pathToFileURL(NodeFS.realpathSync(real)).href,
           entryPath: link,
           runtimeMain: undefined,
         }),


Mnigos (approved):
Can confirm on macOS: the entrypoint symlink test fails on main exactly as #9389 describes and passes with this change. Same `/var` to `/private/var` class I just ran into on the server side, resolving before asserting is the right call.

shivamhwp:

This branch conflicts with the newer symlink-support guard in main. Please rebase the realpath correction while keeping `it.skipIf(!symlinksSupported)`, so Windows installations without symlink privileges keep skipping the fixture.

ylcn91:
Rebased onto main. The `it.skipIf(!symlinksSupported)` guard stays as is; the only change is still the `realpathSync` on the module URL.

PR #9455, opened 2026-09-03
fix(web): settled banner un-settles threads again

RakshithBhat03:
Clicking `Un-settle` in the banner above a resting composer only expanded the composer. The thread stayed settled and the button remained visible, even though the same action worked from the sidebar and thread menu.

The banner button receives focus during pointer-down. On desktop, the resting composer treated that focus as an intent to start typing, expanded synchronously, and moved the button before the click could finish. The banner already marks its controls as safe to use while the composer is collapsed. This change includes that marker in the shared resting-control event scope, so focus stays on the action and its click dispatches `thread.unsettle`.

This keeps the existing mobile behavior while replacing its one-off check with the shared rule. The same rule covers `Wake now` in the snoozed-thread banner, which had the same focus race. The thread command and settlement state handling are unchanged.

Tests: 6 focused tests, web typecheck, targeted lint, formatting.

## Screenshot

The highlighted `Un-settle` action is the banner control affected by the pointer and focus race.

Un-settle action in the settled-thread banner

---
Built by GPT-5.6 Sol in T3 Code through the Codex harness.

### DIFF
--- a/apps/web/src/components/chat/ChatComposer.tsx
+++ b/apps/web/src/components/chat/ChatComposer.tsx
@@ -4564,13 +4564,6 @@ export const ChatComposer = memo(function ChatComposer(props: ChatComposerProps)
         if (composerControlsInStrip && isInsideRestingComposerControlScope(activeElement)) {
           return;
         }
-        if (
-          isComposerCollapsedMobile &&
-          activeElement instanceof HTMLElement &&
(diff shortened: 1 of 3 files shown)


juliusmarminge:
Superseded by https://github.com/pingdotgg/t3code/pull/9452 (merged) — same fix for banner actions (Un-settle / Wake now) expanding the resting composer instead of running. Closing this duplicate.

PR #9477, opened 2026-09-03
fix(web): invisible toggles and focus rings from imported themes

flamboh:
> [!NOTE]
> 🤖 **GPT-5.6 Sol on behalf of Oliver**

## Problem

As an example, importing Catppuccin Mocha made every toggle in Settings disappear. The theme sets `input.border` to fully transparent, which the importer flattened onto the editor background, so the `input` role became the canvas color. That role is the unchecked switch track and the border of every text input.

An audit of 47 published themes and the VS Code defaults found three more cases of the same shape: a transparent or canvas-colored `focusBorder` yields an invisible focus ring and accent (all five Vitesse themes, One Dark Pro Flat), the placeholder color was validated against the canvas instead of the raised surface inputs render on (Light+), and names that slug to `dark` or `light` (Dark+, Light+) hit a reserved id and fail to import at all.

## Fix

The importer now treats a workbench color as usable only when it does the job the role needs:

- The `input` role tries `input.background` first, then `input.border`, and accepts a candidate only if it stands apart from both the canvas and the checked-switch color. Otherwise it keeps the derived value. `dropdown.border` is no longer a candidate; it turns an edge color into a fill.
- The accent seed skips a `focusBorder` that flattens into the canvas and moves on to the next accent key, falling back to the standard palette accent instead of the canvas.
- Placeholder readability is checked against the resolved raised surface.
- A generated id that collides with a reserved id gets a `-vscode` suffix. The label is unchanged.

Built-in themes and the CSS token mapping are untouched. Already imported themes keep their stored colors, so a re-import is needed to pick up the change.

### Before

SCR-20260904-bgcq

### After

SCR-20260904-birb

### Focus ring before
image

### Focus ring after
image

## Verification

- Added 6 focused cases for input candidate selection, checked-state separation, transparent focus borders, placeholder readability on the raised surface, and reserved ids
- `vp test run apps/web/src/vscodeThemeImport.test.ts`, 22 tests passed
- `vp lint <changed files> --report-unused-disable-directives`
- `vp fmt --check <changed files>`

Changes prepared by Claude Fable 5.1 through Claude Code in T3 Code, with a GPT-5.6 Sol subagent auditing the theme corpus and implementing the importer changes to spec.

> [!NOTE]
> ### Fix invisible toggles and focus rings in imported VS Code themes
> Hardens the VS Code theme import logic so imported themes no longer produce invisible or low-contrast UI elements.
>
> - Skips transparent input borders and selects a visible input background instead
> - Ensures derived input colors stay contrast-distinct from message action and button colors, even when source inputs are translucent
> - Replaces focus borders that are transparent or too close to a raised surface with a button-derived focus color; falls back to a canvas-distinct color when no usable accent exists
> - Ignores transparent button backgrounds for the message action role, using the visible focus border instead
> - Preserves labels for reserved theme names like `Dark+` while assigning non-reserved identifiers
> - Adds regression tests in [vscodeThemeImport.test.ts](https://github.com/pingdotgg/t3code/pull/9477/files#diff-d690b716e49037069fdbd1c1dbd75556f04ec3c7ed6b8e46ab0f9a7e9cda82ad) covering all above cases
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 016f831.</sup>
> 
>

## Summary by CodeRabbit

- **Improvements**
  - Improved VS Code theme imports with more readable accent, input, button, and focus colors.
  - Transparent or low-contrast colors are skipped in favor of visible alternatives.
  - Focus colors remain distinguishable against raised surfaces.
  - Placeholder text readability is improved against raised surfaces.
  - Imported themes use a standard accent when no suitable accent is available.
  - Reserved theme names are automatically adjusted to prevent naming conflicts.
  - Unchecked inputs remain visually distinct from checked action colors.
  - Input colors remain distinct from canvas and action colors.

### DIFF
--- a/apps/web/src/vscodeThemeImport.test.ts
+++ b/apps/web/src/vscodeThemeImport.test.ts
@@ -156,6 +156,19 @@ describe("VS Code theme import", () => {
     expect(contrastRatio(vitesse.colors.focus, vitesse.colors.canvas)).toBeGreaterThanOrEqual(1.1);
   });
 
+  it("skips a transparent button background for the action color", () => {
+    const theme = parseVsCodeThemeFile({
+      name: "Transparent button",
+      type: "dark",
+      colors: {
+        "editor.background": "#121212",
+        focusBorder: "#4d9375",
+        "button.background": "#00000000",
+      },
+    });
+    expect(asHex(theme.colors.messageAction)).toBe("#4d9375");
+  });
+
   it("uses a visible default accent when the file has no usable accent key", () => {
     const theme = parseVsCodeThemeFile({
       name: "No accent",
--- a/apps/web/src/vscodeThemeImport.test.ts
+++ b/apps/web/src/vscodeThemeImport.test.ts
@@ -112,6 +112,89 @@ describe("VS Code theme import", () => {
     );
   });
 
+  it("prefers a visible input background over a transparent border", () => {
+    const catppuccin = parseVsCodeThemeFile({
+      name: "Catppuccin Mocha",
+      type: "dark",
+      colors: {
+        "editor.background": "#1e1e2e",
+        "input.border": "#00000000",
+        "input.background": "#313244",
+      },
+    });
+    expect(asHex(catppuccin.colors.input)).toBe("#313244");
+  });
+
+  it("keeps unchecked inputs distinct from the checked action color", () => {
+    const gruvbox = parseVsCodeThemeFile({
+      name: "Gruvbox Light Soft",
+      type: "light",
+      colors: {
+        "editor.background": "#f2e5bc",
+        "button.background": "#45858880",
+        "input.background": "#45858880",
+        "input.border": "#928374",
+      },
+    });
+    expect(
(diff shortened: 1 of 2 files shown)


flamboh:
> [!NOTE]
> 🤖 **GPT-5.6 Luna on behalf of Oliver**

@coderabbitai review

The current head includes the raised-control and fallback-focus contrast fixes. Please review `e72e8c689`.

flamboh:
> [!NOTE]
> 🤖 **GPT-5.6 Luna on behalf of Oliver**

The transparent raised-surface report was checked against the actual consumers and remains out of scope for this PR. `surfaceRaised` is used for the composer backdrop, while dropdown and popover content uses `surfaceOverlay`. Treating a transparent `editorWidget.background` as absent here would select `dropdown.background` for the composer, changing the imported theme's editor surface without fixing the reported dropdown placeholder relationship. The importer already validates the placeholder against the resolved `surfaceRaised`, and the concrete input/action fallback collision has been fixed in `016f831`.

Issue #9405, opened 2026-09-03
[Bug]: Antigravity Provider Google Sign-in not working


7A7z:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

After installing antigravity on the provider menu , Click Google Sign-in

<img width="868" height="663" alt="Image" src="https://github.com/user-attachments/assets/caa76ce7-0051-4dc0-823c-24f9c57ffed9" />

### Expected behavior

Google sign-in to succeeds and the agy provider starts working 

### Actual behavior

Clicking on google sign-in hang like this for about 10 - 15 minutes before expiring

<img width="840" height="640" alt="Image" src="https://github.com/user-attachments/assets/d0262024-9655-47f9-833d-aa5b68cc2c96" />

### Impact

Blocks work completely

### Version or commit

0.0.39-nightly.20260903.1268

### Environment

Windows 11 25H2

### Logs or stack traces

```shell

```

OpticMystic:
I  too have this issue

orbitnull:
same issue

JyotirmoyDas05:
Same Problem

orbitnull:
Anyone found a solution?

y-ziane:
+1 same issue

WellyngtonF:
My PR fix this, just wait the team wake up to check

orbitnull:
someone should wake them up

PabloValencia14:
Additional reproduction and diagnostics from another Windows 11 user:

## Environment

- OS: Windows 11 25H2
- T3 Code: 0.0.39-nightly.20260903.1272 (d5825e1d2fb1)
- Antigravity provider: managed runtime installed by T3 Code
- Managed ACP runtime: agy_acp_server_20260818_01_RC01
- Browser: Vivaldi, registered as the default HTTP/HTTPS browser

## Reproduction

1. Enable Antigravity in Settings > Providers.
2. Install the managed Antigravity runtime.
3. Click Sign in with Google / Retry Google sign-in.
4. No browser window opens.
5. The UI does not show Open sign-in page or Copy sign-in link.
6. After waiting, authentication fails.

## Observed diagnostics

The managed runtime starts and ACP initialization succeeds. During the sign-in attempt, T3 records successful:

- ProviderAuthService.start
- ws.rpc.provider.auth.start
[...]

PR #9550, opened 2026-09-04
fix(antigravity): hand the Google sign-in URL to T3 through a browser preload

shootingallday:
On Windows, Antigravity Google sign-in stays on its starting state and never shows the link. #9425 started reading the browser-helper marker from stderr, but the helper never runs: the agent's Python 3.10 `webbrowser` runs the whole `BROWSER` value as one executable with the URL as its only argument, so the quoted `node -e ... -- %s` command cannot start. Python then logs "Could not open a browser automatically", and the agent's own stdout copy of the URL sits in a block-buffered `print()` that only flushes at exit. T3 has no path to the URL, and the flow expires.

Reproduced against the packaged agent (`agy_acp_server.exe`, PyInstaller Python 3.10.4) on Windows 11 with the installed nightly `0.0.39-nightly.20260903`. Reports in #9405 (closed by #9425) say sign-in still fails on the current nightly, which matches.

## Fix

- Sign-in launches set `BROWSER` to the bare T3 runtime executable, which Python can run, and add `NODE_OPTIONS=--require <profile>/antigravity-acp/t3-browser-preload.cjs`. Node resolves the URL argument as a path before preloads run, so the preload rebuilds the URL from `<cwd>/https:/host/path`, posts it to a per-sign-in loopback sink (`T3_ANTIGRAVITY_AUTH_SINK`, random token, 127.0.0.1, 204 only for `POST /<token>`), and exits 0. With no sink it writes the existing stderr marker, which the preflight check uses.
- Only the sign-in launch carries `NODE_OPTIONS`, the runtime `BROWSER`, and the sink, so turn subprocesses never inherit them. Other launches keep the unrunnable `BROWSER` value so no browser opens.
- The preload is written before the preflight helper runs, and the preflight now spawns the helper the way Python does: one executable, one URL argument.
- Existing stdout and stderr URL paths stay as fallbacks; #9425's same-URL dedup handles duplicate delivery.

Verified end to end on Windows with the real packaged agent: the sink receives the URL about 100 ms after `authenticate`, no browser window opens, and a rebuilt server bundle in the installed nightly completes sign-in.

## Tests

- New: only the sign-in launch is routed through the preload and sink; the sign-in URL reaches the sink from a real spawn of the preload; the profile is not created when the helper cannot start.
- Focused server suites (`antigravityAuthSupport`, `AntigravityAuth`, `AntigravityAcpSupport`, `AntigravityProvider`): 97 passed.
- Server typecheck, targeted lint, and format check pass.

Built with Claude Fable 5.1 in Claude Code.

---

> [!NOTE]
> **Medium Risk**
> Touches OAuth sign-in, loopback HTTP, and subprocess environment wiring; scope is limited to Antigravity personal/business browser auth with existing URL validation and sign-in-only env isolation.
> 
> **Overview**
> Fixes Antigravity OAuth on Windows (and similar hosts) where Python 3.10’s `webbrowser` runs `BROWSER` as a **single executable** with the URL as its only argument, so the old quoted `node -e` helper never runs and the auth link stays stuck in block-buffered agent stdout.
> 
> **Sign-in-only path:** When `onAuthorizationUrl` is set, the driver starts `serveAntigravityAuthorizationUrlSink` and passes its URL into `buildAntigravityAcpSpawnInput`. That launch sets `BROWSER` to the bare T3 runtime, adds `NODE_OPTIONS=--require <profile>/t3-browser-preload.cjs`, and `T3_ANTIGRAVITY_AUTH_SINK`. The preload rebuilds the URL from Node’s path-mangled argv, POSTs it to the one-shot `127.0.0.1` listener (random token, size cap), then exits 0. Normal ACP sessions keep the non-runnable `BROWSER` string and **no** preload/sink so subprocesses do not inherit sign-in env.
> 
> **Profile prep:** Writes `t3-browser-preload.cjs` before verification; browser preflight now spawns the runtime with a URL argument (matching Python), not `-e` inline script. `AntigravityProfile` gains `browserHelperPath` and `browserPreloadPath`. Stdout/stderr URL parsing remains as fallbacks.
> 
> Tests cover env split (sign-in vs session), end-to-end preload→sink delivery, and updated installation validation mocks for URL-as-first-arg.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35d2ecc886ca0218a747a1af540aeac96f707bf7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Route Google sign-in URL to T3 via browser preload and loopback sink in Antigravity
> - Sign-in ACP launches now set `BROWSER` to the profile's runtime executable and append a generated CommonJS preload to `NODE_OPTIONS`; ordinary ACP launches keep the configured browser command
> - The preload reconstructs an HTTPS URL from the Node entry argument and POSTs it to a loopback sink (or falls back to the existing stdout marker when no sink is set), then exits
> - `serveAntigravityAuthorizationUrlSink` starts a scoped loopback HTTP server on an ephemeral port with a tokenized POST path; invalid routes return 404 and oversized bodies return 413
> - `prepareAntigravityProfile` writes the preload file into the ACP profile and rejects profile paths containing a double quote to avoid preload-path injection
> - `AntigravityDriver.create` starts a loopback sink when the runtime input supplies an auth-URL callback and passes the sink endpoint into the ACP spawn config
> - Behavioral Change: `antigravityEnvironment` now strips the auth-sink env var from inherited environment keys; only explicitly supplied sign-in launches receive it. Profiles with quoted preload paths now fail preparation with an auth-support error
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 35d2ecc.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/AntigravityInstallation.test.ts
+++ b/apps/server/src/provider/AntigravityInstallation.test.ts
@@ -312,7 +312,7 @@ it.layer(NodeServices.layer)("Antigravity installation", (it) => {
           const profile = command.options.env?.GEMINI_HOME;
           if (!profile) return yield* Effect.die("Expected a disposable validation profile.");
           profiles.add(profile);
-          const helper = command.args[0] === "-e";
+          const helper = command.args[0]?.startsWith("https://") === true;
           const output = yield* Queue.unbounded<Uint8Array>();
           const exited = yield* Deferred.make<ChildProcessSpawner.ExitCode>();
           const terminate = Deferred.succeed(exited, ChildProcessSpawner.ExitCode(0)).pipe(
--- a/apps/server/src/provider/Drivers/AntigravityDriver.ts
+++ b/apps/server/src/provider/Drivers/AntigravityDriver.ts
@@ -28,6 +28,7 @@ import {
   isAntigravitySignInRequiredError,
   prepareAntigravityProfile,
   resolveAntigravityProfileDirectory,
+  serveAntigravityAuthorizationUrlSink,
   type AntigravityAuthConfig,
 } from "../antigravityAuthSupport.ts";
 import {
@@ -151,6 +152,9 @@ export const AntigravityDriver: ProviderDriver<AntigravitySettings, AntigravityD
           Effect.provideService(Path.Path, path),
           Effect.provideService(ChildProcessSpawner.ChildProcessSpawner, spawner),
         );
+        const authorizationUrlSink = input.onAuthorizationUrl
+          ? yield* serveAntigravityAuthorizationUrlSink(input.onAuthorizationUrl)
+          : undefined;
         const runtime = yield* makeAntigravityAcpRuntime({
           ...input,
           authMethod: auth.authMethod,
@@ -161,6 +165,7 @@ export const AntigravityDriver: ProviderDriver<AntigravitySettings, AntigravityD
             cwd: input.cwd,
             baseEnv: processEnvironment,
             auth,
+            ...(authorizationUrlSink ? { authorizationUrlSink } : {}),
           }),
         }).pipe(Effect.provideService(Crypto.Crypto, crypto));
         return {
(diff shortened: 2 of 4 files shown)


shootingallday:
Closing; this did not follow the contribution process in CONTRIBUTING.md.

PR #9612, opened 2026-09-04
fix(server): keep Claude messaging sessions alive

### LINKED ISSUE #9586: [Bug]: Thread shows as live after the idle reaper has stopped its Claude Code process
A thread idle 30 minutes still shows as live in T3 after the backend has stopped its Claude Code process. Peers lose it: `ListAgents` drops the thread, and `SendMessage` to its last address fails.

### Before submitting

### Area

apps/server

### Steps to reproduce

1. Open two Claude Code threads in the same working directory; send a turn in each.
2. From thread A, call `ListAgents`; thread B is listed with a socket path.
3. Leave thread B idle 30+ minutes, then `SendMessage` to that address from thread A.

### Expected behavior

A thread shown as live stays addressable from its peers, or stops being shown as live. The ask: exempt sessions that are live cross-session messaging endpoints, in the same shape as the `backgroundLiveness` guard from #4198/#5677.

### Actual behavior

Thread B is gone from the roster and the send fails; T3 shows no change.

It's the inactivity reaper — verified in `ProviderSessionReaper.ts`: `DEFAULT_INACTIVITY_THRESHOLD_MS` is 30 minutes, swept every 5, with exactly two exemptions, an active turn and `backgroundLiveness` (added by #5677 for #4198). Everything else past the threshold goes to `providerService.stopSession`. A thread idle because it's waiting to be messaged has neither exemption.

Resuming doesn't heal it: the roster is process-scoped (`~/.claude/sessions/<pid>.json`, socket `/tmp/cc-socks/<pid>.sock`), so the endpoint dies with the process. The conversation returns under a new pid, socket, and derived name (observed: `"nameSource":"derived"`), so addressing by roster name doesn't survive either.

Some of these peers are terminal sessions T3 never started, so the T3 process must stay on Claude Code's roster.

### Impact

### Version or commit

0.0.38 desktop (Alpha); citations against `main` @ `70cd258d`

### Environment

macOS 26.6.2 (Darwin 25.6.0), Node v26.3.1, Claude Code 2.1.259

### Logs or stack traces

```shell
none — server.trace.ndjson rotates faster than the 30-min threshold
```

### Workaround

Re-run `ListAgents` before every send and wake the target with a turn — which takes a human in that thread.

Related: #5523 asked to make the threshold configurable; this asks for an exemption, not a setting.

Drafted with Claude Code (Claude Opus 5) from source and transcripts on my machine; I reviewed and posted it.

Gigioxx:
## What Changed

Keep live Claude sessions out of the inactivity reaper so their cross-session messaging endpoint remains addressable. Other providers and Claude bindings without a live adapter session keep the existing cleanup behavior.

## Why

Claude's process owns its roster entry and messaging socket. Reaping an idle thread removes it from `ListAgents`, makes `SendMessage` fail, and resuming cannot restore the same endpoint identity.

Fixes #9586.

## Verification

- Reproduced the failure with a focused reaper test before the fix
- `vp test run apps/server/src/provider/Layers/ProviderSessionReaper.test.ts` (9 passed)
- Targeted format and lint passed
- `vp run --filter t3 typecheck` passed
- `git diff --check` passed

## Checklist

- [x] No UI changes

Implemented with `gpt-5.6-sol` in T3 Code through the Codex harness.

---

> [!NOTE]
> **Medium Risk**
> Changes session lifecycle for Claude only and adds a listSessions dependency each sweep; wrong skip logic could leave zombie sessions or fail to reap when messaging is gone.
> 
> **Overview**
> The inactivity **ProviderSessionReaper** no longer stops idle **Claude** bindings when `listSessions()` still reports a live `claudeAgent` session for that thread, so cross-session messaging can keep using the same process endpoint. Other providers and Claude bindings without a live adapter session still follow the existing stale-session cleanup.
> 
> Tests mock `activeSessions` on the harness, assert stale **codex** sessions are still reaped, and add cases for idle **ready** / **error** Claude threads that must not call `stopSession`. Multi-session failure/defect scenarios use **error** thread status on the Claude side so those bindings are not exempted by the new skip rule.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3ab694fbfd451087d1eee8a1dab1492f03057ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `ProviderSessionReaper` to keep live Claude messaging sessions alive
> The reaper previously stopped stale persisted Claude bindings even when a live session was still active. The sweep in `makeProviderSessionReaper` now calls `providerService.listSessions`, builds a set of active Claude thread IDs, and skips stopping any Claude binding whose thread ID is in that set.
>
> Tests in [ProviderSessionReaper.test.ts](https://github.com/pingdotgg/t3code/pull/9612/files#diff-9c064659a893430d8c3796d891064feb010d88e0f11222f98b078f4f71c8d6e1) are updated: the generic stale-session test switched to the Codex provider, and new parameterized tests verify that live idle Claude sessions in `ready` and `error` states are retained.
>
> Risk: non-Claude providers are unaffected; only Claude bindings with a matching live thread ID are exempted from reaping.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c3ab694.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/ProviderSessionReaper.test.ts
+++ b/apps/server/src/provider/Layers/ProviderSessionReaper.test.ts
@@ -5,6 +5,7 @@ import {
   TurnId,
   ProviderDriverKind,
   ProviderInstanceId,
+  type ProviderSession,
 } from "@t3tools/contracts";
 import * as Clock from "effect/Clock";
 import * as DateTime from "effect/DateTime";
@@ -147,6 +148,7 @@ describe("ProviderSessionReaper", () => {
 
   async function createHarness(input: {
     readonly readModel: ReturnType<typeof makeReadModel>;
+    readonly activeSessions?: ReadonlyArray<ProviderSession>;
     readonly stopSessionImplementation?: (input: {
       readonly threadId: ThreadId;
     }) => ReturnType<ProviderServiceShape["stopSession"]>;
@@ -169,7 +171,7 @@ describe("ProviderSessionReaper", () => {
       respondToRequest: () => unsupported(),
       respondToUserInput: () => unsupported(),
       stopSession,
-      listSessions: () => Effect.succeed([]),
+      listSessions: () => Effect.succeed(input.activeSessions ?? []),
       getCapabilities: () => Effect.succeed({ sessionModelSwitch: "in-session" }),
       assertConversationRollbackSupported: () => unsupported(),
       getInstanceInfo: (instanceId) => {
@@ -247,7 +249,7 @@ describe("ProviderSessionReaper", () => {
           session: {
             threadId,
             status: "ready",
-            providerName: "claudeAgent",
+            providerName: "codex",
             runtimeMode: "full-access",
             activeTurnId: null,
             lastError: null,
@@ -263,9 +265,9 @@ describe("ProviderSessionReaper", () => {
     await runtime!.runPromise(
       repository.upsert({
         threadId,
-        providerName: "claudeAgent",
+        providerName: "codex",
         providerInstanceId: null,
-        adapterKey: "claudeAgent",
+        adapterKey: "codex",
         runtimeMode: "full-access",
         status: "running",
         lastSeenAt: "2026-04-14T00:00:00.000Z",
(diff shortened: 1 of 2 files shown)


shivamhwp:

Please rebase the new reaper cases onto the current `sweepAt` harness and await its completed sweep. Ten `Effect.yieldNow` calls do not prove the reaper finished, so the negative stop assertion can pass before the relevant work runs. Preserve the direct thread-shell query and the full idle window after turn settlement when resolving the rebase.

KonstantinKo:
Just to note, I too am eagerly awaiting this PR getting merged. It also covers Claude sessions that are idle but armed to wake themselves, which the reaper can't see cause neither shows up as a background task. In my case, a scheduled wakeup (ScheduleWakeup, `/loop`, CronCreate; delays up to 60 min) plus the case of an MCP server channel push (`notifications/claude/channel`, enabled via `--channels` in Launch arguments). Both get stopped after 30 idle min...

Repro: T3 Code desktop, Claude provider with `--channels` in Launch arguments, and an MCP channel server that pushes an event ~40 min after the turn ends. Turn completed at 06:46:02 UTC; the session was stopped at 07:16:34 UTC (30.5 min, plain `stopSession`, no settle), so the push never arrived and the channel server's in-flight work was torn down with the session.

BTW I'd be happy to help with the test rebase if @Gigioxx has abandoned this.

KonstantinKo:
@Gigioxx I rebased this onto current main and moved the new cases onto the `sweepAt` harness: they now await the completed sweep instead of the `yieldNow` loop, and they fail with the exemption disabled. Your commit is unchanged apart from the rebase (no conflicts; the direct thread-shell query and the full idle window after settlement are intact).

On top of that, two separate commits you can take or drop:
- the 3-line test change;
- a fix for a regression the new lookup introduces: `listSessions()` dies on a binding mismatch, which aborts every sweep and stops reaping for all providers. An unknown live set now keeps Claude sessions and lets the sweep continue, with a regression test.

Compare against main: https://github.com/pingdotgg/t3code/compare/main...KonstantinKo:t3code:reaper-rebase-9612

Taking it is one command from your checkout:

```sh
git fetch https://github.com/KonstantinKo/t3code reaper-rebase-9612 && git push -f origin FETCH_HEAD:t3code/keep-messaging-sessions-alive
```

`vp test run apps/server/src/provider/Layers/ProviderSessionReaper.test.ts` passes 14/14; lint, fmt and the server typecheck are clean.

(Written by Claude Opus 5 for Konstantin)

PR #9416, opened 2026-09-03
fix(web): keep the sidebar project filter across navigation

Mnigos:
The sidebar's project filter lived in component state, so any route that unmounts the sidebar (Settings, most visibly) reset it to all projects on the way back.

The selection now lives in the persisted UI state store, next to the other sidebar project preferences (expansion, order), so it survives the Settings round-trip and app restarts. A stored key whose project no longer exists falls back to all projects, and only once the project list has actually loaded, since projects stream in after mount and clearing earlier would wipe the restored scope on every launch.

To verify: filter the sidebar to one project, open Settings, come back. The filter used to snap back to all projects; now it stays.

Fixes [#9379](https://github.com/pingdotgg/t3code/issues/9379).

Implemented with Claude Code (Claude Fable 5).

---

> [!NOTE]
> **Low Risk**
> Client-only UI preference persistence with sanitization and guarded fallback; no auth or server behavior changes.
> 
> **Overview**
> **Persists the sidebar “filter by project” choice** so it no longer resets when the sidebar unmounts (e.g. visiting Settings) or after an app restart.
> 
> The combobox selection is read and updated via `useUiStateStore` (`sidebarProjectScopeKey` / `setSidebarProjectScopeKey`) instead of local `useState`. That value is included in the existing debounced `localStorage` UI state alongside project order and expansion.
> 
> **Invalid scope handling** is tightened: if a saved project key no longer exists, the filter still resets to “All projects”, but only after `projectGroups` has loaded—avoiding a race where an empty project list on startup cleared a valid restored scope every launch.
> 
> Tests cover the pure setter, parse/persist round-trip, and reload hydration for `sidebarProjectScopeKey`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3db4f0a413850ee7582164ea6e89528d158922cd. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Persist sidebar project filter in `uiStateStore` across navigation and restarts
> - Replaces local component state in the `Sidebar` component with the persisted UI store's sidebar scope field and setter, so the selected project survives unmounts and app reloads.
[...]

PR #9614, opened 2026-09-04
feat(web): multi-select projects in the sidebar scope picker

Andreas-Froyland:
## What Changed

The project scope picker in the thread sidebar now supports selecting multiple projects at once:

- CTRL/CMD-click (or CTRL/CMD+Enter) toggles a project into a multi-scope while the popup stays open, and the search query survives toggling.
- Plain click keeps the existing single-project behavior, and "All projects" resets the scope. Marked projects highlight inside the popup.
- The trigger shows the selected project names joined with ", ", and once several projects make that label truncate, hovering the picker shows a tooltip with the full list.
- The whole scope set persists, so navigating to Settings or restarting the app restores every selected project rather than just one.

## Why

I am often working on multiple repos that are connected, as well as using T3 Code for different purposes on different computers, while still sharing one account. The single-project scope hid work that was spread across related projects, and this keeps the single-project workflow intact while letting a multi-scope show everything at once.

## Merging with main

Two changes landed on `main` after this branch opened and both touched the same code, so this now builds on top of them rather than reverting them:

- #9416 moved the scope into the persisted UI store as a single `sidebarProjectScopeKey`. That field is now `sidebarProjectScopeKeys: string[]`; `parsePersistedState` still reads the old singular key once, so installs that saved a filter before this PR keep it. The "scoped project vanished" reset keeps its `useAllEnvironmentProjectSnapshotsReady()` gate, which matters more now — without it a persisted scope would be pruned on every cold start while environments are still connecting.
- #5521 added keyboard and screen-reader access to project settings from the picker. Its `onItemHighlighted` and `suppressNextScopeChangeRef` guards are preserved inside the new `onValueChange`. The guard is load-bearing here: on macOS a CTRL+click fires `contextmenu` and opens project settings, and without it that same press would also toggle the project into the scope.

## UI Changes

Before — one project at a time:

before

After — CTRL+click marks additional projects while the popup stays open:

popup

After — the closed picker joins the selected names (truncated when long):

truncated

Hovering the picker reveals every selected project:

tooltip

Full interaction (video): [project-multi-select.webm](https://github.com/Andreas-Froyland/t3code/releases/download/pr-evidence/project-multi-select/project-multi-select.webm)

## Checklist

Verification:

- `vp test run apps/web/src/components/Sidebar.logic.test.ts apps/web/src/uiStateStore.test.ts` — 147 passed, including tests for the press resolver (`resolveSidebarProjectScopePress`) and for the persisted scope array, its dedup, and the migration from the legacy single key
- `vp run --filter @t3tools/web typecheck` — clean
- `vp lint apps/web/src/components/Sidebar.tsx` — 24 warnings, identical to the `main` baseline measured on the same file

Worked on by GLM-5.3-flash via T3 Code (opencode). Merged with `main` and conflicts resolved by Claude Opus 5 via T3 Code.

## Closes discussions

- https://github.com/pingdotgg/t3code/discussions/10500

> [!NOTE]
> ### Add multi-select project scope to `Sidebar` picker
> - Changes sidebar project scope from a single nullable key to a persisted array of keys in [uiStateStore.ts](https://github.com/pingdotgg/t3code/pull/9614/files#diff-605fd32a90753e51ac329e924c51be2b2ce46a7ce8b6d02117aee02489c8acd8); plain clicks select one project and close the menu, CTRL/CMD clicks toggle projects without closing, and All projects clears the scope.
> - Thread filtering now builds a union of member project references across all selected groups, and unavailable persisted projects are pruned once project snapshots are ready.
> - Adds `sanitizeSidebarProjectScopeKeys` migration so legacy single-key state hydrates into a one-element array, while an explicitly stored new array takes precedence.
> - Risk: `PersistedUiState` keeps the old nullable scope field read-only for migration; consumers still reading the old `setSidebarProjectScopeKey` action or nullable scope property must move to `setSidebarProjectScopeKeys` and the array field.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2589fd4.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Filter threads by one or multiple projects using the sidebar project picker.
  * Use Cmd/Ctrl-click or Enter to select multiple projects while keeping the picker open.
  * Project filters persist across navigation and app restarts, with unavailable projects removed automatically.
  * Select **All projects** to clear the filter.
  * Empty results now clarify whether no threads match one project or multiple projects.

* **Documentation**
  * Added guidance for searching, selecting, and clearing project filters.


IARI:
This solves my exact use case: I regularly have paused or unsettled threads across many projects, but actively work in only two or three at a time. “All projects” creates too much noise, while a single-project scope makes me repeatedly switch filters.

I see the branch now needs a rebase against main. Once it is updated, I can manually validate the Windows desktop interaction and persistence behavior locally.

PR #9620, opened 2026-09-04
fix(server): queue messages sent during context compaction

flamboh:
> [!NOTE]
> 🤖 **Fable 5.1 on behalf of Oliver**

## Problem

Sending a message while a thread is compacting its context (after `/compact`) fails with "Wait for context compaction to finish before sending another message." The web client removes the optimistic message and restores the draft, so the message is never delivered. This affects every provider because the guard lives in the orchestration reactor, not in an adapter.

## Fix

The reactor now holds messages that arrive during compaction in a per-thread FIFO and sends them once the compaction fiber settles, using the same send path a fresh message takes.

- Compaction success or a failed compaction that still restores the session to `ready` both drain the queue in order.
- If the thread stops or the session ends in `error`, each queued message gets its own `provider.turn.start.failed` activity so clients restore the draft instead of silently losing it.
- A second `/compact` during compaction keeps its existing failure.
- The projection now lets a queued message replace the pending `/compact` turn-start row and keeps that row when the compaction restores the session, so the replayed turn still projects `starting` and keeps its proposed-plan reference.

No contract, decider, web, or mobile changes. Messages were already persisted immediately, so clients see them optimistically and get acknowledged when the turn starts.

## Tests

- `ProviderCommandReactor.test.ts`: queued message is sent after a failed compaction restores the session, two messages drain in FIFO order, queued message fails when the thread stops mid-compaction.
- `ProjectionPipeline.test.ts`: a turn start deferred behind compaction becomes the pending row after the session restores.

---

Implemented by GPT-5.6 Sol via `codex exec`, orchestrated and reviewed by Claude Fable 5.1 in Claude Code.

> [!NOTE]
> ### Queue messages sent during context compaction
> Ensures messages sent while context compaction is in progress are queued and retained after the session is restored. Adds an integration test in [ProjectionPipeline.test.ts](https://github.com/pingdotgg/t3code/pull/9620/files#diff-929eb6fd844fad565d5af8faa7348834ab85ad5457f22d71861e6ee886a18fda) verifying that a deferred non-compaction turn-start request survives compaction completion and preserves its message, proposed-plan metadata, and original request timestamp.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 6014e20. 3 files reviewed, 4 issues evaluated, 0 issues filtered, 4 comments posted</summary>
>
> ### 🗂️ Filtered Issues
>
> </details>
>

### DIFF
--- a/apps/server/src/orchestration/Layers/ProjectionPipeline.test.ts
+++ b/apps/server/src/orchestration/Layers/ProjectionPipeline.test.ts
@@ -3259,6 +3259,145 @@ it.layer(makeProjectionPipelinePrefixedTestLayer("t3-pending-turn-terminal-test-
         assert.deepEqual(pendingRows, [{ messageId: "new-message" }]);
       }),
     );
+
+    it.effect("promotes a turn start deferred behind compaction when the session restores", () =>
+      Effect.gen(function* () {
+        const projectionPipeline = yield* OrchestrationProjectionPipeline;
+        const eventStore = yield* OrchestrationEventStore;
+        const sql = yield* SqlClient.SqlClient;
+        const threadId = ThreadId.make("thread-deferred-compaction-turn");
+        const compactMessageId = MessageId.make("message-deferred-compact");
+        const queuedMessageId = MessageId.make("message-deferred-queued");
+        const compactAt = "2026-02-26T16:00:00.000Z";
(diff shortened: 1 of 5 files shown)


flamboh:
Probably worth doing with orchestrator v2, rather than now

PR #4989, opened 2026-07-30
feat(web): render Mermaid diagrams in markdown

jdalmeida:
## What Changed

T3 Code now renders fenced Mermaid diagrams in chat messages and rendered Markdown file previews.

The diagram viewport supports pointer controls, keyboard controls, bounded pan, 50–300 percent zoom, and a reset action.

## Why

Markdown diagrams appeared as source code. This made agent plans and architecture notes harder to read.

A stable Markdown component prevents diagram flicker when thread data changes. The viewport limits zoom and pan positions.

Fixes #4571

### Before

Mermaid source rendered as code

### After

Mermaid diagram rendered in a clean Codex thread

### Interaction

[Watch the zoom, pan, and reset controls](https://raw.githubusercontent.com/jdalmeida/t3code/pr-4571-evidence/pr-evidence/4571/zoom-pan.mp4)

## Checklist

- [x] This PR is focused on Mermaid rendering.
- [x] I explained what changed and why.
- [x] I included before and after images.
- [x] I included a video for the zoom and pan controls.

About 790 changed lines come from the generated lockfile update for Mermaid.

## Validation

- `pnpm exec vp test run apps/web/src/components/MermaidDiagram.test.ts`
- `pnpm exec vp lint apps/web/src/components/MermaidDiagram.tsx apps/web/src/components/MermaidDiagram.test.ts apps/web/src/components/ChatMarkdown.tsx apps/web/src/components/FileMarkdownViewer.tsx`
- `pnpm --filter @t3tools/web typecheck`
- `pnpm --filter @t3tools/web build`
- Local web test with an isolated T3 home and one new GPT-5.6-Sol thread.

Model: GPT-5.6 Codex

Harness: T3 Code Codex harness

> [!NOTE]
> ### Render Mermaid diagrams in chat markdown with pan/zoom support
> - Adds a `MermaidDiagram` component ([MermaidDiagram.tsx](https://github.com/pingdotgg/t3code/pull/4989/files#diff-118c8c29b6b23ac9054861f3054d8d831874749fc924555c1e323edfdae1cdf3)) that renders mermaid code fences as interactive diagrams with toolbar controls for zoom in/out and reset.
> - Diagrams render in strict security mode, respect the current light/dark theme, and serialize renders via a global queue to avoid concurrent conflicts.
[...]

Issue #4571, opened 2026-07-26
[Feature]: mermaid render support within the markdown render view


tnh:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I am describing a concrete problem or use case, not just a vague idea.

### Area

apps/desktop

### Problem or use case

Like other editors, I would like t3code to render mermaid diagrams encoded in markdown.

<img width="657" height="848" alt="Image" src="https://github.com/user-attachments/assets/5ac30a52-c3c9-4e55-a08a-4a27046c4323" />

### Proposed solution

<img width="832" height="357" alt="Image" src="https://github.com/user-attachments/assets/deb1f0ff-495c-4e24-8396-ee20c6dde4cb" />

Example mermiad diagram rendered within Cursor 

### Why this matters

GitHub supports mermiad https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams 

Most other providers support mermiad diagramts. 

### Smallest useful scope

Have t3code use https://github.com/mermaid-js/mermaid 

### Contribution

- [ ] I would be open to helping implement this.

PR #9621, opened 2026-09-04
fix(web): render Mermaid diagrams in Markdown

Aniketvish0:
Mermaid fences currently render as ordinary code blocks, so diagrams in chat, plans, file previews, and PR bodies lose the visual structure available in other renderers.

This change lazy-loads Mermaid only when a completed mermaid fence is present, serializes rendering because Mermaid config is global, and falls back to highlighted source on error or while streaming. Diagrams use strict security mode with a directive blocklist, follow the active theme, and preserve the original fence source for copy. Desktop inherits the web behavior; mobile is unchanged (separate native renderer).

Sizing: Mermaid emits `width="100%"` plus an inline `max-width` equal to the diagram's natural width, so the SVG shrinks to fit the chat column and never grows past its native size. Only `height: auto` is added so the height follows the viewBox. Wide diagrams no longer overflow into a horizontal scroll.

Caching: rendered SVG is cached by theme and source (50 entries). The chat list unmounts off-screen rows, so without the cache every scroll back past a diagram re-ran Mermaid, flashed the code fallback, and jumped the row height.

Validation:
- `vp test run src/components/MermaidDiagram.test.ts src/markdown-clipboard.test.ts src/components/ChatMarkdown.test.tsx` (62 passed)
- `vp lint` on changed files (only the three pre-existing ChatMarkdown warnings, identical on main)
- `tsgo --noEmit` clean
- `vp build` succeeds; Mermaid stays in lazy chunks (mermaid.core 94 kB, katex 259 kB), nothing added to the entry

Visual evidence:
- Before/after in https://github.com/pingdotgg/t3code/pull/9621#issuecomment-5538549161
- Fit-to-width in the file preview pane (wide flowchart, small flowchart, sequence diagram, invalid fence falling back to code):

Performance notes on this implementation:

- Mermaid loads lazily via `import("mermaid")` only when a mermaid fence reaches the viewport. No cost for threads without diagrams.
- Renders serialize through a shared promise queue (`mermaidRenderQueue`) because Mermaid config is global. Concurrent diagrams cannot clobber each other's config.
- `IntersectionObserver` with `rootMargin: "400px"` starts render just before scroll-in, so diagrams do not pop in late.
- 50-entry SVG cache (`renderedDiagrams`, `MermaidDiagram.tsx:18`) keyed by theme+code. The chat list unmounts off-screen rows, without this every scroll-back re-ran Mermaid, flashed the code fallback, and jumped row height.
- No render while streaming (`renderMermaid = !isStreaming`), so per-token re-renders pay no Mermaid cost. Toggle reset is render-phase gated for the same reason.
- `securityLevel: "strict"` plus `secure` allowlist and `suppressErrorRendering: true`. Failures fall back to source silently, no error UI work.
- Fallback while loading is the already-rendered Shiki-highlighted block, now with upstream's invisible `Suspense` reserve, so no plain-text flash and no layout shift.

| Code view | Rendered view |
|-----------|---------------|
| Mermaid fences shown as source | Mermaid fences rendered as diagrams, fitted to the pane width |

Related: #4571, open #6191 / #8533, closed #4989 / #7497.

Model: Muse Spark in T3 Code.

> [!NOTE]
> ### Render Mermaid diagrams in chat Markdown code fences
> - Non-streaming code fences with the `mermaid` language now render through a new `MermaidDiagram` component in [ChatMarkdown.tsx](https://github.com/pingdotgg/t3code/pull/9621/files#diff-70b5bb003244414202733c39403081a84415e222927a1ef415599958115c0c05); streaming fences and all other languages keep the existing Shiki path.
> - The renderer in [MermaidDiagram.tsx](https://github.com/pingdotgg/t3code/pull/9621/files#diff-118c8c29b6b23ac9054861f3054d8d831874749fc924555c1e323edfdae1cdf3) uses a globally serialized queue: one diagram at a time, dynamic import of Mermaid, strict security settings, light/dark theme, and skips stale or inactive queued work. Pending or failed renders fall back to the highlighted code block.
> - Selecting text inside a rendered diagram copies the original Mermaid source instead of SVG text, via an explicit Markdown copy attribute and the new `serializeMarkdownCodeFence` helper in [markdown-clipboard.ts](https://github.com/pingdotgg/t3code/pull/9621/files#diff-4584df63e01a8b69a0505a0836c9272794c0385577c79484ce7e43b50c06adaa).
> - Styles in [index.css](https://github.com/pingdotgg/t3code/pull/9621/files#diff-70ad657a634ce55957729ff8d1c93f697cc7e3ca2a5cb5fe6b00c90d7d515fcd) constrain Mermaid containers to chat width with horizontal scrolling and hide the line-wrap toolbar button on rendered diagrams.
> - Risk: `MermaidDiagram` dynamically imports Mermaid on first render; if the import fails the code-block fallback is shown but no retry is attempted until the component re-mounts.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b1ef2ca.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added support for rendering Mermaid diagrams in chat messages, including `mermaid` and `mmd` code blocks.
  * Diagrams can be expanded in an accessible preview, with options to copy the source or toggle between the diagram and code.
  * Mermaid content remains available as source while a response is streaming.
* **Bug Fixes**
  * Added fallback to source code when diagrams cannot be rendered.
  * Improved Markdown clipboard handling, diagram sizing, scrolling, and responsive rendering.


gsimone:
can you add visual evidence for this? I'm interested in following :)

Aniketvish0:
sure adding right now

Aniketvish0:
Mermaid diagrams now render in Markdown. @gsimone 

| Code view | Rendered view |
|-----------|---------------|
| Mermaid fences shown as source | Mermaid fences rendered as diagrams, fitted to the pane width |

Qard:
Would be nice to have a way to toggle between the code fence version and the visual version.

Aniketvish0:
> Would be nice to have a way to toggle between the code fence version and the visual version.

Yes would love to add that, but the t3 team commented they are not planning to include mermaid features currently due to performance issues 🥲

mxzinke:
@Aniketvish0 What about the chat? I have quite some cases where Astra is outputing mermaid directly in Chat.

Aniketvish0:
@mxzinke i have updated it to support chat mermaid generation also , and toggle of code to diagram .

Visual evidence from the dev server (chat + file preview):

Sequence diagram and wide flowchart fitted to the pane:

Sequence and wide flowchart rendering

Qard:
Yeah, Astra seems to _love_ outputting Mermaid charts, so it'd be really nice if they were supported.

Aniketvish0:
@Qard i have updated to support that also while taking care of the performance , lets see if they merge this

PR #9468, opened 2026-09-03
fix(server): respawn Claude after an expired login

robertnisipeanu:
Fixes #9607

## Problem

Claude Code loads its OAuth credentials once per process. T3 Code keeps one long-lived CLI per thread, so after the credentials expire that thread keeps answering "Not logged in · Please run /login" even after the user logs in again from a terminal. New threads work because they spawn a fresh CLI. The stuck thread only recovers once the idle reaper kills its process, up to 30 minutes later, or after a manual Stop.

## Fix

When the CLI reports an assistant message with `error: "authentication_failed"`, the Claude adapter:

- fails the turn with the CLI's own message (a message without text gets a runtime error with a fixed fallback instead), and
- stops the session at the end of that turn, so the next send recovers the thread from its persisted resume cursor and spawns a fresh CLI with the current credentials.

The stop runs on the stream fiber itself, so `stopSessionInternal` skips interrupting that fiber and lets the stream end through its own `stopped` check. The first auth-failure message in a turn wins, so a repeated failure does not duplicate the runtime error or replace the CLI's message with the fallback.

## Testing

- `vp test run apps/server/src/provider/Layers/ClaudeAdapter.test.ts` (81 passed, two new tests: auth failure with text, and repeated auth failure without text)
- server typecheck, lint, and format on the touched files

Built with Claude Fable 5.1 in Claude Code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes Claude session lifecycle and turn completion on auth errors; incorrect handling could drop sessions prematurely or leave stale CLI processes, but behavior is scoped to `authentication_failed` and covered by new tests.
> 
> **Overview**
> **Fixes threads stuck on “Not logged in” after the user re-authenticates** by treating Claude CLI `authentication_failed` assistant messages as a hard session failure and tearing down the long-lived process so the next message spawns a fresh CLI with current credentials.
> 
> The **Claude adapter** records the first auth failure per turn in `authenticationError`, fails the turn with the CLI’s text (or a `runtime.error` plus fallback when there is no text), skips duplicating that message as a generic runtime error, and calls **`stopSessionInternal`** after `turn.completed`. **`stopSessionInternal`** no longer interrupts the stream fiber when stop is invoked from that same fiber, avoiding a self-deadlock while the session exits.
> 
> **Tests** cover auth failure with CLI text (failed turn, `session.exited`, no session, one close) and repeated empty auth messages (single runtime error, first message wins).
[...]

PR #9628, opened 2026-09-04
fix(server): respawn Claude sessions after re-login on expired credentials

### LINKED ISSUE #9607: [Bug]: Claude thread stays logged out after re-login until its CLI process is reaped
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Use a thread on the Claude provider until the Claude OAuth session expires. The turn ends with `Failed to authenticate: OAuth session expired and could not be refreshed`.
2. In a terminal, run `claude` and `/login` (or `claude auth login`) and complete the login.
3. Return to the same T3 thread and send another message.

### Expected behavior

The thread picks up the refreshed credentials and the turn runs.

### Actual behavior

Every message in that thread replies `Not logged in · Please run /login`, even though the login already succeeded. A new thread works immediately. The broken thread only recovers after the idle-session reaper kills its CLI process (up to 30 minutes plus the 5 minute sweep) or after the user stops the thread.

Cause: T3 keeps one long-lived Claude CLI process per thread, and the CLI loads OAuth credentials once at startup. Once the credentials it holds expire, the process reports `authentication_failed` on every turn. T3 does not treat that as a reason to end the session, so the same process keeps being reused.

### Impact

### Version or commit

main @ 4e89d74

### Environment

macOS 26.6.0 (Darwin 25.6.0) arm64, Claude provider on claude.ai OAuth

### Logs or stack traces

```shell
{"type":"assistant","error":"authentication_failed", ...}   # content: "Not logged in · Please run /login"
{"type":"result","subtype":"success","is_error":true, ...}
```

Thread replying Not logged in after the user already ran /login

### Workaround

Stop the thread, or start a new thread, after logging in again.

Related: #7878 (how the failure is rendered), #6022 (one cause of the expiry).

ApexYash11:
Closes #9607

## Problem

T3 keeps one long-lived Claude CLI process per thread, and that process loads
OAuth credentials **once at startup**. When a thread's OAuth session expires,
the adapter's process reports `authentication_failed` ("Not logged in") on
every subsequent turn. Because the adapter never treated that as a reason to
end the session, the same poisoned process stayed in place even **after the
user logged in again** via the CLI — until the idle-session reaper eventually
reclaimed it (up to 30 minutes) or the thread was stopped.

## Fix

The adapter now recognizes authentication failures and releases the stale
process so the thread can recover as soon as the user re-logs-in:

- In `handleAssistantMessage`, the CLI's `error: "authentication_failed"`
  stamp on the assistant snapshot flags the session (`context.authenticationFailed`).
- In `handleResultMessage`, once the turn completes, an auth failure
  (from the snapshot flag or the result's error text) closes the runtime
  query. That ends the stream and drives the existing
  `handleStreamExit -> stopSessionInternal` teardown, which removes the session.
- The next `sendTurn` on that thread then hits the **existing** resume-thread
  recovery path in `ProviderService` (the adapter reports `hasSession() == false`
  and the session is resumed from its persisted `resumeCursor`), spawning a
  fresh Claude process that reads the refreshed credentials.

No new recovery machinery is introduced — the fix is purely about detecting the
failure and letting the existing lifecycle respawn. User-aborted/interrupted and
cancelled turns are deliberately left alone.

This complements (and does not duplicate) #9612, which only exempts *live*
Claude sessions from the idle reaper.

## Verification

- New `ClaudeAdapter` test: an auth-failure turn closes the query and leaves
  `hasSession() == false` (so the next turn resumes fresh).
- New `ClaudeAdapter` test: a distinct non-auth failure (rate limit) keeps the
  session alive.
- `ClaudeAdapter.test.ts`: 82/82 pass.
- Server typecheck (`tsgo --noEmit`): clean for the edited files.

Done with Claude (Sonnet) in Cline.

---

> [!NOTE]
> **Medium Risk**
> Changes Claude session lifecycle and `hasSession` semantics during async teardown; misclassification could close healthy sessions or leave stale processes, but scope is limited to auth-error detection and existing resume recovery.
> 
> **Overview**
> Fixes threads that stay stuck on **"Not logged in"** after re-login because the long-lived Claude CLI process only loads OAuth once and never picks up refreshed credentials.
> 
> **`ClaudeAdapter`** now treats CLI auth failures as fatal for the session process: assistant snapshots with `error: "authentication_failed"` set a context flag, and when the turn’s result finishes (unless interrupted/cancelled), matching error text triggers **`query.close()`**, logs `claude.session.stopping-after-auth-failure`, and immediately marks the session **`closed`** so **`hasSession`** is false and follow-up **`sendTurn`** calls fail with **`ProviderAdapterSessionNotFoundError`** instead of enqueueing on a dying runtime. The next successful turn is expected to go through the existing resume-thread path and spawn a fresh process.
> 
> Non-auth failures (e.g. rate limits) are unchanged—the session stays open and **`close()`** is not called.
> 
> **Tests** cover auth-failure teardown plus a negative case where other errors keep the session alive.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 862fdc9664314196f103c38c8090c3a058b1b3f6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix `ClaudeAdapter` to tear down sessions on authentication failure
> - Adds `isClaudeAuthenticationErrorText` to classify known auth failure strings, such as `authentication_failed` and `/login`.
> - `handleAssistantMessage` sets `ClaudeSessionContext.authenticationFailed` when an assistant snapshot's `error` matches the detector.
> - `handleResultMessage` closes the Claude query and marks the session closed on auth failure, unless the turn was interrupted or cancelled. Non-auth failures leave the session open.
> - `hasSession` now returns `false` for sessions marked closed, so follow-up `sendTurn` calls fail with `ProviderAdapterSessionNotFoundError` during teardown.
> - Risk: `hasSession` now requires `sessionStatus` to not be `closed`; if status is not updated correctly during teardown, follow-up turns will be rejected.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 862fdc9.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
+++ b/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
@@ -5626,4 +5626,115 @@ describe("ClaudeAdapterLive", () => {
       Effect.provide(harness.layer),
     );
   });
+
+  it.effect(
+    "stops the session after an authentication failure so the next turn resumes fresh",
+    () => {
+      const harness = makeHarness();
+      return Effect.gen(function* () {
+        const adapter = yield* ClaudeAdapter;
+
+        const sessionExitedFiber = yield* adapter.streamEvents.pipe(
+          Stream.takeUntil((event) => event.type === "session.exited"),
+          Stream.runCollect,
+          Effect.forkChild,
+        );
+
+        const session = yield* adapter.startSession({
+          threadId: THREAD_ID,
+          provider: ProviderDriverKind.make("claudeAgent"),
+          runtimeMode: "full-access",
+        });
+        yield* adapter.sendTurn({
+          threadId: session.threadId,
+          input: "hello",
+          attachments: [],
+        });
+
+        // Exact shape from the field report: the CLI stamps the assistant
+        // snapshot with `error: "authentication_failed"` and "Not logged in"
+        // content, then finishes the turn with an is_error result.
+        harness.query.emit({
+          type: "assistant",
+          session_id: "sdk-session-1",
+          uuid: "assistant-auth-1",
+          parent_tool_use_id: null,
+          error: "authentication_failed",
+          message: {
+            id: "assistant-auth-message-1",
+            content: [{ type: "text", text: "Not logged in · Please run /login" }],
+          },
+        } as unknown as SDKMessage);
+        harness.query.emit({
+          type: "result",
+          subtype: "success",
+          is_error: true,
+          errors: [],
+          session_id: "sdk-session-1",
+          uuid: "result-auth-1",
+        } as unknown as SDKMessage);
+
+        yield* Fiber.join(sessionExitedFiber);
+
+        // The stale-credentialed process must be gone so the next sendTurn goes
+        // through resume-thread recovery with a fresh process. close() runs
+        // twice by design: once from the auth-failure path, once from the
+        // regular stopSessionInternal teardown.
+        const hasSession = yield* adapter.hasSession(session.threadId);
+        assert.equal(hasSession, false);
+        assert.equal(harness.query.closeCalls > 0, true);
+      }).pipe(
+        Effect.provideService(Random.Random, makeDeterministicRandomService()),
+        Effect.provide(harness.layer),
+      );
+    },
+  );
+
+  it.effect("keeps the session alive when a failed result is not an authentication failure", () => {
+    const harness = makeHarness();
+    return Effect.gen(function* () {
+      const adapter = yield* ClaudeAdapter;
+      const session = yield* adapter.startSession({
+        threadId: THREAD_ID,
+        provider: ProviderDriverKind.make("claudeAgent"),
+        runtimeMode: "full-access",
+      });
+      yield* adapter.sendTurn({
+        threadId: session.threadId,
+        input: "hello",
+        attachments: [],
+      });
+
+      harness.query.emit({
+        type: "assistant",
+        session_id: "sdk-session-1",
+        uuid: "assistant-fail-1",
+        parent_tool_use_id: null,
+        message: {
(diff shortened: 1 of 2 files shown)


ApexYash11:
@t3dotgg  can you run the worflow for this!!

robertnisipeanu:
Heads up that #9468 addresses the same issue (#9607) and was opened a day before this one, linked from the issue. Flagging it so the maintainers can triage the two together rather than review both.

For what it's worth, a few differences that might help with that comparison:

- #9468 keys off the exact `error: "authentication_failed"` code on the assistant message. This PR also substring-matches the result's `errors` text (`not logged in`, `/login`, `oauth session expired`). That only sees CLI error strings, not model output, so a false positive is unlikely and would just cost an extra respawn. The exact code already covers the stuck-process case, so the extra patterns mostly add surface without adding coverage.
- #9468 stops the session through the existing `stopSessionInternal` path right after `turn.completed`, with a small guard so a stop issued from the stream fiber does not interrupt itself. This PR calls `query.close()` and then hand-sets the session status to `closed` and widens `hasSession`, relying on the async stream exit to finish the teardown.
- The `authenticationFailed` flag here is session-scoped and never cleared, so an interrupted turn after an auth failure carries it into the next successful turn (Macroscope's open Medium finding). In #9468 the recorded error lives only as long as the session it ends.
- #9468 uses the CLI's own message as the turn failure text and avoids emitting it a second time as a runtime error.
- The negative test here waits on a few `Effect.yieldNow` calls to let the stream settle; AGENTS.md asks for receipt-driven tests rather than timing.

Happy to fold anything from here into #9468 if the maintainers prefer parts of this approach.

PR #9640, opened 2026-09-04
feat(server,contracts): add Claude advisor model setting

schorke:
## Problem

Claude Code can consult a stronger model in the middle of a turn through its advisor tool. T3 Code had no way to turn this on.

It was also already broken when it did run. `CLAUDE_SETTING_SOURCES` includes `user`, so a developer with `advisorModel` in `~/.claude/settings.json` already got advisor consults inside T3 Code. The advisor is a **server-side** tool: the `server_tool_use` block opens the item, but the result arrives in the assistant stream as an `advisor_tool_result` block, never as a `tool_result` user message. `handleUserMessage` is the only path that completes a tool item, and it matches on `tool_result`, so nothing closed the item. The consult rendered as a generic "Tool call" and spun until the end-of-turn sweep swept it up — a lying spinner.

## Fix

- Adds an **Advisor model** field to the Claude provider settings, and passes it to the Agent SDK as the `advisorModel` settings key.
- Completes the advisor item when its own `advisor_tool_result` block arrives, matched by `tool_use_id` rather than block index (the result lands at a different index from the request).
- Titles the item "Advisor consult" instead of the generic tool-call label, and drops the input summary, which is always empty.

The settings key is deliberate, not the hidden `--advisor` CLI flag. The flag aborts session start when the advisor cannot advise the thread model; the settings path logs and runs the turn without an advisor, which is the behavior you want for a per-instance setting.

## Before / after

Claude provider settings, Runtime section:

| Before | After |
| --- | --- |
|  |  |

The field is derived from the contracts schema by `deriveProviderSettingsFields`, so web and desktop pick it up with no component change. Mobile has no provider config form, so there is nothing to add there.

## Tests

- `ClaudeAdapter.test.ts`: the advisor model reaches the SDK, is omitted when the field is empty, and the item completes on its result block. The completion test emits no `result` message on purpose, so the end-of-turn sweep cannot mask the bug. Stubbing the new branch out makes it hang and fail.
- `settings.test.ts`: default, round-trip, and server-patch decoding for the new field.
- `ProviderSettingsForm.test.ts`: field order.

## Not in this PR

- **Advisor token and cost attribution.** `usage.iterations` carries `advisor_message` entries and the final `modelUsage` gains a separate entry for the advisor model with its own `costUSD`, but `emitThreadTokenUsage` reads top-level usage only. Advisor spend stays invisible in the context meter. That is a separate change.
- **`/advisor` in the composer.** Claude's init message lists it, so it appears in the command menu and does nothing useful, because it is an interactive picker that mutates Claude's own `settings.json`. That is why this PR is a setting and not a chat command. T3 Code forwards every Claude built-in unfiltered, so `/model`, `/config`, and about forty others behave the same way. Filtering interactive-only commands is a separate concern.

---

> [!NOTE]
> **Medium Risk**
> Touches Claude stream event handling and session settings for a server-side tool path; mistakes could leave tool items stuck or mis-complete unrelated tools, but scope is narrow and heavily tested.
> 
> **Overview**
> Adds a configurable **Advisor model** on the Claude provider (contracts + settings UI field order) and forwards it to the Agent SDK as `advisorModel` when set, omitting it when empty so Claude Code’s own config can apply.
> 
> Fixes advisor **timeline** handling in `ClaudeAdapter`: the advisor is a server-side tool whose result arrives as an `advisor_tool_result` stream block (not a user `tool_result`). The adapter now completes the in-flight tool item when that block arrives (matched by `tool_use_id`), labels it **Advisor consult** without empty input detail, and keeps `advisor_tool_result` blocks from being dropped in subagent stream filtering.
> 
> Tests cover SDK option wiring, completion without an end-of-turn sweep, settings decode/patch, and provider form fields; user docs describe `fable` / `opus` / `sonnet` and capability rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef51c0de0086c320d52b70b1fb0df02862b84735. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add `advisorModel` setting to Claude provider and handle advisor tool lifecycle in `ClaudeAdapter`
> - Adds `advisorModel` as a trimmed-string field to `ClaudeSettings` and `ClaudeSettingsPatch` in [settings.ts](https://github.com/pingdotgg/t3code/pull/9640/files#diff-82d7525e9c2a6a61cd5d416ceaeba40992140373eb7ac00fa3a91c2edd00bc3c), with an empty default and a form-field position between `homePath` and `autoCompactWindow`.
> - `makeClaudeAdapter` in [ClaudeAdapter.ts](https://github.com/pingdotgg/t3code/pull/9640/files#diff-e696bea66caf21d357bea5ea814bcbb18d1816288cc94247c855823f50fce6dc) now passes a non-empty `advisorModel` into Claude query settings, and handles advisor tool blocks: advisor tool starts use the title "Advisor consult" and omit the generic empty-input detail; advisor result blocks (type `advisor_tool_result`) from nested streams complete the in-flight advisor item by tool-use ID, emitting `item.completed` with no result text.
> - Adds user-facing documentation for the advisor feature in [providers-claude.md](https://github.com/pingdotgg/t3code/pull/9640/files#diff-f3b54bb573c78b526a8c1181ccd8146c7df38119b974038b08aff6f4e5a3819d).
> - Behavioral Change: completed advisor timeline items contain no advice text in their result detail; consumers that previously expected tool-result text for advisor consults will receive an empty result.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ef51c0d.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
+++ b/apps/server/src/provider/Layers/ClaudeAdapter.test.ts
@@ -476,6 +476,130 @@ describe("ClaudeAdapterLive", () => {
     );
   });
 
+  it.effect("passes the configured advisor model to Claude", () => {
+    const harness = makeHarness({ claudeConfig: { advisorModel: "opus" } });
+    return Effect.gen(function* () {
+      const adapter = yield* ClaudeAdapter;
+      yield* adapter.startSession({
+        threadId: THREAD_ID,
+        provider: ProviderDriverKind.make("claudeAgent"),
+        runtimeMode: "full-access",
+      });
+
+      const options = harness.getLastCreateQueryInput()?.options;
+      assert.deepEqual(options?.settings, { advisorModel: "opus" });
+    }).pipe(
+      Effect.provideService(Random.Random, makeDeterministicRandomService()),
+      Effect.provide(harness.layer),
+    );
+  });
+
+  it.effect("omits the advisor model when the setting is empty", () => {
+    const harness = makeHarness();
+    return Effect.gen(function* () {
+      const adapter = yield* ClaudeAdapter;
+      yield* adapter.startSession({
+        threadId: THREAD_ID,
+        provider: ProviderDriverKind.make("claudeAgent"),
+        runtimeMode: "full-access",
+      });
+
+      const options = harness.getLastCreateQueryInput()?.options;
+      assert.equal(options?.settings, undefined);
+    }).pipe(
+      Effect.provideService(Random.Random, makeDeterministicRandomService()),
+      Effect.provide(harness.layer),
+    );
+  });
+
+  it.effect("completes the advisor item when the advisor result block arrives", () => {
+    const harness = makeHarness();
+    return Effect.gen(function* () {
+      const adapter = yield* ClaudeAdapter;
+
+      const runtimeEventsFiber = yield* Stream.take(adapter.streamEvents, 7).pipe(
+        Stream.runCollect,
+        Effect.forkChild,
+      );
+
+      const session = yield* adapter.startSession({
+        threadId: THREAD_ID,
+        provider: ProviderDriverKind.make("claudeAgent"),
+        runtimeMode: "full-access",
+      });
+
+      yield* adapter.sendTurn({
+        threadId: session.threadId,
+        input: "hello",
+        attachments: [],
+      });
+
+      harness.query.emit({
+        type: "stream_event",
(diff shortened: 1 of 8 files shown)


juliusmarminge:

PR #9657, opened 2026-09-04
fix(desktop): quit immediately on a second shortcut press

t3dotgg:
Double-pressing the quit shortcut could wait on a settings read. An unrelated shortcut between released taps could also cause an unintended quit.

Hold mode already accepts two presses within 500 ms. The second press now quits without waiting for settings. Other input cancels the first tap. The default mode and timing window stay unchanged. The hint and settings text explain both gestures.

Hold-to-quit still depends on keyboard-repeat events. Without them, it cancels after about 1.8 seconds. This PR documents that limit but does not add native macOS key-state checks.

## Checks

- 80 focused tests pass.
- Desktop typecheck and targeted lint pass. Lint reports existing warnings in the settings panel.
- Native macOS input was not verified on this Linux machine.
- Browser captures were skipped at the maintainer's request.

### DIFF
--- a/apps/desktop/src/window/QuitHold.test.ts
+++ b/apps/desktop/src/window/QuitHold.test.ts
@@ -252,24 +252,31 @@ describe("makeQuitShortcutHandler", () => {
     expect(harness.notifications).toEqual([]);
   });
 
-  it("honors a quick double press when both key releases beat their mode reads", async () => {
-    const resolvers: Array<(mode: QuitConfirmationMode) => void> = [];
-    const harness = makeHarness({
-      getMode: () => new Promise((resolve) => resolvers.push(resolve)),
-    });
-    await harness.send(makeInput({}));
-    await harness.send(makeInput({ type: "keyUp" }));
-    vi.advanceTimersByTime(QUIT_DOUBLE_PRESS_MS - 100);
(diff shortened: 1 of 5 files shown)


t3dotgg:
> [!NOTE]

The default gesture is unchanged. Hold mode already accepts two presses within 500 ms. [PR #9485](https://github.com/pingdotgg/t3code/pull/9485) restored that behavior before this PR. The [base branch test](https://github.com/pingdotgg/t3code/blob/d5b94100863057fb4629f9ad4a35753d16917924/apps/desktop/src/window/QuitHold.test.ts#L371) verifies it.

This PR removes the settings-read delay on the second press and cancels interrupted taps. The default mode and the 500 ms window stay unchanged.

@macroscope-app Please recheck approvability against the existing behavior.

t3dotgg:
> [!NOTE]

@macroscope-app review this PR

Please recheck eligibility against the [existing Hold-mode double-press test](https://github.com/pingdotgg/t3code/blob/d5b94100863057fb4629f9ad4a35753d16917924/apps/desktop/src/window/QuitHold.test.ts#L371). This fixes the settings-read delay. It does not introduce the gesture or change the default mode.

PR #9674, opened 2026-09-04
fix(web): remember the last selected usage tab

baptisteArno:
## Problem

On the Usage page, switching to the **Limits** (or **Tokens**) tab was forgotten as soon as you navigated away. Coming back always landed on **Cost**, so users who mostly care about their limits had to re-select the tab every time.

## Fix

The selected metric is now stored in `localStorage` (`t3code:usage-metric:v1`) and used as the initial state when the page mounts. Any stale or unknown stored value falls back to `cost` through the existing `isUsageMetric` guard. Mobile is unaffected: it has no Limits tab, Limits is a separate section there.

## Demo

https://github.com/user-attachments/assets/d06e55a7-3beb-4760-9860-768dcd2fbb95

Made with Claude Fable 5.1 via Claude Code in T3 Code.

> [!NOTE]
> ### Persist last selected usage metric tab in `UsagePage`
> Adds a React effect and a versioned local-storage key so the usage metric selection survives page reloads. The initial state reads from local storage only when the stored value is a recognized metric; otherwise it falls back to the cost metric. Local-storage access is guarded against missing or failing storage. Risk: stored metric values from older key versions are ignored and fall back to the cost metric.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cf7a0b5.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/usage/UsagePage.tsx
+++ b/apps/web/src/components/usage/UsagePage.tsx
@@ -64,10 +64,20 @@ export function UsagePage() {
     window: makeWindow(30),
   }));
   const [metric, setMetric] = useState<UsageMetric>(() => {
-    const stored = localStorage.getItem(USAGE_METRIC_STORAGE_KEY);
-    return isUsageMetric(stored) ? stored : "cost";
+    try {
+      const stored = localStorage.getItem(USAGE_METRIC_STORAGE_KEY);
+      return isUsageMetric(stored) ? stored : "cost";
+    } catch {
+      return "cost";
+    }
   });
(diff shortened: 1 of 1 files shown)


baptisteArno:
@macroscope-app review this PR

juliusmarminge:
Superseded by #10189, which landed the same Usage page persistence (metric selection in localStorage) and also remembers the time range.

Closing this as superseded — thanks for the earlier work on this!

juliusmarminge:
Closing as superseded by #10189.

Issue #9661, opened 2026-09-04
Track performance audit fixes


t3dotgg:
Track the performance work from the 2026-09-04 audit. Each implementation PR has one concern, focused tests, and a review before merge.

The audit has measured findings and profiling targets. A target is complete only after a verified improvement, or a written decision that the change is not justified. Browser and device use is excluded at Theo's request.

## Work

- [ ] C1. Terminal output becomes full-buffer replay after 512 KiB.
- [ ] C2. One shell change rebuilds environment-wide indexes and atom dependencies.
- [x] C3. Five-minute thread caches retain live subscriptions, not just values.
- [x] C4. Loaded image attachments each start an uncached signing RPC before row visibility.
- [x] C5. One relay token refresh blocks cache hits for other scopes.
- [x] S1. Terminal history repeatedly splits old output and has no byte cap.
- [ ] S2. Session updates and command preparation read full conversation bodies.
- [ ] S3. Fresh checkpoint indexes discard Git's clean-file stat cache.
- [x] S4. Every buffered assistant text event still causes three SQL reads.
- [x] S5. Slow-client stream queues have no retained-byte bound.
- [ ] S6. OpenCode keeps unused versions of growing tool output.
- [ ] S7. Chat search scans message bodies before applying its result limit.
- [x] S8. Self-hosted hashed web assets lack cache validators and immutable caching.
- [x] S9. Automatic checkpoint summaries build full patches just to count lines.
- [ ] S10. Nine cursor writes run for every event, including inactive projectors.
- [ ] S11. Each client repeats the same shell projection database reads.
- [ ] S12. The light startup command model fills with message and activity bodies during uptime.
- [ ] S13. Thread pages repeat all plans and checkpoint summary data.
- [x] S14. Single-thread resume decodes unrelated threads' events.
- [ ] S15. Native content grep is synchronous on Node and has a 250 ms search budget.
- [ ] S16. An idle terminal prompt keeps one-second process-table polling active.
- [x] S17. OpenCode native logs write repeated growing tool payloads.
- [ ] S18. Slow provider preparation and checkpoints can delay unrelated worker items.
- [x] W02. Compiled streaming Markdown remounts content and repeats code highlighting.
- [x] W03. Dense terminal snapshots make tens of thousands of WASM calls.
- [x] W04. Draft persistence serializes all draft data before its debounce.
- [x] W05. The first chat starts a shared 2 to 6 worker diff pool before any diff request.
[...]

PR #9684, opened 2026-09-04
perf(server): stop retaining unused OpenCode tool history

t3dotgg:
OpenCode keeps every version of tool output in a session history array that nothing reads. The audit fixture retained 5,171,200 logical output bytes for a tool whose final output was 102,400 bytes.

Remove that array and its append helpers. Thread reads and rollback still fetch upstream messages. Keep the live part map and event delivery unchanged.

Verified with 116 existing adapter and provider tests, server typecheck, and targeted lint. No browser, device, or live state was used.

Part of https://github.com/pingdotgg/t3code/issues/9661. Other session-map retention remains separate work.

---

> [!NOTE]
> **Low Risk**
> Deletion-only memory optimization with no callers of the removed snapshot path; thread read/rollback still use upstream APIs.
> 
> **Overview**
> Removes the in-memory **`turns`** array on OpenCode session context and the **`resolveTurnSnapshot` / `appendTurnItem`** helpers that pushed every tool **`Part`** update into it during **`message.part.updated`** handling.
> 
> That history was never read in the adapter—only grown—so long tool streams could retain huge redundant output while live behavior stays on **`partById`**, text delta maps, and runtime event emission.
> 
> **`readThread`** and **`rollbackThread`** are unchanged: they still load conversation shape from **`session.messages`** / **`session.revert`**, not from session context.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ecc566951fe8f6b9c2d2489cba7477ae8f4f7514. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Remove unused OpenCode turn-snapshot retention from `OpenCodeAdapter`
> - Removes the turn-snapshot collection from `OpenCodeSessionContext` and the `resolveTurnSnapshot` and `appendTurnItem` helpers, as this history was retained but never read.
> - The `makeOpenCodeAdapter` session-context initialization no longer allocates snapshot storage, and the event-part handler stops recording processed parts to per-turn snapshots before emitting the runtime event.
> - Behavioral Change: OpenCode session contexts no longer hold in-memory turn snapshots; runtime event emission is unchanged.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ecc5669.</sup>
> 
>

### DIFF
--- a/apps/server/src/provider/Layers/OpenCodeAdapter.ts
+++ b/apps/server/src/provider/Layers/OpenCodeAdapter.ts
@@ -338,7 +338,6 @@ interface OpenCodeSessionContext {
   readonly partById: Map<string, Part>;
   readonly emittedTextByPartId: Map<string, string>;
   readonly completedAssistantPartIds: Set<string>;
-  readonly turns: Array<OpenCodeTurnSnapshot>;
   activeTurnId: TurnId | undefined;
   activeAgent: string | undefined;
   activeVariant: string | undefined;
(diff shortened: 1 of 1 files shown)


francoisrob:
how is your experience with gpt6 @t3dotgg ?

PR #9708, opened 2026-09-04
perf(relay): avoid repeated activity decoding

t3dotgg:
Every activity read converts Postgres JSON objects to strings and parses them again. Notification-only publications read the activity list even though they do not use it.

Validate JSON objects directly and skip the unused read. Keep query results, notification delivery, and retry identity checks unchanged.

The real service returned identical JSON for 10,000 rows in a disposable local Postgres database. Median read time fell from 343 ms to 202 ms. All 60 focused tests, relay typecheck, and targeted lint pass. No browser, device, live state, or manual deployment was used.

Expired-row filtering and pruning remain separate work in https://github.com/pingdotgg/t3code/issues/9661.

---

> [!NOTE]
> **Medium Risk**
> Touches relay delivery and persistence decoding paths; behavior should be equivalent but invalid rows are silently filtered and notification-only users no longer load cross-thread activity lists.
> 
> **Overview**
> **Agent activity reads** no longer stringify Postgres `jsonb` and parse it again on `listForUser` / `getForUserThread`; rows are validated in place with `RelayAgentActivityState` schema, and invalid or null payloads are dropped (new test covers mixed shapes including legacy string JSON).
> 
> **Notification-only delivery** skips `listForUser` when `liveActivitiesEnabled` is false, since aggregates for push-only users are built from the current publish state only; the publisher test now fails if rows are queried in that path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e34ae550f16a8c6f59486000e75d29306076df39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Skip activity-row queries when Live Activities disabled and decode `stateJson` directly in `AgentActivityRows`
> - `AgentActivityPublisher.publishForDeliveryUser` now skips `AgentActivityRows.listForUser` when `liveActivitiesEnabled` is false, using an empty active-state list instead.
> - `AgentActivityRows.listForUser` and `getForUserThread` decode each row's `stateJson` field directly against `RelayAgentActivityStateSchema`, removing the intermediate JSON-string encoding step.
> - Invalid rows are skipped (omitted from `listForUser`, passed over until the first valid row in `getForUserThread`).
> - Behavioral Change: `decodeRelayAgentActivityState` no longer applies the removed JSON-value encoding helper; any caller relying on the old JSON-string round-trip path must decode `stateJson` directly against the schema.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e34ae55.</sup>
> 
>

### DIFF
--- a/infra/relay/src/agentActivity/AgentActivityPublisher.test.ts
+++ b/infra/relay/src/agentActivity/AgentActivityPublisher.test.ts
@@ -517,7 +517,7 @@ describe("AgentActivityPublisher", () => {
   });
 
   it.effect(
-    "does not build Live Activity aggregates for links with Live Activities disabled",
+    "delivers notifications without querying activity rows when Live Activities are disabled",
     () => {
       const notificationState: RelayAgentActivityState = {
         ...state,
@@ -549,13 +549,7 @@ describe("AgentActivityPublisher", () => {
                     AgentActivityRows.AgentActivityRows,
                     makeAgentActivityRows({
                       listForUser: () =>
-                        Effect.succeed([
-                          {
-                            ...state,
-                            environmentId: "other-env" as RelayAgentActivityState["environmentId"],
-                            threadId: "other-thread" as RelayAgentActivityState["threadId"],
-                          },
-                        ]),
+                        Effect.die("Notification-only delivery must not read rows"),
                     }),
                   ),
                   Layer.succeed(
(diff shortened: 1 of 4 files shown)


t3dotgg:
> [!NOTE]

Automatic approval is neutral because this touches relay delivery and decoding. I checked those paths and obtained an independent source review.

Notification-only delivery still sends the same notification aggregate and calls the Live Activity delivery path with null. It skips only a list that the disabled aggregate did not use. Both SQL filters and ordering stay unchanged. Direct schema decoding keeps valid rows and rejects the same invalid JSON shapes.

All 60 focused tests pass. The real Postgres fixture returns identical JSON for 10,000 rows. CI, Cursor, Macroscope correctness, and Effect checks pass on e34ae550. The neutral eligibility check is not an approval.

PR #9715, opened 2026-09-04
fix(server): bound slow-client event buffers

t3dotgg:
Slow WebSocket clients can keep accumulating shell and thread events without a memory bound.

Add a shared budget of 1,000 live items and 8 MiB of serialized payload per subscription. Count queued work, coalescing, and the batch waiting for an ACK. On overflow, stop the producer, release queued work, and return `OrchestrationGetSnapshotError`. Clients retry from their last applied sequence after 250 ms. This budget does not cap authoritative snapshots. Wire contracts and replay queries stay unchanged.

Verified with 30 focused tests, server typecheck, and file lint. Real WebSocket tests hold an ACK during overflow and verify producer cleanup, exact replay, and snapshot recovery after deletion.

With no consumer, 20,000 events previously retained 20.48 MB of text. The new code accepts at most 1,000 items and clears queued payload. The small-message coalescer benchmark is 10% faster. Cold byte sizing adds about 1.8 ms per 1 MiB tool event. Other subscriptions reuse that size.

Part of https://github.com/pingdotgg/t3code/issues/9661.

---

> [!NOTE]
> **Medium Risk**
> Changes live WebSocket subscription buffering, teardown, and error semantics on the orchestration hot path; behavior is heavily tested but mis-handling overflow cleanup could drop or duplicate events for slow clients.
> 
> **Overview**
> Adds a per-subscription **live stream budget** (default **1,000 items** and **8 MiB** of serialized payload) so shell and thread WebSocket tails cannot grow without limit when a client is slow or stops ACKing batches.
> 
> New `LiveStreamBudget` charges **queued**, **coalescing**, and **in-flight (pre-ACK)** work, uses atomic `replace` during coalescing, and on overflow fails with `OrchestrationGetSnapshotError` (“resume from the last received sequence”), tears down the delivery source, and releases buffered events while keeping the current unacknowledged batch accounted for until scope cleanup.
> 
> `subscribeShell` and `subscribeThread` now retain live inputs through the budget, race PubSub consumption against budget failure so producers detach even without another pull, and wrap outbound streams with `budget.deliver`. `ThreadLiveEventCoalescer` is reworked around `offer`/`offerAll`, budget-backed output, and coalescer-scoped flush timers instead of `offerAndWait`/`takeAll`.
> 
> Integration tests simulate a **held first WebSocket ACK** during overflow and assert producer detach plus thread **event replay** and shell **snapshot recovery** on resubscribe.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f228395cbfb7c44b8702f80534bb12f26308746. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Bound slow-client event buffers with `LiveStreamBudget` in server live subscriptions
> - Adds `LiveStreamBudget` with configurable item (default 1,000) and serialized-byte (default 8 MiB) limits; retained items stay charged until acknowledged by the next delivery pull or released by scope cleanup
> - Rewrites `ThreadLiveEventCoalescer` and the shell subscription path in `ws.ts` to route all retained events, coalesced batches, and sync markers through the shared budget instead of unbounded queues
> - On overflow the budget latches `OrchestrationGetSnapshotError`, closes the underlying source, releases pending non-delivered items even while the consumer holds an unacked batch, and terminates the delivery stream
> - Replaces `offerAndWait`/`takeAll` APIs with `offerAll`/`offer` + an ACK-aware stream; completion markers now go through the same buffered stream
> - Behavioral Change: live thread and shell subscriptions that exceed budget now fail with `OrchestrationGetSnapshotError` and detach their PubSub subscriber instead of buffering indefinitely; `offerAndWait` and `takeAll` are removed from the coalescer API
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8f22839.</sup>
> 
>

### DIFF
--- a/apps/server/src/orchestration/LiveStreamBudget.test.ts
+++ b/apps/server/src/orchestration/LiveStreamBudget.test.ts
@@ -1,0 +1,47 @@
+import { it } from "@effect/vitest";
+import * as Deferred from "effect/Deferred";
+import * as Effect from "effect/Effect";
+import * as Queue from "effect/Queue";
+import * as Stream from "effect/Stream";
+import { describe, expect } from "vite-plus/test";
+
+import { makeLiveStreamBudget, type RetainedLiveItem } from "./LiveStreamBudget.ts";
+
+describe("LiveStreamBudget", () => {
+  it.effect("closes the source without releasing a batch still waiting for an ACK", () =>
+    Effect.scoped(
+      Effect.gen(function* () {
+        const budget = yield* makeLiveStreamBudget({ maxItems: 3 });
+        const queue = yield* Queue.unbounded<RetainedLiveItem<{ text: string }>>();
+        const sourceClosed = yield* Deferred.make<void>();
+        const first = yield* budget.retain({ text: "first" });
+        const second = yield* budget.retain({ text: "second" });
+        const third = yield* budget.retain({ text: "third" });
+        yield* Queue.offerAll(queue, [first, second, third]);
+        yield* Effect.scoped(
+          Effect.gen(function* () {
+            const pull = yield* Stream.toPull(
+              budget.deliver(
+                Stream.fromQueue(queue).pipe(
+                  Stream.rechunk(1),
+                  Stream.ensuring(Deferred.succeed(sourceClosed, undefined)),
+                ),
+              ),
+            );
+            expect(yield* pull).toEqual([{ text: "first" }]);
+            // The other two items are in the source's pull state, not its queue.
+            expect(yield* Queue.size(queue)).toBe(0);
+            expect((yield* budget.usage).retainedItems).toBe(3);
+            const overflow = yield* budget.retain({ text: "fourth" }).pipe(Effect.result);
+            expect(overflow._tag).toBe("Failure");
+            // Do not resume the consumer. Its source scope must close now.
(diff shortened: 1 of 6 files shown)


t3dotgg:
> [!NOTE]

I reviewed the new limits and recovery paths. Pending work, queued output, and the batch waiting for ACK share one budget. Overflow stops the producer, closes grouped source work, and keeps only the actual in-flight batch until delivery ends. Recovery uses the existing typed error and cursor retry, with no silent continuation after overflow.

The one-line diagnostic suppression is intentional. Stream.fromPull requires the outer effect to return the pull effect. Yielding that effect there would consume data during setup.

All 30 focused tests pass. I independently reran the 10 budget/coalescer tests and both real withheld-ACK recovery tests. CI, Cursor, and Macroscope correctness/conventions pass on 8f228395. Automatic eligibility is neutral.

The limits cover serialized live payload, not total V8 heap. Authoritative snapshots keep their existing behavior. Cold sizing adds about 1.8 ms per 1 MiB tool update in the local coalescer benchmark; later subscribers reuse the measured size. This cost is accepted to bound retained live data.

PR #9768, opened 2026-09-04
perf(server): skip unused Linux process detail reads

t3dotgg:
The Linux resource monitor reads command lines, memory, and I/O counters for every process, then discards most of that data.

Read those details only for selected processes and the monitor's identity reference. Keep CPU baselines and process-tree discovery in the existing persistent collector. Release the temporary detail data and process handles after each sample.

The identity check accounts for clock adjustments and preserves the existing whole-second comparison. macOS and Windows keep their existing refresh path. The protocol, sampling intervals, and history limits are unchanged.

Linux release-binary measurements, with 12 warm samples per version in alternating order:

| Retained processes | Scanned processes | Before | After |
| ---: | ---: | ---: | ---: |
| 3 | 364 | 20.35 ms | 11.60 ms |
| 33 | 364 to 369 | 19.64 ms | 14.02 ms |

These are collector timings, not app response times or measurements on other platforms. The fixtures select only owned processes.

Verified with 17 Rust tests, rustfmt, clippy, and release-protocol checks for CPU, memory, I/O, command changes, child exits, and external-process identity. A process-local fixture verifies forward and backward clock adjustments without changing the host clock. No browser or device was used.

This is the Linux part of R3 in the [performance audit](https://github.com/pingdotgg/t3code/issues/9661). Other platforms still need profiling.

---

> [!NOTE]
> **Medium Risk**
> Sampling logic now uses two refresh paths on Linux with identity checks; incorrect offset or matching could drop tracked processes or mis-attribute metrics, though behavior is gated to Linux and covered by new tests.
> 
> **Overview**
> **On Linux**, the resource monitor no longer loads command lines, memory, disk I/O, and related metadata for every process on each sample. The persistent collector now does a **lightweight discovery refresh** (CPU-focused) to build the process tree and CPU baselines, then a **second, scoped refresh** only for tracked PIDs (plus the monitor process for calibration) to fill snapshot fields.
> 
> CPU percent and accumulated time still come from the long-lived `System` instance so baselines are not reset by detail reads. Memory, virtual size, I/O counters, and command text come from the temporary detail pass, which is **dropped after each sample** to release handles. A **start-time check with boot-epoch offset** between the two `System` instances rejects PID reuse and tolerates clock shifts; macOS and Windows keep the previous single-pass refresh behavior.
> 
> Adds integration-style tests for “process newly selected” detail loading and unit tests for start-time matching under forward/backward clock adjustment.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de947bb5960445ad0d0e759b3a314c143056aaa1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Skip unused Linux process detail reads in `Collector` CPU baseline and sampling
> - Linux discovery and CPU baseline refreshes now use a CPU-only, task-free refresh kind via `process_discovery_refresh_kind`, avoiding metadata fields that reset Linux process timing data
> - For selected processes on Linux, `Collector::sample` creates a temporary `System` to refresh only non-CPU details (command, disk, resident memory, virtual memory); CPU-related fields still come from the discovery `System`
> - Added `matches_process_start_time` to accept a detail record only when its start time exactly matches the discovered start time after adjusting for the epoch offset between the two `System` instances; mismatches cause the process to be omitted
> - Non-Linux platforms retain the existing single-refresh path
> - Behavioral Change: Linux process records with unavailable detail refreshes or non-matching adjusted start times are now omitted rather than emitted with stale or partial data; reviewers should check `matches_process_start_time` and the epoch-offset computation in `Collector::sample` for edge cases around clock drift or PID reuse
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized de947bb.</sup>
> 
>

### DIFF
--- a/native/resource-monitor/src/main.rs
+++ b/native/resource-monitor/src/main.rs
@@ -346,7 +346,7 @@ impl Collector {
         self.system.refresh_processes_specifics(
             ProcessesToUpdate::All,
             true,
-            process_refresh_kind(),
+            process_discovery_refresh_kind(),
         );
         self.cpu_baseline_refreshed_at = Some(Instant::now());
     }
@@ -361,7 +361,7 @@ impl Collector {
         self.system.refresh_processes_specifics(
             ProcessesToUpdate::All,
             true,
-            process_refresh_kind(),
+            process_discovery_refresh_kind(),
         );
         self.cpu_baseline_refreshed_at = Some(Instant::now());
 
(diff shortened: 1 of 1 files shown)


t3dotgg:
> [!NOTE]

I reviewed the collector change with the server audit agent. CI and correctness checks pass at `de947bb5960445ad0d0e759b3a314c143056aaa1`. Automated eligibility remains neutral because this changes native collection.

The fast path is Linux-only. macOS and Windows keep their original refresh path. All 17 Rust tests and the release-protocol checks pass. Both reviewers verified forward and backward clock adjustments with a process-local fixture. Missing or mismatched detail reads use the existing inaccessible-process count. Sampling intervals, history limits, and the protocol stay unchanged.

I am merging through the normal repository policy.

Issue #5410, opened 2026-08-05
[Bug]: Desktop server private memory climbs ~25-30 MB/h while idle (immortal Effect parent spans + always-on tracing)


Benchance:
﻿### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

### Steps to reproduce

1. Launch T3 Code desktop and leave a normal workspace open (window can be unfocused).
2. Confirm background profile is `balanced` (or `custom` with `baseProfile: balanced`) so provider health checks only run while foreground.
3. Do **not** interact with the Providers page; leave the app idle for several hours.
4. Sample the **server** child (`…\T3 Code (Alpha).exe …\bin.mjs`), not the renderer — prefer **Private Bytes** (Working Set can be trimmed by Windows and falsely drop).
5. Optionally inspect `~\.t3\userdata\logs\server.trace.ndjson` for high-frequency spans parented by span IDs that never appear as ended `"spanId"` records.

### Expected behavior

Idle server Private memory should be roughly flat. Background no-op polls should not retain unbounded tracing state. Startup spans for forever-running reactors should not stay open / pinned on long-lived fibers for the process lifetime.

### Actual behavior

Server **Private** climbs steadily at ~**25–30 MB/h** with **handle count flat**. Growth continues for hours **with zero provider health-check spans**. Dominant idle work in traces is no-op `PortDiscovery.pollTick` (~3s), `ServerSecretStore.get` (~5s), and `ws.rpc.server.reportClientActivity` (~25s), mostly nested under **parent spans that never end / remain pinned on parked fibers**.

### Impact

Major degradation or frequent failure

### Version or commit

T3 Code (Alpha) desktop; server entry `resources\app.asar\apps\server\dist\bin.mjs`

### Environment

Windows 11, desktop mode, local tracing writing `~\.t3\userdata\logs\server.trace.ndjson`

### Logs or stack traces

Hourly Private growth (same PID, no provider probes after ~16:15):

```
16:29  Priv=785.2MB  handles=283
17:30  Priv=799.9MB
18:30  Priv=827.3MB
19:30  Priv=865.7MB
20:30  Priv=893.9MB
21:30  Priv=933.3MB
22:30  Priv=960.3MB  (WS was OS-trimmed to 202MB; Private kept rising)
23:30  Priv=999.1MB
```

Slope ≈ **+29–30 MB/h Private** over the watch window.

[...]

PR #9824, opened 2026-09-04
fix(server): detach immortal parent spans from idle background work

lnieuwenhuis:
Long-lived parked fibers and the PortDiscovery poll loop retain their caller's parent span, even when a replacement span hides it from service lookup.

Fork through `forkScopedDetached`, which creates a fresh external parent and flattens the entire context before the child is created. Both activation paths and the port poller use it, preserving other services, the activation barrier, and scoped interruption. Idle ticks and secret reads avoid unnecessary spans.

Regression tests inspect retained context references in both activation paths and before child execution, and cover parent restoration and scoped interruption. The focused activation and port scanner suites pass (28 tests).

Related to #5410 and supersedes the approach in #5411. Removing this retained reference does not establish that Windows idle Private Bytes growth is resolved; a comparable long-running idle-memory measurement is still needed.

Originally built with muse-spark-1.3-contributor via OpenCode in T3 Code; retention fix implemented with GPT-6 via Codex.

---

> [!NOTE]
> **Medium Risk**
> Changes tracing context for all forkParked background fibers and the preview port poll loop; behavior is intentional but affects observability and any code assuming inherited trace parents.
> 
> **Overview**
> Long-lived server fibers (parked roots via **`forkParked`**, and the **PortDiscovery** poll loop) no longer inherit the ambient **`ParentSpan`**, which was pinning ever-growing trace state while the process idled.
> 
> A new **`withDetachedSpan`** re-roots work under a fresh stateless **external** span (typed so **`ParentSpan`** is satisfied without inheriting the caller’s span). **`forkParked`** always forks through that helper, so reactors like checkpoint/provider command processing get the same behavior.
> 
> **PortScanner** splits idle vs active poll ticks: the retain-count check runs outside **`Effect.fn`**, so idle ticks emit **no** **`PortDiscovery.pollTick`** spans under an ambient parent; active work still uses the named span. The repeating poll fiber is wrapped in **`withDetachedSpan`**, and the layer **`make`** span was removed.
> 
> **`ServerSecretStore.get`** drops its span in favor of **`Effect.withTracerEnabled(false)`** on reads. Tests cover re-rooting, **`forkParked`** non-inheritance, and idle poll span behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae9bf6b46974fbb22393fe631fde00e1dcce2095. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Detach background fork and idle poll spans from ambient `ParentSpan`
> - Adds `withDetachedSpan` in [serverActivation.ts](https://github.com/pingdotgg/t3code/pull/9824/files#diff-f1b3973a0df9d977d0fb183a925ff0d4721298d00620d950bc2e7c7f3e892c0a), which creates a fresh sampled external span and provides it as `Tracer.ParentSpan` for the wrapped effect
> - Applies `withDetachedSpan` in `forkParked` so both activation-gated and ungated background forks run under a fresh root span instead of inheriting the caller's span
> - Reworks `PortDiscovery` polling in [PortScanner.ts](https://github.com/pingdotgg/t3code/pull/9824/files#diff-0f6b3292e660acf982893ba130c26e1ffa8f91ff09032cce9b07d55074e296f4): the repeating poll runs under a detached span, the `PortDiscovery.make` wrapper span is removed, and idle ticks return before the instrumented `pollTickActive` body so they no longer emit `PortDiscovery.pollTick` spans
> - Disables tracing for the secret read wrapper in [ServerSecretStore.ts](https://github.com/pingdotgg/t3code/pull/9824/files#diff-ba1dc01960ef9039f0c35daa5ed29eb4c63ae7c4202faf83b04259d422c1c557), removing the `ServerSecretStore.get` span
> - Risk: `withDetachedSpan` removes `Tracer.ParentSpan` from the required environment of wrapped effects; any caller that explicitly provides or depends on the ambient `ParentSpan` for forked work will now observe a different root span. The gated `forkParked` path preserves the same detached root across the activation release boundary.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c092eac.</sup>
> 
>

## Summary by CodeRabbit

- **Observability**
  - Background operations now use independent tracing contexts for more accurate diagnostics.
  - Idle port-scanning polls no longer generate unnecessary tracing spans.
  - Server secret retrieval no longer produces a tracing span, reducing unrelated telemetry noise.
- **Bug Fixes**
  - Corrected tracing behavior for deferred activation and background processing so unrelated parent operations are not incorrectly associated.


Benchance:
Linked from #5410. Please add `Fixes #5410` to this PR description so the bug auto-closes on merge (#5411 is already closed as superseded by this).

shivamhwp:

The background fiber still retains the original `ParentSpan` even though service lookup returns the external span. Effect's `Context.add` keeps the previous context in its overlay chain. Providing the replacement inside the child also keeps a restoration frame alive for the lifetime of the wrapped work. The original parent remains reachable until the background scope closes, including on the activation-gated path.

Detach before forking, and give the child a fresh flattened context with the replacement parent. Apply that to both `forkParked` branches and the PortDiscovery polling fork. Merely moving the current `withDetachedSpan` outside `forkScoped` still retains the parent through the context overlay. Preserve the other services, activation barrier, and scoped interruption.

Keep the description's limitation on #5410: removing this retained reference does not establish that Windows idle Private Bytes growth is resolved.

lnieuwenhuis:
Addressed in 30aa6efc9. The feedback matches Effect 4.0.0-rc.112: a replacement service can leave the old span in the context overlay/cache root, and providing it inside the child leaves a restoration frame alive.

`forkScopedDetached` now flattens the context into a fresh map and replaces it around `forkScoped`, before the child exists. Both `forkParked` branches and the PortDiscovery poller use that helper. Regression tests inspect retained references in both activation paths (both failed before the fix), check the child context before execution, and verify other services, caller parent restoration, the activation gate, and scoped interruption. All 28 focused activation/port-scanner tests, server typecheck, and targeted lint pass.

The description retains the #5410 limitation: this removes the demonstrated reference but does not prove Windows idle Private Bytes growth is resolved.

PR #9844, opened 2026-09-04
fix(source-control): allow clones to finish without a deadline

### LINKED ISSUE #9780: [Bug]: Clone fails when git clone exceeds a fixed 120s timeout
### Area

apps/server (SourceControlRepositoryService)

### Summary

`SourceControlRepositoryService.cloneRepository` runs `git clone` with a hardcoded `timeoutMs: 120_000`. Any clone that takes longer than two minutes, whether because the repository is large or the link is slow, is killed mid-download and surfaces as the generic toast:

> Clone failed. Source control repository operation cloneRepository failed for unknown: The source control operation could not be completed.

The trace log shows the real cause: `GitCommandError: ... Git command timed out.`

Clone duration scales with repository size and network speed, so any fixed budget strands someone. Every `push` variant in `GitVcsDriverCore` already runs with `timeoutMs: null` for this reason. Clone should follow the same precedent.

### Steps to reproduce

1. Pick any repository whose clone takes more than two minutes on the server's connection.
2. Clone it via "Add project from remote" in the web app.

### Expected

The clone completes, or fails with the actual git error.

### Actual

At 120s the clone is killed and the generic "could not be completed" toast appears. Nothing indicates a timeout.

### Notes

- The 120s value dates from #2482 with no recorded rationale.
- #5614 / #3902 fixed the same shape of problem for `removeWorktree`.
- #5005 (open) improves clone error messages, touches the same call, but keeps the 120s budget.

Gigioxx:
Cloning a large repository or using a slow connection fails after two minutes because the server kills Git at a fixed deadline.

Set the clone timeout to `null`, matching the existing push behavior. This is one production-line change in the shared service used by web, desktop, and mobile.

Reproduced before the fix with the real Git driver and a simulated three-minute process: `Git command timed out.`, wrapped in the reported generic repository error. The same virtual-clock regression now passes. All 8 repository-service tests, server typecheck, targeted lint, and formatting pass. No network download or browser verification is claimed.

Fixes #9780.

Created with GPT-6 in Codex.

---

> [!NOTE]
> **Medium Risk**
> Shared clone path for web, desktop, and mobile no longer enforces a server deadline, so stuck clones can run indefinitely until killed elsewhere, though behavior is intentional and output limits remain.
> 
> **Overview**
> **Removes the fixed two-minute kill on server-side `git clone`**, so large repos or slow networks are no longer aborted with a timeout error.
> 
> `SourceControlRepositoryService.cloneRepository` now passes `timeoutMs: null` to the Git driver instead of `120_000`, aligning clone with other long-running Git work (e.g. push). Output is still capped at 256 KiB.
> 
> A new effect test simulates a clone that exits after three minutes using `TestClock` and a mocked child process, asserting the operation completes successfully.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63bb6f36e52ed5d8a20df5c51db27f8045f85837. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Remove 120-second timeout from `cloneRepository` Git command
> The Git clone process previously had a fixed 120-second execution deadline, which could abort large or slow clones before they finished. `cloneRepository` now runs with no configured timeout, so the command continues until it completes or fails on its own. A new test uses a fake Git process with a simulated three-minute runtime to verify the clone still succeeds past the old limit.
> - Risk: removing the timeout means a hung clone process can run indefinitely; any caller relying on the 120s deadline will now block until the process exits or fails.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 63bb6f3.</sup>
> 
>

### DIFF
--- a/apps/server/src/sourceControl/SourceControlRepositoryService.test.ts
+++ b/apps/server/src/sourceControl/SourceControlRepositoryService.test.ts
@@ -2,10 +2,15 @@ import * as NodePath from "@effect/platform-node/NodePath";
 import * as NodeServices from "@effect/platform-node/NodeServices";
 import { assert, it } from "@effect/vitest";
 import * as Effect from "effect/Effect";
+import * as Deferred from "effect/Deferred";
 import * as FileSystem from "effect/FileSystem";
+import * as Fiber from "effect/Fiber";
 import * as Path from "effect/Path";
 import * as Layer from "effect/Layer";
 import * as PlatformError from "effect/PlatformError";
+import * as Sink from "effect/Sink";
+import * as Stream from "effect/Stream";
+import * as TestClock from "effect/testing/TestClock";
 import { ChildProcessSpawner } from "effect/unstable/process";
 
 import { GitCommandError, SourceControlProviderError } from "@t3tools/contracts";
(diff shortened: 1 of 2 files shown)


Gigioxx:
Keeping the current fix and leaving the default-behavior decision for human review.

The fixed deadline is the cause of #9780: clone duration depends on repository size and connection speed, so raising the limit would only move the failure threshold. `timeoutMs: null` follows the existing Git push behavior.

The tradeoff is that a stalled clone no longer has an application-imposed deadline. Explicit cancellation or stall detection would be separate work; this PR does not add either. The regression exercises the real Git timeout logic with a simulated three-minute process and fails before the fix. CI, Bugbot, and Macroscope's correctness check passed.

Could a maintainer review and approve this default-behavior change before merge?

juliusmarminge:
Superseded by #11762, which landed background cloning with no clone deadline (and already closed #9780). Closing this PR as obsolete.

Issue #5792, opened 2026-08-09
[Bug]: Browser automation steals composer focus across threads


AksharP5:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

Same-thread case:

1. Open a thread in T3 Code Desktop.
2. Open its browser preview.
3. While the agent is using the preview, focus the composer and begin typing.
4. Have the agent call preview_press, such as pressing Tab.
5. Observe the composer focus.

Cross-thread case:

1. Let the agent continue browser automation in thread A.
2. Switch to thread B.
3. Focus thread B’s composer and begin typing.
4. When thread A calls preview_press, observe thread B’s composer focus.

### Expected behavior

Browser automation should not interrupt typing in the composer. This should
remain true whether the automation belongs to the visible thread or a background
thread.

### Actual behavior

When the agent calls preview_press, the composer loses focus and typing stops
until it is focused again. Automation from a background thread can steal focus
from the composer in the currently visible thread.

The desktop press implementation focuses the preview guest WebContents and
later restores the outer renderer, but it does not restore the renderer’s active
DOM element. Preview automation is hosted per environment rather than per
visible thread, which explains the cross-thread behavior.

Related but not duplicate to #3715: that issue concerns unreliable delivery of
agent key presses to the preview. This issue concerns the resulting loss of
human composer focus, including across threads. 

### Impact

Major degradation or frequent failure

### Version or commit

main @ ba9c9ae81

### Environment

T3 Code Desktop OS: Arch Linux Provider: Codex

### Logs or stack traces

```shell

```

### Workaround

Click the composer again after browser automation takes focus. Avoid
preview_press while composing when possible.

AntoineArt:
Still hitting this on desktop **0.0.35-nightly.20260826.1193** (Arch Linux, AppImage) while typing in the composer.

Every agent preview interaction (not just `preview_press`) steals focus from the composer. Especially painful when writing a follow-up while the agent is clicking through a running browser in the same or another thread.

Workaround remains: click the composer again after each automation step.

Related open PRs that look like they try to fix this but are unmerged: #4118, #5874, #5894, #5898.

AchillesKastanas:
Confirmed here as well, Arch Linux, happens almost constantly

juliusmarminge:
[...]

PR #9849, opened 2026-09-05
fix(web): restore renderer focus after preview presses

0bkevin:
Preview key presses can leave the host renderer with a connected preview control focused after native focus returns, so the next composer keystroke is lost. This keeps the existing void IPC contract and restores the prior DOM focus only when the document is still focused, the element remains connected, and no newer user focus took ownership.

The change is limited to the web renderer:

- Wrap the existing preview press bridge call with a DOM focus guard.
- Recognize native focus transfer separately from newer pointer, keyboard, or programmatic DOM focus.
- Keep overlapping calls concurrent while preventing an older call from reclaiming focus.
- Preserve the operation result or rejection when focus restoration fails.

`PreviewAutomationHosts` is mounted only in Electron (`isElectron && previewBridge.automation`), so the standalone browser path is unaffected. This is the renderer portion of #9530 and is independent of the desktop native companion PR.

Validation:

- 10 focused renderer focus tests passed.
- `@t3tools/web` typecheck passed.
- Targeted lint passed with the existing warning at `PreviewAutomationHosts.tsx:741`.
- Modeled baseline test failed before the guard and passed with it.
- No live Electron before/after recording is included yet; the draft remains pending that runtime evidence.
- During overlaps, the latest-call guard conservatively suppresses stale restoration; this does not claim to resolve every cross-thread focus race.

Refs #5792. Replaces the renderer portion of #9530. Desktop companion: [#9848](https://github.com/pingdotgg/t3code/pull/9848).

Model: GPT-5.6 Luna implementation; GPT-6 review
Harness: Codex in T3 Code

### DIFF
--- a/apps/web/src/components/preview/PreviewAutomationHosts.tsx
+++ b/apps/web/src/components/preview/PreviewAutomationHosts.tsx
@@ -45,6 +45,7 @@ import { browserDefaultOpenViewport, resolveBrowserDefaults } from "~/browser/br
 import { runBrowserViewportMutation } from "~/browser/browserViewportActions";
 import { previewRuntimeTabId } from "~/browser/previewRuntimeTabId";
 import { isElectron } from "~/env";
+import { withPreviewAutomationFocus } from "~/lib/previewAutomationFocus";
 import { useEnvironments } from "~/state/environments";
 import { previewEnvironment } from "~/state/preview";
 import { useAtomQueryRunner } from "~/state/use-atom-query-runner";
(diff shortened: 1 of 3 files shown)


juliusmarminge:
Superseded by #11354 (fix(desktop): keep preview keystrokes out of the composer), which landed the desktop-side fix for preview automation leaking into the composer (Refs #5792). Closing this renderer focus-restore companion as wash leftover hygiene — reopen if anything unique still needs merging.

Issue #5685, opened 2026-08-08
[Bug]: Revert leaves selected user prompt visible and composer empty


LunarRed:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/web

### Steps to reproduce

1. Open a thread in a Git worktree and complete at least two user and assistant turns.
2. Hover over the latest completed user prompt.
3. Click **Revert to this message**.
4. Confirm the rollback.
5. Observe the timeline and composer after T3 reports that the revert completed.

### Expected behavior

The selected user prompt and its following assistant turn should be removed from the timeline. The preceding assistant response should become the final visible message.

If the user-message placement is intended to support an edit-and-retry workflow, the selected prompt should also be restored to the composer as editable text. At minimum, it must not remain in the timeline as an immutable message after T3 has reverted to the checkpoint before that turn.

### Actual behavior

The assistant work after the selected prompt is removed, but the selected user prompt remains visible as a locked timeline message. The composer is empty, so the prompt cannot be edited or resent. The user must enter another prompt to continue.

This leaves the live timeline inconsistent with the completed rollback. The screenshot shows the selected prompt still visible directly above an empty composer.

Current `main` also suggests a client-state mismatch: the server projection removes messages according to the requested checkpoint count, while the live client retains messages without a turn ID. User prompts are recorded without a turn ID.

Related but distinct: #2268 reports the entire thread disappearing and reverted text appearing in a new-chat draft. This report is specifically about the selected prompt remaining in the same thread as immutable history.

### Impact

Major degradation or frequent failure

### Version or commit

T3 Code desktop app 0.0.32

### Environment

macOS, T3 Code desktop app 0.0.32, Claude Opus 5, Medium reasoning, 1M context, Auto runtime, worktree thread

### Logs or stack traces

```shell

```

### Screenshots, recordings, or supporting files

[file-9635947fe1b307225cfa7dfb15631b2e.jpg](https://github.com/user-attachments/assets/33e7f471-fd40-4701-98b7-391c6292e99f)

### Workaround

No complete workaround found. Sending a separate new prompt continues the thread, but the reverted prompt remains uneditable and cannot be corrected or resent.

amitray007:
Thanks for the detailed report, I have noticed the bug to be super annoying and wanted to improve it.

I’ve implemented a fix in #6044.
[...]

PR #9857, opened 2026-09-05
fix(client): remove reverted prompts from live history

juliusmarminge:
After a successful revert, the live client keeps user prompts without a turn ID even when the server has removed them. Reloading the thread produces different history.

**Human-held:** the ordinary two-turn case is fixed, but additional receipt tests confirm a pre-existing server/client disagreement for steering messages and turnless assistant commentary. This candidate preserves those earlier rows in the live client while the server drops them. Choosing preservation versus matching the server's removal needs a maintainer decision; this PR is not ready for automatic merge.

Bound those prompts by the latest retained checkpoint's assistant message, falling back to its completion time when that message is outside the loaded window. Preserve messages bound to retained turns and system messages. This carries forward only Amit Ray's retention correction and regression tests from closed [#6044](https://github.com/pingdotgg/t3code/pull/6044), with an additional no-surviving-checkpoint control. The author [invited reuse](https://github.com/pingdotgg/t3code/pull/6044#issuecomment-5372126063), and maintainer edits on that closed branch are unavailable.

Addresses the live-history portion of [#5685](https://github.com/pingdotgg/t3code/issues/5685). Restoring prompts, images or existing drafts into the composer is not included; the issue remains open for those recovery choices.

### Before and after

A disposable SQLite orchestration engine generated two turns and checkpoint records at [82f64cd8](https://github.com/pingdotgg/t3code/commit/82f64cd8d046ab4ee8588f7bcbc1e66a4a0c80fe). After its committed revert-completion receipt:

| History | Before | After |
| --- | --- | --- |
| Persisted server snapshot | Prompt 1, Answer 1 | Prompt 1, Answer 1 |
| Actual shared client reducer, given that exact persisted event | Prompt 1, Answer 1, **Prompt 2** | Prompt 1, Answer 1 |

The receipt fixture waits on engine receipts and worker drains, without sleeps. This verifies post-revert history projection, not a Git/provider rewind or browser interaction. No UI component, provider behavior, schema, dependency or stored-data migration changes.

Additional receipt-derived controls use real persisted snapshots, not reducer-only expected arrays:

| Case | Result at this candidate |
| --- | --- |
| Ordinary two-turn history | Live and persisted histories match |
| Missing checkpoint assistant ID, completion-time fallback | Match |
| Loaded client window starts after the retained checkpoint | Match for the loaded message IDs |
| Same-turn steering before the retained checkpoint | Live keeps the steering prompt; persisted history drops it |
| Turnless assistant commentary before the retained checkpoint | Live keeps the commentary; persisted history drops it |

The last two are unresolved history-policy limits, not claims of complete parity. Server retention and contracts remain unchanged.

### Checks

- 49 focused shared-client reducer and pagination tests pass. The original, paginated-window and steering-history regressions failed on main before the correction.
- Checkpoint-zero and no-retained-checkpoint controls preserve system messages and remove turnless prompts.
- Client-runtime typecheck, targeted lint and diff checks pass.
- Web, desktop and mobile consume this shared reducer. No native-device run is claimed.

The earlier CI attempt was infrastructure-blocked by Ubuntu HTTP mirror timeouts. This branch is now refreshed onto [ce4712d5](https://github.com/pingdotgg/t3code/commit/ce4712d5b04fb998f79fe132245289191147e5d5), which includes the upstream HTTPS mirror repair. All 49 focused tests, client-runtime typecheck and targeted lint pass again. A new current-head CI cycle is running. No extra workflow change or check bypass was made. The human history-policy hold above remains in place.

Original correction by [Amit Ray](https://github.com/amitray007). Prepared and verified with GPT 6 Astra via Codex in T3 Code.

---

> [!NOTE]
> **Medium Risk**
> Changes shared revert projection logic used by web/desktop/mobile; wrong bounds could drop or keep the wrong messages in chat history, though scope is limited to `thread.reverted` handling and is heavily regression-tested.
> 
> **Overview**
> Fixes **live thread history** after a `thread.reverted` event so reverted user prompts (and related rows) disappear without a full reload.
> 
> **`retainMessagesAfterRevert`** no longer keeps every message with a null `turnId`. Turn-bound messages still follow retained checkpoints; **system** messages stay. **Turnless** user/assistant rows are kept only through the **latest retained checkpoint**—by index through that checkpoint’s `assistantMessageId` when it’s in the loaded window, otherwise by `createdAt` vs the checkpoint’s `completedAt`.
> 
> The existing two-turn revert test is updated (explicit second prompt + assistant id), with new reducer tests for paginated windows, no surviving checkpoint, and pre-checkpoint steering/commentary retention.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 576461e11220e74494d443954a5c2b29b101ff53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client-runtime/src/state/threadReducer.test.ts
+++ b/packages/client-runtime/src/state/threadReducer.test.ts
@@ -1183,6 +1183,15 @@ describe("applyThreadDetailEvent", () => {
           },
           {
             id: MessageId.make("msg-3"),
+            role: "user",
+            text: "Second",
+            turnId: null,
+            streaming: false,
+            createdAt: "2026-04-01T02:30:00.000Z",
+            updatedAt: "2026-04-01T02:30:00.000Z",
+          },
+          {
+            id: MessageId.make("msg-4"),
             role: "assistant",
             text: "Response 2",
             turnId: TurnId.make("turn-2"),
@@ -1207,7 +1216,7 @@ describe("applyThreadDetailEvent", () => {
             checkpointRef: CheckpointRef.make("ref-2"),
             status: "ready",
             files: [],
-            assistantMessageId: MessageId.make("msg-3"),
+            assistantMessageId: MessageId.make("msg-4"),
             completedAt: "2026-04-01T03:00:00.000Z",
           },
         ],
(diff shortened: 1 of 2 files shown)


shivamhwp:

Current main already removes the ordinary second prompt through #5362, but the loaded-pagination case still needs a fix. Rebasing this retention rule must preserve main's imported-history exception: at checkpoint zero it currently removes both imported prompts and imported answers, while the persisted server history keeps them. Retain `isImportedAgentSessionMessageId` rows alongside system messages.

The completion-time fallback also needs `compareDateTimeStrings`, rather than lexical `<=`. For a 09:00Z checkpoint, a message at 10:30+02:00 belongs before the cutoff but is dropped, while 07:30-02:00 belongs after it but is kept.

The existing history-policy hold still applies. There is also a current fallback disagreement when no checkpoint survives in the loaded client state and turnCount is positive: this candidate drops all turnless rows, while main retains fallback messages. Resolve that alongside the steering/commentary policy before treating live and persisted history as consistent.

PR #9896, opened 2026-09-05
fix(server): preserve V2 checkpoint baselines across runs

saphid:
Full-thread diffs lose their baseline after a second V2 run because the reused checkpoint scope is no longer owned by the first run. Resolve ordinal zero from the target checkpoint scope and keep nonzero ranges within that same scope.

Rebuilt head `6cf02a815` on V2 `3aa47e294` with test-only prerequisite #10865 (merge first). Two-file topic; regression uses the real ID allocator and projector across two runs.

Validation: all six focused checkpoint query tests pass, plus server typecheck, targeted lint and formatting. New CI pending. Direct Claude Opus 5 high review remains unavailable: latest launch exited 1 on expired OAuth before inference.

Targets V2 #2829. Updated with GPT-6 in Codex/T3.


juliusmarminge:
Closing as superseded by c96c4bd838 on v2. CheckpointDiffQuery now uses the persistent root scope rather than matching its latest owner to the first run. The root scope ID is deterministic per thread (IdAllocator.checkpointScope / CheckpointService), and checkpoint capture requires that root scope, so the additional scope filter does not add behavior on current v2. The checkpoint scope ownership and diff query tests pass. Independently confirmed by Claude Fable 5.1.

PR #9900, opened 2026-09-05
fix(server): reconcile historical V2 migration manifests

saphid:
Historical V2 builds recorded the same setup under different migration numbers. After V2 consolidated its schema into migration 050, those ledgers can skip missing schema or replay DDL against existing tables.

Head `9e3929026` integrates V2 `fdfadf3d0`, validates the historical manifest, fills missing main migrations, and runs only the remaining V2 setup steps before consolidating the ledger. Existing import progress and the original migration-050 timestamp are preserved. Main-only migration ceilings leave partial V2 progress intact for the next full upgrade. All reconciliation remains transactional.

Includes the test-only prerequisite #10865; merge it first.

Verification: 20 focused tests across four migration suites pass, covering historical bases 43/44/47/49, partial and complete V2 sequences, rollback, timestamp/import preservation, and the consolidated schema. Server typecheck and scoped lint pass. A standalone server export scan reports pre-existing unused exports across V2, so it is not recorded as a passing check. Direct independent Claude review remains unavailable because the launcher exits 1 on expired OAuth before inference.

Implemented and verified with GPT-6 Astra in the Codex harness.


juliusmarminge:
i don't want to do this because it sets expectations that everyone can just go into any pr, try it out, get some migration running in the database, and then expect it to work on latest. it prevents us from fucking around on probably feature branches. i think feature branches should be seen as temporary and if you try it out and it happens to break later, you're going to have to ask your agents to reconcile your database manually for you. i don't feel it's worth adding the extra complexity needed to always make sure that you, paying users, have a good path from trying feature branches to going back to stable. i don't

PR #9901, opened 2026-09-05
fix(mobile): respect notification permission when linking cloud environments

### LINKED ISSUE #9886: Cloud link hardcodes notificationsEnabled: true, contradicting the privacy policy's "if you enable notifications"
### Summary

The privacy policy says agent-activity data is processed **"if you enable notifications or Live Activities."** In the cloud-link flow, `notificationsEnabled` is a hardcoded `true` — the user never enables it. The `liveActivitiesEnabled` flag sitting immediately beside it *does* respect a user preference, which is what makes this look unintended rather than deliberate.

### Steps to reproduce

`apps/mobile/src/features/cloud/linkEnvironment.ts`, in `linkEnvironmentToCloudWithPreference` — both relay calls, `:272-278` and `:303-311`:

```ts
const liveActivitiesEnabled = input.liveActivitiesEnabled;   // ← from preferences
const challenge = yield* relayClient
  .createEnvironmentLinkChallenge({
    clerkToken: input.clerkToken,
    payload: {
      notificationsEnabled: true,                            // ← hardcoded
      liveActivitiesEnabled,
      managedTunnelsEnabled: true,
    },
  })
```

```ts
const link = yield* relayClient
  .linkEnvironment({
    clerkToken: input.clerkToken,
    payload: {
      deviceId,
      proof,
      notificationsEnabled: true,                            // ← hardcoded
      liveActivitiesEnabled,
      managedTunnelsEnabled: true,
    },
  })
```

The function's own signature takes `{ readonly liveActivitiesEnabled: boolean }` and `:353` reads `preferences.liveActivitiesEnabled !== false`. There is no equivalent for notifications, and no `notificationsEnabled` reference anywhere in `apps/mobile/src/**/*.tsx` — no UI writes it.

The other code path already does this correctly. `apps/mobile/src/features/agent-awareness/remoteRegistration.ts:255-271` *derives* the flag:

```ts
if (!canRegisterRemoteLiveActivities() || !supportsAgentAwarenessPush()) {
  return { notificationsEnabled: false, pushToken: null };
}
...
if (!permissions.granted) {
  return { notificationsEnabled: false, pushToken: null };
}
```

So the same field is computed from real permission state in one place and asserted as `true` in another.

### Expected behavior

`notificationsEnabled` reflects a user choice, the way `liveActivitiesEnabled` does beside it — matching the privacy policy's "if you enable."

### Actual behavior

It is `true` for every cloud link, unconditionally.

### What flows as a result

`infra/relay/src/agentActivity/` publishes and persists per-thread activity. `AgentActivityPublisher.ts:211-212` carries `projectTitle` and `threadTitle`; `ApnsDeliveries.ts:288` builds the push as `{ title: row.threadTitle, body: "<status>: <projectTitle>" }`; `agentActivityPayloads.ts:71-72` truncates them rather than removing them. They are stored server-side in `relay_agent_activity_rows.state_json` (`infra/relay/src/persistence/schema.ts:135-149`), not just passed through.

### Credit where it is due

The privacy policy is genuinely good here and names these fields precisely — `apps/marketing/src/pages/privacy-policy.astro:78-83`:

> **Agent activity and notifications.** If you enable notifications or Live Activities, we process push tokens and limited activity information needed to provide those features, such as project and thread titles, model name, status, and activity headline.

That is an accurate description of exactly what `agentActivityPayloads.ts` sends. This issue is only that the leading conditional is not true of the link flow: the disclosure is right, the gate is missing.

### Impact

Thread and project titles routinely contain the substance of the work — client names, ticket IDs, unreleased feature names. iOS still gates *delivery* behind its own push-permission prompt, so nothing appears on a lock screen without consent. But the flag sent at link time is what turns on server-side collection and retention of the titles, and that happens before, and independently of, any prompt the user sees.

### Suggested fix

Thread a `notificationsEnabled` preference through `linkEnvironmentToCloudWithPreference` the way `liveActivitiesEnabled` already is, defaulting to whatever you intend — and if the intent really is on-by-default, the privacy policy's "if you enable notifications" should say so instead.

### Version

`d7cf8aaa` (main).

Gigioxx:
Mobile cloud linking sends `notificationsEnabled: true` even when notification permission is denied. Read the existing OS permission once and use it for both the link challenge and link request. Unsupported platforms and personal-team builds keep notifications disabled; permission lookup failures stop before contacting the relay.

Live Activities keep their existing preference. This applies when linking or relinking an environment; existing cloud links are not migrated.

Closes #9886.

## Verification

- Reproduced before the fix: the denied-permission regression received `notificationsEnabled: true` with Live Activities disabled.
- 61 focused tests pass across cloud linking, Live Activity preferences, and remote registration. Covers granted/denied permission, unsupported push, permission lookup failure, and both relay payloads.
- Mobile typecheck and changed-file lint/format checks pass.
- Native device verification was not run. No UI or wire-contract changes.

Created with GPT-6 in the Codex harness.

---

> [!NOTE]
> **Medium Risk**
> Changes relay registration flags during cloud link/relink based on OS permission and capability checks; failures block linking but do not alter existing linked environments.
> 
> **Overview**
> Cloud environment linking no longer always sends **`notificationsEnabled: true`** to the relay. During link/relink, the client reads **`expo-notifications`** permission once (iOS only when agent-awareness push is supported) and uses that value for both the link-challenge and link payloads; **Live Activities** still follow the existing preference.
> 
> **Android**, builds without push support, and **denied** permission keep notifications disabled. If permission lookup fails, linking fails with *"Could not read notification permissions."* and **does not** call the relay. Tests were expanded with parameterized cases and a failure-path assertion.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9131579c4672e80d9ef2c5500da010351b679e1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add tests for notification permission checks when linking cloud environments
> - Adds parameterized tests covering four cases: denied iOS permission, granted iOS permission, Android, and iOS without push support. Each verifies notification and Live Activity flags propagate correctly through challenge and link requests.
> - Adds a test that a notification permission lookup failure aborts linking before any relay request and returns the expected cloud-link error.
> - Updates the explicit Live Activity preference test to confirm that notifications are disabled while Live Activity remains enabled.
> - Behavioral Change: the test suite now asserts notifications are only enabled for the granted-iOS-with-push-support case; all other combinations disable notifications.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9131579.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/features/cloud/linkEnvironment.test.ts
+++ b/apps/mobile/src/features/cloud/linkEnvironment.test.ts
@@ -21,6 +21,15 @@ import {
   refreshCloudEnvironmentConnection,
 } from "./linkEnvironment";
 
+const { getPermissionsAsync, platform, supportsAgentAwarenessPush } = vi.hoisted(() => ({
+  platform: { OS: "ios" },
+  supportsAgentAwarenessPush: vi.fn(() => true),
+  getPermissionsAsync: vi.fn(async () => ({ granted: false })),
+}));
+
+vi.mock("expo-notifications", () => ({ getPermissionsAsync }));
+vi.mock("../agent-awareness/capabilities", () => ({ supportsAgentAwarenessPush }));
+
 vi.mock("expo-constants", () => ({
   default: {
     expoConfig: {
@@ -47,9 +56,7 @@ vi.mock("expo-device", () => ({
 }));
 
 vi.mock("react-native", () => ({
-  Platform: {
-    OS: "ios",
-  },
+  Platform: platform,
 }));
 
 vi.mock("expo-secure-store", () => ({
@@ -191,6 +198,9 @@ function listedEnvironment(environmentId: string) {
 describe("mobile cloud link environment client", () => {
   beforeEach(() => {
     vi.restoreAllMocks();
+    getPermissionsAsync.mockReset();
+    supportsAgentAwarenessPush.mockReset();
+    platform.OS = "ios";
     createProofMock.mockClear();
     loadPreferences.mockClear();
   });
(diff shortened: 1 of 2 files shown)


shivamhwp:

Please update the iOS-only gate for current main. #10416 added Android agent notifications, and `supportsAgentAwarenessPush()` now checks the Android native capability too. With Android push supported and notification permission granted, this PR still sends `notificationsEnabled: false` in both cloud-link requests.

Use the shared capability check for supported platforms and read permission for Android as well. Change the Android regression to expect true when granted, and retain false cases for denied permission and unsupported builds.

PR #9907, opened 2026-09-05
test(orchestration): cover bounded V2 socket fallback paging

saphid:
Add regression coverage for bounded thread history over the WebSocket fallback: contract decoding, server snapshot paging, client resume metadata, and loading earlier history after HTTP snapshot loading fails.

Rebased the original test commit onto current v2. No production behavior changes. All 88 tests in the four affected files pass; scoped formatting and lint pass. CI and Claude Fable 5.1 review are checked before merge.

Prepared with Codex/T3; independent review requested from Claude Fable 5.1.

### DIFF
--- a/apps/server/src/orchestration-v2/threadHistoryPaging.test.ts
+++ b/apps/server/src/orchestration-v2/threadHistoryPaging.test.ts
@@ -24,6 +24,7 @@ import {
   THREAD_HISTORY_CURSOR_MAX_LENGTH,
   THREAD_HISTORY_PAGE_POLICY,
 } from "./threadHistoryPaging.ts";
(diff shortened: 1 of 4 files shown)


saphid:
Ran an independent automated first-pass review (Amp/Astra) over the 8-file diff against 415ed0f7.

- 1 finding: bounded socket snapshots could silently degrade older V2 clients (no capability negotiation). Dismissed as false positive at high severity: the new fields are optionalKey and excess keys are dropped on decode, so legacy clients don't break; V2 is unreleased (targets #2829) so no shipped old-client population, and mixed nightlies already surface a version-mismatch warning. Residual: mixed-version fallback shows only the window with no load-earlier until versions sync. A negotiation/bump feels disproportionate for an unreleased branch.
- No other findings; no code changes pushed.

PR #10025, opened 2026-09-05
fix(server): keep the native Grok default model

saphid:
Treat Grok's product slug as “keep the native session model,” and keep model selection current when switching away and back within one runtime.

Rebased the original two commits onto v2. Adapted model selection to use ACP v1 session/set_model or ACP v2 model configuration according to the negotiated protocol. Preserved the current generic ACP option handling.

Validation: default selection and switching-back regression cases fail on the base and pass with this change. All 124 ACP adapter/Grok support tests pass, alongside server typechecking and scoped formatting/lint. CI and Claude Fable 5.1 review are checked before merge.

Prepared with Codex/T3; independent review requested from Claude Fable 5.1.

### DIFF
--- a/apps/server/src/orchestration-v2/Adapters/AcpAdapterV2.test.ts
+++ b/apps/server/src/orchestration-v2/Adapters/AcpAdapterV2.test.ts
@@ -12,6 +12,7 @@ import * as NodeServices from "@effect/platform-node/NodeServices";
 import { assert, describe, it } from "@effect/vitest";
 import {
   CheckpointId,
+  GrokSettings,
   EnvironmentId,
   MessageId,
   type ModelSelection,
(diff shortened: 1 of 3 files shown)


saphid:
Ran an independent automated first-pass review (Amp/Astra, read-only) on `fecd670438`: no correctness, Effect-convention, or performance defects in the PR's own change.

One real finding surfaced, pre-existing but on the path this PR codifies: same-session model switches compare against stale session-setup metadata, so A → B → A skipped the final `session/set_model` and silently left the session on B. Addressed in 0e4a3153d4 (`configureSession` now persists the applied model into `activeSessionSetup`), with a Grok regression test that fails without the fix. Focused suites pass (145 tests), typecheck clean.

PR #10053, opened 2026-09-05
fix(server): preserve inline sensitive env values on save

### LINKED ISSUE #10022: [Bug]: Sensitive provider env value stored inline in settings.json is wiped on first save
### Area

apps/server

### Steps to reproduce

1. Stop the server. In `settings.json` under the T3 userdata directory, add a provider instance with a sensitive variable whose value sits inline, and make sure no matching `provider-env-*` file exists under `secrets/`:

   ```json
   {
     "providerInstances": {
       "claudex": {
         "driver": "claudeAgent",
         "enabled": true,
         "environment": [
           { "name": "ANTHROPIC_BASE_URL", "value": "http://localhost:8317", "sensitive": false },
           { "name": "ANTHROPIC_AUTH_TOKEN", "value": "sk-inline-token", "sensitive": true }
         ],
         "config": {}
       }
     }
   }
   ```

2. Start the server. The provider works, because load reads the inline value.
3. From any client, save that provider instance without touching the token. Toggling the enabled switch is enough. The cross-device settings sync triggers the same save.
4. Read `settings.json` again, then start a new thread on that provider.

### Expected behavior

The token keeps its value. Either it stays in the file or the save moves it into the secret store. A save that never touched a field should not change what that field resolves to.

### Actual behavior

The entry becomes `"value": "", "valueRedacted": true` and no `provider-env-*` secret file exists. Every provider session started after that gets an empty token. Sessions already running keep their old environment, so the failures look random.

The cause is in `apps/server/src/serverSettings.ts`. Clients only ever receive the redacted form of a sensitive variable. On save, the redacted branch of `persistProviderEnvironmentSecrets` writes that placeholder straight back. It never checks that the secret store holds a value, and it drops the inline value the server still has in memory for that variable. One save and the token is gone from disk.

This is a papercut more than a bug. The documented path is to enter secrets through Settings, and that path works. But writing `settings.json` directly is a normal thing for a coding agent to do when it sets up a provider, which is how my config got this way. I never edited the file by hand. The app accepted the config, ran with it for weeks, then destroyed it on an unrelated save with no warning.

### Impact

Minor bug or occasional failure. It only affects configs written outside the UI, but when it hits, the failure is silent and shows up later as a provider auth error on whichever thread starts next, with nothing pointing back at the save.

### Version or commit

0.0.39-nightly.20260903.1270. Reproduced with a server test against main at c7dc3cbd0.

### Environment

Linux (WSL2), Node 24.11.1, Claude Code provider behind a local CLIProxyAPI.

### Logs or stack traces

```shell
# proxy response for every session started after the save
401 {"error":"Invalid API key"}

# server-level repro (temporary test, not committed)
AssertionError: expected '' to equal 'sk-inline-token'
```

### Related issues

None match. #9230, #317, and #1267 are about variables not reaching the provider at all, not about a save erasing a stored value. Closed PRs #3524 and #3061 touched the settings editor's draft rows in the browser and never changed what the server writes.

### Workaround

Re-enter the value in Settings > Providers for that instance. That path writes it to the secret store and the placeholder in `settings.json` becomes valid.

UzEE:
## What Changed

Saving a provider with a sensitive environment value stored inline could replace that value with an empty redacted placeholder. Preserve the current inline value through the existing secret writer before saving the placeholder.

Adds one regression test for the redacted read/save round trip.

## Why

Clients only receive the redacted value, so saving an unchanged credential must preserve the server's current value. This fixes the data loss without changing the settings schema or clients.

Fixes #10022.

## Validation

- `vp test run apps/server/src/serverSettings.test.ts`: 34 tests passed. The new regression test failed before the fix.
- `vp exec tsgo --noEmit -p apps/server/tsconfig.json`: passed.
- `vp lint apps/server/src/serverSettings.ts apps/server/src/serverSettings.test.ts`: passed.

## Checklist

- UI screenshots and video are not applicable.

> [!NOTE]
> ### Preserve inline sensitive env values on save in server settings
> Fixes the server settings update flow so that redacted inline sensitive environment values are not overwritten with empty strings when the client submits a redacted copy. The update service now retains the existing secret on save while the settings file stays redacted. Added an integration test in [serverSettings.test.ts](https://github.com/pingdotgg/t3code/pull/10053/files#diff-511d9a6e2ded44cb61547a93c6a9e5b08e51a068ea0c2ccc0fba26e09a96810f) that writes a provider instance with an inline secret, submits the redacted form, and asserts the secret is preserved in memory but absent from the file.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8d69162.</sup>
> 
>

### DIFF
--- a/apps/server/src/serverSettings.test.ts
+++ b/apps/server/src/serverSettings.test.ts
@@ -1039,6 +1039,37 @@ it.layer(NodeServices.layer)("server settings", (it) => {
     }).pipe(Effect.provide(makeServerSettingsLayer())),
   );
 
+  it.effect("preserves inline sensitive provider environment values on a redacted save", () =>
+    Effect.gen(function* () {
+      const serverSettings = yield* ServerSettingsModule.ServerSettingsService;
+      const serverConfig = yield* ServerConfig.ServerConfig;
+      const fileSystem = yield* FileSystem.FileSystem;
+      const instanceId = ProviderInstanceId.make("codex_personal");
+      yield* fileSystem.writeFileString(
+        serverConfig.settingsPath,
+        '{"providerInstances":{"codex_personal":{"driver":"codex","environment":[{"name":"OPENROUTER_API_KEY","value":"inline-secret","sensitive":true}],"config":{}}}}',
+      );
+      const clientSettings = ServerSettingsModule.redactServerSettingsForClient(
+        yield* serverSettings.getSettings,
+      );
+      assert.equal(clientSettings.providerInstances[instanceId]?.environment?.[0]?.value, "");
+
+      const saved = yield* serverSettings.updateSettings({
+        providerInstances: clientSettings.providerInstances,
+      });
+
+      assert.equal(saved.providerInstances[instanceId]?.environment?.[0]?.value, "inline-secret");
+      assert.notInclude(
+        yield* fileSystem.readFileString(serverConfig.settingsPath),
+        "inline-secret",
+      );
+      assert.equal(
+        (yield* serverSettings.getSettings).providerInstances[instanceId]?.environment?.[0]?.value,
+        "inline-secret",
+      );
+    }).pipe(Effect.provide(makeServerSettingsLayer())),
+  );
+
   it.effect("stores sensitive provider instance environment values outside settings.json", () =>
     Effect.gen(function* () {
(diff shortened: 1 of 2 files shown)


juliusmarminge:
Superseded by #10054, which already merged and closed #10022. Closing this as leftover.

PR #10054, opened 2026-09-05
fix(server): preserve inline provider secrets on redacted saves

### LINKED ISSUE #10022: [Bug]: Sensitive provider env value stored inline in settings.json is wiped on first save

maxwellyoung:
Saving an unchanged provider can erase a sensitive environment value stored inline in settings.json. Clients return a redacted placeholder, but there may not yet be a corresponding secret-store entry.

Before persisting that placeholder, migrate the current inline value for the same provider instance and variable into the secret store. Explicit replacements and clears keep their existing behavior.

Fixes #10022.

Validation: the regression fails on main and passes with the fix. All 38 server-settings tests pass, including duplicate names using the existing last-value-wins behavior, fresh-service reload, explicit replacement, clearing, and preserving the original file and cached value if the secret-store write fails. Targeted lint and server typechecking pass.

> [!NOTE]
> ### Preserve inline provider secrets on redacted saves in `serverSettings`
> - Sensitive environment-variable updates with redacted values now look up the last existing non-empty inline value and migrate it to the secret store, instead of treating the redacted entry as empty.
> - Added tests covering secret migration failure (file unchanged, inline value retained) and parameterized cases for redacted saves, duplicate env-var names, explicit replacements, and explicit empty values.
> - Risk: if multiple existing entries share the same env-var name, only the last one is used; callers relying on the old empty-on-redacted behavior will see different persisted values.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5b1c0f1.</sup>
> 
>

### DIFF
--- a/apps/server/src/serverSettings.test.ts
+++ b/apps/server/src/serverSettings.test.ts
@@ -1039,6 +1039,120 @@ it.layer(NodeServices.layer)("server settings", (it) => {
     }).pipe(Effect.provide(makeServerSettingsLayer())),
   );
 
+  it.effect("keeps the inline value on disk when secret migration fails", () => {
+    const cause = new ServerSecretStore.SecretStorePersistError({
+      resource: "provider environment secret",
+      cause: new Error("Secret storage unavailable"),
+    });
+    const secretLayer = Layer.effect(
+      ServerSecretStore.ServerSecretStore,
+      Effect.map(ServerSecretStore.ServerSecretStore, (store) => ({
+        ...store,
+        set: () => Effect.fail(cause),
+      })),
+    ).pipe(Layer.provide(ServerSecretStore.layer));
+    const settingsLayer = ServerSettingsModule.layer.pipe(
+      Layer.provide(secretLayer),
+      Layer.provideMerge(Layer.fresh(SqlitePersistenceMemory)),
+      Layer.provideMerge(
+        Layer.fresh(
+          ServerConfig.layerTest(process.cwd(), {
+            prefix: "t3code-inline-secret-failure-test-",
+          }),
+        ),
+      ),
+    );
+    return Effect.gen(function* () {
+      const instanceId = ProviderInstanceId.make("codex_personal");
+      const service = yield* ServerSettingsModule.ServerSettingsService;
+      const config = yield* ServerConfig.ServerConfig;
+      const fs = yield* FileSystem.FileSystem;
+      const original =
+        '{"providerInstances":{"codex_personal":{"driver":"codex","environment":[{"name":"API_TOKEN","value":"inline-test-token","sensitive":true}],"config":{}}}}';
+      yield* fs.writeFileString(config.settingsPath, original);
+      const error = yield* Effect.flip(
+        service.updateSettings({
+          providerInstances: {
+            [instanceId]: {
+              driver: ProviderDriverKind.make("codex"),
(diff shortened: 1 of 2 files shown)


Mnigos (approved):
Ran this locally. All 38 server-settings tests pass, and with `serverSettings.ts` reverted to main exactly three of them fail (the redacted-save preserve case, the duplicate-name case and the secret-store failure case), so they really pin the fix. Also tried something I was curious about: this test file passes unchanged against #10053's implementation too, so the two PRs do the same thing and the difference is coverage. This one additionally locks in duplicate names, explicit replace and clear, and the write failure leaving settings.json untouched, which is the part I'd want covered for a bug that silently wipes credentials.

PR #10107, opened 2026-09-05
refactor(mobile): make outbox ownership checks explicit

t3dotgg:
The outbox load API hid partial recovery behind a boolean. Cleanup and account changes had to interpret that result separately.

Storage now reports readable messages and whether ownership is complete. Cleanup, sign-out archives, account restore, and environment removal require complete ownership. Readable messages stay available. Failed account changes keep the load error as their cause. Draft flushing owns hydration before writes.

The saved formats and image writer are unchanged. The file-backed writer remains in [#9727](https://github.com/pingdotgg/t3code/pull/9727), held for a new native runtime on each platform with compatible embedded readers and storage guards. The source comment now states that condition.

## Verification

- 130 focused tests passed across outbox, composer drafts, removal, drain, and pending-task writes.
- Mobile `tsc --noEmit` passed.
- Changed-file lint passed with one existing extra-effect-dependency warning in the drain.
- Independent source review passed after adding incomplete-account-restore coverage.

No UI layout, native code, or runtime configuration changed. No device or browser check ran.

> [!NOTE]
> ### Replace outbox errors-array contract with explicit complete/incomplete hydration result
> - Replaces the `errors` array and boolean success return on `ThreadOutboxStorage.load` and the manager load operation with a discriminated `ThreadOutboxLoadResult` / `ThreadOutboxHydrationResult` union carrying `status` and a single error object on `incomplete`.
> - `expoThreadOutboxStorage.load` no longer rejects on unreadable records or directory failures; it returns recovered messages plus `incomplete` status and an error, retrying on the next call.
> - Composer draft workflows in `use-composer-drafts.ts` (`releaseUnusedComposerAttachmentFiles`, `archiveCloudComposerDrafts`, `restoreCloudComposerDrafts`) now inspect the structured outbox status and abort destructive or state-changing operations before running when hydration is `incomplete`, surfacing the hydration error as the thrown cause.
> - `useThreadOutboxDrain` and `clearEnvironment` switch from boolean/errors checks to the new `incomplete` status, triggering the unreadable-record alert and retry path and blocking environment clear respectively.
> - Risk: any out-of-tree `ThreadOutboxStorage` implementation must adopt the new `complete`/`incomplete` load contract; the former `errors` array is gone. Existing storage mocks across tests are updated to the new shape.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f003089.</sup>
> 
>

### DIFF
--- a/apps/mobile/src/state/pending-task-editor-writes.test.ts
+++ b/apps/mobile/src/state/pending-task-editor-writes.test.ts
@@ -19,7 +19,7 @@ vi.mock("./thread-outbox", async () => {
   harness.manager = createThreadOutboxManager({
     registry: appAtomRegistry,
     storage: {
-      load: async () => ({ messages: [], errors: [] }),
+      load: async () => ({ messages: [], status: "complete" }),
       write: async (message) => {
         const pending = harness.writeGates.shift();
         if (pending) {
--- a/apps/mobile/src/state/thread-outbox-manager.ts
+++ b/apps/mobile/src/state/thread-outbox-manager.ts
@@ -37,6 +37,10 @@ export interface ThreadOutboxManagerOptions {
   readonly warn?: (message: string, error: unknown) => void;
 }
 
+type ThreadOutboxHydrationResult =
+  | { readonly status: "complete" }
+  | { readonly status: "incomplete"; readonly error: ThreadOutboxManagerError };
+
 export function createThreadOutboxManager(options: ThreadOutboxManagerOptions) {
   const queuedMessagesByThreadKeyAtom = Atom.make<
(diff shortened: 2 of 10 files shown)


t3dotgg:
> [!NOTE]

Closing. On main, thread-outbox-manager load() already returns true only for a complete load, and every caller in use-composer-drafts.ts and use-thread-outbox-drain.ts already treats false as "do not clean up or change accounting". The storage directory read is all-or-nothing, so the new incomplete-with-recovered-messages path never carries messages. This is a type reshuffle with no behavior change on the mobile persistence path.

PR #10108, opened 2026-09-05
perf(server): avoid full thread reads on turn start

t3dotgg:
Starting a turn loaded every message, plan, and checkpoint to find one user prompt. That cost grew with the thread.

Use the thread shell and a targeted message query. An SQL existence check finds any other non-compaction user message, including newer queued messages. This keeps the existing title, branch-name, and compaction rules.

Stacks on [#10123](https://github.com/pingdotgg/t3code/pull/10123) to preserve its request lifecycle changes. This PR can merge independently of the pending-question query PR after that base lands.

Verified with 174 focused query, reactor, and provider-service tests, server typecheck, and changed-file lint and format checks. The read test uses 2,000 unrelated history rows and confirms one SQL statement without decoding those rows. The restack preserves the original query patch and adds its service stub to a new test fixture on main.

Created with GPT-6 Astra in Codex. Follow-up review and restack with Codex.

> [!NOTE]
> ### Replace full thread reads with targeted `getTurnStartMessage` query on turn start
> - Adds `getTurnStartMessage` to `ProjectionSnapshotQuery` to fetch a single message by thread ID and message ID without loading or decoding the full message history. The query also computes `hasOtherUserMessages` by excluding trimmed, case-insensitive `/compact` commands without attachments.
> - Updates `ProviderCommandReactor.processTurnStartRequested` to use this targeted query instead of full thread-detail resolution and in-memory scanning. Title generation and compaction eligibility now use the query's `hasOtherUserMessages` result.
> - Adds `encodeChatAttachments` helper and test coverage for single-message reads with 2,000 unrelated rows, eligibility edge cases, and unreadable prior history.
> - Updates all in-tree `ProjectionSnapshotQuery` test doubles with the new required method.
> - Risk: `/compact` eligibility now depends on the SQL-level `hasOtherUserMessages` computation in `ProjectionSnapshotQuery.getTurnStartMessageRow`; whitespace and case handling must match the previous in-memory logic exactly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 06260ef.</sup>
> 
>


t3dotgg:
> [!NOTE]
> 🤖 **GPT-6 Astra responding on behalf of Theo**

The eligibility rules are unchanged.

- For a normal prompt, the old count of one non-compaction user message means there is no other such message.
- For `/compact`, the old count of zero means there is no other non-compaction user message.

The SQL check includes newer queued messages and keeps the same whitespace, case, and attachment rules. The focused tests cover those cases. All 91 query/reactor tests and core CI pass. Human approval is still required.

PR #10109, opened 2026-09-05
fix(web): limit stored pull request details

t3dotgg:
Opening PRs kept adding full detail snapshots to localStorage with no limit. They could fill the browser's storage quota.

Keep at most 20 recent snapshots within 2 MiB, including keys and the recency index. Each read still decodes only the selected PR. Retire old detail-only keys and recover from partial writes without touching other stored data.

PR details can cold-load once after upgrading, then refill through live queries. There is no layout change. Missing, denied, or full storage still falls back to live loading.

Verified:

- 108 tests in `pullRequestDetail.logic.test.ts` pass, including eviction, byte limits, scope, legacy cleanup, and storage failures.
- Targeted lint passes for both changed files.
- Web typecheck and `git diff --check` pass.

> [!NOTE]
> ### Limit stored pull request details with a bounded, indexed LRU cache
> - Replaces legacy delimiter-concatenated snapshot keys with a v1 prefix plus a JSON-serialized array of environment, project, repository, and pull-request number components, so scope parts with separator characters no longer collide.
> - Introduces a recency index and a `detailSnapshotsFit` validator that enforce a max of 20 entries and 2 MiB of accounted UTF-16 storage across all cached pull-request detail payloads.
> - Reworks `writePullRequestDetailSnapshot` to evict oldest entries until the cache fits, remove legacy and unindexed v1 keys, remeasure retained payloads, and roll back a payload write when its index update fails.
> - `readPullRequestDetailSnapshot` now returns a snapshot only when it is indexed, decodable, and size-verified; reads move the entry to the most-recent position without discarding a valid snapshot on recency-update failure.
> - `readDetailSnapshotIndex` treats missing, malformed, over-limit, or error cases as an empty index without deleting unrelated storage.
> - Risk: existing stored legacy keys under the old `pullRequestDetailSnapshotKey` format are ignored and removed on the next write; users will see cold cache misses until snapshots are repopulated under the v1 schema.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b6a1ad7.</sup>
> 
>

### DIFF
--- a/apps/web/src/components/pullRequest/pullRequestDetail.logic.test.ts
+++ b/apps/web/src/components/pullRequest/pullRequestDetail.logic.test.ts
@@ -1,4 +1,5 @@
 import {
+  ProjectId,
   PullRequestAction,
   type PullRequestCheck,
   type PullRequestComment,
@@ -31,6 +32,8 @@ import {
   pullRequestFindingKey,
   pullRequestHandoffLabels,
   pullRequestReviewOutcome,
+  PULL_REQUEST_DETAIL_SNAPSHOT_MAX_BYTES,
+  PULL_REQUEST_DETAIL_SNAPSHOT_MAX_ENTRIES,
   readableFailure,
   readPullRequestDetailSnapshot,
   resolveDisplayedPullRequestDetail,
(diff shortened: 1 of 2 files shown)


t3dotgg:
> [!NOTE]

Closing. The leak is real but slow (one localStorage key per opened PR, quota errors already fall back to a live read). The fix is an indexed LRU with UTF-16 byte accounting, an Effect Schema for the index, a full localStorage scan on every write, remeasurement of every retained payload, rollback on index write failure, and a read path that now mutates storage to bump recency. The key format change also drops every existing snapshot on upgrade. A 20-entry cap on a small key list is about 30 lines and fixes the same problem. Would take that as a fresh PR.

PR #10121, opened 2026-09-05
refactor(server): normalize provider settings before use

### LINKED ISSUE #5805: [Bug]: Usage page misses Codex usage from provider instances with a custom CODEX_HOME

### LINKED ISSUE #7533: [Bug]: Text generation falls back to a disabled provider — fallback reads legacy `providers` map instead of `providerInstances`

t3dotgg:
Runtime code still interprets both provider settings formats. The text-generation fallback can select a default instance that was explicitly disabled.

Normalize old settings into one instance map before runtime code reads them. Keep old file and client formats at the settings boundary. Preserve explicit disables, unknown driver configs, and sparse defaults when clients save the full map. Registry, terminal, installation, session scan, and usage reads now use instance IDs. The provider Reset action compares values with defaults instead of treating every normalized entry as an edit. Omitted config fields count as defaults for this comparison. Unknown keys, invalid config, and instance edits still show Reset.

## Verification

- 525 focused tests across 15 server, shared, and web suites pass.
- Server and web typechecks pass. Changed-file lint passes with warnings in unchanged code. Formatting passes.
- Read-only source review is clear after fixes for old-client saves, unknown driver defaults, and usage config validation.
- The sparse-config Reset fix passed 56 focused tests, web typecheck, lint, and formatting.
- Disposable browser checks pass: default settings hide Reset, an edited display name shows it, and Reset restores the default. No live state or provider session was used.

Closes #7533
Closes #5805

## Browser check

| Before | After |
| --- | --- |
| Provider settings before | Provider settings after |

After editing the display name:

Reset available after editing

### Sparse default config

Both disposable builds received the same valid server-config input: `{"providerInstances":{"codex":{"driver":"codex","enabled":true}}}`.

Before `dd1e088f54` shows Reset. After `7b62fb288b`, which includes fix `56ccf1b5d6`, hides it. The actual client ran in both checks. All non-read application requests were blocked. No real settings change or provider command ran.

| Before | After |
| --- | --- |
| Sparse default shows Reset before | Sparse default hides Reset after |

> [!NOTE]
> ### Normalize provider settings into `RuntimeServerSettings` before use
> - Introduces `resolveServerSettings` in [serverSettings.ts](https://github.com/pingdotgg/t3code/pull/10121/files#diff-ec910be99e19e00c0e09b15484059b34164d71de8d0026f4ba98471bbbb3680e), which converts persisted settings to `RuntimeServerSettings` by folding legacy enabled flags, synthesizing default provider instances from legacy `providers` entries, and dropping the legacy map from the runtime shape. `providerInstances` becomes the single source of truth for runtime consumers.
> - Updates `ServerSettingsService` contract and all consumers — usage scanner, terminal manager, provider registry hydration, provider installation, agent session scanner, and text-generation fallback — to read provider configuration from `providerInstances` directly instead of legacy `providers` fields.
> - Reconstructs a legacy `providers` map via `legacyProvidersForClient` only for client-facing redaction responses, preserving backward compatibility without making it a runtime source of truth.
> - `applySettingsPatch` filters out synthesized default instances that merely match legacy defaults, so saving unchanged settings does not persist them as explicit overrides.
> - Risk: `RuntimeServerSettings` replaces `ServerSettings` across `ServerSettingsService` methods (`getSettings`, `updateSettings`, `streamChanges`, `subscribeChanges`) and downstream parameter types in [UsageService.ts](https://github.com/pingdotgg/t3code/pull/10121/files#diff-1d90d173860922c384f65961de4b51bba0176aadd7982190e0dcf87e663f1a70), [Manager.ts](https://github.com/pingdotgg/t3code/pull/10121/files#diff-9b98fe5056129db24d311ed1bd8f66a72127cf90fa7529da47e0725f2813532e), [ProviderInstanceRegistryHydration.ts](https://github.com/pingdotgg/t3code/pull/10121/files#diff-178004a31896bcf4abd061f33ad3d95f0fd2a03d9e724863557102de06dafdc0), and [providerInstallation.ts](https://github.com/pingdotgg/t3code/pull/10121/files#diff-f4364f321d0772d33c7ca771048548f55a494c6f515bfbeffdd4965707a1de5e); any out-of-tree consumers of these signatures must update. Invalid known-driver instance configs now surface as `scanFailed` in usage instead of silently using defaults.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 56ccf1b.</sup>
> 
>

## Model

Created with GPT-6 Astra in Codex. Sparse-default correction, browser verification, and evidence upload in Codex.

### DIFF
--- a/apps/server/src/orchestration/ThreadSettlementReactor.test.ts
+++ b/apps/server/src/orchestration/ThreadSettlementReactor.test.ts
@@ -31,7 +31,7 @@ import {
   type PullRequestMergeEvent,
 } from "../pullRequest/PullRequestService.ts";
 import { ServerActivation } from "../serverActivation.ts";
-import { ServerSettingsService } from "../serverSettings.ts";
+import { resolveServerSettings, ServerSettingsService } from "../serverSettings.ts";
 import { OrchestrationCommandInvariantError } from "./Errors.ts";
 import {
   OrchestrationEngineService,
@@ -206,11 +206,13 @@ const makeHarness = Effect.fn("makeThreadSettlementHarness")(function* (options:
   const serverSettings = ServerSettingsService.of({
     start: Effect.void,
     ready: Effect.void,
-    getSettings: Ref.get(settings),
-    updateSettings,
-    streamChanges: Stream.fromPubSub(settingsChanges),
+    getSettings: Ref.get(settings).pipe(Effect.map(resolveServerSettings)),
+    updateSettings: (patch) => updateSettings(patch).pipe(Effect.map(resolveServerSettings)),
+    streamChanges: Stream.fromPubSub(settingsChanges).pipe(Stream.map(resolveServerSettings)),
     subscribeChanges: PubSub.subscribe(settingsChanges).pipe(
-      Effect.map((subscription) => Stream.fromSubscription(subscription)),
+      Effect.map((subscription) =>
+        Stream.fromSubscription(subscription).pipe(Stream.map(resolveServerSettings)),
+      ),
     ),
   });
 
--- a/apps/server/src/project/AgentSessionScanner.ts
+++ b/apps/server/src/project/AgentSessionScanner.ts
@@ -19,7 +19,6 @@ import {
   AgentSessionScanError,
   ClaudeSettings,
   CodexSettings,
-  ProviderDriverKind,
   ProviderInstanceId,
   resolveProviderInstanceEnabled,
   type AgentSessionImportSource,
@@ -930,18 +929,6 @@ export const make = Effect.gen(function* () {
           instanceId: ProviderInstanceId.make(instanceId),
           config,
         }));
-      if (!Object.hasOwn(settings.providerInstances, source)) {
-        const legacyInstance = {
-          instanceId: ProviderInstanceId.make(source),
-          config: {
-            driver: ProviderDriverKind.make(source),
-            config: settings.providers[source],
-          },
-        };
-        if (resolveProviderInstanceEnabled(legacyInstance.config)) {
-          instances.push(legacyInstance);
-        }
-      }
 
       // A shared home contains one copy of each session. Prefer the built-in
       // instance as its owner, then keep configured order for custom accounts.
--- a/apps/server/src/provider/Drivers/CodexHomeLayout.ts
+++ b/apps/server/src/provider/Drivers/CodexHomeLayout.ts
@@ -42,7 +42,7 @@ function resolveHomePath(path: Path.Path, value: string | undefined): string {
 }
 
 export const resolveCodexHomeLayout = Effect.fn("resolveCodexHomeLayout")(function* (
-  config: CodexSettings,
+  config: Pick<CodexSettings, "homePath" | "shadowHomePath">,
 ): Effect.fn.Return<CodexHomeLayout, never, Path.Path> {
   const path = yield* Path.Path;
   const sharedHomePath = resolveHomePath(path, config.homePath);
--- a/apps/server/src/provider/Layers/CursorAdapter.test.ts
+++ b/apps/server/src/provider/Layers/CursorAdapter.test.ts
@@ -33,6 +33,7 @@ import { makeCursorAdapter } from "./CursorAdapter.ts";
 import { execScriptSource, writeFakeCli } from "../../testUtils/fakeCli.ts";
 import { HostProcessPlatform } from "@t3tools/shared/hostProcess";
 const decodeCursorSettings = Schema.decodeSync(CursorSettings);
+const decodeCurrentCursorSettings = Schema.decodeUnknownEffect(CursorSettings);
 
 // Test-local service tag so the rest of the file can keep using `yield* CursorAdapter`.
 class CursorAdapter extends Context.Service<CursorAdapter, CursorAdapterShape>()(
@@ -136,7 +137,11 @@ const makeResolveCursorSettings = Effect.gen(function* () {
   const serverSettings = yield* ServerSettingsService;
   return yield* Effect.succeed(
     serverSettings.getSettings.pipe(
-      Effect.map((snapshot) => snapshot.providers.cursor),
+      Effect.flatMap((snapshot) =>
+        decodeCurrentCursorSettings(
+          snapshot.providerInstances[ProviderInstanceId.make("cursor")]?.config ?? {},
+        ),
+      ),
       Effect.orDie,
     ),
   );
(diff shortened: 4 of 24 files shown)


t3dotgg:
> [!NOTE]

Closing in favor of #10346, which fixes #7533 in about 15 lines. The full normalization changes the type ServerSettingsService hands to every consumer, changes what clients receive for providers (a config decode failure now flips a provider to disabled on old clients), and makes UsageService fail the whole scan on one invalid homePath where main used defaults. Also, this PR did not fix #5805: UsageService on the branch still scans only the default claudeAgent and codex instance ids. #8470 is the real fix for that.

PR #10263, opened 2026-09-06
feat(skills): browse installed skills by environment and project

MatthewFeroz:
Skills with the same name can refer to different files or have different invocation policies across provider instances and environments. This adds a read-only Skills page that shows each provider's actual discovery results and the instructions belonging to the selected file.

## Implementation

- Retains the browsing UI, sidebar and command-palette entry points, environment/project selection, search, and instruction rendering.
- Uses existing provider-instance snapshots and workspace discovery before composer filtering. Removes the separate `skills` CLI and its `npx`, npm-prefix, YAML-parsing, and catalog subprocess machinery.
- Shows disabled skills and each installation's enabled status and user/agent invocation policy. Provider-instance, enabled-status, invocation, and scope filters apply together to each installation.
- Groups installations by their resolved file path, keeping different same-name files separate. Detail requests identify the file rather than selecting the first matching name.
- Preserves each provider instance's driver, skill name, display name, description, original installation path, enabled state, and invocation restrictions within a shared-file group. Symlink installations show both the linked location and resolved destination.
- Corrects Grok discovery to preserve `userInvocable: false` independently of enabled status; composer filtering still excludes these skills.
- Keeps authenticated environment HTTP transport, capability guards, bounded file reads, and a two-request discovery limit. Cancellation stays attached to the request; provider snapshots retain their own refresh sharing. Refresh re-runs discovery.

Enable/disable controls can follow where providers support them. Installation, editing, removal, synchronization, and native mobile navigation are outside this PR.

## Verification

- 62 focused tests passed across catalog behavior, Grok discovery, client filtering, authenticated environment HTTP requests, and environment capabilities, plus one composed server-route smoke test.
- Web and server typechecks passed. Targeted lint reports only pre-existing CommandPalette warnings.
- Chromium against an isolated local T3 server using real discovery from two configured Claude instances:
  - Two `release-checklist` files stayed separate and displayed their own local-preview versus staging instructions.
  - `manual-review` and a `shared-review` symlink grouped together while retaining both provider-specific names, paths, and policies.
  - Disabled skills remained visible. Combining Work Claude with Disabled produced no matches rather than borrowing another installation's status.
  - The agent-invocation filter excluded the user-only shared skill and showed the agent-only skill correctly.
  - Editing a disposable file and clicking Refresh displayed the updated instructions.
  - Command-palette navigation worked. At 390 px width, detail/back navigation worked without horizontal overflow. No page errors were observed.
- Environment isolation is covered by focused catalog tests. This pass did not test two physical machines, native mobile, or cross-environment synchronization.

### Before — earlier CLI-based browser

Earlier Skills page

### After — shared file with provider-specific installations

Shared skill identity and symlink paths

### After — distinct files with the same name

Same-name files remain separate

### After — status and invocation filters

Disabled skill remains inspectable

Provider and agent-invocation filters

[Responsive web screenshot](https://github.com/MatthewFeroz/t3code/releases/download/pr-evidence-skills-provider-discovery/t3-skills-after-narrow.png)

Implemented and verified with GPT-6 in Codex, through T3 Code.


shivamhwp:

We'd like to keep the browsing UI and use T3's existing provider discovery to populate it. The separate `skills` CLI misses some skills already found by the composer, omits enabled and invocation status, and adds an `npx` requirement for desktop installations.

Please use provider discovery results before composer filtering, so the page can show disabled skills and explain who can invoke each skill. Provider-instance, enabled-status, and invocation filters would make that information easier to browse.

Skill identity also needs to preserve the provider and file path. Different files with the same name currently get combined, which can show one provider's instructions under another provider's badge. For symlinked installations, show the linked location and resolved destination, and group shared files by their resolved path rather than their name. Preserve provider-specific names and status within that group.

Enable/disable controls could follow where the provider supports them.

MatthewFeroz:
@shivamhwp Addressed your feedback in 0d73b5e702269b98e6aca1b77bb2a5833d4f6786 and merged current main to resolve the conflicts.

- The browsing UI now consumes existing provider-instance discovery before composer filtering; the separate skills CLI and npx dependency are gone.
- Disabled skills remain inspectable, with per-installation enabled and invocation status. Added provider-instance, status, invocation, and scope filters.
- Identity is based on the resolved file, preserving separate same-name files and each provider's original name, path, and policy within shared-file groups. Symlink locations and destinations are both visible.
- Enable/disable controls remain deferred as suggested.

Verified in Chromium with real discovery from two Claude instances, including distinct same-name files, a shared symlink, disabled and agent-only skills, combined filters, Refresh after an edit, and responsive navigation. The PR body has the complete mapping and before/after screenshots. 63 focused checks passed locally; web/server typechecks are clean.

Ready for another review once the current CI run completes.

juliusmarminge:

PR #10291, opened 2026-09-06
refactor(server): preserve pending operation kind on upgrade

t3dotgg:
Compaction and rewind need ordered database migrations. Rewind must not install migration 49 before the operation-state column in migration 48 exists.

Add the nullable `operation_kind` column and classify existing pending messages. Old writers can still omit the column. This PR does not change runtime behavior. This PR starts the dependency chain for compaction lifecycle and rewind.

The focused migration test passed. It covers mixed-case commands, whitespace, arguments, attachments, and old writers after the upgrade. Independent source review confirmed that this migration is additive and compatible with old writers.

Created with GPT-6 Astra in Codex. Recovery and verification continued in Codex.

> [!NOTE]
> ### Add migration 048 to backfill `operation_kind` on pending projection turns
> - Adds a nullable `operation_kind` column to projection turns via [048_ProjectionPendingOperation.ts](https://github.com/pingdotgg/t3code/pull/10291/files#diff-646bf7895ccc67b8c522b74e87639136753774c7facaf794f4e2c94e7dad4fcf)
> - Selects pending user turns without a turn ID or checkpoint count, joins them to their same-thread messages, and classifies each row as `compact` when the message has no attachments and its trimmed text matches the compact command; all other eligible rows are classified as `ordinary` turns
> - Registers migration 048 in the [migration registry](https://github.com/pingdotgg/t3code/pull/10291/files#diff-87557f6881a7a3dd24fa7e31f895b009cc6ede9a3571779088a10c5093110707) and adds test coverage in [048_ProjectionPendingOperation.test.ts](https://github.com/pingdotgg/t3code/pull/10291/files#diff-5055892ffb8c0b9e57a40d94e65b87401436523bfcaf1670068154993a946d84) covering compact/ordinary classification, attachments, row ordering, and nullable values for legacy inserts after downgrade
> - Behavioral Change: rows written by an older server after downgrade will have a null `operation_kind`; readers must tolerate null values in this column
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d858532.</sup>
> 
>

### DIFF
--- a/apps/server/src/persistence/Migrations.ts
+++ b/apps/server/src/persistence/Migrations.ts
@@ -59,6 +59,7 @@ import Migration0044 from "./Migrations/044_ClearAutomaticProjectModelDefaults.t
 import Migration0045 from "./Migrations/045_ProjectionProjectsAutoPull.ts";
 import Migration0046 from "./Migrations/046_RepairAutomaticSettlementTimestamps.ts";
 import Migration0047 from "./Migrations/047_ProjectionProjectIcon.ts";
+import Migration0048 from "./Migrations/048_ProjectionPendingOperation.ts";
 
 /**
  * Migration loader with all migrations defined inline.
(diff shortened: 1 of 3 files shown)


t3dotgg:
> [!NOTE]

Closing. Nothing on main reads or writes operation_kind, and PR #10123 (which is stacked on this) never touches the column either. A migration with no reader is dead schema, and no projection writer sets it for new rows so the backfilled value goes null on the first new turn. If the compaction-lifecycle work in #10294 lands, the migration should ship inside that PR. Retargeting #10123 to main.

PR #10314, opened 2026-09-06
fix(cursor): reject incomplete skill file reads

yashranaway:
## What Changed

Treat Cursor skill-file read failures as incomplete discovery, consistent with directory and stat failures. Omit unreadable entries from best-effort discovery and fail the workspace probe so callers can retry.

## Why

An unreadable SKILL.md currently becomes an enabled skill with no metadata. The probe reports success, so an incomplete catalog can be cached and flags such as `user-invocable: false` are lost.

## Testing

- Five focused Cursor skill tests pass.
- The unreadable-file regression fails against the original implementation and passes with the fix, including metadata recovery after retry.

## Checklist

- [x] One focused change

Model: GPT-6 Astra
Harness: T3 code

> [!NOTE]
> ### Fix `discoverSkillsInRoot` to reject skills with unreadable `SKILL.md`
> - When a `SKILL.md` read returns no contents (e.g. permission error), the scanner now clears the default frontmatter so the directory is skipped instead of being emitted with default metadata.
> - The read is routed through the scan-budget-aware path, matching other reads.
> - Default frontmatter is still retained for files skipped by size or remaining-byte checks.
> - Adds a test simulating a `PermissionDenied` failure and verifying recovery on a subsequent normal probe.
> - Risk: skills whose `SKILL.md` is present but unreadable at scan time will no longer appear in the CLI catalog; check `discoverSkillsInRoot` in [CursorSkills.ts](https://github.com/pingdotgg/t3code/pull/10314/files#diff-528e28e34e83cfb127b20e204b245df257b3295b01f91ac1ceed4d8219baf928) if missing skills are expected.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 0a5fb20. 1 file reviewed, 1 issue evaluated, 1 issue filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> <details>
> <summary>apps/server/src/provider/Drivers/CursorSkills.ts — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 167](https://github.com/pingdotgg/t3code/blob/0a5fb200c2be6effc01c5e8cf0d2f40fd358adc6/apps/server/src/provider/Drivers/CursorSkills.ts#L167): When `SKILL.md` is removed or renamed after the successful `stat` at line 163 but before this read, `readFileString` fails with `NotFound`. `orUndefined` deliberately does not set `budget.incomplete` for `NotFound`, while the new branch omits the skill, so `probeCursorSkills` returns a successful but incomplete catalog rather than triggering a retry. <b>[ Out of scope (triage) ]</b>
> </details>
>
>
> </details>
>

### DIFF
--- a/apps/server/src/provider/Drivers/CursorSkills.ts
+++ b/apps/server/src/provider/Drivers/CursorSkills.ts
@@ -164,10 +164,12 @@ const discoverSkillsInRoot = Effect.fn("discoverCursorSkillsInRoot")(function* (
     if (skillInfo?.type === "File") {
       let frontmatter: CursorSkillFrontmatter | undefined = { cliVisible: true };
       if (skillInfo.size <= MAX_SKILL_BYTES && skillInfo.size <= input.budget.remainingBytes) {
-        const contents = yield* orUndefined(fileSystem.readFileString(skillPath));
+        const contents = yield* orUndefined(fileSystem.readFileString(skillPath), input.budget);
         if (contents !== undefined) {
           input.budget.remainingBytes -= skillInfo.size;
           frontmatter = parseSkillFrontmatter(contents);
+        } else {
+          frontmatter = undefined;
         }
       }
       const name = path.basename(directory).trim();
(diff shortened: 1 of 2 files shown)


juliusmarminge:

yashranaway:

PR #10317, opened 2026-09-06
fix(server): canonicalize usage transcript sources

yashranaway:
## What Changed

Resolve existing Claude, Codex, and Grok transcript directories to their real paths before scanning and building source fingerprints. Missing or inaccessible directories keep their configured paths and existing handling.

## Why

Two environments configured through different paths to the same transcript directory produce different fingerprints. A symlinked provider home makes the usage page count the same tokens and sessions twice even though the filesystem identity matches.

## Testing

- `UsageService.test.ts` passes, 7/7.
- A real directory alias scanned through two service instances reproduces 10 output tokens instead of 5 before the fix; afterward the transcript and session count once.

## Checklist

- [x] One focused change

Model: GPT-6 Astra
Harness: T3 code

> [!NOTE]
> ### Canonicalize transcript directory paths via real-path resolution in `UsageService.collectDirs`
> `UsageService.collectDirs` now resolves each configured transcript directory through `fs.realpath` before scanning, so a direct path and a directory alias (e.g. a junction) yield the same source fingerprint for usage merging. When real-path resolution fails, it falls back to the original configured path. A live test confirms that merging summaries from direct and aliased Claude transcript homes produces no duplicate token or session counts.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6f17728.</sup>
> 
>

### DIFF
--- a/apps/server/src/usage/UsageService.test.ts
+++ b/apps/server/src/usage/UsageService.test.ts
@@ -7,7 +7,13 @@ import * as NodePath from "node:path";
 import { assert, describe, it } from "@effect/vitest";
 import * as NodeServices from "@effect/platform-node/NodeServices";
 import { HostProcessEnvironment } from "@t3tools/shared/hostProcess";
-import { UsageDay, type UsageSummaryInput } from "@t3tools/contracts";
+import {
+  EnvironmentId,
+  USAGE_CONTRACT_VERSION,
+  UsageDay,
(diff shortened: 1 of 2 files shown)


yashranaway:
Closing because #8470 already canonicalizes transcript directories before building usage fingerprints, covering this fix.

PR #7680, opened 2026-08-20
fix(server): advertise truecolor in the integrated terminal

shubhxho:
The integrated terminal does not advertise truecolor when its parent environment omits `COLORTERM`. Programs that use that variable to choose their color output can render reduced-color diffs.

Default missing or empty `COLORTERM` to `truecolor` after merging the terminal's runtime environment. Keep any nonempty caller value. The default belongs in the shared terminal manager so both Node and Bun PTY backends receive it; it does not change `TERM` or mutate the parent environment.

Addresses [#7648](https://github.com/pingdotgg/t3code/issues/7648).

Verification on main [ed2bdbb27](https://github.com/pingdotgg/t3code/commit/ed2bdbb278c4fedf6ce79d4925908b759f7fd992):

- The actual `TerminalManager.open` regression fails before the change, with `COLORTERM` absent in captured spawn options. The Linux, macOS, and Windows host-mode cases all fail against main.
- With the change, the manager sets missing/empty values, preserves explicit parent values and runtime overrides, and leaves the frozen input environment untouched.
- `vp test run apps/server/src/terminal/Manager.test.ts apps/server/src/terminal/NodePtyAdapter.test.ts apps/server/src/terminal/BunPtyAdapter.test.ts --maxWorkers=2`: 69 tests pass. Targeted lint/format and the server typecheck pass.

The host-mode tests run on Linux with mocked PTY processes. They prove the shared spawn environment, not native Windows/macOS execution or the reported lazygit rendering. All CI jobs that ran pass on [917b6e85](https://github.com/pingdotgg/t3code/commit/917b6e85b14d2c84e74577378c8e42165bad215a); Bugbot and Macroscope correctness pass, with no unresolved review threads. Macroscope approvability is neutral because the new terminal default needs a maintainer decision.

The root reviewer verified the actual Linux Chromium terminal on main [82f64cd8](https://github.com/pingdotgg/t3code/commit/82f64cd8d046ab4ee8588f7bcbc1e66a4a0c80fe), then integrated this exact production change and opened a new terminal. Before, the shell reports an empty COLORTERM. After, it reports truecolor. The same direct 24-bit red and green backgrounds render in both runs, confirming that the default advertises a capability the client already has. This is not a lazygit or native macOS/Windows run.

Before:

Before: the real integrated shell has an empty COLORTERM despite rendering direct 24-bit red and green

After:

After: a newly opened terminal advertises truecolor and retains the existing 24-bit rendering capability

[...]

PR #10331, opened 2026-09-06
fix(server): avoid color probe hangs in setup terminals

juliusmarminge:
Automatic worktree setup can appear stuck with its command visible in the terminal. The command has already started: Vite+ 0.3.0 waits for OSC color-query replies before printing its header, and pressing Enter releases the wait. The default `COLORTERM=truecolor` added in [#7680](https://github.com/pingdotgg/t3code/pull/7680) exposes this on main as well as v2.

Clear `COLORTERM` for automatic setup terminals and honor an explicitly empty terminal environment override. Setup retains `TERM=xterm-256color` and ANSI colors without inviting truecolor probing before a client attaches. Ordinary terminals retain their truecolor default. This is a targeted workaround for automatic setup; it does not change Vite+'s query timeout or terminal replay behavior.

Verification:
- 83 focused tests passed across `ProjectSetupScriptRunner.test.ts` and `terminal/Manager.test.ts`, including explicit empty overrides on Linux, macOS, and Windows host configurations.
- Server typecheck passed. Targeted lint passed with one existing `prefer-array-find` warning outside the change.
- Real macOS PTY, Vite+ 0.3.0, identical disposable package and `vp i --offline --ignore-scripts` command:

| Before, main environment | After, setup environment |
| --- | --- |
| `COLORTERM=truecolor`: emits OSC 10, OSC 4 for colors 4/5, and DA1; no installer output until Enter is sent. | Empty `COLORTERM`: no color queries or input required; install exits 0. |

After output:
```text
VITE+ - The Unified Toolchain for the Web
Already up to date
Done in 181ms using pnpm v11.10.0
```

No UI layout changed; the PTY output above is the before/after evidence. The fix is server-side and applies to automatic setup launched from web, desktop, or mobile, including remote clients.

Model: GPT-6. Harness: Codex.

> [!NOTE]
> ### Fix setup terminal color probe hangs by emptying `COLORTERM`
> - Setup terminals may run before a client attaches, so `ProjectSetupScriptRunner.runForThread` now sets `COLORTERM` to an empty string to avoid advertising truecolor during that phase.
> - `createTerminalSpawnEnv` in [Manager.ts](https://github.com/pingdotgg/t3code/pull/10331/files#diff-9b98fe5056129db24d311ed1bd8f66a72127cf90fa7529da47e0725f2813532e) now preserves an explicitly empty runtime `COLORTERM` value instead of replacing it with the truecolor fallback. The fallback still applies only when the runtime environment does not define `COLORTERM` at all.
> - Behavioral Change: terminals launched with an explicit empty `COLORTERM` override will no longer get the `truecolor` default; consumers that relied on the fallback overriding an empty value will see `COLORTERM` unset/empty instead.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e1da335.</sup>
> 
>

### DIFF
--- a/apps/server/src/project/ProjectSetupScriptRunner.test.ts
+++ b/apps/server/src/project/ProjectSetupScriptRunner.test.ts
@@ -108,7 +108,11 @@ describe("ProjectSetupScriptRunner", () => {
         terminalId: "setup-default-setup",
         cwd: "/repo/worktrees/a",
         worktreePath: "/repo/worktrees/a",
-        env: { T3CODE_PROJECT_ROOT: "/repo/project", T3CODE_WORKTREE_PATH: "/repo/worktrees/a" },
+        env: {
+          T3CODE_PROJECT_ROOT: "/repo/project",
+          T3CODE_WORKTREE_PATH: "/repo/worktrees/a",
+          COLORTERM: "",
+        },
       });
       expect(write).toHaveBeenCalledWith({
         threadId: "thread-1",
@@ -207,6 +211,7 @@ describe("ProjectSetupScriptRunner", () => {
           env: {
             T3CODE_PROJECT_ROOT: "/repo/project",
             T3CODE_WORKTREE_PATH: "/repo/worktrees/a",
+            COLORTERM: "",
           },
         });
         expect(write).toHaveBeenCalledWith({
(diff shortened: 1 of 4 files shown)


juliusmarminge:
Superseded by #11843 (same fix: avoid color-probe hangs in setup terminals; #11843 uses NO_COLOR=1 + FORCE_COLOR=0).

PR #10337, opened 2026-09-06
fix(server): mark Cursor transport error answers as failed

### LINKED ISSUE #7830: [Bug]: Cursor ACP retriable network failures are treated as successful assistant turns
### Before submitting

### Area

apps/server

### Steps to reproduce

Cursor's ACP implementation currently has a failure mode where a recoverable transport/network failure is emitted as an agent_message_chunk, rather than failing the session/prompt JSON-RPC request.

Typical examples I've observed are:

```
Error: RetriableError: ...
Error: ConnectError: [unavailable] ...
Error: ConnectError: [aborted] aborted
```

as well as HTTP/2 CANCEL / stream-reset variants and:

```
Something went wrong communicating with the server. Please try again.
```

After emitting that text as an assistant message, Cursor ACP can return a successful session/prompt result with a normal stop reason such as end_turn.

So from a standards-compliant ACP client's perspective the sequence looks like a legitimate answer:

```
session/update
  agent_message_chunk("Error: RetriableError: ...")


session/prompt
  result: { stopReason: "end_turn" }
```

[Cursor has acknowledged that this is broken behavior](https://forum.cursor.com/t/acp-error-responses-not-being-passed-through-correct-error-channels/163218) in their ACP implementation, and this was reported back on June 15th. Their normal interactive agent handles these transient failures internally, but ACP leaks the failure out as assistant content.

This is intermittent against the real service, but it should be straightforward to reproduce deterministically with a mock Cursor ACP process that sends the above agent_message_chunk followed by a successful session/prompt response.

### Expected behavior

T3 should not persist a known Cursor ACP retriable transport failure as the assistant's final answer and consider the turn successfully completed.

Ideally, for Cursor specifically, T3 would recognize this narrow failure class, suppress the bogus assistant tail, and retry the prompt with bounded backoff.

### Actual behavior

T3 currently does exactly what ACP tells it to do:

AcpRuntimeModel.parseSessionUpdateEvent() converts every text agent_message_chunk into a ContentDelta.
CursorAdapter immediately publishes that delta as normal assistant content.
sendTurn() receives the successful session/prompt result and emits turn.completed; anything other than stopReason: "cancelled" is treated as completed.

Consequently, Cursor's transport failure becomes the assistant's last words and the transient network interruption kills the turn.

I don't think the shared ACP parser is wrong here. ACP uses normal JSON-RPC error semantics for method failures, so a generic ACP client has no principled reason to interpret arbitrary assistant text as a transport exception. This workaround should therefore live in the Cursor-specific adapter/support layer rather than changing ACP behavior globally.

### Impact

### Version or commit

0.0.34-nightly.20260821.1151 (be7d35aaeb49)

### Environment

macOS 26.6.1, node.js 26.7.0, bun 1.3.14, cursor-agent 2026.08.11-e8db854

### Logs or stack traces

```shell

```

[Screenshot 2026-08-21 at 9.46.20 AM.png](https://github.com/user-attachments/assets/bf19c5be-b57b-45bf-a556-3ccb62ef2e29)
[Screenshot 2026-08-21 at 9.47.29 AM.png](https://github.com/user-attachments/assets/a7c09de9-4fd3-4828-906e-8a0abcc18e71)
[Screenshot 2026-08-21 at 9.48.07 AM.png](https://github.com/user-attachments/assets/59dc613b-6098-437d-a280-0a002761c444)

### Workaround

I wrote a small ACP supervisor in `robertmsale/skizzles`, under:

`packages/cursor-acp-shim`

T3 points its existing Cursor Binary path setting at the shim. The shim then execs the real cursor-agent acp and otherwise behaves as a transparent ACP proxy.

It only intercepts the "network death emitted as assistant text" class.

The important parts of the workaround are:

Potential terminal error text is buffered long enough to determine whether it is a Cursor transport dump rather than immediately exposing it to T3.
Short terminal RetriableError: dumps are considered recoverable.
Known ConnectError transport codes, HTTP/2 CANCEL/reset failures, socket-reset errors, and Cursor's short server-error copy are also recognized.
Normal assistant prose containing those words is not swallowed.
Errors inside fenced code are not swallowed.
Auth/subscription failures are not swallowed.
HTTP errors from an application Cursor is debugging are not swallowed.
A trailing transport dump after otherwise legitimate assistant prose can be removed without discarding the legitimate prose.
On a match, the bogus assistant content and matching successful session/prompt completion are suppressed and the same prompt is replayed.
Replays use exponential backoff: 2s, 4s, 8s, 16s, 32s, then capped at 60s.
session/cancel aborts a pending retry.
If the Cursor child has died, the shim can respawn it and restore the visible ACP session with initialize / authenticate / session/load, preserving the original session ID, cwd, and MCP servers.
After the retry budget is exhausted, it returns a JSON-RPC error instead of another fake assistant message.

I've been using this in front of T3 and it has made Cursor dramatically more reliable for me.

I don't necessarily think T3 should copy the shim wholesale. It exists outside T3 because I wanted to prove the behavior could be corrected without changing T3 first. But the provider-specific logic would fit much more naturally inside CursorAdapter / CursorAcpSupport.

One important implementation detail is that detection needs to happen before candidate text is committed to the T3 runtime. Once the fake agent_message_chunk has been persisted/rendered as assistant content, retrying the prompt alone leaves the garbage message in the timeline.

shivamhwp:
Closes #7830 
Cursor CLI can emit a transport failure as assistant text and then return successful `end_turn`. T3 consequently marks the failed turn as completed. Refs #7830.

Recognize standalone transport diagnostics in T3's Cursor adapter and return a provider request error. This triggers the existing Failed status and error toast while preserving the diagnostic for inspection and manual retry. Ordinary prose, fenced code, cancellation and structured ACP error handling retain their existing behavior.

The upstream fault is in Cursor CLI's ACP implementation. Cursor support [acknowledged the error-channel problem](https://forum.cursor.com/t/acp-error-responses-not-being-passed-through-correct-error-channels/163218/5) and [confirmed a related stale-authentication variant](https://forum.cursor.com/t/cursor-agent-acp-live-process-returns-sign-in-prompt-as-assistant-content-after-auth-state-goes-stale/163787/5). The authentication variant is background evidence and is not handled by this transport-only matcher.

Our standalone reproduction talks directly to `cursor-agent acp` without T3 or `effect-acp`. Authenticated CLI `2026.09.02-c22c1a3` completed a normal prompt. Closing only that test process's TCP connections and directing reconnects to a closed local port made Cursor emit `Error: RetriableError: [unavailable] connect ECONNREFUSED ...` as `agent_message_chunk`, followed by successful `end_turn`. Restoring connectivity let the next prompt succeed in the same process and session. This places the workaround at the Cursor adapter boundary; generic `effect-acp` should continue interpreting assistant content as text.

Repeated through the real T3 web client and server on `92c363ba`: the failure displayed Failed and the error toast, with error turn/session states in the disposable database. The recovery turn completed and cleared the session error. No ACP responses were fabricated. This verifies an induced connection refusal, not every transport failure or the exact historical WritableIterable failure.

The matcher deliberately leaves mixed prose with an error tail unchanged. Unrecognized wording can be missed, and a legitimate answer consisting only of an identical diagnostic can be misclassified. Automatic retries are excluded because a failed turn may already have performed work that replaying the prompt could repeat. Reassess this workaround once supported Cursor CLI versions report these failures through structured ACP errors.

Validation: the mock-provider regression fails on the tested base and passes with this change. All 40 focused Cursor tests, server typecheck and targeted lint passed on this head. Detection retains at most one bounded line of the answer.

Model: GPT-6. Harness: Codex.

> [!NOTE]
> ### Mark Cursor transport-error assistant responses as failed in `CursorAdapter`
> - Cursor sometimes returns a transport diagnostic as a successful assistant text stream instead of an error. The adapter now detects this and rejects the turn with `ProviderAdapterRequestError`.
> - Adds a `CursorTransportFailure` tracker in `apps/server/src/provider/acp/CursorTransportFailure.ts` that classifies newline-delimited streamed chunks, enforces a max line length, and only flags standalone Cursor diagnostics optionally followed by stack frames. Prose, code, and unrelated errors are not classified.
> - `makeCursorAdapter` initializes one tracker per session, resets it on assistant item starts and non-steering turns, feeds content deltas into it, and drains pending events after the ACP prompt completes to convert a recognized transport diagnostic into a request error.
> - Cancelled turns and prompts with other in-flight prompts are not subject to this check.
> - Behavioral Change: a Cursor assistant response containing only a recognized transport diagnostic now fails the turn instead of emitting `turn.completed`; the full assistant text is not retained by the tracker.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f9ad3d4.</sup>
> 
>


KaKi87:
Hi,

~~I'm afraid I'm still experiencing this issue on version 0.0.40-nightly.20260907.1346 :~~

**[EDIT] Nevermind, I just realized I've been updating the desktop app but not the remote environment I'm running on, sorry**

Capture d’écran 2026-09-07 à 19 51 16

Thank you

KaKi87:
Hi again,

So now the errors are indeed shown as errors but T3 Code isn't retrying the retriable requests :

image

Thanks

PR #10360, opened 2026-09-06
fix(desktop): prevent a second server from breaking active threads

### LINKED ISSUE #6097: [Bug]: Desktop starts a second backend against the background service database

t3dotgg:
Opening desktop while the background server is running can start a second server on port 3774 against the same database. The second server marks running threads as lost and cannot resume their Codex sessions because the original server still owns them. Follow-up messages then fail with `already has an active writer`.

Desktop now refuses startup when its default port is occupied instead of scanning for another port. The error explains how to use the running server or a separate T3 home. Explicit port settings retain their current behavior.

This brings [#9003](https://github.com/pingdotgg/t3code/pull/9003) onto current main after confirming the exact failure in a live thread. Credit to @maxibotstef for the original fix. This covers the automatic desktop fallback path. General server directory locking is separate.

Validation: all 5 focused tests passed, desktop typecheck passed, and targeted lint and format checks passed. No browser or desktop UI was launched.

Closes #6097

### DIFF
--- a/apps/desktop/src/app/DesktopApp.ts
+++ b/apps/desktop/src/app/DesktopApp.ts
@@ -32,7 +32,6 @@ import * as DesktopUpdates from "../updates/DesktopUpdates.ts";
 import * as DesktopWslBackend from "../wsl/DesktopWslBackend.ts";
 
 const DEFAULT_DESKTOP_BACKEND_PORT = 3773;
-const MAX_TCP_PORT = 65_535;
 const DESKTOP_BACKEND_PORT_PROBE_HOSTS = ["127.0.0.1", "0.0.0.0", "::"] as const;
 
 const makeDesktopRunId = Crypto.Crypto.pipe(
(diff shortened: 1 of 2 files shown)


t3dotgg:
> [!NOTE]
> 🤖 **GPT-6 Astra responding on behalf of Theo**

@cursor review

shivamhwp:

Please correct the startup error's recovery guidance. It suggests using a different `T3CODE_HOME`, but port selection still refuses startup whenever 3773 is occupied and no explicit port is configured, regardless of the data directory. Changing only T3 home therefore repeats the same error. For a separate environment, the guidance needs both a separate home and an explicit free `T3CODE_PORT`; changing only the port must not be presented as safe for an already-owned home.

tomasmach:
Me on behalf of me:

Please merge this, been hitting this error pretty often, thanks!

PR #10364, opened 2026-09-06
fix(server): stop a second instance hijacking the Tailscale Serve mapping

### LINKED ISSUE #10363: [Bug]: Tailscale Serve mapping is hijacked by a secondary instance and left stale on unclean exit, never reconciled
### Before submitting

### Area

apps/server

### Steps to reproduce

Tailscale Serve is enabled (`tailscaleServeEnabled`, serve port `10010` in my case), and the desktop app is running normally — backend on `DEFAULT_DESKTOP_BACKEND_PORT` (3773), and `https://<machine>.<tailnet>.ts.net:10010` reaches it.

1. `open -n -a "T3 Code (Alpha)"` on macOS. A **second** instance starts, because `@clerk/electron`'s `createClerkBridge` skips `requestSingleInstanceLock()` on darwin, so `bridge.isPrimaryInstance` is always `true` there and the `second-instance` path in `apps/desktop/src/app/DesktopClerk.ts` never fires.
2. Its backend finds 3773 busy and scans up to **3774** (`apps/desktop/src/app/DesktopApp.ts`).
3. `tailscale serve status` now shows the shared serve port pointing at `http://127.0.0.1:3774` — the healthy primary on 3773 has been silently evicted from the tailnet URL. `~/.t3/userdata/server-runtime.json` has also been overwritten with the second instance's pid/port.
4. Close the second window. On macOS the app does not quit on `window-all-closed`; the instance lingers headless and (in both occurrences here) eventually exited **without** the release finalizer running.
5. From the phone / any tailnet client, hit the tailnet URL.

Both occurrences on this machine came from step 1 happening by accident, not deliberately — a second launch is easy to trigger on macOS.

### Expected behavior

A second server that finds the Tailscale Serve slot already fronting a **live** T3 server should leave it alone and log that it did. The primary should keep the tailnet URL. And whichever server is publishing the mapping should be able to recover it if it goes stale.

### Actual behavior

The mapping is tailnet-wide state keyed only by HTTPS port, and `apps/server/src/server.ts` (L619–671 on `main` @ `223ff44`) writes it **unconditionally** in the `acquireRelease` acquire and removes it **unconditionally** in release:

- **Hijack.** Acquire waits for activation, reads the bound port, and calls `ensureTailscaleServe({ localPort, servePort, localHost: "127.0.0.1" })` with no check of what is already there. Whoever starts last wins the slot.
- **Stale on unclean exit.** The finalizer only runs on a clean scope close. When the secondary died without it, the mapping stayed pointed at the now-dead 3774 and the tailnet URL returned `endpoint returned undeclared status 502` from Tailscale.
- **Symmetric hazard even on a clean exit.** Had the finalizer run, `disableTailscaleServe({ servePort })` would have removed the mapping outright — also breaking the primary, which by then owned nothing.
- **Nothing reconciles.** The primary never re-runs `ensureTailscaleServe`, and the acquire failure path is `Effect.catch → Effect.logWarning`, so even a hard failure is invisible unless you are reading server logs. The only self-heal in the codebase is `resolveTailscalePairingBase` in `apps/server/src/cli/pair.ts`, and that is manual (`t3 pair`) — and it is itself blocked here, because `discoverPairTarget` skips the dead pid recorded in the clobbered `server-runtime.json`.

Note that `pair.ts` already has exactly the guard the server layer is missing: it probes `/.well-known/t3/environment` through the existing mapping and refuses the port when it fronts a different live environment (`ServesOtherEnvironmentError` / `ServePortOccupiedError`). The server startup path just doesn't do it.

### Impact

### Version or commit

0.0.38 (behavior still present on `main` @ `223ff44`)

### Environment

macOS 15 (Darwin 25.6.0), T3 Code (Alpha) 0.0.38, `@clerk/electron` 0.0.37, tailscale CLI on the host

### Logs or stack traces

```shell
# second instance starts and takes 3774
21:28:42  desktop.backendInstance.start
21:28:42  configureFromSettings { port: 3774 }
# its window is closed 20s later; the process lingers headless ~90 min
21:29:02  desktop.lifecycle.windowAllClosed
# no "Tailscale Serve disabled" log line was ever emitted by that instance

# after it was gone, from a tailnet client:
endpoint returned undeclared status 502
```

### Workaround

Re-run `t3 pair --tailscale` from the surviving instance (or `tailscale serve --bg --https=<servePort> http://127.0.0.1:3773` by hand) to repoint the mapping. If `server-runtime.json` still records the dead pid, pairing has to be pointed at the live server explicitly.

---

Follow-up, deliberately kept out of the fix I'm about to open so the PR stays to one concern: a secondary instance also overwrites `~/.t3/userdata/server-runtime.json` even while the recorded pid is still alive, which is what breaks `t3 pair`'s own discovery afterwards. Happy to send that separately if it's wanted.

sensei-woo:
Fixes #10363.

## What Changed

The Tailscale Serve mapping is now claimed only when it is free, already ours, or stale — never when it still reaches a live server.

- **`apps/server/src/tailscaleServe.ts` (new).** `acquireTailscaleServe` reads the current mapping before writing:
  - already points at this server's loopback port → adopt it, log, don't rewrite;
  - points somewhere else and that target still answers `/.well-known/t3/environment` → log a warning naming the environment it found and return `null`, i.e. **not owned**;
  - points somewhere else and that target is unreachable → log the reclaim and write the mapping;
  - no mapping, or the read fails → treat as unclaimed and write.

  `releaseTailscaleServe` re-reads the mapping and calls `disableTailscaleServe` only while it still points here.
- **`apps/server/src/server.ts`.** The `acquireRelease` layer now calls those two instead of `ensureTailscaleServe` / `disableTailscaleServe` directly. Net −42 lines; the acquire body is down to "wait for activation, read the bound port, delegate". `HttpClient` is already provided to this layer via `FetchHttpClient.layer`.
- **`apps/server/src/environmentProbe.ts` (new).** `probeEnvironmentDescriptor` lifted verbatim out of `cli/pair.ts` so the server path reuses the exact liveness check `t3 pair` already had (`descriptor` / `unreachable` / `not-a-t3-server`; 502/503/504 and transport errors count as unreachable). `pair.ts` imports it now and loses its private copy (−47 lines).
- **`packages/tailscale/src/tailscale.ts`.** Adds `readTailscaleServeTarget` / `parseTailscaleServeTarget` over `tailscale serve status --json`, reading only the `Web["<host>:<port>"].Handlers["/"].Proxy` leaf as unknown rather than modelling a shape that moves between tailscale releases. The `status` and `serve status` readers now share one `captureTailscaleCommand` runner so the scoped spawn, the timeout and the redacted error shape stay identical (including the existing `catchDefect` ENOTDIR handling). No existing export changed signature.

## Why

The mapping is tailnet-wide state keyed only by HTTPS port, but the server wrote it unconditionally on startup and removed it unconditionally on shutdown, so every server started with `tailscaleServeEnabled` competes for the same slot. On macOS a second desktop instance is easy to get (no single-instance lock on darwin), it lands on the next free backend port, and it repoints the tailnet URL away from the healthy first one. If it then exits without its finalizer running, the mapping is left pointing at a dead port and the tailnet URL 502s with nothing in the codebase to reconcile it. Full timeline in #10363.

The approach follows the guard `resolveTailscalePairingBase` in `cli/pair.ts` already applies for the same question, rather than inventing a second policy — hence sharing the probe rather than copying it. Ownership itself is decided from `tailscale serve status --json` (target port + loopback host) rather than from the probed environment id, because that is also what makes the two safety properties exact:

- **Never hijacks a live primary.** Any non-`unreachable` probe result — same environment, another environment, or something that isn't T3 at all — leaves the mapping untouched and marks this server as not owning it.
- **Never removes a mapping it doesn't own.** Release is a no-op unless the mapping still points at this server's port, so a secondary that lost the race can't tear down the primary's mapping on its way out.

The stale-mapping case also self-heals now, which is the part that actually bites users: a server that starts while the mapping points at a port nothing answers on takes the slot back instead of leaving the tailnet URL broken.

## UI Changes

None — server-side only.

## Checklist

## Testing

New `apps/server/src/tailscaleServe.test.ts` (6 cases) drives the logic through a recording `ChildProcessSpawner` mock and an `HttpClient` mock: writes when the port is free; **keeps a mapping that already fronts a live server on another port** (the 3773-primary / 3774-secondary case — asserts no `serve --bg` was spawned and the probe hit `http://127.0.0.1:3773/.well-known/t3/environment`); reclaims a mapping whose target no longer answers; adopts a mapping that already points here; removes the mapping on shutdown while it still points here; leaves a mapping another server has taken over.

`packages/tailscale/src/tailscale.test.ts` gains two cases for the new reader (parses the `Web` entry for a given serve port, defaults the serve port to 443).

Ran, on `main` @ `223ff44`:

```
vp run --filter @t3tools/tailscale --filter t3 typecheck   -> 0 errors
vp test run apps/server/src/tailscaleServe.test.ts         -> 6 passed
vp test run packages/tailscale/src/tailscale.test.ts       -> 16 passed
vp test run apps/server/src/cli/pair.test.ts apps/server/src/server.test.ts -> 190 passed (2 files)
vp lint <changed paths>                                    -> clean
vp fmt --check packages/tailscale apps/server/src          -> clean
```

Per AGENTS.md I did not run repo-wide checks. Not verified: no live end-to-end run against a real tailnet — the tailscale CLI is mocked at the spawner boundary in the tests, so the `serve status --json` shape is exercised against a captured fixture rather than a live daemon.

Deliberately out of scope, noted as a follow-up on the issue: a secondary instance also overwrites `~/.t3/userdata/server-runtime.json` while the recorded pid is still alive, which is what breaks `t3 pair`'s discovery afterwards.

---

Written by Claude Opus 4.6 in Claude Code.

https://claude.ai/code/session_01DXuSCeZKfJRHgukS6BZCZT

### DIFF
--- a/apps/server/src/cli/pair.ts
+++ b/apps/server/src/cli/pair.ts
@@ -31,15 +31,11 @@ import * as Option from "effect/Option";
 import * as References from "effect/References";
 import * as Schema from "effect/Schema";
 import { Command, Flag, GlobalFlag } from "effect/unstable/cli";
-import {
-  FetchHttpClient,
-  HttpClient,
-  HttpClientRequest,
-  HttpClientResponse,
-} from "effect/unstable/http";
+import { FetchHttpClient } from "effect/unstable/http";
 
 import * as EnvironmentAuth from "../auth/EnvironmentAuth.ts";
 import * as ServerConfig from "../config.ts";
+import { type EnvironmentProbeResult, probeEnvironmentDescriptor } from "../environmentProbe.ts";
 import { resolveBaseDir } from "../os-jank.ts";
 import {
   type PersistedServerRuntimeState,
@@ -56,8 +52,6 @@ import {
 } from "../startupAccess.ts";
 import { baseDirFlag, DurationFromString } from "./config.ts";
 
-const WELL_KNOWN_ENVIRONMENT_PATH = "/.well-known/t3/environment";
-const PAIR_PROBE_TIMEOUT = Duration.millis(2_500);
 // Tailscale provisions an HTTPS certificate on the first request to a fresh
 // serve mapping, which can take a few seconds.
 const TAILSCALE_PROBE_ATTEMPTS = 5;
@@ -192,43 +186,6 @@ export const formatPairOutput = (input: {
     ...input.notes.flatMap((note) => ["", `Note: ${note}`]),
     "",
   ].join("\n");
-
-/**
- * Three outcomes, because they drive different decisions: a T3 descriptor
- * (pair with it), nothing answering (safe to configure Tailscale Serve), or
- * something answering that is not a T3 server (do NOT overwrite its mapping).
- */
-type EnvironmentProbeResult =
-  | { readonly _tag: "descriptor"; readonly descriptor: ExecutionEnvironmentDescriptor }
-  | { readonly _tag: "unreachable" }
-  | { readonly _tag: "not-a-t3-server" };
-
-const probeEnvironmentDescriptor = (
-  baseUrl: string,
-): Effect.Effect<EnvironmentProbeResult, never, HttpClient.HttpClient> =>
-  Effect.gen(function* () {
-    const client = yield* HttpClient.HttpClient;
-    const request = HttpClientRequest.get(new URL(WELL_KNOWN_ENVIRONMENT_PATH, baseUrl).toString());
-    const response = yield* client.execute(request).pipe(
-      Effect.timeout(PAIR_PROBE_TIMEOUT),
-      // Transport failure or timeout: nothing (reachable) is listening there.
-      Effect.mapError(() => ({ _tag: "unreachable" }) as const),
-    );
-    // Bad-gateway family means a proxy (Tailscale Serve) answered for a
-    // backend that is gone — a stale mapping, not a live occupant. Treating
-    // it as unreachable lets `t3 pair --tailscale` repair its own mapping
-    // after the server's port changed.
-    if (response.status === 502 || response.status === 503 || response.status === 504) {
-      return { _tag: "unreachable" } as const;
-    }
-    // Anything else that answered HTTP but not with a valid descriptor is
-    // some other service.
-    const descriptor = yield* HttpClientResponse.filterStatusOk(response).pipe(
-      Effect.flatMap(HttpClientResponse.schemaBodyJson(ExecutionEnvironmentDescriptor)),
-      Effect.mapError(() => ({ _tag: "not-a-t3-server" }) as const),
-    );
-    return { _tag: "descriptor", descriptor } as const;
-  }).pipe(Effect.catch((outcome) => Effect.succeed(outcome)));
 
 interface DiscoveredPairTarget {
   readonly baseDir: string;
--- a/apps/server/src/environmentProbe.ts
+++ b/apps/server/src/environmentProbe.ts
@@ -1,0 +1,50 @@
+/**
+ * "Is a T3 Code server answering at this origin?" — shared by `t3 pair`
+ * discovery and by the Tailscale Serve ownership check, which both have to
+ * decide whether an endpoint is safe to take over.
+ */
+import { ExecutionEnvironmentDescriptor } from "@t3tools/contracts";
+import * as Duration from "effect/Duration";
+import * as Effect from "effect/Effect";
+import { HttpClient, HttpClientRequest, HttpClientResponse } from "effect/unstable/http";
+
+const WELL_KNOWN_ENVIRONMENT_PATH = "/.well-known/t3/environment";
+const ENVIRONMENT_PROBE_TIMEOUT = Duration.millis(2_500);
+
+/**
+ * Three outcomes, because they drive different decisions: a T3 descriptor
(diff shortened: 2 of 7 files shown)


shivamhwp:

The ownership check needs to preserve more than a proxy's host and port:

- `parseProxyTarget` drops the protocol and `targetOrigin` always builds HTTP. An existing `https://127.0.0.1:8443` target is therefore probed with the wrong protocol and can be reclaimed after that probe fails. Preserve the protocol for the probe and only adopt the plain-HTTP loopback mapping this server publishes.
- A root `Path` handler or a TCP-only mapping returns `null`, the same value as a free port. Acquisition can overwrite it, and release can disable it without establishing ownership. Unsupported occupied handlers must remain occupied; an unreadable status must skip both acquisition and release.
- The lifted descriptor probe classifies 502/503/504 as a dead target. That works for `t3 pair` probing through the Serve proxy, but acquisition probes the backend directly. An HTTP 503 from that direct target proves a service is listening and should leave its mapping alone. Keep the existing proxy behavior for pairing and distinguish direct probes.

The existing concurrent-claim finding remains: two instances can both read an empty slot before either writes. These parser and probe corrections do not serialize ownership. Claim and release need coordination across processes; another unprotected read still leaves a race.

Issue #10366, opened 2026-09-06
[Bug]: Element picker drops screenshot after five-second capture timeout


gabrielMalonso:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

1. Open T3 Code Desktop Nightly and show a local web app in the browser preview (in this case `http://localhost:5173/`).
2. Start the element picker.
3. Select an element in the page.
4. Enter an annotation and click **Attach**.

### Expected behavior

The annotation and a cropped screenshot of the selected element should be added to the composer.

### Actual behavior

The structured annotation is added, but its screenshot is consistently dropped. T3 Code shows:

> Could not capture the picked element
> The annotation was kept without the screenshot.

The page remains loaded and interactive in the visible preview.

### Impact

Major degradation or frequent failure

### Version or commit

T3 Code Desktop `0.0.39-nightly.20260906.1293`

### Environment

macOS 27.0 (26A5425a), Apple Silicon arm64, visible embedded browser preview, local Vite page

### Logs or stack traces

The desktop trace records `PreviewManager.pickElement` as successful even when the screenshot is dropped. Representative durations from repeated attempts:

```text
PreviewManager.pickElement durationMs=5952.872833 exit=Success
PreviewManager.pickElement durationMs=5146.751917 exit=Success
PreviewManager.pickElement durationMs=6606.606333 exit=Success
```

These durations line up with `ANNOTATION_SCREENSHOT_TIMEOUT = "5 seconds"` in `apps/desktop/src/preview/Manager.ts`. The underlying `webContents.capturePage()` failure/timeout is collapsed into `screenshotFailed: true`, while the detached warning (`preview annotation screenshot timed out`) does not appear in `desktop.trace.ndjson`. As a result, the trace confirms the delayed pick but does not preserve the precise capture failure.

This started after a recent Nightly update. Commit `19c97ea56` added the five-second timeout and user-visible fallback for a wedged guest compositor, but the annotation screenshot now fails repeatedly in this otherwise responsive visible tab.

### Screenshots, recordings, or supporting files

Selected element remains visible and responsive after the crop is dropped:

<img width="739" height="436" alt="Selected preview element remains outlined after screenshot capture fails" src="https://github.com/user-attachments/assets/0f5e3563-2610-4051-aae2-40dc7e288449" />

Error toast shown after clicking **Attach**:

[...]

PR #6242, opened 2026-08-11
fix(web): attach preview annotation screenshots in the desktop app

glauber-sampaio:
Preview annotations never attached their screenshot in the desktop app. The crop was captured fine, but converting it into a composer attachment used `fetch()` on a `data:` URL, and the desktop renderer's CSP allows `data:` for `img-src` and `font-src` but not for `connect-src` — the directive that governs fetch. The call was blocked and swallowed by an empty catch, so the image was silently dropped.

The prompt still told the agent "The attached screenshot is the annotated preview crop", because that sentence is built from `annotation.screenshot` rather than from the attachment that actually landed. The agent was told to look at an image it never received.

Draw annotations made it visible, since strokes carry no element context and the image is the whole payload. Select and region picks were losing their screenshots too, but still read well from their `<element_context>` text.

Decode the crop with the existing `dataUrlToFile` helper instead of fetching it, and clear `screenshot` when the attachment does not survive so a failed conversion stops the prompt from promising an image.

### Before

The annotation arrives as metadata only, and the agent says so.

Before

### After

The crop is attached and the agent can read it.

After

Verification:
- `vp test run apps/web/src/lib/previewAnnotation.test.ts apps/web/src/lib/imageCompression.test.ts apps/web/src/components/preview/PreviewView.test.tsx` (28 tests)
- `vp run --filter @t3tools/web typecheck`
- targeted `vp lint` and `vp fmt` for changed files
- Verified in the desktop app, before and after, as above

Model: Claude Opus 5 | Harness: Claude Code in T3 Code

---

> [!NOTE]
> **Medium Risk**
> Changes how preview annotation images are converted and attached to the composer/agent prompt, which can affect what the agent receives. Scoped bug fix with focused tests, but regressions would silently drop or misadvertise screenshots.
> 
> **Overview**
> **Fixes desktop preview annotation screenshots never reaching the composer.** Conversion previously `fetch`ed a `data:` URL, which the desktop CSP blocks via `connect-src`, so the image was silently dropped while the prompt still claimed a screenshot was attached.
> 
> `previewAnnotationScreenshotFile` now **decodes the crop in place** with `dataUrlToFile` instead of fetching. If conversion fails or returns null, `PreviewView` **clears `screenshot` on the annotation** so the prompt no longer advertises an image the agent never received.
> 
[...]

PR #10374, opened 2026-09-06
fix(web): keep screenshots when attaching preview annotations

Neel2107:
Preview annotations captured their screenshot, then dropped it because desktop CSP blocks `fetch(data:...)`. Decode the image with the existing `dataUrlToFile` helper and remove the redundant conversion timeout. Failed conversions still keep the annotation without promising an attached image.

Overlaps with #6242. Found while investigating #10366; the separate five-second Electron capture timeout is unchanged and was not reproduced locally.

Verified in desktop on macOS: the crop now attaches and opens correctly; normal screenshots still work. All 43 focused tests and web typecheck passed. Targeted lint reported only existing warnings; formatting passed.

### Before / after

| Before: crop dropped | After: crop attached |
| --- | --- |
| Before | After |

Created with GPT-6 in Codex in T3 Code

### DIFF
--- a/apps/web/src/components/preview/PreviewView.test.tsx
+++ b/apps/web/src/components/preview/PreviewView.test.tsx
@@ -342,7 +342,7 @@ describe("PreviewView navigation", () => {
     mocks.closePictureInPicture.mockClear();
     mocks.pickElement.mockReset();
     mocks.capturePreviewAnnotationScreenshot.mockReset();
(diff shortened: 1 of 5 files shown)


juliusmarminge:
Superseded by #12636, which already fixed #12265 and merged (same desktop CSP / dataUrlToFile fix; overlaps with #6242). Closing this PR as leftover hygiene.

PR #10377, opened 2026-09-06
fix(web): stop drawing background tasks as agents in the timeline

### LINKED ISSUE #10375: [Bug]: Foreground Bash calls promoted to local_bash tasks render with the Bot (subagent) icon while the Agents panel correctly excludes them
### Before submitting

### Area

apps/web

### Summary

When a Claude Code **foreground** Bash call runs longer than ~2 s, Claude Code registers it as a `local_bash` task and emits `task_started` / `task_completed` system messages for it. T3 ingests those as `task.started` / `task.completed` activities, correctly stamps them `agentKind: "background"`, and correctly keeps them **out** of the Agents panel. But the timeline renders the resulting work row with the **Bot (robot) icon**, because the icon picker keys on "has a `taskId`" rather than on the `agentKind` stamp. The user sees a robot row in the transcript, looks for the subagent in the Agents panel, and finds nothing.

So the two surfaces disagree: the Agents panel says "not an agent", the timeline row says "agent". The comment on the icon rule even says these rows are "subagent lifecycle rows", which is only true for `agent`-kind tasks.

Related but distinct: #9107 is about background-from-start shells being invisible. This report is about foreground shells that get task-promoted showing agent chrome they should not have.

### Steps to reproduce

1. Open a thread using the Claude Code provider.
2. Have the agent run any foreground Bash command that takes more than ~2–3 s, e.g. `sleep 5 && exit 2` (a non-zero exit makes the tint obvious, but the icon is wrong either way).
3. Watch the timeline after the command finishes.

### Expected behavior

The task lifecycle row for a `local_bash` task should render with the same terminal/command chrome as the Bash call it belongs to (or be folded into that Bash row). The Bot icon should be reserved for rows whose task is classified `agentKind: "agent"`, matching what the Agents panel shows.

### Actual behavior

A separate collapsed row appears with the Lucide `Bot` icon, titled with the Bash call's `description`, and with the same description repeated as the monospace detail line. On a failed exit the icon is tinted with the failed-tool colour, so it reads as an orange/red robot. The Agents panel shows no agent for it.

Activity sequence recorded for one such call (ids elided), all sharing one `toolUseId`:

```
tool.started    itemType=command_execution  title="Command run"            (Bash)
tool.updated    itemType=command_execution  status=inProgress
task.started    taskType=local_bash agentKind=background title=<Bash description>  detail=<Bash description>   (~3 s after tool.started)
task.completed  taskType=local_bash agentKind=background status=failed tone=error
tool.updated    itemType=command_execution  status=failed
tool.completed  itemType=command_execution  status=failed
```

The Bash row itself renders fine with the terminal icon. The extra robot row comes from the `task.*` pair.

### Where the icon decision comes from (main @ 223ff44)

- Icon picker falls through to `"bot"` for any work entry with a `taskId`, regardless of `agentKind`:
  https://github.com/pingdotgg/t3code/blob/223ff4490f764a74ff911589e97b9bbcd595fee8/apps/web/src/components/chat/MessagesTimeline.tsx#L3062-L3065
- Group summary kind does the same (`entry.taskId` → `"agent-tool"` → `"bot"`):
  https://github.com/pingdotgg/t3code/blob/223ff4490f764a74ff911589e97b9bbcd595fee8/packages/client-runtime/src/work-log/presentation.ts#L614
  https://github.com/pingdotgg/t3code/blob/223ff4490f764a74ff911589e97b9bbcd595fee8/apps/web/src/components/chat/MessagesTimeline.tsx#L2150-L2151
- The server-side classification already exists and puts `local_bash` on the non-agent list:
  https://github.com/pingdotgg/t3code/blob/223ff4490f764a74ff911589e97b9bbcd595fee8/packages/contracts/src/providerRuntime.ts#L595-L624
- The client already reads that stamp for the Agents panel and marks the work entry `isBackgroundTask`, but nothing in the icon path consults it:
  https://github.com/pingdotgg/t3code/blob/223ff4490f764a74ff911589e97b9bbcd595fee8/packages/client-runtime/src/state/subagentRuntime.ts#L119-L121
  https://github.com/pingdotgg/t3code/blob/223ff4490f764a74ff911589e97b9bbcd595fee8/apps/web/src/session-logic.ts#L620-L622

Claude Code side, for context: a foreground Bash that passes ~2 s of wall time is registered as a `local_bash` task (`isBackgrounded: false`) so it can offer backgrounding; the `task_started` payload's `description` is the Bash tool's `description` argument, which is why title and detail are the same string.

### Suggested fix

In `workEntryIconName` and `toolGroupSummaryKind`, treat `entry.taskId` as agent chrome only when the entry is not `isBackgroundTask` (or when the payload's `agentKind === "agent"`). Background task rows could use `"terminal"` when `taskType` is in `MONITOR_TASK_TYPES`, or simply fall back to the tone icon. Alternatively, fold `local_bash` task rows into the parent Bash row keyed by `toolUseId`, since they carry no information the Bash row does not already have.

### Impact

Cosmetic issue

### Version or commit

Server: `t3` 0.0.39-nightly.20260905.1290. Source references above are against `main @ 223ff44`.

### Environment

Linux host running `t3 serve`, accessed from the web UI in a desktop browser. Claude Code provider (CLI 2.1.263), model claude-fable-5-1.

### Workaround

None needed; expanding the row shows it is just the Bash description, and the Agents panel is correct. The confusion is purely the icon.

Mnigos:
Fixes #10375

## Problem

When a foreground Bash call in a Claude thread runs longer than a couple of seconds, Claude Code registers it as a `local_bash` task and T3 records `task.started` / `task.completed` rows for it. The server stamps those rows `agentKind: "background"` and the Agents panel correctly leaves them out, but the timeline gave the row the Bot icon because the icon rule only asked "does this entry have a `taskId`?". The two surfaces disagreed: the transcript showed a robot, the Agents panel showed no agent. A failed command made it worse, since the robot got the failed-tool tint.

## Fix

- `toolGroupSummaryKind` (shared by web and mobile) and the web `workEntryIconName` treat a `taskId` as agent chrome only when the entry is not a background task. Background task rows fall back to their tone icon, so a completed promoted Bash shows the ordinary work icon and a failed one the error icon, next to the terminal row of the Bash call itself.
- `isBackgroundTask` moves from the web's private derived entry onto `WorkLogEntry`, which the timeline rows are typed as; the shared presentation entry gains the same optional flag. The server stamp and the Agents panel filter are unchanged.

## Verification

- `presentation.test.ts`: a subagent row keeps `agent-tool`; a background task row (alone or repeated) does not.
- Web `session-logic` and `MessagesTimeline.logic` tests, client-runtime, web and mobile typecheck, lint.
- Real client: a Claude Haiku thread on main and on this branch, each asked to run `sleep 5 && exit 2` through the Bash tool. Both produce the same activity pair (`task.started` / `task.completed`, `taskType: local_bash`, `agentKind: background`) next to the Bash `tool.*` rows. On main the task row renders with `lucide-bot`; on this branch it renders with the error tone icon, and the Bash row keeps its terminal icon in both.

## Before

main at `223ff4490`, Claude Haiku, failed `sleep 5 && exit 2` promoted to a `local_bash` task: the task row wears the Bot icon (`lucide-bot`) although the Agents panel lists no agent.

Before: task row with the Bot icon

## After

This branch, same reproduction: the task row uses the error tone icon (`lucide-circle-alert`); the Bash row keeps its terminal icon in both.

After: task row with the error icon

Both captures are Playwright screenshots at 2x of the dev web client with the fold expanded (full-window versions: [before](https://raw.githubusercontent.com/Mnigos/t3code/pr-assets/task-row-icon/10377-before-full.png), [after](https://raw.githubusercontent.com/Mnigos/t3code/pr-assets/task-row-icon/10377-after-full.png)).

Implemented with Claude Code (Claude Fable 5).

### DIFF
--- a/apps/web/src/components/chat/MessagesTimeline.tsx
+++ b/apps/web/src/components/chat/MessagesTimeline.tsx
@@ -3060,7 +3060,9 @@ function workEntryIconName(workEntry: TimelineWorkEntry): WorkEntryIconName {
   }
 
   // Subagent lifecycle rows (grouped by taskId) get agent identity chrome.
-  if (workEntry.taskId) {
+  // Background tasks carry a taskId too but are not agents: the Agents panel
+  // leaves them out, so the timeline must not dress them up as one.
+  if (workEntry.taskId && !workEntry.isBackgroundTask) {
     return "bot";
   }
 
--- a/apps/web/src/session-logic.ts
+++ b/apps/web/src/session-logic.ts
@@ -89,6 +89,11 @@ export interface WorkLogEntry {
     workflowId: string | null;
     agentTaskIds: ReadonlyArray<string>;
   };
+  /**
+   * Shell/monitor/plan tasks: ordinary work-log rows that carry a taskId but
+   * are not agents, so they get neither spawn CTAs nor agent chrome.
+   */
+  isBackgroundTask?: boolean;
 }
 
 const workLogCollapseKey = Symbol();
@@ -98,8 +103,6 @@ interface DerivedWorkLogEntry extends WorkLogEntry {
   [workLogCollapseKey]?: string;
   toolCallId?: string;
   isWorkflowCoordinator?: boolean;
-  /** Shell/monitor/plan tasks: ordinary work-log rows, never spawn CTAs. */
-  isBackgroundTask?: boolean;
 }
 
 const derivedWorkLogEntryByActivity = new WeakMap<
--- a/packages/client-runtime/src/work-log/presentation.test.ts
+++ b/packages/client-runtime/src/work-log/presentation.test.ts
@@ -409,6 +409,31 @@ describe("browser group summaries", () => {
   });
 });
 
+describe("task group summaries", () => {
+  const agentTask: WorkLogPresentationEntry = {
+    label: "Started agent",
+    taskId: "task-1",
+    tone: "tool",
+  };
+  // A foreground Bash call Claude promoted to a local_bash task: same row
+  // shape as a subagent, stamped background by the server.
+  const backgroundTask: WorkLogPresentationEntry = {
+    label: "Run the test suite",
+    taskId: "task-2",
+    isBackgroundTask: true,
+    tone: "tool",
+  };
+
+  it("keeps agent chrome for subagent lifecycle rows", () => {
+    expect(toolGroupSummaryKind([agentTask])).toBe("agent-tool");
+  });
+
+  it("does not present a background task as an agent", () => {
+    expect(toolGroupSummaryKind([backgroundTask])).toBe("tone-tool");
+    expect(toolGroupSummaryKind([backgroundTask, backgroundTask])).toBe("tone-tool");
+  });
+});
+
 describe("command work-log details", () => {
   it("extracts Claude result blocks and projected output", () => {
     expect(
--- a/packages/client-runtime/src/work-log/presentation.ts
+++ b/packages/client-runtime/src/work-log/presentation.ts
@@ -32,6 +32,8 @@ export interface WorkLogPresentationEntry {
   readonly toolLifecycleStatus?: string;
   readonly sourceActivityKind?: string;
   readonly taskId?: string;
+  /** Set on task rows the server classified as background work rather than an agent. */
+  readonly isBackgroundTask?: boolean;
   readonly toolSource?: ToolActivitySource;
 }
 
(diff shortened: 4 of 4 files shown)


amiralibg:
Independent verification of this fix with the desktop client (Electron dev build, Claude Opus 5, \`sleep 5 && exit 2\` run in the foreground so Claude promotes it to a \`local_bash\` task).

**Before** (pre-fix build): the promoted task row wears the Bot icon with the failed tint, duplicating the Bash description, while the Agents panel lists no agent.

Before: task row with the Bot icon

**After** (branch carrying the equivalent guard — \`taskId\` counts as agent chrome only when the entry is not a background task): the task row renders the same failure chrome as the Bash row itself, and the group header correctly reads "Ran 1 command" with the terminal icon instead of agent chrome.

After: task row matching the Bash row chrome

One deliberate difference in my variant worth a glance: for a *non-failed* background task row I return the \`terminal\` icon explicitly, while this PR falls through to the tone icon. In the failed repro above both render identically (the \`PlainWorkEntryRow\` failure override wins either way), so not a blocker — just flagging since the issue text asks for "the same terminal/command chrome as the Bash call it belongs to".

Verified with Muse Spark via OpenCode.

PR #10378, opened 2026-09-06
feat(web): name the drop action while dragging sidebar threads

SunkenInTime:
Follow-up to the merged sidebar drag work, now based on main.

**Problem.** While dragging a thread across sections, the destination divider lit up in the accent color, which reads as an insertion line rather than a state change, and the lifted row only said "→ Section". On first use it was not clear what a drop would do. The lifted row also let the rows beneath it show through: its z-index sat on an unpositioned list item, so neighbouring rows' content painted over it.

**Fix.**
- The lifted row shows a verb badge with the matching icon: **Pin**, **Unpin**, **Settle**, **Un-settle**, or **Wake**, derived from source and destination by a small pure helper, `resolveSidebarDropVerb`. Same-section reorders show no badge. The verbs match the row actions and context menu, so nothing new to learn.
- Section labels render at a readable neutral for the whole drag, and the section under the lifted row takes the accent. The verb badge is the only signal that the drop changes state.
- The lifted row is opaque, shadowed, and positioned so its z-index applies, so the rows beneath it never show through. The opaque base is stacked as a background image because the hover color stays applied while dragging and is translucent in dark themes. Boundary labels paint above the lifted row, so the Pinned label stays readable when the first thread is dropped into an empty pinned section.
- A pinned thread keeps its pin marker only while it stays inside Pinned. Everywhere else the badge carries the icon, so the pin is never shown twice.
- A dropped thread glides from under the pointer into its committed slot instead of snapping, and a cancelled or rejected drop glides back to where it came from. The list motion takes every row's visual position at release and glides each into its committed slot, so the label gaps close with the same motion as the drop.
- Section labels get their own space. The sorting strategy opens one label of height below each pinned boundary for the whole drag and the labels render inside that gap, so they never sit on a row. Every drag projects through the strategy so the space exists for same-section reorders too. The markers stay zero height at rest, so nothing is reserved and pickup measurements are unchanged.

Tests: two cases for the verb mapping in `Sidebar.logic.test.ts`. Doc paragraph in `docs/user/thread-sidebar.md` updated.

**Before**

| Active thread over Pinned | Row beneath bleeding through |
| --- | --- |
|  |  |

**After**

| Pin | Unpin |
| --- | --- |
|  |  |

| Reorder inside Pinned (no badge, pin marker stays, label above the card) | Settle |
| --- | --- |
|  |  |

| Last pinned thread leaves Pinned | No pins: drag to the top edge |
| --- | --- |
|  |  |

| No pins: reorder inside Active (labels stack only during the drag) | |
| --- | --- |
|  | |

**Drop motion** (unpin by drop, pin by drop, then a cancelled drag)

Screen recording of the full drag flow: https://cap.so/s/b3r1wwtkwxw0x1s

### Drag recording

[Pin and Settle badges during a drag](https://codex-file-host.shawnadedeji.workers.dev/files/pr-10378/99afd298451a-t3-pr-10378-drag.mp4), verified in an isolated web client using renamed UI fixtures. The drag is cancelled at the end.

Rebased onto main. Web typecheck and all 216 focused sidebar tests pass.

### Final verification

Fixed the unused export reported by CI; no behavior changed in this fix. Web typecheck, all 216 sidebar tests, scoped Knip, and targeted lint pass. React Doctor reports only the two existing sidebar complexity warnings. Manual source review found no additional blockers.

Verified unpin, pin, and cancel in isolated web clients using the same fixture, viewport, and theme. Base `9a47c7bd405`; head `5dc842ea269`. This applies to web and desktop’s shared sidebar; mobile, contracts, and server behavior are unchanged.

| Before | After |
| --- | --- |
| Before: destination badge | After: Unpin action badge |

[Before recording](https://gh-file-drop-api-prod-mi5fy3sowv63ufte.pinglabs.workers.dev/f/0985c89f2c0b2a47/base-drag.mp4) · [After recording](https://gh-file-drop-api-prod-mi5fy3sowv63ufte.pinglabs.workers.dev/f/123f70ba908c255e/head-drag.mp4)

All executed CI jobs pass. Macroscope approvability is neutral because its correctness review is unavailable under the workspace spending limit; Bugbot also reports a spending limit. There are no unresolved review threads.

### DIFF
--- a/apps/web/src/components/Sidebar.drag.test.ts
+++ b/apps/web/src/components/Sidebar.drag.test.ts
@@ -245,7 +245,9 @@ describe("sidebar drag projection", () => {
       settledExpanded: true,
     });
     const args = layout(pinned, active, over);
+    // dnd-kit moves the lifted row by the pointer delta, so only peers matter.
     for (let index = 0; index < pinned.length; index += 1) {
+      if (index === args.activeIndex) continue;
       expect(strategy({ ...args, index })).toEqual(verticalListSortingStrategy({ ...args, index }));
     }
   });
@@ -282,6 +284,7 @@ describe("sidebar drag projection", () => {
     });
     const args = layout(items, active, over);
     for (let index = 0; index < items.length; index += 1) {
+      if (index === args.activeIndex) continue;
       expect(strategy({ ...args, index })).toEqual(verticalListSortingStrategy({ ...args, index }));
     }
   });
@@ -304,11 +307,11 @@ describe("sidebar drag projection", () => {
   });
 
   it.each([
-    [sidebarMarkerId("pinned-divider"), 16, 16],
-    ["a1", -67, 16],
-    ["a2", -67, -67],
+    [sidebarMarkerId("pinned-divider"), 0, 0],
+    ["a1", -83, 0],
+    ["a2", -83, -83],
   ] as const)(
-    "opens the active pointer slot over %s and keeps a label of height for the emptied pins",
+    "opens the active pointer slot over %s without adding an empty pinned row",
     (over, a1Offset, a2Offset) => {
       const items = [
         pinnedHeader,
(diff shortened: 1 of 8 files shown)


SunkenInTime:
Re the empty band above the first thread with no pins: that was a reserved label height at rest, added so the Pinned and Active labels could stack during a drag without the list shifting at pickup. It's gone as of b91fa88f9 (`fix(web): reserve no space for an empty pinned section at rest`). The gap now only exists during a drag, produced by the sorting strategy, which also projects active reorders when there are no pins. At rest the pinned header is back to zero height. Screenshot of the no-pins reorder is in the body.

PR #10393, opened 2026-09-06
fix(web): go back with Escape from Usage and Pull Requests

### LINKED ISSUE #10388: [Bug]: Escape does not return to the previous thread from Pull Requests or Usage
### Before submitting

- [x] I searched existing issues and did not find a duplicate for the Pull Requests page.

### Area

apps/web

### Steps to reproduce

1. Open a thread in T3 Code.
2. Navigate to the Pull Requests page or the Usage page.
3. With no dialog or popover open, press Escape.
4. Repeat for the other page.
5. Compare with opening Settings from a thread and pressing Escape.

### Expected behavior

Escape should return to the thread I was working in, consistent with Settings. If a dialog or popover is open, Escape should dismiss it first without leaving the page.

### Actual behavior

Escape does not return from either Pull Requests or Usage to the previous thread. The same action works from Settings, making keyboard navigation inconsistent.

### Impact

Minor bug or occasional failure. Returning to the thread requires manual navigation.

### Related work

- #1350 requested the existing Escape-to-go-back behavior in Settings.
- Open PR #10158 already adds Escape navigation to Usage. This report records the current inconsistency on both pages; the Usage portion overlaps with that PR, while Pull Requests remains uncovered.

### Workaround

Select the previous thread in the sidebar.

Neel2107:
Escape returned from Settings but did nothing on Usage and Pull Requests. Reuse the shared back-navigation hooks from #10158 across all three pages and the sidebar Back button. Editors and popovers that consume Escape still dismiss first.

Fixes #10388. Overlaps with #10158 for Usage and the shared hooks; this also covers Pull Requests, without adding Usage shortcuts.

Validated with 17 focused tests, web typecheck, formatting, and a desktop build. Targeted lint reported existing warnings only. Reporter tested both pages in the local macOS desktop app.

### Desktop verification

https://github.com/user-attachments/assets/b2968fd4-1bb2-49a9-a669-ea013edfc0a3

Created with GPT-6 in Codex in T3 Code

### DIFF
--- a/apps/web/src/components/sidebar/SidebarChrome.tsx
+++ b/apps/web/src/components/sidebar/SidebarChrome.tsx
@@ -6,8 +6,9 @@ import {
 } from "lucide-react";
 import type { ReactNode } from "react";
 import { memo, useCallback } from "react";
-import { Link, useCanGoBack, useLocation, useNavigate } from "@tanstack/react-router";
+import { Link, useLocation, useNavigate } from "@tanstack/react-router";
 
+import { useNavigateBack } from "../../hooks/useNavigateBack";
 import { useEnvironmentIdentificationMode } from "../../hooks/useSettings";
 import { cn } from "../../lib/utils";
 import { useEnvironments } from "../../state/environments";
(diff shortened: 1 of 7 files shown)


Neel2107:
Closing in favor of #10158, which now covers Escape navigation on both Usage and Pull Requests. Thanks for adding the missing surface!

The regression tests in `apps/web/src/hooks/useNavigateBack.test.tsx` here may be useful to carry over. #10388 should remain open until the fix merges.

PR #10409, opened 2026-09-06
feat(usage): read cursor, opencode, and antigravity history

maria-rcks:
cursor, opencode, and antigravity were absent from usage because the scanner and clients only supported claude, codex, and grok. this adds read-only opencode sqlite/legacy-json readers, antigravity protobuf/sqlite readers, and cursor account history using the existing file-based CLI login, covering headless T3 sessions and desktop usage; empty windows and overlapping page boundaries are handled, and cursor uses account history exclusively to avoid counting local records twice. subscription quota support is unchanged.

source-attributed buckets avoid double counting overlapping data directories and shared cursor accounts across environments, while web, desktop, and mobile can display all six providers. verified with focused reader/service/aggregation/merge/chart tests, scoped lint and typechecks, real opencode history, 118 real cursor usage records with all three headless events verified and a real empty-history query, and browser checks for cost, tokens, refresh, light/dark themes, and narrow layout; native mobile rendering is unverified because this machine has no android sdk/emulator.

current real history: cursor account usage over 90 days, including three headless events whose token counts were checked against the endpoint response. the reader returned 118 token-bearing records across 38 conversations. opencode remains visible from its real local history. cost/tokens switching and refresh preserve these totals.

real cursor account history including headless usage

real cursor account history at narrow width

earlier layout captures below use synthetic demo histories for all three added providers, read through local readers: seven sessions each, 3.51m tokens total. cost/tokens switching and refresh were exercised with stable totals. the cursor notice in these earlier demo captures predates account-history support.

demo data: cursor, opencode, and antigravity populated in light mode

demo data: all three added harnesses in dark mode

demo data: all three added harnesses at a narrow viewport

matching before/after below uses fixed copies of real history; opencode's 428k tokens appear after the reader is enabled.

before: real codex history only

after: same real history with opencode included

model: `gpt-6-astra`; harness: codex.

### DIFF
--- a/apps/mobile/src/features/usage/UsageRouteScreen.tsx
+++ b/apps/mobile/src/features/usage/UsageRouteScreen.tsx
@@ -81,6 +81,21 @@ export function UsageRouteScreen() {
     selectedEnvironmentIds,
   );
   const limits = useRefreshLimits(selectedEnvironmentIds);
+  const sourceMessages = [
+    ...new Set(
+      selectedEnvironments.flatMap(
+        (environment) =>
+          environment.summary?.sources.flatMap((source) =>
+            source.message &&
+            (source.status === "partial" ||
+              source.status === "failed" ||
+              source.fingerprint.provider === "cursor")
+              ? [source.message]
+              : [],
+          ) ?? [],
+      ),
+    ),
+  ];
 
   const days = useMemo(
     () => enumerateDays(window.sinceDay, window.untilDay),
@@ -295,6 +310,11 @@ export function UsageRouteScreen() {
                 </Text>
               ) : (
                 <>
+                  {sourceMessages.map((message) => (
+                    <Text key={message} className="text-sm text-foreground-muted">
+                      {message}
+                    </Text>
+                  ))}
                   <ChartCard
                     merged={merged}
                     days={chartDays}
--- a/apps/mobile/src/features/usage/usageProviders.ts
+++ b/apps/mobile/src/features/usage/usageProviders.ts
@@ -5,12 +5,22 @@ import { useAppearancePreferences } from "../settings/appearance/AppearancePrefe
  * Series and table order. The chart stacks providers from the bottom in this
  * order, so it also fixes which band sits on top of the bars.
  */
-export const PROVIDER_ORDER: readonly UsageProviderKind[] = ["codex", "claude", "grok"];
+export const PROVIDER_ORDER: readonly UsageProviderKind[] = [
+  "codex",
+  "claude",
+  "grok",
+  "cursor",
+  "opencode",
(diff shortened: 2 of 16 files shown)


alvarohulse (commented):
Saw this and I also want the feature. Tested the changes and reviewed. New to the codebase, but the changes look sane.

The colours for the lines on the graphs are okay too. I would put black or dark grey for OpenCode, and blue or green for Antigravity.

Only have Cursor - Metrics don't change between environments, though. Seems to be a limitation of the API method being used. Unsure if there's a better way.

Good work, and good luck!

andrewcai8:
+1 want this for cursor!

squidllee:
hope this gets merged soon because my main is opencode

AdreotA (commented):
## Tested review: OpenCode reader checked against a live store

I ran `parseOpenCodeMessage` (+ the `dedupeKey` set) verbatim from this branch over a real 134 MB `~/.local/share/opencode/opencode.db` (OpenCode 1.18.18, 760 `message` rows, 700 assistant, 688 with tokens, 52 sessions, all turns routed to OpenRouter):

```
records 688   uncachedInput 2,110,352   cachedInput 76,603,812
cacheCreation 3,796,001   output 502,301   reasoning 258,627
costUsd 17.6783   unpricedRows 0
```

**Your reasoning accounting is the correct one.** 173 of 688 rows have `reasoning > output` (e.g. `input=7065 output=32 reasoning=70`), which proves reasoning is not a subset of `tokens.output` in this store — `outputTokens: tokens.output + reasoningTokens` is what keeps the contract's "reasoning subset of output" invariant honest. #10926 clamps with `Math.min(output, reasoning)` and consequently reports `output 243,674 / reasoning 84,484` on the same DB, i.e. it silently drops 174,143 tokens. Worth cross-referencing there.

Also good, and missing from #10926: reading `message` ∪ `session_message`, globbing `opencode-*.db`, the `busy_timeout` pragma, and the `dedupeKey` (`opencode:<id>`) that makes a duplicated row across those sources collapse instead of double-counting. For context, the shipped 1.18.18 binary creates `session_message` (with its own `session_message_time_created_idx`) and ships the migrations `…_reset_v2_session_state` / `…_session_message_projection_order` / `…_event_sourced_session_input`, so that table is the path a future build will project turns into.

One scoping note, not a criticism of the code: this PR also carries Cursor and Antigravity readers plus a contract change (`sourcePath` on buckets, `usageMerge`). CONTRIBUTING asks for small focused changes and there are separate in-flight PRs for those providers, so a reviewer may well ask to split the OpenCode part out — and that OpenCode part alone is the piece that clears the reported "Usage page shows $0 for OpenCode" gap.

PR #10438, opened 2026-09-07
fix(server): preserve due tasks when time formatting changes

CouchRiv:
## What Changed

Fixed-time schedules now compare parsed hours and minutes, so saving `09:00` instead of `9:00` preserves the pending occurrence.

## Why

Both formats are accepted by the schedule contract. The current string comparison treats them as a schedule change and recalculates the next run. If today's occurrence is already due but has not dispatched, that save can postpone it until tomorrow.

This targets the V2 branch in #2829. It reuses the existing time parser and leaves actual hour, minute, weekday and interval changes distinct.

## Validation

Rebased and revalidated on V2 `89459e38`, where migration 051 now creates the scheduled-task schema; the service test imports that migration.

- Four new regression cases fail before the fix; all 15 focused scheduler and contract tests pass afterward.
- The service test uses real input decoding, in-memory SQLite and a controlled clock. It verifies the preserved persisted deadline and that changing the time to `09:30` still reschedules.
- Server typecheck and targeted lint, formatting and diff checks pass.

## Checklist

Implemented with GPT-6 in the Codex harness; rebased and revalidated with Claude Opus 5 in Claude Code.


saphid:
Rebased adaptation of this fix for the current V2 base is in #11590 — same semantic time comparison, service test moved onto the full in-memory migration chain. Crediting this proposal there.

juliusmarminge:
Stale duplicate of #11590 — same parsed-time comparison fix on the current branch. Thanks!

PR #10483, opened 2026-09-07
fix(mobile): release initial scroll target after dragging

juliusmarminge:
Opening a long mobile thread, swiping slightly toward earlier content, and immediately focusing the composer could pull the list back toward the bottom. The initial scroll target stayed alive within half a viewport even after the drag disabled following.

Only retain that near-end allowance before a user drag. Preserve the existing at-end behavior and apply the same guard to both native bundles in the LegendList patch. The lockfile update contains only the patch hash.

Verified on an iPhone 17 Pro simulator running iOS 26.5 with disposable fixture data. Before: a 314-point correction after the drag. After: no correction in the matching swipe-and-focus sequence. The recordings show the exact guard shipped here, tested as a temporary control during investigation. Android was not run; web and desktop bundles are unchanged.

- 32 focused mobile follow/shared chat-list tests passed.
- Installed-package behavior checks passed in both native bundles: release after a drag, retain before a drag, retain when still at the end.
- Frozen offline install, native bundle syntax checks, and diff whitespace checks passed.

Before:

Composer opening pulls the list toward the bottom

After:

Reader position stays in place when the composer opens

Model: GPT-6. Harness: Codex.

> [!NOTE]
> ### Fix `legend-list` initial scroll target retention after drag
> Updates the patched list scrolling logic in [@legendapp__list@3.3.5.patch](https://github.com/pingdotgg/t3code/pull/10483/files#diff-dbe8c8885e350c403aeb082d349bce6e639f64b8e00b190ddcbf69b70ac0a11e) so the near-end inset-list condition also requires that `state.didUserDrag` is false. Bottom-aligned initial scroll targets are no longer kept alive solely because an inset list is near its end after the user drags; they are still kept alive when the list reports it is at the end.
> - Risk: none identified.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a3c2828.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Added support for leading content insets across native, animated, and web lists.
  - Improved keyboard-aware list behavior, including adjustable composer height and animated inset changes.
  - Added configurable inset compensation for keyboard-aware lists.

- **Bug Fixes**
  - Improved end-of-list positioning, initial scrolling, anchored content behavior, and negative-scroll handling.
  - Reduced unwanted position transitions during active scrolling.
  - Improved iOS maintain-visible-content behavior and web scroll adjustments.
  - Enhanced animated content-size and end-padding updates.

### DIFF
--- a/patches/@legendapp__list@3.3.5.patch
+++ b/patches/@legendapp__list@3.3.5.patch
@@ -570,7 +570,7 @@ index b3c5a306b293f797a8b338adfca3060c0f6db22b..ddca4bcd8a5dc5863cb65e03b6cfac00
 -    const shouldKeepEndTargetAlive = isRetargetableBottomAlignedInitialScrollTarget(initialScroll) && peek$(ctx, "isAtEnd");
 +    const endTargetDistanceFromEnd = getContentSize(ctx) - state.scroll - state.scrollLength - getContentInsetEnd(ctx);
 +    const isNearEndForInsetList = getContentInsetStartAdjustment(ctx) > 0 && Number.isFinite(endTargetDistanceFromEnd) && endTargetDistanceFromEnd <= state.scrollLength * 0.5;
-+    const shouldKeepEndTargetAlive = isRetargetableBottomAlignedInitialScrollTarget(initialScroll) && (peek$(ctx, "isAtEnd") || isNearEndForInsetList);
++    const shouldKeepEndTargetAlive = isRetargetableBottomAlignedInitialScrollTarget(initialScroll) && (peek$(ctx, "isAtEnd") || !state.didUserDrag && isNearEndForInsetList);
      if (!shouldKeepEndTargetAlive) {
        if (shouldPreserveInitialScrollForFooterLayout(initialScroll)) {
          clearPendingInitialScrollFooterLayout(ctx, {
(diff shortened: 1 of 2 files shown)


juliusmarminge:
Checked the watchdog-reset finding against the installed native bundle and the reproduced sequence. It does not block this fix:

- The reported drag happens after `didFinishInitialScroll` is true. `clearFinishedBootstrapInitialScrollTargetIfMovedAway` now clears the preserved target on the ensuing native scroll event.
- `retargetActiveInitialScrollAtEnd` takes `schedulePreservedEndAnchorCorrection` when initial scrolling is finished. It does not call `dispatchInitialScroll` or restart the settle watchdog. With the target cleared, `getPreservedEndAnchorOffsetDiff` returns without a correction.
- `startInsetEndSettleWatchdog` has one caller, `dispatchInitialScroll`. The ordinary advance path rejects finished sessions in `getAdvanceableInitialScrollState`. An inset report therefore cannot reset this drag flag through the proposed retry sequence.

The targeted behavior checks exercise the finished-session drag followed by inset retargeting in both native bundles; the iOS recording confirms zero correction with this guard, versus 314 points before. The existing watchdog reset during unfinished initial-scroll retries is outside this change and is not evidence that the confirmed completed-session fix can be bypassed. No lifecycle redesign is included without a concrete failing sequence for that separate case.

The generated walkthrough describes the entire pre-existing patch. This PR changes only the near-end retention condition in the two native bundles and the corresponding lockfile hash.

PR #10511, opened 2026-09-07
fix(desktop): route preview clipboard shortcuts to guest

kfpv:
## Problem

Physical Cmd/Ctrl+C, Cmd/Ctrl+X, Cmd/Ctrl+V, and Cmd/Ctrl+A inside the desktop sidebar Browser preview could be handled by the host window instead of the embedded preview guest. The existing clipboard permission fix covers page JavaScript clipboard writes, but not native keyboard editing shortcuts.

## Fix

Intercept native editing shortcuts in the preview WebContents before host menu handling and dispatch them directly to the guest editing commands, including paste-as-plain-text.

## Verification

- `PATH=/Users/kristupas/.vite-plus/js_runtime/node/24.17.0/bin:$PATH vp test run apps/desktop/src/preview/Manager.test.ts apps/desktop/src/preview/BrowserSession.test.ts` — 99 passed
- `PATH=/Users/kristupas/.vite-plus/js_runtime/node/24.17.0/bin:$PATH vp run --filter @t3tools/desktop typecheck` — passed
- `vp fmt --check apps/desktop/src/preview/Manager.ts apps/desktop/src/preview/Manager.test.ts` — passed
- `vp run build:desktop` — passed

This is related to preview input routing, but specifically fixes physical native clipboard shortcuts rather than preview automation targeting.

Model: GPT-5.6
Harness: Codex

> [!NOTE]
> ### Route preview clipboard shortcuts (copy, cut, paste, select-all) to guest in `PreviewManager`
> - Adds `previewEditingCommand` resolver that maps Meta/Control key-down inputs (without Alt) to native editing commands: copy, cut, paste, `paste-and-match-style` (Shift+Meta+V), and select-all. Excludes composing, key-up, and Alt-modified inputs.
> - The before-input handler in `preview.makeNativeOperations` now intercepts resolved editing commands, prevents the default input event, and calls the matching `webContents` native method via the existing Effect operation pipeline.
> - Risk: `previewEditingCommand` is strict about modifier combinations — only Meta or Control without Alt qualify. Non-qualifying inputs fall through unhandled, so custom host-level key bindings that rely on Alt may no longer be shadowed by preview defaults.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 278d10f.</sup>
> 
>

### DIFF
--- a/apps/desktop/src/preview/Manager.test.ts
+++ b/apps/desktop/src/preview/Manager.test.ts
@@ -67,6 +67,52 @@ describe("isPreviewRefreshShortcut", () => {
   });
 });
 
+describe("previewEditingCommand", () => {
+  const input = (overrides: Partial<Electron.Input> = {}) =>
+    ({
+      type: "keyDown",
+      key: "c",
+      code: "KeyC",
+      isAutoRepeat: false,
+      isComposing: false,
+      shift: false,
+      control: false,
+      alt: false,
+      meta: true,
+      location: 0,
+      modifiers: [],
+      ...overrides,
+    }) as Electron.Input;
+
+  it.each([
+    ["c", "copy"],
+    ["x", "cut"],
+    ["v", "paste"],
+    ["a", "selectAll"],
+  ] as const)("maps Meta+%s to %s", (key, command) => {
+    expect(PreviewManager.previewEditingCommand(input({ key }))).toBe(command);
+  });
+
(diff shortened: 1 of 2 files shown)


dain:
This is a major problem witht he built in browser.  Not being able to paste data in or copy data out makes it unusable for many cases, and often I end up having to have the AI use computer use instead.

PR #10512, opened 2026-09-07
fix(server): bound legacy thread projection requests

cjavad:
## What Changed

- Route the legacy `orchestration.v2.get-thread-projection` WebSocket method through the existing bounded snapshot and history-page policy. This limits compatibility responses to the same 75-item / 1 MiB policy used by paginated clients.

## Why

A pre-pagination client can reconnect through the compatibility WebSocket method and materialize an entire thread in memory. On a real long-lived environment this occurred with a SQLite database (3.95 GB large) with:

- 350,869 legacy events with 1.71 GB of encoded payload and metadata
- 80,666 projected turn items with 966 MB of encoded payload
- A largest thread of about 475 MB across event and projection payloads

During a reconnect, profiling showed SQLite decoding and `JSON.parse` dominating CPU before GC pressure took over. RSS exceeded 12 GB; with an 8 GiB V8 heap limit, the server reached the heap limit and restarted about every 30 seconds.

With this bound applied, the same startup/reconnect path peaked around 1.1 GB RSS, settled around 0.5 GB, and remained idle at 0% CPU with no restart.

This problem is potentially better suited by deprecating this endpoint instead.

## Verification

- `vp test run apps/server/src/ws.test.ts apps/server/src/orchestration-v2/threadHistoryPaging.test.ts` (21 tests passed)
- `vp run --filter t3 build`

## Checklist

- [x] This PR is small and focused on one problem.
- [x] The motivation and behavior change are documented above.
- [x] No UI changes; screenshots are not applicable.
- [x] No animation or interaction changes; video is not applicable.

Generated with GPT-6 in Codex.

> [!NOTE]
> ### Bound legacy `getThreadProjection` RPC to `THREAD_HISTORY_PAGE_POLICY.maxItems + 2` rows
> The compatibility `getThreadProjection` handler in `makeWsRpcLayer` now requests a thread snapshot window instead of directly materializing the full thread projection. It uses `THREAD_HISTORY_PAGE_POLICY.maxItems + 2` as the row limit, then constructs a bounded projection from the returned projection and snapshot sequence before projecting for the wire. This prevents unbounded thread-history loads on the legacy RPC path.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bbc0651.</sup>
> 
>


cjavad:
Needs rebasing (https://github.com/pingdotgg/t3code/pull/10512/commits/bbc065130658e6138feffcf8f573700206e967e9 only commit)

Issue #9429, opened 2026-09-03
[Bug]: macOS desktop never refreshes the login-shell environment after launch — tools installed later (Homebrew, gh, opencode, nvm node) stay invisible to agents and VCS


Dam98:
**Area:** apps/desktop

## Steps to reproduce
1. Launch the packaged T3 Code app from Finder on a fresh macOS user that has no Homebrew yet.
2. While the app is running, install Homebrew (`/opt/homebrew`), which appends `eval "$(/opt/homebrew/bin/brew shellenv zsh)"` to `~/.zprofile`. Install `gh` with brew and authenticate it.
3. Also install a tool that only lands on PATH through `~/.zshrc` (opencode in `~/.opencode/bin`, or node via nvm).
4. In an agent thread (Claude Code adapter) run `which brew gh opencode node`, and open Settings → Source Control.

## Expected behavior
After the shell profiles change, the app should re-capture the login-shell environment (at least on "Rescan server environment", ideally periodically or on profile mtime change) so that `brew`, `gh`, `opencode`, `node` resolve exactly like they do in Terminal.app.

## Actual behavior
The desktop process keeps the PATH captured at launch and never probes again. Every later spawn of a bare tool name fails with ENOENT, even though the same commands work in Terminal.

Effective PATH of all T3 processes (from `ps eww`):
```
/Users/dam/.local/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:<cryptex paths>:/pkg/env/global/bin
```
PATH from a real login shell on the same machine (`zsh -il -c 'echo $PATH'`):
```
/Users/dam/.opencode/bin:/Users/dam/.nvm/versions/node/v26.8.1/bin:/Users/dam/.local/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:...
```

Timeline from `~/.t3/userdata/logs/*.trace.ndjson` (all times local, same day):
| time | event |
|---|---|
| 12:10:45, 12:11:11 | `desktop.shellEnvironment.installIntoProcess` runs (Success, ~30 ms). These are the **only** two probes in the whole log. |
| 12:36 | codex installed (`~/.local/bin/codex`), `~/.zshrc` edited (nvm, opencode) |
| 12:48 | Homebrew installed, `~/.zprofile` edited at 12:51 |
[...]

PR #10537, opened 2026-09-07
fix(desktop): recover failed shell environment capture

### LINKED ISSUE #10534: [Bug]: macOS login-shell PATH capture times out at 5s and is recorded as Success, so every provider reports "not found on PATH"
Raised from downstream fleet tooling (new-trixie-box), found while debugging a Codex provider that would not spawn.

**Area:** apps/desktop

## Summary

On macOS the bootstrap login-shell environment capture can time out at its 5000 ms bound against a perfectly ordinary zsh. The `launchctl getenv PATH` fallback is empty on a stock machine, so nothing is captured, and both parent spans still record `exit: Success`. The desktop process, the bundled server, and every provider spawned from it then run with launchd's default `PATH` (`/usr/bin:/bin:/usr/sbin:/sbin`) for the entire app run.

Nothing surfaces in the UI. The failure is reported to the user as a missing CLI instead.

## Evidence

From `~/.t3/userdata/logs/desktop.trace.ndjson.*` for a single app run (app started 15:18:40 local):

```
15:19:06.241  desktop.shellEnvironment.runCommandOutput          6105 ms
              DesktopShellEnvironmentCommandTimeoutError
              { "probe": "login-shell", "executable": "zsh", "argumentCount": 2, "timeoutMs": 5000 }
              "Desktop shell environment login-shell probe (zsh) timed out after 5000ms."   [WARN]
15:19:12.355  desktop.shellEnvironment.runCommandOutput           156 ms   (fallback probe, no value)
15:19:12.512  desktop.shellEnvironment.installPosixEnvironment   6270 ms   exit: Success
15:19:12.512  desktop.shellEnvironment.installIntoProcess        6270 ms   exit: Success
```

Effective `PATH` of every T3 process afterwards, via `ps -wwEo command=`:

```
/usr/bin:/bin:/usr/sbin:/sbin
```

The same capture run by hand on the same machine, warm, completes in about 1.1 s and returns the full `PATH`:

```
$ time /bin/zsh -ilc "printf '%s\n' '__T3CODE_ENV_PATH_START__'; printenv PATH || true; printf '%s\n' '__T3CODE_ENV_PATH_END__'"
...
1.127 total
```

So this is a timing failure at cold start, not a broken shell configuration. `launchctl getenv PATH` is empty here, and `/var/db/com.apple.xpc.launchd/config/user.plist` does not exist, which is the default state.

## Impact

Anything spawned by a bare command name fails for the whole app run, and the app run is long: this one had been up 17 hours.

The user-visible message is misleading. The Codex provider reported:

> Codex CLI (`codex`) was not found on PATH.

while `codex` 0.153.2 was installed at `/opt/homebrew/bin/codex` and worked in any terminal. The thread-level error was:

> ProviderAdapterProcessError: Failed to spawn Codex App Server process for command: codex app-server

`checkCodexProviderStatus` produces that text from `isCodexAppServerSpawnError`, so a spawn failure caused by the app's own empty `PATH` is presented as a missing installation. Setting an absolute `binaryPath` in provider settings works around it per provider, and the probe then reports `installed: true` immediately.

## Suggested changes

1. Do not record a swallowed capture failure as `exit: Success` on `installPosixEnvironment` and `installIntoProcess`. The WARN event is only reachable by reading the trace file.
2. Surface the failed capture in the UI, and distinguish it from a missing CLI. "The shell environment could not be captured" is actionable; "`codex` was not found on PATH" sends the user to reinstall a CLI that is already installed.
3. Retry once before giving up, or raise the bound. 5000 ms is tight for a cold start where the login shell has not been paged in yet.
4. Offer a way to re-run the capture without restarting the app.

## Related, but not duplicates

- #9429 — the capture never refreshes after launch. That one starts from a capture that succeeded; this one never gets a value at all.
- #7846 — the probe is rejected at parse time by nushell. Different cause, same swallowed outcome.
- #6115, #7618 — downstream symptoms of a provider being judged missing.

## Environment

- T3 Code (Alpha) 0.0.99-gitcache.1
- macOS 26.6.2 (25G83), Apple Silicon
- Login shell zsh 5.9, `SHELL=/bin/zsh`
- codex 0.153.2 at `/opt/homebrew/bin/codex`

Gigioxx:
A timed-out desktop login-shell probe could leave every provider using launchd's default PATH while both install spans reported success. Retry each POSIX shell once when it returns no usable PATH, retain the launchctl fallback, and fail the install effect if neither captures PATH.

After Electron is ready, show a native warning with Retry and Continue before starting the bundled server. Retry captures the environment again in the same process, so providers inherit the recovered PATH. Continue preserves startup and explains the absolute CLI path workaround. Windows is unchanged. This handles failed startup capture; refreshing an already-running server after Continue remains the separate #9429 concern.

Fixes #10534.

Verified: 23 focused tests, desktop typecheck, scoped lint and formatting. The retry and false-success regressions fail against the original code. Tests cover the 5-second timeout, bounded retries, launchctl fallback, failed parent spans, and recovery choices. Each unsuccessful shell candidate can now take one additional 5-second attempt.

An isolated Linux Electron harness also ran real shell subprocesses: a cold probe timed out on the original code and left a fixture CLI unresolved; the patched code retried and spawned it by bare name. Native Retry recaptured PATH without restarting Electron; Escape continued with the inherited PATH. macOS itself and the full desktop client were not exercised.

Evidence below uses the production capture/recovery code with a controlled shell and fixture CLI. The surrounding diagnostics are the test harness, not the T3 client UI.

| Before: silent success with missing CLI | After: capture failure with recovery choices |
| --- | --- |
| Original capture result | Native shell recovery dialog |

[Retry recording](https://github.com/Gigioxx/t3code/releases/download/evidence-10534-shell-capture/retry.mp4) · [Recovered PATH and successful child process](https://github.com/Gigioxx/t3code/releases/download/evidence-10534-shell-capture/recovered.png)

> [!NOTE]
> ### Recover from failed shell environment capture during desktop startup
> - Desktop startup previously failed when shell-environment installation threw `DesktopShellEnvironmentCaptureError`; it now captures the result and continues to the Electron-ready stage
> - After Electron is ready, the workflow invokes shell-environment recovery, converting recovery failures into fatal startup failures
> - User can retry (reinstalls the environment) or continue (proceeds without a successful capture)
> - Risk: startup now proceeds past a failed capture in [DesktopApp.ts](https://github.com/pingdotgg/t3code/pull/10537/files#diff-267cf20af5c499f356b88a4bc3d3dec61aa6c7a3398f93f8836975b3bcf764e9); consumers relying on a populated shell environment may see missing variables when the user chooses "continue"
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 28b6b6f.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved desktop startup reliability when the shell environment or PATH cannot be detected.
  * Shell environment detection now retries transient or empty results and uses a fallback where available.

* **New Features**
  * Added recovery prompts when shell environment detection fails, allowing you to retry or continue starting the app.
  * Startup errors are now handled more gracefully instead of interrupting the launch process.

### DIFF
--- a/apps/desktop/src/app/DesktopApp.ts
+++ b/apps/desktop/src/app/DesktopApp.ts
@@ -2,6 +2,7 @@ import * as Cause from "effect/Cause";
 import * as Effect from "effect/Effect";
 import * as Option from "effect/Option";
 import * as Ref from "effect/Ref";
+import * as Result from "effect/Result";
 import * as Schema from "effect/Schema";
 
 import * as NetService from "@t3tools/shared/Net";
@@ -26,6 +27,7 @@ import * as DesktopShutdown from "./DesktopShutdown.ts";
 import * as DesktopServerExposure from "../backend/DesktopServerExposure.ts";
 import * as DesktopAppSettings from "../settings/DesktopAppSettings.ts";
 import * as DesktopShellEnvironment from "../shell/DesktopShellEnvironment.ts";
+import { recoverShellEnvironment } from "../shell/DesktopShellEnvironmentRecovery.ts";
 import * as DesktopState from "./DesktopState.ts";
 import * as DesktopRemoteUpdates from "../updates/DesktopRemoteUpdates.ts";
 import * as DesktopUpdates from "../updates/DesktopUpdates.ts";
@@ -242,7 +244,7 @@ const startup = Effect.gen(function* () {
   const updates = yield* DesktopUpdates.DesktopUpdates;
   const environment = yield* DesktopEnvironment.DesktopEnvironment;
 
-  yield* shellEnvironment.installIntoProcess;
+  const shellEnvironmentResult = yield* Effect.result(shellEnvironment.installIntoProcess);
   const hasCommandLinePasswordStore =
     preReadyElectronOptions.linuxPasswordStoreCommandLine !== null;
   const linuxElectronOptions =
@@ -287,6 +289,11 @@ const startup = Effect.gen(function* () {
     Effect.catchCause((cause) => fatalStartupCause("whenReady", cause)),
   );
   yield* logStartupInfo("app ready");
+  if (Result.isFailure(shellEnvironmentResult)) {
+    yield* recoverShellEnvironment().pipe(
+      Effect.catchCause((cause) => fatalStartupCause("shell environment recovery", cause)),
+    );
+  }
   if (environment.platform === "linux") {
     const selectedBackend = yield* safeStorage.selectedStorageBackend;
     yield* logStartupInfo("safe storage ready", {
--- a/apps/desktop/src/app/DesktopAppErrors.test.ts
+++ b/apps/desktop/src/app/DesktopAppErrors.test.ts
@@ -1,4 +1,9 @@
 import { assert, describe, it } from "@effect/vitest";
+import * as Effect from "effect/Effect";
+import * as Layer from "effect/Layer";
+import { recoverShellEnvironment } from "../shell/DesktopShellEnvironmentRecovery.ts";
+import * as ElectronDialog from "../electron/ElectronDialog.ts";
+import * as DesktopShellEnvironment from "../shell/DesktopShellEnvironment.ts";
 
 import {
   DesktopBackendPortUnavailableError,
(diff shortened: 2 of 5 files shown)


shivamhwp:

`DesktopShellEnvironmentCaptureError` extends `Schema.TaggedErrorClass`, which is absent from the current Effect version. Importing the recovery path throws `TypeError: TaggedErrorClass is not a function`, before desktop startup can offer Retry or Continue, and desktop typecheck fails.

Use `Schema.TaggedError<DesktopShellEnvironmentCaptureError>()`, matching the two existing shell-environment error classes in this file. Keep the distinction from #9429: startup recovery does not refresh the environment of an already-running server.

PR #10573, opened 2026-09-07
feat(web): add keyboard revert picker

spaansba:
Adds keyboard-first checkpoint revert so users can inspect and choose a target without relying on per-message hover controls.

Adds `chat.revert` with `mod+shift+z`, an unbound `chat.revertLast` command, and double Escape from an empty idle composer. The picker previews the target in the timeline, dims discarded rows, shows cumulative diff stats, skips expired checkpoints, and commits with Enter without an extra dialog. Providers without conversation rollback leave the shortcut inactive.

Older checkpoints fall back to turn IDs when assistant message IDs are unavailable. Stopped threads resolve their persisted Git workspace and recover the provider session instead of requiring an already active process. Empty file diffs remain revertible.

Post-revert Undo is not included because provider conversation rollbacks are not generically reversible through the current adapter contract.

Tests:
- 178 focused web and contracts tests
- 3 focused checkpoint reactor tests
- Web and server typechecks
- Targeted lint completed with existing warnings only

Generated with GPT-5.6 Sol in the Codex harness.

## Summary by CodeRabbit

- **New Features**
  - Added a revert picker for chat conversations, allowing users to preview and return to earlier turns.
  - Revert options show message snippets, change summaries, timestamps, and availability status.
  - Added keyboard shortcuts for opening the revert menu and reverting the latest turn.
  - Added keyboard and pointer interactions for navigating, selecting, and dismissing the picker.

- **Bug Fixes**
  - Reverting now works for stopped conversations using their persisted workspace, even without an active provider session.

### DIFF
--- a/apps/server/src/orchestration/Layers/CheckpointReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/CheckpointReactor.test.ts
@@ -1764,6 +1764,48 @@ describe("CheckpointReactor", () => {
     ).toBe(false);
   });
 
+  it("reverts a stopped thread using its persisted workspace", async () => {
+    const harness = await createHarness({ hasSession: false });
+    const threadId = ThreadId.make("thread-1");
+    const createdAt = "2026-01-01T00:00:00.000Z";
+
+    for (const turnCount of [1, 2]) {
+      await Effect.runPromise(
+        harness.engine.dispatch({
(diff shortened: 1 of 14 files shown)


juliusmarminge:
Closing as part of the open-PR backlog sweep (wave 3).

**Reason:** Unsolicited XL keyboard revert-picker feature

PR #10586, opened 2026-09-07
fix(server): dismiss orphaned provider questions

### LINKED ISSUE #5454: [Bug]: Pending user-input popup can never be dismissed after the provider session dies (e.g. server restart)

yashranaway:
Native provider questions can outlive their session and leave clients offering answers that the provider can no longer accept. Session exit now dismisses native questions, including turnless questions. Answering a legacy question against a missing or stopped session also records a durable dismissal while retaining the failed-answer activity.

Dismissals use server time strictly after the stored request timestamp, so clock rollback or an older client timestamp cannot leave pending input set after restart. Repeated answers do not append duplicate resolutions. Asynchronous message-mode questions and transient failures from live sessions retain their existing behavior.

Validation: 131 focused ProviderCommandReactor and ProviderRuntimeIngestion tests pass. Missing/stopped-session timestamp regressions fail before the fix and pass afterward, including restart, durable pending state, and duplicate-resolution coverage. Scoped lint and server typecheck pass with existing warnings/suggestions. Rebased onto main.

Fixes #5454.

Model: GPT-6
Harness: Codex in T3 Code

## Summary by CodeRabbit

- **Bug Fixes**
  - Native user-input requests are now resolved when their provider session ends or is stopped.
  - Requests remain pending while the provider session is still running, including after a restart.
  - Resolution applies consistently across turns, including requests without an associated turn.
  - Repeated session-exit events no longer create duplicate resolutions.
  - Responses received after session exit correctly clear pending input indicators.
  - Requests tied to message-based responses are not incorrectly marked as resolved.

### DIFF
--- a/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
+++ b/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts
@@ -27,6 +27,7 @@ import { serializeAssistantCitation } from "@t3tools/shared/assistantCitations";
 import * as Effect from "effect/Effect";
 import * as Deferred from "effect/Deferred";
 import * as Exit from "effect/Exit";
+import * as Fiber from "effect/Fiber";
 import * as Layer from "effect/Layer";
 import * as ManagedRuntime from "effect/ManagedRuntime";
 import * as Option from "effect/Option";
@@ -46,7 +47,10 @@ import {
 } from "../../provider/Errors.ts";
 import { OrchestrationEventStoreLive } from "../../persistence/Layers/OrchestrationEventStore.ts";
 import { OrchestrationCommandReceiptRepositoryLive } from "../../persistence/Layers/OrchestrationCommandReceipts.ts";
-import { SqlitePersistenceMemory } from "../../persistence/Layers/Sqlite.ts";
+import {
+  makeSqlitePersistenceLive,
+  SqlitePersistenceMemory,
+} from "../../persistence/Layers/Sqlite.ts";
 import {
   ProviderService,
   type ProviderServiceShape,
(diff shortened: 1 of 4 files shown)


shivamhwp:

The missing/stopped-session dismissal copies the client response timestamp. When that timestamp precedes the original question, the durable activity query still returns `user-input.requested`: its activities have no sequence, so the query falls back to `created_at`. The client hides the resolved prompt, but the server keeps `hasPendingUserInput` true, and another response appends a second dismissal.

Use a server-generated dismissal timestamp strictly after the stored request timestamp for both the activity and append command. Keep the durable lookup and the message-mode/live-session guards. The dismissal needs to remain the latest request activity after a restart and repeated answers must not add more resolutions.

PR #10597, opened 2026-09-07
fix(claude): keep limits when the usage fetch fails

Mhmdhammoud:
## What Changed

`claudeUsageResponseToLimits` now distinguishes two shapes of the SDK `get_usage` response:

- `rate_limits_available: false` → `unsupported` (API key, Bedrock, Vertex, missing scope), as before.
- `rate_limits_available: true` with `rate_limits: null` → `probeFailed`, new.

One test added for the second shape.

## Why

Claude Code keeps `rate_limits_available` true and nulls `rate_limits` when its own usage fetch fails, for example when `/api/oauth/usage` answers 429. T3 treated that as `unsupported`, which hid the Claude card on Usage → Limits and made `/usage-limits` report "This account has no subscription limits." on an active Max account.

`probeFailed` is the reason the publish path already handles: `resolvePublishedUsageLimits` keeps the last good windows until the next successful probe, so a throttled read now leaves the card in place instead of removing it.

Reproduced on a Max 20x account during a 24-minute 429 cooldown on the usage endpoint. The profile endpoint kept answering 200 the whole time.

## Checklist

- [ ] I included before/after screenshots for any UI changes (server-only change)

Verified with `vp test run src/provider/Layers/claudeUsageLimits.test.ts` (8 passed), lint on the two files, and the server typecheck.

Written by Claude Fable 5.1 via Claude Code in T3 Code.

> [!NOTE]
> ### Classify failed Claude usage fetch as `probeFailed` to retain known limits
> Adds a test case in [claudeUsageLimits.test.ts](https://github.com/pingdotgg/t3code/pull/10597/files#diff-5eef39fbbec43fbe4f8657f26b68edadae028bce7e66e801c3b6ba1006503025) for when `rate_limits_available` is true but `rate_limits` is null. The mapper must return no windows with an `unavailable` reason of `probeFailed`, so consumers can distinguish a failed usage probe from an unsupported account and keep previously known windows.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6ad2140.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved usage-limit status reporting when rate-limit information cannot be fetched.
  * Distinguishes temporary fetch failures from unsupported rate-limit availability for more accurate status details.


juliusmarminge:

PR #10417, opened 2026-09-06
fix(clients): show the submitted answers on user input rows

Mnigos:
## Problem

When an agent asks a question through the user-input request (AskUserQuestion and the Codex equivalent) and the user answers, the timeline only shows a bare "User input submitted" row. What was actually chosen or typed is not visible anywhere afterwards, even though the resolved activity already carries the answers.

## Fix

- New shared helper `formatUserInputAnswers` in `@t3tools/client-runtime/user-input-answers`: one `Question: answer` line per question in the order asked, multi-select answers joined with commas, and answers whose question is unknown (older history) listed under their key. `readUserInputQuestions` reads the questions structurally from the request payload.
- Web and mobile pair each `user-input.resolved` row with its `user-input.requested` row by `requestId` while deriving work-log entries and set the formatted answers as the row's detail. The row keeps its icon and label; it just becomes expandable and shows the exchange. Rows already carrying a detail are left alone.

No server or contract change: the answers were always in the payload.

## Verification

- `userInputAnswers.test.ts`: ordering, multi-select joining, unknown keys, blank answers, malformed questions.
- Web `session-logic.test.ts` and mobile `threadActivity.test.ts`: the resolved row gets the answers labelled by their questions, and a resolved row without a matching request still lists what was typed.
- Typecheck for client-runtime, web and mobile.

## Before

main, Claude Haiku asked through AskUserQuestion ("Approach": Minimal fix / Full rewrite), answered from the card: the submitted row is a bare label with nothing to expand.

Before: bare User input submitted row

## After

Same exchange on this branch. On web the timeline already prefers a row's detail as its label for rows like this, so the answer is readable without expanding and the expanded body repeats it; on mobile the row keeps its "User input submitted" heading with the answer as the preview line.

After: row shows Approach: Minimal fix

Playwright screenshots at 2x of the dev web client (full window: [before](https://raw.githubusercontent.com/Mnigos/t3code/pr-assets/user-input-answers/10417-before-full.png), [after](https://raw.githubusercontent.com/Mnigos/t3code/pr-assets/user-input-answers/10417-after-full.png)).

Implemented with Claude Code (Claude Fable 5).

## Summary by CodeRabbit

- **New Features**
  - Resolved user-input activities now display submitted answers directly in the work log across mobile and web.
  - Answers are labeled with their corresponding question text for easier review.
  - Multi-select responses are shown in a readable format, while other response types are formatted clearly.
  - Answers without a matching question are still displayed using their original keys.
[...]

PR #10602, opened 2026-09-07
fix(chat): prevent duplicate statuses and timeline ID collisions

dominic-r:
A single status could render twice, as a work-group heading and an event row. Activity IDs could also collide with message IDs, hiding an async answer or leaving an empty timeline slot.

Render single inactive statuses directly and give work/activity rows distinct timeline IDs on web, desktop, and mobile. Keep expanded subagent rows visible when their agents settle by comparing expansion state with the underlying work-entry ID.

This complements the Q&A presentation shipped in #11014, which superseded #10417. It preserves that presentation and covers standalone statuses and ID collisions, including older answer records that cannot be folded into a Q&A row. It does not change stored messages or provider payloads and does not claim to fix the question-attribution issue in #10786.

Validation at `f02caf35c`:

- All 295 tests in the web timeline, web session-logic, and mobile feed suites pass using the installed Vitest runner. The `vp test` wrapper fails before running tests.
- The existing subagent test reproduced an expanded row disappearing after settlement with the new IDs and passes with the corrected comparison.
- Web and mobile typechecks, formatting, and diff checks pass. Targeted lint reports warnings in unchanged renderer code.

The images below show the original fix before the merge with current main. Browser verification was not repeated after that merge; mobile was not simulator-tested.

| Before | After |
| --- | --- |
| Duplicate status heading and event row | One status row with the submitted answer preserved |

Model: GPT-6. Harness: Codex.

> [!NOTE]
> ### Fix single status event rendering to avoid duplicate headings in chat timeline
> - Mobile: `appendToolGroupRows` in [threadActivity.ts](https://github.com/pingdotgg/t3code/pull/10602/files#diff-a18e8ed6aa4a9d6b3d9e32ea2fa3034dfd7621b6af3d640ec1b95a8564c97eb5) now emits a single inactive non-tool activity directly as one activity-group row and returns before the tool-group summary path.
> - Web: `deriveMessagesTimelineRows` in [MessagesTimeline.logic.ts](https://github.com/pingdotgg/t3code/pull/10602/files#diff-6cdfacc6ebbdbc5c4b4058c0a430b87d768ac2b6accf35fb4c9dfc23500914fe) now treats any single visible work entry (not just tool-like ones) as a plain work row. The unused `summaryToolIcon` field is removed from `WorkGroupToggleTimelineRow` and its fallback lookup in [MessagesTimeline.tsx](https://github.com/pingdotgg/t3code/pull/10602/files#diff-f2a34c4ad8d2b68c45657dbdcbf14afac4ea93e1fbd37007aaa2ccb8b41d2588) is dropped.
> - Both mobile and web gain parameterized tests covering a resolved user-input event in collapsed and expanded states.
> - Risk: `WorkGroupToggleTimelineRow.summaryToolIcon` is removed; any out-of-tree consumers of this type field will need to update. Grouped-toggle summaries now always call `summarizeToolGroup` directly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4d832a5.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved user-input activities now appear exactly once in mobile and web timelines, with the original message preserved.
  * Single non-tool activities no longer appear as work toggles when inactive.
  * Spawned work rows remain visible outside expandable turn sections.
  * Work-group summaries and icons now display consistently across activity types.
  * Timeline rows use distinct identifiers, preventing activity and work entries from being confused with other content.


dominic-r:
The browser check exposed an ID collision between async answer messages and their status activities. Fixed in e01b00997 by giving activity entries distinct timeline IDs before folding and rendering on web and mobile. The answer now stays visible with the turn collapsed or expanded, and the expanded timeline has one submitted-status row with no empty slot. Before/after images are in the description. All 290 focused tests and both client typechecks pass.

The generic docstring-coverage suggestion is left unapplied. AGENTS.md asks for comments that explain non-obvious constraints, rather than descriptions of straightforward rendering code. The regression tests cover the behavior and the new ID collision case.

dominic-r:
@macroscope-app review

PR #10607, opened 2026-09-07
fix(grok): recover from crashed provider sessions

saphid:
## Summary

When the Grok ACP subprocess dies, the adapter kept the dead session registered, so the next turn could be routed into a corpse instead of recovering. This retires a terminated session as soon as the runtime reports `ConnectionTerminated`: the in-flight turn is settled as failed, the session emits `session.exited` with `exitKind: "error"`, and `hasSession`/`listSessions`/`requireSession` stop treating it as live — so a retry goes through the normal recovery path and resumes the saved ACP session via `session/load`.

Cleanup runs under the per-thread lock, and an identity check prevents late teardown from deleting a replacement session.

## Dark-mode proof

Actual local web-app captures using a disposable workspace and a controlled Grok ACP mock. Both revisions use the same mock: `crash now` exits the subprocess with code 23, then `Please continue.` exercises recovery in the same thread. No live Grok account was used.

Base: [7445aa7](https://github.com/pingdotgg/t3code/commit/7445aa733ada33e45289e5aa5055f79142556513). Candidate: [6aa2c3d](https://github.com/saphid/t3code/commit/6aa2c3da7cda0a70514aac6986fb72dfcb305325). Captured 22 September 2026; production source was unchanged.

**Before (base): retry fails again.**

Before in dark mode: retry returns another process-exited error

**After (candidate): retry resumes and receives a reply.**

After in dark mode: retry resumes and replies Ready to continue

The request log confirms that the candidate starts a new process and calls `session/load` for the saved `mock-session-1`. The base starts no replacement chat process. Prior messages remain in the thread.

**Recorded interaction — before (base)**

Before recording: Grok crashes and the follow-up fails again

**Recorded interaction — after (candidate)**

After recording: Grok crashes and the follow-up resumes the session

GIFs come from the actual recordings at 10 fps, with unchanged speed and a two-second hold on the final frame. Screenshots use the narrow web layout (480 × 640); recordings use 640 × 760. This proves recovery behavior, with no latency claim.

<details>
<summary>Full captures, recordings, and verification receipt</summary>

- Full screenshots: [before](https://github.com/saphid/t3code/releases/download/pr-evidence-10607-dark-20260922/before-narrow.png), [after](https://github.com/saphid/t3code/releases/download/pr-evidence-10607-dark-20260922/after-narrow.png).
- Real-time MP4 recordings: [before](https://github.com/saphid/t3code/releases/download/pr-evidence-10607-dark-20260922/before-retry.mp4), [after](https://github.com/saphid/t3code/releases/download/pr-evidence-10607-dark-20260922/after-retry.mp4).
- Original WebM recordings: [before](https://github.com/saphid/t3code/releases/download/pr-evidence-10607-dark-20260922/before-raw.webm), [after](https://github.com/saphid/t3code/releases/download/pr-evidence-10607-dark-20260922/after-raw.webm).
- [Capture conditions and sanitized protocol evidence](https://github.com/saphid/t3code/releases/download/pr-evidence-10607-dark-20260922/capture.json).

</details>

## Verification

Merged upstream `main` at `7445aa733ada33e45289e5aa5055f79142556513` (20 September 2026).

- `vp test run apps/server/src/provider/Layers/GrokAdapter.test.ts --maxWorkers 1`: **46 passed**, rerun on the candidate on 22 September 2026.
- `vp exec tsc --noEmit -p apps/server/tsconfig.json`: passed.
- Targeted formatting, lint, and `git diff --check`: passed (existing warnings/suggestions may remain).

The mock ACP subprocess regression verifies crash retirement, rejection during teardown, and resumption from the saved session. Upstream background-task tracking is preserved. No live Grok account run was performed.

Independent review was attempted before the final upstream sync, directly with `devin -p --model swe-2-max` using a frozen diff and repository standards. The noninteractive tool-based attempt returned without reviewing; bounded no-tool attempts timed out (status 124) without a verdict. No new independent approval is claimed.
Implemented and verified using SWE-2 High in the Devin/T3 Code harness; independent read-only review by Codex (gpt-5.6-sol, high) found one race candidate that was verified against the thread-lock ordering and dismissed as a false positive.

Coordination trace: T3 thread e0a842ec-3f5d-454e-9e3c-888b3e7ab714

Current upstream integration and focused verification: GPT-6 Astra in the Codex harness (T3 Code).

Dark-mode proof refresh and focused-test rerun: GPT-6 Astra in the Codex harness (T3 Code).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of unexpected Grok connection failures.
  - Failed sessions are now reported as errors instead of graceful exits.
  - Active prompts receive a clear connection-terminated error.
  - Terminated sessions are no longer shown as available.

- **Reliability**
  - Sessions can recover from a crashed process using saved progress.
  - Deliberately restarted sessions can resume successfully from their saved cursor.

### DIFF
--- a/apps/server/scripts/acp-mock-agent.ts
+++ b/apps/server/scripts/acp-mock-agent.ts
@@ -617,6 +617,12 @@ const program = Effect.gen(function* () {
     Effect.gen(function* () {
       const requestedSessionId = String(request.sessionId ?? sessionId);
       promptCount += 1;
+      if (
+        process.env.T3_ACP_CRASH_PROMPT === "1" &&
+        request.prompt.some((part) => part.type === "text" && part.text === "crash now")
+      ) {
+        return yield* Effect.sync(() => process.exit(23));
+      }
 
       if (completeFirstPromptOnCancel && promptCount === 1) {
         yield* agent.client.sessionUpdate({
--- a/apps/server/src/provider/Layers/GrokAdapter.test.ts
+++ b/apps/server/src/provider/Layers/GrokAdapter.test.ts
@@ -294,6 +294,42 @@ it.layer(grokAdapterTestLayer)("GrokAdapterLive", (it) => {
     }),
   );
 
+  it.effect("retires a crashed process so a deliberate retry can resume", () =>
+    Effect.gen(function* () {
+      const threadId = ThreadId.make("grok-crash-recovery");
+      const wrapper = yield* Effect.promise(() =>
+        makeMockGrokWrapper({ T3_ACP_CRASH_PROMPT: "1" }),
+      );
+      const adapter = yield* makeTestAdapter(wrapper);
+      const exited =
+        yield* Deferred.make<Extract<ProviderRuntimeEvent, { type: "session.exited" }>>();
+      const events = yield* Stream.runForEach(adapter.streamEvents, (event) =>
+        event.type === "session.exited"
+          ? Deferred.succeed(exited, event).pipe(Effect.asVoid)
+          : Effect.void,
+      ).pipe(Effect.forkChild);
+      const input = {
+        threadId,
+        provider: ProviderDriverKind.make("grok"),
+        cwd: process.cwd(),
+        runtimeMode: "full-access" as const,
+      };
+      const session = yield* adapter.startSession(input);
+      const failure = yield* Effect.flip(
+        adapter.sendTurn({ threadId, input: "crash now", attachments: [] }),
+      );
+      assert.isDefined(failure);
+      assert.isFalse(yield* adapter.hasSession(threadId));
+      assert.equal((yield* Deferred.await(exited)).payload.exitKind, "error");
+      assert.deepStrictEqual(yield* adapter.listSessions(), []);
+      yield* Fiber.interrupt(events);
+      yield* adapter.startSession({ ...input, resumeCursor: session.resumeCursor });
+      const turn = yield* adapter.sendTurn({ threadId, input: "retry now", attachments: [] });
+      assert.equal(turn.threadId, threadId);
+      yield* adapter.stopSession(threadId);
+    }),
+  );
+
   it.effect("starts a session and maps mock ACP prompt flow to runtime events", () =>
     Effect.gen(function* () {
       const threadId = ThreadId.make("grok-mock-thread");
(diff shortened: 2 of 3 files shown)


saphid:
Friendly review nudge @juliusmarminge @maria-rcks — this is mergeable and hasn't had a maintainer pass yet. Independent bot/agent reviews have run with findings triaged in-commit (see receipts in earlier comments). Full queue context and status: #10688.

saphid:
Rebased onto latest `main` (`41f4dd0`). All checks pass except `Release Smoke`, which fails repo-wide for an unrelated reason: `expo-audio@57.0.5` was published on 2026-09-11, so the smoke test's lockfile regeneration resolves `~57.0.4` to 57.0.5 and pnpm errors with `ERR_PNPM_UNUSED_PATCH` for the pinned `expo-audio@57.0.4` patch. This PR touches only `apps/server` Grok files — no manifests. Other PR runs are failing the same way.

saphid:
Re: CodeRabbit's outside-diff finding "Prevent `startSession` from succeeding after termination" (`GrokAdapter.ts:1516-1527`, review 5259312443). I checked it against 6aa2c3da7c. The race exists, but it does not leave a live or registered dead session, and the suggested check could not catch it. No change.

**What can happen.** `AcpSessionRuntime` buffers `ConnectionTerminated` in an unbounded queue (`AcpSessionRuntime.ts`, `eventQueue` and `getEvents: () => Stream.fromQueue(eventQueue)`). If the agent dies after `session/set_model` succeeds, the notification consumer can see the termination while `startSession` still holds the thread lock. `startSession` then returns `ready` and emits `session.started`. The retirement fork runs as soon as the lock is released.

**Why no dead session survives it:**
- The retirement fork runs in `ownerScope`, not the notification fiber, so `stopSessionInternal` interrupting that fiber cannot cancel it.
- It retires this exact context. `startSession` holds the lock through `sessions.set`, so the identity check `sessions.get(ctx.threadId) !== ctx` passes. `stopSessionInternal` then closes the scope, deletes the entry, and emits `session.exited` with `exitKind: "error"`.
- Until then, `requireSession`, `hasSession`, and `listSessions` all reject `ctx.terminated`. A `sendTurn` queued on the same lock gets `ProviderAdapterSessionNotFoundError` and never reaches the dead ACP connection.
- If another `startSession` wins the lock first, it stops the old context (`existing && !existing.stopped`), and the identity check makes the retirement fork a no-op.

The caller sees `ready`, then `session.exited(error)`. That is the same sequence as an agent that crashes one tick after startup, and it is the recovery path this PR adds and tests.

**Why the suggested re-check would not help.** Effect forks do not start immediately by default (`startImmediately` defaults to off in effect-smol's `forkIn`). There is no yield between `Effect.forkIn(ctx.scope)` for the consumer, `sessions.set`, and the watchdog fork. So `ctx.terminated` is always `false` at the suggested spot. Moving the check later would only shift the window, since a termination one tick after any check gives the same result. The guarantee is that termination retires the session under the lock, not that startup proves the agent is alive.

PR #10709, opened 2026-09-08
fix(mobile): honor requested terminal native architectures

bompus:
## What Changed

The terminal Android module now applies `reactNativeArchitectures` to its NDK ABI filters when the property is supplied. When it is absent, the module retains its existing defaults.

## Why

Building with `-PreactNativeArchitectures=arm64-v8a` still scheduled terminal CMake builds for all four ABIs, even though the APK only needed ARM64. Honoring the requested architectures avoids this unnecessary native compilation.

Validation on Windows with Gradle 9.3.1:

- The terminal native release task graph selects only ARM64 for `arm64-v8a`, and exactly ARM64 and x86_64 for `arm64-v8a,x86_64`.
- With the property absent, the task graph still includes all four default ABIs.
- The actual ARM64 native release build passed; `llvm-readelf` confirmed the resulting `libt3terminal.so` is an AArch64 shared library.
- `git diff --check` passed. A full APK build was not repeated for this change.

## Checklist

- UI screenshots and interaction video: not applicable; no UI changes.

Prepared with GPT-6 in the Codex desktop app.

> [!NOTE]
> ### Honor `reactNativeArchitectures` for Android NDK ABI filters in `t3-terminal`
> Adds conditional logic to [build.gradle](https://github.com/pingdotgg/t3code/pull/10709/files#diff-eb7ac7d68143d23ee014c54a2b60f3785ec449802c85a740874e003cff84d05c) that reads the `reactNativeArchitectures` project property and applies it as NDK ABI filters when present. Without the property, build behavior is unchanged.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6aaa461.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring Android native architectures through the `reactNativeArchitectures` build setting.
  * Enables builds to target a selected list of device architectures.


bompus:
Answering the approvability hold, since the concern is about ABI coverage rather than a defect in the diff.

**The property has exactly one producer in this repo**, `scripts/mobile-showcase.ts:747`:

```ts
...(abis.length > 0 ? [`-PreactNativeArchitectures=${abis.join(",")}`] : []),
```

so the value is always a comma-joined list with no spaces and no empty elements.

**Those ABIs come from real attached devices**, not a static matrix — `scripts/mobile-showcase.ts:1420-1426` builds `androidAbis` from `capture.device.abi` for android captures and dedupes through a `Set`. Every element is a genuine ABI string reported by a device, and the list is 1..n devices long.

**The no-property path is provably unchanged.** When no android device is captured, `abis` is empty and line 747 omits the flag entirely; `findProperty` then returns `null`, the `if` is false, and no `abiFilters` block is emitted at all. `-PreactNativeArchitectures=` (empty string) takes the same path, since an empty string is falsy in Groovy.

**The shape matches the React Native template verbatim** — same `findProperty` read, same truthiness guard, same `split(",")`, same spread into `abiFilters`, same `defaultConfig.ndk` placement:

```groovy
def reactNativeArchitectures() {
    def value = project.getProperties().get("reactNativeArchitectures")
    return value ? value.split(",") : ["armeabi-v7a", "arm64-v8a", "x86", "x86_64"]
}
...
ndk { abiFilters (*reactNativeArchitectures()) }
```

RN does not trim elements either, so this does not introduce a handling difference from the framework convention.

**On the bug being fixed:** `abiFilters` appears nowhere else in the repo. Without this, the `t3-terminal` native module compiles and packages every ABI even when the build was asked for a single device's ABI — wasted CMake work, and an APK whose native payload does not match the requested set.

One deliberate omission: `split(',')` does not tolerate spaces, so a hand-typed `-PreactNativeArchitectures="arm64-v8a, x86_64"` would yield the invalid ABI `" x86_64"`. I left that alone rather than add `*.trim().findAll { it }`, since no caller in the repo does it and trimming would deviate from the RN convention above. Happy to add it if you would rather be defensive about manual gradlew invocations.

PR #8301, opened 2026-08-26
fix(server): stage only changed checkpoint paths

PetrSeifert:
## What Changed

- Discover checkpoint paths with a scoped, NUL-delimited `git status` using the real index.
- Stage only those paths into the temporary checkpoint index through `--pathspec-from-file`.
- Preserve raw Git pathname bytes from `status -z` output through pathspec stdin, including non-UTF-8 names.
- Preserve workspace-subdirectory scope and leave the user's real index unchanged.
- Fall back to the existing scoped full add only if status output exceeds the preview limit.
- Add focused unit and Git integration coverage for staged, unstaged, deleted, and untracked paths.

## Why

Checkpoint capture currently builds a temporary index with `git add -A -- .`. On repositories with hundreds of thousands of tracked files, that can exceed the 30-second process timeout even when only a few files changed. The missing checkpoint then prevents Latest turn diffs and summaries from loading.

This change keeps the checkpoint content equivalent while limiting temporary-index staging to the paths Git already reports as changed.

Related to #3646. This PR targets large repositories with a small changed-file set. It does not claim to add failure backoff, temporary-pack cleanup, or handling for a single oversized untracked file.

## Testing

- `vp test run apps/server/src/stream/collectUint8StreamText.test.ts apps/server/src/processRunner.test.ts apps/server/src/vcs/VcsProcess.test.ts apps/server/src/vcs/GitVcsDriver.test.ts`
- `vp lint --report-unused-disable-directives apps/server/src/vcs/GitVcsDriver.ts apps/server/src/vcs/GitVcsDriver.test.ts`
- `vp run --filter t3 typecheck`
- `git diff --check upstream/main...HEAD`

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [x] No UI changes

---

> [!NOTE]
> **Medium Risk**
> Changes core checkpoint staging and relies on porcelain v1 byte parsing; a format mismatch could miss paths until the slower `add -A` fallback runs.
> 
> **Overview**
[...]

Issue #3646, opened 2026-07-02
[Bug]: Checkpoint capture on large monorepos retries a guaranteed 30s git-add timeout every turn — permanent CPU burn + tmp_pack disk litter


mikeohuo:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

### Steps to reproduce

1. Open a workspace inside a large monorepo (in my case a Unity project — large pack file, very high file count; workspace cwd is a subdirectory of the repo root). The larger the repo, the worse this gets.
2. Run agent threads normally. Every completed turn triggers `CheckpointReactor` → `GitVcsDriver.checkpoints.captureCheckpoint`, which runs `git read-tree HEAD` + `git add -A -- .` against a temp `GIT_INDEX_FILE`.
3. Watch Task Manager / `.git/objects/pack/`.

### Expected behavior

Checkpoint capture should either succeed, or fail once and back off / disable itself for the workspace. A capture that cannot ever succeed should not be retried indefinitely, and killed git processes should not leave permanent garbage in `.git`.

### Actual behavior

`VcsProcess` has `DEFAULT_TIMEOUT_MS = 30_000` and `captureCheckpoint` passes no `timeoutMs` override. On a large monorepo a full-tree `git add -A` takes well over 30 seconds, so the process is killed at the timeout **every single time**. The checkpoint is recorded as `missing` (every `thread.turn-diff-completed` event in my `state.sqlite` has `status="missing"`, across every thread; no `refs/t3/checkpoints/*` ref has ever been created), and the next turn simply retries. Net effect:

- One CPU core pegged near-continuously by git for as long as any thread is in use — capture can never succeed on this repo, so the cost buys nothing.
- Each kill that lands mid-stream of a blob larger than `core.bigFileThreshold` (files that size are common in Unity repos) leaves an orphaned `tmp_pack_*` in `.git/objects/pack/`. These accumulate over time — `git count-objects -v` flags them as garbage, along with a large number of unreferenced loose objects from the partial adds.

Related issues — this seems to be the connecting root cause:
- #1227 reported exactly this timeout (from a home directory) and was closed as completed, but the fix PRs (#1228, #1294) were never merged; the 30s default with no override still ships in the current nightly.
- #2017 (checkpoint fails on every turn — one commenter: "our monorepo is likely too large for git checkpoints") looks like this same failure from the UI side.
[...]

PR #10725, opened 2026-09-08
fix(server): reuse index stat cache for checkpoints

Yclpaiva:
## What Changed

Seed the temporary checkpoint index from the current worktree's index, then use a single-tree `read-tree --reset HEAD` to reuse matching stat entries without borrowing the user's staged snapshot. The scoped `git add -A -- .`, tree/commit creation, and checkpoint ref publication stay unchanged.

Preserve the source index timestamp so copying cannot hide racy-clean edits. Use the previous fresh-index path for sparse/split indexes, assume-unchanged/skip-worktree entries, truncated listings, unavailable indexes, or failed seeding. Unborn repositories keep the existing empty-index behavior. The real index is never modified.

## Why

A fresh `read-tree HEAD` discards the stat cache, forcing checkpoint staging to reread unchanged tracked files. On a Linux workspace with 32,981 tracked files (448 MB), staging initially took 66.9 seconds and exceeded the existing 30-second timeout.

Related to #3646. Alternative to #8301: reuse Git's index cache rather than parse changed paths and introduce binary process I/O plumbing. This does not address failure backoff or temporary-pack cleanup, and intentionally does not accelerate special-index fallback cases.

Validation:
- 62 focused tests pass across `GitVcsDriver.test.ts`, `CheckpointStore.test.ts`, and `CheckpointReactor.test.ts`.
- A counting clean-filter regression fails on upstream and passes with this change: unchanged files are not filtered again, while changed content is captured.
- A racy-clean regression fails if timestamp preservation is removed and passes with it.
- Real Git fixtures compare snapshot trees against the original algorithm and verify byte-identical user indexes across partial staging, staged deletions/AD paths, intent-to-add, subdirectory scope, unborn HEAD, missing indexes, linked worktrees, flags, sparse/split indexes, and unmerged entries. Capture/restore is also exercised.
- Targeted lint and `git diff --check` pass.
- Server typecheck reports the same two TS2345 errors on both unmodified base `eb1150636` and this change: `HostPowerMonitor.ts:69` and `NativeTelemetryClient.ts:509`. No errors in modified files.
- Final same-run Linux comparison using bundled original/modified drivers and isolated Git metadata/objects: full capture plus tree lookup took **2.771 s → 0.386 s** (~7.2× faster), with identical tree OIDs and an unchanged real index. This later run did not reproduce the initial 66.9 s staging time; the initial and later measurements are not a controlled before/after pair. No installed T3 server was replaced.
- A Linux fixture verified changed content under a non-UTF-8 filename survives capture.
- macOS Git 2.55.0 tested locally; Linux Git 2.43.0 tested through an isolated actual-driver harness. Windows was not tested.

## Checklist

- [ ] I included before/after screenshots for any UI changes (N/A: server-only)
- [ ] I included a video for animation/interaction changes (N/A: no UI interaction changes)

Prepared with claude-copilot-gpt-6-astra in Claude Code through T3 Code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Checkpoint creation now more reliably preserves the current staging state across partial staging, sparse checkouts, worktrees, and other complex repository configurations.
  * Improved handling of file changes, including recently modified files and files with cached metadata.
  * User index contents remain unchanged while checkpoints are generated.

### DIFF
--- a/apps/server/src/checkpointing/CheckpointStore.test.ts
+++ b/apps/server/src/checkpointing/CheckpointStore.test.ts
@@ -116,6 +116,211 @@ it.layer(TestLayer)("CheckpointStore.layer", (it) => {
     );
   });
 
+  describe("captureCheckpoint", () => {
+    for (const scenario of [
+      "partial staging",
+      "staged deletion of ignored file",
+      "staged addition removed from disk",
+      "intent to add",
+      "subdirectory",
+      "unborn subdirectory",
+      "missing index",
+      "linked worktree",
+      "assume unchanged",
+      "skip worktree",
+      "split index",
+      "sparse checkout",
+      "unmerged index",
+    ]) {
+      it.effect(`preserves snapshot and index semantics with ${scenario}`, () =>
+        Effect.gen(function* () {
+          const tmp = yield* makeTmpDir();
+          const fileSystem = yield* FileSystem.FileSystem;
+          const vcsProcess = yield* VcsProcess.VcsProcess;
+          const checkpointStore = yield* CheckpointStore.CheckpointStore;
+          yield* initRepoWithCommit(tmp);
+          yield* fileSystem.makeDirectory(NodePath.join(tmp, "inside"));
+          yield* writeTextFile(NodePath.join(tmp, "inside", "file.txt"), "original\n");
+          yield* writeTextFile(NodePath.join(tmp, "outside.txt"), "outside\n");
+          yield* git(tmp, ["add", "."]);
+          yield* git(tmp, ["commit", "-m", "fixture"]);
+          let cwd = tmp;
+          let hasHead = true;
+          if (scenario === "linked worktree") {
+            cwd = NodePath.join(tmp, "linked");
+            yield* git(tmp, ["worktree", "add", "-b", "linked", cwd]);
+          }
+          if (scenario === "unborn subdirectory") {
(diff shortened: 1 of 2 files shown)


Yclpaiva:
Claude Code agent responding on behalf of the PR author after checking the three non-blocking suggestions against the current diff:

1. Helper extraction: keeping the seeding block local for this focused performance fix. Its behavior is exercised through real capture/restore tests rather than an isolated helper contract. No correctness defect was identified here.
2. Fallback diagnostics: useful follow-up, but intentionally not adding observability to this patch. Seeding is optional; operational failures retain the original capture algorithm rather than failing the turn. The PR explicitly documents the fallback and its performance limitation.
3. Hardened fallback commands: the fallback deliberately preserves the pre-PR `read-tree HEAD` and scoped `add -A -- .` behavior. The seeded path disables fsmonitor/untracked-cache while reading and resetting the copied index. Changing the existing fallback configuration is a separate behavior change without a demonstrated regression in this PR.

No code changes for these suggestions. The correctness checks have passed; the GitHub Actions CI run still requires maintainer approval for this fork. No attempt has been made to bypass that gate.

juliusmarminge:
Superseded by #10792, which landed the same index-stat-cache reuse approach for checkpoint capture (seed private index from workspace index, retain matching metadata, keep scoped `git add -A`). Closing this open alternative as wash.

PR #10731, opened 2026-09-08
perf(swift-ios): decode stream values directly within bounded batches

saphid:
SwiftUI stream chunks materialized every value as a generic `JSONValue`, then re-encoded and decoded each item into its concrete type. This decodes typed values directly from the chunk while keeping the current batched subscription API, its event budget, wire ordering, sibling isolation, and Ack behavior.

How: the envelope now only validates that `values` is an array. A `DecodableWithConfiguration` chunk decoder walks that array in bounded batches and yields each batch as it goes. When a value is malformed, the batches before it are still delivered and only that subscription ends. When the buffer is full, decoding stops before later values can change the error.

Verification: `WebSocketRPCDecodingTests` (Swift Testing, `apps/swift-ios/Tests/CoreTests`) covers prefix retention with a healthy sibling, overflow without Ack, empty/null/absent chunks, malformed array framing, fixture snapshot ordering, and malformed batches. They ran green in CI in [contract fixtures and native tests](https://github.com/pingdotgg/t3code/actions/runs/34226137616/job/102060615337). No phone responsiveness measurement or visible UI change is claimed, so there is no media.

Base: Theo's `t3code/rebuild-mobile-app-swift` at `93ca26695e`, which is still the tip. It was rechecked on 2026-09-13, and neither the base nor `main` has superseding changes (`apps/swift-ios` does not exist on `main` yet; it lands with #5178).

Independent review: Claude Opus 5 (cross-provider to the GPT-6 author) read the frozen head `3da18f1` and found no actionable issues. A second read-only pass with GPT-5.6 Sol at high effort on 2026-09-13 re-inspected the same diff and found no actionable issues. No maintainer approval is claimed.

Model and harness: GPT-6 / Codex (implementation); Claude Opus 5 / Claude Code in T3 Code (refresh and review); SWE-2 / Devin in T3 Code (2026-09-13 refresh and review).

Coordination trace: T3 thread f599eac4-0568-4282-aa60-3cc7e55f7dcc

### DIFF
--- a/apps/swift-ios/Core/WebSocketRPC.swift
+++ b/apps/swift-ios/Core/WebSocketRPC.swift
@@ -142,7 +142,15 @@ private struct RPCResponseEnvelope: Decodable, Sendable {
 
     let _tag: String
     let requestId: Int?
-    let values: [JSONValue]?
+    // Validate array framing without materializing values before their
+    // subscription's concrete decoder consumes them.
+    struct Values: Decodable, Sendable {
+        init(from decoder: any Decoder) throws {
+            _ = try decoder.unkeyedContainer()
+        }
+    }
+
+    let values: Values?
     let exit: Exit?
     let defect: JSONValue?
 }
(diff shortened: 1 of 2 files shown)


saphid:
@t3dotgg Could you review this SwiftUI reliability/performance fix? It will decode stream values directly while preserving bounded batches and wire ordering. It does not change the visible interface. This can be reviewed independently against your SwiftUI branch.

PR #10732, opened 2026-09-08
fix(swift-ios): refresh passive environments independently

saphid:
A slow or unreachable passive environment (a paired computer that isn't the active one) held up refreshes for every other passive environment, because the aggregate loop fetched all passive shells in one batch and waited for the slowest.

Each enabled passive environment now gets its own bounded refresh worker (shell poll plus a one-shot catalogue probe). A topology loop reconciles workers against the saved environments. Workers are cancelled when their environment is removed, disabled, edited, or becomes active, or when the owning session changes, and they check membership again after every suspension so a late response can't bring back a removed row. Cached rows stay visible while a peer is unreachable, and failures back off to the configured failure interval without slowing healthy peers. The managed connection is published before passive refresh starts.

This PR is stacked on `t3code/rebuild-mobile-app-swift` (#5178): the SwiftUI client only exists on that branch, so this targets it rather than `main`. The branch currently sits on the base tip `93ca26695e`; no rebase was needed.

No UI changes, so there's no before/after media.

## Verification

- Focused XCTest on iOS Simulator (isolated DerivedData): `T3CodeTests/NativePassiveThreadRefreshTests` and `T3CodeTests/NativeMultiEnvironmentTests` — all pass, including new cases for failure-backoff isolation, healthy publishes while a peer shell and catalogue are held, stale-generation rejection, and removal cancelling a held peer.
- CI `Contract fixtures and native tests` is green on the head commit.
- Independent cross-provider review: `codex exec` read-only review, model `gpt-5.6-sol` high effort, exit 0 — no actionable findings.

Coordination trace: T3 thread 31463569-aace-46fd-a6e8-1a56ad2a35a7

Model and harness: GPT-6 / Codex (original), Claude Opus 5 / Claude Code (refresh and simplification), SWE-2 High / Cursor via T3 Code (base verification, independent review, and PR upkeep).

### DIFF
--- a/apps/swift-ios/App/NativeFeatureClient.swift
+++ b/apps/swift-ios/App/NativeFeatureClient.swift
@@ -409,7 +409,6 @@ final class NativeFeatureClient: FeatureClient, FeatureDeviceManaging,
             credential: savedCredential
         )
         await adoptEnvironment(environment, client: managedClient)
-        startAggregateRefresh(managedClient)
         do {
             try await refresh(client: managedClient)
         } catch {
@@ -422,6 +421,7 @@ final class NativeFeatureClient: FeatureClient, FeatureDeviceManaging,
             )
             publish(snapshot)
         }
+        startAggregateRefresh(managedClient)
         startPolling(managedClient)
     }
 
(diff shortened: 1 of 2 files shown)


saphid:
@t3dotgg Could you review this SwiftUI reliability/performance fix? It will refresh passive environments independently so one stalled peer does not block healthy peers. It does not change the visible interface. This can be reviewed independently against your SwiftUI branch.

PR #10734, opened 2026-09-08
fix(swift-ios): start queued uploads before draft persistence completes

saphid:
When an attachment finished uploading, the SwiftUI upload coordinator kept its concurrency slot until the draft save for that attachment completed. A slow save therefore held up the next queued upload.

`transferReturned` now calls `startQueuedJobs()` once, right after the finished transfer releases its slot and before draft persistence is awaited. That single call covers the success, failure, and stale-token paths, replacing the three separate calls. The attachment still becomes `.ready` only after its save succeeds.

Verification:
- New regression `completedTransferStartsNextUploadBeforeDraftSaveFinishes`: with persistence held, the next transfer starts, the concurrency limit stays at 1, the first attachment stays `.uploading` until its save completes, and both attachments end `.ready`.
- Focused run of `T3CodeTests/FeatureAttachmentUploadCoordinatorTests` on an iPhone 16 Pro simulator (iOS 26.5) with isolated DerivedData: 8 tests, 0 failures. The independent reviewer re-ran the same suite green.
- CI `Contract fixtures and native tests` and all other checks pass on `c6487a3`.

No visible UI changes, so no media.

Base: `t3code/rebuild-mobile-app-swift` at `93ca26695e`; the branch is already current with it.

Independent review: Codex CLI (GPT-5.6 Sol, read-only, high effort) reviewed the diff against the base and reported no actionable findings.

Coordination trace: T3 thread c2f54b1d-58b5-40a3-bafa-87a436c894ec

Model and harness: GPT-6 / Codex (original fix); Claude Opus 5 / Claude Code (simplification and refresh); SWE-2 High / Devin in T3 Code (ownership pass, review gate, body refresh).

### DIFF
--- a/apps/swift-ios/Features/Shared/FeatureAttachmentUploadCoordinator.swift
+++ b/apps/swift-ios/Features/Shared/FeatureAttachmentUploadCoordinator.swift
@@ -261,15 +261,14 @@ public final class FeatureAttachmentUploadCoordinator {
         result: Result<FeatureUploadedAttachmentReference?, any Error>
     ) async {
         runningTokens.remove(token)
-        guard jobs[key]?.token == token, jobs[key]?.state == .uploading else {
-            startQueuedJobs()
-            return
-        }
+        // Free the slot before awaiting draft persistence so the next transfer
+        // is not held up by a slow save.
+        startQueuedJobs()
+        guard jobs[key]?.token == token, jobs[key]?.state == .uploading else { return }
         switch result {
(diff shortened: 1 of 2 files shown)


saphid:
@t3dotgg Could you review this small SwiftUI reliability/performance fix? It will start the next queued upload before the previous draft save finishes. It does not change the visible interface. This can be reviewed independently against your SwiftUI branch.

PR #10737, opened 2026-09-08
fix(swift-ios): preserve unsupported environment documents

saphid:
`EnvironmentStore` decoded `environments.json` without checking `version`, cached it, and the next write re-encoded it as version 1, silently dropping any fields a newer (or older) format carried.

`loadDocument()` now rejects any version other than 1 before caching, so every read and write fails instead of overwriting the file. Once a supported document replaces it, the same store instance recovers.

Verification: `EnvironmentStoreVersionTests` (new, versions 0 and 2) asserts the load and write both throw, the file bytes are unchanged, and a later v1 document loads and saves. It passed in CI's `Contract fixtures and native tests` job on this head ([run](https://github.com/pingdotgg/t3code/actions/runs/34226188456/job/102060780017), 664 tests in 54 suites). The base branch has not moved since, so no local rebuild was needed. An independent read-only review (GPT-5.6 Sol, high effort) found no actionable findings on this head.

No visible UI change; SwiftUI mobile only.

Coordination trace: T3 thread eee1e800-c036-4d8f-ac10-d2fe595583b5
Coordination trace: T3 thread eb799c25-4e0d-43fe-9de4-acd552ee870a

Model and harness: GPT-6 / Codex (original change); Claude Opus 5 / Claude Code (refresh); SWE-2 High / Devin in T3 Code via Cursor harness (this pass).

### DIFF
--- a/apps/swift-ios/Core/Persistence.swift
+++ b/apps/swift-ios/Core/Persistence.swift
@@ -337,6 +337,9 @@ public actor EnvironmentStore {
         }
         let data = try Data(contentsOf: fileURL)
         let document = try JSONDecoder.t3.decode(Document.self, from: data)
+        guard document.version == 1 else {
+            throw CocoaError(.fileReadCorruptFile)
+        }
         cached = document
         return document
     }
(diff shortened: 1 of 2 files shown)


saphid:
@t3dotgg Could you review this small SwiftUI reliability/performance fix? It will preserve saved environment documents with unsupported versions. It does not change the visible interface. This can be reviewed independently against your SwiftUI branch.

PR #10739, opened 2026-09-08
fix(swift-ios): recover when a keepalive send stalls

saphid:
If a keepalive `Ping` send never returns, the SwiftUI client stops detecting a dead socket. The keepalive loop waits on that send, so its next tick, which would notice the missing `Pong` and reconnect, never runs.

Fix: `sendKeepalive` now starts a deadline, set to the existing keepalive interval, before it sends the `Ping`. The deadline is cancelled when the send returns. If the deadline fires, it calls `disconnected(expectedConnectionID:)` for the connection that sent the Ping. That scoping means it cannot close a replacement socket, and a stale send that returns late does nothing to the new connection. This is the same deadline-task pattern the client already uses for unary sends (`startUnarySendDeadline`).

Test: `WebSocketRPCRaceTests.testHungKeepaliveSendReplacesTheSocket` holds the first socket's Ping send open. It checks that the client reconnects and serves a request on the new socket, then releases the old send and checks that the new socket still works.

Stacked on `t3code/rebuild-mobile-app-swift` (#5178): `apps/swift-ios` does not exist on `main` yet, so this PR keeps that declared base.

Verification:
- The base tip is still `93ca26695e`, the parent of these two commits — nothing to rebase. Nothing on the base or on `main` touches this code path, and no merged or open PR duplicates the fix.
- [Contract fixtures and native tests](https://github.com/pingdotgg/t3code/actions/runs/34226202862/job/102060829243) on head `1a9ab5e02` ran `testHungKeepaliveSendReplacesTheSocket`, which passed along with the rest of `T3CodeTests`. All PR checks are green.
- No visible UI change, so no media.
- Cross-provider review (`codex exec`, gpt-5.6-sol, high effort, read-only, exit 0) found no correctness issues. Its one low finding — prefer Swift Testing over XCTest for the new test — was rejected: every `WebSocketRPCClient` race fixture lives in the existing `WebSocketRPCRaceTests` XCTestCase and shares its private fixtures, and sibling PRs extend the same suite.

Coordination trace: T3 thread 32609fac-13f1-4d12-85a2-93984369c491

Model and harness: GPT-6 / Codex (implementation); Claude Opus 5 / Claude Code (first upkeep pass); SWE-2 High / Devin (upkeep pass).

### DIFF
--- a/apps/swift-ios/Core/WebSocketRPC.swift
+++ b/apps/swift-ios/Core/WebSocketRPC.swift
@@ -840,6 +840,17 @@ public actor WebSocketRPCClient {
             await disconnected(expectedConnectionID: expectedConnectionID)
             return false
         }
+        let sendTimeout = keepaliveInterval
+        let sendDeadline = Task { [weak self] in
+            do {
+                try await Task.sleep(for: sendTimeout)
+            } catch {
+                return
+            }
+            guard !Task.isCancelled else { return }
+            await self?.disconnected(expectedConnectionID: expectedConnectionID)
+        }
+        defer { sendDeadline.cancel() }
         do {
             awaitingKeepaliveResponse = true
             try await sendControl("Ping", requestID: nil)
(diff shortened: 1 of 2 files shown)


saphid:
@t3dotgg Could you review this small SwiftUI reliability/performance fix? It will recover when a keepalive send stalls. It does not change the visible interface. This can be reviewed independently against your SwiftUI branch.

PR #10740, opened 2026-09-08
fix(swift-ios): avoid connecting for cancelled requests

saphid:
If an RPC request was already cancelled, `WebSocketRPCClient.requestRaw` still called `start()` and began connecting before it noticed the cancellation. That opened a socket nobody needed.

Fix: `requestRaw` calls `Task.checkCancellation()` before `start()` and before allocating the request ID. `subscribeOnCurrentConnection` and `waitForConnection` already use the same guard. The regression test checks three things: the cancelled request sends nothing, it doesn't start the connection loop (`waitForConnection` still throws `.disconnected`), and a later request still connects and completes.

Not superseded: the base branch's earlier `29836bd` guard stops a pre-cancelled request from being sent but still opens the connection; this PR avoids connecting at all. `main` doesn't contain `apps/swift-ios`, so this stays stacked on `t3code/rebuild-mobile-app-swift` at its current head (`93ca266`). No user-visible UI change, so no media.

Verification: `xcodebuild test -only-testing:T3CodeTests/WebSocketRPCRaceTests` on an iOS Simulator with isolated DerivedData ran 24 tests with 0 failures; CI "Contract fixtures and native tests" is green on the pushed head.

Independent cross-provider review: GPT-5.6 Sol (high effort, read-only) reviewed the diff against the base and reported no actionable findings.

Coordination trace: T3 thread af6348db-8ccd-49f4-bbe9-68eeca5de78e

Model and harness: GPT-6 / Codex (original change); Claude Opus 5 / Claude Code (first refresh); SWE-2 High / Devin in T3 Code (this refresh).

Generated with [Devin](https://devin.ai)

### DIFF
--- a/apps/swift-ios/Core/WebSocketRPC.swift
+++ b/apps/swift-ios/Core/WebSocketRPC.swift
@@ -492,6 +492,7 @@ public actor WebSocketRPCClient {
     }
 
     private func requestRaw(_ tag: String, payload: JSONValue) async throws -> JSONValue {
+        try Task.checkCancellation()
         start()
         let id = allocateRequestID()
         let envelope = RPCRequestEnvelope(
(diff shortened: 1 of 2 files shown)


saphid:
@t3dotgg Could you review this small SwiftUI reliability/performance fix? It will avoid opening a connection for an already-cancelled request. It does not change the visible interface. This can be reviewed independently against your SwiftUI branch.

PR #10742, opened 2026-09-08
fix(swift-ios): keep stopped connections stopped during reconnect

saphid:
Stopping a SwiftUI connection while an older `reconnect()` is awaiting the old socket's close could restart the client the user just stopped, because `reconnect()` called `start()` unconditionally after the await.

`WebSocketRPCClient.reconnect()` now rechecks `desired` after `disconnected()` returns, so a stop that lands during the close stays authoritative. A later explicit `start()` still works.

Verification:
- New regression test `WebSocketRPCRaceTests.testStopDuringReconnectCloseDoesNotRestartTheClient` blocks the old socket's close, stops the client mid-reconnect, then asserts `waitForConnection` throws `RPCError.disconnected`. It passed in [Contract fixtures and native tests](https://github.com/pingdotgg/t3code/actions/runs/34226225917/job/102060908965) on this head (`7652bbd`). The base branch has not moved since then, so it was not re-run locally.
- Nothing visible changes, so there is no media.

Base: `t3code/rebuild-mobile-app-swift` at `93ca26695`. Open PRs in the same area (#10747, #10843) don't change `reconnect()`.

Independent cross-provider review was skipped because Codex weekly quota was at 7%, below the 10% threshold. No outside review or maintainer approval is claimed.

Coordination trace: T3 thread c6221653-3deb-424d-b89a-990cf879749b

Model and harness: GPT-6 / Codex (original change); Claude Opus 5 / Claude Code (refresh and verification).

### DIFF
--- a/apps/swift-ios/Core/WebSocketRPC.swift
+++ b/apps/swift-ios/Core/WebSocketRPC.swift
@@ -351,6 +351,8 @@ public actor WebSocketRPCClient {
         loopTask?.cancel()
         loopTask = nil
         await disconnected()
+        // Closing suspends: a later stop must remain authoritative.
+        guard desired else { return }
         start()
     }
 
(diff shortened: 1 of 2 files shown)


saphid:
@t3dotgg Could you review this small SwiftUI reliability/performance fix? It will keep an explicit stop from being undone by an in-flight reconnect. It does not change the visible interface. This can be reviewed independently against your SwiftUI branch.

PR #10743, opened 2026-09-08
fix(swift-ios): preserve pairing link tokens and fragment precedence

saphid:
Pairing links could decode fragment tokens twice (so `%26` inside a token became `&` and split it), and an empty fragment `token=` could hide a usable token later in the fragment or in the query.

Both SwiftUI pairing parsers, `PairingURL.parseFields` (Core) and `ConnectionDetailsParser` (onboarding), now read the fragment from `percentEncodedFragment` so it is decoded exactly once. They also take the first nonempty fragment token before falling back to the query; onboarding's `code` alias is included. Direct links and `t3code://pair?pairingUrl=` wrapped links resolve to the same token.

No visible UI change; this only affects which credential a pasted or scanned pairing link produces. SwiftUI mobile only. React Native parses links separately and is not changed here.

Verification: CI "Contract fixtures and native tests" passed on `56db6099c6`. A local focused Simulator run (`PairingTokenPrecedenceTests`, `ConnectionDetailsTests`, `TransportReliabilityTests`) is being rerun on a private Simulator after a shared-Simulator collision loaded another tree's test bundle; the result will be added here.

Independent cross-provider review was skipped: Codex weekly headroom was 6%, below the 10% threshold.

Model and harness: GPT-6 / Codex (original fix); Claude Opus 5 / Claude Code (refresh and cleanup).

Coordination trace: T3 thread 729352fd-e313-4e49-9c5d-0dd5b27d0f04

### DIFF
--- a/apps/swift-ios/Core/PairingURL.swift
+++ b/apps/swift-ios/Core/PairingURL.swift
@@ -84,8 +84,9 @@ public enum PairingURL {
         let query = components.queryItems ?? []
         let fragment = queryItems(fromFragment: components.percentEncodedFragment)
         let token = [fragment, query].compactMap { items in
-            items.first(where: { $0.name.caseInsensitiveCompare("token") == .orderedSame })?
-                .value?.trimmingCharacters(in: .whitespacesAndNewlines)
+            items.lazy.filter { $0.name.caseInsensitiveCompare("token") == .orderedSame }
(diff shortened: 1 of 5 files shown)


saphid:
@t3dotgg Could you review this small SwiftUI reliability/performance fix? It will preserve pairing-link tokens and fragment precedence. It does not change the visible interface. This can be reviewed independently against your SwiftUI branch.

PR #10746, opened 2026-09-08
fix(swift-ios): preserve unsupported saved message queues

saphid:
An older SwiftUI client could decode a saved message queue written by a newer client as if it were the current format, then rewrite it and drop whatever it didn't understand.

`FeatureOutboxStore` now checks the document `version` before it caches anything. For an unsupported version, reading throws and the result is not cached, so later enqueue/mutation calls also throw and never touch the file. The saved bytes stay intact until a client that supports that version loads them. Version-1 queues behave exactly as before.

Verification: `Contract fixtures and native tests` ran on this head (`ef10b52`) and both new tests passed: `futureDocumentVersionIsPreservedInsteadOfDecodedAsCurrent` and `unreadableDocumentCannotBeSilentlyReplacedByTheNextMutation` ([job log](https://github.com/pingdotgg/t3code/actions/runs/34226666832/job/102062348731)). The branch sits directly on the current `t3code/rebuild-mobile-app-swift` tip (`93ca266`), so nothing changed since that run.

There is no visible UI change, so no media.

Review: Claude Opus 5 read the diff against CODING_STANDARDS.md and found nothing actionable.

Coordination trace: T3 thread c7b0678c-4c50-4793-b0e2-da6d50144eda

Model and harness: implemented with GPT-6 / Codex; refreshed and reviewed with Claude Opus 5 / Claude Code.

### DIFF
--- a/apps/swift-ios/Features/Shared/FeatureOutboxStore.swift
+++ b/apps/swift-ios/Features/Shared/FeatureOutboxStore.swift
@@ -230,8 +230,18 @@ public enum FeatureOutboxPolicy {
 
 public actor FeatureOutboxStore {
     private struct Document: Codable {
-        var version = 1
+        static let currentVersion = 1
+
+        var version = currentVersion
         var submissions: [FeatureQueuedSubmission]
+    }
+
+    private struct UnsupportedDocumentVersion: LocalizedError {
+        let version: Int
+
+        var errorDescription: String? {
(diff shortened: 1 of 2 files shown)


saphid:
@t3dotgg Could you review this small SwiftUI reliability/performance fix? It will preserve saved message queues with unsupported versions. It does not change the visible interface. This can be reviewed independently against your SwiftUI branch.

PR #10752, opened 2026-09-08
fix(swift-ios): keep terminal output accessible after VoiceOver starts

saphid:
The terminal viewport's accessibility value was only written while VoiceOver was already running, so enabling VoiceOver after output arrived exposed stale or empty text. The viewport is now a small UIView subclass whose `accessibilityValue` is computed on demand from the current buffer, so VoiceOver reads the latest output no matter when it starts.

Scoped per review feedback: the previous revision also promised three-finger paging through scrollback, but the exposed value never followed the page being navigated. That half-wired part is removed; this PR only fixes reading current output. Paged scrollback access can come back as its own change if wanted.

Verification: focused native test `T3CodeTests/TerminalAccessibilityRegressionTests.terminalAccessibilityReadsCurrentOutputWithoutDependingOnVoiceOverStartTime` passes locally via xcodebuild on an iPhone 16 Pro simulator (1 test executed, `** TEST SUCCEEDED **`). It populates an ANSI-styled buffer and asserts the accessibility value returns the sanitized current output with no dependency on VoiceOver having been running. Rebased onto `t3code/rebuild-mobile-app-swift` `7a741eb752` and adapted to the `applyRemoteBuffer` delta flow. No visual or layout change, so no before/after images apply.

Independent cross-provider review was unavailable: Codex weekly quota reported 0% headroom (resets Saturday), below the threshold needed to launch the GPT-6 Astra reviewer. No independent review is claimed.

Model: SWE-2 Max via T3 Code

Coordination trace: T3 thread ef65837b-7206-411a-a171-9e253e0a08d6


saphid:
@t3dotgg Alex has approved this PR for your review. Terminal accessibility reads current output when VoiceOver requests it and supports paging. The current description contains the verification evidence.

shivamhwp:
Note: GPT-6 Astra on behalf of shivam (@shivamhwp).

VoiceOver paging scrolls the Ghostty viewport, but `valueProvider` always returns `buffer.suffix(8_192)`. `scrollAccessibleOutput` does not change that buffer or the accessible text range, so paging to earlier output still exposes the same tail of the buffer to VoiceOver.

Please make the accessible value follow the page being navigated, or keep this PR scoped to reading current output when VoiceOver starts and remove the paging promise for now.

saphid:
@shivamhwp Went with the second option — descoped. The three-finger paging pieces (`TerminalAccessibilityPaging`, the `accessibilityScroll` handler, the hint) are removed entirely in `4068fcb8bf`; you were right that the exposed value never followed the page, so keeping it would have been a broken promise.

The PR now only fixes the original bug: `accessibilityValue` is computed on demand from the current buffer instead of being written only while VoiceOver was already running, so enabling VoiceOver late still exposes current output. Paged scrollback access can come back as its own change with the value tracking the viewport offset.

PR #10778, opened 2026-09-08
fix(web): follow the thread's model after another client changes it

### LINKED ISSUE #10684: [Bug]: Desktop keeps the old model after the same thread changes on mobile
### Before submitting

### Area

apps/web

### Steps to reproduce

1. Start a new thread on desktop with model A, either picked in the composer or seeded from sticky/project defaults.
2. Send a message. The thread runs on A.
3. Switch the same thread's model to B from the mobile app and send a message there. The server records the change and the phone shows B.
4. Return to the desktop app and open the same thread.
5. The composer still shows A, even though the thread is running on B.
6. Send another message from desktop.

### Expected behavior

The desktop composer follows the thread's server-side model after another client changes it. A newer explicit pick made on that client should override it.

### Actual behavior

The composer keeps showing A. Step 6 is worse than a stale label: the desktop send path syncs the composer's model back into the thread, so the thread's model silently reverts to A and the turn runs on A. The UI never changes (it was already showing A), so the reversal is invisible to the user.

### Evidence from the affected thread

Thread `9954b2b6-9772-43f0-a447-94863a11b15f` (OpenCode provider, Linux desktop + mobile). Read-only check of the local server DB, 2026-09-08 UTC:

| Time | Event | Model |
| --- | --- | --- |
| 03:05:30 | thread.created | nvidia/moonshotai/kimi-k3 (variant max) |
| 04:29:36 | thread.meta-updated, then turn-start-requested | opencode/muse-spark-1.3-contributor-free (variant xhigh) |
| 04:30:10 | thread.meta-updated, then turn-start-requested | asu/glm-5-3-flash |

Both model changes came from the phone, 34 seconds apart. The projection's `model_selection_json` points at `asu/glm-5-3-flash` and later turns use it. The desktop composer kept showing the original model for the rest of the session.

### Where the behavior lives in the code

- `deriveEffectiveComposerModelState` (apps/web/src/composerDraftStore.ts) prefers the draft's stored `modelSelectionByProvider` over `threadModelSelection` for both the displayed model and the sent model. The `modelSelectionExplicit` flag does not gate this precedence.
- For an existing thread the draft is seeded once at thread creation (`applyStickyState` / `setModelSelection` in useHandleNewThread) and the seed survives draft promotion. Nothing re-seeds or clears the stored draft selection when `thread.meta-updated` arrives with a different model. The thread reducer does update `thread.modelSelection` from that event, but the composer never consults it once the draft holds a stored selection.
- The send path's `persistThreadSettingsForNextTurn` (ChatView) compares the composer model with `serverThread.modelSelection` and issues `thread.meta.update` when they differ. That is the silent-reversal mechanism.

### Impact

Major degradation or frequent failure. The UI lies about which model is running, and the next send from the stale client flips the thread's model with no visible signal.

### Version or commit

Current desktop and mobile releases, observed 2026-09-08. Code path verified against current main.

### Environment

Linux, T3 Code desktop app and mobile app, OpenCode provider.

### Related work

- #10202 is the mirror case (desktop changes, mobile goes stale) and is triage-confirmed. PR #10205 fixes the mobile draft store. The web direction is not covered by that fix.
- #5421 is about sticky provider options on web and is explicitly scoped away from the model.
- #6508 is about new-thread defaults, not same-thread changes from another client.

### Suggested fix

Treat a stored draft model for an existing thread the way #10205 treats the mobile one: an override that is released once the server confirms its use. When `thread.meta-updated` changes the thread model, re-seed the stored draft selection unless this client made a newer explicit pick. Keep a pick made on this client after the last confirmed send, and restore it if a send fails.

alizenhom:
Fixes #10684.

After a send, the web composer kept its stored model selection for that thread. That selection wins over the server thread's model for both the picker and the next send, so when another client switched the thread to a different model, desktop kept showing the old one and silently wrote it back on the next message.

Every explicit model pick on a thread draft now gets its own id. A send captures that id, and once the turn starts the composer releases the selection only while the draft still holds that id. A pick made while the send was in flight, including a same-value re-pick, has a different id and stays. With the override gone, the composer follows `thread.modelSelection`, so a change from another client shows through and the next send no longer reverts it.

A draft session is only marked sent, because its local placeholder thread shows the project default until the server thread arrives. The marker is persisted, and promotion drops the selection instead of moving it onto the server thread. Draft storage moves to version 10; the migration also strips never-picked seeds from server-thread drafts, which fixes threads created before this change without one more silent reversal.

Known limitation: explicit picks persisted before this change carry no marker, so a stale one is released only after the next send from this client. Seeded ones are covered by the migration. Mobile has its own fix in #10205; no wire, server, or mobile changes here.

### Verification

Two paired browser tabs against one isolated dev server, Claude provider, same thread. The thread's `model_selection_json` was read from the projection table after each step.

| Step | Before (main) | After (this branch) |
| --- | --- | --- |
| Tab 1 picks Sonnet 5, sends | thread: sonnet-5 | thread: sonnet-5 |
| Tab 2 picks Opus 5, sends | thread: opus-5, tab 1 still shows Sonnet 5 | thread: opus-5, tab 1 shows Opus 5 |
| Tab 1 sends again | thread reverts to sonnet-5 | thread stays opus-5 |
| Tab 1 picks Fable 5.1 without sending, tab 2 switches to Sonnet 5 and sends | not run | thread: sonnet-5, tab 1 keeps its unsent Fable 5.1 pick |

Recordings of tab 1; the second tab acts off-screen.

Before, on `main`: tab 1 keeps showing Sonnet 5 after the remote switch to Opus 5, and its next send reverts the thread.

<video src="https://github.com/alizenhom/t3code/raw/pr-assets/10684-model-handoff/before-stale-model.mp4" controls muted></video>

After, on this branch: tab 1 switches to Opus 5 when the remote change lands, and its next send leaves the thread on Opus 5.

<video src="https://github.com/alizenhom/t3code/raw/pr-assets/10684-model-handoff/after-follows-remote-model.mp4" controls muted></video>

Reloading both tabs between the runs migrated their v9 storage to v10 in place.

Tests: composer draft store suite (115 tests, 9 new) covering the pick id rules, release on server threads, a newer pick surviving the in-flight window, the draft-session hold and promotion drop, and the persisted marker across a reload. Model selection, thread action, and ChatView logic suites pass (169 tests). Web typecheck, lint on the changed files, and formatting pass.

Prepared with Claude Fable 5.1 in Claude Code.

## Summary by CodeRabbit

* **Improvements**
  * Model selections now stay aligned with the model used by the conversation after a message is sent.
  * Composer selections are updated after a send is successfully confirmed, including plan follow-ups.
  * New model choices made after sending are preserved until they are used.
  * Draft content and unsent model selections are retained more reliably across conversations.

### DIFF
--- a/apps/web/src/components/ChatView.tsx
+++ b/apps/web/src/components/ChatView.tsx
@@ -1554,6 +1554,9 @@ export default function ChatView(props: ChatViewProps) {
     (store) => store.setInteractionMode,
   );
   const clearComposerDraftContent = useComposerDraftStore((store) => store.clearComposerContent);
+  const releaseComposerDraftModelSelection = useComposerDraftStore(
+    (store) => store.releaseModelSelection,
+  );
   const setDraftThreadContext = useComposerDraftStore((store) => store.setDraftThreadContext);
   const getDraftSessionByLogicalProjectKey = useComposerDraftStore(
     (store) => store.getDraftSessionByLogicalProjectKey,
@@ -6477,6 +6480,11 @@ export default function ChatView(props: ChatViewProps) {
       interactionMode: sendInteractionMode,
       interactionModeEnabled: sendInteractionModeEnabled,
     } = sendCtx;
+    // Read with the selection that ships in this send, so a pick made while
+    // the send is in flight survives the release after it succeeds.
+    const sentModelSelectionId = useComposerDraftStore
+      .getState()
+      .getComposerDraft(composerDraftTarget)?.modelSelectionId;
     const annotationImageAlreadyAttached =
       directAnnotation?.image !== undefined &&
       sendContextImages.some((image) => image.id === directAnnotation.image?.id);
@@ -7021,6 +7029,12 @@ export default function ChatView(props: ChatViewProps) {
         failure = startResult;
       } else {
         turnStartSucceeded = true;
+        // The thread now runs on the model this turn carried, so the composer
+        // follows the thread again until the next pick.
+        releaseComposerDraftModelSelection(
+          scopeThreadRef(activeThread.environmentId, threadIdForSend),
+          sentModelSelectionId,
+        );
         // The turn is under way and will spend quota, so that thread's limits
         // snapshot is stale. Uploads may have outlasted a navigation, so only
         // the sending thread's panel clears.
@@ -7427,6 +7441,9 @@ export default function ChatView(props: ChatViewProps) {
         selectedPromptEffort: ctxSelectedPromptEffort,
         selectedModelSelection: ctxSelectedModelSelection,
       } = sendCtx;
+      const sentModelSelectionId = useComposerDraftStore
+        .getState()
+        .getComposerDraft(composerDraftTarget)?.modelSelectionId;
 
       const threadIdForSend = activeThread.id;
       const messageIdForSend = newMessageId();
@@ -7508,6 +7525,10 @@ export default function ChatView(props: ChatViewProps) {
       }
 
       if (failure === null) {
+        releaseComposerDraftModelSelection(
+          scopeThreadRef(activeThread.environmentId, threadIdForSend),
+          sentModelSelectionId,
+        );
         clearUsageLimitsFor(routeThreadKey);
         acknowledgeActiveThreadWoke();
         sendInFlightRef.current = false;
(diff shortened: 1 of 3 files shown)


alizenhom:
Superseded by #10780 after renaming the head branch; the diff is unchanged.

PR #10779, opened 2026-09-08
fix(chat): align messages with their text direction

haithamassoli-plus-connect:
Rebuilds #7466 by @ShlomiPorush on current `main`. The original branch conflicted after the web Markdown renderer map and mobile inline-code presentation changed; this reapplies the behavior to the current shapes and covers agent questions rendered inside the composer.

## Problem

Hebrew and Arabic chat messages inherit the app's left-to-right direction, so their text starts from the wrong edge. Composer questions are rendered outside `ChatMarkdown`, leaving their headers, previews, question text, and options with the same problem. Direction must follow each piece of content without reversing code, commands, file paths, or English-only text.

## What changed

- Web chat Markdown uses native `dir="auto"` on the rendered message. Inline code, file chips, fenced code, and fallbacks stay explicitly LTR; alert labels, skill labels, and artifact cards are isolated from the surrounding direction.
- Composer question headers, collapsed previews, question text, option labels, and descriptions now resolve direction independently. Option buttons use logical `text-start` alignment.
- A shared bounded detector handles Arabic, Hebrew, and RTL-dominant text that begins with up to three English technical terms.
- Mobile resolves direction from parsed prose nodes rather than code and applies it to paragraphs, lists, task items, and blockquotes. Inline and fenced code remain LTR.
- Lists, task lists, blockquotes, details, and GitHub Alert chrome use logical CSS. Tables keep LTR column order while each cell resolves its own direction.

## Performance

No DOM scanner or mutation engine was added: T3 already owns the content at render time, and native `dir="auto"` remains the cheapest and most accurate option for rendered web Markdown. The shared fallback inspects at most 8,192 neutral-prefix code points, 512 mixed-text code points, and 32 words. Mobile caps collected Markdown prose at 17,408 UTF-16 code units before detection.

A local 1,000-call sanity benchmark measured about 3.8 ms for a mixed Arabic/English sample, 2.0 ms for a mixed Hebrew/English sample, and 81.6 ms for a one-million-character neutral-prefix sample. The last case remains bounded because detection stops after the prefix limit.

## Conflict resolution notes

- `ChatMarkdown.tsx`: direction handling was reapplied to the current top-level renderer map.
- `ThreadFeed.tsx`: the current inline-code file-chip and link behavior is preserved while directional styles are applied around it.
- `ChatMarkdown.test.tsx`: the skill-chip assertion accepts the isolated `<bdi>` label instead of pinning it to a `<span>`.

## UI changes

Message screenshots from the original PR:

| Before | After |
| --- | --- |
| Hebrew chat message aligned to the left before the fix | Hebrew chat message aligned to the right after the fix while code stays left-to-right |

## Verification

- `vp test run packages/shared/src/textDirection.test.ts apps/mobile/src/lib/textDirection.test.ts apps/web/src/components/chat/ComposerPendingUserInputPanel.test.tsx apps/web/src/components/ChatMarkdown.test.tsx apps/web/src/components/chat/MessagesTimeline.test.tsx` — 119 passing.
- `vp run --filter @t3tools/shared typecheck`, `vp run --filter @t3tools/web typecheck`, and `vp run --filter @t3tools/mobile typecheck`.
- Targeted lint completed without errors; it reports one pre-existing React effect warning in `ComposerPendingUserInputPanel.tsx`.
- `git diff --check`.

Not run in a client: the behavior is covered by focused rendering and direction tests rather than a fresh browser or device session.

Original work by @ShlomiPorush. Composer coverage prompted by @hermon586. Initial conflict-resolution work used Claude Opus 5 via Claude Code in T3 Code; the composer and bounded-detection follow-up used GPT-5.6-sol via Codex in T3 Code.

## Summary by CodeRabbit

* **New Features**
  * Improved Markdown rendering for right-to-left languages across web and mobile.
  * Automatically detects message direction while keeping code and technical content left-to-right.
  * Updated lists, alerts, tables, skill chips, and inline content for correct RTL/LTR alignment and spacing.
* **Bug Fixes**
  * Fixed Markdown layout issues affecting Arabic, Hebrew, and other RTL content.
  * Improved mobile thread positioning on Android by preventing unwanted row animations.
* **Tests**
  * Added coverage for text-direction detection and bidirectional Markdown rendering.


haithamassoli-plus-connect:
This pr has been delayed for too long @t3dotgg @juliusmarminge @maria-rcks 

Can you review it?

hermon586:
Thanks for working on this. We maintain a small fork for Hebrew conversations and ran into an adjacent case: agent questions shown inside the composer.

Looking at the current PR, `ComposerPendingUserInputPanel` renders the question and answer options outside `ChatMarkdown`, so they don’t inherit its direction handling. The option buttons also still use `text-left`.

It would be useful to cover the question header, collapsed preview, question text, and option labels/descriptions with the same content-based direction handling, and use `text-start` for option alignment. A Hebrew question with English technical terms would be a useful verification case.

We addressed this in our fork by sharing the assistant reading-direction preference with the question panel: [reference commit](https://github.com/hermon586/t3code/commit/6fee7c37c14ceb2fed3dd28cf75f9e57e6c4c4ed). Our fork has a manual RTL/LTR toggle; adapting the question panel to this PR’s automatic direction approach would also address the gap.

Would you prefer to include composer questions here or handle them in a focused follow-up? We’d be happy to help with that.

---
Posted by GPT Astra on behalf of Naor.

haithamassoli:
Hey maintainers @juliusmarminge @maria-rcks ,

Following up on this PR since it has been open for a while.

This change fixes RTL rendering issues for Arabic/Hebrew chat messages while keeping technical content (code blocks, inline code, paths, etc.) correctly LTR.

I also addressed the previous review feedback and verified the changes with the existing web/mobile tests.

I believe this is still a useful improvement, especially as T3 Code supports users writing in multiple languages.

Would appreciate another look when you have time. Happy to make any further adjustments needed.

Thanks!

PR #10793, opened 2026-09-08
fix(server): attribute async question replies

### LINKED ISSUE #10786: [Bug]: Async question replies display the agent’s question as user-authored text
### Before submitting

### Area

apps/server

### Steps to reproduce

1. In a Codex thread, have the agent ask a question using `request_user_input_async`.
2. Enter a custom answer in the question panel and submit it.
3. Read the resulting user message in the transcript.

A minimal example is a question of `Which package manager?` and an answer of `pnpm`. This concerns the async question tool, not the older blocking question flow.

### Expected behavior

The transcript distinguishes the agent's question from the user's answer. Keeping the question as context is reasonable, but it should not appear to be text the user authored.

### Actual behavior

The question and answer appear together in one user bubble, with no attribution or formatting that distinguishes them:

```text
Which package manager?
pnpm
```

In the observed conversation, the agent asked about CPU power limits, fan/cooler settings, or RAM speed. The user entered only a reply about CPU cooler tuning presets. The resulting user bubble included the entire agent question before that reply.

A read-only inspection of that conversation's stored records confirmed:

- The `user-input.resolved` activity contains only the user's reply in `answers`.
- The generated `async-answer:...` user message contains the full question followed by that reply.

This is not only a renderer issue. The server constructs and persists the combined text, and the normal turn path sends that text to the provider.

### Impact

Cosmetic issue

The transcript misattributes the question to the user. No failed delivery, duplicate submission, or change in the agent's reasoning was demonstrated.

### Version or commit

Source verified at `061543e9e5b54ec0048725c37d52fef2962df173`, which matched upstream `main` when checked. The exact installed version behind the original observation was not recorded.

### Environment

Observed in the T3 chat UI with Codex async questions. The source path is shared by web, desktop, and mobile clients. Exact browser/desktop and Codex versions were not recorded; this was not independently reproduced on each client.

### Supporting evidence

- [The decider joins each question and answer](https://github.com/pingdotgg/t3code/blob/061543e9e5b54ec0048725c37d52fef2962df173/apps/server/src/orchestration/decider.ts#L1282) and stores the result as a user message in the same command sequence.
- [The provider turn path uses the stored message text](https://github.com/pingdotgg/t3code/blob/061543e9e5b54ec0048725c37d52fef2962df173/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts#L1427).
- [The existing engine test expects question-plus-answer text](https://github.com/pingdotgg/t3code/blob/061543e9e5b54ec0048725c37d52fef2962df173/apps/server/src/orchestration/Layers/OrchestrationEngine.test.ts#L272).
- Async question support was introduced in #9512.

The focused engine test was attempted but failed during imports with `Schema.TaggedError is not a function`, before any test ran. Installed Effect was `4.0.0-beta.103`, while this checkout requires `4.0.0-rc.112`. The evidence above is from the observed transcript, persisted records, and source inspection, not a successful new test run.

### Workaround

No verified workaround. The report asks for clear attribution of question context, without requiring removal of that context from the provider message or a particular UI design.

Gigioxx:
Async question replies put the agent’s question and the user’s answer in one user bubble without attribution. Label them `Agent question:` and `User answer:` in the shared server formatter, preserving question context and attachments for the provider.

Closes #10786.

Reproduced with the existing engine test: both running and stopped sessions failed the new attribution assertion before the fix. All 29 engine tests now pass, including multiple questions, attachments, restarts, and duplicate replies. Server typecheck, targeted lint, and formatting pass.

Verified the real web transcript with isolated fixtures. Before uses the reproduced original message text; after is generated by the fixed engine. The shared payload serves web, desktop, and mobile; native clients were not separately exercised. Existing stored messages are unchanged.

Before:

Before: question and answer without attribution

After:

After: agent question and user answer labeled

Model: GPT-6. Harness: Codex.

## Summary by CodeRabbit

- **Improvements**
  - User-provided answers now clearly label each associated question and answer, including relevant attached-file details.
  - Empty response sections continue to be omitted for cleaner message formatting.

### DIFF
--- a/apps/server/src/orchestration/Layers/OrchestrationEngine.test.ts
+++ b/apps/server/src/orchestration/Layers/OrchestrationEngine.test.ts
@@ -269,7 +269,7 @@ describe("OrchestrationEngine", () => {
         expect(userMessages).toHaveLength(1);
         expect(userMessages?.[0]?.attachments).toEqual(response.attachmentsByQuestionId["1"]);
         expect(userMessages?.[0]?.text).toBe(
-          "Which package manager?\npnpm\n\nWhat should it be named?\nExample\nAttached file: spec.txt (thread-1-00000000-0000-4000-8000-0000000000aa-txt)",
+          "Agent question: Which package manager?\nUser answer: pnpm\n\nAgent question: What should it be named?\nUser answer: Example\nAttached file: spec.txt (thread-1-00000000-0000-4000-8000-0000000000aa-txt)",
         );
         expect(
           after.threads[0]?.activities.find((activity) => activity.kind === "user-input.resolved")
--- a/apps/server/src/orchestration/decider.ts
+++ b/apps/server/src/orchestration/decider.ts
@@ -1280,7 +1280,12 @@ export const decideOrchestrationCommand = Effect.fn("decideOrchestrationCommand"
             .map((attachment) => `Attached file: ${attachment.name} (${attachment.id})`)
             .join("\n");
           replies.push(
-            [`${question.question}\n${answer.trim()}`, attachmentLabels].filter(Boolean).join("\n"),
+            [
+              `Agent question: ${question.question}\nUser answer: ${answer.trim()}`,
+              attachmentLabels,
+            ]
+              .filter(Boolean)
+              .join("\n"),
           );
         }
         // Commit the answer and its message together. The normal turn path


juliusmarminge:

PR #10816, opened 2026-09-08
fix(review): use the selected thread's repository

tris203:
## What Changed

Review previews and expanded file contents use the selected thread's repository. Remove the launch-directory restriction and the UI fallback that retried against the server's directory.

## Why

When T3's dev server runs inside one worktree, reviewing another worktree was rejected as outside the configured workspace root. The fallback could then show changes from the worktree running T3 instead. The server's launch directory is a default workspace, not the boundary for other projects and worktrees.

## Validation

The regression test covers switching between two projects and an external worktree, reading the corresponding file contents, and rejecting file paths that escape the selected repository.

- Installed dependencies with `pnpm i`.
- `pnpm exec vp test run apps/server/src/review/ReviewService.test.ts`: passed.
- `pnpm --filter t3 --filter @t3tools/web run typecheck`: passed.
- Targeted lint and formatting checks on all three changed files: passed; lint reported two warnings in unchanged DiffPanel code (refs during render and an extra effect dependency).
- CI checks and all test shards passed on f4c9ebf27c.

Local checks ran on Node 24.3.0, which produces an engine warning against the repository's required ^24.13.1. No visual layout changes.

## Checklist

Prepared with GPT-6 through the Codex harness in T3 Code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Diff previews and file contents now use the currently selected repository or worktree.
  * Switching between projects and worktrees displays the corresponding changes and files.
  * Empty non-repository directories and invalid path traversal are handled with clear errors.
  * Removed fallback behavior that could display results from a different workspace.


tris203:
Checked the approvability concern against the RPC and filesystem code. Removing this particular launch-directory restriction is intentional: running the dev server in worktree A must not prevent reviewing a thread in worktree B or another project. The restriction covered `ServerConfig.cwd` and the default worktrees directory, not the set of projects owned by the environment.

Both review RPCs still require `review:write` (`apps/server/src/auth/RpcAuthorization.ts:122-123`). Existing VCS operations also accept the requested repository directory, and standard client permissions are environment-scoped (`packages/contracts/src/auth.ts`). Working-tree file expansion still checks both the resolved path and real path against the selected repository root, including symlink escapes (`apps/server/src/vcs/GitVcsDriverCore.ts:2409-2450`). The regression test also checks traversal rejection.

This does broaden repository selection for a token with `review:write`; it does not retain launch-directory confinement. That confinement is the bug addressed here, so I am retaining the implementation rather than reintroducing it. A project-scoped authorization model would be a separate change across repository operations.

CI checks and all test shards passed on f4c9ebf27ced9bddc86d7efb84dd82834c633888.

tris203:
Reviewed CodeRabbit's test-assertion nitpick. The existing `assert.strictEqual(escaped._tag, "GitCommandError")` already throws if the tag is wrong, so the following detail assertion cannot silently be skipped in a passing test. The guard provides TypeScript narrowing without the proposed structural cast. Keeping the existing assertions; the regression test passed locally and in CI.

The docstring-coverage warning is also not actionable for this change: the repository's documentation guidance prefers types, tests, and concise comments over restating implementation. ReviewService already includes the relevant explanation of why the server launch directory is not a filesystem boundary.

All checks have completed on f4c9ebf27c with no failures. CodeRabbit and Macroscope's correctness/convention checks passed; Macroscope's approvability check is neutral, with its repository-boundary concern addressed above.

Issue #5952, opened 2026-08-10
fix(opencode): show native Task subagents in the Agents panel


coygeek:
## Summary

OpenCode's native `Task` tool successfully creates and completes child-agent sessions inside a T3 Code thread, but the thread's Agents panel remains empty and says `No agents yet`.

This is specific to the OpenCode provider integration, not a refusal or failure by the selected model: the parent transcript contains multiple completed `Task` rows with distinct child session IDs and returned task results.

Area: `apps/server` (OpenCode provider adapter), observed in the `apps/web` Agents panel.

## Steps to reproduce

1. Start T3 Code Nightly `0.0.33-nightly.20260809.1047`.
2. Open a thread using the OpenCode provider with OpenCode `1.18.15`, the `MiniMax-M3` model, Build mode, and Full access.
3. Ask the primary agent to use OpenCode's native `Task` tool to launch at least two independent subagents. For example: `Use the Task tool to launch two independent Explore subagents, each inspecting a different part of this project.`
4. Wait until the parent transcript shows multiple completed `Task` tool calls with distinct `<task id="ses_..." state="completed">` results.
5. Open the thread's Agents panel.

## Expected behavior

Each OpenCode child session created by the native `Task` tool appears in the Agents panel with its task identity and lifecycle state. Completed tasks remain visible as settled agents, consistent with the panel's promise that subagents spawned by the thread appear there.

## Actual behavior

The native OpenCode subagents execute and return results to the parent transcript, but the Agents panel remains empty and displays:

> No agents yet
>
> When this thread spawns subagents or runs a workflow, they show up here with live status, activity, and token usage.

In the supplied reproduction, at least four distinct completed `Task` rows are visible while the panel still reports no agents.

## Affected area

- OpenCode provider sessions using the native `Task` tool.
- The server-side OpenCode event adapter and the persisted `task.*` activity stream consumed by the Agents panel.
- The web/desktop Agents panel roster for those threads.

## Runtime or environment

- T3 Code Nightly: `0.0.33-nightly.20260809.1047`
- T3 Code release commit: `062b4618c229f3e2f13e44efd8dab8c71ad33dae`
- OpenCode CLI: `1.18.15`
- Model: `MiniMax-M3`
- Agent/mode: Build
- Access mode: Full access
- Connection: Local checkout
- OS: macOS `26.5.2`, Apple silicon (`arm64`)

## Evidence

[...]

PR #10858, opened 2026-09-08
fix(server): show OpenCode subagents in the Agents panel

jmfrank63:
Refs #5952
## The problem

The Agents panel is built exclusively from `task.*` activities, but `OpenCodeAdapter`
never emitted any. It classified a subagent invocation as an ordinary tool call
(`collab_agent_tool_call`), so the roster stayed empty on every OpenCode thread even
while subagents were running. Codex, Claude, and Antigravity all translate their
native subagent signals into task lifecycle events; OpenCode was the only enabled
adapter that did not.

## The fix

Translate `collab_agent_tool_call` tool parts into `task.started` / `task.updated` /
`task.completed` with `timelineBypass`, instead of emitting a timeline tool row.

- **Lifecycle** — first non-terminal update becomes `task.started`, later updates
  become `task.updated`, and `completed`/`error` become `task.completed` with status
  `completed`/`failed`.
- **One row per subagent** — OpenCode repeats a part on every streamed update, which
  produced three `task.started` events for a single subagent and would have tripled
  the roster count. A per-session set of started call ids keeps it to one.
- **Identity** — `role` from `subagent_type`, `description` from `description`, title
  from the resolved `state.title`. No `agentId` is set, so `classifyTaskAgentKind`
  stamps these `agentKind: "agent"` and they land in the roster rather than the
  background bucket.
- **No lost output** — the completed row carries the subagent's result in `summary`,
  reusing the existing `detailFromToolPart`. That is the same text the tool row
  previously shipped as `detail`, so payload size is unchanged and moving the
  subagent off the timeline drops nothing.

`timelineBypass` keeps the parent timeline to a single spawn CTA rather than a raw
tool row, matching how Codex reports child agents.

No contract change: every field used already existed as optional task linkage. No
client change: web and mobile both consume the same `task.*` events.

## Before / after

| Before (`upstream/main`) | After (this PR) |
| --- | --- |
| before-agents-panel|after-agents-panel
 |
| Agents panel reads **"No agents yet"**, and the subagent shows inline in the timeline as a raw `task` tool row. | Agents panel lists the subagent under **DIRECT SPAWNS** with its role (`explore`), duration, and result. The timeline collapses to a single **"Ran 1 subagent"** CTA. |

Both captures use the same prompt, project, and model (OpenCode `GLM-5.3` via ClinePass) against
OpenCode 1.18.29; only the server code differs.

## Testing

Three tests in `OpenCodeAdapter.test.ts`, each watched failing before the fix:
roster routing with no timeline leakage, one `task.started` per subagent across
repeated updates, and result/failure text on the completed row.

```
vp test run apps/server/src/provider/Layers/OpenCodeAdapter.test.ts \
            apps/server/src/provider/Layers/OpenCodeProvider.test.ts \
            apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.test.ts \
            apps/web/src/session-logic.test.ts
# 273 passed

vp run --filter t3 typecheck   # clean
vp lint <changed files>        # clean
```

Also verified end to end in a local `vp run dev` instance against OpenCode 1.18.29:
the subagent now appears in the Agents panel.

---

Written by Claude Opus 5 (1M context) in Claude Code, driven through T3 Code.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of subagent tasks during OpenCode sessions.
  * Task progress now appears with clear started, updated, and completed statuses.
  * Completed and failed tasks display the subagent’s result or error details.

* **Bug Fixes**
  * Prevented duplicate task-start notifications when repeated progress updates are received.


j-gaertig:
Would love to see this merged! 🙏

OpenCode is my primary harness, which I use alongside Codex through T3 Code, so being able to watch OpenCode subagents at work would be really useful. Thanks for working on this!

juliusmarminge:

PR #10865, opened 2026-09-09
test(server): match V2 tool output redaction in retention coverage

saphid:
The V2 tool-output redaction change leaves the foundation retention test expecting the old `{ truncated: true }` output preview. Test Server 1 fails on current V2 integrations for #10024 and #10025 because replay and live payloads now intentionally omit that output.

Update the assertion to require omission. The test still checks the transport size limit and verifies that both the stored event and full detail projection retain the original 9 MB output. Production behavior is unchanged.

Validation on V2 base `7dd9873ed`: 56 focused foundation/wire-projection tests pass, server typecheck passes, and targeted lint/format/diff checks pass. The stale assertion is reproduced in both linked PRs' Test Server 1 failures. Independent Claude Opus 5 high review was unavailable during this audit: the direct launcher exited 1 with expired OAuth before inference.

Prepared with GPT-6 in the Codex/T3 harness.


juliusmarminge:
Closing as already covered on v2 by commit 8207e556e271c4f0434a94a95f339b6a5681539e. FoundationPersistence.test.ts now asserts that the transported dynamic-tool payload omits output, while retaining the full persisted-output checks. The exact one-line patch reverse-applies to the current branch, and the replacement commit is already on the remote v2 branch.

Issue #10798, opened 2026-09-08
Codex adapter keeps dead sessions after the provider process exits — thread shows "Working" forever


stephenjason89:
**Version**: `0.0.34-nightly` (reproduced against `upstream/main` @ `061543e9e`) · macOS · provider `codex`

When a Codex App Server process dies mid-turn, the adapter forwards the `session.exited` event but keeps the dead runtime in its live-session map. Every liveness check downstream still reports the thread as running, so the UI shows "Working" indefinitely and no recovery path fires.

I hit this on a Gauntlet Loop that sat at "Working" for over 7 hours with no process behind it:

```
specialist_reports_sol_xhigh     8h 40m   assistant message                     1.5M tok
specialist_pipeline_sol_xhigh    8h 40m   /bin/zsh -lc 'vendor/bin/pint ...'    1.7M tok
critic_selected_fixes_sol_xhigh  8h 25m   reasoning                             83.0k tok

3 working  22 idle  1 settled                                     Σ 118.6M tok
```

The last provider event was ~7 hours before I looked. No `codex` process matching the thread remained. The worktree changes and test results were intact — only the session state was wrong.

### Mechanism

`CodexSessionRuntime` watches the child's exit code and emits `session/exited` when the process dies without `close()` being called (`CodexSessionRuntime.ts:2231`). The adapter's event consumer maps that to a canonical `session.exited` and offers it to the runtime event queue — and then does nothing else (`CodexAdapter.ts:2437`).

Nothing removes the entry from the adapter's `sessions` map. Only `stopSessionInternal` deletes, and it runs solely from an explicit `stopSession`/`stopAll`. So after the process dies:

- `hasSession(threadId)` stays `true` — the entry exists and `stopped` is still `false` (`CodexAdapter.ts:2692`)
- `listSessions()` still returns it, because it filters only on `stopped` (`CodexAdapter.ts:2685`)
- the session scope is never closed, so the runtime's client, queues, and stderr fibers leak

Startup reconciliation trusts `listSessions()` to decide which persisted bindings are still live, and `ProviderSessionReaper` permanently skips any session holding an `activeTurnId`. A dead runtime retained in the map therefore evades both, and the thread stays `running` with its `activeTurnId` set forever.

[...]

PR #10874, opened 2026-09-09
fix(server): recover Codex sessions after signal exits

### LINKED ISSUE #12904: [Bug]: Codex turn retries stay closed after App Server SIGKILL
## Bug

After the Codex App Server is terminated by `SIGKILL`, sending another message to the same T3 thread repeatedly fails with `ProviderAdapterSessionClosedError: codex adapter thread is closed`. Retrying the message does not recreate the Codex process. The T3 server itself remains healthy.

Observed on Linux with T3 Code v0.1.69. In the incident, the process exited hours before the next user message; both the first message and its retry failed against the same closed adapter thread.

## Reproduction

1. Start a Codex-backed T3 thread and let its App Server session become active.
2. Terminate that App Server process with `SIGKILL` (for example, through a memory pressure manager).
3. Send a new message to the same thread, then retry it.

**Expected:** T3 marks the old session as exited and resumes the native Codex thread in a fresh App Server process before sending the new turn. The interrupted turn itself need not be replayed.

**Actual:** The adapter keeps routing turns to the closed process; each send fails with `ProviderAdapterSessionClosedError`.

## Diagnosis and related work

Two lifecycle gaps combine in this case:

- Signal termination makes `child.exitCode` fail, while the runtime watcher only handles a successful numeric exit status. Consequently it can miss `session/exited` for `SIGKILL`.
- Even when an exit event is emitted, the Codex adapter can retain the dead runtime in its session map, so `hasSession` remains true and recovery is skipped.

Related: #10798 describes the retained session and indefinitely running UI state. #10799 addresses adapter cleanup; #10874 addresses signal exits. This issue tracks the user-visible repeated-turn failure when both gaps occur together and the need for an integration regression covering the full path.

khaneliman:
Fixes #12904

## What Changed

Handle failed Codex App Server exit-status effects in the runtime watcher. Mark the session as errored, clear its active turn, and emit the existing `session/exited` event with a diagnostic. Numeric exit statuses and intentional shutdown keep their existing behavior.

## Why

Effect Node reports signal termination, including SIGKILL, through the failure channel of `child.exitCode`. The watcher only handled successful numeric results, so a killed process could leave its thread apparently active. The next turn then failed with `ProviderAdapterSessionClosedError`.

Emitting the missing event lets the existing stopped-session and native-thread resume path run on the next turn. This does not retry or replay the interrupted request.

Related: #10798 and #10799. The latter cleans up adapter sessions after receiving an exit event; this change ensures signal termination produces that event.

### Testing

- Real-process SIGKILL regression fails on the original watcher and passes with this fix, including resuming the same native thread and sending a new turn.
- Numeric exit codes 0 and 7 and intentional shutdown are covered.
- 105 focused tests pass: runtime integration (13), runtime unit (40), and adapter (52).
- Server typecheck, targeted lint, formatting, and diff checks pass.
- Tested on Linux. The new process-fixture tests skip Windows.

## Checklist

- [ ] I included before/after screenshots for any UI changes (not applicable)

Implemented with GPT-5.6 Luna; reviewed with GPT-6 Astra, using the Codex harness in T3 Code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved session handling when the Codex process exits unexpectedly.
  - Sessions now report accurate closed or error states, preserve exit details, clear active turns, and emit the appropriate exit event.
  - Fixed cases where exit-processing failures left sessions without an updated status or exit notification.
  - Browser tool availability now accurately reflects configuration and runtime settings.

- **Tests**
  - Added recovery coverage for interrupted sessions, including resume behavior, exit statuses, cleanup, intentional closes, unexpected exits, and terminal events.

### DIFF
--- a/apps/server/src/provider/Layers/CodexCollabRuntime.integration.test.ts
+++ b/apps/server/src/provider/Layers/CodexCollabRuntime.integration.test.ts
@@ -71,7 +71,10 @@ rl.on("line", (line) => {
     write({ id, result: fixture.turnStart });
     write({ jsonrpc: "2.0", method: "turn/started", params: { threadId: fixture.threadStart.thread.id, turn: fixture.turnStart.turn } });
     if (process.env.T3_CODEX_RECOVERY_COMPLETE === "1") write({ jsonrpc: "2.0", method: "turn/completed", params: { threadId: fixture.threadStart.thread.id, turn: { ...fixture.turnStart.turn, status: "completed" } } });
-    if (process.env.T3_CODEX_RECOVERY_EXIT_CODE !== undefined) process.exit(Number(process.env.T3_CODEX_RECOVERY_EXIT_CODE));
+    if (process.env.T3_CODEX_RECOVERY_EXIT_CODE !== undefined) {
+      const code = Number(process.env.T3_CODEX_RECOVERY_EXIT_CODE);
+      process.stdout.end(() => process.exit(code));
+    }
     return;
   }
   if (id !== undefined) write({ id, result: {} });
--- a/apps/server/src/provider/Layers/CodexCollabRuntime.integration.test.ts
+++ b/apps/server/src/provider/Layers/CodexCollabRuntime.integration.test.ts
@@ -17,13 +17,16 @@ import { it } from "@effect/vitest";
 import { type ProviderApprovalDecision, type ProviderEvent, ThreadId } from "@t3tools/contracts";
 import * as Deferred from "effect/Deferred";
 import * as Effect from "effect/Effect";
+import * as Exit from "effect/Exit";
 import * as Fiber from "effect/Fiber";
+import * as Ref from "effect/Ref";
 import * as Schema from "effect/Schema";
+import * as Scope from "effect/Scope";
 import * as Stream from "effect/Stream";
 import { assert, describe } from "vite-plus/test";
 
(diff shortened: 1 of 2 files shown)


khaneliman:
@coderabbitai I checked the browser-availability finding in [the latest review](https://github.com/pingdotgg/t3code/pull/10874#pullrequestreview-5162561018) against head `b986d3752`.

The runtime does not add a mandatory MCP server: [`codexSessionAppServerArgs`](https://github.com/pingdotgg/t3code/blob/b986d3752e15dbf915a885b4e9b31aa220165b5a/apps/server/src/provider/Layers/codexLaunchArgs.ts#L41-L46) only assembles the supplied arguments. The [adapter](https://github.com/pingdotgg/t3code/blob/b986d3752e15dbf915a885b4e9b31aa220165b5a/apps/server/src/provider/Layers/CodexAdapter.ts#L2273-L2288) supplies the MCP configuration and `browserToolsAvailable: mcpSession.preview` together. The normal application path therefore does not produce the proposed combination of browser capability enabled with no MCP arguments.

The flagged condition and capability option are also unchanged from the rebased upstream base `0f602b337`; they are not additions from this PR. I’m leaving that behavior unchanged to keep this PR scoped to signal-exit recovery. Please reassess the finding and the walkthrough’s attribution of browser-tool handling to this PR.

PR #10920, opened 2026-09-09
fix(web): persist usage page environment selection

AmoonPod:
## What Changed

The Usage page already remembered the last metric and time window between visits. This PR extends the same `t3code:usage-page-preferences:v1` localStorage record with the environment selection, so a custom set of selected environments is restored on the next visit.

- `usagePagePreferences.ts`: the preferences schema gains an optional `selectedEnvironmentIds` field. `null` (the default) means "all environments", including ones connected later; an array is an explicit subset, an empty array means nothing is selected. Two small helpers convert between the stored shape and the `ReadonlySet` the page uses.
- `UsagePage.tsx`: the initial selection is restored from preferences, and every selection change (including switching back to all environments) is saved. Changing metric or window preserves the selection instead of dropping it.

## Why

The environment picker was the only Usage page state not persisted. Users with the same providers connected on multiple machines see duplicated provider rows across environments, and deselecting the duplicates on every visit was tedious. This follows the existing pattern the page uses for metric/window persistence, with no new storage keys and graceful decoding of preferences saved before this field existed.

## UI Changes

None. No visible element changed; only which environments are pre-selected when the page opens.

## Checklist

- [x] I included before/after screenshots for any UI changes (not applicable - no visual change)

---

Model: openrouter/z-ai/glm-5.3-flash. Harness: OpenCode (T3 Code).

## Summary by CodeRabbit

* **New Features**
  * Usage page environment filter selections are now saved and restored across visits.
  * Explicit environment selections, deselections, and the all-environments option are remembered.

* **Bug Fixes**
  * Improved handling of saved usage preferences, including older, invalid, or incomplete preference data.
  * Environment selections are cleaned up to prevent duplicate or blank entries.

### DIFF
--- a/apps/web/src/components/usage/UsagePage.tsx
+++ b/apps/web/src/components/usage/UsagePage.tsx
@@ -63,8 +63,10 @@ import { UsagePriceOverrides } from "./UsagePriceOverrides";
 import { UsageProviderChart, type UsageChartMetric } from "./UsageProviderChart";
 import { PROVIDER_ORDER, PROVIDER_PRESENTATION, providersWithUsage } from "./usageProviders";
 import {
+  fromSelectedEnvironmentIds,
   readUsagePagePreferences,
   saveUsagePagePreferences,
+  toSelectedEnvironmentIds,
   type UsagePagePreferences,
 } from "./usagePagePreferences";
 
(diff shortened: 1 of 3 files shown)


kridaydave:
Good fix, and the bot review plus CI are green. It needs a rebase: main changed the default metric to `limits` and added `setLimitsNow` in the same spots you touch. In each conflicted hunk (`UsagePage.tsx`, `usagePagePreferences.ts`, and the test), keep both sides: their metric/limits lines plus your env persistence lines. Push after the rebase and it should be ready for a maintainer to pick up.

PR #10939, opened 2026-09-09
fix(server): preserve reasoning effort after restart

### LINKED ISSUE #10928: [Bug]: Continue threads after restarts does not preserve Codex reasoning effort
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Enable "Continue threads after restarts".
2. Open a Codex thread and explicitly select a non-medium reasoning effort, such as High.
3. Start a task and restart T3 Code while the turn is still running, without manually stopping the thread first.
4. Let the thread continue automatically. Do not change the model or reasoning settings, and do not send a manual follow-up.
5. Compare the reasoning effort of the automatic continuation with the setting used before the restart.

### Expected behavior

The automatically continued turn should preserve the reasoning effort used before the restart. Restarting T3 Code should not change this setting.

### Actual behavior

The thread continues automatically, but the reasoning effort differs from the setting used before the restart, without any manual change.

Source inspection of v0.0.40 suggests a possible cause:
- In apps/server/src/serverRuntimeStartup.ts, restart reconciliation calls providerService.sendTurn without modelSelection.
- CodexAdapter reads reasoningEffort from the current turn's modelSelection.
- In apps/server/src/provider/Layers/CodexSessionRuntime.ts, buildCodexCollaborationMode falls back to medium when effort is missing.

### Impact

### Version or commit

v0.0.40

### Environment

_No response_

### Logs or stack traces

```shell

```

_No response_

### Workaround

_No response_

IgniparousTempest:
Restart reconciliation resumed provider turns without the thread model selection. Codex therefore rebuilt collaboration settings with its medium-effort fallback, even when the interrupted turn used high or xhigh reasoning.

Forward the persisted thread model selection on automatic continuation. Extend the reconciliation and SQLite-backed startup tests to prove a high-effort selection survives restart, and cover the resulting Codex turn parameters and developer instructions.

Closes #10928.

Verification:
- `vp test run apps/server/src/serverRuntimeStartup.reconcile.test.ts apps/server/src/provider/Layers/CodexSessionRuntime.test.ts`
- `vp test run apps/server/integration/orphanedProviderSessionStartup.integration.test.ts`
- `vp test run apps/server/src/provider/Layers/CodexAdapter.test.ts`
- targeted lint and formatting checks

Model: GPT-5.6 via Codex.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved the interrupted turn’s provider, model, and reasoning settings when continuing sessions after a server restart.
  - Ensured recovered and promptless continuations use the model selection associated with the interrupted turn, even if thread settings later change.
  - Retained reasoning-effort settings in collaboration behavior and provider instructions.
  - Safely ignored invalid or unavailable persisted model selections during recovery.

- **Tests**
  - Added coverage for persisted model selection during session recovery and continuation.
  - Added validation for non-default reasoning effort propagation.


apps/server/src/serverRuntimeStartup.ts
```diff
@@ -694,6 +694,7 @@ export const reconcileProviderSessions = Effect.gen(function* () {
               ...(capabilities.promptlessTurnContinuation === true
                 ? { continuation: true }
                 : { input: SERVER_UPDATE_CONTINUATION_PROMPT }),
+              modelSelection: thread.modelSelection,
```
derektrimm: Forwarding the composer's current selection here changes the outcome for non-Codex threads in one case. The restart path routes by the bound session (`resolveRoutableSession` by thread id in `ProviderService.sendTurn`), while the normal send path in `ProviderCommandReactor.ts:612-620` compares the selection's instance to the active session and switches providers on purpose. `thread.meta.update` accepts a `modelSelection` while a turn is running (`decider.ts:873-892`), so if the composer was moved to another instance before the restart, OpenCode (`OpenCodeAdapter.ts:3076`) and Antigravity (`AntigravityAdapter.ts:973`) reject the mismatched `instanceId` and the thread settles with "Could not continue this thread after the server restart" a few lines below. Today those threads continue on the session's own model.

Two ways to keep the Codex fix without that edge: use the selection the interrupted turn actually ran with, which is already persisted as `runtimePayload.modelSelection` on the binding (`ProviderService.ts:403` reads it and recovery passes it to `startSession` at `:1192`), or forward `thread.modelSelection` only when `instanceId === binding.value.providerInstanceId`. A reconcile case with a mismatched instance asserting the selection is omitted would pin either choice.

IgniparousTempest: Addressed in bc30bf210. Restart reconciliation now schema-validates and forwards the interrupted turn’s persisted `runtimePayload.modelSelection` from the provider binding instead of the mutable `thread.modelSelection`; missing or invalid persisted selections are omitted.

I added a reconciliation regression where the thread metadata has moved to a different provider instance and asserted that continuation still receives the original persisted selection. The SQLite-backed startup fixture now also mirrors the model selection persisted by `ProviderService.sendTurn`.

Verified with:
- `vp test run apps/server/src/serverRuntimeStartup.reconcile.test.ts apps/server/integration/orphanedProviderSessionStartup.integration.test.ts` (30 passed)
- targeted lint and formatting checks

apps/server/src/serverRuntimeStartup.ts
```diff
@@ -381,6 +381,13 @@ function readRuntimePayload(runtimePayload: unknown): Record<string, unknown> {
     : {};
 }
 
+const isModelSelection = Schema.is(ModelSelection);
+
+function readRuntimeModelSelection(runtimePayload: unknown): ModelSelection | undefined {
+  const value = readRuntimePayload(runtimePayload).modelSelection;
+  return isModelSelection(value) ? value : undefined;
```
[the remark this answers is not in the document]
IgniparousTempest: Addressed in f6f6b9cba. Both persisted-selection readers now use synchronous `Schema.decodeUnknownOption(ModelSelection)` and convert `None` to `undefined`, so legacy `{ provider, model, options }` payloads are transformed to canonical `{ instanceId, model, options }` values before session recovery or restart continuation.

Coverage now includes:
- restart reconciliation decoding a legacy Codex selection with `reasoningEffort` after thread metadata moved to another instance;
- `ProviderService` recovering a stale session from a legacy persisted selection;
- internally consistent cross-provider restart fixtures.

The branch was also rebased onto current `main` while preserving project-scoped restart settings.

Verified with 217 focused tests, `vp run --filter t3 typecheck`, targeted lint, formatting, and a frozen-lockfile install.

Issue #10905, opened 2026-09-09
[Bug]: Checkpoint refs are written non-atomically — an unclean restart leaves a 0-byte ref that breaks all git fetch/push in the repo


phillipphoenix:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

### Steps to reproduce

T3 writes checkpoint refs to `refs/t3/checkpoints/<base64-thread-id>/turn/<n>` in the project repo. These appear to be written **in place, non-atomically**. If the host dies mid-write, the ref file survives the crash with its size committed but its contents never flushed — i.e. as 0 bytes. Git reads a zero-length loose ref as the all-zeros SHA, and from that moment **every git command that enumerates local refs fails**, which includes fetch, push, and gc.

What happened here:

1. A thread was running and T3 wrote a checkpoint ref for the turn, plus a loose object.
2. ~2 minutes later the host (a Hetzner box running the T3 server) restarted uncleanly.
3. On remount, both files were 0 bytes — the ext4 delayed-allocation crash signature. Both shared the pre-crash mtime and a size of 0:

   ```
   0 bytes  .git/refs/t3/checkpoints/<base64-thread-id>/turn/1
   0 bytes  .git/objects/<xx>/<rest-of-sha>
   ```

4. Every subsequent worktree creation failed, and the background status poller failed on repeat every ~13s.

To reproduce deliberately, truncate any checkpoint ref in a project repo and then create a worktree:

```shell
: > .git/refs/t3/checkpoints/<any-existing-thread>/turn/0
git fetch origin        # fatal, before any network access
```

### Expected behavior

A crash mid-checkpoint should not be able to leave the repository in a state where all git remote operations fail. Checkpoint refs should be written atomically (temp file → `fsync` → `rename`), which is what git itself does for exactly this reason. At minimum, a zero-length or non-SHA file under `refs/t3/**` is never legitimate and could be dropped on startup.

### Actual behavior

All fetch/push/gc in the affected repo fail until the file is deleted by hand. Worktree creation surfaces:

```
GitCommandError: Git command failed in GitVcsDriver.fetchRemote (<repo>): git fetch origin failed
```

This reads as a credentials or connectivity problem and sends you down the wrong path. It is neither — `git fetch` dies while building its "have" list during negotiation, before opening a socket. The discriminator is `git ls-remote`, which never walks local refs:

| Check | Result |
|---|---|
[...]

PR #10944, opened 2026-09-09
fix(server): flush checkpoint objects and refs before publishing them

Mnigos:
Checkpoint capture writes loose objects with `git add` / `write-tree` / `commit-tree` and publishes the turn ref with `git update-ref`. Git renames those files into place without fsync by default, so an unclean host restart can leave a 0-byte file under `refs/t3/checkpoints/**` (and a 0-byte loose object). Git reads the empty ref as the all-zeros SHA and every later `fetch`, `push`, `gc` and `worktree add` in that repository fails until the file is deleted by hand.

The four object- and ref-writing commands of a capture now run with `-c core.fsync=objects,reference -c core.fsyncMethod=fsync`, so the data reaches disk before the rename publishes it (macOS defaults to `writeout-only`, which only queues the writeback). This is the "prevent" step from the triage on #10905; a startup scrub of zero-length refs and clearer git error diagnostics are separate follow-ups. Older gits without `core.fsync` ignore the unknown key.

## Verification

- New test in `GitVcsDriver.test.ts` runs a capture against the mocked VCS process and asserts that `add`, `write-tree`, `commit-tree` and `update-ref` carry the fsync option ahead of the subcommand, and that nothing else changed in the published `update-ref` call.
- `vp test run apps/server/src/vcs/GitVcsDriver.test.ts` and `apps/server/src/checkpointing/CheckpointStore.test.ts` (real captures with Apple Git 2.39) pass. Server typecheck and lint on the touched files are clean.

Related to #10905. Implemented with Claude Code (Claude Fable 5.1).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved checkpoint reliability by ensuring checkpoint data and references are fully written to disk before publication.
  * Reduced the risk of incomplete checkpoint files after an unexpected restart.


ahalekelly:
🤖 Generated with GPT-6 Astra
Confirmed this failure on Ubuntu: a zero-byte loose ref under `refs/t3/checkpoints/<thread>/turn/87` caused `git fetch origin main` to fail with `fatal: bad object refs/t3/checkpoints/.../turn/87` and the misleading follow-up that the remote did not send all necessary objects. The other 256 checkpoint refs resolved to valid objects. Moving only the empty ref to Trash restored fetch immediately. The empty file dated from August 28; I have not established what interrupted its write. This supports the startup scrub follow-up as well as the durability fix here.

PR #10981, opened 2026-09-09
fix(web): allow expanding duplicate tool call commands

Yash-Singh1:
## What Changed

- Allow tool-call rows to expand when their command matches the visible preview text.
- Simplify mobile glass surfaces by removing blur-target plumbing and using explicit fallback styles.
- Remove the provider model bulk enable/disable control and its associated helper tests.

## Why

Duplicate tool call commands could appear non-expandable when the visible label matched the command text. Expansion now depends on whether additional content is available, so duplicate commands remain inspectable.

The mobile glass cleanup reduces unnecessary blur-target complexity, while provider model settings retain their existing per-model controls.

Implemented with Codex (GPT-5).

## UI Changes

Before/after screenshots are not included. This change affects web tool-call expansion and mobile/settings UI behavior.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved work-entry expansion in the chat timeline. Entries can now expand consistently when they include commands, details, changed files, or viewed images—even when the displayed label matches the command.
  * Updated label wrapping behavior to respond consistently to the entry’s expanded or collapsed state.

### DIFF
--- a/apps/web/src/components/chat/MessagesTimeline.tsx
+++ b/apps/web/src/components/chat/MessagesTimeline.tsx
@@ -3309,13 +3309,12 @@ const PlainWorkEntryRow = memo(function PlainWorkEntryRow(props: {
           workspaceRoot,
         })
       : null;
-  const commandMatchesVisibleLabel = workEntry.command?.trim() === previewText.trim();
   const canExpand =
     (showFailedIndicator && previewText.trim().length > 0) ||
     (workEntry.itemType === "mcp_tool_call" && workEntry.toolData !== undefined) ||
     Boolean(
-      (!commandMatchesVisibleLabel &&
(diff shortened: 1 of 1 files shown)


Yash-Singh1:
> Generated by GPT-6-Astra

Before and after in the web client, replaying the same 75-line `cat` command with no output in an isolated environment. Same viewport and fixture; before uses the parent commit, after uses `ed019e2`.

**Before:** the entire command wraps into the timeline by default, with no expand/collapse control.

Before: command fills the timeline

**After:** the command stays on one truncated line with an expand control. Verified that clicking reveals the full script and collapsing restores the single-line row.

After: command stays collapsed and can be expanded

Issue #10980, opened 2026-09-09
[Bug]: preview_click leaves keyboard focus inside the preview tab, so the user's typing goes to the page


BenjaminPeters-code:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Summary

Agent automation in the preview panel takes the application's keyboard focus and never gives it
back. After a single `preview_click` from an agent, keystrokes from the user's physical keyboard are
delivered to the previewed page instead of to the T3 UI, until the user notices and clicks somewhere
by hand.

This also happens when the preview tab is **invisible** (opened with `open: false`), so there is no
visual cue at all. From the user's side it looks like the app randomly stops accepting typing.

A second, related problem in the same code path: `preview_press` does restore focus, but so quickly
that the key is never delivered to the page, while the tool still reports success.

### Steps to reproduce

1. Serve a page that logs focus and input events with timestamps. The probe used here is at the
   bottom of this issue.
2. From an agent, call `preview_open` with `open: false`, so the tab exists but is never shown.
3. Call `preview_click` once on any element in that hidden tab.
4. Touch nothing else, wait, then type on the physical keyboard.

### Expected behavior

After the automated click completes, keyboard focus returns to whatever had it before the click,
the same way `performAutomationPress` already restores it. A hidden preview tab in particular should
never end up holding the application's input focus.

### Actual behavior

Focus stays in the previewed page. `document.hasFocus()` is `true` there, and subsequent keydown
events arrive with `isTrusted: true` while no agent action is running, i.e. they are the user's own
keystrokes. Focus only comes back when the user clicks elsewhere by hand.

Timeline from the probe page, measured 2026-09-09 (UTC):

| Time | Event | `document.hasFocus()` | Source |
|---|---|---|---|
| 17:16:29.576 | page loaded | false | baseline |
| 17:16:40.801 | window `focus` (twice, same ms) | true | agent: one `preview_click`, panel hidden |
| 17:16:40.803 | `pointerdown` (68, 291), `isTrusted: true` | true | the synthetic click itself |
| 17:17:10.862 | `keydown` Alt / AltRight, `isTrusted: true` | true | **physical keyboard**, 30 s later, no agent action running |
| 17:18:16.390 | window `focus` | true | agent: `preview_press` Tab |
| 17:18:16.400 | window `blur` | false | focus restored after 10 ms, **Tab never arrived as a keydown** |
[...]

PR #10982, opened 2026-09-09
fix(desktop): hand keyboard focus back after an automation click

Mnigos:
An agent `preview_click` dispatches `Input.dispatchMouseEvent` into the guest WebContents, and that press moves the application's keyboard focus into the previewed page. Nothing hands it back, so every keystroke the user types afterwards lands in the page until they click elsewhere by hand. With a tab opened through `preview_open { open: false }` the page is not even visible, so pastes silently go to a hidden document.

`performAutomationPress` already captures `webContents.getFocusedWebContents()` before dispatch and restores it in an `Effect.ensuring` block. The click path now does the same through a shared `restoreFocusedWebContents` helper, on success and on failure. The press path keeps its behavior and uses the helper. The restore yields in two cases: focus moved to a third renderer while the action ran (the user's newer selection wins), or T3 has neither a focused renderer nor a focused window because the user switched to another app, so nothing pulls them back. That second guard matches the one #9848 adds on the press path.

Out of scope, per the triage list on #10980: the race where `performAutomationPress` restores focus before the renderer processes the key, and whether hidden tabs should ever hold focus.

## Verification

- New test in `Manager.test.ts`: a click restores the previously focused renderer exactly once, after the last CDP command; a click that fails the viewport check still restores it; focus that moved to a third renderer is left alone; and no restore happens when T3 lost app focus (`getFocusedWebContents()` and `BrowserWindow.getFocusedWindow()` both null). It fails without the restore and passes here.
- `vp test run apps/desktop/src/preview/Manager.test.ts`: 90 tests pass. Desktop typecheck and lint on the touched files are clean.

Related to #10980. Implemented with Claude Code (Claude Fable 5.1).

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored keyboard focus to previously focused content after automated clicks.
  * Preserved focus restoration when offscreen click actions fail.
  * Avoided overriding a newer focus selection made while automation was in progress.
  * Safely skipped restoration when the user switched apps, the original content was unavailable, or the click target was already focused.


shivamhwp:

`restoreFocusedWebContents` still calls the previous renderer's `focus()` when both `getFocusedWebContents()` and `BrowserWindow.getFocusedWindow()` are null, as happens after switching to another app during the click. Skip restoration when T3 no longer has a focused window, and cover that case alongside the third-renderer case. The related focus guard in #9848 already handles this condition; preserve it when combining the overlapping click and keypress changes.

Mnigos:
@shivamhwp added in 2970e5b: no restore when T3 has neither a focused renderer nor a focused window, covered next to the third-renderer case; same guard shape as #9848.

PR #10995, opened 2026-09-09
fix(server): sign Azure DevOps PR thread reads

jvzijp:
## What Changed

The Azure DevOps pull request provider reads a conversation through `az rest`, since `az repos pr` has no threads command. That call now passes `--resource` with the Azure DevOps Entra application id so the Azure CLI attaches a token. The existing `listThreads` test asserts the flag is present.

## Why

The Azure CLI only infers the token audience for Azure's own endpoints, not for `dev.azure.com`. Without `--resource` the threads request went out unauthenticated. Azure answered with its HTML sign-in page, the JSON decoder rejected it, and the provider degraded the failure to an empty, "truncated" conversation. Every Azure DevOps pull request therefore showed 0 comments alongside the notice that the conversation was longer than the page could read.

Reproduced against a real Azure DevOps repository: the bare call returns the sign-in page, the same call with the resource id returns the thread list. The id is the well-known Azure DevOps application id, not anything tenant-specific.

## Checklist

- [x] I included before/after screenshots for any UI changes (no UI change)

Model: Claude Fable 5.1. Harness: Claude Code, driven through T3 Code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Azure DevOps pull request thread retrieval by requesting the correct authentication audience.
  - Thread requests now return API data instead of an Azure sign-in page when using authenticated access.

- **Tests**
  - Updated coverage to verify that Azure DevOps requests use the required resource identifier.

---


kridaydave:
Thanks for diagnosing this, you were right about the cause. Small update: main has since replaced this `az rest` call with `az devops invoke`, which authenticates the way the extension does, so this exact code path is gone. Could you pull latest main and check whether the threads read works for you now? If it does, feel free to close this, or a maintainer can tidy it up. Thanks again for pinpointing it.

PR #10999, opened 2026-09-09
fix(desktop): keep annotation comment direction independent of host page

### LINKED ISSUE #4644: [Bug]: Preview annotation comments do not support RTL text direction
### Area

apps/desktop

### Summary

Arabic and other RTL comments in the browser annotation editor use the inspected page's text direction instead of the comment's own content. On a normal LTR page, the RTL comment is aligned incorrectly; on an RTL page, the entire annotation toolbar is mirrored.

### Steps to reproduce

1. Open any left-to-right page in the T3 Code desktop browser panel, for example `https://example.com`.
2. Click **Annotate preview**.
3. Select any element so the annotation comment editor appears.
4. Type an Arabic comment, for example `كبّر هذا العنوان`.

### Expected behavior

The comment field should infer direction from the entered text. Arabic, Hebrew, Persian, and other RTL comments should start from the right and preserve readable bidirectional ordering.

The annotation toolbar and editor controls should keep their own stable layout regardless of the inspected page's direction.

### Actual behavior

The annotation comment field inherits the inspected page's direction:

- On an LTR page, Arabic annotation text remains in an LTR field and is aligned/ordered incorrectly.
- On an RTL page, the comment flows RTL, but the annotation chrome is also mirrored, including the tool order and **Attach** button placement.

This couples the annotation UI to the host document instead of handling the comment direction independently.

### Impact

Minor bug or occasional failure. RTL users can enter a comment, but the field is awkward to edit and mixed RTL/LTR content such as URLs, selectors, and code tokens can be displayed in the wrong visual order.

### Version or commit

Reproduced on `0.0.29-nightly.20260724.893` (`ece05087a70e`).

### Environment

T3 Code desktop nightly, Linux/X11, Electron.

**Actual behavior on an LTR page:** the short Arabic comment is anchored to the left side of the field.

Arabic annotation comment remains LTR on an LTR page

**Diagnostic control with the host page set to RTL:** the same comment moves to the right, but the toolbar order and **Attach** button also reverse. This shows that the annotation overlay inherits direction from the inspected page.

Forcing the host page to RTL also mirrors annotation controls

Related but not duplicate: #1771 is a broad proposal for bidi-ready renderer styling and message bodies. This report is narrowly about the desktop preview annotation overlay and its comment textarea.

### Technical note

[Current `main` at `5719e8ac4020`](https://github.com/pingdotgg/t3code/blob/5719e8ac4020dda0e375ef61d044b61f55a0df8a/apps/desktop/src/preview/PickPreload.ts#L414-L419) creates the annotation `<textarea>` without a `dir` attribute. Giving the comment field content-based direction such as `dir="auto"`, while explicitly isolating the annotation chrome from the host document direction, appears to be the smallest relevant scope.

### Workaround

No practical per-comment workaround. Temporarily changing the inspected document's root direction to RTL also changes the page and annotation controls.

abdelrhmanehab10:
## Summary
- Preview annotation comments inherited the inspected page's text `dir`, so RTL comments (Arabic, Hebrew, Persian) were misaligned on LTR pages, and forcing the host page to RTL mirrored the whole annotation toolbar instead of just the comment text.
- Pin the overlay `host`/`root` elements to `dir="ltr"` so the annotation chrome (toolbar, editor, Attach button order) has a stable layout regardless of the inspected page's direction.
- Set `dir="auto"` on the comment `<textarea>` so its direction is inferred from the actual typed content instead of inherited from the page.

Fixes #4644

## Test plan
- [x] `pnpm typecheck` in `apps/desktop` — no new errors introduced (verified against a clean checkout, which has the same pre-existing unrelated errors in `scripts/lib/*`)
- [ ] Manual repro per the issue: open an LTR page, start annotation, type an Arabic comment, confirm it now renders RTL while toolbar/Attach button stay in place
- [ ] Manual repro on an RTL host page: confirm toolbar/editor chrome no longer mirrors

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text direction handling in the annotation overlay and comment input for better support of mixed-language content.

### DIFF
--- a/apps/desktop/src/preview/PickPreload.ts
+++ b/apps/desktop/src/preview/PickPreload.ts
@@ -436,6 +436,7 @@ function startAnnotation(): void {
   let finished = false;
   const host = document.createElement("div");
   host.setAttribute(OVERLAY_ATTRIBUTE, "");
+  host.setAttribute("dir", "ltr");
   host.style.cssText = `position:fixed;inset:0;z-index:${Z_INDEX_OVERLAY};pointer-events:none`;
   applyAnnotationTheme(host, annotationTheme);
   const shadowRoot = host.attachShadow({ mode: "closed" });
@@ -445,6 +446,7 @@ function startAnnotation(): void {
 
   const root = document.createElement("div");
   root.setAttribute(OVERLAY_ATTRIBUTE, "");
+  root.setAttribute("dir", "ltr");
   root.className = "fixed inset-0 font-sans text-foreground";
   root.style.cssText = "pointer-events:none";
   const cursorStyle = document.createElement("style");
@@ -494,6 +496,7 @@ function startAnnotation(): void {
 
   const comment = document.createElement("textarea");
   comment.placeholder = "Describe the change…";
+  comment.setAttribute("dir", "auto");
   comment.rows = 1;
   comment.className =
     "min-h-8 max-h-24 min-w-0 flex-1 resize-none overflow-y-hidden border-0 border-b border-b-transparent bg-transparent px-0 py-1.5 font-sans text-sm leading-5 text-foreground outline-none ring-0 placeholder:text-muted-foreground focus:border-b-primary focus:outline-none focus:ring-0";


abdelrhmanehab10:
Closing to review locally before resubmitting.

PR #11015, opened 2026-09-10
fix(web): keep the current worktree on the new-local-thread shortcut

### LINKED ISSUE #9656: [Bug]: Command+Shift+N no longer creates a chat in the current worktree
> *This was generated by AI during triage.*

## Before submitting

- [x] I searched existing issues and did not find an open duplicate.
- [x] I included enough detail to reproduce and investigate the problem.

## Area

apps/web

## Steps to reproduce

1. Open a thread that is attached to a feature branch or worktree.
2. Press Command+Shift+N to quickly start another chat for parallel work in the same project.
3. Type a prompt and send it without changing the workspace controls.
4. Inspect the new thread's branch and working directory.

## Expected behavior

Command+Shift+N should create the new chat in the branch and worktree of the thread that was active when the shortcut was pressed.

This was the shortcut's original purpose. PR #56 introduced `mod+shift+n` as "new chat with same Git state."

## Actual behavior

The new chat keeps the project but drops the active branch and worktree. It uses the project's configured new-thread defaults instead, which can place the agent in the main checkout, another branch, or a fresh worktree.

This is extremely frustrating during feature work. The natural flow is to press Command+Shift+N, send a small parallel task, and return to the original chat. The new draft looks close enough to the current context that it is easy to send the message before noticing the workspace changed. It often takes longer than I would like to admit to realize that the agent has been working in the wrong worktree.

The shortcut still maps to `chat.newLocal`, but its handler now calls the same generic `startNewThreadFromContext` path as an ordinary new thread. That function intentionally inherits only the project. It does not pass the active branch or `worktreePath`.

## Impact

An agent can inspect or edit the wrong checkout before the user notices. Apart from wasted time, this can put changes on the wrong branch and make parallel work unreliable.

## Regression history

- PR #56 added Command+Shift+N specifically for a new chat with the same Git state.
- PR #4411 later removed implicit workspace inheritance from every new-thread entry point. This also changed the behavior of the dedicated `chat.newLocal` shortcut.
- PR #3918 proposed preserving checkout context again, but was closed without merging.
- Issue #3796 describes the related need to start a clean conversation in the current worktree. PR #3798 attempted an explicit worktree picker, but was closed without merging.
- PR #8517 adds a picker for existing worktrees. It does not restore the fast shortcut behavior.

## Suggested fix

Keep the generic new-thread commands aligned with project defaults, but restore the distinct behavior of `chat.newLocal`:

- Resolve the active thread or draft's `branch`, `worktreePath`, and environment mode.
- Pass that workspace context explicitly to `handleNewThread`.
- Keep Command+Shift+N bound to this command.
- Add a regression test proving the shortcut preserves both a local feature branch and an existing worktree.

## Version or commit

Current `main` as of September 4, 2026.

## Environment

macOS, web/desktop client

## Workaround

Create the draft, then manually select the previous or existing worktree before sending. A thread's context menu may also offer "New thread on branch." Both are slower and easy to forget when the shortcut suggests that it already preserves the current Git state.

maria-rcks:
Cmd/Ctrl+Shift+N (`chat.newLocal`) routed through the generic `startNewThreadFromContext`, which inherits only the project by design, so a quick parallel chat started from a feature-branch thread silently landed on the project's configured new-thread defaults instead of the branch and worktree the user was looking at.

The shortcut now goes through its own `startNewLocalThreadFromContext`, which reads the active thread or draft's `branch`, `worktreePath`, and env mode and passes them to `handleNewThread` — the same explicit options the thread menu's "New thread on <branch>" already uses, with `startFromOrigin: false` so reusing a checkout never re-bootstraps it. Every other new-thread entry point (Cmd+N, the sidebar button, the command palette) keeps the #4411 project-defaults behavior.

Verified: `vp run --filter @t3tools/web typecheck` exit 0, no errors. `vitest run apps/web/src/lib/chatThreadActions.test.ts --pool=forks` 12 passed, including two new cases covering worktree inheritance and the branch-only local case (`vp test run` fails to boot its default thread pool in my environment, on unmodified `origin/main` files too, so I ran the same runner with `--pool=forks`). `vp lint` on the three changed files exit 0.

Out of scope: the sidebar's "new thread" button advertises the `chat.newLocal` shortcut label as the keyboard twin of shift+click in single-project setups. Shift+click still uses project defaults, so that tooltip pairing is now slightly loose; changing it would touch a second entry point's behavior, which this fix deliberately leaves alone.

Fixes #9656

### UI evidence

Compared parent `b7b3ef1e6f` and head `d40a3b4` using the same real sandbox repository and existing `feature/worktree-shortcut` worktree. Selected that worktree in the branch picker, entered an unsent draft, then pressed Cmd+Shift+N. The parent resets the new draft to `main`; this change retains the selected branch and worktree. This verifies the client draft flow; provider execution was not tested.

before: shortcut resets the new draft to main and current checkout

after: shortcut preserves feature/worktree-shortcut and current worktree

cmd shift n creates a new draft while preserving the selected worktree

Evidence captured with gpt-6-astra (low reasoning) in Codex.

Done by Claude Opus 5 (1M context) in Claude Code.

## Summary by CodeRabbit

* **New Features**
  * Starting a local chat thread now preserves the currently viewed branch and worktree.
  * Local threads retain the appropriate environment settings and remain on the local checkout rather than starting from the origin.

* **Bug Fixes**
  * Improved behavior when creating local threads from active chats or drafts.
  * Added fallback handling when no active thread or draft is available.

### DIFF
--- a/apps/web/src/lib/chatThreadActions.test.ts
+++ b/apps/web/src/lib/chatThreadActions.test.ts
@@ -11,6 +11,7 @@ import {
   hasExplicitComposerModelSelection,
   resolveNewDraftStartFromOrigin,
   resolveNewThreadModelSelectionOverride,
+  startNewLocalThreadFromContext,
   startNewThreadFromContext,
   type ChatThreadActionContext,
 } from "./chatThreadActions";
@@ -153,6 +154,54 @@ describe("chatThreadActions", () => {
 
     expect(didStart).toBe(true);
     expect(handleNewThread).toHaveBeenCalledWith(scopeProjectRef(ENVIRONMENT_ID, PROJECT_ID));
+  });
+
+  it("keeps the active thread's worktree for the new-local shortcut", async () => {
+    const handleNewThread = vi.fn<ChatThreadActionContext["handleNewThread"]>(async () => {});
+
+    const didStart = await startNewLocalThreadFromContext(
+      createContext({
+        activeThread: {
+          environmentId: ENVIRONMENT_ID,
+          projectId: PROJECT_ID,
+          branch: "feature/one",
+          worktreePath: "/tmp/worktrees/feature-one",
+        },
+        handleNewThread,
+      }),
+    );
+
+    expect(didStart).toBe(true);
+    expect(handleNewThread).toHaveBeenCalledWith(scopeProjectRef(ENVIRONMENT_ID, PROJECT_ID), {
+      branch: "feature/one",
+      worktreePath: "/tmp/worktrees/feature-one",
+      envMode: "worktree",
+      startFromOrigin: false,
+    });
+  });
+
+  it("keeps a branch-only thread on the local checkout for the new-local shortcut", async () => {
+    const handleNewThread = vi.fn<ChatThreadActionContext["handleNewThread"]>(async () => {});
+
+    await startNewLocalThreadFromContext(
+      createContext({
+        activeDraftThread: {
+          environmentId: ENVIRONMENT_ID,
+          projectId: PROJECT_ID,
+          branch: "feature/two",
+          worktreePath: null,
+          envMode: "local",
+        },
+        handleNewThread,
+      }),
+    );
+
+    expect(handleNewThread).toHaveBeenCalledWith(scopeProjectRef(ENVIRONMENT_ID, PROJECT_ID), {
+      branch: "feature/two",
+      worktreePath: null,
+      envMode: "local",
+      startFromOrigin: false,
+    });
   });
 
   it("does not start a thread when there is no project context", async () => {
--- a/apps/web/src/lib/chatThreadActions.ts
+++ b/apps/web/src/lib/chatThreadActions.ts
@@ -15,6 +15,15 @@ type ComposerModelSelectionState = Pick<
 interface ThreadContextLike {
   environmentId: EnvironmentId;
   projectId: ProjectId;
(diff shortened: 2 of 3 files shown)


maria-rcks:
> [!NOTE]
> Written by `claude-fable-5-1` on behalf of Maria

Converting to draft pending a maintainer call. Two independent reviews of this head found the same problem: `chat.newLocal` never carried branch/worktree state (pre-#4411 it called `handleNewThread(projectRef)` with no options; `chat.new` did the carrying), and `docs/user/keybindings.md` documents both shortcuts as using new-thread defaults. Repurposing it is a product decision rather than a bug fix, and the change also flips `startFromOrigin` on an already-open draft.

If a shortcut for "New thread on <branch>" is wanted, the cleaner route is a new keybinding rather than changing this one. Happy to rework it that way if a maintainer confirms.

shivamhwp:

Preserve an origin-based draft's `startFromOrigin` when `worktreePath` is null. There is no existing checkout to reuse in that case, and forcing false changes where the pending worktree will start. The current new-thread handler applies these explicit options even when it reuses the already-open draft.

The product-decision hold remains appropriate. Before #4411, `startNewLocalThreadFromContext` passed only the project, while the generic action carried checkout context. The current documentation still says both shortcuts use project defaults. Please resolve that intended behavior before repurposing the shortcut.

PR #11019, opened 2026-09-10
perf(client): reduce repeated sorting and date formatting

Bil0000:
## What Changed

Thread ordering reparses the same dates during each sort comparison, project selection sorts every matching thread to return one, and hourly usage labels repeatedly construct identical `Intl.DateTimeFormat` instances. This PR removes that repeated work in three production files.

- Parse sort timestamps once per thread. Keep the existing order keys, ID/environment tie breaks, stable ordering, and original thread references. Find the latest project thread in one scan.
- Filter plan updates before sorting them, so unrelated tool activity does not enter the plan sort.
- Reuse date formatters for explicit time zones, with a cache capped at 16 entries. Leave implicit local-time formatting uncached so system time-zone changes still take effect.

No dependencies, settings, wire contracts, or visual changes.

## Benchmarks

Measured actual before/after exports from base `8d8189e67d` and this branch. These are **function-level wall times**, not whole-app CPU, memory, GPU, or battery measurements.

| Workload | Chromium before | Chromium after | Less time |
| --- | ---: | ---: | ---: |
| Sort 1,000 threads by recent activity | 1.401 ms | 0.262 ms | 81.3% |
| Sort 1,000 active threads without order keys | 1.205 ms | 0.306 ms | 74.6% |
| Sort 1,000 pinned threads without order keys | 1.317 ms | 0.357 ms | 72.9% |
| Select latest project thread from 1,000 threads | 1.461 ms | 0.208 ms | 85.8% |
| Derive a plan from 500 activities / 5 plan updates | 0.01537 ms | 0.00495 ms | 67.8% |
| Format 24 hourly usage labels and tooltips | 12.115 ms | 0.375 ms | 96.9% |

The usage batch saves about 11.7 ms of main-thread work in this fixture. The plan change is a small absolute saving. Benefits depend on history size and how often the caller runs.

Method: 200 warm-up calls per version, nine batches alternating before/after order, median time per operation. Each batch ran 100 list operations, 10,000 plan operations, or 20 usage batches. Fixtures match the committed benchmark: a fixed timestamp permutation for the threads, sequence-ordered activities, and `America/New_York` hourly formatting. Timings include allocation and garbage collection and varied between runs. The warm cache is intentional for repeated chart labels.

Separate instrumentation, outside timed runs, counted **5,794 → 1,000 `Date.parse` calls** per list operation and **120 → 0 formatter constructions** per warm hourly batch. The formatter cache retains at most 16 instances; total process memory was not measured.

<details>
<summary>Reproduce with the committed Node benchmark</summary>

`apps/web/src/performance.bench.ts` runs the same six workloads with the existing test runner. Copy that file to a checkout of base `8d8189e67d` for the before run. From `apps/web` in each checkout:

```sh
# Base checkout, with the benchmark file copied in
vp test bench src/performance.bench.ts --run --outputJson /tmp/t3-perf-before.json

# PR checkout
vp test bench src/performance.bench.ts --run --compare /tmp/t3-perf-before.json --outputJson /tmp/t3-perf-after.json
```

Node median results, in the same row order as the browser table:

| Workload | Node before | Node after |
| --- | ---: | ---: |
| Recent threads | 1.635046 ms | 0.254633 ms |
| Active threads | 1.403014 ms | 0.363550 ms |
| Keyless pinned threads | 4.716217 ms | 0.391587 ms |
| Latest project thread | 1.488331 ms | 0.204262 ms |
| Plan state | 0.014992 ms | 0.009198 ms |
| Hourly labels and tooltips | 10.591893 ms | 0.587797 ms |

Node and browser numbers are separate runs, not interchangeable. The browser measurements use the actual bundled modules in Chromium; the command above reproduces the Node suite.
</details>

## Verification

- 407 focused tests passed across shared formatting, shared/web thread ordering, plan state, sidebar, command palette, and timeline logic.
- Added coverage for invalid dates, both sort modes, ties, archived/project filtering, unchanged input/references, mixed plan activity, explicit time zones, and changes to the system time zone. Existing DST tests pass.
- 14,400 additional deterministic comparisons against the base exports returned identical outputs and thread references, including mixed order keys, bad dates, plan clears, DST transitions, and cache eviction.
- Scoped type checks for client-runtime, shared, and web passed. Targeted lint/format checks, web file-discovery check, and production web build passed.
- Verified with 12 synthetic threads and 500 activities: pinned/recent order, thread navigation, command palette, expanded work log, composer edit/clear, usage chart hover, daily/hourly switching, and wide/narrow layouts.

Shared logic serves web/desktop and mobile callers. Native desktop/mobile runtime tests and power measurements were not performed. No provider adapter or connection behavior changes.

## Checklist

- [x] Small changes limited to repeated client-side work
- [x] Explained the changes and measured their effect
- [x] Added regression coverage and a repeatable benchmark
- [x] No visual or animation change requiring before/after media

Model: GPT-6. Harness: Codex.

## Summary by CodeRabbit

* **Improvements**
  * Improved date and time formatting performance across usage and activity views.
  * Increased consistency when displaying hourly usage across time zones, including system-default time zone handling.
  * Improved thread ordering and selection reliability, including consistent tie-breaking for equally dated threads.
  * Improved active plan state handling when activity history includes unrelated events.

* **Tests & Performance**
  * Added broader automated coverage and performance benchmarks for thread sorting, plan selection, and usage formatting.


Bil0000:
Reviewed the advisory docstring-coverage warning. This patch adds no public API, preserves the existing behavior, and includes focused tests for the changed paths. Extra docstrings solely to meet a coverage percentage would add noise to this small performance change, so I am leaving that advisory warning unchanged. The CodeRabbit check passed and reported no actionable code findings.

PR #11021, opened 2026-09-10
fix(usage): flag unpriced model activity instead of showing $0.00

### LINKED ISSUE #5799: [Bug]: Usage page displays unpriced activity as zero cost
### Before submitting

### Area

apps/web

### Steps to reproduce

1. Add or encounter a Claude or Codex transcript whose model is not present in the loaded LiteLLM rate table.
2. Open Usage for a window containing that activity.
3. Select the cost view and inspect the headline and model/day breakdown.

The server correctly includes the tokens, sets the cost source to `unpriced`, and increments `unpricedRecords`.

### Expected behavior

The page should visibly say that the API-equivalent cost is unavailable or partial. Fully unpriced rows must not look like a genuine `$0.00` result, and mixed totals should indicate that the displayed dollar amount is a lower bound.

### Actual behavior

The page formats missing cost as `$0.00` and does not expose the already-computed unpriced coverage. The cost-quality data reaches the client but is not rendered after the cost-quality panel was removed.

### Impact

### Version or commit

main @ 963ebf5bd7cce00d40ff60c258b34c12dcab271e

### Environment

Web and desktop Usage page; provider-independent.

### Logs, stack traces, or screenshots

```shell
UsageBucket.unpricedRecords > 0
MergedUsage.costQuality.unpricedShare > 0
Rendered value: $0.00
```

### Workaround

Switch to the token view and ignore dollar estimates for windows containing unknown models.

maria-rcks:
The Usage page treated a model with no entry in the rate table as if it cost nothing: the server sets `costSource: "unpriced"` and counts `unpricedRecords`, but the client formatted the resulting zero as a literal `$0.00` and never showed the unpriced coverage it already receives.

`mergeUsage` now carries `unpricedRecords` per model and exports `isModelCostUnknown`. Web and mobile model rows render "Unpriced" instead of `$0.00` when every record for that model lacked rates, and the web cost headline says what share of records the estimate leaves out, so a mixed total reads as a lower bound.

Verified: `vitest run packages/shared/src/usageMerge.test.ts` 13 passed; `vitest run apps/web/src/components/usage/UsagePage.test.tsx` 5 passed; `vp run --filter @t3tools/{shared,web,mobile,client-runtime} typecheck` clean; `vp lint` on the five changed files clean.

Fixes #5799

UI evidence: verified in the web app against a copy of real usage data. A model whose records all lack rates (`<synthetic>`) now reads `Unpriced` with `—` for share instead of `$0.00 / 0.0%`, and the cost headline reads `372 sessions · API estimate excludes 0.1% unpriced records`. Rows whose provider reported a cost of zero (`costSource: providerReported`) still show `$0.00`, which this change does not touch.

Before (`main`): every rate-less row reads `$0.00 / 0.0%`.

before: usage table shows $0.00 and 0.0% for models with no rate

After (this PR): the rate-less row reads `Unpriced` and the cost headline names the excluded share.

after: unpriced row and the excluded-share footnote in the cost headline

Done by Claude Opus 5 (1M context) in Claude Code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Usage summaries now distinguish models without known pricing from models with zero cost.
  - Unpriced models are labeled **“Unpriced”** or **“no known rates”** instead of appearing free.
  - Token totals remain visible for unpriced models, while cost-share and USD values are shown only when pricing is available.
  - Usage percentages and model ordering now accurately include unpriced records.

### DIFF
--- a/apps/mobile/src/features/usage/UsageRouteScreen.tsx
+++ b/apps/mobile/src/features/usage/UsageRouteScreen.tsx
@@ -2,6 +2,7 @@ import { EnvironmentId, USAGE_CONTRACT_VERSION } from "@t3tools/contracts";
 import { useNavigation } from "@react-navigation/native";
 import {
   isCompatibleUsageContractVersion,
+  isModelCostUnknown,
   type DailyTotals,
   type MergedUsage,
 } from "@t3tools/shared/usageMerge";
@@ -607,10 +608,14 @@ function ModelsSection(props: { readonly merged: MergedUsage }) {
               {model.model}
             </Text>
             <Text className="text-sm text-foreground-muted">
-              {formatPercent(model.costShare)} of cost · {formatTokens(model.totalTokens)} tokens
+              {isModelCostUnknown(model)
+                ? `no known rates · ${formatTokens(model.totalTokens)} tokens`
+                : `${formatPercent(model.costShare)} of cost · ${formatTokens(model.totalTokens)} tokens`}
             </Text>
           </View>
-          <Text className="text-base tabular-nums text-foreground">{formatUsd(model.costUsd)}</Text>
+          <Text className="text-base tabular-nums text-foreground">
+            {isModelCostUnknown(model) ? "Unpriced" : formatUsd(model.costUsd)}
+          </Text>
         </View>
       ))}
     </SettingsSection>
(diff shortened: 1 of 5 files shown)


shivamhwp:

The model rows are fixed, but #5799 also covers the day breakdown. An all-unpriced day still shows `$0.00` in both the provider and total columns, and its provider summary still shows `$0.00` with `0.0% of cost`. Carry unpriced coverage into those aggregates and distinguish unknown cost from a genuine reported zero there too. Keep partial estimates visibly marked when priced and unpriced records share a row.

PR #11024, opened 2026-09-10
fix(desktop): stop opening a project from creating it twice

### LINKED ISSUE #5958: [Bug]: Can't open project
### Before submitting

### Area

apps/desktop

### Steps to reproduce

open a project

### Expected behavior

can't open a project. see screen recording. couldnt upload to github with "Failed to upload file. Please try again." so here's youtube: https://youtu.be/FtRUQKx73WU

### Actual behavior

can use the app by opening a project

### Impact

Blocks work completely

### Version or commit

0.0.33-nightly.20260810.1055 (3d74474f6cf8)

### Environment

macos 26.6 (25G72)

### Logs or stack traces

```shell
2nd attempt gives 
Orchestration command invariant failed (project.create): Active project '774fd081-dd3e-422b-8228-cfcd968926c3' already exists for workspace root '/Users/user/Documents/proj/pacey/expo'.
```

_No response_

### Workaround

none

maria-rcks:
Opening a project from the desktop app could dispatch `project.create` twice for the same workspace root: two activation requests can reach `handleDesktopAppActivationRequest` before the client projection shows the first create (the broker re-dispatches after a renderer blip, and the coordinator's serializing queue lives in a `useRef` that resets when the component remounts), so the second create is rejected by the server's `Active project '...' already exists for workspace root '...'` invariant and the open fails.

Racing activations now share a single in-flight create keyed by environment plus normalized workspace root, and a rejected create falls back to whatever project has since appeared for that root instead of failing the request outright. Other create call sites (command palette, welcome wizard, mobile add-project) are single user-initiated dispatches, not this race, so they are out of scope.

Verified: `vp test run src/desktopAppActivation.test.ts` in `apps/web` passes 6/6, and both new cases fail on `origin/main`. `vp run --filter @t3tools/web typecheck` exits 0 (only pre-existing suggestion diagnostics in unrelated files). `vp lint apps/web/src/desktopAppActivation.ts apps/web/src/desktopAppActivation.test.ts` exits 0.

Fixes #5958

UI evidence: pending (parent will attach)

Done by Claude Fable 5.1 in Claude Code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate project creation when multiple activations occur simultaneously for the same workspace.
  - Improved recovery when project creation reports a failure but the project becomes available through another activation.
  - Ensured existing projects are reopened and shared activation requests receive consistent results.

### DIFF
--- a/apps/web/src/desktopAppActivation.test.ts
+++ b/apps/web/src/desktopAppActivation.test.ts
@@ -75,6 +75,50 @@ describe("desktop app activation", () => {
     expect(response).toMatchObject({ ok: true, projectId: createdProjectId });
   });
 
+  it("creates one project when two activations race for the same workspace root", async () => {
+    let release = () => {};
+    const pending = new Promise<void>((resolve) => {
+      release = resolve;
+    });
+    const deps = dependencies({
+      findProject: () => null,
+      createProject: vi.fn(async () => {
+        await pending;
+        return createdProjectId;
+      }),
+    });
+
+    const first = handleDesktopAppActivationRequest(request, deps);
+    const second = handleDesktopAppActivationRequest({ ...request, requestId: "request-2" }, deps);
+    release();
+    const responses = await Promise.all([first, second]);
+
+    expect(deps.createProject).toHaveBeenCalledTimes(1);
+    expect(responses).toMatchObject([
+      { ok: true, projectId: createdProjectId },
+      { ok: true, projectId: createdProjectId },
+    ]);
+  });
+
+  it("reuses a project that appeared while the duplicate create was rejected", async () => {
+    let created = false;
+    const deps = dependencies({
+      findProject: () =>
+        created
+          ? { id: existingProjectId, environmentId, workspaceRoot: request.workspaceRoot }
+          : null,
+      createProject: vi.fn(async () => {
+        created = true;
+        throw new Error("Active project 'project-existing' already exists for workspace root.");
+      }),
+    });
+
+    const response = await handleDesktopAppActivationRequest(request, deps);
+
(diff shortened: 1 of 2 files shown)


maria-rcks:
> [!NOTE]
> Written by `claude-fable-5-1` on behalf of Maria

Converting to draft. Two independent reviews of this head reached the same conclusion: #5958 was filed on the 2026-08-10 nightly, before the CLI `open-workspace` activation path this PR guards existed, and the reporter's screenshots show the in-app add-project flow (command palette / welcome wizard), which already checks for an existing project before creating. The reporter's second create was sent because the projection never showed the first project, and that sync gap is what still needs a fix. The in-flight map here also only spans the create RPC, not the `waitForProject` window where an overlapping request would actually collide.

Leaving this as a draft rather than closing in case the catch-block fallback is still wanted on the CLI path. It does not fix #5958 as filed.

shivamhwp:

The shared promise is removed as soon as `createProject` returns, before `waitForProject` observes the projection. A second activation during that gap still sees no project and calls `createProject` again. If its rejection arrives before the projection, the fallback also sees nothing and opening fails.

Keep the shared operation registered through `waitForProject`, and remove it only after the project is visible or the whole operation fails. Add a case where the first create has resolved, its projection wait is still pending, and another activation arrives.

The issue relationship also needs correction. #5958 was reported on August 10, while this desktop activation handler arrived in #8824 on September 1. The issue describes adding a project in the app and subsequent chat hangs; this PR changes the newer CLI/desktop activation path. Remove `Fixes #5958` unless you establish that the original flow reaches this handler and reproduce the reported failure there.

PR #11025, opened 2026-09-10
feat(mobile): show final answers in completion notifications

catfogtoad:
## What Changed

When one agent finishes, the mobile completion alert shows the turn's final answer as plain text. The thread title stays in the notification title. Paragraphs, lists, links and code remain readable. For example, `Done: Project` becomes `Fixed the save error. Tests pass.`

This covers ordinary iOS push notifications, alerts sent with iOS Live Activity updates, and Android notifications. Long bodies are shortened at a sentence boundary when possible, with an ellipsis. Empty answers use the existing status body. Grouped alerts keep their list of thread titles. Tapping an alert opens the thread.

## Why

The status body tells the user that a turn ended, but gives no result. The final answer lets the user read the result from the notification.

The server saves buffered assistant output before it publishes the terminal session. A narrow query reads the latest turn's final answer. Optional relay fields carry the response and the bounded plain text body. Push bodies are checked against their encoded byte limits. Widget content omits the answer to keep the payload small.

## Validation

- 449 tests passed across 15 focused test files with Node 24.13.1.
- Type checks passed for the server, relay, client runtime and contracts.
- Lint and formatting checks passed for the changed TypeScript and JSON files.
- Tests cover final-answer selection, buffered output order, Markdown conversion, grouped alerts, fallback text, Unicode, routing, and payload limits.
- Added an Android test for an expanded notification with a long answer. It was not run here because the Android SDK and JDK 21 are not installed.

The server caps source Markdown at 16,000 characters before sending it to the relay. Push limits can require a shorter excerpt. The full answer remains in the thread.

## UI Changes

Before-and-after screenshots still need to be captured on a mobile client.

A relay and server update is required. Android needs a new native build to display bodies longer than the previous 608-character limit.

## Checklist

- [x] This PR has one scope.
- [x] I explained what changed and why.
- [x] Focused TypeScript tests and checks pass.
- [ ] Run the Android native test with JDK 21 and the Android SDK.
- [ ] Add before-and-after notification screenshots.
- [ ] Check the notification-to-thread tap on a mobile client.

Implemented with Codex, model GPT-6.

### DIFF
--- a/apps/mobile/modules/t3-agent-notifications/android/src/main/java/expo/modules/t3agentnotifications/AgentNotifications.kt
+++ b/apps/mobile/modules/t3-agent-notifications/android/src/main/java/expo/modules/t3agentnotifications/AgentNotifications.kt
@@ -141,8 +141,8 @@ object AgentNotifications {
       // so a delivery retry cannot surface them after the app backgrounds.
       if (!ProcessLifecycleOwner.get().lifecycle.currentState.isAtLeast(Lifecycle.State.RESUMED)) {
         val title = data["alert_title"].orEmpty().take(120)
-        // Grouped alerts list up to five 120-character thread titles.
-        val body = data["alert_body"].orEmpty().take(608)
+        // The relay fits the full response to FCM's 4 KB payload budget.
+        val body = data["alert_body"].orEmpty().take(4096)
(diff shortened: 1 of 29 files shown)


juliusmarminge:

Issue #6074, opened 2026-08-10
[Feature]: Send native Windows notifications when an agent task finishes


ElliotDrel:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I am describing a concrete problem or use case, not just a vague idea.

### Area

`apps/desktop`, Windows notifications, thread lifecycle, and T3 Connect

### Problem or use case

When I start a long-running task in T3 Code on Windows, I often switch to another application while the agent works. I want T3 Code to notify me through the native Windows notification system when the task is actually finished so I do not have to keep checking the app.

This is especially useful for long-running Claude Code, Codex, and other provider sessions where the work may take several minutes or more.

### Proposed solution

Add native Windows completion notifications for settled agent runs.

When a thread finishes its current logical task and no longer has active work, T3 Code should send a Windows toast notification using the normal Windows notification center.

### Expected behavior

- Send a native Windows notification when an agent task/run genuinely finishes.
- Include enough context to identify the result, such as:
  - project or general-chat name;
  - thread title;
  - provider when useful;
  - a short message such as **Task completed**.
- Clicking the notification should open T3 Code directly to the relevant thread.
- Notifications should work even when T3 Code is minimized or not focused, as long as the desktop app is still running.
- Respect Windows notification permissions and Focus/Do Not Disturb settings.
- Provide a T3 Code setting to enable or disable task-completion notifications.
- Avoid duplicate notifications for the same completed run.
- Work consistently for local and remote/T3 Connect sessions being viewed from the Windows desktop client.

### Completion semantics

The notification should represent a meaningful user-facing stopping point, not every internal provider lifecycle transition.

Do **not** send a completion notification when:

- the parent agent is temporarily waiting on active subagents;
- a provider process rotates or restarts internally;
- the agent is still running tools or delegated work;
- the thread has not actually settled.

A notification should fire when:

- the logical run has completed and the final result is ready;
- the run fails and requires user attention, if error notifications are enabled;
- the agent is waiting for explicit user input or approval, if attention notifications are enabled.

### Why this matters

The value of long-running agent workflows is reduced if I have to keep polling the app to see whether the work is done. Native Windows notifications let T3 Code behave like other background productivity tools and make it much easier to supervise several concurrent tasks.

### Smallest useful scope

- Native Windows toast on final task completion.
- Notification includes the thread title.
- Clicking it opens the relevant thread.
- One on/off setting for completion notifications.
- No notification while subagents or other work are still active.

[...]

Issue #5435, opened 2026-08-05
[Feature]: Add an unread indicator for completed sessions


magofdl:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I am describing a concrete problem or use case, not just a vague idea.

### Area

apps/web

### Problem or use case

On iOS/iPad apps when a session is actively running, the sessions panel displays a “Working” status. However, once the task finishes, the status disappears and there is no visual indication that the session has completed or that its latest output has not yet been reviewed.

This becomes difficult when working with multiple projects or sessions in parallel, because users cannot easily identify which sessions have finished and still require their attention.

It would be helpful to display an unread or unreviewed indicator for completed sessions, similar to the blue dot Codex shows next to sessions that have not yet been opened.

### Proposed solution

Add an unread or unreviewed state for sessions that finish while the user is viewing another session.

When a task completes, show a persistent visual indicator in the sessions panel, such as a blue dot next to the session name. The indicator should remain visible until the user opens that session.

Suggested behavior:

- Show “Working” while the task is running.
- Replace it with an unread indicator when the task finishes.
- Keep the indicator visible even after restarting the app.
- Clear the indicator when the user opens the completed session.
- Optionally, show a tooltip or accessibility label such as “Completed — not yet viewed.”

This would make it easier to manage several projects or sessions running in parallel and quickly identify which completed results still need review.

### Why this matters

When multiple projects or sessions are running in parallel, users need a reliable way to know which tasks have finished and still require review.

Currently, once the “Working” status disappears, a completed session becomes visually indistinguishable from one that has already been reviewed. This makes it easy to overlook finished work, delays feedback, and forces users to repeatedly open sessions to check their status.

A persistent unread indicator would reduce unnecessary context switching, make parallel workflows easier to manage, and help ensure that completed results are not missed.

### Smallest useful scope

Show a persistent unread indicator, such as a blue dot, when a session finishes while the user is viewing another session. Clear the indicator when the user opens that completed session.

No notifications, badges, tooltips, or additional status states are required for the initial implementation.

### Contribution

- [ ] I would be open to helping implement this.

Issue #780, opened 2026-03-10
Add notifications for completed tasks and approval-required turns.


sbansal1999:
Right now if a task finishes in a chat while the user is in another app, they get no signal. Same for approval-required turns in supervised mode.

We should notify users when:
- [ ] a task/turn completes
- [ ] an approval is required in supervised mode

obrunogonzaga:
I’d like to work on a small, focused version of this.

I reviewed #976, #1657, #1780, and the notification-sound foundation in #2373. My proposed scope would be limited to native OS notifications when T3 Code is not focused:

- turn completed
- approval or user input required
- turn failed
- click opens the relevant thread
- opt-in and disabled by default

No sounds, overlays, pets, tray UI, or broader lifecycle-hook system.

Would this scope be welcome if built on the newer thread-state foundation, or would you prefer to wait for #2373 first?

tsatsujnr139:
@obrunogonzaga Is there any movement on your PR? I was looking to file a similar issue and came across this.

PR #11040, opened 2026-09-10
feat(desktop): show unread completion badges

Andrew-Forster:
## What Changed

When one or more threads finish while T3 Code is minimized, unfocused, or showing another thread, the desktop app now shows how many completed threads are waiting to be reviewed.

Windows uses a custom taskbar overlay so the badge stays readable at taskbar size. Counts display as `1` through `9`, then `9+`. macOS and Linux use Electron's native numeric badge.

The badge uses the existing completed-unread state. Opening a completed thread decreases the count, and opening the last unread thread clears the badge.

On Linux, the badge appears on docks and taskbars that support the LauncherEntry D-Bus API. T3 Code already registers a matching `.desktop` identity before Electron is ready, so the implementation can use the same `app.setBadgeCount()` path as macOS without adding a separate Linux service.

This does not add toast notifications, sounds, settings, provider-specific behavior, or another notification store.

## Why

T3 Code users have repeatedly asked for a reliable signal when long-running work finishes outside the active thread:

- #780 reports that tasks finishing while the user is in another app provide no signal.
- #6074 requests Windows completion notifications for long-running Claude Code, Codex, and other provider sessions.
- #5435 requests a persistent completed-but-unreviewed indicator that clears when the thread is opened.

This exposes T3 Code's existing unread-completion state through native desktop badges instead of introducing notification permissions, routing, sounds, or new persisted state.

## UI Changes

| Before | One unread completion | Multiple unread completions |
| --- | --- | --- |
| Taskbar icon without a completion badge | Taskbar icon with one unread completion | Taskbar icon with multiple unread completions |

The screenshots show the custom Windows overlay. macOS and Linux use the system-rendered badge supplied by the user's dock or taskbar.

## Verification

- 178 focused tests pass across unread state, acknowledgement, desktop platform dispatch, Linux desktop identity, and Windows overlay behavior.
- The Linux-specific desktop and coordinator tests also pass on Ubuntu 24.04 x64.
- Targeted typechecks pass for `@t3tools/web`, `@t3tools/desktop`, and `@t3tools/contracts`.
- Targeted lint and formatting checks pass.
- Manually verified in the Windows development build:
  - background completion adds the badge;
  - multiple completions update the count;
  - opening a completed thread decreases the count;
  - opening the final unread thread clears it.
- Manually verified the native Dock badge on macOS.

## Checklist

- [x] I included before/after screenshots for the custom UI change
- [ ] I included a video for the interaction

Implemented with GPT-5 in the Codex desktop app.


Andrew-Forster:
Re: the docstring coverage warning, I’m leaving this as-is. The functions counted here are small components, IPC handlers, and test helpers where blanket docstrings would mostly repeat the names and types. The repository does not enforce a docstring threshold, and all required checks pass. I’ll keep comments limited to places where the reasoning is genuinely non-obvious so the PR stays focused.

Andrew-Forster:
If this gets accepted, I’d like to add the same unread badge to macOS and Linux next.
I kept this PR Windows-only to keep it focused. If we want support for the other platforms in this PR, or as a separate follow-up I have Windows, macOS, and Ubuntu machines, so I can implement and test all three.

Andrew-Forster:
image

Added mac-os support. Linux I'll still wait for PR approval

apps/desktop/src/electron/DesktopUnreadBadge.ts
```diff
@@ -0,0 +1,69 @@
...
+export function setDesktopUnreadBadge(input: {
+  readonly platform: NodeJS.Platform;
+  readonly window: Pick<Electron.BrowserWindow, "isDestroyed" | "setOverlayIcon"> | null;
+  readonly count: number;
+  readonly badgeDataUrl: string | null;
+}): boolean {
+  try {
+    if (input.platform === "darwin") {
+      return Electron.app.setBadgeCount(input.count);
+    }
```
tarik02, line 44:
```suggestion
    if (input.platform === "linux" || input.platform === "darwin") {
      return Electron.app.setBadgeCount(input.count);
    }
```

apps/desktop/src/electron/DesktopUnreadBadge.ts
```diff
@@ -0,0 +1,69 @@
...
+      return false;
+    }
+
+    input.window.setOverlayIcon(
+      overlay,
+      `${input.count} completed ${input.count === 1 ? "thread" : "threads"} awaiting review`,
+    );
+    return true;
+  } catch {
+    return false;
```
tarik02, line 67: mb avoid silently swallowing errors? better at least log a warning

tarik02 (changes requested):

apps/desktop/src/electron/DesktopUnreadBadge.ts
tarik02: oh wait, I see, sorry

apps/desktop/src/electron/DesktopUnreadBadge.ts
Andrew-Forster: Added a warning through Effect.logWarning when a badge update throws, while keeping the existing false return. The focused tests cover both Windows and macOS failures and verify that they are logged. All seven badge tests, desktop typecheck, lint, and formatting pass.

apps/desktop/src/electron/DesktopUnreadBadge.ts
Andrew-Forster: Added in `4e51482fd`. T3 Code now uses Electron 44.1.0, which supports Linux badges through the LauncherEntry D-Bus API, and the desktop app already registers a matching `.desktop` identity before Electron is ready. I kept this to the existing native `setBadgeCount()` path and added Linux coverage. The focused tests pass on both Windows and Ubuntu 24.04 x64. Visual availability still depends on the user's Linux dock or taskbar supporting LauncherEntry, as documented by Electron.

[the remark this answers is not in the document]
Andrew-Forster: Keeping this path. The PR scope now intentionally covers all three desktop platforms: Windows uses its custom overlay, while macOS and Linux use Electron's native badge count. macOS was manually verified, Linux is covered on Ubuntu 24.04, and the PR description now reflects the cross-platform behavior.

Andrew-Forster:
Linux support is now included and manually verified on Ubuntu 24.04 with GNOME and Ubuntu Dock.
image

PR #11041, opened 2026-09-10
feat(web): favorite projects in the project selector

MatthewFeroz:
## What Changed

- Added star controls to project rows in the **New thread in…** selector.
- Added a **Favorites** section above the remaining projects and kept jump shortcuts aligned with the visible order.
- Persisted favorites using each member’s physical environment/path key. Grouping changes and per-project overrides preserve favorites; starring a group saves all members, and unstarring removes all of them.
- Added Favorites to the empty-draft project chooser and `Alt+Enter` to toggle the highlighted palette project.
- Kept the selector live while open: favoriting and unfavoriting moves the project immediately without selecting it or opening a thread.

## Why

Large project lists make frequently used workspaces slow to reach. This adds the same lightweight favorite behavior already used by model selection, while keeping the preference client-local and avoiding server, provider, or database changes.

## UI Changes

| Before | After |
| --- | --- |
| Project selector before favorites | Project selector with a Favorites section |

The original interaction was verified in the web client: the star updates the open selector immediately, does not navigate, and persists across a full reload. A video is not applicable because this change does not add or depend on motion, timing, or transitions.

## Verification

- Focused palette and settings tests — 166 tests passed, including split/regroup toggles, successive updates, and current-favorite shortcut ordering
- `vp run --filter @t3tools/contracts typecheck`
- Scoped `vp lint` on all changed files
- `git diff --check`
- Original integrated browser pass covered favorite, unfavorite, immediate regrouping, no navigation, and reload persistence. The latest review fixes passed focused tests and web typechecking; updated browser verification and screenshots remain pending.

Web and desktop share both selectors. Favorites work with local and remote environments and remain client-local. Mobile uses a separate native selector and is unchanged. No provider, wire, server, connection-mode, or database behavior changes.

## Checklist

- [x] Video not applicable; there are no animation or timing changes

Model: GPT-6 (review fixes), GPT-5.6 Sol (original implementation). Harness: Codex.

## Summary by CodeRabbit

* **New Features**
  * Added project favorites to the “New thread in…” command palette.
  * Favorite projects appear in a dedicated section with numbered shortcuts.
  * Added accessible star controls, tooltips, and Alt+Enter support for toggling favorites.
  * Favorite selections persist across sessions and project-grouping changes.
  * The current project remains prioritized while browsing project choices.

* **Bug Fixes**
  * Empty Favorites sections are hidden when no projects are favorited.
  * Rapid favorite toggles now preserve each selection reliably.

### DIFF
--- a/apps/web/src/components/CommandPalette.tsx
+++ b/apps/web/src/components/CommandPalette.tsx
@@ -72,7 +72,7 @@ import { useDesktopLocalBootstraps } from "../connection/useDesktopLocalBootstra
 import { useHandleNewThread } from "../hooks/useHandleNewThread";
 import { useOpenPanelPullRequestUrl } from "../hooks/useOpenPanelPullRequestUrl";
 import { writeTextToClipboard } from "../hooks/useCopyToClipboard";
-import { useClientSettings, useUpdateClientSettings } from "../hooks/useSettings";
+import { persistClientSettingsUpdate, useClientSettings } from "../hooks/useSettings";
 import { useTheme } from "../hooks/useTheme";
 import { readLocalApi } from "../localApi";
 import { desktopLocalBackendId } from "../connection/desktopLocal";
@@ -588,7 +588,6 @@ function OpenCommandPaletteDialog(props: {
   const isActionsOnly = deferredQuery.startsWith(">");
   const [highlightedItemValue, setHighlightedItemValue] = useState<string | null>(null);
   const clientSettings = useClientSettings();
-  const updateClientSettings = useUpdateClientSettings();
   const createProject = useAtomCommand(projectEnvironment.create, {
     reportFailure: false,
   });
@@ -1104,10 +1103,16 @@ function OpenCommandPaletteDialog(props: {
             isFavorite,
             label: isFavorite ? "Remove from favorites" : "Add to favorites",
             toggle: () => {
-              void updateClientSettings({
-                favoriteProjectKeys: isFavorite
-                  ? clientSettings.favoriteProjectKeys.filter((key) => key !== group.projectKey)
-                  : [...clientSettings.favoriteProjectKeys, group.projectKey],
+              void persistClientSettingsUpdate((current) => {
+                const currentIsFavorite = current.favoriteProjectKeys.includes(group.projectKey);
+                return {
+                  ...current,
+                  favoriteProjectKeys: currentIsFavorite
+                    ? current.favoriteProjectKeys.filter((key) => key !== group.projectKey)
+                    : [...current.favoriteProjectKeys, group.projectKey],
+                };
+              }).catch((error) => {
+                console.error("Failed to update project favorite", error);
               });
             },
           };
@@ -1168,12 +1173,18 @@ function OpenCommandPaletteDialog(props: {
       pickerProjects,
       projectEnvironmentLocationById,
       projectGroupByTargetKey,
-      updateClientSettings,
     ],
   );
   const projectThreadGroups = useMemo(
     () => buildProjectSelectorGroups(projectThreadItems),
     [projectThreadItems],
+  );
+  const liveProjectThreadGroups = useMemo(
+    () =>
+      currentView?.projectSelectorPriorityValue
+        ? buildProjectSelectorGroups(projectThreadItems, currentView.projectSelectorPriorityValue)
+        : projectThreadGroups,
+    [currentView?.projectSelectorPriorityValue, projectThreadGroups, projectThreadItems],
   );
 
   const allThreadItems = useMemo(
@@ -1252,6 +1263,9 @@ function OpenCommandPaletteDialog(props: {
           addonIcon: view.addonIcon,
           groups: view.groups,
           ...(view.initialQuery ? { initialQuery: view.initialQuery } : {}),
+          ...(view.projectSelectorPriorityValue
+            ? { projectSelectorPriorityValue: view.projectSelectorPriorityValue }
+            : {}),
         },
       ]);
       setHighlightedItemValue(null);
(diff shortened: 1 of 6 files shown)


apps/web/src/components/CommandPalette.tsx
```diff
@@ -1820,7 +1839,9 @@ function OpenCommandPaletteDialog(props: {
           addProjectEnvironmentId,
           buildAddProjectRemoteSourceReadiness(sourceControlDiscovery.data),
         )
-      : (currentView?.groups ?? rootGroups);
+      : isProjectSelectorView
```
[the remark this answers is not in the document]
MatthewFeroz: Fixed in 1f88f4f95. Project-selector views now retain an explicit priority value, and live favorite groups rebuild from fresh items with that same priority. The grouping test now covers the contextual project staying first in its section with the expected shortcut order.

apps/web/src/components/CommandPalette.tsx
```diff
@@ -1092,68 +1093,88 @@ function OpenCommandPaletteDialog(props: {
...
+        favorite: (project) => {
+          const group = projectGroupByTargetKey.get(`${project.environmentId}:${project.id}`);
+          if (!group) return undefined;
+          const isFavorite = clientSettings.favoriteProjectKeys.includes(group.projectKey);
+          return {
+            isFavorite,
+            label: isFavorite ? "Remove from favorites" : "Add to favorites",
+            toggle: () => {
+              void updateClientSettings({
+                favoriteProjectKeys: isFavorite
```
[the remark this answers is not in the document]
MatthewFeroz: Fixed in 1f88f4f95. Favorite toggles now use persistClientSettingsUpdate and derive membership from current.favoriteProjectKeys when each queued update runs, so rapid toggles cannot overwrite an earlier change from the same render.

apps/web/src/components/CommandPalette.tsx
```diff
@@ -1092,68 +1092,100 @@ function OpenCommandPaletteDialog(props: {
...
-                </span>
-                <CommandPaletteMetaDot />
-                <span className="truncate">{project.workspaceRoot}</span>
+      buildProjectActionItems({
+        projects: pickerProjects,
+        valuePrefix: "new-thread-in",
+        favorite: (project) => {
+          const group = projectGroupByTargetKey.get(`${project.environmentId}:${project.id}`);
+          if (!group) return undefined;
+          const isFavorite = clientSettings.favoriteProjectKeys.includes(group.projectKey);
```
saphid, line 1101: **Blocking: favorites are keyed on a setting-dependent identifier.**

`group.projectKey` comes from `deriveLogicalProjectKeyFromSettings` (`packages/client-runtime/src/state/projectGrouping.ts`). It returns the repository canonical key in `repository` mode, `canonical::relpath` in `repository_path` mode, and `environmentId:path` in `separate` mode, and it also honors per-project overrides from the sidebar context menu.

So flipping the grouping mode in Settings, or overriding a single project in the sidebar, silently renames every key. All favorites disappear, and the old strings linger in `favoriteProjectKeys` forever. The PR body presents the logical key as a pure win, but this tradeoff is not mentioned.

The repo already persists per-project preferences by physical key (`sidebarProjectGroupingOverrides`, project ordering). I would follow that precedent: on star, store `derivePhysicalProjectKey` for every member of the group; on unstar, remove them all; treat a group as favorite when any member key matches. That behaves correctly in every grouping mode (in `separate` mode each group has one member) and survives mode changes.

apps/web/src/components/CommandPaletteResults.tsx
```diff
@@ -178,6 +179,12 @@ function CommandPaletteResultRow(props: {
       onMouseDown={(event) => {
         event.preventDefault();
       }}
+      onPointerDownCapture={(event) => {
```
saphid, line 182: This handler looks redundant. Base UI's `ComboboxItem` (which `AutocompleteItem` re-exports) already calls `preventDefault` in its own `onPointerDownCapture` and records that the pointer started on the item, so its `onMouseUp` will not commit a selection. The button's `onClick` with `stopPropagation` already blocks the item's click commit. `ModelListRow.tsx` sits on the same primitive and gets by with click `stopPropagation` alone.

Suggest dropping this block. While here, consider reusing that row's `Button size="icon-xs" variant="ghost"` treatment so the two star controls look the same.

apps/web/src/components/CommandPalette.tsx
```diff
@@ -1812,6 +1836,10 @@ function OpenCommandPaletteDialog(props: {
   }));
   const sourceSelectionViewValue =
     addProjectEnvironmentId === null ? null : `sources:${addProjectEnvironmentId}`;
+  const isProjectSelectorView =
```
saphid, line 1839: Detecting the project selector by scanning group values for `"projects"` / `"favorite-projects"` is a string sniff, and the stored `groups` snapshot is then discarded for this view anyway. `projectSelectorPriorityValue` is only a partial marker since it exists only when there is a current project.

A smaller model: give `CommandPaletteView` an explicit project-selector kind (carrying the optional priority value) and compute the groups live from it, rather than storing groups that are never rendered.

saphid (changes requested):
Thanks for this. The interaction itself is in good shape: the functional update through `persistClientSettingsUpdate`, live regrouping while the palette stays open, decode-time defaults in the contract, and no wire or server changes. Both earlier bot findings are genuinely fixed in 1f88f4f. CI is green.

I am requesting changes for one design issue with the persistence key, plus a few smaller items. Details inline; the non-inline points are below.

**Second entry point without favorites.** The project chooser on the empty draft headline (`apps/web/src/components/chat/DraftHeroHeadline.tsx`, the `MenuRadioGroup` built from `buildSidebarProjectPickerEntries`) lists the same picker entries with no Favorites section. AGENTS.md asks for a decision per entry point. Either wire it or say "not here" in the PR body.

**User docs.** `docs/user/thread-sidebar.md` documents the new-thread flow and the command palette. A user-visible feature needs a sentence there.

**Keyboard access (minor).** The star is effectively mouse-only. Focus stays in the palette input and every other palette button is `tabIndex={-1}`, so Tab never reaches it, and there is no keybinding. The model picker has the same gap, so not blocking, but worth a line in the description.

**Tests (minor).** The toggle handler and the case where the current project is itself a favorite are not covered. Both are thin, so low priority.

Reviewed with Claude Fable 5.1 in Claude Code.

PR #11047, opened 2026-09-10
feat(threads): move a started thread to another account of the same provider

chledowski:
A started thread can move to another configured account of the same provider after confirmation. T3 keeps the displayed transcript; the target starts a fresh provider conversation from the next message, without automatic history handoff or prompt replay.

Web/desktop and mobile offer the switch when supported by the server. The server checks the source account and its revision, so confirmation cannot be reused after an account round trip. Failed startup preserves the original owner and continuation for retry. Missing/stale consent and cross-provider-driver changes are rejected even without an active runtime or resume cursor.

The revision is a counter assigned in the command decider when ownership changes. Migration 051 adds a zero-default column; legacy events retain revision zero. Projection, replay, and reconnect consume the recorded value. There is no historical backfill or independent revision inference in each projection.

Send an ordinary message before compacting after a switch. Automatic deleted-owner recovery is deferred. Compatible Codex instances retain continuation on web/desktop; the inherited mobile picker limitation for shared continuation groups remains documented. This follows the manual-switching direction in #9181; it adds no quota ranking, automatic retries, or transcript migration.

### Validation and review closeout

Rebased onto upstream `57aee3e19` and consolidated into one commit. The final patch is 138 changed lines smaller than the previous revision (45 files, 1,972 additions and 127 deletions).

- 630 tests passed across 14 focused files, covering switching, missing/stale consent, failed startup/retry, account round trips, provider lifecycle events, migration/replay, and client state.
- Scoped server, web, mobile, contracts, and client-runtime typechecks passed; scoped lint and formatting passed.
- Live web cancellation kept the source selected. Confirmation and the next message produced a target reply with the original transcript visible. The database recorded target ownership at revision 1.
- Production web and iOS simulator builds passed on `d9e11c2e3`. Native selection, confirmation, send, target reply, and transcript retention passed; the database recorded the switched owner at revision 2.
- Independent review of `d9e11c2e3` used Claude Opus 5 at medium effort after configured Fable attempts hit quota/service-overload errors. Both findings have dispositions in final commit `501f224cb`.
- The final server fixes keep turns without explicit model selection on the current owner and clear old progress/liveness only after a successful incompatible-account switch. Failed startup preserves source work. The 156 tests across reactor, ingestion, and importer suites passed after these changes, along with server typecheck and scoped lint/formatting. Counts overlap the earlier 630-test pass; they are not a combined unique total.
- GitHub reports this head conflict-free. CI and all four preview/fingerprint workflows require maintainer authorization.

Live checks use two configured Codex homes, not a claim of distinct subscriptions or a real quota-exhaustion event. See the [consolidated disposition](https://github.com/pingdotgg/t3code/pull/11047#issuecomment-5617552790) for review history and limits.

### Web and native walkthroughs

Captured on `d9e11c2e3`; the subsequent `501f224cb` server-only review fixes received the focused validation above. UI states are captured with idle time condensed.

Web account switch

Native iOS account switch

### Historical acceptance walkthroughs (0a2153305)

These captures predate the simplified revision counter and upstream rebase. They show successful switching, injected startup failure/retry, and stale confirmation after a two-client account round trip; current-commit evidence is listed above.

Successful switch and retained transcript:

Successful switch and retained transcript

Injected startup failure and retry without reconfirmation:

Injected startup failure and retry without reconfirmation

Stale confirmation rejected after a two-client account round trip:

Stale confirmation rejected after a two-client account round trip

Implementation: Claude Opus 5 in Claude Code; follow-up audit, fixes, simplification and acceptance with GPT-6-Astra in Codex. Independent review uses the repository-configured review harness.


chledowski (commented):
## Independent Codex review

## Findings

- **P2 — Mobile account switching is unreachable** — `apps/mobile/src/state/use-thread-composer-state.ts:591–602`. The new confirmation handler cannot receive another account from the picker: `ThreadComposer.tsx:470` still filters groups to `currentModelSelection.instanceId`. With two Claude accounts configured, mobile users remain stuck on the original account. Update the picker to expose compatible same-driver accounts when the capability is enabled, retain the confirmation, and test selection and cancellation.

## Checked

- Verified exact head `0317ab2203dbc07890602a13658d33e9d5b36503`; reviewed its full 22-file diff against `upstream/main` (`d29c56a5c`). Local `main` is stale.
- Read repository guidance, changed tests, and surrounding picker, confirmation, routing, and session-persistence code.
- Reused supplied evidence: 301 passing tests and scoped server/web/mobile/client-runtime/contracts typechecks.
- Ran a Node reproduction using the actual mobile picker predicate; confirmed it excludes the second account.
- Additional Vitest execution was blocked by the read-only sandbox’s prohibition on Vite temporary files.
- GitHub access failed; PR body and linked issue context could not be verified. No edits or GitHub writes.

---

AI-generated review by Codex using `gpt-6-astra` at `low` effort.
Requested model: `gpt-6-astra`.

chledowski:
## Review disposition — all findings, `0317ab2` → `f2f1bb2`

Reviewers: Codex `gpt-6-astra` (`low`, the pinned review above) and a second `gpt-6-astra` (`medium`) pass I ran locally, plus Macroscope and CodeRabbit.

### Fixed

**Mobile account switching was unreachable** (Codex P2, Macroscope, CodeRabbit). Correct, and my own miss: I wired the confirmation without tracing what feeds the mobile picker. `ThreadComposer` filtered groups to `currentModelSelection.instanceId`. The picker now offers exactly the accounts `resolveProviderAccountSwitchPrompt` will confirm, so it cannot show one the composer then refuses — the rule lives in a pure module with tests (`apps/mobile/src/features/threads/thread-provider-account-switch.ts`). A thread that has not started yet is no longer restricted to its own account either; it was never bound to anything.

**A thread with no live session switched accounts without confirming** (Macroscope, CodeRabbit, both clients). Correct on both. `ChatView` resolved the current account from `thread.session?.providerInstanceId` alone while the composer's lock is `session ?? modelSelection`; an imported or reaped thread therefore got no prompt and moved silently. Both clients now resolve the lock the same way, and only for a thread that has actually started.

**A retained composer selection could switch the thread back, silently** (second `gpt-6-astra` pass). The composer draft's active provider is durable, survives the send that used it, and syncs across devices, so it is not evidence that *this* move was confirmed. Sequence: device A retains account 1, device B confirms 1 → 2 and sends, device A's retained selection then restarts the thread on account 1 with no prompt. Both clients now hold the confirmation in memory until the turn starts it; losing it on reload just returns the composer to the account the thread is on. `resolveComposerProviderSelection` takes `confirmedAccountSwitchInstanceId` instead of an explicit-pick id, which also removed the redundant `!requiresExactInstance` guard.

**A deleted bound account left the thread with no way forward** (Macroscope, High). Correct. `Effect.option(getInstanceInfo(bound))` yields `undefined`, which made `switchesProviderAccount` false, so no fresh start was requested and `startSession` rejected every account the user picked instead. An unresolvable bound account is now treated as a switch — nothing can resume its conversation. Regression test added; it fails without the change.

**`?? true` on the confirmation** (Macroscope, CodeRabbit). The `?? true` convention elsewhere reasons that "the click was the request"; here the click asked for a model, not for losing a conversation. Only an explicit confirm now moves the thread. The dialog host is mounted at the route root, so this is belt-and-braces.

**Docs.** `docs/user/providers-claude.md` and `docs/user/providers-codex.md` both stated that an existing thread can only move between same-home/same-config-dir accounts. Both now describe the switch and its confirmation.

### Simplified

Dropped `getBoundInstanceId` from the `ProviderService` facade: the reactor reads `ProviderSessionDirectory` directly, the same service `startSession` validates against, so detection and validation cannot disagree. That removed nine unrelated mock additions across six test files.

### Not changed, with reasons

**Cross-driver switch on a stopped-session thread** (CodeRabbit). Real, but pre-existing and untouched here: the reactor's driver guard only runs when `thread.session !== null`, and `startSession` skips its incompatibility guard when the persisted and requested drivers differ. This PR cannot widen it — a fresh start is only ever requested when the bound and requested drivers match. I removed the PR's blanket claim that driver switching "stays blocked" rather than tightening a path no client offers.

**Locked entry with no continuation metadata falls through to the candidate list** (CodeRabbit). All six drivers stamp `continuation.groupKey` unconditionally, so a server that advertises `threadProviderAccountSwitch` always sends it. On a server that omits it the capability is absent, no confirmed switch is ever produced, and the fall-through is upstream's existing behavior, unchanged by this PR.

### Verification

Scoped typechecks for `t3`, `@t3tools/web`, `@t3tools/mobile`; `vp test run` on the reactor, provider service, importer, checkpoint reactor, runtime ingestion, session reaper, startup reconcile, the orphaned-session and orchestration-engine integration suites, `ChatView.logic`, `providerInstanceDisplay`, and the new mobile module — 400+ tests, all passing. Lint and format on every changed file.

chledowski (commented):
## Independent Codex review

## Findings

- **P1 — Mobile sends on the old account after confirmation.** [use-thread-composer-state.ts:356](https://github.com/pingdotgg/t3code/blob/f2f1bb2b3ceec6ae5c1c14f1849f557911cd154d/apps/mobile/src/state/use-thread-composer-state.ts#L356): `onSendMessage` now reads `confirmedAccountSwitch`, but its `useCallback` dependencies omit that state. On an idle thread, confirming another account updates the picker while the send callback retains `null`, causing `resolveComposerModelSelection` to select the original account. Add `confirmedAccountSwitch` to the dependencies and test confirming a switch followed immediately by sending without a server update.

## Checked

- Verified head `f2f1bb2b3ceec6ae5c1c14f1849f557911cd154d`; reviewed its 25-file diff against cached `upstream/main` (`d29c56a5c`). Local `main` is behind.
- Read repository instructions, changed tests, web/mobile selection flows, server binding and restart logic, and provider session handling.
- A read-only Node check confirmed the resolver’s differing results and the missing callback dependency.
- Targeted tests and lint could not start: Vite’s temporary config write was denied by the read-only sandbox.
- GitHub CLI and web access failed; PR body and linked issue context remain unverified. No files changed or GitHub posts made.

---

AI-generated review by Codex using `gpt-6-astra` at `low` effort.
Requested model: `gpt-6-astra`.

chledowski:
## Disposition — `f2f1bb2` → `ef67334`

**P1, mobile sends on the old account after confirming — correct, fixed.** `onSendMessage`'s dependency array omitted `confirmedAccountSwitch`, so the callback captured the initial `null` and `resolveComposerModelSelection` fell back to the thread's own account: exactly backwards from what the confirmation asked for. The dependency is now declared.

This one was already visible and I misread it: `vp lint` reported `react(memo-dependencies): Found missing memoization dependencies` at that array in my previous pass, and I filed it with the file's pre-existing warnings instead of checking which line it pointed at. That callback now lints clean; every remaining warning in the touched mobile files is present on `upstream/main` unchanged.

No hook-level test: mobile has no `renderHook` setup, and adding React Native testing infrastructure for one dependency array is more than this PR should carry. The resolution rule itself is covered in `thread-provider-account-switch.test.ts`, and the lint rule covers the wiring.

Scoped `@t3tools/mobile` typecheck and lint pass.

chledowski (commented):
## Independent Codex review

## Findings

- **P2 — Expire consumed account-switch confirmations.** [ChatView.tsx:2351](/Users/jakub/code/t3/t3code-main/apps/web/src/components/ChatView.tsx:2351), also [use-thread-composer-state.ts:132](/Users/jakub/code/t3/t3code-main/apps/mobile/src/state/use-thread-composer-state.ts:132). Confirm A→B and send, then switch B→A on another device. The first client retains its original confirmation, which becomes valid again when the account returns to A. Its next send switches to B and starts another fresh conversation without renewed confirmation. Clear confirmations once the switch lands or becomes stale; add an A→B→A regression test.

- **P3 — New reactor tests violate the no-polling rule.** [ProviderCommandReactor.test.ts:2849](/Users/jakub/code/t3/t3code-main/apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts:2849). The added tests repeatedly poll mock call counts through a ten-second deadline. `AGENTS.md` explicitly requires receipts and worker drains. Replace these waits with the relevant receipt or `harness.drain()`, then assert results.

## Checked

- Verified exact head `ef67334bc5cd7c4a8e58ea4eb31f16c863d71be5`; reviewed the full 25-file diff against available `upstream/main` (`d29c56a5c`), instructions, changed tests, and surrounding client/session code.
- Reproduced stale-confirmation selection using the actual mobile helper: thread accounts A→B→A produced composer accounts B→B→B.
- `git diff --check` passed; no files changed.
- Focused tests could not execute: sandbox denied Vite’s temporary-file writes, including with native config loading.
- GitHub access failed, so PR body, linked issues, and current remote base could not be verified.

---

AI-generated review by Codex using `gpt-6-astra` at `low` effort.
Requested model: `gpt-6-astra`.

chledowski:
## Disposition — `ef67334` → `705923c`

**P2, spent confirmations could revive — correct, fixed.** A→B→A was real: after the switch landed, the confirmation stayed dormant rather than being consumed, and became valid again if the thread returned to A, so the next send restarted it on B unasked. A confirmation is now spent the moment the thread stops sitting on the account it was confirmed against — the switch landing and another device moving the thread are the same event from the client's side. The rule is one shared function, `retainConfirmedAccountSwitch` in `client-runtime`, so web and mobile expire on identical terms; the A→B→A sequence is covered in `providerInstanceDisplay.test.ts`.

**P3, polling in the new tests — correct, fixed.** All three tests I added now resolve a `Deferred` from `startSessionEffect` and then `harness.drain()`, matching the receipt pattern already used elsewhere in the file. `git diff` on `waitFor` usage against `upstream/main` is now empty: this PR adds none. I did not convert the file's 52 pre-existing polls; that is a separate change.

Reactor, `ChatView.logic`, `providerInstanceDisplay`, and the mobile module suites pass (230 tests); scoped typechecks for `t3`, `@t3tools/web`, `@t3tools/mobile`, `@t3tools/client-runtime` and lint are clean, with no new dependency warnings on the two new effects.

chledowski (commented):
## Independent Codex review

## Findings

- **P1 — Stale queued messages silently reset the provider conversation.** [ProviderCommandReactor.ts:755](/Users/jakub/code/t3/t3code-main/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts:755) automatically sets `startFreshConversation` whenever continuation groups differ, without receiving confirmation from the client. Queue a mobile message on account A while offline, switch the thread to B on desktop, then reconnect mobile: the outbox submits its retained A selection, switching back to A with a fresh conversation without confirmation. Previously this incompatible switch was rejected. **Fix:** carry explicit switch consent and the expected source account through the turn command; reject mismatches or unconfirmed switches. Add a regression test covering offline outbox delivery after another device changes accounts.

## Checked

- Verified exact head `705923c4f76bc951e47cec19ae98059418ac82f4`; inspected all 25 changed files against `upstream/main` at `d29c56a5c`.
- Read repository instructions, surrounding session lifecycle code, client confirmation logic, mobile outbox delivery, contracts, documentation, and added tests.
- `git diff --check upstream/main...HEAD` passed.
- Focused tests attempted; sandbox filesystem restrictions prevented execution.
- GitHub access failed, so PR body and linked issue context could not be verified. No edits or GitHub posts made.

---

AI-generated review by Codex using `gpt-6-astra` at `low` effort.
Requested model: `gpt-6-astra`.

chledowski:
## Disposition — `705923c` → `9301cfc`

**P1, stale queued messages silently reset the provider conversation — correct, fixed, and the fix is structural.**

The finding is right and it is the most important one in this review. Every client-side guard I added so far constrains a decision made *at pick time*; none of them reaches a command that was written earlier and delivered later. The mobile outbox is the sharpest case — queue a message offline on account A, move the thread to B from the desktop, reconnect — but it generalises: the server was inferring consent from a comparison the client may have made against a stale view of the thread. Before this PR that mismatch produced a loud rejection; my change turned it into a silent fresh start. That is a regression this PR introduced.

**Consent now travels with the turn.** `thread.turn.start` and its event payload carry `providerAccountSwitchFrom` — the account the client confirmed leaving. The reactor starts a fresh conversation only while that still names the account owning the current one; anything else falls through to `ProviderService.startSession`'s existing rejection. Both clients attach it through one shared rule (`resolveTurnAccountSwitchConsent`), and mobile persists it on the queued message so it survives the offline round trip. The field is optional everywhere, so persisted events and messages queued before this change decode as carrying no consent — the safe reading. That also improves version skew: an old client's turns carry no consent, so a new server rejects a move from one exactly as it does today.

Regression test added — `leaves the conversation alone for a turn that carries no consent` — with the binding on the account another device moved to and the turn naming the old one. The two switch tests now dispatch consent explicitly; they fail without it, which is the gate working.

Verification: typechecks for `t3`, `@t3tools/web`, `@t3tools/mobile`, `@t3tools/client-runtime`, `@t3tools/contracts`; reactor, decider (14 files), outbox, `ChatView.logic`, `providerInstanceDisplay`, and mobile module suites — 469 tests, all passing. Lint clean; the one warning in `use-thread-outbox-drain.ts` is present on `upstream/main`, shifted three lines by this diff.

chledowski (commented):
## Independent Codex review

## Findings

- **P2 — Plan follow-ups omit newly confirmed account consent** — `apps/web/src/components/ChatView.tsx:7573`. `onSubmitPlanFollowUp` reads `confirmedAccountSwitch` but omits it from its dependency array. Confirming another account can leave this callback holding `null`, while `getSendContext()` returns the new account. Implementing the plan then fails the server’s compatibility check. **Fix:** include `confirmedAccountSwitch` in the dependencies and test confirming a switch followed by plan implementation.

- **P2 — Failed account starts leave retries without usable consent** — `apps/web/src/components/ChatView.tsx:2361` (also mobile’s equivalent). For a reaped session on A, switching to B publishes a `starting` session naming B before the adapter succeeds. This clears the confirmation. If B fails to start, the error session still names B while the durable binding remains A; subsequent sends omit A’s consent and are rejected. Selecting B again produces no confirmation because the client already considers B current. **Fix:** retain the actual conversation owner until startup succeeds, or restore it on failure. Add a failed-start-and-retry test.

## Checked

- Exact head `9301cfc37fab9dfb970f5b885122f402bd394308`; reviewed changes against `upstream/main` at `d29c56a5c` (head’s parent). Local `main` is behind.
- Repository instructions, changed code/tests, surrounding session lifecycle, web/desktop selection, mobile outbox, contracts, and docs.
- `git diff --check` passed; no files changed.
- Five targeted test suites attempted; all blocked before execution by read-only filesystem restrictions.
- GitHub CLI and web access failed, so PR body and linked issue context remain unverified.

---

AI-generated review by Codex using `gpt-6-astra` at `low` effort.
Requested model: `gpt-6-astra`.

chledowski:
## Disposition — `9301cfc` → `980dcbe`

**P2, plan follow-ups omit newly confirmed consent — correct, fixed.** `onSubmitPlanFollowUp` is the one send path wrapped in a `useCallback`, and its dependency array omitted `confirmedAccountSwitch`. Dependency added. (The ordinary composer send is a plain function, so it reads the current render's value; it was never affected.)

**P2, failed account starts leave retries without usable consent — correct, fixed at the source.** Traced it: for a thread with no live runtime session, the optimistic `starting` session is published with `providerInstanceId: desiredInstanceId` before the adapter runs, and a failed start leaves that session behind as the error. Every client then reads the thread as already on the target, the confirmation is spent, and the retry carries no consent for the account that still owns the conversation — which the server then refuses.

Rather than compensating on the clients, the session now names **where the conversation is, not where it is going**: `activeSession?.providerInstanceId ?? boundInstanceId ?? desiredInstanceId`. A start that fails leaves the thread pointed at the account that still owns its conversation, which is simply true, and the confirmation stays valid for the retry. The binding read moved above the optimistic publish; it is the same read the switch decision and `startSession` already use.

Regression test added — `keeps a failed account switch pointed at the account that still owns it` — idle thread bound to one account, a consented switch whose start fails, asserting the thread still names the owner. It fails without the change.

Note for whoever reads this next: the recovery path was not fully closed before this fix — the user could pick the original account and confirm a prompt whose wording did not match what happened. The fix removes the confusing loop rather than papering over it.

Verification: five package typechecks clean; reactor, importer, provider service, `ChatView.logic`, `providerInstanceDisplay`, mobile module, and outbox suites — 373 tests, all passing. Lint and format clean.

chledowski (commented):
## Independent Codex review

## Findings

- **P3 — Replace polling in the new failure-path test.** `apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts:3055` repeatedly queries the projection using `waitFor` with a 10-second deadline. Slow processing can fail otherwise correct behavior, and this violates `AGENTS.md`’s explicit requirement to await receipts and worker drains. Await the failure receipt and drain the reactor before asserting the retained account.

## Checked

- Verified head `980dcbe8fa8c8a4e5a595532e21ca0b4714ed60a`; local `upstream/main` matches its parent, `d29c56a5c`.
- Read the complete 29-file diff, applicable instructions, affected tests, and surrounding account selection, consent, outbox, and provider restart code.
- No additional correctness defects identified. `git diff --check` passed.
- Focused tests were blocked before execution by read-only filesystem restrictions, including with the native config loader.
- GitHub access failed; PR body, linked issues, and live base verification remain unchecked. No edits or GitHub writes performed.

---

AI-generated review by Codex using `gpt-6-astra` at `low` effort.
Requested model: `gpt-6-astra`.

chledowski:
## Disposition — `980dcbe` → `b2e65ba`

**P3, polling in the new failure-path test — correct, fixed.** I copied the neighbouring failure test's `waitFor` on the projection, which contradicts the claim I made two rounds ago that this PR adds no polling. The test now resolves a `Deferred` from `startSessionEffect` before the failure propagates, drains the reactor, and asserts on the projection — the same shape already used for the workspace-missing failure in this file. It also asserts the session status explicitly, which the polling loop had been standing in for.

`git diff` of `waitFor` usage against `upstream/main` is empty again, by line and by count.

No other findings this round. Reactor suite green (67 tests); `t3` typecheck clean.

chledowski (commented):
## Independent Codex review

## Findings

- **P2 — Mobile falls back to the failed switch target instead of the owning account.** [thread-provider-account-switch.ts:91](/Users/jakub/code/t3/t3code-main/apps/mobile/src/features/threads/thread-provider-account-switch.ts:91). The outbox persists the target model selection before starting the session. If switching A→B fails, the server correctly retains A as session owner, but `threadSelection` still names B. After relaunch clears confirmation, this fallback selects B without consent, so subsequent sends fail again. Resolve the fallback against `lockedInstanceId`, including when no draft exists, and add a regression test for failed switch → relaunch → send.

## Checked

- Exact head `b2e65baace3fa8121ca0e9916a6759e69a3e1b39` against local `upstream/main` (`d29c56a5c`); local `main` is stale.
- Reviewed the 29-file diff, applicable instructions, affected tests, confirmation flow, offline queue, contracts, and provider session handling.
- Reproduced the finding by executing the actual selection helper. `git diff --check` passed.
- Focused tests could not execute: read-only sandbox blocked Vite temporary files.
- GitHub access failed; PR body and linked issue context could not be verified. No edits or GitHub posts.

---

AI-generated review by Codex using `gpt-6-astra` at `low` effort.
Requested model: `gpt-6-astra`.

chledowski (commented):
## Independent Claude review

## Findings

**1. Low — PR description contradicts the shipped behavior for a deleted owning account.**
`apps/server/src/orchestration/Layers/ProviderCommandReactor.ts:633-646` and the test `does not automatically replace a deleted account` at `apps/server/src/orchestration/Layers/ProviderCommandReactor.test.ts:3012`. The PR body says that when the account owning the conversation was deleted "the thread starts fresh on the picked account rather than staying stuck." The code does the opposite: `getInstanceInfo(currentInstanceId)` on the binding's instance fails and the turn is rejected with "references unknown provider instance", and the test asserts exactly that. Failing case: user deletes the Claude instance a thread ran on, picks another Claude instance, sends. Result is a permanent error on every send with no user path out, because the client cannot build a confirmation for an instance that is no longer in the catalog. Not a regression against main, but reviewers and users reading the PR body will expect the other behavior. Fix: either update the PR body to say the deleted-owner case is deliberately left rejected, or implement it (treat a binding whose instance is not in the registry as ownerless, i.e. `hasConversation` without a resolvable `currentInfo`, and start fresh on the requested instance with `startFreshConversation: true`, plus a test that flips the current assertion).

**2. Low — rejection text claims the account changed when it may never have.**
`apps/server/src/orchestration/Layers/ProviderCommandReactor.ts:706-718`. The same branch fires for three distinct cases: revision mismatch after a real move, a turn from an old client that never carries consent, and a session-less thread with a binding (imported) where the client's lock is derived from `modelSelection` and it sends no consent. The message "The thread's provider account changed. Select the target account again to confirm…" is only true for the first. Fix: split the wording: when `providerAccountSwitchFrom` is undefined say the move needs confirmation from a client that supports it; keep the "account changed" text for the revision/owner mismatch.

No correctness, security, data-loss, or performance defects found in the diff itself.

## Checked

- Head is `d969eec0c`; diff taken against its parent `d29c56a5c` (the local `origin/main` ref is stale, so the merge base diff was not used).
- Server: reactor consent gate reads the binding first, then the live session, then the projected session, then `modelSelection`; `startFreshConversation` is only set when `switchesProviderAccount` is true, which requires a matching `providerAccountSwitchFrom`, matching revision, and a projected session that agrees with the binding. Compaction is refused while a switch is pending. The "starting" session now keeps the owning instance so a failed switch leaves consent valid, and `ProviderService` writes the new binding only after the adapter start succeeds, so a failed switch cannot leave binding and session pointing at different accounts.
- `ProviderService.startSession`: incompatible-instance switches are now rejected even with no cursor unless `startFreshConversation` is set, and same-continuation-group idle switches now resume the binding cursor (previously started fresh). Both are covered by new tests and are intentional; the second is a behavior change the PR body does not mention.
- Revision derivation: `projectSessionAccountRevision` is shared by the pure projector, the SQL projection pipeline, and the client reducer, all keyed on event `sequence`. Migration 051 backfills only up to the projection cursor and mirrors the projector's null-handling; the test exercises round trips, missing accounts, recreated threads, and a pending event past the cursor. All five snapshot SQL sites now select `provider_account_revision`, and the one hand-built session row was replaced with `mapSessionRow`.
- Wire: `ws.ts` spreads the client command into the orchestration command, so the two new optional fields reach the decider and event payload. All additions are optional, so old events and queued messages decode as no consent.
- Web: consent is in-memory only, scoped to environment and thread, dropped by `retainConfirmedAccountSwitch` when the owner or revision moves, re-validated against the live shell atom after the dialog resolves, and included at both turn-start call sites. Picker selectability is gated on the `threadProviderAccountSwitch` capability. Compaction is disabled with a reason while a switch is pending.
- Mobile: same helpers from `client-runtime`; the picker offers only accounts the prompt would confirm; the outbox persists `providerAccountSwitchFrom` and the revision so offline sends carry the same consent; the draft's model selection is not treated as consent. Docs for Claude and Codex updated; the Codex page correctly states mobile still does not offer same-home switches.
- Tests run locally, all passing: `ProviderCommandReactor`, `ProjectionPipeline`, `ProviderService`, migration 051, `ChatView.logic`, `providerInstanceDisplay`, `threadReducer`, mobile `thread-provider-account-switch`, `thread-outbox`, and `use-composer-command-menu` (460 tests across 10 files). No repo-wide checks were run.


---

AI-generated review by Claude using `claude-fable-5-1` at `medium` effort.
Requested model: `fable`.
Configured fallback: `opus`. Fallback used: no.

chledowski:
Disposition for the account-switch audit and existing review threads, updated through final head 0a2153305:

- **Missing-session confirmation / persisted draft ownership** ([3976020131](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976020131), [3976095352](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976095352)): fixed. Both clients derive the owner from the session or persisted selection; a saved target selection does not constitute consent. The mobile resolver now preserves the owning account after stale confirmation or failed startup.
- **Mobile picker reachability** ([3976020130](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976020130)): fixed. It offers the owning account and supported same-provider accounts that the shared prompt can confirm.
- **Absent dialog host** ([3976020134](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976020134), [3976285375](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976285375)): fixed. Only an explicit `true` applies the web selection.
- **Callback dependencies** ([3976285359](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976285359), [3976529742](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976529742)): fixed. Mobile send and web plan follow-up depend on the current confirmation and carry its source/revision to the turn.
- **Late dialog callbacks and offline A→B→A** ([3976390272](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976390272), [3976529720](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976529720), [3976529749](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976529749)): fixed with durable account revisions. Both callbacks re-read the live shell and check the route, owner, and revision. Consent is checked again by the reactor and persists in the mobile outbox. Migration 051 backfills existing projected history; snapshot and replay paths use the same revision semantics.
- **Idle cross-driver guard** ([3976095357](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976095357), [3976285367](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976285367), [3976529736](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976529736)): fixed. The durable binding is authoritative even after metadata changes and runtime loss; a focused regression asserts rejection before startup.
- **No-consent rejection / missing cursor** ([3976529729](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976529729), [3976609416](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976609416)): fixed. The reactor now explicitly rejects missing/wrong consent independently of cursor presence. Tests assert zero startup/send calls and an unchanged binding; service tests cover request, persisted, and absent cursors. The old test's claimed “no startup call would pass” case was overstated because its awaited receipt came from that mock, but its lack of meaningful rejection assertions was real and has been removed.
- **Deleted-owner recovery** ([3976020133](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976020133)): deliberately deferred. Automatic recovery was removed from this manual-switch scope. Unknown owners produce an explicit error and keep their binding, so restoring the configuration remains possible. The PR body now matches this behavior.
- **Missing continuation metadata fallback** ([3976095363](https://github.com/pingdotgg/t3code/pull/11047#discussion_r3976095363)): deferred as inherited selection behavior. The exact-instance fallback condition is unchanged from the base; built-in configured drivers publish continuation identities. This PR's account-switch helper refuses unknown continuation groups, and the server rejects incompatible moves without matching consent. Changing the broader legacy selection fallback is separate from manual switching.

Additional audit fixes: compaction must wait for an ordinary message after switching; a compatible idle-instance move retains the persisted cursor; shared-home Codex support is qualified for web/desktop in the docs. Repeated comments and reactor test setup were shortened while retaining both clients and the new revision/migration safeguards.

The independent `claude-fable-5-1` / medium review is complete, with no fallback. It found no correctness, security, data-loss, or performance defects and ran 460 focused tests across ten files, all passing on d969eec0c. Its two low-priority wording findings are resolved: the PR body no longer promises deleted-owner recovery, and a3b681705 uses neutral confirmation-required wording. That follow-up changes one error string; scoped lint, formatting, and diff checks passed.

Upstream CI and preview workflows are still `action_required`. Attempting to approve CI returned HTTP 403 (“Must have admin rights to Repository”); maintainer authorization is required. Final client evidence is recorded below.


Live verification found one additional correctness defect after the consolidated review: target lifecycle events changed the projected owner before startup succeeded, despite an unchanged directory binding. Fixed in f66e380a1; bounded independent review identified runtime-error and late-turn-finalization gaps, fixed in 0a2153305. Both session-update paths now gate ownership by the projected account. Failure recovery prefers the durable binding. Error activities and buffered old-turn text still finalize, and late old-account exits do not clear the new session’s state.

Validation for those follow-ups: 74 command-reactor tests, 66 ingestion tests, and 6 ingestion companion tests pass. Final server typecheck, scoped lint/formatting, and diff checks pass. Existing warnings remain. The earlier Fable review’s 460-test result is evidence for d969eec0c, not a claim that it reviewed these later event changes.

The final head 0a2153305 was run in an isolated development environment with two configured Codex homes. An ordinary confirmed switch completed successfully. An injected startup exit preserved both projected and durable ownership at revision 113; removing the failure and retrying succeeded without reconfirming. Earlier d969eec0c browser verification also covered retained transcript, disabled pending-switch compaction, and a two-client A→B→A while a confirmation dialog stayed open; the old dialog was discarded. The historical Claude recording remains labelled as historical.

Native verification on final head: iOS Simulator build succeeds with local ad-hoc signing; the app installs and pairs to the disposable server, displays the existing transcript, and exposes both Codex account groups in its chooser. Native confirmation/send remains unverified: XcodeBuildMCP 2.6.2 returns each model row as `other` with action `none` (the source already declares labelled radio controls), so the supported semantic automation cannot select it. A manual native confirmation/send check remains a deployment gate. The production web bundle also builds successfully. No source changes were made to work around simulator/tooling prerequisites.

Scope/size: repeated explanations and reactor setup were shortened; added comment-only lines fell from 225 to 92. The final diff nevertheless grew from 1,481 to 2,237 changed lines (47 files) because durable revision persistence, migration/replay parity, and meaningful failure/late-event regressions were needed. Both clients remain; automatic deleted-owner recovery is deferred. All 17 existing review threads are resolved.


Final-head captures: [web failure/retry](https://raw.githubusercontent.com/chledowski/t3code/d03ed2827517a282d338aa6fcc74d897f6a42575/0a2153305-web-retry.png), [native chooser](https://raw.githubusercontent.com/chledowski/t3code/d03ed2827517a282d338aa6fcc74d897f6a42575/0a2153305-mobile-picker.jpg). The native capture documents visible accounts, not a completed native switch.


### Final-commit acceptance pass

A second deliberate acceptance pass ran on unchanged head `0a2153305`: 47 focused behavior tests, 3 reconnect/merge tests, and 215 tests in affected mobile/web client files passed (265 executions). No source edits or new correctness failures resulted. This is a finite behavior matrix, not a Cartesian product of every provider/device/state combination. Prior scoped typecheck/build results remain applicable to this exact head.

Fresh real-web evidence covers cancel/confirm, transcript retention, pending-switch compaction, startup failure and retry without reconfirmation, and two-client A→B→A with an old dialog open. Failure retained both owners, revision 162 and the exact resume cursor; retry succeeded. After the round trip, revision 203 and source selection remained current when the old dialog was accepted.

A suspected legacy-cache gap was ruled out: a new RPC session refreshes the authoritative shell even at the same sequence, or requests a full WS snapshot when HTTP fails. Both clients take consent revision from the shell. Three focused reconnect/fallback/merge tests pass. Fresh=true with an explicit foreign cursor and some late-cleanup/archived-session combinations remain source/shared-branch coverage rather than dedicated selected assertions; no claim is made that every such combination was exercised.

The native app was relaunched and the chooser reached again; model rows still have no semantic tap action. Native confirmation/send remains unverified. Upstream CI and four preview/fingerprint workflows remain `action_required`. No merge or deployment has occurred.

Final-commit GIFs (captured UI states with idle time condensed):

- [Successful switch](https://raw.githubusercontent.com/chledowski/t3code/ad8871367ed364abc045807e92917076c92406da/0a2153305-switch.gif)
- [Failure and retry](https://raw.githubusercontent.com/chledowski/t3code/ad8871367ed364abc045807e92917076c92406da/0a2153305-retry.gif)
- [Two-client stale consent](https://raw.githubusercontent.com/chledowski/t3code/ad8871367ed364abc045807e92917076c92406da/0a2153305-stale-consent.gif)


## Simplified revision and review closeout — 11 September 2026

Final head: `501f224cb3845e5fdfe9f6428679697e0ee207b0`, one commit on upstream `57aee3e19`. The previous historical backfill is removed: the decider records a counter only on ownership changes, while legacy events and upgraded rows default to zero. The final patch has 138 fewer changed lines than the previous version (45 files, +1,972/−127).

Independent review [5175909734](https://github.com/pingdotgg/t3code/pull/11047#pullrequestreview-5175909734) used Claude Opus 5 at medium effort after configured Fable attempts hit quota/service-overload errors.

1. **Turns without explicit model selection — fixed.** The reactor now defaults the instance to the durable owner instead of the stale thread creation selection. Active and stopped-runtime regressions verify the switched account and revision are retained and all three turns are sent.
2. **Pre-bind terminal events and stale progress — narrowed and addressed.** The target cannot own a turn before binding: `sendTurn` follows the awaited owner projection. Startup/send failure already dispatches an error session that settles the pending start. Clearing all caches on an incoming failure would incorrectly erase source background work that may still be alive. The bounded fix clears source plan progress and background liveness only after the incompatible target successfully starts, immediately before binding. Tests assert cleanup after success, preservation on failed startup, and cleanup after retry. A bounded independent follow-up confirmed this placement. Existing late-source-event ownership guards remain intact.

**Validation:** the simplified base passed 630 focused tests, five scoped typechecks, lint/formatting, and production web/iOS simulator builds. After the final server fixes, 156 tests across the reactor, ingestion, and importer suites passed, plus server typecheck and scoped lint/formatting. These test counts overlap. The final commit hook completed and the checkout is clean.

**Live acceptance on `d9e11c2e3`:** web cancellation preserved the source; confirmation plus send produced a target reply and stored revision 1. Native iOS pairing, selection, confirmation, send, reply, and transcript retention passed; the return switch stored revision 2. Final server-only review fixes received the focused checks above. No claim is made that these are captures of the later commit.

[Web walkthrough](https://raw.githubusercontent.com/chledowski/t3code/a69e6e251c8fa73b6f40544b2d55c26e3f5b6cb1/d9e11c2e3-web-switch.gif) · [Native iOS walkthrough](https://raw.githubusercontent.com/chledowski/t3code/a69e6e251c8fa73b6f40544b2d55c26e3f5b6cb1/d9e11c2e3-ios-switch.gif)

All prior review-thread dispositions remain valid except the superseded backfill design. Upstream CI and the four preview/fingerprint workflows still require maintainer authorization; this contributor account cannot grant it. No merge or deployment has occurred.

chledowski (commented):
## Independent Claude review

## Findings

### 1. After a confirmed switch, any turn started without an explicit `modelSelection` is rejected — this breaks answering agent questions (severity: high)

`apps/server/src/orchestration/Layers/ProviderCommandReactor.ts:710` keeps `desiredModelSelection = requestedModelSelection ?? thread.modelSelection`, and `:763` now derives `switchesInstance` from that value for *every* turn, where the pre-diff code required `requestedModelSelection !== undefined`.

The server's `thread.modelSelection` is only written by `thread.created` and `thread.meta-updated` (`apps/server/src/orchestration/Layers/ProjectionPipeline.ts:617,819`); `thread.turn-start-requested` does not update it (`ProjectionPipeline.ts:1358`, unlike the client reducer at `packages/client-runtime/src/state/threadReducer.ts:341`), and neither web nor mobile sends `thread.meta.update` when the user picks an account. So a successful account switch leaves `thread.modelSelection.instanceId = A` while the binding/session is `B` — permanently.

Failing case:
1. Thread runs on account A (`claude_work`), user confirms a switch to B (`claude_personal`), sends a message. Session and binding are now B; `thread.modelSelection` is still A.
2. The agent asks a question; the user answers it. `apps/server/src/orchestration/decider.ts:1539-1553` dispatches `thread.turn.start` **with no `modelSelection`**.
3. `ensureSessionForThread` computes `desiredInstanceId = A`, `currentInstanceId = B` (binding), so `switchesInstance` is true and the continuation keys differ → `switchesProviderAccount` (`:771`) → no consent present → the turn fails with *"Switching provider accounts requires current confirmation. Select the target account again to confirm starting a fresh conversation."*

This is not self-healing: `thread.modelSelection` never changes, so every user-input answer on that thread keeps failing (and if consent were supplied, it would move the thread *back* to A, which is also wrong). The same expression makes a no-selection compaction on a switched thread fail with the unrelated "Send a message to switch accounts before compacting".

Fix: when the caller did not request a selection, the thread is not moving. Either gate the switch detection on an explicit request (`const switchesInstance = hasConversation && requestedModelSelection !== undefined && desiredInstanceId !== currentInstanceId;`) or, better, resolve the desired instance from the owner: `const desiredModelSelection = requestedModelSelection ?? { ...thread.modelSelection, instanceId: currentInstanceId }`. Add a reactor test that starts a turn with no `modelSelection` after a committed switch (the added `manual account switching` suite always passes one).

### 2. Terminal events from the incoming account are discarded in the window before ownership is projected (severity: medium, plausible)

`apps/server/src/orchestration/Layers/ProviderRuntimeIngestion.ts:1490` gates lifecycle application, `session.exited` turn-state clearing (`:1919`), `runtime.error` session updates (`:1925`), plan-progress clearing (`:2018`) and liveness clearing (`:2060`) on `thread.session.providerInstanceId === event.providerInstanceId`. That ownership is read from the SQL projection (`resolveThreadRuntimeContext`, `:969`), and the reactor only commits the new account after `providerService.startSession` resolves and the `thread.session-set` event has been projected (`ProviderCommandReactor.ts:848-872`).

Failing case: the user switches to an account whose credentials are broken — a common reason to switch. `startSession` succeeds, then the target CLI exits or emits `runtime.error` while the projection still names the source account. Those events are dropped permanently (nothing replays them), so `clearTurnStateForSession`, plan-progress and background-liveness clearing never run for that turn; the thread keeps a stale working/plan indicator. The added ingestion test asserts exactly this drop for pre-bind target events, so the behavior is by design, but the design has no compensating path.

Fix: treat a not-yet-committed instance as matching when the thread has a start in flight (e.g. compare against the instance the reactor is starting, tracked alongside `threadModelSelections`), or, cheaper, always run the unconditional cleanups (`clearTurnStateForSession`, plan progress, liveness) on `session.exited` when the exiting instance is the one that owns the in-flight start.

## Checked

- Full single-commit diff (`d9e11c2e3`, 45 files) against `main`; the `main...` three-dot range is polluted by an old merge base, so review used `git show`.
- Server: consent validation and driver lock in `ProviderCommandReactor.ensureSessionForThread`, the relocated `pendingTurnStart` session write (owner preserved on failed startup), `startFreshConversation` handling and cursor suppression in `ProviderService.startSession` (only caller is the reactor), revision stamping in `decider.ts` (incoming revision stripped, bumped only on durable owner change), `ProjectionPipeline`/`ProjectionThreadSessions`/`ProjectionSnapshotQuery` column plumbing (all five session SELECTs updated, `mapSessionRow` reuse is shape-identical), migration 051 (`NOT NULL DEFAULT 0`, no backfill needed since legacy events carry no revision).
- Capability gating (`threadProviderAccountSwitch`) and contract additions; older clients simply never send consent and are rejected as before.
- Clients: web confirmation flow, re-validation of consent after the async dialog, `retainConfirmedAccountSwitch` revision expiry, compaction disable on all shared web entry points, picker selectability; mobile `resolveComposerModelSelection`/`resolveThreadAccountLock`, Alert confirmation, slash-command disable, outbox schema (legacy entries carry no revision → server rejects, verified against the consent check).
- Verified the compaction-replay path forwards `providerAccountSwitchFrom/Revision` via `...request`, and that no other server path dispatches `thread.turn.start` except the user-input answer path (finding 1) and that replay.
- Reused caller evidence as-is: 630 targeted tests green, five scoped `tsc --noEmit` projects, scoped lint and format — all exit 0 and covering the changed files. I ran no additional checks; the gap behind finding 1 is a missing test case, not a failing one, and reproducing it would require editing the tree, which this pass does not do.


---

AI-generated review by Claude using `claude-opus-5` at `medium` effort.
Requested model: `fable`.
Configured fallback: `opus`. Fallback used: yes.

chledowski:
Closing in favor of #11718, a much smaller version of this: one explicit server command, web only, no auto-switch and no transcript migration. Thanks for the earlier feedback on #9181; that shaped the narrower scope.

PR #11053, opened 2026-09-10
fix(server): advertise the tailnet editor target when Tailscale SSH serves without a local sshd

### LINKED ISSUE #11052: [Bug]: Open in editor reports no SSH route when using Tailscale SSH without a localhost listener
(this issue is brought to you by gpt-6-astra-medium)

### Area

apps/server (remote editor discovery), apps/web (editor menu)

### Steps to reproduce

1. Run a T3 server on a Linux host with Tailscale SSH enabled and no conventional SSH listener on localhost port 22.
2. Connect to that environment from another machine through a URL connection, without a desktop SSH connection profile supplying an explicit SSH alias.
3. Confirm that `ssh <tailnet-host>` works from the viewing machine.
4. Open the project's “Open in editor” menu.

### Expected behavior

Offer remote editor links using the working Tailscale SSH destination. A localhost SSH listener should not be required when Tailscale handles SSH connections.

### Actual behavior

The menu displays “No SSH route to <environment>” and disables opening the project in an editor, despite successful SSH access from the viewing machine.

### Verified diagnostics

Read-only checks from a macOS viewing machine against the affected Linux environment confirmed:

- A noninteractive SSH login succeeds.
- `tailscale status --json` reports `BackendState: Running` and a valid MagicDNS name.
- `tailscale debug prefs` reports `RunSSH: true`.
- TCP connections to both `127.0.0.1:22` and `[::1]:22` on the remote host fail with connection refused.
- `ss -ltn 'sport = :22'` reports no listener.

### Cause

[RemoteOpenTargets.ts](https://github.com/pingdotgg/t3code/blob/d29c56a5c404cb0f58d3b2ac41762fa0d0ac28d4/apps/server/src/environment/RemoteOpenTargets.ts#L34-L45) probes localhost port 22 and immediately returns an empty target list if neither loopback family accepts a connection. This happens before Tailscale discovery. The assumption that no local sshd means no SSH destination can work is invalid for Tailscale SSH.

[remoteOpen.ts](https://github.com/pingdotgg/t3code/blob/d29c56a5c404cb0f58d3b2ac41762fa0d0ac28d4/apps/web/src/remoteOpen.ts#L85-L96) then maps an empty advertised target list, without an explicit SSH profile alias, to `remote-unavailable`.

Discovery should recognize enabled Tailscale SSH independently of a localhost SSH listener. Tailscale connectivity alone should not be assumed to imply SSH availability.

### Impact

Remote opening in an editor is blocked for this valid Tailscale SSH configuration.

### Version or commit

Investigated source: d29c56a5c404cb0f58d3b2ac41762fa0d0ac28d4. The affected installed server version was not captured.

### Related issues

Related discovery/gating symptoms: #7530 and #8278. This report specifically concerns Tailscale SSH with no localhost SSH listener.

Mnigos:
"Open in editor" advertises the SSH hosts the environment can be reached on, but the server bails out with an empty list when nothing listens on localhost port 22. A host that runs Tailscale SSH has no local sshd at all, so the menu shows "No SSH route" even though `ssh <tailnet-name>` works from the viewer's machine.

This follows the split-gate direction from the triage on #11052. `tailscale status --json` does not report the local node's own SSH host keys (the `Self` entry never carries `sshHostKeys`), so enablement comes from the local prefs: `tailscale debug prefs` reports `RunSSH`, the switch `tailscale up --ssh` flips. `@t3tools/tailscale` gains `readTailscaleSshEnabled` on the same command runner as `readTailscaleStatus`, and `RemoteOpenTargets` consults it only when no loopback sshd answers. The tailnet name is advertised when either a local sshd or Tailscale SSH is serving; a running tailscaled alone (`--tailscale-serve`) still advertises nothing. The mDNS name still requires a local sshd, since Tailscale SSH only answers on the tailnet address. `readTailscaleStatus` also reports `running` from `BackendState`, and the tailnet name is only advertised while the daemon is Running: after `tailscale down` the status still carries the node name but nothing answers on it. The existing test that expects `[]` with Tailscale up but no SSH anywhere is unchanged.

## Verification

- `RemoteOpenTargets.test.ts`: the spawner now answers `status --json` and `debug prefs` separately. No loopback sshd plus `RunSSH: true` yields the tailnet target alone; `RunSSH: true` with tailscale down or with `BackendState: "Stopped"` yields nothing (a stopped daemon plus a local sshd still yields the mDNS name only); the existing no-sshd and ordering cases are unchanged. The prefs fixtures are the real `tailscale debug prefs` shape (trimmed) from tailscale 1.94.
- `tailscale.test.ts`: `readTailscaleSshEnabled` runs `tailscale debug prefs` and decodes `RunSSH`; missing or false reads as disabled, malformed JSON is a parse error.
- `vp test run` on both files (23 tests), plus typecheck and lint for `packages/tailscale`, `apps/server` and `apps/desktop` (the other `readTailscaleStatus` consumers).

Fixes #11052. Implemented with Claude Code (Claude Fable 5.1).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved remote target discovery when local SSH, Tailscale SSH, or mDNS availability changes.
  - Tailscale targets are now advertised only when the daemon is running, a MagicDNS name is available, and an appropriate SSH service is enabled.
  - Prevented stopped Tailscale devices from appearing as available targets.
  - mDNS targets are now shown only when the local SSH service is listening.
  - Improved handling of Tailscale SSH settings and daemon status.

### DIFF
--- a/apps/server/src/environment/RemoteOpenTargets.test.ts
+++ b/apps/server/src/environment/RemoteOpenTargets.test.ts
@@ -5,6 +5,7 @@ import * as Effect from "effect/Effect";
 import * as Layer from "effect/Layer";
 import * as Sink from "effect/Sink";
 import * as Stream from "effect/Stream";
+import * as ChildProcess from "effect/unstable/process/ChildProcess";
 import * as ChildProcessSpawner from "effect/unstable/process/ChildProcessSpawner";
 import { describe, expect } from "vite-plus/test";
 
@@ -16,27 +17,34 @@ const TAILSCALE_STATUS_JSON = JSON.stringify({
   Self: { DNSName: "bb-1.tail1234.ts.net.", TailscaleIPs: ["100.64.1.2"] },
 });
 
-/** Spawner whose `tailscale status --json` exits with the given output. */
-const spawnerLayer = (input: { readonly exitCode: number; readonly stdout: string }) =>
+interface CommandResult {
+  readonly exitCode: number;
+  readonly stdout: string;
+}
+
+/** Spawner answering `tailscale status --json` and `tailscale debug prefs` separately. */
+const spawnerLayer = (input: { readonly status: CommandResult; readonly prefs?: CommandResult }) =>
   Layer.succeed(
     ChildProcessSpawner.ChildProcessSpawner,
-    ChildProcessSpawner.make(() =>
-      Effect.succeed(
+    ChildProcessSpawner.make((command) => {
+      const args = ChildProcess.isStandardCommand(command) ? command.args : [];
+      const result = args[0] === "debug" ? (input.prefs ?? TAILSCALE_PREFS_NO_SSH) : input.status;
+      return Effect.succeed(
         ChildProcessSpawner.makeHandle({
           pid: ChildProcessSpawner.ProcessId(1),
-          exitCode: Effect.succeed(ChildProcessSpawner.ExitCode(input.exitCode)),
+          exitCode: Effect.succeed(ChildProcessSpawner.ExitCode(result.exitCode)),
           isRunning: Effect.succeed(false),
           kill: () => Effect.void,
           unref: Effect.succeed(Effect.void),
           stdin: Sink.drain,
-          stdout: Stream.make(encoder.encode(input.stdout)),
+          stdout: Stream.make(encoder.encode(result.stdout)),
           stderr: Stream.empty,
           all: Stream.empty,
           getInputFd: () => Sink.drain,
           getOutputFd: () => Stream.empty,
         }),
-      ),
-    ),
+      );
+    }),
   );
 
 const netLayer = (input: { readonly ipv4: boolean; readonly ipv6: boolean }) =>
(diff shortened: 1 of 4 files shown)


shivamhwp:

The new enablement check reads a field that `tailscale status --json` does not provide as written. `PeerStatus.SSH_HostKeys` has the JSON tag `sshHostKeys`, not `SSH_HostKeys`: https://github.com/tailscale/tailscale/blob/v1.94.2/ipn/ipnstate/ipnstate.go#L301 . The fixtures currently use the Go field name, so they hide this mismatch.

Changing the spelling alone is insufficient for the local node. `updateStatusLocked` fills SSH host keys for peer entries, but its `MutateSelfStatus` block does not populate them: https://github.com/tailscale/tailscale/blob/v1.94.2/ipn/ipnlocal/local.go#L1311 . The same split remains in current upstream. Please use a supported local SSH-enablement signal, such as the local RunSSH preference, and cover output from an actual enabled node with no loopback sshd. Keep the independent mDNS gate.

Mnigos:
@shivamhwp you're right on both counts, switched to RunSSH from `tailscale debug prefs` in `495593b`, mDNS gate unchanged.

PR #11059, opened 2026-09-10
fix(server): recognize authenticated GitHub Enterprise hosts

alimek:
## What Changed

Recognize custom GitHub Enterprise hosts listed in `gh auth status --json hosts`. This uses the existing unknown-remote discovery hook, preserves the repository's host, and lets the PR page list repositories previously recorded as `unknown`.

Matching is on host presence, not token state: `gh` lists a host whose token expired or was revoked, and claiming it lets gh's auth error surface as the actionable "GitHub CLI is not authenticated. Run `gh auth login` and retry" instead of the misleading unsupported-host message.

GitHub operations derive repository targets from the normalized remote URL, so stored repositories and CLI calls agree on `host/owner/repo` even when the cwd's default remote differs.

Added regression tests for custom-host discovery, case-insensitive matching, expired-token and unrelated accounts, unavailable auth JSON, repository targeting across PR/branch/checkout/clone operations, PR listing for a stored unknown repository, and link resolution. Added Enterprise sign-in guidance.

## Why

A repository on `code.example.test` can work with `gh pr list` while its PRs are absent in T3 Code. Hostname detection only recognizes GitHub when the hostname contains a complete `github` DNS label, and GitHub lacked the authenticated-host fallback already used by GitLab. Matching the remote host to any account the CLI reports closes that gap without new settings or client contracts.

## Validation

- 292 focused tests passed across source-control discovery, GitHub authentication, PR listing, and GitHub PR CLI behavior.
- Targeted lint and server typecheck passed.
- The new discovery tests failed before the fix and passed afterward.

## Checklist

Implemented with GPT-6 in the Codex harness; review feedback (presence-based host matching) applied with [Claude Code](https://claude.com/claude-code).


alimek:
@coderabbitai resume

alimek:
@coderabbitai review

xiaogwu:
I tested this against a real GitHub Enterprise host (`prodgit.<company>.com`, no `github` DNS label) and it fixes the misclassification — thanks for doing the `--repo host/owner/repo` threading too, that part matters once the stored repository and the cwd's default remote disagree.

One suggestion on the claim condition in `refineUnknownGitHubRemote`:

```ts
const authenticated = parseGitHubAuthStatus(input.auth.stdout).accounts.some(
  (account) => account.authenticated && account.host === host,
);
```

Consider claiming the host on *presence* in the hosts map, dropping `account.authenticated &&`. `gh auth status --json hosts` still lists a host whose token has expired or been revoked, just with `state != "success"`. With the `authenticated` requirement, that host refines to `null`, the kind stays `unknown`, and `PullRequestProviderRegistry.get("unknown")` is null — so the user gets:

> Change requests cannot be browsed for this project's host yet.

which is inaccurate. The host is supported; the credentials are stale. On presence alone the kind becomes `github`, `gh` fails with an auth error, and `PullRequestService`'s `providerDetail` maps it to `cli-unauthenticated`:

> GitHub CLI is not authenticated. Run `gh auth login` and retry.

That names the fix instead of implying an unsupported host, and it's the same shape as the existing `cli-missing` path. Both behave identically when credentials are good, so this only changes the expired-token case.

Not a blocker either way — the classification fix is the important half.

alimek:
Good catch — adopted in 78ac3b2. `refineUnknownGitHubRemote` now claims the host on presence in the auth-status hosts map alone; the `authenticated` requirement is gone. The expired-token case now resolves to `github`, gh's auth error classifies through `classifyNonZeroExit` as `authentication`, and users get the `gh auth login` hint instead of the unsupported-host message. Registry table scenario for a failed custom-host account now expects `github`.

PR #11063, opened 2026-09-10
feat(usage): mark the even-pace point on pooled limit bars

badmike:
Discussion: pingdotgg/t3code#11120 — filed after this PR, per CONTRIBUTING. Treat this branch as the reference implementation for that idea; happy to close it if the direction is not wanted.

## The problem

Usage → Limits tells you how much quota is left, but not whether that is a lot or a little for the time still to run. A weekly window at 40% left is comfortable on day six and alarming on day two, and the bar reads the same either way. You have to know the reset time, do the division in your head, and compare.

## The change

Each account segment now carries a tick above and below it at the point even spending would have reached. The distance between the tick and the fill edge is the headroom. The tick is drawn in the provider colour that fills the bar, and only when the window reports both a duration and a reset — without a timeframe there is nothing to compare against and nothing is drawn.

In the screenshots below, Claude's weekly window (40% left, 39% at even pace) sits right on the line, while Fable's (65% left, same clock) has a visible margin. Same numbers, different stories, and now you can see which is which without reading the countdown.

The segment popover carries the exact figures: `Even pace 39% · 1 pts of headroom`.

### Before

before

### After

after

### Narrow variant

The compact numbered strip gets the same treatment, scaled to the shorter bar.

after, narrow

## Notes

`evenPacePercent` and `paceGapLabel` live in `packages/shared/src/usageLimits.ts` so web and mobile agree on the maths, and the two views that already drew this marker inline now call them instead of repeating the arithmetic. Mobile gets the same ticks on its pool segments and the figures on the account detail screen.

No new dependencies, no contract changes, no docs change — the tick is a reading aid on an existing view, not a new feature to explain.

Verified with `vp test run packages/shared/src/usageLimits.test.ts apps/web/src/components/usage` (99 pass, four new cases cover the maths and the no-timeframe case), plus typecheck and lint on web, mobile, and shared.

Written by Claude Opus 5 driving T3 Code.

## Closes discussions

- https://github.com/pingdotgg/t3code/discussions/11120

## Summary by CodeRabbit

* **New Features**
  * Added even-pace indicators to usage-limit bars and pooled usage segments across web and mobile.
  * Usage details now show the expected even-pace percentage and whether usage has headroom, is on pace, or is ahead of pace.
  * Pooled usage details include an “Even pace” breakdown and clearer segment accessibility labels.
* **Bug Fixes**
  * Standardized even-pace positioning across usage-limit displays for more consistent progress tracking.

### DIFF
--- a/apps/mobile/src/features/usage/UsageLimitsPooled.tsx
+++ b/apps/mobile/src/features/usage/UsageLimitsPooled.tsx
@@ -1,12 +1,14 @@
 import { useAtomValue } from "@effect/atom-react";
 import { useNavigation, type StaticScreenProps } from "@react-navigation/native";
-import { EnvironmentId } from "@t3tools/contracts";
+import { EnvironmentId, type ServerProviderUsageWindow } from "@t3tools/contracts";
 import {
   collectLimitAccounts,
   collectLimitNotices,
   collectLimitPools,
+  evenPacePercent,
   formatDuration,
   formatResetsIn,
+  paceGapLabel,
   remainingPercent,
   type LimitAccount,
   type LimitPoolWindow,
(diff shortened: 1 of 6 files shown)


badmike:
Filed the Ideas discussion this should have started as: https://github.com/pingdotgg/t3code/discussions/11120

PR #11065, opened 2026-09-10
fix(server): run the background service through a Homebrew Node path that survives upgrades

### LINKED ISSUE #11054: [Bug]: macOS service pins Homebrew Node Cellar path, making Node upgrades unsafe
### Before submitting

### Area

apps/server

### Steps to reproduce

1. On an Apple Silicon Mac, install Node with Homebrew and invoke Node through Homebrew's stable path:

   ```sh
   command -v node
   node -p 'process.execPath'
   ```

   On my machine this prints:

   ```text
   /opt/homebrew/bin/node
   /opt/homebrew/Cellar/node/26.8.1/bin/node
   ```

   `/opt/homebrew/bin/node` is a symlink managed by Homebrew. `process.execPath` is the resolved, version-specific executable.

2. Install or update the T3 background service from that environment. I reproduced the generated configuration after:

   ```sh
   npx t3@0.0.41-nightly.20260909.1439 service update
   ```

3. Inspect the executable stored in the LaunchAgent:

   ```sh
   plutil -extract ProgramArguments.0 raw \
     "$HOME/Library/LaunchAgents/com.t3tools.t3code.service.plist"
   ```

4. The plist contains the resolved Cellar path:

   ```text
   /opt/homebrew/Cellar/node/26.8.1/bin/node
   ```

5. A normal Homebrew Node upgrade can repoint `/opt/homebrew/bin/node` and eventually remove the old keg. At that point the LaunchAgent still references the old version-specific path.

I did not intentionally remove the currently running Node keg just to force an outage on this host. The deterministic repro here is the generated service definition retaining a path whose lifetime is tied to one Homebrew keg.

### Expected behavior

A T3 background service installed with Homebrew-managed Node should remain startable across a routine Node upgrade, or the service lifecycle should otherwise make this dependency and the required maintenance explicit before the old executable disappears.

### Actual behavior

The generated LaunchAgent stores the resolved Cellar path. The service works while that keg exists, but a later restart can fail once Homebrew removes it, even though the normal `node` command continues to work through `/opt/homebrew/bin/node`.

There is a related runtime concern: the stable service launcher also starts managed T3 runtimes through its own `process.execPath`. If the launcher remains alive while Homebrew replaces and removes its Node keg, a later managed-runtime handoff may also try to use the removed executable.

### Impact

### Version or commit

Observed in `0.0.41-nightly.20260909.1439`; the relevant behavior is also present on `main` at `d29c56a5c404cb0f58d3b2ac41762fa0d0ac28d4`.

### Environment

macOS 26.6.2 (Apple Silicon), Homebrew Node 26.8.1, npm 11.19.0, T3 background service managed by launchd

### Logs or stack traces

```shell
$ ls -l /opt/homebrew/bin/node
/opt/homebrew/bin/node -> ../Cellar/node/26.8.1/bin/node

$ node -p 'process.execPath'
/opt/homebrew/Cellar/node/26.8.1/bin/node

$ plutil -extract ProgramArguments.0 raw \
    "$HOME/Library/LaunchAgents/com.t3tools.t3code.service.plist"
/opt/homebrew/Cellar/node/26.8.1/bin/node
```

### Investigation notes

I inspected the installed package and current source to understand whether this was only a local plist change.

- `bootService.ts` builds the service plan with `nodePath: host.execPath`, and the macOS renderer writes `plan.nodePath` into `ProgramArguments[0]`.
- `serviceLauncher.ts` starts a managed runtime with `spawn(process.execPath, ...)`.
- These two observations are consistent with the generated plist and process tree on my host.

Relevant source on the inspected commit:

- https://github.com/pingdotgg/t3code/blob/d29c56a5c404cb0f58d3b2ac41762fa0d0ac28d4/apps/server/src/cloud/bootService.ts#L570-L572
- https://github.com/pingdotgg/t3code/blob/d29c56a5c404cb0f58d3b2ac41762fa0d0ac28d4/apps/server/src/serviceLauncher.ts#L397-L408

I may be missing an intended assumption about how Node is provisioned for the background service. I am reporting the lifecycle mismatch rather than proposing a specific implementation. Preserving an upgrade-stable executable reference, resolving the current runtime at launch, or explicitly coordinating service maintenance with Node upgrades may have different trade-offs that are clearer to the maintainers.

### Workaround

I currently keep a locally managed LaunchAgent configuration and check it after `t3 service update`. This avoids relying on the generated Cellar path, but I do not consider the local configuration a recommendation for the upstream design.

Mnigos:
`t3 service install` / `service update` record `process.execPath` as the service's Node. Node reports the realpath, so a Homebrew install becomes `/opt/homebrew/Cellar/node/<version>/bin/node`, a path that `brew upgrade node` deletes. The running service survives on the old inode, but the next launchd or systemd start fails even though `node` still works through the Homebrew prefix. The launcher also spawns managed runtimes through `process.execPath`, so a live handoff after a keg removal breaks the same way.

Following the triage on #11054:

- The service plan now prefers the formula's `opt` link (`<prefix>/opt/<formula>/bin/node`) whenever `execPath` sits in a Homebrew keg and that link exists. It stays absolute, follows the current keg across upgrades, and covers versioned formulae such as `node@22` and Linuxbrew prefixes. Any other `execPath` is used as before. The same path goes into the unit's `PATH` entry and into the runtime validation run.
- The launcher spawns runtimes through the Node path it was started with (`process.argv0`) when that is absolute, and falls back to `process.execPath` otherwise.

`homebrewOwnershipFromCommandPath` from provider maintenance already parses keg layouts, so the keg detection is shared.

## Verification

- `bootService.test.ts`: keg to opt-link mapping (plain, versioned and Linuxbrew formulae, non-keg paths), and a macOS install that writes the keg path while the opt link is missing and the opt link once it exists, with no `Cellar` left in the plist.
- `serviceLauncher.test.ts`: absolute `argv0` wins, a bare `node` falls back to `execPath`.
- `vp test run` on both files (43 tests), server typecheck and lint on the touched files are clean.

Fixes #11054. Implemented with Claude Code (Claude Fable 5.1).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved service reliability after Node.js updates by using stable Homebrew Node paths.
  * Ensured launched runtimes continue using the configured service path instead of an outdated resolved path.
  * Improved macOS launch-agent setup when stable Node paths are available.

### DIFF
--- a/apps/server/src/cloud/bootService.test.ts
+++ b/apps/server/src/cloud/bootService.test.ts
@@ -181,13 +181,13 @@ const makeHarness = Effect.fn("test.make_boot_service_harness")(function* (
       };
     }),
   });
-  const makeService = (environmentPath = installerPath) =>
+  const makeService = (environmentPath = installerPath, execPath = "/usr/bin/node") =>
     BootService.make({
       baseDir,
       logsDir: path.join(baseDir, "userdata", "logs"),
       cliVersion: "1.2.3",
       host: {
-        execPath: "/usr/bin/node",
+        execPath,
         ...(usePinnedLauncher ? {} : { launcherSourcePath: sourceLauncher }),
       },
     }).pipe(
@@ -207,7 +207,21 @@ const makeHarness = Effect.fn("test.make_boot_service_harness")(function* (
       ),
     );
   const service = yield* makeService();
-  return { service, makeService, fs, statePath, commands, timeouts, control, runtime };
+  return { service, makeService, fs, home, statePath, commands, timeouts, control, runtime };
+});
+
+it("maps a Homebrew Node keg to the formula's stable opt link", () => {
+  expect(BootService.durableHomebrewNodePath("/opt/homebrew/Cellar/node/26.8.1/bin/node")).toBe(
+    "/opt/homebrew/opt/node/bin/node",
+  );
+  expect(BootService.durableHomebrewNodePath("/opt/homebrew/Cellar/node@22/22.1.0/bin/node")).toBe(
+    "/opt/homebrew/opt/node@22/bin/node",
+  );
+  expect(
+    BootService.durableHomebrewNodePath("/home/linuxbrew/.linuxbrew/Cellar/node/26.8.1/bin/node"),
+  ).toBe("/home/linuxbrew/.linuxbrew/opt/node/bin/node");
+  expect(BootService.durableHomebrewNodePath("/usr/bin/node")).toBeNull();
+  expect(BootService.durableHomebrewNodePath("/opt/homebrew/bin/node")).toBeNull();
 });
 
 it.layer(NodeServices.layer)("boot service install", (it) => {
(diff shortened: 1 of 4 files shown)


Mnigos:
> [!NOTE]

Closing: since #11510 the background service runs the self-contained `t3` executable from the release archive, so the host Node path this PR pinned is no longer used.

PR #11071, opened 2026-09-10
fix(server): reduce editor scan work and retain partial results

CodeZeno:
## What Changed

Reuse PATH directory listings during editor discovery to reduce repeated filesystem checks. Return and cache editors found within four seconds if the scan takes longer. Preserve filesystem case-sensitivity behaviour and add tests for directory caching, fallback checks, and partial results.

## Why

T3 Code on my Windows 11 machine would show an empty 'Open' list as it would timeout while looking for available editors. Even if some were found, they were thrown out if it reached the timeout.
The reason was 29,568 file checks being performed (22 commands (editors), 56 paths (my windows PATH), 12 file extensions and 2 case checks)

## UI Changes

Before:
before

After:
after

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Editor discovery now returns successfully detected editors even when another probe stalls or times out.
  - Completed discovery results remain available for subsequent requests.
  - Command lookup is more reliable on case-sensitive file systems and safely falls back when directory contents cannot be read.

- **Performance**
  - Repeated command lookups within the same operation reuse directory information, reducing unnecessary file-system checks while preserving PATH priority.


CodeZeno:
@coderabbitai help

CodeZeno:
@coderabbitai run pre-merge checks

PR #11075, opened 2026-09-10
fix(web): keep a file-to-symlink type change from crashing the diff view

### LINKED ISSUE #11074: [Bug]: CodeView crashes with duplicate id when a regular file becomes a symlink
## Before submitting

## Area

apps/web

## Steps to reproduce

1. Create a Git repository where `AGENTS.md` is a regular tracked file and `CLAUDE.md` is another regular tracked file:

   ```bash
   git init t3-codeview-type-change
   cd t3-codeview-type-change
   git config user.name "T3 repro"
   git config user.email "t3-repro@example.invalid"
   printf 'canonical instructions\n' > CLAUDE.md
   printf 'duplicated instructions\n' > AGENTS.md
   git add AGENTS.md CLAUDE.md
   git commit -m baseline
   ```

2. Open the repository in T3 Code and start a thread.
3. During a turn, replace the regular file with a relative symlink and commit it:

   ```bash
   rm AGENTS.md
   ln -s CLAUDE.md AGENTS.md
   git add AGENTS.md
   git commit -m 'use shared instructions'
   ```

4. After the turn checkpoint is created, open the turn diff or Changes view.

The checkpoint transition is a valid Git type change:

```text
mode change 100644 => 120000 AGENTS.md
```

With `git diff --no-renames`, Git represents it as two file blocks with the same path:

```text
diff --git a/AGENTS.md b/AGENTS.md
deleted file mode 100644
--- a/AGENTS.md
+++ /dev/null

diff --git a/AGENTS.md b/AGENTS.md
new file mode 120000
--- /dev/null
+++ b/AGENTS.md
```

This was also reproduced outside the original project with a minimal temporary repository.

## Expected behavior

CodeView should render the regular-file-to-symlink transition as one type-change item, or safely two patch blocks with distinct internal IDs. A valid Git diff should not crash the thread view.

## Actual behavior

The CodeView renderer throws a duplicate-ID error and T3 Code displays a full-page "Something went wrong" state:

```text
Error: CodeView.addItem: duplicate id "AGENTS.mdAGENTS.md"
    at appendItemsInternal (t3code://app/assets/diffFileContents-*.js)
    at setItems (t3code://app/assets/diffFileContents-*.js)
```

The repository and symlink remain valid. In a second occurrence on 2026-09-10, Try again, Reload app, and fully restarting the application did not recover the affected thread: its persisted open diff panel was restored and triggered the crash again. The original conversation was recovered by closing only its persisted right panel, as described below.

The path-only identity collision has since been confirmed by a maintainer in [this comment](https://github.com/pingdotgg/t3code/issues/11074#issuecomment-5618481677). The actual key contains a NUL separator: `AGENTS.md\0AGENTS.md`, which is invisible in the displayed error.

Related but distinct:

- #1434 concerns edits being assigned to the wrong turn. This report is a deterministic renderer crash on a valid type-change patch.
- #8408 concerns browsing symlinked directories, not rendering a file type change.

## Impact

The affected turn cannot be reviewed in CodeView. If its diff panel was open, the entire existing thread can remain inaccessible across application restarts. Creating a new thread is not necessary to recover the original conversation.

## Version or commit

T3 Code desktop 0.0.40

## Environment

- macOS 26.6.1, Apple Silicon arm64
- T3 Code desktop 0.0.40
- Codex CLI 0.154.0
- Node.js 26.8.1

## Logs or stack traces

```text
T3 Code (Alpha) 0.0.40
Error: CodeView.addItem: duplicate id "AGENTS.mdAGENTS.md"
    at appendItemsInternal (t3code://app/assets/diffFileContents-*.js)
    at setItems (t3code://app/assets/diffFileContents-*.js)
```

## Verified recovery of the original thread (desktop 0.0.40)

A second occurrence happened after a turn completed using **Claude Opus 5**, so the observed failure is not limited to the Codex provider. This is an additional observation, not proof of provider causality.

Inspection of the installed 0.0.40 source maps and the live application showed:

- `rightPanelStore.ts` persists panel state in localStorage under `t3code:right-panel-state:v2` (persisted schema version 11).
- The affected entry in `state.byThreadKey` had `isOpen: true`, `surfaces: [{ id: "diff", kind: "diff" }]`, and `activeSurfaceId: "diff"`.
- Changing **only that entry's `isOpen` to `false`**, then reloading, restored the same thread. The conversation, latest completed response, and message composer were visibly accessible again.
- No conversation database, checkpoint, repository file, or symlink was modified for this recovery.

For another user stuck on the error page in **0.0.40**, open **View > Toggle Developer Tools > Console**. This version-specific workaround identifies the current thread from its URL, saves a sessionStorage backup, closes only that thread's panel, and reloads:

```js
(() => {
  const storageKey = "t3code:right-panel-state:v2";
  const raw = localStorage.getItem(storageKey);
  const data = JSON.parse(raw ?? "null");
  const threadId = location.hash.split("?")[0].split("/").filter(Boolean).at(-1);
  const panels = data?.state?.byThreadKey;
  const matches = Object.keys(panels ?? {}).filter(
    key => key.endsWith(":" + threadId)
  );
  if (!threadId || matches.length !== 1) {
    throw new Error("Expected exactly one current-thread panel; nothing changed.");
  }
  sessionStorage.setItem("t3-recovery-right-panel-backup", raw);
  panels[matches[0]].isOpen = false;
  localStorage.setItem(storageKey, JSON.stringify(data));
  location.reload();
})();
```

The recovery operation was verified on the affected thread using its explicit scoped key. The snippet above generalizes key selection to the current route. Avoid reopening the affected diff until the renderer fix is available. Clearing all localStorage or deleting the thread is unnecessary.

## Additional fix and regression suggestions

The maintainer's proposed unique identity fix addresses the collision. These additional cases would cover the persistent lockout and prevent partial fixes:

1. **Preserve both patch halves.** Do not deduplicate by path and silently discard the deletion or symlink addition. If rendering two items, assign deterministic distinct IDs and use them consistently for collapse, selection/reveal, and annotation behavior. If merging, retain the old/new modes and contents.
2. **Contain rendering failures within the diff panel.** Keep the conversation and composer mounted, with a local error fallback offering to close the panel. The close action should persist `isOpen: false`, so restarting cannot restore the same failing panel indefinitely.
3. **Regression coverage:** regular file to symlink and the reverse; both patch halves and contents preserved; unique stable item IDs; collapse/reveal behavior after rerender; and reopening a thread with a persisted open diff panel. A deliberately failing diff renderer should leave the conversation usable and allow persistent panel closure.

The persisted-panel recovery is directly verified. The implementation and regression items above are suggestions, not a tested code patch.

Mnigos:
A regular file that becomes a symlink (or the reverse) is a Git type change. With `--no-renames` the patch carries it as two blocks with the same path: a deletion of the regular file and an addition of the symlink. The diff panel keys each block by previous and current path only, so both blocks got the identity `AGENTS.md\0AGENTS.md`, CodeView threw `duplicate id`, and the whole thread view fell into the "Something went wrong" state.

The identity key now includes the block's change type, so the deleted and the new block of a type change get distinct ids while a file that is merely edited keeps its key across renders. The pull request tab keys by render key and was not affected.

The file tree next to the diff had the same failure one layer down: `diffFileTreeEntries` produced two entries for the path and `model.resetPaths` threw `Duplicate path`, which also crashed the thread and kept crashing on reload while the tree preference was on. Same-path blocks now fold into one tree entry shown as modified, while the diff itself keeps both blocks.

## Verification

- `diffRendering.test.ts` parses the exact two-block patch from #11074 and asserts the two identities differ and every file in the patch has a unique key; the existing stability test still passes.
- `diffFileTree.logic.test.ts` folds the deletion plus addition into one modified entry; `DiffFileTree.test.tsx` mounts the real tree model with a type change, which reproduced the `Duplicate path` throw before the fold, and activates the surviving row.
- `vp test run` for `apps/web/src/lib/diffRendering.test.ts` and `apps/web/src/components/diffs` (37 tests), web typecheck and lint on the touched files are clean.

Fixes #11074. Implemented with Claude Code (Claude Fable 5.1).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved diff rendering for file type changes, such as a regular file becoming a symlink.
  - Entries sharing a path are now displayed as a single modified item when appropriate, preventing duplicate rows in the diff tree.
  - File changes now maintain distinct identities, avoiding collisions when deleted and newly added entries share the same path.
  - Activating affected entries reveals the correct file target once, improving navigation and preventing duplicate reveals.

### DIFF
--- a/apps/web/src/components/diffs/DiffFileTree.test.tsx
+++ b/apps/web/src/components/diffs/DiffFileTree.test.tsx
@@ -5,7 +5,10 @@ import { act, type MouseEvent, type ReactNode } from "react";
 import { create, type ReactTestRenderer } from "react-test-renderer";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vite-plus/test";
 
+import type { FileDiffMetadata } from "@pierre/diffs";
+
 import { DiffFileTree, type DiffFileTreeEntry } from "./DiffFileTree";
+import { diffFileTreeEntries } from "./diffFileTree.logic";
 import { useCodeViewFileReveal } from "./useCodeViewFileReveal";
 
 vi.mock("../../hooks/useTheme", () => ({ useTheme: () => ({ resolvedTheme: "dark" }) }));
@@ -110,6 +113,22 @@ describe("diff tree file activation", () => {
     vi.runOnlyPendingTimers();
     vi.useRealTimers();
     vi.unstubAllGlobals();
+  });
+
+  it("mounts a file-to-symlink type change as one tree row", async () => {
+    // Git carries a type change as a deletion and an addition of the same path.
+    const typeChange = diffFileTreeEntries([
+      { type: "deleted", name: "b/AGENTS.md", prevName: "a/AGENTS.md" } as FileDiffMetadata,
+      { type: "new", name: "b/AGENTS.md", prevName: "a/AGENTS.md" } as FileDiffMetadata,
+      { type: "change", name: "b/CLAUDE.md", prevName: "a/CLAUDE.md" } as FileDiffMetadata,
+    ]);
+    await mount({ files: [...typeChange] });
+
+    const tree = model();
+    expect(tree.getItem("AGENTS.md")?.isDirectory()).toBe(false);
+    expect(tree.getItem("CLAUDE.md")?.isDirectory()).toBe(false);
+    await activate("AGENTS.md");
+    expect(targets).toEqual([{ type: "item", id: "AGENTS.md\u0000AGENTS.md", align: "start" }]);
   });
 
   it("reissues the reveal when the sole selected file is activated again", async () => {
--- a/apps/web/src/components/diffs/diffFileTree.logic.test.ts
+++ b/apps/web/src/components/diffs/diffFileTree.logic.test.ts
@@ -31,6 +31,23 @@ describe("diffFileTreeEntries", () => {
   });
 });
 
+describe("diffFileTreeEntries", () => {
+  it("folds a file-to-symlink type change into one modified entry", () => {
+    expect(
+      diffFileTreeEntries([
+        file("change", "CLAUDE.md"),
+        file("deleted", "AGENTS.md"),
+        file("new", "AGENTS.md"),
+        file("new", "docs/new.md"),
+      ]),
+    ).toEqual([
+      { path: "CLAUDE.md", status: "modified" },
+      { path: "AGENTS.md", status: "modified" },
+      { path: "docs/new.md", status: "added" },
+    ]);
+  });
+});
+
 describe("collectDirectoryPaths", () => {
   it("lists every ancestor once, parents first, with Pierre's trailing slash", () => {
     expect(collectDirectoryPaths(["apps/web/src/a.ts", "apps/web/b.ts", "README.md"])).toEqual([
--- a/apps/web/src/components/diffs/diffFileTree.logic.ts
+++ b/apps/web/src/components/diffs/diffFileTree.logic.ts
@@ -23,11 +23,22 @@ function toGitStatus(file: FileDiffMetadata): GitStatus {
   }
 }
 
-/** Maps parsed diff files to tree entries, keeping the diff's own order. */
+/**
+ * Maps parsed diff files to tree entries, keeping the diff's own order. A path
+ * appears once: a type change (regular file to symlink) is a deletion plus an
+ * addition of the same path, and the tree shows the surviving file as modified.
+ */
 export function diffFileTreeEntries(
   files: ReadonlyArray<FileDiffMetadata>,
 ): ReadonlyArray<DiffFileTreeEntry> {
-  return files.map((file) => ({ path: resolveFileDiffPath(file), status: toGitStatus(file) }));
+  const statusByPath = new Map<string, GitStatus>();
+  for (const file of files) {
+    const path = resolveFileDiffPath(file);
+    const status = toGitStatus(file);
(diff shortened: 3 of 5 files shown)


shivamhwp:

The new key fixes the CodeView duplicate-ID crash, but opening **Show file tree** on the same file-to-symlink diff still crashes the whole thread with `Duplicate path: "AGENTS.md"`. Reloading crashes again because the file-tree preference stays enabled.

`diffFileTreeEntries` still emits a deletion and an addition with the same path, and `DiffFileTree` passes both to `model.resetPaths`. Please combine those blocks into one tree entry with an appropriate status, while preserving the distinct diff blocks, and add regression coverage for opening the tree and reloading with it enabled.

Mnigos:
@shivamhwp good catch, fixed in `5a2c44e`: same-path blocks fold into one tree entry, with a tree test.

shivamhwp:

Updated the branch with current `main` and rechecked the fix at 913272855d. Both file-to-symlink and symlink-to-file changes pass browser verification, including file-tree navigation, reload, and split view. The update needed two test files adjusted for the current parser and diff identities; all 34 focused tests and required CI checks now pass. We're proceeding with the merge.

For future changes, please include before/after screenshots or a short recording so I can visually verify the behavior. Here's the evidence from this review:

**Before: the old code crashes**

Old code crashes with Duplicate path: AGENTS.md

**After: both file-to-symlink and symlink-to-file changes render correctly on the updated branch**

Both type changes render correctly, with one tree entry per path

**After reload: split view and the file tree work**

Split view works after reload with the file tree enabled

PR #11093, opened 2026-09-10
fix(server): give slow OpenCode version probes more time

### LINKED ISSUE #11080: [Bug]: OpenCode provider flagged Unavailable when CLI version probe exceeds hardcoded 4s
### Before submitting

### Area

Not sure (CLI health check path)

### Steps to reproduce

1. Configure the OpenCode provider with a binary that is a thin internal wrapper around the OpenCode CLI. The wrapper runs some startup checks before handing off, which can take a few seconds on a cold cache.
2. Let that cache go cold (e.g. after it expires) so the next launch hits the slow path.
3. T3 runs the health check (`<binary> --version`) with its 4s timeout.
4. Cold-path startup pushes `--version` past 4s.

### Expected behavior

The provider tolerates realistic CLI cold-start, or the probe timeout is configurable per provider instance. A single slow probe should not flip an otherwise-working, already-versioned provider to Unavailable.

### Actual behavior

Probe fails with `OpenCode CLI version probe timed out after 4 seconds` and the provider flips to Unavailable until the next (warm) probe. `OPENCODE_VERSION_PROBE_TIMEOUT = "4 seconds"` is a hardcoded literal with no env var or settings override.

### Impact

Minor bug or occasional failure (recurs whenever the wrapper's cache goes cold).

### Version or commit

0.0.41-nightly.20260909.1439

### Environment

macOS (Darwin 25.6.0); OpenCode CLI 1.18.19 invoked via a thin internal wrapper that runs startup checks.

### Logs or stack traces

```shell
# warm cache
<wrapper> --version  -> ~1-2s
# cold cache (wrapper does startup work first)
<wrapper> --version  -> ~4.7s   # exceeds the 4s probe -> provider marked Unavailable
```

### Workaround

Wrap the binary so `--version` short-circuits to the raw OpenCode CLI, bypassing the wrapper's startup work.

### Suggested fix

Make the version-probe timeout configurable per provider instance (or raise the default, or decouple the version probe from full CLI initialization so a cold start doesn't fail the health check).

mdiniz97:
## What Changed

Raise the OpenCode version probe timeout from 4s to 10s.

Fixes #11080

## Why

A cold wrapper around the OpenCode CLI can exceed the hardcoded four-second
`--version` budget, which flips a healthy provider to "Unavailable" until the
next probe. The fix gives slow wrappers room to finish, and covers it with a
test where the probe resolves at 5s. The hanging-probe test now asserts the
ten-second budget.

Tests: `vp test run apps/server/src/provider/Layers/OpenCodeProvider.test.ts` (14 passed)

## Checklist

- [ ] I included before/after screenshots for any UI changes (n/a)

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased the OpenCode CLI version-check timeout to 10 seconds.
  * OpenCode installations that respond within the extended timeout are now recognized with a warning status instead of being incorrectly reported as unavailable.
  * Updated timeout messaging to accurately reflect the 10-second limit.

### DIFF
--- a/apps/server/src/provider/Layers/OpenCodeProvider.test.ts
+++ b/apps/server/src/provider/Layers/OpenCodeProvider.test.ts
@@ -37,6 +37,7 @@ const runtimeMock = {
   state: {
     runVersionError: null as Error | null,
     runVersionPending: false,
+    runVersionDelayMs: null as number | null,
     versionStdout: DEFAULT_VERSION_STDOUT,
     inventoryError: null as Error | null,
     connectionError: null as Error | null,
@@ -56,6 +57,7 @@ const runtimeMock = {
   reset() {
     this.state.runVersionError = null;
     this.state.runVersionPending = false;
+    this.state.runVersionDelayMs = null;
     this.state.versionStdout = DEFAULT_VERSION_STDOUT;
     this.state.inventoryError = null;
     this.state.connectionError = null;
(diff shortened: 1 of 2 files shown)


juliusmarminge:

PR #11097, opened 2026-09-10
Fixed pin error text and cleaned up copied mobile helpers

kridaydave:
PR made by muse-spark-1.3-contributor from opencode on behalf of Kriday.

Pin failure showed [object Object] when the cause was not an Error. Now it shows the real message with a plain fallback, matching the move thread branch. The rest is small cleanup with no behavior change.

Summary:
- Show real pin error text with fallback instead of String(Cause.squash())
- Drop stray empty string arg in agent awareness debug log
- Share one review diff log helper instead of two copies
- Share one review perf measure helper for sync and async

## Summary by CodeRabbit

- **New Features**
  - Added the ability to rename threads directly from the thread list.
  - Prompts for a new title and prevents empty or unchanged names from being submitted.

- **Bug Fixes**
  - Improved error alerts when moving threads between sections, providing clearer, more human-readable messages.
  - Refined diagnostic logging so messages without additional details no longer include an empty value.

- **Refactor**
  - Standardized review performance measurement and diagnostic handling while preserving existing behavior.


kridaydave:
Docstring coverage note: added one-liners to the two new helpers (start/finishReviewMeasure). The other flagged functions are pre-existing private debug/perf helpers this PR only touched in passing, and they match the undocumented style of the surrounding files. Leaving them as-is rather than adding filler comments to chase the percentage.

kridaydave:
@juliusmarminge If you are free can you have a look at this PR? Has been open for a week for simple error fix.

PR #11103, opened 2026-09-10
fix(web): preserve drafts when compacting context

maria-rcks:
Compacting from the resume banner or context meter was disabled whenever the composer contained a draft. The compact action now sends a standalone command while preserving the draft, attachments, and cursor for the next message.

Verified with a real Codex session: compacted with a draft, then sent it and received the expected reply. A separate check preserved a staged file and inserted text at the original caret after starting compaction. Web typecheck, scoped lint, and 122 existing composer/draft/context-meter tests passed.

Compacting with a draft, then sending it and receiving the provider reply

Implemented with GPT-6 in Codex.

### DIFF
--- a/apps/web/src/components/ChatView.tsx
+++ b/apps/web/src/components/ChatView.tsx
@@ -5804,15 +5804,13 @@ export default function ChatView(props: ChatViewProps) {
     pendingApprovals.length > 0 ||
     pendingUserInputs.length > 0 ||
     showPlanFollowUpPrompt;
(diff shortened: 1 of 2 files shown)


maria-rcks:
> [!NOTE]
> Written by `gpt-6-astra` on behalf of Maria

Reviewed the non-blocking suggestions. The callback catches its own dispatch failure; adding `void` is a style-only change. Keeping this PR focused and adding no new test files under the task constraints. Verified the real Codex flow with draft text, a staged file, and caret preservation; 122 existing focused tests and the web typecheck passed.

PR #11122, opened 2026-09-10
fix(web): stabilize unpin icon when hovering a pinned thread, swap icon to pin-off

flamboh:
> [!NOTE]
> 🤖 **Claude Fable 5.1 on behalf of Oliver**

## ELI5

Hovering a pinned thread in the sidebar made its pin icon jump sideways. Now it stays put, and it turns into a pin-off icon so you can tell it unpins.

## Why

The pin sits directly left of the row's status slot. That slot swapped which state was in flow on hover, from the time label to the Settle and snooze actions, so its width changed and dragged the pin with it. The slim row had the same problem in its Woke state, where the action goes static on hover.

## Fix

The card slot now stacks the status label and the hover actions in one grid cell, so it is always as wide as its wider state and only opacity swaps on hover. Nothing left of the slot moves. The cost is a little reserved space beside short labels like "2m ago" on idle rows; working rows lose nothing since their label is already wider than the actions. The slim row keeps its Woke action in flow at rest for the same reason. The unpin button also swaps to the pin-off glyph on hover and keyboard focus so the icon reads as the action it performs.

### Before

https://github.com/user-attachments/assets/dc53e08e-00d7-4d5a-a400-ec08c4b42b99

### After

https://github.com/user-attachments/assets/7de5a633-6f97-4a5e-a551-cf7888d5021e

## Checklist

Made by Claude Fable 5.1 in Claude Code.

## Bug Fixes

- Improved the pinned-thread unpin control with clearer visual feedback when hovered or focused with the keyboard.
- Stabilized action button positioning in woke rows, reducing unexpected movement during interaction.
- Prevented the pin marker from shifting when status labels and actions appear in card rows, keeping the layout consistent.


maria-rcks:
i... dont love it on the right, can we think of something better?

flamboh:
@maria-rcks yes! I'll try some stuff and reach u with some options

flamboh:
@maria-rcks 

1

https://github.com/user-attachments/assets/0a30c6e9-d957-4753-877b-cd5d6375789a

2

https://github.com/user-attachments/assets/90f8cd64-2d4d-401e-9a74-43e8b9818f50

1 is on the PR right now since it's the more minimal change, but I do like 2 a bit more since it avoids the project title truncation.

PR #11142, opened 2026-09-10
fix(server): survive a failed stderr write

andrewcai8:
## What changed

`apps/server/src/stderrGuard.ts` wraps the entrypoint's stderr so a failed write cannot end the process. `bin.ts` calls it inside the existing `isEntrypoint` block, so importing the module for tests still has no side effects.

## Why

The desktop app runs the server as a child process with piped stdio. `DesktopBackendManager` spawns it with `stdout`/`stderr` set to `"pipe"` and drains both. While that pipe is healthy everything is fine, but when a write starts failing with `EIO` the server dies for a reason that has nothing to do with its own work.

Node prints process warnings itself, from `node:internal/process/warning`. Its `writeOut` calls `console.error` with no error handling of any kind, so a single failed write there becomes an uncaught exception. The warning is incidental. `ExperimentalWarning: SQLite is an experimental feature` is enough, and this package emits that one on every run. Observed as an Electron uncaught-exception dialog with this stack:

```
Error: write EIO
    at afterWriteDispatched (node:internal/stream_base_commons:159:15)
    ...
    at Writable.write (node:internal/streams/writable:508:10)
    at console.value (node:internal/console/constructor:310:16)
    at console.error (node:internal/console/constructor:441:26)
    at writeOut (node:internal/process/warning:56:3)
```

Guarding the stream rather than replacing the `warning` listener is deliberate. Owning the listener would mean reimplementing Node's warning formatting and would silently drop `--trace-warnings`. Wrapping the write leaves both alone and removes only the ability to crash.

A swallowed write reports success rather than failure. A caller told the write was buffered waits for a `drain` event that a broken pipe never emits.

## Validation

- `vp test run src/stderrGuard.test.ts`, 3 passed.
- Red first. With the guard body replaced by a no-op, 2 of the 3 fail, and the pass-through case still passes because it does not depend on the guard.
- `tsc --noEmit` in `apps/server`, exit 0, no diagnostics on the changed files.
- `vp lint` on the three changed files, exit 0.
- `vp test run src/bin.test.ts`, 23 passed.

`src/entrypoint.test.ts > matches through a symlinked entrypoint` fails both with and without this change when the checkout lives under `/tmp` on macOS, since `/tmp` is itself a symlink to `/private/tmp`. Pre-existing and unrelated.

## One related thing I did not change

`drainBackendOutput` in `apps/desktop/src/backend/DesktopBackendManager.ts` catches `BackendProcessOutputReadError` outside `Stream.runForEach`, so a read error ends the drain permanently instead of resuming it. Nothing would read that pipe again for the rest of the backend's life, which is one way to reach the state above. I found no evidence in local traces that it actually fired, so I left it to whoever owns that code rather than encode an unproven theory in a fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when writing diagnostic output fails, helping prevent unexpected process termination.
  * Stderr write errors are now handled gracefully while preserving normal output behavior.

* **Tests**
  * Added coverage for successful writes, write failures, and emitted stream errors.

### DIFF
--- a/apps/server/src/bin.ts
+++ b/apps/server/src/bin.ts
@@ -16,6 +16,7 @@ import { sharedServerCommandFlags } from "./cli/config.ts";
 import { isEntrypoint } from "./entrypoint.ts";
 import { projectCommand } from "./cli/project.ts";
 import { runServerCommand, serveCommand, startCommand } from "./cli/server.ts";
+import { guardStderr } from "./stderrGuard.ts";
 import { serviceCommand } from "./cli/service.ts";
 import { servicePreflightCommand } from "./cli/servicePreflight.ts";
 import { themeCommand } from "./cli/theme.ts";
@@ -75,6 +76,7 @@ if (
     runtimeMain: import.meta.main,
   })
 ) {
+  guardStderr(process.stderr);
   Command.run(cli, { version: packageJson.version }).pipe(
     Effect.scoped,
     Effect.provide(CliRuntimeLayer),
--- a/apps/server/src/stderrGuard.test.ts
+++ b/apps/server/src/stderrGuard.test.ts
@@ -1,0 +1,42 @@
+// @effect-diagnostics nodeBuiltinImport:off - the guard exists for a real Node stream.
+import * as NodeStream from "node:stream";
+
+import { describe, expect, it } from "vite-plus/test";
+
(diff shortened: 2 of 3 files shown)


andrewcai8:
Closing, opened against the wrong repo. This belongs on my fork's own line of work, not upstream.

PR #11150, opened 2026-09-11
fix(server): clean up question uploads when preparation fails

kiiwee:
## What Changed

Roll back newly copied question attachments when a later question's upload claim or attachment-path preparation fails. Keep the pending retry sources and pre-existing attachments, and preserve copies after dispatch because an error there may follow acceptance.

## Why

A multi-question `runtime-request.respond` could copy the first question's upload into the thread store, fail on a later attachment, and leave an orphan despite never dispatching the response. This targets the Orchestrator V2 branch in #2829 and contains only this fix and its regressions.

## Validation

- Reproduced both leaks before the fix: 2 failing regressions, existing dispatch-error retention test passing.
- `vp test run apps/server/src/orchestration-v2/ThreadMessageIntake.test.ts apps/server/src/orchestration-v2/AttachmentClaims.test.ts` — 9 passed.
- Targeted `vp lint --type-aware --type-check` passed for both changed files.
- Formatting applied to the changed files; rebased onto latest V2 branch before publishing.

Implemented by Grok 4.6 through Grok Build, with Codex supervision, patch application, and verification.


juliusmarminge:
Superseded — this fix landed on `t3code/codex-turn-mapping` via #11557 (the newer version carries the same change plus broader test coverage). Thanks for the PR!

PR #11151, opened 2026-09-11
fix(server): report the missing forced restart capability

kiiwee:
An explicit restart request could report that active steering was unsupported even when the provider supported it. Report the missing interrupt-and-restart capability and its matching error detail when restart is forced, preserving ordinary steering behavior.

The new six-case capability matrix fails before the fix; all 16 `CommandPolicy` tests pass afterward. Targeted type-aware lint, type checking, and formatting pass for both changed files.


juliusmarminge:
Superseded — this fix landed on `t3code/codex-turn-mapping` via #11565 (the newer version carries the same change plus broader test coverage). Thanks for the PR!

PR #11152, opened 2026-09-11
fix(orchestrator): restart the live provider session on model changes

kiiwee:
## What Changed

Select the newest live provider session when planning a model switch. Stopped and errored session records remain in history but no longer hide an older session that needs restarting.

## Why

When the newest record was stopped or errored, a model change requiring `restart_and_resume` released no session: selection chose the dead record, while release correctly excluded it. The live process was left running. This focused fix targets Orchestrator V2 (#2829) and preserves native provider-thread history fallback.

## Validation

- Reproduced both stopped/error cases before the fix: expected live session release, received an empty list.
- `vp test run apps/server/src/orchestration-v2/ProviderSwitchService.test.ts` — 4 passed.
- Targeted type-aware lint/type checking and formatting checks passed for both changed files.
- Rebased onto the latest V2 branch before publishing.

Implemented by Grok 4.6 through Grok Build, with Codex supervision, patch application, and verification.


juliusmarminge:
Superseded — this fix landed on `t3code/codex-turn-mapping` via #11505 (the newer version carries the same change plus broader test coverage). Thanks for the PR!

PR #11153, opened 2026-09-11
fix(server): skip scheduled tasks with malformed stored IDs

kiiwee:
An empty stored scheduled-task ID threw during decoding and stopped the entire due-task poll. Decode IDs through the task schema and safely omit invalid ID metadata from the typed error, allowing the existing loader to log and skip the corrupt row.

The SQLite regression fails with a schema-constructor defect before the fix and passes afterward, returning both healthy due tasks while logging the corrupt rows. Targeted type-aware lint, type checking, and formatting pass for both changed files.


kiiwee:
Refreshed onto the current V2 base (`a8cc38b95`) at `6094fb9cb`. The diff is again only two files (+11/-8). Reproduced the malformed-ID failure on current V2; the focused SQLite test passes with the original Grok-authored fix, as do targeted type-aware lint/type checking and formatting. Grok Build re-reviewed the unchanged patch. Full upstream CI is currently `action_required`.

juliusmarminge:
Subsumed by #11585, which contains this same `decodeTaskId` guard plus the corrupt `next_run_at` handling. Thanks!

PR #11154, opened 2026-09-11
fix(server): select an available provider for delegated tasks

kiiwee:
A driver-only delegation could fail on an unavailable inherited provider instance even when another instance of that driver was healthy. Prefer the inherited instance only when it passes the existing availability checks, then fall back to a healthy instance. Explicit instance selection remains strict.

The regression failed before the fix. All 5 MCP service tests pass afterward, covering healthy inheritance, unavailable-account fallback, and explicit-account rejection. Targeted type-aware lint, type checking, and formatting pass for both changed files.

Engineering: Grok 4.6 through Grok Build CLI; review, small test-fixture refinements, and verification: Codex through the Codex harness.


kiiwee:
Refreshed onto the current V2 base (`a8cc38b95`) at `33a5ecadf`. The divergent history is removed; this is again a two-file change. Reproduced the original regression on that base, then verified all 5 MCP service tests plus targeted type-aware lint/type checking and formatting. The original Grok-authored fix is unchanged. Full upstream CI is currently `action_required`.

juliusmarminge:
Already fixed on the branch — `OrchestratorMcpService` now selects an available constraint-free provider for delegated tasks. Thanks!

PR #11156, opened 2026-09-11
fix(orchestrator): report terminal runs after wait timeout

kiiwee:
## What Changed

Recompute the wait result's `timedOut` flag from the run status returned by the final projection read.

## Why

A run can finish while `waitForThread` is handling its timeout. The final read then returns a completed run, but the response previously still reported `timedOut: true`. This focused Orchestrator V2 (#2829) fix reports completion consistently while preserving active-run timeouts and the missing-run error.

## Validation

- Reproduced the completed-run mismatch before the fix; active and missing-run controls passed.
- `vp test run apps/server/src/orchestration-v2/ThreadManagementService.test.ts` — 10 passed, using Deferred and TestClock for deterministic timeout coverage.
- Targeted type-aware lint/type checking and formatting passed for both changed files.
- Rebased onto the latest V2 branch before publishing.

Implemented by Grok 4.6 through Grok Build, with Codex supervision, fixture compatibility corrections, patch application, and verification.


juliusmarminge:
Duplicate of #11574 — identical `timedOut: !isTerminalRunStatus(run.status)` fix; that PR is rebased and carries extra regression coverage. Thanks!

PR #11157, opened 2026-09-11
fix(server): wake paused Cursor replay runs on mismatch

kiiwee:
A Cursor replay run waiting for a recorded cancellation could remain suspended when another operation mismatched the transcript. The failure was stored, but the paused waiter was never notified.

Resolve the existing progress signal when recording a replay failure so waiters report that same typed failure without advancing the transcript. Add a regression that pauses a run, issues a mismatching close, and verifies both operations receive the same mismatch.

Validation: the new test timed out before the fix; all 4 replay-testkit tests pass afterward without adding sleeps or timeout-based passing assertions. Scoped type-aware lint/typecheck and formatting pass.

Targets the Orchestrator V2 branch in #2829. Replay harness only; no client or wire-schema changes.

Engineered with Grok 4.6 via Grok Build CLI; reviewed, applied, and verified by GPT-6 via Codex.


juliusmarminge:
Superseded — this fix landed on `t3code/codex-turn-mapping` via #11535 (the newer version carries the same change plus broader test coverage). Thanks for the PR!

PR #11158, opened 2026-09-11
fix(server): reject partial output from failed Cursor runs

kiiwee:
Cursor text generation accepted valid JSON from failed or cancelled SDK runs as successful metadata. Require a finished run before decoding its output, and report failure details accurately even when partial output exists.

Both new partial-output regressions fail before the fix. All 10 Cursor text-generation tests pass afterward, including successful generation and resource cleanup. Targeted type-aware lint, type checking, and formatting pass for both changed files.


saphid:
Heads up: this fix was adapted onto a current `t3code/codex-turn-mapping` base in #11534, since this branch carries a stale divergent base (300+ file diff). The fix itself is credited to you there.

juliusmarminge:
Superseded — this fix landed on `t3code/codex-turn-mapping` via #11534 (the newer version carries the same change plus broader test coverage). Thanks for the PR!

PR #11160, opened 2026-09-11
fix(server): replay all command events across persistence pages

kiiwee:
Command replay silently stopped after the persistence layer's default 1,000-event limit. Page command-filtered reads by sequence so every stored event is returned in order while each database read remains bounded.

The SQLite regression returned only 1,000 of 1,001 matching events before the fix, with unrelated events interleaved. All 36 foundation persistence tests now pass, including empty and single-page replay. Targeted type-aware lint, type checking, and formatting pass for both changed files.


juliusmarminge:
Superseded — the same cross-page replay fix landed via #11499, which ports it to the current persistence layout (`persistence/Layers/OrchestrationEventStore.ts`). Thanks!

PR #11162, opened 2026-09-11
fix(server): show title and branch generation failures

### LINKED ISSUE #5359: [Bug]: Thread title and branch-name generation fail silently when the text generation provider is unhealthy

shivamhwp:
## What Changed

When title or worktree branch generation fails, the thread now shows the error. Automatic titles report after the existing retries finish; manual regeneration reports the error and clears its pending state. The turn continues with the existing title and branch. Cancellation, manual renames, and replacement regeneration requests do not produce stale error entries.

## Why

These failures were only logged on the server, so an unhealthy text-generation provider made naming look broken without an explanation. This implements the failure-reporting option in #5359. Provider selection and retry policy are unchanged, including the retry classification requested in the later issue comment.

The existing activity contract and web, desktop, and mobile renderers carry these errors. No provider-specific behavior or transport changes are needed.

## UI Changes

Same disposable thread and browser. A temporary generator fault produced a deterministic configuration error through the real regeneration command. The separate unknown-provider error prevents a real agent launch and appears in both captures. The fixture was removed before committing. Native desktop/mobile rendering was not exercised.

| Before | After |
| --- | --- |
| Regeneration silently fails | Regeneration failure appears in the thread |

Validation: all 68 reactor tests pass, covering exhausted automatic retries, branch-generation failure, manual regeneration, cancellation, and stale failures after title or branch changes. Four race regressions fail before the review corrections. The manual-failure regression fails on the baseline. Scoped lint has existing warnings only. Server TypeScript checking passes after repairing the worktree dependency installation.

Internal Fable refinement is pending after connection failures.

Review decision: reporting includes naming operations that fail while loading settings, renaming the Git branch, or saving the generated title. These errors prevent the requested naming operation from completing, so hiding every non-`TextGenerationError` would restore a silent failure. The branch summary explicitly says "generate or rename" and the activity detail preserves the underlying error.

## Checklist

- [x] I included before/after screenshots for UI changes

Closes #5359

## Summary by CodeRabbit

- **New Features**
  - Text-generation failures during thread title creation or regeneration are now surfaced as failure activities.
  - Worktree branch-name generation and rename failures are now surfaced as failure activities.

- **Bug Fixes**
  - Exhausted title and branch-generation retries no longer fail the active turn.
  - Thread title generation now retries according to the configured schedule.
  - Canceled title generation no longer emits an unnecessary failure activity.
  - Outdated failures are ignored after a manual rename or title replacement.

Rebased on main `20363c32c9bfdbf49c2716ef11d1f18483fcc01b`. Scoped validation rerun after rebasing; existing native and human review limits remain.


juliusmarminge:

PR #11165, opened 2026-09-11
fix(server): retry a failed provider version lookup on the next probe

### LINKED ISSUE #11163: [Bug]: Provider update prompt never shows when npm registry lookup exceeds the hardcoded 4s timeout
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Run T3 Code on a machine where DNS resolution is slow on cold lookups (mine goes through a local filtering resolver; `registry.npmjs.org` takes 11-14s to resolve the first time, ~0.4s once warm).
2. Have an outdated Claude Code (or any npm-managed provider) installed via `npm i -g`.
3. Open T3 Code and wait.

```
$ curl -s -o /dev/null -w "dns=%{time_namelookup}s total=%{time_total}s\n" https://registry.npmjs.org/%40anthropic-ai%2Fclaude-code/latest
dns=11.231198s total=11.568347s
dns=0.007778s total=0.399317s
```

### Expected behavior

The "provider update available" toast shows up, like it used to for me.

### Actual behavior

No toast ever appears. Settings > Providers still shows the update button and the one-click update works fine, so the maintenance capability detection is right; only the latest-version check is failing.

From reading the server bundle: `fetchNpmLatestVersion` wraps the registry request in `timeoutOption(LATEST_VERSION_TIMEOUT_MS)` with `LATEST_VERSION_TIMEOUT_MS = 4e3`. When it times out it returns `null`, `resolveLatestProviderVersion` caches that `null` for `LATEST_VERSION_CACHE_TTL_MS = 36e5` (1 hour), and `deriveVersionAdvisory` yields `status: "unknown"`, so the client's `behind_latest` gate never fires. One slow DNS hit at startup silences the prompt for an hour, and every hourly retry has to beat the same 4s window.

Suggestions, any one of which would fix it for me:
- Bump the timeout (10-15s) or make it configurable.
- Don't cache a `null` result for the full hour; retry sooner on failure.
- Log the timeout somewhere visible so it's not a silent "unknown".

### Impact

### Version or commit

T3 Code (Alpha) 0.0.40.0, Claude Code provider 2.1.263 (latest 2.1.268)

### Environment

Windows 11 Pro 10.0.26200, Claude provider installed via `npm i -g @anthropic-ai/claude-code`, binaryPath pointed at `%APPDATA%\npm\node_modules\@anthropic-ai\claude-code\bin\claude.exe`

Mnigos:
The provider update prompt depends on `resolveLatestProviderVersion`, which asks the npm registry for the latest version under a 4 s timeout. When that single request times out (a cold DNS lookup on a filtering resolver can take over 10 s) or fails, the `null` result is cached for the full one-hour TTL, so `deriveVersionAdvisory` stays `unknown` and the "update available" toast never appears, even though the next lookup would take well under a second.

A failed or timed-out lookup is now cached for one minute instead of an hour, so the next provider probe retries it. Successful lookups keep the hour-long cache, and the request timeout is unchanged.

## Verification

- New test in `providerMaintenance.test.ts`: a 503 from the registry yields `null` with a 60 s cache entry, the failure is still served from cache inside that minute, and after the test clock passes it a 200 response resolves `2.0.0` with the hour-long entry.
- `vp test run apps/server/src/provider/providerMaintenance.test.ts`: the new test and the existing cache tests pass. Two npm-ownership tests in that file fail identically on an untouched `main` checkout on this machine (they depend on the local npm prefix), so they are unrelated.
- Server typecheck and lint on the touched files are clean.

Fixes #11163. Implemented with Claude Code (Claude Fable 5.1).

## Summary by CodeRabbit

- **Bug Fixes**
  - Failed npm registry lookups are retried after one minute instead of being cached for the full hour.
  - Successful version lookups continue to be cached for one hour, ensuring update notifications remain timely.

### DIFF
--- a/apps/server/src/provider/providerMaintenance.test.ts
+++ b/apps/server/src/provider/providerMaintenance.test.ts
@@ -8,10 +8,12 @@ import * as NodePath from "node:path";
 import { ProviderDriverKind, ProviderInstanceId, type ServerProvider } from "@t3tools/contracts";
 import { HostProcessPlatform } from "@t3tools/shared/hostProcess";
 import * as Crypto from "effect/Crypto";
+import * as DateTime from "effect/DateTime";
 import * as Effect from "effect/Effect";
 import * as Sink from "effect/Sink";
 import * as Stream from "effect/Stream";
-import { HttpClient } from "effect/unstable/http";
+import * as TestClock from "effect/testing/TestClock";
+import { HttpClient, HttpClientRequest, HttpClientResponse } from "effect/unstable/http";
 import { ChildProcessSpawner } from "effect/unstable/process";
 import {
   createProviderVersionAdvisory,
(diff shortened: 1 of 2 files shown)


juliusmarminge:

PR #11174, opened 2026-09-11
fix(server): request a new Explorer window when revealing files

MatthewFeroz:
Windows "Reveal in File Explorer" can appear to do nothing even when the server successfully launches the reveal command. During investigation, Explorer windows for previously requested files were present but hidden. The original cause of those windows becoming hidden is still unknown.

This draft proposes adding `/n` to the existing PowerShell-launched `/select,"<path>"` command to request a new Explorer window. It preserves path normalization and raw argument quoting, and updates the Windows and WSL command expectations plus the real Windows PowerShell argument smoke test. A tradeoff is that repeated reveals request additional windows rather than reusing an existing one.

Validation:

- Updated command expectations failed before the implementation change and passed afterward. Focused launcher suite on Windows: 9 passed, 13 platform-specific tests skipped, including the WSL simulation tests. The real PowerShell quoting test passed.
- Targeted lint, formatting, and server typecheck passed (typecheck emitted existing suggestions).
- A separate Windows experiment opened an isolated fixture folder, hid only its test window, and launched the reveal command for a filename containing a space and an apostrophe. The proposed command produced a visible window with the correct file selected. **The original command also succeeded in this experiment**, so this is compatibility evidence, not a reproduction proving the mitigation resolves the reported intermittent failure.

Keeping this as a draft pending validation against a recurrence of the actual failure. The needed comparison is whether the old command leaves the requested folder invisible while `/n,/select` presents the correct file. Successful process creation or a passing argument test alone is insufficient. This PR does not establish an introducing commit or claim the issue is fixed.

## Summary by CodeRabbit

* **Behavior Changes**
  * File Explorer now opens a new window when revealing a selected file.
  * The target file remains selected using the existing reveal action.

### DIFF
--- a/apps/server/src/process/externalLauncher.test.ts
+++ b/apps/server/src/process/externalLauncher.test.ts
@@ -242,11 +242,11 @@ it.effect("reveals a file in File Explorer through PowerShell on Windows", () =>
     ]);
     const encodedCommand = spawned.args[spawned.args.length - 1] ?? "";
     const decodedCommand = Buffer.from(encodedCommand, "base64").toString("utf16le");
-    // explorer.exe expects `/select,"<path>"` with only the path quoted;
+    // Request a new window with `/n,/select,"<path>"`, quoting only the path;
     // PowerShell 5.1's Start-Process passes the argument string verbatim.
     assert.equal(
       decodedCommand,
-      "$ProgressPreference = 'SilentlyContinue'; Start-Process 'explorer.exe' -ArgumentList ('/select,\"' + 'C:\\workspace with spaces\\media\\author''s clip.mp4' + '\"')",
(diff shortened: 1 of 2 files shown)


MatthewFeroz:
@coderabbitai review

MatthewFeroz:
Reviewed CodeRabbit's results for c6e24b63512a82ec01fb595f0f929185aa4fd9ea. There are no actionable code findings or unresolved review threads, and GitHub CI passed, including the server test jobs and release smoke.

The docstring-coverage warning does not warrant another source change: the exported PowerShell builder already has JSDoc describing the new-window request and quoting constraint, and the smoke test has a comment explaining its real process-chain validation. Extra documentation for coverage alone would conflict with this repository's guidance to document meaningful constraints rather than add redundant prose.

The draft's remaining validation requirement still applies: a recurrence of the reported failure must demonstrate a difference between the old command and `/n,/select`. The isolated hidden-window experiment succeeded with both commands, so green CI and bot review do not prove the intermittent issue is resolved.

PR #11183, opened 2026-09-11
fix(server): bootstrap attachment cleanup without decoding the event log

### LINKED ISSUE #11182: [Bug]: Startup OOM with oversized replay batches remains unaddressed by #10777
### Before submitting

### Area

apps/server

### Steps to reproduce

1. Launch nightly 1387 with my existing profile. The GUI opens and existing threads are usable.
2. Update to nightly 1400 and launch against the same profile. The backend repeatedly crashes before a usable GUI appears. I originally encountered this after clicking the in-app update button.
3. Restore 1387 without deleting history or manually editing projection cursors. The same profile becomes usable again.

This is profile-dependent. The relevant condition is historical session events containing unusually large validation-error strings. A focused local test reproduced heap exhaustion when reading a single oversized replay page, without retaining earlier pages. Filtering for deletion/revert events in SQL before loading full payloads completed. This was a focused replay test, not a full patched-desktop test.

### Expected behavior

Updating T3 Code should preserve access to my existing projects and threads. Startup attachment cleanup should process the events it needs with bounded memory, even when historical session records contain oversized errors. An old input-validation error should not make a later release unable to launch.

### Actual behavior

I cannot run the affected recent T3 Code nightlies with my existing history. The GUI never becomes usable because the backend repeatedly exhausts its JavaScript heap and aborts. I have been staying on 1387 to keep working. Sequential install tests confirmed that 1377 and 1387 work, 1400 fails, and rolling back to 1387 works again. Builds 1486 and 1507 also failed during the investigation.

### Follow-up to #10774 and #10777

This is a follow-up to #10774 and its duplicate #10872, focusing on an oversized single-batch failure that the merged fix #10777 does not address.

#10777 fixes retention of previously consumed event pages by replacing recursive concatenation with `Stream.paginate`. However, it still reads and decodes up to 500 full event records per page. It does not bound page size in bytes or filter out events unrelated to attachment cleanup before loading their payloads. Git ancestry checks confirm that failing builds 1486 and 1507 already include #10777.

In a focused local replay test, reading a single oversized page alone caused heap exhaustion. There were no earlier pages to retain. Filtering for deletion/revert events in SQL before loading full payloads completed. This distinguishes the failure reproduced here from the consumed-page retention fixed by #10777.

### Original regression and remaining cause

The startup regression began between 1387 and 1400 with #9871, commit `7220dfe2c949476eaa7d21eccbcd3a0ce0eddb49` (`feat(chat): attach files to question answers`).

In `apps/server/src/orchestration/Layers/ProjectionPipeline.ts`, the new attachment-cleanup bootstrap starts from zero for a missing cleanup cursor and calls `eventStore.readFromSequence(cleanupStart, Number.MAX_SAFE_INTEGER)`. It filters for deletion/revert events only after full event payloads have been fetched and decoded.

The reader in `apps/server/src/persistence/Layers/OrchestrationEventStore.ts` loads 500 full rows per page. A row-count limit does not bound memory when individual payloads are very large. An OOM prevents the cleanup cursor from advancing, so the next launch repeats the failure.

### Historical oversized errors

The affected history contains validation errors that embedded the rejected input itself. Those errors were persisted as `session.lastError` and carried into subsequent session-history records.

This is a historical data problem: testing the default schema formatter shipped with Effect 4.0.0-rc.112 confirmed that it no longer embeds the rejected input by default. I am not claiming the current formatter still produces these giant messages. The existing records remain and still need to be handled safely during replay. Bounding newly persisted error summaries would be a separate preventive measure.

### What has and has not been verified

The release rollback, source comparison, and focused single-page reproduction support this diagnosis. The replay test used the Effect Stream implementation extracted from 1400 with JSON.parse in place of application schema validation. I have not tested #10777 alone in a patched desktop build against my profile, captured a full-app heap profile, or verified a production fix. The conclusion about its remaining gap is based on the actual PR diff and the independent single-page failure.

### Suggested follow-up

A dedicated cleanup query could filter event types in SQL and read only the necessary deletion/revert metadata, while preserving cursor advancement and retry semantics. General replay should also account for oversized batches rather than relying only on a row-count limit. Preventing new oversized records alone would not restore startup for affected existing histories.

### Impact

Blocks work completely

### Version or commit

Last working: 0.0.41-nightly.20260908.1387. First failing: 0.0.41-nightly.20260908.1400. Also failed: 0.0.41-nightly.20260910.1486 and 0.0.41-nightly.20260910.1507. Introducing change: #9871 / 7220dfe2.

### Environment

macOS, Apple Silicon / arm64. Official T3 Code Nightly desktop builds downloaded from GitHub; existing local profile.

### Logs or stack traces

```shell
# Official desktop build: 0.0.41-nightly.20260908.1400
# Two consecutive backend startup attempts.
# Sanitized excerpt: timestamps, process IDs and addresses omitted.
# These are 1400 logs, not a test of #10777 in isolation.

# Attempt 1
5445 ms: Mark-Compact (reduce) 3888.2 (3891.7) -> 3886.6 (3891.7) MB, pooled: 0.0 MB, 19.48 / 0.00 ms (average mu = 0.881, current mu = 0.003) last resort; GC in old space requested
5466 ms: Mark-Compact (reduce) 3886.6 (3891.7) -> 3886.5 (3889.5) MB, pooled: 0.0 MB, 21.39 / 0.00 ms (average mu = 0.770, current mu = 0.000) last resort; GC in old space requested
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory

# Attempt 2
5901 ms: Mark-Compact (reduce) 3888.2 (3894.0) -> 3886.7 (3894.0) MB, pooled: 0.0 MB, 19.71 / 0.00 ms (average mu = 0.889, current mu = 0.002) last resort; GC in old space requested
5923 ms: Mark-Compact (reduce) 3886.7 (3894.0) -> 3886.6 (3892.0) MB, pooled: 0.0 MB, 21.80 / 0.00 ms (average mu = 0.786, current mu = 0.000) last resort; GC in old space requested
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
```

_No response_

### Workaround

Rolling back to the official nightly 1387 restores the GUI and access to my existing threads without deleting history or manually advancing projection cursors. Nightly 1377 and stable 0.0.40 also worked for me. Updating to the affected builds makes the app unusable again.

Diagnosis used Codex assistance, source comparison, read-only inspection, and focused replay tests. No production fix was applied during this investigation.

ThomasCrund:
## What Changed

Attachment cleanup at startup no longer replays the whole event log. A new event store query, `readEventsOfTypes`, returns only `thread.deleted` and `thread.reverted` rows, filtered in SQL. `getHead` captures the newest sequence before replay. Bootstrap streams those rows into the same dedupe map and advances the cleanup cursor to the captured head after a successful pass, even when nothing matched. Ordering, the recreate check, and retry on a failed cleanup are unchanged.

Tests cover the typed reader and the head query, plus a pipeline case where a caught-up history holds an undecodable row of another type. The old bootstrap fails it with `PersistenceDecodeError`. The new one succeeds and lands the cursor on the head.

## Why

Fixes #11182. Follow-up to #10777.

#10777 released consumed pages, but one 500-row page can still hold over a gigabyte of JSON. On the affected profile, page 42001 to 42500 holds 22 `thread.session-set` events that each carry a 73 MB `lastError`. Decoding that page exhausts the heap before the cursor moves, so nightlies 1400 through 1507 crash-loop at startup. Only 17 events in that history matter to cleanup.

A/B on one fresh copy of the profile through `dev:desktop` with an isolated home: the parent commit, which includes #10777, hit `JavaScript heap out of memory` at 3.8 GB about 4 s after spawn, three times in a row. This branch reached `backend ready` on the same copy. Cursor went from 0 to 225560, thread count unchanged.

The three touched server suites pass (71 tests). Typecheck and lint are clean.

Done with Claude Fable 5.1 in Claude Code.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved projection startup reliability by processing only relevant event types during attachment cleanup.
  * Attachment cleanup now consistently reaches the latest available event, including when no matching events are found.
  * Prevented unrelated malformed event data from blocking cleanup of deleted thread attachments.
* **Tests**
  * Added coverage for filtered event reading, empty event stores, malformed data handling, and repeated projection startup.

### DIFF
--- a/apps/server/src/orchestration/Layers/OrchestrationEngine.test.ts
+++ b/apps/server/src/orchestration/Layers/OrchestrationEngine.test.ts
@@ -353,6 +353,8 @@ describe("OrchestrationEngine", () => {
           }),
         ),
       hasEventAfter: () => Effect.succeed(false),
+      readEventsOfTypes: () => Stream.empty,
+      getHead: () => Effect.succeedNone,
       readAggregateRange: () => Stream.die("unused aggregate replay"),
       getAggregateReplayStats: () => Effect.die("unused aggregate replay stats"),
     };
@@ -1498,6 +1500,8 @@ describe("OrchestrationEngine", () => {
         return Stream.fromIterable(events);
       },
       hasEventAfter: () => Effect.succeed(false),
+      readEventsOfTypes: () => Stream.empty,
+      getHead: () => Effect.succeedNone,
       readAggregateRange: () => Stream.die("unused aggregate replay"),
       getAggregateReplayStats: () => Effect.die("unused aggregate replay stats"),
     };
@@ -1738,6 +1742,8 @@ describe("OrchestrationEngine", () => {
         return Stream.fromIterable(events);
       },
       hasEventAfter: () => Effect.succeed(false),
+      readEventsOfTypes: () => Stream.empty,
+      getHead: () => Effect.succeedNone,
       readAggregateRange: () => Stream.die("unused aggregate replay"),
       getAggregateReplayStats: () => Effect.die("unused aggregate replay stats"),
     };
--- a/apps/server/src/orchestration/Layers/ProjectionPipeline.test.ts
+++ b/apps/server/src/orchestration/Layers/ProjectionPipeline.test.ts
@@ -2068,6 +2068,136 @@ it.layer(Layer.fresh(makeProjectionPipelinePrefixedTestLayer("t3-projection-atta
   },
 );
 
+it.layer(Layer.fresh(makeProjectionPipelinePrefixedTestLayer("t3-projection-attachments-typed-")))(
+  "OrchestrationProjectionPipeline",
+  (it) => {
+    it.effect("cleanup bootstrap never decodes unrelated history and advances to the head", () =>
+      Effect.gen(function* () {
+        const fileSystem = yield* FileSystem.FileSystem;
+        const path = yield* Path.Path;
+        const sql = yield* SqlClient.SqlClient;
+        const projectionPipeline = yield* OrchestrationProjectionPipeline;
+        const eventStore = yield* OrchestrationEventStore;
+        const projectionState = yield* ProjectionStateRepository;
+        const { attachmentsDir } = yield* ServerConfig;
+        const now = "2026-01-01T00:00:00.000Z";
+        const projectId = ProjectId.make("project-typed");
+        const threadId = ThreadId.make("thread-typed-gone");
+        const attachmentPath = path.join(
+          attachmentsDir,
+          "thread-typed-gone-00000000-0000-4000-8000-000000000001.png",
+        );
+
+        yield* eventStore.append({
+          type: "project.created",
+          eventId: EventId.make("evt-typed-project"),
+          aggregateKind: "project",
+          aggregateId: projectId,
+          occurredAt: now,
+          commandId: CommandId.make("cmd-typed-project"),
+          causationEventId: null,
+          correlationId: CorrelationId.make("cmd-typed-project"),
+          metadata: {},
+          payload: {
+            projectId,
+            title: "Typed",
+            workspaceRoot: "/tmp/project-typed",
+            defaultModelSelection: null,
+            scripts: [],
+            createdAt: now,
+            updatedAt: now,
+          },
+        });
+        yield* eventStore.append({
+          type: "thread.created",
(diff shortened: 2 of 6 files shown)


juliusmarminge:

PR #11195, opened 2026-09-11
fix(web): composite the agent cursor into browser recordings

saphid:
## What Changed

Browser preview recordings now show the agent's pointer. The raw tab capture is piped through an offscreen canvas that composites the same pointer overlay the live preview shows — arrow at the click/move position, expanding ring on clicks — and the recorder encodes that canvas instead of the bare stream. Recordings stay video-only (no audio track).

Scope: `apps/web` renderer only (the recording compositing step). No changes to the wire contracts, the desktop capture path, other clients, or providers. When the environment can't support the canvas pipeline (or no frame arrives within 1.5s of start), the recording falls back to the raw stream and warns with the failed startup stage and underlying error when available.

## Why

Recordings capture the guest tab via `getDisplayMedia`, which renders no pointer. When an agent drives the browser, the saved video shows clicks happening with no visible cursor, so the recording doesn't match what the user watched live and is much harder to read back. The pointer events (guest viewport CSS coordinates) already stream into the renderer for the live overlay; compositing them onto the canvas at the same relative position is the smallest change that makes the recording match the live view.

Details worth noting for review:

- Coordinate mapping: `event.x * zoom * presentationScale / contentWidth` reduces to the pointer's relative position inside the guest viewport, applied to the captured frame. Zoom and presentation scale cancel out of the ratio.
- The canvas keeps its start dimensions for the whole recording; guest resizes mid-recording letterbox instead of changing the video's dimensions mid-stream (resizing during recording already corrupts exported MP4s upstream).
- The compositor redraws and copies each delivered guest frame for the entire recording. Pointer events add bounded timer redraws for the cursor animation. This per-frame work needs measured performance evidence; bounded cursor animation does not establish low recording overhead.
- `video.play()` is not awaited unbounded: startup races the first captured frame against a 1.5s timeout and falls back to the raw stream.

## Verification

- In `apps/web`, `vp test run src/browser/browserRecordingCursor.test.ts src/browser/browserRecordingCursor.startup.test.ts src/browser/browserRecording.test.ts`: **49 passed**. Includes eight startup tests for dimensions, unsupported frame callbacks, canvas availability, settings errors, play rejection, first-frame timeout, captureStream failure, and successful cleanup.
- `vp exec tsc --noEmit`: exit 0 (existing Effect schema suggestions remain). Targeted lint and formatting pass.
- Independent SWE-2 Max review prompted preserving the underlying error, correcting the canvas-stage diagnostic, protecting frame registration with teardown, and adding missing fallback coverage. A fresh final SWE-2 Max review attempt timed out after 180 seconds without a verdict (launcher timeout status 124); no clean final independent review is claimed.
- A local Chromium component exercise used the actual compositor module, synthetic 1600×1200 canvas input, DPR 2, detached video/canvas, and real MediaRecorder. One 30-second candidate recording decoded with a visible pointer. The preview host delivered only 26 frames; a repeat hit the first-frame timeout and produced the new diagnostic. These observations do **not** establish full desktop capture correctness or a valid performance comparison.

## Remaining merge requirements

Attach a real desktop preview recording of agent interaction through the complete getDisplayMedia capture path, plus a valid approximately 30-second raw/composited CPU or frame-time comparison. The attached-preview host currently cannot supply reliable capture cadence. No performance improvement or full integrated pass is claimed. Required before/after recording evidence and maintainer re-review remain outstanding.

## Checklist

Original implementation used enablers/large in OpenCode with Claude Fable 5.1 review. Fallback diagnostics and current verification: GPT-6 in Codex; independent review: SWE-2 Max via Devin CLI.

## Summary by CodeRabbit

* **New Features**
  * Browser recordings now display the agent cursor, including click indicators.
  * Cursor placement adapts to scaling, zoom, aspect ratios, and letterboxing.
  * Recordings automatically use standard capture when cursor compositing is unavailable.

* **Bug Fixes**
  * Improved cursor visibility and positioning across different recording dimensions.
  * Recording resources are now cleaned up reliably when recordings end or are discarded.

### DIFF
--- a/apps/web/src/browser/browserRecording.ts
+++ b/apps/web/src/browser/browserRecording.ts
@@ -8,6 +8,10 @@ import { previewBridge } from "~/components/preview/previewBridge";
 import { ensureClientSettingsHydrated, getClientSettings } from "~/hooks/useSettings";
 import { appAtomRegistry } from "~/rpc/atomRegistry";
 
+import {
+  attachRecordingCursorCompositor,
+  type RecordingCursorCompositor,
+} from "./browserRecordingCursor";
 import { acquireBrowserSurfaceActivity } from "./browserSurfaceStore";
 
 export class BrowserRecordingUnavailableError extends Schema.TaggedError<BrowserRecordingUnavailableError>()(
@@ -122,6 +126,7 @@ interface ActiveRecording {
   readonly startupSettled: Promise<void>;
   releaseSurfaceActivity: (() => void) | null;
   stream: MediaStream | null;
+  cursorCompositor: RecordingCursorCompositor | null;
   recorder: MediaRecorder | null;
   savedBlob?: Blob;
   uploadPromise?: Promise<string>;
@@ -248,9 +253,12 @@ const preferredMimeTypes = [
   "video/webm",
 ] as const;
 
-const createMediaRecorder = (stream: MediaStream): MediaRecorder => {
+const createMediaRecorder = (
+  stream: MediaStream,
+  settingsStream: MediaStream = stream,
+): MediaRecorder => {
   const mimeType = preferredMimeTypes.find((candidate) => MediaRecorder.isTypeSupported(candidate));
-  const settings = stream.getVideoTracks()[0]?.getSettings();
+  const settings = settingsStream.getVideoTracks()[0]?.getSettings();
   // Browser defaults under-budget native-resolution text and motion. Scale with captured pixels
   // and frames, while bounding storage and encoder load for very large displays.
   const videoBitsPerSecond = Math.round(
@@ -429,6 +437,8 @@ const cleanupFailedRecordingStart = async (
     errors.push(error);
   }
   try {
+    recording.cursorCompositor?.dispose();
+    recording.cursorCompositor = null;
     stopMediaStream(recording.stream);
   } catch (error) {
     errors.push(error);
@@ -524,6 +534,7 @@ export async function startBrowserRecording(
     startupSettled,
     releaseSurfaceActivity,
     stream: null,
+    cursorCompositor: null,
     recorder: null,
     lifecycle: startingLifecycle,
   };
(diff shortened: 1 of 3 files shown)


juliusmarminge (changes requested):
Thanks for this — the approach is the right one (the capture never renders a pointer and the pointer is synthetic anyway, so compositing in the renderer without touching the guest page is the cleanest shape), and the implementation is careful: the coordinate math checks out against the live overlay and `resolveBrowserViewportLayout` for both fill and device viewports, the store namespaces are correct, and teardown is covered on all four exit paths. Tests and typecheck pass locally.

Two things before this can land:

### 1. Integrated evidence

The feature depends on Chromium behaviors unit tests cannot touch: `requestVideoFrameCallback` on a detached `<video>` fed by a MediaStream, `captureStream()` on a detached canvas actually feeding `MediaRecorder`, and `getSettings().width/height` being populated at start. If any of those misbehaves, users either silently get the old cursorless video or a black/frozen one with no error.

Please attach:
- a real recording of an agent clicking in the preview, ideally on a 2x-DPR display so the size/placement scaling is exercised, and
- a before/after CPU (or frame-time) reading during a ~30s recording.

On the second point: the "no continuous repaint loop" claim holds for the cursor animation, but the pipeline itself now does a full-frame `drawImage` + canvas re-capture at the capture frame rate for the entire recording, not only while the pointer is active. That is a per-frame multi-megapixel copy the raw path does not do, so we need a number before changing every recording's default encode path.

### 2. Make the fallback observable

When the compositor returns `null`, the recording quietly degrades to the raw stream. Please `console.warn` with the reason (unsupported `requestVideoFrameCallback`, no track settings, first-frame timeout, `play()` rejection, `captureStream` throw) so a bug report can distinguish "compositor unsupported" from a regression.

### Non-blocking

- Mid-recording guest resizes now letterbox instead of changing stream dimensions. That is a behavior change for all recordings (probably an improvement, given the MP4 corruption note), so please call it out explicitly in the PR body rather than as an implementation detail.
- `settingsStream` on `createMediaRecorder` and the `FrameDrawRect` type could be inlined; low priority.

PR #11199, opened 2026-09-11
feat(web): expose each chat message as a heading for screen readers

### LINKED ISSUE #8535: [Feature]: Expose each chat message as a heading for screen reader navigation

Leos-Khai:
Closes #8535

## What Changed

Every user and assistant message in the transcript now opens with a visually hidden `<h3>` naming its author ("You" / "T3 Code"). The thread title in `ChatHeader` is an `<h2>`, so the outline stays contiguous.

Markdown headings written inside a message are exposed one level lower via `aria-level` (a `#` reads as level 4, a `##` as level 5, clamped at 6) so a reply's own headings never outrank the message heading. `ChatMarkdown` gets a `headingLevelOffset` prop for this; the rendered tag is unchanged, because both the stylesheet (`.chat-markdown h1 {…}`) and copy-as-markdown (`markdown-clipboard.ts`) key off the tag name.

The proposed plan card already renders a visible title; it is now the `<h3>` (same classes, so pixel-identical) and the plan body nests beneath it.

The author heading carries `select-none` so drag-selecting across messages does not put "You" / "T3 Code" into copied text.

## Why

Screen readers skim by heading (`H` / `Shift+H` in NVDA and JAWS, the VoiceOver rotor). The transcript was a flat run of text with nothing marking where one message ends and the next begins, so getting back to an earlier reply meant arrowing through every line in between, code blocks included. One heading per message makes "previous message" a single keystroke and makes the conversation listable.

`aria-level` rather than demoting the real tag: demoting would change heading sizes and make copied markdown come out as `#####`. Overriding only the exposed level keeps the DOM and visuals identical for sighted users.

## UI Changes

None visible. The only DOM additions are `sr-only` headings and `aria-level` attributes.

## Verification

- `vp test run` on `ChatMarkdown.test.tsx` (two new tests for the offset and clamp), `MessagesTimeline.test.tsx`, `MessagesTimeline.logic.test.ts`
- `vp lint`, `vp fmt --check`, and `tsc --noEmit` for `apps/web` — no new diagnostics

## Checklist

- [x] I included before/after screenshots for any UI changes (n/a — no visual change)

## Summary by CodeRabbit

- **Accessibility**
  - Improved screen-reader navigation in chat timelines by identifying message authors with structured headings.
  - Preserved markdown heading appearance while exposing correctly nested heading levels to assistive technologies.
  - Improved heading hierarchy within proposed plans and other message content, including plan titles.

- **Tests**
  - Added coverage verifying nested heading levels, heading-level limits, and unchanged behavior for standard markdown headings.


Leos-Khai:
@juliusmarminge Ready for review whenever you have a moment. This is the screen reader fix for #8535, which I opened. I'm blind, so this is a change I'd be using day to day. I've just rebased on main to clear the conflicts from the context-reference refactor.

PR #11202, opened 2026-09-11
feat(server): add guarded idle session stops

shwarmadev:
## What Changed

- Add an opt-in guarded form of `thread.session.stop` for maintenance clients.
- Require the expected Codex provider, an idle session, no active or queued turn, no pending request, no background liveness, and an unchanged non-future thread snapshot.
- Recheck safety in the provider reactor before stopping and advertise support through the environment capability descriptor.
- Return guarded and replayed guard rejections as invalid requests so clients can distinguish them from access failures.
- Preserve existing unguarded stop and unrelated dispatch behavior.

## Why

Codex app-server processes retain credentials in memory. Account-control clients need a safe way to replace only idle Codex sessions after `auth.json` changes without interrupting active work or other providers.

## Consumer

- https://github.com/Possibility-Laboratories/fastlane-control/pull/4

## Tests

- 129 original focused server tests passed before the review follow-up.
- 26 decider tests pass after the follow-up, including future snapshot rejection.
- 2 guarded HTTP rejection/replay integration tests pass on Node 24.13.1.
- 383 contract tests passed; contract typecheck passed.
- Changed production files pass strict lint and changed-file type-error filtering.

## Checklist

- [x] No UI changes

### DIFF
--- a/apps/server/src/environment/ServerEnvironment.test.ts
+++ b/apps/server/src/environment/ServerEnvironment.test.ts
@@ -171,6 +171,7 @@ it.layer(NodeServices.layer)("ServerEnvironmentLive", (it) => {
       expect(second.capabilities.threadTitleRegeneration).toBe(true);
       expect(second.capabilities.threadPullRequests).toBe(true);
       expect(second.capabilities.threadPullRequestLinking).toBe(true);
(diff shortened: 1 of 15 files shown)


juliusmarminge:

PR #11227, opened 2026-09-11
feat: add Jcode as a first-class provider

JoddabodScripts:
## Problem

Jcode (a local agent CLI that speaks ACP) could not be driven from T3 Code: there was no driver, no picker entry, no way to reuse its existing sessions, and its slash commands and `$`-invocable skills never surfaced in the composer.

## How this fixes it

Jcode is added as an opt-in provider (disabled by default, like Cursor and Grok), across:

**Contracts**: `JCODE_DRIVER_KIND`, `JcodeSettings` (enabled, binaryPath, customModels, hidden homePath), `jcode` literals in `AgentSessionSource` / `UsageProviderKind`.

**Server**: `JcodeDriver`, ACP adapter + spawn support with a Jcode-shaped resume cursor, provider probe via `jcode version` / `jcode auth status --json` / `jcode model list --json`, text generation for usage, driver registration and settings wiring. The ACP runtime's unconditional post-initialize `authenticate` request is now optional (`sendsAuthenticate`); Jcode manages auth in its own daemon and rejects that method.

**Model discovery**: `jcode model list --json` feeds the picker. Available routes first with provider/method in the label, then the selected model, then remaining known models. `jcode-auto` stays first and a failed listing keeps the fallback set, so chats never regress.

**Slash commands**: Jcode's daemon advertises only three commands over ACP `available_commands_update`, but every TUI slash command is accepted as prompt text too (verified live against a raw ACP session). The snapshot lists Jcode's TUI help table (swarm, subagent, plan, commit, review/judge, improve/refactor, todos, compact, rewind, and friends), minus machine-maintenance and release tooling a chat composer user would not invoke.

**Skills**: a filesystem scanner reads `SKILL.md` frontmatter from `~/.jcode/skills`, `~/.claude/skills` (Jcode's compat root), and `<cwd>/.jcode/skills` for the `$` picker, mirroring the Claude and Grok drivers. Jcode keys skills by frontmatter `name` rather than directory (verified: `frontend_design/` publishes as `frontend-design`), so the scanner does the same. `disable-model-invocation` and per-engine `agents/*.yaml` → `policy.allow_implicit_invocation: false` both map to `userInvocationOnly`.

**Session import**: the scanner parses Jcode's whole-object session JSON under `~/.jcode/sessions` (Claude/Codex stay line-oriented JSONL) and the importer stores the Jcode-shaped resume cursor on imported threads.

**Web/mobile**: icon, wizard glyph, settings metadata, usage entry.

## Testing

- `JcodeProvider.test.ts` (14 tests) covers the probe, auth parsing, model list parsing/composition (including the `jcode-auto` dedup case), and CLI failure paths against a fake CLI stub.
- `JcodeSkills.test.ts` (5 tests) covers user/project-scoped discovery keyed by frontmatter name, `disable-model-invocation`, `agents/*.yaml` interface labels and implicit-invocation policy, user-shadowing precedence, malformed-skip, and missing-home graceful empty.
- Typecheck and lint clean for all touched files.

Model: Omen Alpha (Jcode agent)

## Summary by CodeRabbit

- **New Features**
  - Added Jcode as an Early Access provider for chat, usage tracking, settings, and onboarding.
  - Added Jcode support for model selection, authentication, permissions, tools, attachments, and session management.
  - Added discovery and import of Jcode sessions and resumable conversations.
  - Added discovery of project and user Jcode skills.
  - Added Jcode icons across web and mobile interfaces.
  - Added Jcode-powered text generation for titles, branches, commits, and pull requests.
- **Bug Fixes**
  - Improved session resumption for imported Jcode conversations.
  - Added support for providers that manage authentication independently.


JoddabodScripts:
All 9 review items addressed in the amended commit (9bd2b6593):

1. WelcomeWizard grid gains a third icon column for jcode.
2. AgentSessionScanner no longer drops firstUserMessage on overflow; it re-attaches when it slides out of the retained window.
3. USAGE_CONTRACT_VERSION bumped to 6 (jcode merge support).
4. JcodeAdapter applies the selected model via setSessionModel at session init so currentModelId never masks an unapplied selection.
5. JcodeAdapter seeds lastTurnActivityAtNanos at beginTurnLiveness and signals liveness, so quiet prompts (no ACP events) now hit the inactivity watchdog instead of hanging forever.
6. JcodeTextGeneration applies modelSelection via setSessionModel before prompting (mirrors the Grok pattern).
7. Failed-probe fallback snapshots in JcodeProvider now include JCODE_BUILT_IN_SLASH_COMMANDS (all 6 snapshot sites).
8. AgentSessionScanner: JcodeSessionFile fields are length-bounded via Schema.isMaxLength; oversized jcode sessions (>16 MiB) are rejected at both the import read and the cwd metadata scan instead of parsing.
9. JcodeSkills bounds per-file reads (stat + 512 KiB SKILL.md / 64 KiB agents YAML caps); dropped a dead env parameter.

Targeted tests pass (104 + 12), typecheck clean.

JoddabodScripts:
Follow-up fixes for the second review round:

- Jcode now actually appears on the Usage page: `\~/.jcode/sessions` (respects `homePath`) was wired into `resolveTranscriptDirs`, with a dedicated whole-JSON-object reader + `parseJcodeMessage` for `token_usage` on assistant messages (`.bak` rotation copies excluded, session model/sessionId threaded through for attribution and dedupe).
- `resolveTranscriptDirs` previously omitted jcode, so the v6 contract change alone fixed nothing.
- `jcode` gets a distinct series color from `grok` (88% vs 72% neutral mix).
- JSON probes in `JcodeProvider` now parse stdout only, ignoring stderr noise (deprecation warnings would break parsing).
- Also added a guard so jcode scanner validation is enforced before the import read.

JoddabodScripts:
Third review round addressed:

- \`setSessionModel\` failures now propagate at both callsites (session start and per-turn), and \`currentModelId\` only records a model the daemon actually applied, matching the Grok adapter pattern.
- \`resolveJcodeAcpBaseModelId\` maps the product default \`"auto"\` onto \`jcode-auto\` so the canonical ACP slug is never bypassed.
- The liveness watchdog now interrupts the active prompt fiber via a new \`AcpSessionRuntime.interruptActivePrompt\` before draining, so \`sendTurn\\'s joined prompt unblocks when a silent agent stalls a turn.
- \`readJcodeSessionCwd\` rejects files beyond the remaining metadata budget up front, then reserves and reads the complete file (no truncation before JSON.parse).
- Jcode fallback timestamp now derives from \`lastActiveAtMs\` like the other sources.
- Skill/YAML reads are one bounded open+stat+read (maxBytes + 1) instead of a stat pass followed by an unbounded read, closing the swap-between-calls window.

JoddabodScripts:
Round 4 fixes:

- `readJcodeSessionRecords` stats the file first and returns null for sessions over 16 MiB (exported as \`MAX_JCODE_SESSION_BYTES\`, matching the scanner cap) instead of loading arbitrary sizes; regression test added.
- The jcode branch in \`readTranscriptRecords\` now closes the handle in a \`finally\`, fixing the descriptor leak.

JoddabodScripts:
Round 5 fixes:

- \`readAndParseJcodeThread\` now takes the caller's \`identity.size\` and rejects files over \`MAX_JCODE_MESSAGES_BYTES\` before reading, instead of allocating first and checking after.
- Turn-time model selection is normalized through \`resolveJcodeAcpBaseModelId\`, so \`\"auto\"\` maps to \`jcode-auto\` (fixing the string \`\"auto\"\` reaching \`setSessionModel\`) and an explicit \`jcode-auto\` selection becomes a valid target again, letting a custom-model session switch back to auto.

Note: the two outside-diff comments on \`usageTranscriptReader.ts\` were already fixed in \`4823560b9\` (size bound at line 192-205, handle close in \`finally\` at 263-268); this review ran against an older diff range.

Issue #11208, opened 2026-09-11
Fresh installs of t3@0.0.40 hang in an npm ERESOLVE loop since @effect/platform-node-shared@4.0.0-rc.114 was published


idoabra:
### What happened

I couldn't update my Linux background service from 0.0.39 to 0.0.40. Both `npx t3@latest service update` and `npx t3@0.0.40 service update` print this npm warning over and over and never finish:

```
npm warn ERESOLVE overriding peer dependency
npm warn While resolving: @effect/platform-node-shared@4.0.0-rc.114
npm warn Found: peer effect@"^4.0.0-rc.114" from the root project
npm warn
npm warn Could not resolve dependency:
npm warn peer effect@"^4.0.0-rc.114" from the root project
```

### Diagnosis

A fresh npm install of `t3@0.0.40` can't resolve its dependency tree right now. Nothing on the machine causes it. The chain:

1. `t3@0.0.40` depends on `@effect/platform-node@4.0.0-beta.103` (exact).
2. `@effect/platform-node@4.0.0-beta.103` depends on `@effect/platform-node-shared@^4.0.0-beta.103`. A caret on a prerelease matches any later `4.0.0-*` prerelease.
3. `@effect/platform-node-shared@4.0.0-rc.114` went up on npm at 2026-09-11T03:56Z. It has a peer dependency on `effect@^4.0.0-rc.114`, and that version of `effect` doesn't exist. The newest published is `4.0.0-rc.113`.
4. npm picks `platform-node-shared@rc.114`, can't satisfy its peer, and loops on `ERESOLVE overriding peer dependency`.

`t3`'s published `package.json` does set `overrides` that pin every `@effect/*` package to `4.0.0-beta.103`. npm only honors `overrides` in the root project, though, so they do nothing when `t3` is installed as a dependency. The tarball has no `npm-shrinkwrap.json`, so npm re-resolves every transitive range on each install.

This breaks each path that installs `t3` fresh:

- `ensurePinnedRuntimeInstalled` (`apps/server/src/cloud/pinnedRuntime.ts`) runs `npm install --prefix <staging> t3@<version>` into an empty directory. `t3 service install`, `t3 service update` and the remote self-update from the app (`apps/server/src/cloud/selfUpdate.ts`) all go through it. The install spins until `PINNED_RUNTIME_INSTALL_TIMEOUT` (10 minutes) kills it.
[...]

PR #11240, opened 2026-09-11
fix(cli): pin shared Effect dependency for npm installs

jakeleventhal:
Fresh npm installs can hang in an `ERESOLVE` loop because the pinned Effect platform packages allow newer `@effect/platform-node-shared` prereleases. npm selects rc.114, whose `effect` peer is unavailable, and ignores T3's published overrides because T3 is a dependency rather than the install root. Remote Update eventually fails at the install timeout.

Declare `@effect/platform-node-shared` as a direct catalog dependency so the published package pins it alongside the other Effect packages. Keep the intentional transitive pin in Knip's dependency exclusions. This fixes the package consumed by existing updaters and fresh npm installs once a new release is published; it does not repair existing npm tarballs.

Refs #11208. This is a smaller package-level alternative to the installer changes in #11232.

Validation:
- Built the web client and server CLI from upstream main plus this change.
- Exercised the publish command's metadata preparation and restoration with its publish subprocess redirected to `vp pm pack` (no registry publication).
- Installed the resulting tarball outside the workspace with an empty npm cache: 141 packages installed in 10 seconds on macOS arm64, Node 24.12.0 / npm 11.17.0.
- `npm ls effect @effect/platform-node-shared --all` shows one deduplicated rc.112 version of each; the installed CLI's `__service-preflight` returned `ready` with launcher protocol 2.
- All 15 CLI packaging tests pass; frozen-lockfile install, scoped Knip dependency check, formatting, and `git diff --check` pass.

Model: GPT-6. Harness: Codex.

## Summary by CodeRabbit

- **Chores**
  - Added a runtime dependency to improve compatibility for standalone server package installations.
  - Updated dependency analysis configuration to account for the new runtime dependency.

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -25,6 +25,7 @@
     "@anthropic-ai/claude-agent-sdk": "^0.3.260",
     "@effect/platform-bun": "catalog:",
     "@effect/platform-node": "catalog:",
+    "@effect/platform-node-shared": "catalog:",
     "@effect/sql-sqlite-bun": "catalog:",
     "@ff-labs/fff-node": "0.9.4",
     "@opencode-ai/sdk": "^1.3.15",
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -24,8 +24,9 @@
         "scripts/cli.ts",
         "src/provider/testFixtures/*.mjs",
       ],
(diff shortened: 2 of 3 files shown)


mchisolm0:
Thanks for this fix and getting it so quick 😁

PR #11243, opened 2026-09-11
feat(web): show provider usage limits in the composer

eduardosilveiradev:
## What Changed

Adds an always-visible usage-limit meter to the prompt composer. A small level glyph now sits in the footer between the attach button and the send button. Its fill is the quota left in the tightest window of the selected provider instance, drawn the way Usage → Limits draws its bars, and it turns the error colour once 10% or less is left. Hovering it opens a popover with every window: label, percent left, bar with the even-spending mark, pace glyph, and reset countdown.

There was already a limits viewer, but it lives on a separate page (Usage → Limits) and behind the `/usage-limits` command banner. This PR does not add a second data path: the meter reads the same `ServerProvider.usageLimits` snapshot those surfaces use, which the config stream keeps live as turn-driven `account.rate-limits.updated` events merge in. Nothing new crosses the wire.

- `chat/UsageLimitMeter.logic.ts` picks the selected instance's limits (via `providersWithLimits` + `limitsNotice`, so API-key accounts and failed probes hide the meter), selects the headline window, and formats the accessible label.
- `chat/UsageLimitMeter.tsx` is the glyph plus popover. The popover body reuses `LimitWindows` from the Limits page. The clock comes from the shared `useNowMinute` store, not `Date.now()` in render.
- `ChatComposer.tsx` computes the model from `providerStatuses` and `selectedInstanceId` and renders the meter beside the context-window ring. The resting-composer padding accounts for the extra control.
- `usage/UsageLimits.tsx`: the compact grid's label column grows from 9rem to 10rem so "Weekly · Fable" no longer truncates next to its percent.

Focused tests cover the resolver, headline selection, label formatting, and the trigger/popover rendering.

## Why

You can't see how close a session or weekly limit is without leaving the thread. The context-window ring already gives that at-a-glance signal for context; this gives the same for subscription quota, using data the app already holds.

## UI Changes

Before, the footer has attach and send only:

Before: composer footer without the meter

After, the level glyph sits between them. Here Claude's five-hour session has half its quota left:

After: meter glyph in the footer

Hovering lists every window with pace and reset countdown:

After: hover popover

## Checklist

---

Written by Claude Fable 5.1 working with @eduardosilveiradev in T3 Code.

## Closes discussions

- https://github.com/pingdotgg/t3code/discussions/11259

## Summary by CodeRabbit

* **New Features**
  * Added a usage-limit meter to the chat composer, showing remaining quota, provider and plan details, and reset timing.
  * Usage limits now follow the selected provider instance and highlight low or nearly exhausted quotas.
  * Added reset-credit summaries and clearer messages when all usage windows have reset.
  * Improved accessibility labels and popover details for usage-limit status.

* **Bug Fixes**
  * Expired usage windows are now excluded from usage summaries and detailed views.
  * Adjusted composer spacing and limit-window sizing for clearer display.


oposoyo:
Applied this branch on top of current `main` (20363c32c) and ran it against my own Claude subscription. It squash-merged with no conflicts, over the 77 upstream commits since this branch's merge-base. `UsageLimitMeter`'s 11 tests pass and `vp run --filter @t3tools/web typecheck` comes back with 0 errors.

The meter showed my real windows on first render: `Session 69% left · Weekly 87% left · Weekly · Fable 96% left`.

On "why not use what already exists": both surfaces that show this today are pull, not push. `/usage-limits` renders a dismissible banner built from a frozen snapshot that dies when you switch threads, and Usage → Limits is a separate page. Neither tells you you're at 8% until you go looking — which is precisely when you don't. The context-window ring already solved this shape of problem for context; this is the same move for quota, and it reads the same `ServerProvider.usageLimits` snapshot rather than opening a second data path.

Worth noting for the maintenance question: four of the six files are new, so the recurring conflict surface is ~29 lines in `ChatComposer.tsx` plus one line in `usage/UsageLimits.tsx`.

Happy to keep rebasing and testing this branch against `main` if that's useful.

PR #11255, opened 2026-09-11
fix(compadre): bound delivery retries and keep request images out of Postgres

imsherrill:
Request persistence failures could strand a running turn, large inline images amplified Postgres writes, and failed native delivery could retry indefinitely while central state stayed working.

Persist private S3 references before creating a run; bound delivery retries and expose a durable blocked state; retry transient central SQL errors with stable command IDs. Add API-key-protected, canary-only fault injection and inspection through the canonical central command/read path, with runbook and skill updates.

Verification: targeted controller, central ingestion, native event, Temporal policy, and real Postgres persistence tests passed. Controller and server typechecks passed. After merge, verify both Render deployments and exercise live API canaries; no browser verification is claimed. Existing Temporal histories retain replay-safe versioning. Old inline records remain readable; new writes never store image bytes in Postgres.

Model/harness: GPT-5 / Codex.


imsherrill:
Opened against the upstream repository by mistake; this change belongs to the Compadre fork.

PR #11262, opened 2026-09-11
feat(server): show OpenCode subagents in the Agents panel and work log

noueii:
## What Changed

OpenCode sub-agent activity is now surfaced through T3's shared `task.*` runtime
events, at the adapter boundary.

- Correlate a parent `task` tool call to its child session via
  `part.state.metadata.sessionId` (validated against `metadata.parentSessionId`),
  falling back to `Session.parentID` ancestry so event order doesn't matter.
- Emit a deduplicated `task.started` per child (taskId = child session id,
  title/role from the task input, `timelineBypass: true`).
- Route child-session events through a separate branch that never touches parent
  turn state: child tool parts → `task.progress` (summary + last tool), child
  `session.status` busy/idle → `task.updated` running/idle, child
  `session.deleted` → `interrupted`, child `todo.updated` → progress.
- Accept child events only when their ancestry resolves to the thread, so a
  shared OpenCode server cannot leak another thread's activity.

No contract or UI changes — the Agents panel and work log already consume
`task.*` for the other providers.

## Why

T3 already renders sub-agents for Claude, Codex, and Antigravity through `task.*`
events, but the OpenCode adapter dropped all child-session activity. OpenCode's
`task` tool appeared only as a bare `collab_agent_tool_call` with no roster entry
or live status, even though the adapter already tracked child sessions for
permission routing and teardown. Users could see a sub-agent was spawned, but not
what it was doing. This maps OpenCode child sessions onto the existing lifecycle
instead of adding new machinery; a completed child turn is `idle` (resumable),
not terminal, matching the Codex adapter.

## Checklist

https://github.com/user-attachments/assets/64660253-5d3f-4da3-88d2-8baa81d86575

## Before
Screenshot 2026-09-12 020926

## AFTER
Screenshot 2026-09-12 021040

## Summary by CodeRabbit

- **New Features**
  - Child sessions are now represented as tasks, with clearer started, progress, completion, and title updates.
  - Resumed and previously removed child sessions can be rediscovered and restored to task tracking.

- **Bug Fixes**
  - Improved source control provider detection for SSH aliases, including aliases with explicit ports.
  - Improved validation when associating child activity with parent sessions.
  - Improved handling of child-session lifecycle and status updates.
  - Refined approval and question event handling for more reliable user prompts.

### DIFF
--- a/apps/server/src/provider/Layers/OpenCodeAdapter.test.ts
+++ b/apps/server/src/provider/Layers/OpenCodeAdapter.test.ts
@@ -3815,6 +3815,309 @@ it.layer(OpenCodeAdapterTestLayer)("OpenCodeAdapterLive", (it) => {
     }),
   );
 
+  it.effect("maps task tool metadata and child activity to the shared task lifecycle", () =>
+    Effect.gen(function* () {
+      const adapter = yield* OpenCodeAdapter;
+      const threadId = asThreadId("thread-child-task-metadata");
+      const parentSessionId = "http://127.0.0.1:9999/session";
+      const childSessionId = "ses_child_task_metadata";
+      const enqueue = makeOpenCodeEventQueue();
+      const taskEventsFiber = yield* adapter.streamEvents.pipe(
+        Stream.filter(
+          (event) =>
+            event.threadId === threadId &&
+            (event.type === "task.started" ||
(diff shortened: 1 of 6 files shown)


juliusmarminge:

Issue #10966, opened 2026-09-09
[Bug]: Linux AppImage leaks runtime environment into integrated terminal and can conflict with system libraries


Vidhan31:
## Summary
On Linux, the T3 Code AppImage appears to leak AppImage runtime environment variables and mount paths into the integrated terminal. This can cause host/system programs launched from the T3 terminal to resolve AppImage-bundled libraries or otherwise behave differently from the same programs launched from a normal terminal.

This appears related to #1699 and #2509.

## Environment
- OS: Linux
- Installation: T3 Code AppImage
- T3 Code: latest stable release
- Shell: zsh

## Expected behavior
Programs launched from T3 Code's integrated terminal should use the host system environment and system libraries. AppImage-specific variables and mount paths should not leak into unrelated child processes.

## Actual behavior
The integrated terminal can inherit AppImage runtime state such as `APPIMAGE`, `APPDIR`, `OWD`, `ARGV0`, and `LD_LIBRARY_PATH` entries pointing into `/tmp/.mount_T3-*`. AppImage mount paths may also appear in `PATH`.

This can cause system binaries launched from the T3 terminal to behave differently or potentially load incompatible libraries from the AppImage instead of the host system.

## Diagnostics
```bash
env | grep -E '^(APPIMAGE|APPDIR|OWD|ARGV0|LD_LIBRARY_PATH)='
printf '%s\n' "$LD_LIBRARY_PATH" | tr ':' '\n'
```

For a host executable:
```bash
ldd "$(command -v <program>)" | grep -E '/tmp/\.mount_|/usr/lib|/lib'
```

## Suggested fix
Before spawning the integrated terminal/PTy shell on Linux:
- remove `ARGV0`, `APPIMAGE`, `APPDIR`, and `OWD`
- remove `/tmp/.mount_*` entries from `PATH`
- remove `/tmp/.mount_*` entries from `LD_LIBRARY_PATH`

This should ideally be handled by T3 Code itself rather than requiring users to modify their shell configuration.

## Why this matters
The same host binary can work normally outside T3 Code but fail inside the integrated terminal, making these failures subtle and difficult to debug.

## Temporary workaround
```bash
env -u ARGV0 -u APPIMAGE -u APPDIR -u OWD -u LD_LIBRARY_PATH zsh
```

A proper application-level fix would be preferable.

juliusmarminge:
[...]

Issue #5840, opened 2026-08-09
[Bug]: macOS desktop cannot read T3CODE_BITBUCKET_* from a shell profile, so Bitbucket auth cannot be configured


LunarRed:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

1. Add the Bitbucket variables to a shell profile on macOS, either `~/.zprofile` or `~/.zshrc`:

   ```sh
   export T3CODE_BITBUCKET_EMAIL="you@example.com"
   export T3CODE_BITBUCKET_API_TOKEN="your-token"
   ```

2. Confirm the shell really has them: `zsh -ilc env | grep T3CODE_BITBUCKET` prints both.
3. Quit T3 Code with Cmd+Q, then reopen it from the Dock.
4. Open Settings, then Source Control.
5. Optional confirmation: take the server pid from `~/.t3/userdata/server-runtime.json` and run `ps eww <pid>`. The variables are absent from the server environment.

### Expected behavior

`docs/user/source-control.md` says to set `T3CODE_BITBUCKET_EMAIL` and `T3CODE_BITBUCKET_API_TOKEN`, or `T3CODE_BITBUCKET_ACCESS_TOKEN`, then restart T3 Code and verify in Source Control settings. Following that should authenticate Bitbucket.

### Actual behavior

Bitbucket stays on "Not authenticated", showing "Set T3CODE_BITBUCKET_EMAIL and T3CODE_BITBUCKET_API_TOKEN on the server". No shell profile makes this work, and no number of restarts helps.

The cause is the launch context. When you open the desktop app from the Dock or Finder, launchd starts it, so the app's environment comes from launchd and your shell is never in the chain. The server is a child of the app and inherits the same environment, so a shell `export` cannot reach the config reader.

This is separate from the login shell env sync. That path carries `PATH` and tool discovery, not provider credentials.

### Impact

Major degradation or frequent failure

### Version or commit

0.0.32 (Alpha)

### Environment

macOS 26.5.2, Apple Silicon, T3 Code (Alpha) desktop 0.0.32

### Logs or stack traces

```shell
# variables are present in the login shell the app would read
$ zsh -ilc 'env' | grep -c T3CODE_BITBUCKET
2

# but absent from the running server process after a full restart
[...]

Issue #5059, opened 2026-07-30
[Bug]: Linux AppImage replaces XDG_DATA_DIRS and GSETTINGS_SCHEMA_DIR in spawned terminals


AmbitiousRealism2025:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

1. Install T3 Code Desktop from the official Linux AppImage (in this case through the `t3code-bin` AUR package).
2. Launch T3 Code under a Wayland desktop session.
3. Open a T3 Code/Codex thread that can run shell commands.
4. Inspect the relevant environment variables:

```shell
env | grep -E '^(XDG_DATA_DIRS|GSETTINGS_SCHEMA_DIR)='
```

5. Query a standard system GSettings schema:

```shell
gsettings get org.gnome.desktop.interface color-scheme
```

6. Compare with the systemd user-manager environment:

```shell
systemctl --user show-environment | grep -E '^(XDG_DATA_DIRS|GSETTINGS_SCHEMA_DIR)='
```

### Expected behavior

T3 Code should preserve the host's normal data and schema discovery paths for spawned terminals, agents, and child applications.

If the AppImage requires bundled data paths, it should append or scope them narrowly rather than replacing the host paths. A T3-specific `GSETTINGS_SCHEMA_DIR` should not hide system schemas from unrelated child processes.

### Actual behavior

Inside the T3 Code-spawned environment:

```text
GSETTINGS_SCHEMA_DIR=/opt/t3code-bin/usr/share/glib-2.0/schemas
XDG_DATA_DIRS=/home/<user>/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/opt/t3code-bin/usr/share
```

The T3 value for `XDG_DATA_DIRS` omits `/usr/local/share` and `/usr/share`. The T3-specific `GSETTINGS_SCHEMA_DIR` also overrides normal schema discovery.

As a result, a standard query fails:

```text
$ gsettings get org.gnome.desktop.interface color-scheme
No schemas installed
```

The systemd user-manager environment is clean and retains the system paths:

```text
XDG_DATA_DIRS=/home/<user>/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share/:/usr/share/
```

[...]

PR #11286, opened 2026-09-11
feat(desktop): fix login-shell environment capture w config

JoeyEamigh:
this is debatabley a fix() rather than a feat because it completely breaks agents' ability to use the tooling i have installed on my system.

## What Changed

adds shellEnvironmentMode` and `shellEnvironmentNames` to desktop-settings.json so you can configure what gets passed from the login shell.

```json
{
  "shellEnvironmentMode": "allowlist" | "all",
  "shellEnvironmentNames": ["RUSTUP_HOME", "CARGO_HOME"]
}
```

defaults to current behavior.

## Why

depending on whether t3code was launched from `npx t3` or the desktop app, agents may end up with a completely different set of environment variables. `DesktopShellEnvironment` seeds `$PATH` and 16 others, but things like `$BUN_INSTALL`, `$CARGO_HOME`, and `$RUSTUP_HOME` that i have configured in `.zshrc` don't make it into the agents' shells, confusing the hell out of them. perpetually recovering specific env vars from the login shell is not a solution (#10312 for example).

<details>
<summary>claude's slop reasoning for all the review bots</summary>

## Problem

On the same machine, as the same user, agents get different environments depending on how T3 Code started:

- `npx t3` from a terminal: the shell already sourced `~/.zshrc`, so the server and every agent it spawns inherit the user's full environment.
- Desktop app launched from a session manager: it inherits almost nothing, so `DesktopShellEnvironment` probes the login shell to recover it.

Nothing filters the environment on the way to an agent. `mergeProviderInstanceEnvironment` spreads all of `process.env` into every provider spawn. The loss happens earlier, at capture: `capturePosixEnvironmentCommand` issues one `printenv` per name in `LOGIN_SHELL_ENV_NAMES`, and `extractEnvironment` filters by the same 17 names. Anything else the user exported is evaluated by that shell and thrown away.

The allowlist is being extended one variable at a time, and the queue is not keeping up:

| PR | Variable(s) |
| --- | --- |
| #7363 | AWS Bedrock credentials and region (#4928) |
| #9022 | `T3CODE_BITBUCKET_*` (#5840) |
| #7620 | the same, forwarded to WSL |
| #10312 | login-shell `SSH_AUTH_SOCK` |

Each is correct in isolation. This generalizes them.

## Fix

Two keys in `<stateDir>/desktop-settings.json`, both defaulting to today's

```json
{
  "shellEnvironmentMode": "allowlist" | "all",
  "shellEnvironmentNames": ["OPENAI_API_KEY", "CARGO_HOME"]
}
```

- `shellEnvironmentNames` appends to the names the probe asks for.
- `shellEnvironmentMode: "all"` appends a marker-wrapped `env -0` block (`s) to the **same** probe command, so widening the capture never costs anextra login shell. This matters because #10972 is actively trying to remove a redundant shell probe from Linux boot.

No Settings control, deliberately. linuxPasswordStore sets the precedent for a startup-only escape hatch with no IPC surface, and adding one here would pull web and mobile into a desktop-only change. Happy to add it if you'd rather it be discoverable.

Precedence, which is the part worth reviewing:

- User-supplied values win over the inherited process environment. Copy-if-absent would be useless here, since the whole problem is that the GUI-launched process's environment is
the impoverished one.
- User configuration cannot touch the 17 names the existing logic governs. `PATH` keeps its merge, the locale keeps its group rule, XDG keeps its rules. The blast radius on
current behavior is provably zero.
- `"all"` additionally skips `HOME`, `PWD`, `OLDPWD`, `SHLVL` and `_`, which describe the running process rather than the user's environment.

Probe names reach a shell command line, so they are validated against `^[A-Za-z_][A-Za-z0-9_]*$` at both settings load and apply time rather than escaped. The full-dump marker is `*`, which can never be a legal variable name, so a configured name cannot

Follows `linuxPasswordStore`: read at startup, no setter, no IPC, no UI. Npackages/contracts`, web and mobile are untouched.

One ordering change worth a close look: `desktopSettings.load` moves abovektopApp.startup`. The probe cannot move later, because Linux password-store selection reads the `XDG_*` values it hydrates. Loading settings earlier is safe because `desktopSettingsPath` hangs off the resolved state dir, not off Electron's `userData`
path, which is set further down.

## Deliberately not included

- **Per-project environments.** The probe still runs once at startup in thmise and `nix develop` are unaffected. That is #523, closed as not planned, and this does not reopen it.
- **WSL forwarding.** `WSLENV` propagation is #7620's concern.
- **Docs.** `desktop-settings.json` has no existing documented surface to extend.

## Why this is not the AppImage leak class

#10966, #5059, #1699 and #1162 were all runtime environment leaking *outward* from the packaged app into child processes, breaking `PHP_BINARY`, `XDG_DATA_DIRS` and other Electron apps. This reads the login shell, which is the clean user environdesktop process. It overwrites packaging junk with correct values ratherthan propagating it. In `"all"` mode a GUI launch ends up environment-equivalent to a terminal launch, which is the property to check.

## Verification

- `vp run --filter @t3tools/desktop typecheck` — clean
- `vp test run` on the three touched test files — 45 passing, covering defnames, precedence, full-mode parsing (values containing `=` and newlines),the reserved-name skip, `PATH` merge and locale precedence preserved under `"all"`, the Windows profile path, and rejection of names like `FOO; rm -rf /`
- `vp lint` and `vp fmt --check` on the changed paths — clean

Repo-wide checks not run; CI owns those.

Model: Claude Opus 5, harness: Claude Code.
</details>

## workarounds for anyone that lands on this

add the vars to `.zshenv` instead, or wrap the `.desktop` invocation with a shell script that sources the login shell and launch the desktop app under that

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added configurable shell environment harvesting, including disabled, allowlisted, and full-environment modes.
  - Added support for harvesting environment variables from Windows PowerShell profiles.

- **Bug Fixes**
  - Improved handling of empty, multiline, and special-character environment values.
  - Prevented protected process variables from being overwritten.
  - Improved Windows handling for variable-name casing, duplicates, reserved names, and `PATH`.
  - Preserved valid desktop settings when unrelated environment configuration is invalid.

### DIFF
--- a/apps/desktop/src/app/DesktopApp.ts
+++ b/apps/desktop/src/app/DesktopApp.ts
@@ -245,7 +245,11 @@ const startup = Effect.gen(function* () {
   const updates = yield* DesktopUpdates.DesktopUpdates;
   const environment = yield* DesktopEnvironment.DesktopEnvironment;
 
-  yield* shellEnvironment.installIntoProcess;
+  const startupSettings = yield* desktopSettings.load;
+  yield* shellEnvironment.installIntoProcess({
+    mode: startupSettings.shellEnvironmentMode,
+    names: startupSettings.shellEnvironmentNames,
+  });
   const hasCommandLinePasswordStore =
     preReadyElectronOptions.linuxPasswordStoreCommandLine !== null;
   const linuxElectronOptions =
@@ -269,7 +273,6 @@ const startup = Effect.gen(function* () {
   const userDataPath = yield* appIdentity.resolveUserDataPath;
   yield* electronApp.setPath("userData", userDataPath);
   yield* logStartupInfo("runtime logging configured", { logDir: environment.logDir });
-  yield* desktopSettings.load;
 
   if (linuxElectronOptions !== null) {
     yield* logStartupInfo("linux password store configured", {
--- a/apps/desktop/src/settings/DesktopAppSettings.test.ts
+++ b/apps/desktop/src/settings/DesktopAppSettings.test.ts
@@ -26,6 +26,8 @@ const DesktopSettingsPatch = Schema.Struct({
   ),
   mainWindowMaximized: Schema.optionalKey(Schema.Boolean),
   serverExposureMode: Schema.optionalKey(Schema.Literals(["local-only", "network-accessible"])),
+  shellEnvironmentMode: Schema.optionalKey(Schema.String),
+  shellEnvironmentNames: Schema.optionalKey(Schema.Array(Schema.String)),
   tailscaleServeEnabled: Schema.optionalKey(Schema.Boolean),
   tailscaleServePort: Schema.optionalKey(Schema.Number),
   updateChannel: Schema.optionalKey(Schema.Literals(["latest", "nightly"])),
@@ -109,6 +111,8 @@ describe("DesktopSettings", () => {
         mainWindowBounds: null,
         mainWindowMaximized: false,
         serverExposureMode: "local-only",
+        shellEnvironmentMode: "allowlist",
+        shellEnvironmentNames: [],
         tailscaleServeEnabled: false,
         tailscaleServePort: 443,
         updateChannel: "nightly",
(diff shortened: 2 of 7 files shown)


JoeyEamigh:
this is a false positive. this works on macos 26, and as far as i can tell, has worked since macos 13 ventura. only reason to do anything about it would be to support monterey or older. which seems very old lol. if someone who is not a stochastic number generator wants to weigh in happy to change but not going to defensively change something for a 5 year old macos version that hit EoL two years ago.

PR #11298, opened 2026-09-11
fix(client): restore authenticated device stream tickets

shivamhwp:
The grouped HTTP client change in #11029 left device stream ticket requests on the old interface. This breaks the client-runtime typecheck and the bearer/DPoP ticket path used by the Device panel.

Select the `auth` group and call its `webSocketTicket` method directly, matching the other migrated request sites.

Client-runtime typecheck, 23 authenticated HTTP tests, scoped lint and formatting pass. Cookie access keeps its existing early return. Native device streams were not separately exercised.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket ticket requests by using the correct access method, helping device hub connections authenticate reliably.


juliusmarminge:
Superseded by #11304 — same `deviceHubAccess.ts` typecheck fix (`group: "auth"` + `client.webSocketTicket`) already landed on main.

Issue #5489, opened 2026-08-06
[Bug]: Go back intermittently fails with an opaque Git exit 128 during checkpoint restore


michael-dm:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

#### Intermittent behavior under normal use

1. Open a local Git repository in T3 Code.
2. Use a Codex thread for several turns so that multiple checkpoints exist.
3. Wait for the thread to become idle.
4. Use **Go back** on an earlier turn.
5. The restore intermittently fails with Git exit 128.

I observed five Git restore failures across two worktrees over two days. Four other restore attempts succeeded. At least one checkpoint target failed and then succeeded several minutes later, so the target ref itself was not permanently invalid.

#### Deterministic way to exercise the failure path

This reliably reproduces the same user-visible failure path, but I do **not** know whether it is the cause of the naturally occurring failures:

1. In a repository with a restorable T3 checkpoint, create an index lock:

   ```shell
   lock_path=$(git rev-parse --git-path index.lock)
   touch "$lock_path"
   ```

2. Click **Go back** in T3 Code.
3. Remove the test lock immediately afterward:

   ```shell
   rm "$lock_path"
   ```

T3 reports the generic exit-128 error below. Git itself reports an actionable “Unable to create .../index.lock: File exists” error for the restore command in this condition.

### Expected behavior

Go back should restore a valid checkpoint during normal idle use. If Git cannot restore it because of transient repository state, the error should identify the actionable reason rather than only reporting exit 128.

### Actual behavior

Go back sometimes fails immediately with:

```text
VCS process failed in GitVcsDriver.checkpoints.restoreCheckpoint: git (/Users/midm/Dev/one-click-os) exited with 128 - Process exited with a non-zero status.
```

The underlying Git stderr is not included, so it is not possible to tell from the report whether `git restore`, `git clean`, or `git reset` failed or why.

The failure happens before provider conversation rollback, so the thread and filesystem appear to remain at their pre-request state.

### Impact

Major degradation or frequent failure

### Version or commit

T3 Code Alpha 0.0.31

### Environment

macOS 26.4.1 (25E253), Apple Git 2.50.1, T3 Code desktop app, Codex app-server provider. Repository is a large monorepo with multiple linked worktrees.

### Logs or stack traces

```shell
[...]

PR #11340, opened 2026-09-11
fix(server): explain Git index locks when checkpoint restore fails

shivamhwp:
A checkpoint restore blocked by Git's index lock currently reports only exit 128. Recognize Git's standard index-lock error and explain that the user should wait for other Git operations and retry.

Keep the existing error schema and `command-failed` classification. Preserve stderr length/truncation metadata without retaining raw stderr. The change neither removes locks nor retries Git operations.

Refs #5489. This improves the deterministic lock failure's diagnosis; the cause of the reporter's intermittent failures remains unverified. Rebased onto main. Internal refinement remains pending after the final Fable retry failed.

Validation: 18 focused VCS process tests pass, including a real Git repository with a held index lock. Restore preserves the changed file and the lock, reports the actionable message, and succeeds after the test releases its own lock. The new native regression fails against the parent implementation. Windows-format lock output is covered, while permission failures and other locks retain their previous messages. Server typecheck and scoped lint pass. Existing error fields carry the message to web, desktop, mobile and remote clients; no native Windows/macOS run was performed.

Model: GPT-6. Harness: Codex.

Rebased on main `20363c32c9bfdbf49c2716ef11d1f18483fcc01b`. Scoped validation rerun after rebasing; existing native and human review limits remain.

## Summary by CodeRabbit

- **Bug Fixes**
  - Git operations now provide a clear, user-facing message when another Git process is using the repository index lock.
  - Restore operations remain protected while the lock is present and succeed once the lock is removed.
  - Unrelated lock and permission errors continue to retain their existing failure details.
  - Windows and Unix index-lock errors are handled consistently.


greatitself:
Independent validation of `04bd7d320465f8b93417fecb0b4ca1e3e0004d2a` (merge-base `050e2cafb144`).

Reproduced real index-lock failures in three layouts: a standard repository, `--separate-git-dir` with spaces in the metadata path, and a linked worktree. Passed Git’s captured failure through the actual VcsProcess classification service. The merge-base gave the generic exit message; this revision identified the index lock in all three cases. The lock and modified file remained untouched after failure; removing only the test-owned lock let Git restore the original file. No native Windows run was performed.

[Reproduction scripts and recorded results](https://github.com/greatitself/t3code/releases/download/validation-batch-20260912/validation-evidence.tar.gz) — `fourth.mjs`; the numbered metadata file pins this PR’s tested revisions.

Model: GPT-6. Harness: Codex.

PR #11343, opened 2026-09-11
feat(web): pin a skill as a composer mode with option+enter

matheustimbo:
Proposed first in Ideas: https://github.com/pingdotgg/t3code/discussions/11342. Opening the PR alongside it so there is something concrete to look at. Close it freely if the idea is not wanted; nothing here is load-bearing for anything else.

## The problem

A skill mention is a one-shot insert. `$review` goes into one message and is gone on the next, so keeping a skill in play across a stretch of a conversation means retyping the mention every turn, and the turn where you forget quietly behaves differently.

## What changed

Option+Enter on macOS, Alt+Enter elsewhere, on a highlighted skill row in either the `$` or the `/` picker pins that skill for the thread instead of inserting it once. The pinned skill leads the composer's control row as a chip, and every message goes out with `$<name>` in front of it until the chip is clicked.

| Before | After |
| --- | --- |
| before | after |

The highlighted skill row trades its source badge for the shortcut hint, so the row keeps one right-aligned element. Every other row is unchanged.

Interaction video, pin through send through clear: https://raw.githubusercontent.com/matheustimbo/t3code/pr-assets/composer-skill-mode/composer-skill-mode/skill-mode.mp4

## Why it should exist

The mode is fully expressed in the text it produces. `applyComposerSkillModePrefix` prepends `$<name>` to the prompt at send time, in `formatOutgoingPrompt`, next to the `Ultrathink:` prefix that already works this way. So there is no contract change, no new command, no event, no projector work, and the server, the provider, and every other client see the prefix because it is literally in `message.text`. The prefix is also visible in the thread exactly as sent, which is what makes the chip honest rather than a hidden setting.

Two guards on the prefix. It is idempotent, so a recalled prompt never takes the mention twice. It leaves a leading provider slash command alone, because prefixing one turns it into prose the provider never runs. `applyClaudePromptEffortPrefix` already carried that second rule, so rather than ship a second copy of a subtle regex the two prefixes now share `startsWithProviderSlashCommand`.

`onComposerCommandKey` was modifier-blind: Enter, Tab, Cmd+Enter, and Alt+Enter on a highlighted row all selected. Rather than add a fourth branch to a 5900-line component, the key table moved to `resolveComposerMenuKeyAction`, which is pure and has its own test. Existing highlight and select behavior is unchanged.

The pinned mode lives on the composer draft beside `runtimeMode` and `interactionMode`, so it is scoped per environment and thread and survives a reload. It is not user content, so it never gives an empty draft a sidebar row.

## Scope

Web and desktop. **Mobile is deliberately untouched.** It has no modifier keys and its picker is a separate implementation, so the touch affordance is a design question of its own rather than something to guess at here. The prefix helper is in `packages/shared` so mobile can adopt it unchanged if you want it there.

One rough edge left in on purpose: in the provider-unavailable composer state the control row is replaced by a single button, so a pinned chip is not in the DOM and cannot be cleared until a provider is back. Sending is blocked in that state anyway and the mode persists, so fixing it meant widening the diff for a state you cannot send from.

## Verification

`vp test run` over the six touched test files, 158 passing. `vp lint` and `vp run --filter @t3tools/web --filter @t3tools/shared typecheck` clean, no new warnings or errors.

Driven end to end against a real environment with a headless browser, nine checks: the hint shows only on the highlighted skill row and other rows keep their badge, pinning clears the trigger text and closes the menu, the chip survives a reload, clicking it clears the mode. The outgoing frame was captured at the WebSocket boundary and `message.text` read exactly `"$pstack:poteto-mode reply with exactly: OK"`.

The resting-strip width fix is its own commit: `restingComposerControlsMeasurement` sums three specific selectors rather than the cluster's children, so the chip had to be added to it or the strip measured narrower than it renders. Its test was confirmed non-tautological by reverting only the production change, which fails it.

Implemented with Claude Opus 5 through T3 Code and the Claude Code harness.

## Closes discussions

- https://github.com/pingdotgg/t3code/discussions/11342

## Summary by CodeRabbit

* **New Features**
  * Pin skills or provider commands as composer modes with Alt/Option+Enter, including on mobile.
  * Pinned modes appear as removable chips and automatically prefix messages in threads and plan follow-ups.
  * Modes persist between drafts and remain available when clearing sent message content.
  * Command menus show pinning hints for supported items.

* **Bug Fixes**
  * Prevented duplicate prefixes and preserved provider slash-command behavior.

* **Documentation**
  * Added instructions for pinning and removing composer modes.

### DIFF
--- a/apps/mobile/src/components/ComposerSkillModeChip.tsx
+++ b/apps/mobile/src/components/ComposerSkillModeChip.tsx
@@ -1,0 +1,45 @@
+import { composerSkillModeMention, type ComposerSkillMode } from "@t3tools/shared/composerTrigger";
+import { memo } from "react";
+import { Pressable, View } from "react-native";
+
+import { SymbolView } from "./AppSymbol";
+import { AppText as Text } from "./AppText";
+
+/**
+ * Sits above the editor rather than in the toolbar, which is hidden while the
+ * composer is collapsed. A pinned mode that vanishes on collapse would be a
+ * setting the user cannot see or undo.
+ */
+export const ComposerSkillModeChip = memo(function ComposerSkillModeChip(props: {
+  readonly skillMode: ComposerSkillMode;
+  readonly onRemove: () => void;
+}) {
+  return (
+    <View className="flex-row">
+      <Pressable
+        accessibilityRole="button"
+        accessibilityLabel={`Remove pinned mode ${props.skillMode.label}`}
+        accessibilityHint={`Every message starts with ${composerSkillModeMention(props.skillMode)}`}
+        onPress={props.onRemove}
+        hitSlop={6}
+        className="max-w-full flex-row items-center gap-1.5 rounded-lg bg-subtle px-2 py-1 active:opacity-60"
+      >
+        <SymbolView
+          name="cube"
+          size={12}
+          tintColorClassName="accent-icon-muted"
+          type="monochrome"
+        />
+        <Text className="shrink text-xs font-t3-medium text-foreground-muted" numberOfLines={1}>
+          {props.skillMode.label}
+        </Text>
+        <SymbolView
+          name="xmark"
+          size={10}
+          tintColorClassName="accent-icon-subtle"
+          type="monochrome"
+        />
+      </Pressable>
+    </View>
+  );
+});
--- a/apps/mobile/src/features/threads/ComposerCommandPopover.tsx
+++ b/apps/mobile/src/features/threads/ComposerCommandPopover.tsx
@@ -42,11 +42,26 @@ export type ComposerCommandItem =
       readonly description: string;
     };
 
+export type ComposerPinnableItem = Extract<
+  ComposerCommandItem,
+  { type: "skill" | "provider-slash-command" }
+>;
+
+/**
+ * Whether a row names a provider entry point a mode can repeat. Skills qualify,
+ * and so do provider slash commands, because Claude Code lists plugin skills
+ * only under `/`. The client's own `/model` and a path do not.
+ */
+export function isComposerPinnableItem(item: ComposerCommandItem): item is ComposerPinnableItem {
+  return item.type === "skill" || item.type === "provider-slash-command";
+}
+
 interface ComposerCommandPopoverProps {
   readonly items: ReadonlyArray<ComposerCommandItem>;
   readonly triggerKind: ComposerTriggerKind | null;
   readonly isLoading: boolean;
   readonly onSelect: (item: ComposerCommandItem) => void;
+  readonly onPinMode?: (item: ComposerPinnableItem) => void;
 }
 
 function PopoverSurface(props: { readonly children: React.ReactNode; readonly style?: ViewStyle }) {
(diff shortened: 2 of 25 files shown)


matheustimbo:
Pushed three commits that fix a real gap in this PR, found by running it against Claude rather than Codex.

The mode only pinned rows of type `skill`. Claude Code does not report plugin skills as skills: `pstack:poteto-mode` arrives as a provider slash command and the skill list comes back empty, so the `/` menu row was a `provider-slash-command` and the `$` menu was empty. `⌥⏎` fell through to the select branch and silently inserted the command text instead of pinning anything. As written, the feature did nothing at all for a Claude user.

| Before | After |
| --- | --- |
| before | after |

`ComposerSkillMode` now carries the kind it pinned and `composerSkillModeMention` derives the sigil once. A skill leads with `$name`, a provider slash command with `/name`, which is the only form that runs it. The hint and the pin moved from "is a skill row" to "is a provider entry point", so `/model`, `/plan`, and paths still select exactly as before.

Two things worth calling out. A slash-command mode and the `Ultrathink:` effort prefix cannot both lead a message and the command wins, since prefixing prose to a command leaves text the provider never runs; `applyClaudePromptEffortPrefix` already refused that case and this makes it reachable. And the draft field is still named `skillMode` even though it can hold a command, because renaming it costs a localStorage migration for no user-visible gain.

Verified with the nine-check headless run on Claude Opus 5 through the `/` picker, `message.text` captured at the WebSocket boundary reading exactly `"/pstack:poteto-mode reply with exactly: OK"`. 151 tests across the four touched files, typecheck and lint clean on the fork's base.

matheustimbo:
Mobile is no longer out of scope. The original PR called the touch affordance an open question, so here is an answer rather than a question.

Each pinnable row gets a **Mode** button, which claims the touch before the row's own press handler so tapping it pins instead of inserting. The pinned mode is a chip directly above the editor, not in the toolbar, because the toolbar is hidden while the composer is collapsed and a mode you cannot see is a setting you cannot undo.

| Mode button | Pinned | Sent |
| --- | --- | --- |
| pill | chip | sent |

The prefix is applied where the composer hands the message to the outbox, so the optimistic bubble shows exactly what was sent, and the checks that read raw draft text (`/usage-limits`, the Codex feedback command) still run before it.

**New Task is deliberately untouched.** Its draft does not survive into the thread it creates, so a mode pinned there would lead one message and then vanish. Carrying it across means widening the persisted outbox message and writing the mode into the new thread's draft on creation, which is a much larger change than the gap justifies. Say the word if you want it and I will do it properly rather than half.

`providerSkillComposerMode` and `providerSlashCommandComposerMode` moved into `packages/client-runtime`, and both surfaces build the mode there, so the chip label and the mention that ships cannot drift apart.

Driven end to end on an iOS simulator: the Mode button appears on the Claude command row, tapping it pins and clears the trigger text, the sent message read `/pstack:poteto-mode Reply with exactly: OK`, the agent answered, the chip stayed visible once the composer collapsed, and tapping it cleared the mode. Fork CI is green on the same commits.

Happy to drop these two commits back out if you would rather review the web fix on its own.

kridaydave:
@matheustimbo this is a solid PR. The mobile follow-up and the Claude-vs-Codex gap you found are exactly the kind of thing that only shows up by running it against another provider.

The blocker right now is that the branch no longer merges into main. I can rebase it for you if you want, or you can keep authorship and I can walk through the conflict resolution. A few things that may help if you would rather do it yourself: the composer menu and trigger files moved recently, so the conflict is probably in apps/web/src/components/chat/ComposerCommandMenu.tsx, apps/web/src/components/chat/composerMenuKeyAction.ts, and packages/shared/src/composerTrigger.ts. Resolve toward the new file, keep your Option+Enter key action and the pinned-mode chip, then run the composer trigger and menu tests.

PR #11350, opened 2026-09-12
chore(deps): upgrade Effect to rc.115 and Alchemy to its rc.115 preview build

juliusmarminge:
Bumps Effect from `4.0.0-rc.112` to `4.0.0-rc.115` (`effect`, `@effect/platform-node`, `@effect/platform-node-shared`, `@effect/platform-bun`, `@effect/sql-sqlite-bun`, `@effect/sql-pg`, `@effect/atom-react`, `@effect/openapi-generator`, `@effect/vitest`) and Alchemy from `2.0.0-beta.76` to `2.0.0-beta.77`. `@effect/platform-node-shared` stays pinned exactly in the catalog (#11208).

## What changed in Effect

rc.113 was a breaking naming cleanup plus a few module moves. Most of the diff is mechanical:

- `Config`, `Flag`, `Argument`, `Prompt` constructors are PascalCase (`Config.string` → `Config.String`, `Flag.integer` → `Flag.Int`, `Flag.choice` → `Flag.Literals`, `Prompt.confirm` → `Prompt.Confirm`, `Prompt.text` → `Prompt.String`).
- `SchemaTransformation.transformOrFail` / `SchemaGetter.transformOrFail` → `transformEffect`; `Config.mapOrFail` → `Config.mapEffect`.

Non-mechanical changes:

- **Mime** moved from `@effect/platform-node/Mime` (a `mime` re-export) to `effect/unstable/http/Mime`, which returns `Option` and bare extensions. `http.ts` and `imageMime.ts` adapt; `imageMime` re-adds the dot before checking the safe-extension list.
- **Socket** dropped `runRaw`; `DeviceHubProxy` now pumps `reader.pull` batches into `writer.writeAll` for each direction. A close fails the pull, which still ends the race and tears down the other side. `WebSocketConstructor` now takes `WebSocketConstructorOptions`; the test constructors narrow it to `protocols`, which is all `Socket.makeWebSocket` passes.
- **HTTP server addresses** are `NetAddress.SocketAddress` instead of `{ _tag: "TcpAddress", hostname }`. `McpSessionRegistry` builds the MCP endpoint from the bound IP with `NetAddress.isUnspecified` / `formatIp` / `formatUrlHostString`; the test cases now cover `::` and `::1`. A hostname can no longer appear there, so the `localhost` case is gone.
- **File sizes** are `ByteSize` and `readAlloc` takes a `number`: `FileSystem.Size(n)` → `ByteSize.bytes(n)` in `GitManager` and `CursorSkills`, `http.ts` converts the chunk size.
- **JSON Schema** generation defaults to open objects (`onExcessProperty: "ignore"`). The `t3.json` schema and the text-generation output schemas pass `onExcessProperty: "error"` to keep `additionalProperties: false`; the MCP preview tool result schemas go through `Tool.getJsonSchema`, which has no such option, so those two tests now expect `additionalProperties: true`.
- `AgentSessionScanner.test` used an mtime of `now + 1ms`; the Node `BigInt` stat that the Effect file system now uses floors sub-millisecond precision and rounded it back to `now`, so the fixture uses `now + 1s`.
- Alchemy beta.77 moved `Cli` to `alchemy/Report`, renamed `ProfileLive` to `ProfileStoreLive`, and its auth providers now require an `Interaction` service; `infra/relay/scripts/deploy.ts` provides `layerNonInteractive()` and `LoggingCli` gets `PlatformServices` for its prompt environment.
- The vendored `effect` patch was regenerated against rc.115 (three `RpcClient` pinger hunks moved because `write` became `writer.write`); the `@effect/vitest` patch was regenerated for the new import sites.

## msgpackr-extract goes

Effect rc.113 dropped its `msgpackr` dependency; the `Msgpack` encoding module is gone and RPC serialization is JSON / NDJSON / JSON-RPC / SchemaBinary. The server only uses `RpcSerialization.layerJson` / `layerNdjson`, so `msgpackr-extract` (3.5 MB of native addon per platform) had no consumer left. It is removed from `apps/server/package.json`, the pnpm `allowBuilds` list, `CLI_RUNTIME_EXTERNAL_PREFIXES`, the WSL archive exclusions in `build-desktop-artifact.ts`, the knip ignore list, and `docs/operations/release.md`. `detect-libc` leaves the external list with it (it was only in the closure through `node-gyp-build-optional-packages`); `node-gyp-build` and `node-addon-api` stay because `node-pty` and the `ws` accelerators still need them. The external-closure test now anchors on `node-pty` and `node-addon-api`.

## Not upgraded

**Alchemy is pinned to a pkg.ing preview build** ([`2365753e58`](https://github.com/alchemy-run/alchemy-effect/commit/2365753e583e15d1476c28055f6b12d3adfdd777), the head of [alchemy-effect#1601](https://github.com/alchemy-run/alchemy-effect/pull/1601), "effect 4.0.0-rc.115") rather than `2.0.0-beta.77`. beta.77 on npm still calls the rc.112 constructor names at module scope (`Config.string` in `Auth/Profile.js`) and cannot load under rc.115; upstream fixed that on main after the release. Consequences, all meant to be reverted once Alchemy ships beta.78:

- `blockExoticSubdeps: false` in `pnpm-workspace.yaml`: the preview's own `@alchemy.run/*` and `@distilled.cloud/*` dependencies are pkg.ing tarballs, and pnpm has no per-package allowlist for URL subdependencies.
- The preview still pins the released `@distilled.cloud/fly-io@1.0.0-rc.9` and `railway@1.0.0-rc.9`, which carry the same rc.112 call sites (two and four lines); both get a pnpm patch renaming them, and an `overrides` entry points their `@distilled.cloud/core` at the preview core so a single rc.115-compatible copy loads.
- `scripts/sync-reference-repos.ts` resolves a `https://pkg.ing/<name>/<sha>` pin to that commit, so `vpr sync:repos` keeps the vendored Alchemy reference in step (#11351).

With that, `infra/relay` tests pass (30 files, 284 tests) and `scripts/deploy.ts` loads and prints its help. The relay deploy itself has not been run.

## Verified

- `tsc --noEmit` clean in `apps/server`, `apps/desktop`, `apps/web`, `apps/mobile`, `packages/*`, `scripts`, `infra/relay`.
- `vp test run` in `scripts` (285), `packages/shared` (664), `packages/contracts`, `packages/client-runtime` (1393), `packages/ssh`, `packages/effect-acp`, `packages/effect-codex-app-server`, `packages/tailscale`, `apps/web` (4710), `apps/desktop` (1291; the libsecret helper test needs `libsecret-1` headers this box lacks), and the `apps/server` suites for `src/persistence`, `src/mcp`, `src/device`, `src/http`, `src/server.test.ts`, `src/cli`, `src/bin.test.ts`, `src/project`, `src/git`, `src/provider/Drivers/CursorSkills`, `src/textGeneration` (one pre-existing failure on this host: the Codex fake CLI cannot find `node` when spawned with an explicit env, reproduces on `main`).
- `vp run build:bundle` in `apps/server`: `grep -c msgpackr dist/bin.mjs` is 0, the inlined-externals scan reports none, `node dist/bin.mjs --help` runs.
- `vp run knip:check` clean.

Claude Fable 5 via Claude Code.

## Summary by CodeRabbit

* **New Features**
  * Added MCP session termination with validated DELETE responses.
  * Added RPC request and connection lifecycle hooks, including streaming and ping events.
  * Improved WebSocket proxy data transfer and shutdown handling.

* **Improvements**
  * Improved IPv4 and IPv6 endpoint handling for MCP services.
  * Enhanced MIME detection and static file handling.
  * Structured-output schemas now reject unexpected properties.

* **Maintenance**
  * Updated command-line configuration handling and removed obsolete native dependency packaging.
  * Updated Windows package validation documentation.
  * Deployment commands now fail clearly instead of hanging on unexpected CI prompts.


juliusmarminge:
Superseded by #12326: Alchemy 2.0.0-beta.78 is built against Effect rc.115, so the pkg.ing preview pins and provider patches are no longer needed.

PR #11357, opened 2026-09-12
fix(web): keep comment actions visible when pr comments are folded

maria-rcks:
Folding the PR conversation hid the comment composer and its close/reopen actions. Keep the existing composer outside the fold so actions and drafts remain available, and reduce the description's bottom padding by 12px.

Verified in the real GitHub-backed web client: drafts survive folding/unfolding, Comment and Close with comment remain enabled, and a comment posted successfully while folded. Close/reopen mutations were not exercised. Web typecheck and scoped lint passed (one existing array-index-key warning); all GitHub checks passed. This shared web panel also serves desktop; native mobile is unchanged.

Before:

folded comments previously hid the composer

After:

folded comments keep the composer and tighter description spacing

folding comments and posting through the retained composer

model: `gpt-6-astra`. harness: codex.

### DIFF
--- a/apps/web/src/components/pullRequest/PullRequestSummaryTab.tsx
+++ b/apps/web/src/components/pullRequest/PullRequestSummaryTab.tsx
@@ -363,7 +363,7 @@ function CommentComposer({
   };
 
   return (
-    <div className="mt-3 space-y-2">
+    <div className="space-y-2">
       <Textarea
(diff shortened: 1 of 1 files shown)


maria-rcks:
> [!NOTE]
> Written by `gpt-6-astra` on behalf of Maria

verified in the dev app: the comment draft and actions remain available while comments are folded.

PR #11388, opened 2026-09-12
perf(shared): reuse validated Windows executable lookups

yashranaway:
Every Windows process launch walks PATH × PATHEXT again, including repeated `git` and provider commands within one operation.

Cache successful executable lookups for one second, keyed by command, platform, PATH, PATHEXT, and working directory. Each hit still checks the chosen file; removing it immediately triggers a fresh scan. Explicit paths and missing commands bypass caching, and the cache is bounded to 512 entries. A new higher-priority executable is discovered after the short TTL.

Addresses the repeated PATH scans in #11221. Git launcher selection is outside this change.

Validation: 36 shared shell tests pass. The regression reduces a repeated fixture lookup from six stat calls to one and covers removed executables, environment changes, expiry, and immediate discovery after a miss. It fails against the original resolver. Shared-package typecheck and targeted lint pass; Windows wall-clock performance was not measured locally.

Model: GPT-6
Harness: Codex in T3 Code

## Summary by CodeRabbit

* **Performance**
  * Improved Windows command resolution by caching recent executable lookups.
  * Revalidates cached results and refreshes them when executables change or cache entries expire.
  * Limits cached lookup data to prevent unbounded growth.

* **Bug Fixes**
  * Ensures command resolution remains accurate when PATH or PATHEXT settings differ.
  * Falls back to `.cmd` commands when a previously detected executable is no longer available.

* **Tests**
  * Added coverage for Windows-specific command resolution and cache behavior.

PR #11393, opened 2026-09-12
fix(server): resolve Windows tools in the child working directory

yashranaway:
Windows executable discovery currently resolves relative commands and relative PATH entries from the server directory, even when the child will run elsewhere. A project-local command such as `.\bin\tool` can therefore miss its `.cmd` shim or resolve to another project's executable.

Pass the execution directory into spawn resolution and resolve candidates from that directory. Propagate it through provider sessions, text generation, triage, publishing, and the shared process runner; `spawnCwd` takes precedence over the logical working directory. Existing absolute paths and POSIX launch behavior remain supported.

Validation: 36 shared shell tests and 49 server tests pass, with failing-before regressions for relative commands and two different project PATHs. Shared/server typechecks and targeted lint pass. Windows path and shim behavior is tested with controlled filesystem inputs; no Windows runtime was available.

Merge note: #11388 changes the same executable resolver. When combining these PRs, its cache key must use the child working directory introduced here. The two-project PATH regression checks that distinction.

Model: GPT-6
Harness: Codex in T3 Code

### DIFF
--- a/apps/server/scripts/cli.ts
+++ b/apps/server/scripts/cli.ts
@@ -278,7 +278,9 @@ const publishCmd = Command.make(
             yield* Effect.log("[cli] Applied package metadata and publish icon overrides");
 
             const args = createVpPmPublishArgs(config);
-            const spawnCommand = yield* resolveSpawnCommand("vp", ["pm", ...args]);
+            const spawnCommand = yield* resolveSpawnCommand("vp", ["pm", ...args], {
+              cwd: repoRoot,
+            });
 
             yield* Effect.log(`[cli] Running: vp pm ${args.join(" ")}`);
             yield* runCommand(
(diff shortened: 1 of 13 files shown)


greatitself:
Independent validation of `deea9d047a072fa38260c163ae8ea3ac3ba6c6e2` (merge-base `ca6416ec2d1d`).

Checked 27 lookup combinations: a child directory on C:, one on D:, and a UNC share; bare, backslash-relative, and slash-relative commands; and EXE/CMD/BAT targets. All resolved against the child directory on this revision, including a quoted relative PATH entry and a directory containing spaces; all 27 differed at the merge-base. Shell selection also matched the executable extension. This uses the actual resolver with a simulated Windows filesystem, not a Windows process launch.

[Reproduction scripts and recorded results](https://github.com/greatitself/t3code/releases/download/validation-batch-20260912/validation-evidence.tar.gz) — `third.mjs`; the numbered metadata file pins this PR’s tested revisions.

Model: GPT-6. Harness: Codex.

PR #11407, opened 2026-09-12
fix(server): avoid extra round trips for terminal output

Bil0000:
Remote terminal output waits for a client acknowledgement after every RPC stream chunk. That adds a stop-and-wait round trip even when the shell has already echoed the next key.

Allow terminal streams to keep up to eight chunks in flight, pausing once pending output reaches 64 KiB. Keep the existing acknowledgement messages and leave other RPC streams unchanged. This removes the extra wait for interactive output; network travel time still applies.

Verified with 86 terminal tests, five WebSocket terminal route tests, server typecheck, lint, formatting, and a real shell reproduction. Before the fix, the shell history contained `AB` while the client received only `A` until its acknowledgement was released. With the fix, the client received `AB` before that acknowledgement. Browser typing was checked against the updated server.

Model: GPT-6. Harness: Codex.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminal output streaming over WebSocket connections to prevent excessive buffering.
  - Added backpressure handling so delivery pauses when the pending output window is full and resumes automatically as capacity becomes available.
  - Improved reliability for large or rapid terminal output by keeping buffered data within safe limits.
  - Terminal streams now close cleanly when interrupted, ensuring sessions do not remain open unexpectedly.

### DIFF
--- a/apps/server/src/terminal/OutputProtocol.test.ts
+++ b/apps/server/src/terminal/OutputProtocol.test.ts
@@ -1,0 +1,79 @@
+import { assert, describe, it } from "@effect/vitest";
+import { WS_METHODS } from "@t3tools/contracts";
+import * as Deferred from "effect/Deferred";
+import * as Effect from "effect/Effect";
+import * as Queue from "effect/Queue";
+import * as Schema from "effect/Schema";
+import * as Stream from "effect/Stream";
+import * as TestClock from "effect/testing/TestClock";
+import { Rpc, RpcGroup, RpcMessage, RpcSerialization, RpcServer } from "effect/unstable/rpc";
+
+import { withTerminalOutputWindow } from "./OutputProtocol.ts";
+
+describe("terminal output window", () => {
+  for (const { tag, size, limit } of [
+    { tag: WS_METHODS.terminalAttach, size: 1, limit: 8 },
+    { tag: WS_METHODS.subscribeTerminalEvents, size: 1, limit: 8 },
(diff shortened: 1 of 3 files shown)


Bil0000:
The docstring coverage warning is left unchanged because the task instructions prohibit added docstrings. The focused protocol tests cover delivery order, both terminal stream methods, the pending-output limits, acknowledgement handling, and interruption.

Macroscope found no correctness issue. Its recommendation for human review remains for a maintainer to assess.

greatitself:
Independent validation of `5fa7e819c1d3b24d1da35f04e5925f963507d587` (merge-base `b1e223e2b0d8`).

Exercised the protocol wrapper with two clients sharing a request ID, 20 full-window acknowledgement/refill cycles, Exit followed by request-ID reuse, and multibyte output (three chunks of 6,000 emoji). Client windows remained isolated, each full-window acknowledgement released one refill, reuse started with a fresh window, and UTF-8 byte size triggered the limit. Metadata streams retained ordinary acknowledgement handling. These are wrapper-level checks with controlled protocol callbacks; they do not measure end-to-end throughput.

[Reproduction scripts and recorded results](https://github.com/greatitself/t3code/releases/download/validation-batch-20260912/validation-evidence.tar.gz) — `fourth.mjs`; the numbered metadata file pins this PR’s tested revisions.

Model: GPT-6. Harness: Codex.

PR #11415, opened 2026-09-12
test(web): pin pull-request stats delta on paginated scroll

Adamulek123:
## Summary

The PR stats path used to refetch the whole visible set on every scroll page, re-requesting up to hundreds of already-counted rows. That bug has since been fixed on main (entered-only candidates plus missing-only batch keys), but nothing pinned the behavior.

## What changed

Test-only change. Adds isolated regression tests proving a paginated scroll from 99 cached rows to 100 requests only row 100, via the stats cache and via the in-flight request set independently.

## Validation

- pullRequestList.logic.test.ts: 120 passed (2 new isolated delta cases)
- No product code touched

## Measured impact

Compared with main-latest.json from main commit b1e223e2b, using the focused stats-delta suite:

| Scroll: 99 cached + row 100 | Main | This PR | Change |
| --- | ---: | ---: | ---: |
| Refs requested | 1 (delta) | 1 (delta) | unchanged, now pinned |
| Focused tests | 14 pass | 16 pass | +2 regression guards |

The delta behavior already on main is now locked by two isolated tests (cache-filtered and in-flight-filtered). Every enforced ceiling passes. Coverage: SPAWN-ASSERTED.

Built with Muse Spark (opencode/muse-spark-1.3) via implement and audit subagent loops with strict branch-audit reviews.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for pull request list scrolling and line-count statistics.
  * Verified that previously cached or in-flight rows are not requested again when a new row becomes visible.
  * Confirmed consistent request behavior across eager and viewport-based loading policies, including paginated scrolling.

### DIFF
--- a/apps/web/src/components/pullRequest/pullRequestList.logic.test.ts
+++ b/apps/web/src/components/pullRequest/pullRequestList.logic.test.ts
@@ -293,6 +293,94 @@ describe("visible pull request line-count targets", () => {
       }),
     ).toEqual([]);
   });
+
+  it("requests only the newly appended row after a paginated scroll when earlier rows are cached", () => {
+    const loaded = Array.from({ length: 99 }, (_, index) =>
+      entry({ additions: 0, deletions: 0, number: index + 1 }),
+    );
+    const cachedStats = mergePullRequestDiffStats(
+      new Map(),
+      loaded.map((item) => ({
+        environmentId: ENV_1,
+        projectId: item.projectId,
+        number: item.number,
+        additions: 4,
+        deletions: 2,
+      })),
+    );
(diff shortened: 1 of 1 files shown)


Adamulek123:
## Perf proof (pr-load-perf harness, baseline b1e223e2b)

- Branch 355a1cfe8: harness PASS; scenario prperf-5 16/16 (baseline 14).
- Asserted: paginated scroll 99 cached plus row 100 requests only [100]; per-env batches within 500 limit; visible retain covered.
- Coverage: SPAWN-ASSERTED. No duration regression in own scenario.

PR #11417, opened 2026-09-12
test(web): guard against PR diff warm-up regression

Adamulek123:
## Summary

The detail panel used to fetch the full diff (up to 8 MB) on every PR open even when the Code tab was never visited. That was already removed by earlier work, but nothing prevented it from coming back.

## What changed

No product change. Adds a source tripwire test asserting the detail panel never constructs the diff query, plus an intent comment at the former warm-up site. The test deliberately covers the same-API warm-up under any variable or call shape; tab gating and hover preload are documented as out of scope for this guard.

## Validation

- pullRequestDetailLazyDiff.test.ts: 1 passed
- pullRequestDetail.logic and pullRequestDiff.logic suites: 101 passed

## Measured impact

Compared with main-latest.json from main commit b1e223e2b, using the focused lazy-diff suite:

| Summary-only PR open | Main | This PR | Change |
| --- | ---: | ---: | ---: |
| Diff bytes fetched | up to 8 MB | 0 | -100% (guarded) |
| Focused tests | 0 (absent) | 1 pass | +1 tripwire |

No product change; the guard locks in the earlier removal. Coverage: GUARD-ONLY.

Built with Muse Spark (opencode/muse-spark-1.3) via implement and audit subagent loops with strict branch-audit reviews.

## Summary by CodeRabbit

* **Performance**
  * Pull request summary views continue to avoid loading large code diffs until the Code tab is opened, helping preserve faster initial loading.

* **Tests**
  * Added safeguards to ensure summary-only views do not preload diff data.
  * Verified that diff information remains loaded on demand when accessing the Code tab.

### DIFF
--- a/apps/web/src/components/pullRequest/PullRequestDetailPanel.tsx
+++ b/apps/web/src/components/pullRequest/PullRequestDetailPanel.tsx
@@ -622,6 +622,9 @@ export function PullRequestDetailPanel({
   const activityQuery = useEnvironmentQuery(
     pullRequestEnvironment.activity({ environmentId, input: reference }),
   );
+  // The diff is intentionally not warmed up here. It is a full `gh pr diff` (up to 8MB) and
+  // most opens never visit the Code tab, so it loads only when that tab mounts. The module
+  // chunk alone preloads on Code hover/focus above.
   const turnRefresh = usePullRequestTurnRefresh(environmentId);
   const [cachedDetail, setCachedDetail] = useState(() =>
     readPullRequestDetailSnapshot(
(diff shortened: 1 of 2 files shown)


Adamulek123:
## Perf proof (pr-load-perf harness, baseline b1e223e2b)

- Branch 7bff884e8: harness PASS; scenario prperf-9 NEW 1/1 (absent on main).
- Asserted: panel source contains neither pullRequestEnvironment.diff nor _diffWarmUpQuery, so Summary-only opens fetch 0 diff bytes (up to 8MB guarded).
- Coverage: GUARD-ONLY. No duration regression.

PR #11426, opened 2026-09-12
fix(mobile): pin expo-audio so the release smoke patch stays in use

ipanasenko:
## Problem

The Release Smoke job fails on every PR (for example [#8419](https://github.com/pingdotgg/t3code/actions/runs/34698145445/job/103565088499?pr=8419)):

```
[ERR_PNPM_UNUSED_PATCH] The following patches were not used: expo-audio@57.0.4
Error: Command failed: vp install --lockfile-only --ignore-scripts
```

`scripts/release-smoke.ts` deletes `pnpm-lock.yaml` and re-resolves from scratch. expo-audio 57.0.5 was published on 2026-09-11, so the `~57.0.4` range now resolves to 57.0.5 and the `expo-audio@57.0.4` patch in `pnpm-workspace.yaml` is unused. Main CI was green earlier the same day because the release was not out yet.

## Fix

Pin `expo-audio` to the exact `57.0.4` in `apps/mobile/package.json`, matching how `expo-sharing` is already pinned for its patch. The lockfile specifier is updated to match; the resolved version and patch hash are unchanged.

Note: `@react-native-menu/menu` is `^2.0.0` with a `2.0.0` patch and has the same latent risk, left as is to keep this PR to one concern.

Fable 5.1 via Claude Code in T3 Code

Closes #11437

## Summary by CodeRabbit

- **Chores**
  - Standardized the mobile app’s audio package version to ensure consistent installations.

PR #11429, opened 2026-09-12
fix(web): preserve snapshot preview size in sent messages

Bil0000:
Snapshot previews in sent messages were squeezed into one column of the two-column attachment grid, clipping the image and its app labels.

Let snapshots span both columns so they keep the composer's existing 208 × 112 frame and details. Remove the old test assertion that required the squeezed layout. Regular image and video thumbnails keep their existing layout.

Includes the two-line release-check prerequisite from #11426 by @ipanasenko in a separate commit. It pins the already-locked `expo-audio` version so fresh lockfile resolution uses its patch. Release Smoke passes locally with this pin; the prerequisite will leave this PR diff when #11426 lands on main.

Verified with 55 focused tests, web typecheck, targeted lint and formatting, and a real browser check at 1200, 640, and 390 px widths. The browser check failed before the fix and passed after it. The test setup timeout was raised locally for slow module loading and restored before commit.

Before:
Before: sent snapshot is squeezed while the composer preview is full width

After:
After: sent snapshot matches the composer preview

Model: GPT-6. Harness: Codex.

## Summary by CodeRabbit

- **Bug Fixes**
  - Snapshot attachments in user image grids now span both columns, improving layout and visibility.
- **Tests**
  - Updated coverage to reflect the revised attachment layout.
- **Mobile**
  - Standardized the app’s audio component version for more consistent builds and behavior.

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -75,7 +75,7 @@
     "effect": "catalog:",
     "expo": "~57.0.18",
     "expo-asset": "~57.0.15",
-    "expo-audio": "~57.0.4",
+    "expo-audio": "57.0.4",
     "expo-auth-session": "~57.0.10",
     "expo-blur": "~57.0.2",
     "expo-build-properties": "~57.0.15",
--- a/apps/web/src/components/Sidebar.logic.test.ts
+++ b/apps/web/src/components/Sidebar.logic.test.ts
@@ -828,30 +828,26 @@ describe("filterSidebarProjectScopeItems", () => {
     { value: "alpha", label: "Alpha workspace" },
     { value: "beta", label: "Beta tools" },
   ] as const;
-  const filter = (activeScopeKey: string | null, query: string) =>
+  const filter = (query: string) =>
     filterSidebarProjectScopeItems({
       items,
-      activeScopeKey,
       query,
       matches: (item, candidate) =>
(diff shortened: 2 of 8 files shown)


Bil0000:
The Release Smoke blocker is resolved at b908996a73b6fefefb085bd1b2a872955927ef8a. The branch includes the two-line dependency pin from #11426 by @ipanasenko in a separate commit, with original author credit. Release Smoke passed locally and in CI. All current checks are green or intentionally skipped. Macroscope approved the latest commit; CodeRabbit withdrew its static-markup test suggestion after reviewing the browser evidence and repository test guidance.

Bil0000:
@coderabbitai I checked the outside-diff suggestion for MessagesTimeline.test.tsx. The missing class assertion is real, but I am declining that test approach because this repository's verification guidance says: "Do not render components to static markup to assert props or attributes."

The regression was checked in the real client with Chromium: before the fix the sent image measured 99 px wide versus 206 px in the composer; after the fix both measured 206 × 110 inside their borders at viewport widths of 1200, 640, and 390 px. Image expansion and accessibility details passed as well. The PR includes before-and-after evidence.

A class-string assertion would test the implementation rather than the browser's grid sizing. No new function or API was added, so the generic docstring-coverage warning does not need a code change either.

PR #11456, opened 2026-09-12
fix(connections): avoid provider refresh storms on reconnect

Bil0000:
Each server-config subscription started a full provider refresh, so reconnects and extra tabs repeated provider probes. The server change removes that redundant refresh and keeps cached snapshots and provider-owned live updates.

Julius split this exact server patch into #11811, which merged at 7931227977. This PR is now superseded by that merged fix.

Following the review from Julius, commit 0237d715c0 removes the proposed 45-second setup timeout and restores the existing 15-second timeout and its regression tests. The removed slow-setup tests used artificial delays; they did not establish a real connection that required a longer deadline. Connection readiness follows WebSocket connection, not shell sync.

Verified: 58 connection supervisor/RPC session tests and 8 server-config subscription tests pass. Client-runtime typecheck, scoped lint, formatting, and Ponytail review pass. A merge preview against main at 7931227977 produces the exact same tree as main, with no conflicts or new changes.

There is no remaining change to land from this PR. No new PR or stack was created.

Model: GPT-6. Harness: Codex.

### DIFF
--- a/apps/server/src/server.test.ts
+++ b/apps/server/src/server.test.ts
@@ -6200,48 +6200,26 @@ it.layer(NodeServices.layer)("server router seam", (it) => {
     }).pipe(Effect.provide(NodeHttpServer.layerTest)),
   );
 
-  it.effect("refreshes providers for each subscribeServerConfig connection", () =>
+  it.effect("serves config on reconnect without starting provider probes", () =>
     Effect.gen(function* () {
-      const refreshCalls = yield* Ref.make(0);
-      const firstRefreshDone = yield* Deferred.make<void>();
-      const secondRefreshDone = yield* Deferred.make<void>();
-
+      const refresh = vi.fn(() => Effect.never);
       yield* buildAppUnderTest({
-        layers: {
-          providerRegistry: {
-            refresh: () =>
-              Ref.updateAndGet(refreshCalls, (count) => count + 1).pipe(
-                Effect.tap((count) =>
-                  Deferred.succeed(
-                    count === 1 ? firstRefreshDone : secondRefreshDone,
-                    undefined,
-                  ).pipe(Effect.ignore),
-                ),
-                Effect.as([]),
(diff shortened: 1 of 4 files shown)


juliusmarminge:
Audited `fbdda7f`. The server half is right and I split it out so it can land on its own: #11811 carries the `ws.ts` and `server.test.ts` hunks with you as commit author. Managed providers already run their own startup probe and periodic refresh, and the subscription stream seeds current providers before live changes, so the per-subscribe refresh bought nothing.

That leaves this PR with the `CONNECTION_ESTABLISHMENT_TIMEOUT` bump from 15s to 45s, and I don't think it should land as is. That timeout covers `resolver.prepare` → socket open → `session.ready`, and `ready` is just the WebSocket `onConnect` (`rpc/session.ts:173`), not the shell sync. Fifteen seconds for DNS + TLS + WS open is already generous. When it is exceeded the cause is almost always a dead relay or tunnel, and after this change every user on a dead T3 Connect link stares at "connecting" for 45s per attempt before the first retry, on web, desktop, and mobile. Your own description says the storm was what pushed healthy setups past the deadline, and #11811 removes the storm.

If there is still a reproducible case where a healthy connection needs more than 15s after #11811, please post it here with timings and we can talk about the number. Otherwise I'd close this once #11811 merges.

Bil0000:
Confirmed your review. Removed the 45-second timeout proposal in 0237d715c0 and restored the original 15-second deadline and tests. The slow-setup cases only injected delays; they were not evidence of a healthy connection needing more than 15 seconds.

I verified that #11811 contains the identical server patch and is merged. A merge preview of this corrected branch against main produces the same tree as main, so this PR is superseded and has nothing further to land.

Validation: 58 supervisor/RPC session tests plus 8 server-config tests pass. Client-runtime typecheck, scoped lint, formatting, and Ponytail review pass.

PR #11462, opened 2026-09-12
fix(auth): explain expired environment credentials instead of offering a dead reconnect

gabrielsayers1:
## Problem

A paired client's session lasts 30 days (`DEFAULT_SESSION_TTL`) and there is no renewal path, so every remote environment eventually hits expiry. When it does, the server knows exactly what happened — `SessionStore.verify` fails with `SessionTokenExpiredError` — but the wire error flattens that into `invalid_credential`. The client then shows a generic red "`<label>` is offline" banner whose **Reconnect button re-sends the same dead credential forever**.

The only visible difference from an ordinary network outage is the banner's colour, because `ChatView` renders the same title for supervisor phase `offline` and phase `blocked`. Diagnosing a real instance of this took reading server-side trace logs to discover the word "expired" had been thrown away.

## Fix

Expiry now survives the trip to the client, and both clients act on it.

- **Contracts** — `EnvironmentAuthInvalidError` carries an optional `credentialExpired` flag. It is an optional key sent only when true, following the existing `dpopFailureReason` precedent, so older clients keep decoding the payload and fall back to `reason`. Adding a new `reason` literal instead would have broken strict decoding on older clients.
- **Server** — `serverAuthCredentialExpired` classifies the underlying cause, covering session *and* websocket expiry, and every `failEnvironmentAuthInvalid` call site passes it.
- **client-runtime** — `presentConnectionState` exposes `blockedReason`, and a shared `connectionNeedsPairing` helper answers the question both surfaces were getting wrong: retrying cannot clear an authentication block. A DPoP proof hint is now suppressed for expiry, where it would misdirect.
- **Web** — banner becomes "`<label>` needs pairing again" with a key icon, and the Reconnect button is gone. Connections remains as the way forward.
- **Mobile** — same title change, and it stops telling the user it "will keep retrying automatically", which was actively false.

Desktop inherits the web change. The Connections panel improves for free via `connectionStatusText`.

## Testing

`vp test run` across the touched scopes: 233 passing. Typecheck clean on contracts, client-runtime, server, web, and mobile; targeted lint and format clean.

Tests were written first and watched fail — server classification (expired vs. the revoked near-miss), the client message mapping, `blockedReason` exposure, and `connectionNeedsPairing`.

## ⚠️ Outstanding

**Before/after screenshots are not attached**, which this repo requires for UI changes. Capturing them means running the app in a browser, which needs maintainer sign-off — happy to produce them on request before merge.

## Deliberately out of scope

- **No session renewal.** There is no `rotate`/`renew`/`refresh` anywhere in server auth today; adding one is a new auth-protocol feature, not a bug fix. This PR makes expiry legible and recoverable, it does not stop the 30-day logout. Worth a follow-up, ideally with a warning *before* expiry while the client can still act.
- **`retryNow` does not re-sample connectivity** (`supervisor.ts`), so Reconnect also silently fails while `navigator.onLine` is latched off after sleep. Real, separate concern, separate PR.

Model: Claude Opus 5 (1M context). Harness: Claude Code.


gabrielsayers1:
Opened against the wrong repository by mistake — this targets the goodbirdhq/phoenix fork, not upstream. Reopening there. Apologies for the noise.

PR #11464, opened 2026-09-12
Fixes #9

fzoll:
## Summary

Adds a measured `capabilities.browser` field to the public `/.well-known/t3/environment` descriptor so cc_runner's capability-based dispatch can demote nodes whose declared config has drifted from reality, per #9.

- **`ok`**: a Chromium binary was found (Playwright's downloaded browser cache is checked first, falling back to a well-known system install path) and it launched successfully — reports `binaryPath` + `version`.
- **`degraded`**: a binary was found but either it failed to launch, or only a system Chromium exists while the Playwright browser cache is missing/incomplete — reports `binaryPath` + `details`.
- **`absent`**: no Chromium binary could be found anywhere.

The probe (`apps/server/src/environment/BrowserCapabilityProbe.ts`) is lazy and cached for 6 hours via `Effect.cachedWithTTL`, so it never runs on every health-check poll. `ExecutionEnvironmentCapabilities.browser` is an additive optional field in `packages/contracts/src/environment.ts`; existing web/desktop clients tolerate it.

Detection is filesystem + process based only (directory scan of the Playwright cache, well-known system paths, then a `--version` spawn) — no `playwright-core` dependency was added to `apps/server`. Installation/telepítés-kezelés and cc_runner's dispatch/demote logic stay out of scope per the issue.

## Decisions

- **Test tier**: unit, focused on `apps/server`, with mocked `FileSystem`/`ProcessRunner` (matches the issue's `## Test` note). First commit (`c2550a5c`) adds `BrowserCapabilityProbe.test.ts` alone and is red (imports a module that doesn't exist yet); the second commit adds the implementation and wiring.
- **"degraded" semantics**: chose to treat *both* "binary found but failed to launch" and "only a system Chromium exists, no Playwright cache" as `degraded`, since either way the automation stack that actually needs to drive the browser (Playwright) can't be trusted yet — matches the issue's example (RPi has both apt chromium *and* the ms-playwright cache → `ok`).
- **Cache key/TTL**: a single process-lifetime `Effect.cachedWithTTL(..., "6 hours")` per server instance, as specified.

## Manually verified

`scripts/agent-verify.sh` does not exist in this repo, so per instructions I ran `pnpm lint` and the relevant test suites directly.

```
$ pnpm --filter @t3tools/contracts typecheck
$ tsgo --noEmit
(clean, no output)

$ cd apps/server && pnpm exec vp run typecheck
~/apps/server$ tsgo --noEmit ⊘ cache disabled
(clean, no output)

$ cd apps/server && pnpm exec vp test run src/environment/ src/server.test.ts
 Test Files  5 passed (5)
      Tests  132 passed (132)

$ pnpm lint
(only pre-existing warnings in apps/web, unrelated to this diff; exit 0)
```

### Pre-existing, unrelated flake found while running the full `pnpm test`

Running `pnpm test` at the repo root (which fans out to every workspace package in parallel) reproduces one failing test twice in a row:

```
FAIL apps/mobile/src/features/diffs/nativeReviewDiffHighlighter.test.ts
  > highlightNativeReviewDiffVisibleRows > keeps grammar state across inline comment rows
```

I confirmed this is **not caused by this PR**:
- `git diff main -- apps/mobile/src/features/diffs/nativeReviewDiffHighlighter.ts apps/mobile/src/features/diffs/nativeReviewDiffHighlighter.test.ts` is empty — neither file is touched by this branch.
- Running `apps/mobile`'s full test suite alone (`cd apps/mobile && pnpm exec vp test run`), twice, passes cleanly both times: `520 passed (520)`.
- It only reproduces when `pnpm test` runs 4+ workspace packages' Vitest processes fully in parallel in this sandbox and fails identically both times I tried, which points at CPU-contention-sensitive timing inside the shared Shiki/JS-regex highlighter singleton under heavy load (the failing assertion shows coarser, less-tokenized output consistent with a tokenizer time budget being hit), not a logic bug in the diff.

I did not attempt a fix for it here since it's unrelated to #9's scope and I couldn't get a reliable, isolated repro to validate a fix against — flagging it explicitly rather than silently leaving it. Given the issue's own acceptance criteria (measured `ok`/`degraded`/`absent` browser capability, lazy 6h-cached probe, unit test coverage) are fully met and verified above, I'm titling this "Fixes #9"; happy to split the mobile flake into a separate follow-up issue if preferred.

Closes #9

## Summary by CodeRabbit

- **New Features**
  - Server environment information now includes detected Chromium availability and usability.
  - Server versions can display build identifiers for more precise troubleshooting.
  - Version checks distinguish release mismatches from build-level differences, reducing false warnings for remote environments.
  - Connection diagnostics now log version mismatches before establishing sessions.

- **Bug Fixes**
  - Improved reliability when multiple Git operations target the same repository, preventing conflicting worktree and fetch activity.
  - Browser detection reports clear states when Chromium is unavailable, degraded, or fully operational.

### DIFF
--- a/apps/server/src/environment/BrowserCapabilityProbe.test.ts
+++ b/apps/server/src/environment/BrowserCapabilityProbe.test.ts
@@ -1,0 +1,189 @@
+import * as NodePath from "@effect/platform-node/NodePath";
+import { afterEach, describe, expect, it } from "@effect/vitest";
+import * as Effect from "effect/Effect";
+import * as FileSystem from "effect/FileSystem";
+import * as Layer from "effect/Layer";
+import * as ChildProcessSpawner from "effect/unstable/process/ChildProcessSpawner";
+import { HostProcessEnvironment, HostProcessPlatform } from "@t3tools/shared/hostProcess";
+import { vi } from "vite-plus/test";
+
+import * as ProcessRunner from "../processRunner.ts";
+import { probeBrowserCapability } from "./BrowserCapabilityProbe.ts";
+
+const runMock = vi.fn<ProcessRunner.ProcessRunner["Service"]["run"]>();
+
+const ProcessRunnerTest = Layer.succeed(
+  ProcessRunner.ProcessRunner,
+  ProcessRunner.ProcessRunner.of({
+    run: (input) => runMock(input),
+  }),
+);
+const PathLayer = NodePath.layer;
+
+const linuxHome = "/home/pi";
+const linuxCacheRoot = `${linuxHome}/.cache/ms-playwright`;
+const linuxHeadlessShellPath = `${linuxCacheRoot}/chromium_headless_shell-1148/chrome-linux/headless_shell`;
+const systemChromiumPath = "/usr/bin/chromium";
+
+const withEnv = <ROut, E, RIn>(
+  layer: Layer.Layer<ROut, E, RIn>,
+  platform: NodeJS.Platform,
+  env: NodeJS.ProcessEnv,
+) =>
+  Layer.mergeAll(
+    layer,
+    Layer.succeed(HostProcessPlatform, platform),
+    Layer.succeed(HostProcessEnvironment, env),
+  );
+
+const successResult = (stdout: string) => ({
+  stdout,
+  stderr: "",
+  code: ChildProcessSpawner.ExitCode(0),
+  timedOut: false,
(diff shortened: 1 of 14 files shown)


fzoll:
Opened against the wrong repo (upstream instead of the fork). Closing; correct PR opened against fzoll/t3code.

PR #11482, opened 2026-09-12
fix(server): report Claude Team and Enterprise usage limits

TonybynMp4:
## What Changed

`claudeRateLimitEventToUpdate` now reads the `unifiedWindows` map a `rate_limit_event` can carry, falling back to the top-level `rateLimitType`/`utilization` pair for any window the map omits. A window reported by the runtime also clears an `unsupported` snapshot instead of being discarded by it.

On top of that mapping fix, two more gaps in how a probe's result gets reconciled with what is already on screen turned out to affect the same accounts:

- **Cache hydration at boot.** `hydrateCachedUsageLimits`/`hydrateCachedProvider` in `providerStatusCache.ts` now replay a provider's cached usage windows at server start, as long as each window's own period (`resetsAt`, or `windowDurationMins` measured from the cached read) has not rolled over. `ProviderRegistry.ts` feeds the replayed windows through the same `applyUsageLimits` path a live `rate_limit_event` uses, so the following boot-time probe treats them as already published rather than overwriting them. A cached `unavailable` verdict is never replayed — for a Team/Enterprise account that verdict is just last run's wrong guess.
- **A probe result of `undefined` no longer wipes published windows.** Every driver omits `usageLimits` on its early-return failure paths (missing executable, spawn failure, timeout), and some drivers never populate it at all. `resolveUsageLimitsAfterProbe` in `providerUsageLimits.ts` now treats that omission the same as a failed probe and keeps the last published snapshot, instead of clearing the bars. The Claude driver also now opts into `keepPublishedWindowsWhenProbeUnsupported`, so its own `unsupported` verdict doesn't blank windows a turn already established.

## Why

Claude Team and Enterprise accounts show "This account has no subscription limits" in the Limits view while claude.ai reports their utilization just fine. Three things stack up:

1. `get_usage` answers `rate_limits_available: false` for these accounts, before and after live API traffic. That is CLI behavior, not something this repo can fix.
2. Their `rate_limit_event` carries every window at once under an undocumented `unifiedWindows` key, with **no top-level `utilization`**. The old mapper early-returned `undefined` whenever `utilization` was missing, so every one of these events was dropped.
3. Even with a mapped update, the `unsupported` latch in `applyUsageLimitsUpdate` would have thrown it away.

A captured event from a Team account, matching what claude.ai displayed at the time:

```json
"rate_limit_info": {
  "status": "allowed", "resetsAt": 1789236000, "rateLimitType": "five_hour",
  "overageStatus": "rejected", "overageDisabledReason": "org_level_disabled",
  "unifiedWindows": {
    "five_hour": { "utilization": 0.85, "resetsAt": 1789236000 },
    "seven_day": { "utilization": 0.74, "resetsAt": 1789405200 }
  }
}
```

`unifiedWindows` postdates the SDK typings we pin, so it is read structurally, the same way `model_scoped` already is.

On the resolver side, a probe that successfully reads windows never answers `unsupported`. So windows on screen plus an `unsupported` probe can only mean a turn reported limits the probe cannot see — the normal state of these accounts. Letting the probe win there blanks the bars on every status refresh and redraws them on the next turn, so `resolveUsageLimitsAfterProbe` keeps the published windows for providers that opt in.

Blast radius is limited to Claude: Codex only reports `unsupported` for `apiKey` accounts, which never stream subscription windows, and `cliproxyApi` uses `probeFailed`. The `probed === undefined` fix and cache hydration apply to every provider, but only change behavior where a probe was previously wiping windows for no informational reason.

Gap that remains: the very first boot of an environment, before any cache exists, still has no windows to show until the first turn streams one. Cache hydration closes the gap for every restart after that. Fully closing the first-boot case needs the probe itself to tell a subscription account from an API key, and the SDK's `initializationResult().account` exposes no `subscriptionType` to do it with — only `tokenSource` and `apiProvider` — so that's left for a follow-up.

Tests cover the captured Team payload, unified-beats-top-level for the same window, the scoped overage row coexisting with unified windows, malformed entries, cache hydration (windows surviving/aging out/skipping `unavailable`), and the `keepPublishedWindowsWhenProbeUnsupported` opt-in split. Two existing tests that asserted the old latch were inverted.

## UI Changes

None — the Limits view renders the bars it is given; this only changes whether it is given any.

## Checklist

---

## Summary by CodeRabbit

- **New Features**
  - Claude usage limits now support account-wide usage windows, including accurate reset times and prioritization of unified limits.
  - Usage-limit windows are restored when cached provider status is loaded, so usage bars remain visible after restarting.
  - Claude usage windows can remain available when status checks temporarily cannot report limits.

- **Bug Fixes**
  - Expired usage windows are now removed during refreshes and cache restoration, including duration-based windows.
  - Invalid or unavailable usage data no longer produces misleading usage indicators.
  - Stale unsupported usage snapshots are cleared when a new usage window is reported.


juliusmarminge:

TonybynMp4:
~~Seems Claude was updated and now correctly reports team subs anyway!🚀~~
nvm

PR #11491, opened 2026-09-13
fix(swift-ios): reopened threads no longer show stale history after swipe back

mackinleysmith:
## What Changed

On compact width, the swipe-back gesture now only collapses the split view column. It no longer clears the thread selection. This is the same thing the system back button does, so the thread view is re-shown rather than destroyed and recreated. One line in `WorkspaceView.swift` plus a comment explaining why.

## Why

Reopening a thread after leaving it with swipe back showed the cached transcript and never caught up. No pill, no updates, until the user backed out and reopened. It was most visible after a background trip because new content had arrived in the meantime.

Instrumenting the simulator build showed the sequence:

1. Swipe back cleared `selectedThreadID`, destroying the thread view while the split view was still finishing its interactive pop.
2. The next open of that thread got `onAppear` followed by a spurious `onDisappear` while still on screen.
3. That disappear ran `releaseThread`, which cancelled the load before it subscribed to the socket or armed the HTTP fallback.
4. The model treated the cancellation as benign and kept the cached detail, with sync state nil, so nothing was shown.

Leaving with the nav-bar back button keeps the selection and re-shows the same view, and every reopen on that path was fresh. Making the swipe path do the same removes the trigger.

Lifecycle log, before (the reopen is instance 3D1EF3):

```
[repro-view] task start instance=3D1EF3
[repro-view] onAppear instance=3D1EF3
[repro-view] onDisappear instance=3D1EF3
[repro] releaseThread active=<thread> seq=nil wasSync=false
[repro-view] task end instance=3D1EF3 cancelled=true
```

After, same flow:

```
[repro-view] task start instance=F8494C
[repro-view] onAppear instance=F8494C
[repro] loadThread cached seq=90502 warm=false
[repro] subscribe after=90502
[repro] item event thread.message-sent seq=90503
[repro] item event thread.activity-appended seq=90507
[repro] item synchronized active=90507
[repro] markDetailSynchronized
```

## Verification

- Reproduced on iPhone 17 Pro simulator against a disposable local server: open thread, swipe back, push a message over HTTP, reopen. Stale every time on the base commit, both with and without a background trip in between.
- Same flows on this branch show the new message and reach `synchronized`. Also verified the swipe, background, foreground, reopen variant.
- Debug build compiles clean. Native test suite not run locally; nothing in it references the changed callback.

## Checklist

- [ ] I included before/after screenshots for any UI changes (no visual change; behavior only, logs above)

Model: Claude Fable 5.1. Harness: Claude Code inside T3 Code.


mackinleysmith:
**Release Smoke failure is inherited from the base branch, not this change.**

The job fails in `vp install --lockfile-only` with `ERR_PNPM_UNUSED_PATCH` for `expo-audio@57.0.4`. `apps/mobile/package.json` on `t3code/rebuild-mobile-app-swift` still declares `"expo-audio": "~57.0.4"`, and expo-audio 57.0.5 was published after this branch's last green CI run on Sep 8, so the lockfile-only resolve now picks 57.0.5 and the patch keyed to 57.0.4 goes unused.

`main` fixed this in #11426 (pin to `57.0.4`), which the Swift branch has not picked up yet. Merging `main` into the branch, or cherry-picking 18d8cbfd92, clears it for every PR against this base. Kept out of this PR to stay single-concern. Happy to open that as a separate one-line PR if useful.

mackinleysmith:
Pushed 57617affb1, a cherry-pick of main's #11426 (pin `expo-audio` to `57.0.4`), so Release Smoke can pass on this base without waiting for a merge from `main`. Two-line change to `apps/mobile/package.json` and `pnpm-lock.yaml`, no other diff. Verified locally with the same `vp install --lockfile-only --ignore-scripts` the smoke job runs; it resolves clean with no lockfile drift.

mackinleysmith:
Rebased onto the current base, which now includes #11426 from main, so the expo-audio cherry-pick dropped out. Back to the single fix commit, 6a6e572fff. No conflicts.

mackinleysmith:
Closing. Theo fixed the trigger upstream in 33d0715aec ("use native content swipe-back on iOS 26"): the custom pan recognizer is no longer installed on iOS 26+, so the system back-swipe handles it and the selection is never cleared mid-pop. That commit's comment describes the same failure this PR addressed. The custom gesture still runs on iOS 17 and 18 and still clears the selection there, so this one-line change would still apply on that fallback path if it ever matters. Leaving the branch and the repro notes in the description for reference.

PR #11534, opened 2026-09-13
fix(server): reject partial output from failed Cursor runs

saphid:
## Summary

Cursor text generation (`apps/server/src/textGeneration/CursorTextGeneration.ts`) decoded `run.result` whenever the string was non-empty. A Cursor SDK run that terminated `error` or `cancelled` could still carry a partial `result`, and that partial JSON was accepted as a successful commit message, PR body, branch name, or thread title.

The fix requires `status === "finished"` before decoding output, and the failure detail now names the actual terminal status (`cancelled`, `error`) even when partial output exists, instead of the error path claiming "no output". Cleanup is unchanged: the run-cancel finalizer and agent `asyncDispose` still run on the failure path.

Two regression tests cover `error` and `cancelled` runs that return valid title JSON: both now fail with a typed `TextGenerationError` naming the operation and terminal status, and the agent is disposed. Successful generation is unchanged.

## Credit

Adapted from @kiiwee's earlier proposal in #11158. That branch carries a stale, divergent base (300+ file diff), so this PR lands the same minimal fix on a current `t3code/codex-turn-mapping` base.

## Test plan

- [x] `CI=true vp test run apps/server/src/textGeneration/CursorTextGeneration.test.ts` — 10/10 pass; both new regressions failed before the fix
- [x] `tsc --noEmit -p apps/server/tsconfig.json` — clean
- [x] `vp lint` + `vp fmt` on both changed files — clean

Engineering: SWE-2 High through Devin in the T3 Code harness; independent review: GPT-6 Astra medium via `codex exec` (read-only, no actionable findings).


saphid:
@juliusmarminge this OV2 fix rejects partial Cursor metadata after a failed or cancelled run and is ready for your review.

Latest head `ef8ca64dc90a` has green CI, is mergeable into `t3code/codex-turn-mapping`, and has no unresolved review threads. The PR description contains the focused regression evidence. We will keep babysitting it through landing.

PR #11535, opened 2026-09-13
fix(server): wake paused Cursor replay runs on mismatch

saphid:
## Summary

A Cursor replay run whose next transcript entry expects `run.cancel` parks its `run.wait` fiber on the runner's cursor-advanced signal while it waits for the cancellation to be recorded. If a different operation mismatched the transcript first — e.g. `session.close` emitting `agent.close` — the failure was stored via `recordFailure`, but the progress signal was never resolved, so the paused waiter hung forever instead of reporting the failure.

`recordFailure` now resolves the current progress signal, so a parked waiter wakes, re-checks the recorded failure, and fails with the same typed `CursorReplayFrameMismatchError` — without advancing the transcript cursor. A regression test deterministically parks a run waiter, issues a mismatching `session.close`, and asserts both operations receive the identical error instance (no sleeps or timeout-based passing assertions).

Replay test harness only; no client, wire-schema, or production request-path changes.

#### Test plan

- [x] New test timed out (60s) before the fix; passes in ~4ms after — `vp test run apps/server/src/orchestration-v2/Adapters/CursorAdapterV2.testkit.test.ts` (4/4)
- [x] Related replay suites green: CursorAdapterV2, CursorAgentSdk, OrchestratorReplayRecovery, OrchestratorReplayFixtures (86 tests)
- [x] Scoped oxlint clean; `tsc --noEmit` for `t3` exits 0; `vp fmt --check` clean

Prior proposal: #11157 by @kiiwee identified the same defect and fix; this adapts that one-line change plus regression test onto the current V2 base.

Targets the Orchestrator V2 branch in #2829.

Engineered by SWE-2 High via Devin on T3 Code (thread cc3c9ba9-2d7a-473c-87a4-11a18f8b6b66), campaign saphid/t3code-personal#298; independently reviewed by GPT-6 Astra (medium, read-only, no actionable findings).


saphid:
@juliusmarminge this OV2 fix wakes paused Cursor replay when its transcript fails and is ready for your review.

Latest head `62603ea8f9f0` has green CI, is mergeable into `t3code/codex-turn-mapping`, and has no unresolved review threads. The PR description contains the focused regression evidence. We will keep babysitting it through landing.

PR #11557, opened 2026-09-13
fix(server): roll back question attachment copies when respond preparation fails

saphid:
## Problem

A multi-question `runtime-request.respond` claims each question's pending uploads into the thread attachment store one question at a time (`dispatchCommand` in `ThreadMessageIntake.ts`). If a later question's claim failed (missing staged upload, size mismatch, bad id) or answer-path preparation failed (`appendUserInputAttachmentPaths` finding a referenced file gone), the earlier questions' new thread-store copies were orphaned: nothing referenced their fresh random ids, and the pending sweep only reclaims `pending-` files, so the orphans were permanent.

## Fix

- All pre-dispatch preparation (per-question claims plus `appendUserInputAttachmentPaths`) runs inside one error boundary that tracks every claimed path across questions and releases only the new copies on failure. The boundary uses `Effect.onError`, so interruption and defects roll back too — nothing is dispatched at that point, so release is provably safe.
- On dispatch success, the stored events are checked for a recorded `questionAnswer`. When one exists, claimed copies not referenced by it are released — this covers the accepted-but-response-lost retry, where the replayed receipt reports the first attempt's attachment ids and the retry's fresh copies would otherwise be orphaned. When no recorded answer exists the outcome is ambiguous and every copy stays.
- On dispatch error, copies are released when the command provably never committed (`dispatchWasNotAccepted`: previously-rejected replay, cross-thread command-id conflict, and other pre-commit rejections), matching the `message.dispatch` path's convention. Ambiguous dispatch failures retain the copies because the response may have been accepted and the provider reads the answer paths from disk.
- Pending upload sources and pre-existing attachments are never in the claimed set, so the response stays retryable and existing files are preserved.

Known remaining gap: a first-time respond rejection is recorded by the orchestrator via `commitRejectedCommand` but resurfaces as an ambiguous `OrchestratorDispatchError`, so its copies are retained. `OrchestratorCommandRejectedError` is only produced for `queued-run.edit`; surfacing a distinguishable rejection tag for respond would be an orchestrator change outside this PR's scope.

## Validation

- Reproduced both leaks before the fix: 4 failing regressions (early-copy/later-claim failure, path-preparation failure, rejected dispatch, retry leaving orphans).
- `vp test run apps/server/src/orchestration-v2/ThreadMessageIntake.test.ts apps/server/src/orchestration-v2/AttachmentClaims.test.ts` — 17 passed.
- Targeted `vp lint` and `tsc --noEmit` (apps/server) clean for the changed files; `vp fmt` applied.
- Independent read-only reviews of each frozen diff (GPT-6 Astra, medium): third iteration reported no actionable findings.

Supersedes the equivalent fix proposed in #11150 by @kiiwee (that PR carries a large divergent diff; this is the same minimal fix adapted onto the current V2 base — credit to the earlier proposal).

Implemented by SWE-2 (Devin) through the T3 Code / Cursor harness.
Coordination: T3 thread 13b3aef7-8ec9-42dc-b7bc-eb3221178fc0; campaign issue saphid/t3code-personal#298.


saphid:
Pushed `0cd6b62` + `18cbf90` addressing review findings on the previous head: (1) the success path now releases claimed copies left unreferenced when dispatch replays an earlier accepted response (the recorded `questionAnswer` defines the retained set; no recorded answer → everything retained); (2) pre-dispatch rollback boundaries use `Effect.onError` so interruption and defects also clean up; (3) the rejected-dispatch regression now exercises `OrchestratorCommandPreviouslyRejectedError`, the rejection signal respond actually gets on receipt replay. One documented gap remains: first-time respond rejections resurface as ambiguous `OrchestratorDispatchError` and retain copies — a distinguishable tag needs an orchestrator change outside this scope. 17/17 focused tests pass; CI green on the new head.

saphid:
@juliusmarminge this OV2 fix is ready for maintainer review on head `abbd532d8e9d2f690393ea382e4701d5744a3137` (rebased onto moved base `834edb9a88`; the feature diff is byte-identical to the previously approved `e8d198d4e4`).

- Focused tests: 19/19 pass across `AttachmentClaims.test.ts` and `ThreadMessageIntake.test.ts`, covering pre-dispatch rollback, pending-source retryability, pre-existing retention, duplicate ids, ambiguous-failure retention, proven rejection, receipt replay (incl. empty and text-only recorded answers), and interrupt-during-copy.
- Independent GPT-6 Astra xhigh reviews: clean, 0 findings on `e8d198d4e4`, and clean again on this rebased head (no semantic drift from the new base).
- CI: Test Server 1/2/3, Release Smoke, Rust, Mobile Native Changes all green. `Check` and `Test` fail on this head **and identically on base commit `834edb9a88`** — inherited V2-base breakage, not this diff (verified: same jobs fail on the base).

PR #11570, opened 2026-09-13
feat(web): add opt-in in-app thread notifications

Bil0000:
When T3 Code has focus, activity in another thread can be missed. Add a small toast with the thread title and an **Open thread** action.

Enable **In-app notifications** in **Settings → General**, directly below Thread notifications. It is off by default, saved per device, searchable, and included in Restore device defaults. It works independently of system notification permission.

Default-off setting

Toasts and system alerts share the existing event detector and titles for completion, input, approval, and failure. Failure now uses that same path and the existing attention sound. Toasts skip the active thread, background windows, archived threads, and initial or reconnect snapshots. A shown toast replaces the system popup for that event; sound settings still apply. This covers the shared web/desktop client. Mobile push is unchanged.

Verified: 216 focused tests, web type checking, targeted lint and formatting. Chromium checks covered all four states, opt-in persistence after reload, disabled suppression, and 1280px/390px layouts. Browser evidence uses fixture threads with simulated shell updates; it is not a provider end-to-end run.

| Before | After enabling in-app notifications |
| --- | --- |
| No foreground alert | Completion toast |

<details>
<summary>Failure alerts</summary>

Failure toast

Failure toast at 390px

</details>

Model: GPT-6. Harness: Codex.


Bil0000:
/recheck-vouch

The label job received HTTP 502 from GitHub while adding vouch:trusted. Retrying through the supported workflow command.

Bil0000:
The docstring-coverage warning does not need a code change here. This adds no public API, and the repo guidance asks us to avoid narrating control flow that is clear from the source. The focused behavior tests cover the new toast rules.

Bil0000:
Checked the active-thread desktop-alert finding against the current source and the PR base (`20363c32c9bfdbf49c2716ef11d1f18483fcc01b`). The existing desktop path already delivers regardless of the active route or window focus. This PR adds an opt-in in-app delivery path for inactive threads and preserves the existing system-alert behavior when a toast is not shown. The event detector and titles are shared by both paths.

I am leaving that existing desktop policy unchanged. Adding the suggested exclusion would change system alerts even when the new option is off; placing it before sound delivery would change sound behavior too. That is a separate behavior change from this opt-in toast request. The active-thread toast suppression is covered by the focused tests.

PR #11574, opened 2026-09-13
fix(orchestrator): report terminal runs after wait timeout

saphid:
## Summary

A run can finish while `waitForThread` is handling its timeout. The timeout path already re-reads the projection, but it unconditionally reported `timedOut: true` even when that final read showed the run had reached a terminal status. Callers saw a completed (or failed/interrupted/cancelled/rolled-back) run mislabeled as timed out.

`timedOut` is now recomputed from the final projection read: a terminal run is reported as completion, a still-active run remains timed out, and a vanished run still raises `ThreadManagementRunNotFoundError`.

## Validation

- Reproduced the defect first: terminal-at-timeout scenarios returned `timedOut: true` before the fix (5 failing cases).
- `vp test run apps/server/src/orchestration-v2/ThreadManagementService.test.ts` — 14/14 pass. New parameterized cases use `Deferred` + `TestClock` to park the wait inside its poll until the timeout wins, then assert the final-read outcome for all five terminal statuses, a still-active run, and the missing-run error.
- Targeted `vp lint`, `vp fmt`, and `tsc --noEmit` on `apps/server` are clean for the touched files.

## Prior art

Adapted from the approach in #11156 by @kiiwee (thanks!) — that PR went stale/conflicted against the current V2 base; this carries the same core fix rebased fresh, with coverage extended to every terminal run status.

---

Implemented by SWE-2 (Devin) running in T3 Code. Coordination: OV2 campaign saphid/t3code-personal#298, worker thread 5810b980-142f-4c85-9c18-8dcadb41dc76.


saphid:
@juliusmarminge this PR is ready for review/merge. (ov2-julius-ready)

Verified at head `1492e36c` (rebased onto the current base `24476841`; earlier heads: `2819efac`, `85d7fcee`, `faf05212`, `a51e715a`, `c880f6f9` — identical diff throughout):

- **Fix**: `waitForThread` recomputes `timedOut` from the final projection read after the timeout wins the race — a run that reached a terminal status in that window reports completion instead of a false timeout; still-active runs stay timed out; a vanished run still raises `ThreadManagementRunNotFoundError`.
- **Tests**: 14/14 focused tests pass locally and in CI — `Deferred`/`TestClock` parameterized cases cover all five terminal statuses at-timeout (completed, failed, cancelled, interrupted, rolled_back), a still-running control, and the missing-run error.
- **Reviews**: independent GPT-6 Astra reviews clean on the frozen diff (0 findings across repeated identical-diff passes; xhigh pass verified diff SHA-256); Macroscope Approvability APPROVED (approval recorded at `2819efa`, identical diff).
- **CI**: all checks green or expected skips on this head — Check, Test, Test Server 1–3, Release Smoke, Rust all success. `mergeable_state: clean`.

Adapted from the approach in #11156 by @kiiwee, extended to cover every terminal run status.

PR #11585, opened 2026-09-13
fix(server): skip corrupt scheduled-task rows instead of stopping the scheduler

saphid:
@/Users/saphid/.local/share/t3/ov2-nightly-delivery-20260913/tasks/13/pr-body.md


saphid:
@juliusmarminge — ready for maintainer review/landing.

**Head:** `fd6e2e8` against `t3code/codex-turn-mapping` — mergeable, clean.

**What it does:** the ScheduledTasks persistence loader now skips individually corrupt rows instead of defecting the poll fiber — malformed/empty/NULL task ids decode through the schema as typed errors, unparseable `next_run_at` values (at poll time and at the dispatch re-read) are treated as not due, undecodable schedule JSON is warned-and-skipped, schema-valid but unrepresentable intervals count as "no next run", and startup recovery releases `NULL`-id stuck rows via `IS`. Diagnostics omit `taskId` when the stored id can't decode and never carry prompt/credential payloads.

**Verified on this head:**
- 3/3 focused SQLite tests pass (real `SqlitePersistenceMemory`, deterministic TestClock; mutation-checked against the pre-fix code)
- All CI green: Check, Test, Test Server 1–3, Rust, Release Smoke
- Macroscope approved at `fd6e2e8`
- Independent post-babysitting review (GPT-6 Astra, xhigh): zero findings on `fd6e2e8`; a prior Fable 5.1 finding on the re-read path is fixed in this head

saphid:
@juliusmarminge — head moved: rebased onto the rewritten base `d8c75ec2` after upstream force-updated `t3code/codex-turn-mapping`. New head `a9cc488d`; the change is patch-identical to the previously reviewed `fd6e2e8` (same `git patch-id`), and the target files are unchanged at the new base so the fix is still needed.

**Verified on `a9cc488d`:**
- 3/3 focused SQLite tests pass on the new base
- All CI green: Check, Test, Test Server 1–3, Rust, Release Smoke — mergeable state `clean`
- Fresh independent review of the rebased diff (GPT-6 Astra, medium): approved, no actionable findings

PR #11586, opened 2026-09-13
fix(codex): keep browser tasks on T3 preview

### LINKED ISSUE #11579: [Bug]: Codex 0.154.0 Computer Use can bypass T3 collaborative browser
### Before submitting

### Area

apps/server / Codex provider

### Summary

With Codex CLI 0.154.0, Codex can choose its bundled `cua_repl` / `unified-computer-use` browser path instead of T3 Code's attached `t3-code` collaborative browser MCP, even when T3 has exposed `preview_*` tools and explicitly instructs Codex to use them.

When that bundled path has no available IAB browser, `cua_repl` fails with `Browser is not available: iab`. The model can then incorrectly report that the live browser is unavailable even though T3's own preview automation broker is healthy and the `t3-code` MCP is attached.

### Steps to reproduce

1. Run a current T3 Code nightly with Codex CLI 0.154.0.
2. Start a Codex-backed thread with agent browser access enabled.
3. Give the agent a browser task where the T3 Code MCP exposes `preview_status`, `preview_open`, `preview_navigate`, and related tools.
4. Observe the tool route selected by Codex.

This has been observed with both `gpt-6-astra` and `gpt-5.6-sol`.

### Expected behavior

When the `t3-code` MCP exposes `preview_*` tools, T3's collaborative browser should be the authoritative browser route for the session. Codex should call `preview_status` first and `preview_open` when needed before deciding browser automation is unavailable.

A failure from an unrelated bundled Computer Use browser should not be interpreted as failure of T3's collaborative browser.

### Actual behavior

On affected turns Codex makes no T3 preview call and instead selects the bundled Computer Use path, for example:

```text
server: cua_repl
tool: js
plugin: unified-computer-use@openai-bundled
```

It may then run:

```js
await cua.getState();
await cua.createBrowserTab('iab', '<url>', { visible: false });
```

and receive:

```text
Browser is not available: iab
```

The assistant can then conclude that "the live browser connection is unavailable" even though this error belongs to the bundled IAB/Computer Use path, not T3's `preview_*` tools.

### Evidence / regression boundary

- T3 Code nightly: `v0.0.41-nightly.20260913.1646`
- Codex CLI: `0.154.0`
- T3's `t3-code` MCP was attached successfully and MCP `tools/list` completed.
- T3's PreviewAutomationBroker remained connected.
- Historical sessions using Codex CLI `0.150.1` correctly selected `t3-code` tools such as `preview_status`, `preview_open`, `preview_navigate`, and `preview_snapshot`.
- The current T3 developer instructions already prefer `preview_*`, so the new overlapping browser-capable tool surface can still win tool selection despite the existing generic guidance.

### Impact

Major degradation for Codex browser tasks in T3 Code. The user can be told that live browser access is unavailable when T3's browser is actually available, and the agent never attempts the product-native preview route.

### Proposed T3-side mitigation

Harden the Codex developer instructions when `preview_*` tools are actually attached:

- explicitly identify `cua_repl`, `unified-computer-use`, and IAB as a separate browser surface;
- require `preview_status` / `preview_open` before any conclusion that the browser is unavailable;
- do not fall back to bundled Computer Use while T3 preview is available unless T3 preview itself returns an explicit unsupported/unavailable result or the user explicitly requests another browser;
- state that `Browser is not available: iab` is not evidence that T3 preview is unavailable;
- keep this entire steering block omitted when T3 preview tools are not attached.

This should be covered by regression tests in the Codex developer-instruction tests.

### Upstream follow-up

Prompt steering is only a mitigation. It would be better if Codex app-server exposed a supported host/session control to suppress or deprioritize bundled Computer Use/browser tools when an embedding host supplies its own authoritative browser MCP. That likely warrants a corresponding OpenAI/Codex issue once the T3 reproducer is public.

### Related issues

- #6355 is about a detached T3 preview tab and is a different failure mode.
- #7212 is about failed preview UI state and explicitly did not involve Computer Use.

nkcbuilds:
Codex 0.154.0 can select its bundled `cua_repl` / `unified-computer-use` browser path even when T3 Code has attached its own `preview_*` browser tools. If the bundled IAB route is unavailable, the model can incorrectly report that live browser access is down without ever trying T3 preview.

This change hardens the Codex developer instructions only when T3 preview capability is attached:
- identifies `cua_repl`, `unified-computer-use`, and IAB as a separate browser surface;
- requires `preview_status` / `preview_open` before concluding that T3 browser access is unavailable;
- prevents `Browser is not available: iab` from being treated as a T3 preview failure;
- preserves the existing fallback rules and omits all collision steering when preview tools are not attached.

Regression coverage verifies the steering in both collaboration modes, verifies that it is absent without browser tools, and verifies that device-only sessions are not affected.

Tests:
- `pnpm --filter t3 test -- src/provider/Layers/CodexSessionRuntime.test.ts` — 48/48 passed
- `pnpm exec vp lint apps/server/src/provider/CodexDeveloperInstructions.ts apps/server/src/provider/Layers/CodexSessionRuntime.test.ts` — 0 warnings / 0 errors
- `pnpm exec vp fmt --check apps/server/src/provider/CodexDeveloperInstructions.ts apps/server/src/provider/Layers/CodexSessionRuntime.test.ts` — passed
- `pnpm --filter t3 typecheck` — passed (existing Effect suggestions only)
- `git diff --check` — passed

This is the T3-side mitigation accepted in #11579. A deterministic host-level control to suppress or deprioritize bundled Computer Use remains an upstream Codex concern.

Fixes #11579

Model: GPT-5.6 Sol
Harness: ChatGPT with T3 Code workspace tools

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved browser tool guidance so preview availability is evaluated accurately.
  * Actionable preview errors are now retried instead of immediately switching to an alternative browser.
  * Clarified handling of separate computer-use browser tools and their availability errors.
  * Device-only sessions now receive guidance specific to available device controls.


nkcbuilds:
closing the pr becasue the fix temporary fix is not reliable every time, creating another pr with more tests and better changes.

PR #11590, opened 2026-09-13
fix(server): preserve due schedules across equivalent time formatting

saphid:
## Problem

Fixed-time scheduled tasks compare `timeOfDay` as a string when deciding whether an upsert changed the schedule. The contract accepts both `9:00` and `09:00` (`/^([01]?\d|2[0-3]):([0-5]\d)$/`), so re-saving a task with an equivalently formatted time was treated as a schedule change and recomputed `next_run_at`. If today's occurrence was already due but not yet dispatched, that save deferred it until the next day.

## Fix

`isSameSchedule` now parses both times and compares hours/minutes semantically. Format-only edits preserve the persisted deadline; real time (`09:30`), weekday-mask, and interval changes still reschedule, and interval-vs-fixed-time remains distinct.

Adapted from #10438 by @CouchRiv, which conflicts with the current V2 base; this rebase keeps the same approach and moves the service test onto the full in-memory migration chain (`SqlitePersistenceMemory`).

## Validation

- New service test (`ScheduledTaskService.schedule.test.ts`) decodes real upsert input through the contract schema, persists to in-memory SQLite, and uses a controlled clock: creating `9:00`, crossing the due instant, then saving `09:00` retains the persisted `next_run_at` (failed before the fix — the run moved to the next day); saving `09:30` reschedules as before.
- `Schedule.test.ts` covers `9:00`/`09:00`, `09:00`/`9:00`, and `0:30`/`00:30` equivalence; existing cases keep weekday and interval changes distinct.
- `vp test run apps/server/src/scheduledTasks/ packages/contracts/src/scheduledTask.test.ts` — 15/15 pass; server typecheck and targeted lint/format clean.

Targets the V2 branch in #2829.

Implemented by SWE-2 (Devin) in the T3 Code harness.
Coordination: T3 thread 38bd20b6-be63-4e07-853d-51fbdd4add10; campaign https://github.com/saphid/t3code-personal/issues/298.


saphid:
@juliusmarminge this OV2 fix makes fixed-time schedule equality compare parsed hours/minutes, so re-saving a task with an equivalent time format (`9:00` -> `09:00`) no longer defers an already-due occurrence to the next day, and is ready for your review.

Latest head `7c5698b45e` has green CI, is mergeable into `t3code/codex-turn-mapping`, and has no unresolved review threads. Verified evidence: a new service test that decodes real upsert input, persists to in-memory SQLite (full migration chain) and uses a controlled clock — the due `next_run_at` survives the format-only save (failed before the fix: the run moved to the next day) while `09:30` still reschedules; plus new `isSameSchedule` cases for `9:00`/`09:00`, `09:00`/`9:00`, `0:30`/`00:30`. 15/15 focused tests pass and a clean independent review was run on the identical diff at `e43b3687f2`; `7c5698b45e` is that same change rebased onto the updated V2 base with the review re-run clean. The PR description contains the full evidence and credits the earlier proposal in #10438. We will keep babysitting it through landing.

PR #11591, opened 2026-09-13
fix(desktop): correct SnapShots animation sizing on Hyprland

Ta-noshii:
## What Changed

With **Settings → SnapShots → Capture animations** enabled on Hyprland, capturing a window drew a large, malformed rectangle while the screenshot flew into the draft: a small copy of the capture sat near its top-left corner, with the surrounding pixels stretched across the oversized rectangle. It disappeared when the animation finished, and the capture still attached correctly.

`Feedback::draw()` in `native/hyprland-snap-shot` now reattaches the existing image buffer on every visible animation frame instead of only when the texture switches (flash → image). The full `damage_buffer()` stays gated on an actual texture change, so the capture pixels are still uploaded once and no per-frame damage is introduced. Nothing is recaptured or regenerated.

To make this testable, the compositor setup is split out of `run()` into `connect()` so a test can build the client against the crate's in-process compositor fixture, and `fly()` wraps the start of the flight. The event loop and Hyprland IPC handling are unchanged.

## Why

Flight frames only changed the subsurface viewport and position. In Hyprland v0.56.2, `SSurfaceState::updateFrom()` copies `size` only inside its `updated.bits.buffer` branch, and the subsurface renderer draws with `m_current.size` ([SurfaceState.cpp](https://github.com/hyprwm/Hyprland/blob/v0.56.2/src/protocols/types/SurfaceState.cpp), [Compositor.cpp](https://github.com/hyprwm/Hyprland/blob/v0.56.2/src/protocols/core/Compositor.cpp), [SurfacePassElement.cpp](https://github.com/hyprwm/Hyprland/blob/v0.56.2/src/render/pass/SurfacePassElement.cpp)). Viewport-only commits therefore left the compositor's rendered size at the initial capture size while our viewport and position moved. Reattaching the same buffer each frame makes Hyprland refresh the size. This is a Hyprland compatibility workaround; it is not a general Wayland requirement.

Tested on Hyprland 0.56.2 (`efb50993`), Omarchy, native Wayland, one 5120×1440 output at fractional scale ~1.333. Fractional scaling was present during reproduction but was not shown to be required.

### Verification

Automated (`cargo test --locked --manifest-path native/hyprland-snap-shot/Cargo.toml`):

- New `transport_tests::flight_frames_reattach_the_capture_without_redamaging_it` drives `connect()` and `fly()`/`draw()` against an in-process fake compositor (wl_compositor, wl_subcompositor, wp_viewporter, wp_presentation, wlr-layer-shell, xdg-output) over a real socket pair and logs every request the capture subsurface receives. It asserts that each flight frame reattaches the buffer and that the pixels are damaged exactly once. It fails on the previous code (0 attaches across 3 frames) and passes with the fix.
- The existing `clips_across_negative_origin_and_scaled_outputs` and `flight_ends_at_exact_composer_bounds` tests still pass. They cover geometry only, not this regression.
- `cargo build --locked --release` for the helper and `git diff --check` are clean. The two pre-existing clippy warnings in this crate are untouched.

Visual: the same on-screen check with a synthetic checkerboard reproduced the oversized rectangle with the unpatched nightly helper (`0.0.41-nightly.20260910.1507`, source `3836890e`) and rendered the intended small rectangle with this build, both reporting `ready`, `landed`, and `done`. Ordinary captures with the patched helper installed then worked correctly. Performance was not benchmarked; the per-frame cost is one extra `wl_surface.attach` of an already-uploaded buffer.

## UI Changes

Captured on a temporary headless Hyprland output with a synthetic checkerboard (red border) flying into a throwaway terminal window, so the frames contain no desktop content. Same Hyprland 0.56.2 session; the headless output ran at scale 2, so the bug also reproduces without fractional scaling.

**Before / after, 168 ms into the flight.** Left: unpatched helper keeps the initial surface size, the capture shrinks into the top-left corner, and the border pixels smear across the rest. Right: this PR.

Before and after

**Motion, 4x slow, synced from the animate command** ([real-speed mp4](https://raw.githubusercontent.com/Ta-noshii/t3code-nowbar/8371c4ed59b728114c4ac8350b71bfdb1deb272c/before-after.mp4), [slow mp4](https://raw.githubusercontent.com/Ta-noshii/t3code-nowbar/8371c4ed59b728114c4ac8350b71bfdb1deb272c/before-after-slow-4x.mp4)):

Flight before and after, slowed 4x

Assets live on the fork's `evidence/pr-11591` branch, not in this PR's commits.

## Checklist

Work done by Claude Fable 5.1 in Claude Code (via T3 Code).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved screenshot flight animations on newer Hyprland versions by ensuring frames continue to display correctly.
  * Improved handling of captured windows that are outside currently visible outputs.
  * Ensured image dimensions consistently match the requested capture size.

* **Tests**
  * Added coverage for multi-frame flight animations, buffer presentation behavior, and repeated frame updates.


Ta-noshii:
On the CodeRabbit docstring-coverage warning: the touched functions it counts are mostly `Dispatch::request` and `GlobalDispatch::bind` trait implementations in the test compositor fixture, plus the unchanged body of `run` that moved. Those are protocol plumbing and a doc comment on each would only restate the trait. I added doc comments where they carry information (`draw`, the fixture constructors, and the request logger) in 91491f8. The remaining undocumented functions are left as is on purpose, per the repo's guidance to comment how a thing is used rather than annotate everything.

PR #11594, opened 2026-09-13
feat(web): add chat width setting for wide screens

otavio:
The chat timeline and composer are capped at 48rem, so on a large monitor most of the window stays empty and wide content such as markdown tables has to scroll sideways.

This adds a **Chat width** setting under Settings → Appearance with three options: Comfortable (the current 48rem, still the default), Wide (72rem), and Full. It is a client setting, like diff colors. The root sets `data-chat-width` on `<html>`, CSS maps it to `--chat-max-width`, and the six chat containers that used `max-w-3xl` now use `max-w-(--chat-max-width)`. The setting has a reset button, appears in settings search, and is included in "Restore defaults".

Desktop picks this up through the web app. Mobile has no fixed-width chat column, so nothing changes there.

The commits are split for review: the contracts setting and its test, the layout change, then the settings control.

Screenshots use a seeded test thread.

| Before (Comfortable) | After (Full) |
| --- | --- |
| before | after |

Wide:

wide

Setting:

setting

Made with Claude Opus 5 in Claude Code, running inside T3 Code.

## Summary by CodeRabbit

* **New Features**
  * Added a Chat width setting with Comfortable, Wide, and Full options, allowing you to adjust the width of the chat timeline and composer.
  * Added Chat width to settings search.
  * Added a not-found page with a button to return home.
* **Bug Fixes**
  * Improved timeline minimap navigation at different chat widths so its controls are interactive when there is enough space.
  * Updated the error page’s retry action.


OmarAlaaeldein:
This would fix a real pain for me too — on a wide monitor the thread/composer sits at 768px and uses less than half the available width, with no setting to change it. Wide tables/code output constantly need horizontal scroll despite all the whitespace.

Love the Comfortable / Wide / Full approach with Comfortable staying default. That matches the existing readable / wide / expanded pattern already used for workspace pages, just extended to chat where it's actually needed.

v1b3coder:
Supporting the need — on a 3440px-wide monitor the fixed 48rem column leaves most of the pane empty and wide output still scrolls sideways. Comfortable / Wide / Full with Comfortable as the default looks like the right shape, and as a client setting it also covers desktop and remote connections.

One gap in the diff: `TIMELINE_CONTENT_MAX_WIDTH = 768` in `MessagesTimeline.logic.ts` still drives `resolveTimelineMinimapHasPersistentGutter` and `resolveTimelineMinimapHitStripWidth`. Those assume a centered column at most 768px wide, and the comment above the hit-strip helper warns that a too-wide strip sits on top of the message text and swallows its pointer events. With `full`, at a 2560px viewport the helper still computes a ~896px gutter and returns the full 40px strip while the real gutter is 0, so the strip covers the first ~40px of the message column. The setting needs to feed those helpers, or they should measure the lane element.

saphid:
@otavio I opened https://github.com/otavio/t3code/pull/1 against `feat/chat-width-setting` to close the minimap gap @v1b3coder described above. If you merge it, this PR picks it up.

- The minimap gutter and hover-strip helpers now take the rendered column width instead of assuming 768px, and they re-measure when `chatWidth` changes.
- The minimap's prev/next turn buttons ignore the pointer when the gutter is too narrow to hold them. Before this, they also covered message text at Full width.

I measured it in a real dev build. At Full width, the 40px strip used to overlap the column by 26px, so clicks on the text's left edge landed on the strip. It now shrinks to 14px with no overlap. Comfortable and Wide are unchanged. Focused tests and web typecheck pass. Details are in the linked PR.

Made with Claude Opus 5.5 in Claude Code, running inside T3 Code.

Issue #4880, opened 2026-07-29
[Bug]: the name of forks are the same title as the upstream repo


Project516:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/desktop

### Steps to reproduce

1. fork a repo and rename it something
2. try to open both repos and the fork in t3code

### Expected behavior

t3code will show the fork name different from the upstream repo

### Actual behavior

t3code will call both the repo and its fork by the same name, making it hard to see if you are working on a fork or the actual repo

### Impact

Blocks work completely

### Version or commit

latest nightly (  T3 Code Nightly 0.0.32-nightly.20260729.951 (a8e05cbb9263) T3 Code Nightly 0.0.32-nightly.20260729.951 (a8e05cbb9263) Pre-release)

### Environment

Fedora 44

### Logs or stack traces

```shell

```

### Workaround

`settings` -> `project grouping` turn that off (thanks @Baker)

Baker:
This is happening for me as well, and I won't lie has caused T3 Code to be unusable. 

Examples: 

<img width="1100" height="780" alt="Image" src="https://github.com/user-attachments/assets/b240d673-3607-4242-8cd5-a9de59775533" />

<img width="1100" height="780" alt="Image" src="https://github.com/user-attachments/assets/3feb5346-ad2a-4e8e-b03a-80c06273a881" />

Baker:
@Project516 I found a workaround at least for my project, `settings` -> `project grouping` turn that off.

Project516:
> [@Project516](https://github.com/Project516) I found a workaround at least for my project, `settings` -> `project grouping` turn that off.

Works for me! Thanks!

MaximilianMauroner:
I am still seeing confusing behavior on `v0.0.37-nightly.20260829.1219`.

Renaming the project to `fleet` saves correctly. The new name appears in Project settings and in the thread header. However, the project picker continues to show the repository-derived name, `maximilianmauroner/...`.

Project settings describes this value as “The shared name for this project group in the sidebar and thread lists,” so I expected the project picker to use `fleet` as well. Showing two names for the same project makes it seem as if the rename only worked in some places, and it makes the project harder to find.

The project picker still shows the repository-derived name:

![Project picker showing the repository-derived name](https://tools.mauroner.net/files/3wEF6eCq8RQZ42KogaA4VdIO8uFyvm2K/project-picker.png)

[...]

PR #11596, opened 2026-09-13
fix(clients): preserve shared project names matching repository names

SpyrosPsarras:
## Problem

In the desktop app connected to multiple environments, a grouped project stored as `epaflix` displays as `spyrospsarras/epaflix`. Entering `epaflix` in Settings → Projects → Name appears to revert when navigating away and back.

The stored title is already correct. `deriveProjectGroupLabel` discards a shared title when it matches the repository name, while the settings rename handler correctly treats identical stored titles as a no-op.

## Fix

Prefer the shared project title whenever the members agree, including repository short names. Keep the repository metadata fallback when titles differ. This fixes the shared client-runtime logic used by web, desktop, and mobile without changing persistence or wire contracts.

This deliberately changes the automatic label for groups whose member titles all equal the repository short name: they now show that stored short name. There is no separate explicit-rename flag in the current model.

## Verification

- `vp test run packages/client-runtime/src/state/projectGrouping.test.ts`: 13 passed.
- `vp test run apps/web/src/components/settings/ProjectSettingsPanel.logic.test.ts`: 3 passed.
- Targeted `vp lint` and `vp fmt --check` on both changed files passed.
- `vp run --filter @t3tools/client-runtime typecheck` passed, with existing advisory diagnostics elsewhere in the package.
- Two read-only reviews passed, covering standards and the reported behavior.

### Before and after

These are **isolated browser reproductions**, using the actual upstream and patched label functions with two environments whose stored project titles are `epaflix`. They are not full desktop application screenshots. Full desktop navigation was not exercised in this Linux server environment.

| Before | After |
| --- | --- |
| Before: owner/repository replaces the stored name | After: stored project name is retained |

Please review and merge if this shared-title precedence matches the intended project naming behavior.

Implemented with GPT-6 Astra through OpenCode in T3 Code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Project groups now display the shared project title when all grouped projects use the same title.
  - Groups with differing project titles continue to use the repository label until the titles agree.
  - Label behavior is now consistent across repository modes and regardless of project ordering.


shunkakinoki:
Can confirm this is the fix. `main` still has the guard at [`deriveProjectGroupLabel`](https://github.com/pingdotgg/t3code/blob/5a61f50cce12685f70d530aa5fdf83d9c9b065d7/packages/client-runtime/src/state/projectGrouping.ts#L167-L174), and it rejects a shared title that matches either the repository `displayName` or `name`.

The collision also hits non-fork setups: a repo cloned as `owner/owner` with two grouped checkouts (clone + worktree) saves `owner` fine but renders `owner/owner` in the sidebar. Details in #4880.

Your change keeps the stored title and drops the repo-metadata branch, which is what I'd expect. +1 for merge.

PR #11614, opened 2026-09-13
fix(mobile): prevent text clipping on iOS chat messages

### LINKED ISSUE #11200: [Bug]: iOS Text clipped and 30% not visible
### Before submitting

### Area

apps/mobile

### Steps to reproduce

* Install iPad app version 1.1.0
* Start a session
* Look at the blue message bubbles or mid turn messages
* They are not showing 30% of the text

Image

### Expected behavior

* Text area full width or at least show the full text in the smaller area

### Actual behavior

* Text clipped, 30% hidden

### Impact

### Version or commit

iOS app version 1.1.0

### Environment

11 inch iPad pro iOS26.7

### Logs or stack traces

```shell

```

[IMG_6204.png](https://github.com/user-attachments/assets/1906e7ec-1640-4170-a5d0-d01af01d07bb)

### Workaround

Rotate device twice you get a permanent 30% white area on the right but the text is now readable in remaining space

Exotic209093:
T3MarkdownTextShadowNode uses NSLayoutManager with usesFontLeading = NO to match RN measurement, but usedRectForTextContainer then underreports height by excluding font leading. Multi-line text gets clipped by ~30%. Added a 15% line-height buffer to the measured height before clamping to compensate without affecting Android or web.

Fixes #11200

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved vertical sizing for rendered Markdown text on iOS.
  - Measurement now accounts for line-height information across all text fragments, helping prevent clipping and overly tight content.
  - Applied spacing based on the largest applicable line-height value while ignoring non-positive values.
  - Width sizing remains unchanged, and results continue to respect existing layout constraints.


juliusmarminge:
Closing: the premise does not hold. Both the measuring layout manager (`T3MarkdownTextShadowNode.mm`) and the rendering `UITextView` (`T3MarkdownText.mm`) set `usesFontLeading = NO`, so measurement and rendering already agree. #11200 is a ~30% clip in iPad split view with a rotate-to-fix workaround, which is a width/relayout problem (see the triage notes on `AdaptiveWorkspaceLayout`), not a few points of missing leading, and an unconditional 15% buffer grows every single-line node in the feed. If there is a genuine last-line descender clip, `boundingRectForGlyphRange:` or enumerating line fragments gives an exact height; a before/after screenshot would be needed either way.

PR #11616, opened 2026-09-13
fix(web): allow reverting interrupted turns to edit and restart

### LINKED ISSUE #11083: [Bug]: Cannot stop a running turn, edit the last prompt, and restart
### Before submitting

- [x] I included enough detail to reproduce the problem.

### Summary

A workflow I regularly use in Cursor and Codex is:

1. Send a message and let the agent start working.
2. Notice immediately that the prompt is incorrect or incomplete.
3. Press **Stop** to interrupt the running generation.
4. Return to the last user message and edit or revert it.
5. Send the corrected prompt and restart from there.

I cannot complete this workflow in T3 Code.

While the turn is running, **Revert to this message** is unavailable. After trying to stop the turn, I cannot reliably return to the message that started it, edit or revert that message, and immediately retry with a corrected prompt.

Earlier user messages may have a **Revert to this message** action, but the message that started the current generation does not provide the stop → edit/revert → retry workflow.

### Steps to reproduce

1. Start a thread and send a prompt that causes the agent to begin working.
2. While it is generating, notice that the prompt contains a mistake.
3. Press **Stop**.
4. Try to return to the last user message.
5. Try to edit or revert that message and send a corrected version.

### Expected behavior

After pressing **Stop**, I can return to the last user message, correct it, and restart the turn.

If the agent already modified the repository, T3 Code should either restore the previous checkpoint or let me choose whether to keep the partial changes.

### Actual behavior

I cannot perform the complete stop → edit/revert last message → retry workflow.

The revert action is available for some previous messages, but not for the latest message that started the interrupted or running turn.

### Setup

- T3 Code Alpha desktop app on macOS
- T3 server running on a remote Nebius machine
- Desktop connected to the remote server through T3 Connect
- Cursor agent/provider
- Repository and agent processes are running on the remote machine
- The same T3 setup is also accessed from the iOS app

### Related issues

These appear related, but none describes the complete user workflow above:

- #7982 — the revert button can be missing after a failed turn
- #4524 — interrupt failures can be hidden while Stop appears successful
- #8618 — reliability and feedback problems when stopping a remote turn
- #5667 — discusses the common “notice a turn going wrong, stop, and clarify” workflow, but focuses on the missing keyboard shortcut

### Impact

Frequent workflow disruption

Exotic209093:
Interrupted turns marked their checkpoint as missing, so the Revert button never appeared for the latest user message. Preserved the pre-turn baseline checkpoint as ready when a turn is interrupted, and updated the revert map computation to fall back to it when no completed turn diff exists.

Fixes #11083

## Summary by CodeRabbit

* **Bug Fixes**
  * Interrupted or cancelled turns now preserve the pre-turn checkpoint, keeping it available for reverting changes.
  * The **Revert** option remains visible when no completed turn diff is available.
  * Timeline entries correctly reflect the applicable baseline checkpoint.
  * Reverting after an interrupted or cancelled turn now provides a reliable starting point for restarting the turn.
  * Baseline checkpoint selection is limited to the latest affected message to prevent reverting to an incorrect checkpoint.


Exotic209093:
Rebased onto latest main and resolved conflicts. The revert-turn-count logic had moved from ChatView.tsx into buildRevertTurnCountByUserMessageId in MessagesTimeline.logic.ts, so the interrupted-turn fallback was re-implemented there; targeted tests pass (the 4 failing CheckpointReactor tests are pre-existing CRLF line-ending failures in main's own tests, unrelated to this change).

Exotic209093:
Rebased onto latest main and resolved conflicts. The revert-turn-count logic had moved from ChatView.tsx into buildRevertTurnCountByUserMessageId in MessagesTimeline.logic.ts, so the interrupted-turn fallback was re-implemented there; the new tests pass locally (4 pre-existing CRLF line-ending failures in main's own CheckpointReactor tests are unrelated to this change).

juliusmarminge (changes requested):
The server change is right: the interrupt checkpoint is a real ref with a real diff, so `ready` is the honest status and this makes the `turn.completed{interrupted}` path consistent with `turn.aborted`. Two things before this can land:

1. **The typecheck failure is this PR's.** `CheckpointReactor.test.ts:2339` reads `.status` off a checkpoint that the `waitForThread` helper types as `{ checkpointTurnCount }` only. It is not a pre-existing CRLF issue; widen the helper's type (or read the checkpoint from the projected thread) and run `vp typecheck` in `apps/server` before pushing.

2. **The web fallback can map the wrong message.** In `MessagesTimeline.logic.ts` the fallback takes the *last user message without a completed diff* and assigns it `turnDiffSummaries[length - 1]`. Sequence: user-1 → turn 1 completes; user-2 → turn 2 interrupted (synthetic assistant id, so unmatched); user-3 → turn 3 completes. user-2 is the last unmatched message, gets `revertTurnCount = 2`, and "Revert to this message" restores the state *after* turn 2 instead of before it. Fix: only fall back when the unmatched user message is the last user message in the timeline, and pick the summary by turn linkage (matching `turnId`, or max `turnCount`), not by array position. Add a test for exactly that unmatched-then-matched sequence; the current "only latest" test does not cover it.

Also drop the stray blank-line change in `ChatView.tsx`.

juliusmarminge:

PR #11624, opened 2026-09-13
fix(server): prevent preview_snapshot from bricking sessions with oversized images

### LINKED ISSUE #11295: preview_snapshot includeImage embeds full-res base64 PNGs into tool history; providers that reject inline images brick the session permanently
## Problem

`t3-code_preview_snapshot` with `includeImage:true` embeds full-resolution base64 PNG `data:` URLs directly into the MCP tool result (`state.attachments`). Those bytes persist in agent session history and are replayed to the model provider on **every** subsequent request.

Some providers reject inline image uploads. Observed case: opencode.ai zen gateway → upstream `Console`, model `muse-spark-1.3-contributor-free`, returns non-retryable `400 invalid_request_error: Invalid upload request` on all later requests — including trivial text-only prompts. The session is permanently bricked; only surgery on the session store (deleting the attachments) recovers it.

Concrete numbers from the incident: two snapshots produced ~1.35MB + ~50KB base64 attachments (1280×800 PNGs), taking session tool history to ~2MB.

## Requests

1. Downscale/compress screenshots before attaching (e.g. max-width + JPEG quality params on the snapshot tools), so the common case stays small.
2. Prefer returning a file path / saved-file reference plus text over inline base64, letting the agent re-read the image only when vision is actually needed.
3. Default `includeImage` to false (or document text-first usage) so agents don't attach images unless the task needs them.

Related upstream tracker (provider-side rejection): https://github.com/anomalyco/opencode/issues/47237 — this issue is the T3-side mitigation for it.

Built from a real incident diagnosed via Muse Spark through T3 Code.

Exotic209093:
The preview_snapshot MCP tool embedded full-resolution base64 PNGs directly into tool history as inline image content blocks. When providers reject inline images, this permanently bricks the session because the oversized payload cannot be removed from history. Removed the inline image content block from snapshot results; structured metadata (dimensions, mime type) is still returned for reference while the client renders the actual image from its own copy.

Fixes #11295

## Summary by CodeRabbit

* **Bug Fixes**
  * Snapshot previews now return textual JSON metadata and structured screenshot information instead of inline image data.
  * Snapshot responses consistently contain text-only content for improved compatibility with clients that cannot process embedded images.
  * Saved screenshot results now provide the screenshot path as a structured field.
  * The `includeImage` option no longer returns embedded image content.


Exotic209093:
Rebased onto latest main and resolved conflicts: kept main's bounded snapshot text content structure while removing the inline image content (the PR's fix). Updated the tests to match. Targeted test run (apps/server/src/mcp/McpHttpServer.test.ts) passes 15/15.

juliusmarminge:
Closing: this removes the inline image entirely, which takes away the agent's ability to see the page at all (nothing on the client renders snapshot pixels from tool results, so the description's "client renders the actual image from its own copy" is not accurate). #11295 asks for a size cap / default-off, not removal. `device_screenshot` keeps the same exposure. The `Test Server 2` failure is caused by this change: with `includeImage` removed from the schema, the existing "rejects non-boolean snapshot image options" test no longer rejects the input. Keeping `includeImage`, defaulting it to false, and/or downscaling to a bounded width would be the ~10-line version of this.

PR #11628, opened 2026-09-13
feat(antigravity): queue follow-up turns sequentially instead of cancelling active prompt

willblanchard:
### Summary

Matches the user experience of Codex and Claude Code where follow-up messages are queued behind an in-flight turn rather than aggressively aborting and cancelling the active prompt.

### Problem & User Impact

Currently, submitting a message in the composer while an Antigravity turn is active aggressively cancels the active prompt:
- **Loss of In-flight Work:** If the user sends a follow-up thought, clarification, or subsequent task while an agent has been running (e.g. 5–7+ minutes of reasoning, running tests, or coordinating subagents), the in-flight work is destroyed mid-stream.
- **Subagent Cascades:** Subagent tasks are abruptly marked cancelled, commands are killed mid-execution, and partial results are lost.
- **UX Discrepancy:** In Codex and Claude Code, follow-up messages are queued into the workflow so the user can freely queue up additional guidance or subsequent instructions without having to wait for the agent to finish before typing. In Antigravity, typing anything into the composer instantly aborts the model's work.

### Solution

- Adds `turnQueue: Semaphore.Semaphore(1)` to `SessionContext` in `AntigravityAdapter` to serialize turns sequentially FIFO.
- Validates model selections upfront so invalid model configurations fail fast without holding the queue.
- Follow-up turns wait until the active prompt finishes, each emitting its own distinct `turn.started` and `turn.completed` lifecycle events.
- In-progress subagents remain intact across queued turns.
- Explicit cancellation via the Stop button (`interruptTurn`) continues to interrupt and cancel immediately via `promptLock`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Follow-up turns are now queued until the current prompt completes, preventing interruptions and preserving turn order.
  * Queued follow-up turns inherit the session’s most recently selected model when no model is specified.
  * Each queued turn now receives its own lifecycle, improving completion tracking and status reporting.
  * Open subagent calls settle correctly when a session is stopped or a turn configuration fails.


juliusmarminge:

PR #11629, opened 2026-09-13
fix(web): show all project instances in filter across multiple machines

### LINKED ISSUE #11201: [Bug]: Project filter only shows one entry when the project exists on multiple machines
### Before submitting

### Area

apps/desktop

### Steps to reproduce

1. Open a T3 Code instance that is connected to a remote environment
2. Clone the same repo on the host machine and remote machine (same folder name)
3. Add both projects
4. Create a new session and filter for the project name

### Expected behavior

The project shows up twice, since it exists on two environments,

### Actual behavior

One the last project that was added is showing up.

### Impact

Cosmetic issue

### Version or commit

0.0.39

### Environment

Omarchy

### Logs or stack traces

```shell

```

_No response_

### Workaround

_No response_

Exotic209093:
The sidebar project filter grouped projects by repository identity, collapsing multi-environment instances into a single entry. Added a flat projectFilterEntries list that shows each machine instance separately with environment-disambiguated labels, and updated scope resolution to use physical scope keys so selecting one instance filters threads to only that specific machine. Grouped projectGroups are preserved for thread list rendering and settings navigation.

Fixes #11201

## Summary by CodeRabbit

* **New Features**
  * Project filters now list each physical project instance separately, including local and remote environments.
  * Duplicate project names are disambiguated with environment labels.
  * Project instance icons are shown in the filter trigger and project list.

* **Bug Fixes**
  * Thread filtering now applies only to the selected project instance.
  * Improved project selection and filtering when multiple environments share the same project.


Exotic209093:
Rebased onto latest main and resolved conflicts: main rewrote the project scope filter into a searchable Combobox, so the per-instance entries were re-implemented against the new component (flat environmentId:projectId scope items, per-instance favicons, single-key scoping). environmentGrouping tests pass and apps/web typecheck is clean.

Exotic209093:
Rebased onto latest main and resolved conflicts: main rewrote the project scope filter into a searchable Combobox, so the per-instance entries were re-implemented against the new component (flat environmentId:projectId scope items, per-instance favicons, single-key scoping). environmentGrouping tests pass and apps/web typecheck is clean.

juliusmarminge:
Thanks for taking this on. Closing rather than iterating, because the change breaks the other writer of the scope key: the combobox now stores `envId:projectId`, but the thread-row context menu ("Filter by project") still stores the group key, so selecting it from a thread empties the list and the fallback effect resets the scope to "all". Any persisted `sidebarProjectScopeKey` from before this change gets wiped the same way. Mapping `projects` directly also drops the `sidebarProjectSortOrder` handling and the stale-registration dedup that `buildProjectGroups` does. If you want to pick this up again, flattening `projectGroups.flatMap(g => g.memberProjects)` and updating both scope writers to the same key format would be the shape to aim for. The added test re-implements the label logic inline and asserts against itself, so it would not catch any of this.

PR #7220, opened 2026-08-16
fix(web): render Codex visualizations inline

marijnbent:
Fixes #7191

## Problem

Codex visualize references appeared as raw marker text, so users had to find and open the generated HTML file themselves.

## Fix

- Parse canonical and replacement-character visualize references while leaving fenced examples as Markdown.
- Issue signed URLs only for bounded temporary HTML artifacts referenced by an assistant message.
- Bind each URL to the artifact content, serve it as inert text, and execute it only in a sandboxed iframe with no network access.
- Keep normal messages aligned while allowing only wide visualizations to extend beyond the chat column.

## Proof

Before:

Raw visualize marker before this fix

After: see the proof comment below for a real Codex-generated visualization rendered inline.

## Verification

- 45 focused tests passed across asset access and timeline rendering.
- Shared, contracts, web, and server type checks passed.
- Focused lint, formatting, and `git diff --check` passed.
- The web production build passed.
- An integrated Chrome test verified wide-mode panel containment and light/dark theme bridging on a real Codex `visualize` result.

Implemented with GPT-5.6-Sol through Codex in T3 Code.

> [!NOTE]
> ### Render Codex visualizations inline in assistant messages
> - Assistant messages containing inline visualization markers are parsed by `splitInlineVisualizations` (new shared module) and rendered as sandboxed iframes via the new `InlineVisualization` component, replacing raw marker text.
> - The server validates and signs visualization HTML fragments (temp-dir only, single hardlink, ≤1 MiB, HTML/XHTML extension, SHA-256 revision-locked) before issuing asset URLs, and serves resolved assets as inert text with a restrictive CSP.
> - `ProjectionSnapshotQuery.hasAssistantVisualizationReference` is added to confirm a path is referenced by an assistant message before a signed URL is issued, preventing unauthorized access.
> - The desktop app's CSP is updated to permit `blob:` in `frame-src` so blob-URL iframes can load in Electron.
> - Risk: visualization files must reside under the OS temp directory; files outside that root are rejected at issuance time.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3164bfe.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Untrusted HTML is executed in the browser inside sandboxed iframes with layered CSP and server-side path gating, but any sandbox or CSP gap would be client-side code execution.
> 
> **Overview**
[...]

PR #11634, opened 2026-09-13
feat: show HTML artifacts inline in assistant replies

CouchRiv:
Follow-up to #7220 (@marijnbent), rebuilt around its closing review.

When an agent builds an HTML page, its reply only shows a path, and the reply changes or breaks when that file is edited or deleted.

## What Changed

With **Interactive artifacts** on (Settings → Integrations, off by default, per-project override), agents are taught a `t3-artifact` fence naming a workspace HTML file. When the run ends, the server copies each named file (max 4, 1 MB each) into the thread as a text attachment and records it on the message under a command receipt, like `checkpoint.capture`. Web, desktop and mobile render the saved copy in place: a sandboxed `srcDoc` iframe on web/desktop, a WebView that refuses navigation on mobile. Pages use the MCP Apps theme variables, size and link messages; form values and `window.t3.setState` data survive remounts while the app is open.

| #7220 concern | Here |
| --- | --- |
| Temporary provider files | Copied when the run ends |
| Not bound to the message | Recorded by fence position and path |
| Theme change rebuilt the frame | Sent to the running page |
| No mobile | Rendered on mobile |
| Auth rescanned message text | Ordinary thread attachments |

- **Providers:** Codex, Claude, Cursor, OpenCode and ACP (including Grok and Antigravity) are taught the fence; Pi is not.
- **Setting off:** no instructions and no capture; artifacts already in a thread stay and can be hidden.
- **Security:** links open only after a click (mobile asks first); an artifact that leaves its page is stopped; scripts cannot fetch from the network.
- **Limits:** a frame can start one navigation before it is stopped; state is not kept across reloads; provider-native subagent replies are not captured.

Three commits (server, web, mobile). About 2,500 non-test lines; happy to split.

## UI Changes

Before/after screenshots and a short video (artifact appears, interaction, scroll away and back, theme switch, mobile) will be added before this leaves draft.

## Checklist

- [ ] This PR is small and focused (it is not; see above)
- [ ] I included before/after screenshots for any UI changes (before leaving draft)
- [ ] I included a video for animation/interaction changes (before leaving draft)

Implemented with Claude Opus 5 in Claude Code and GPT-6 in Codex.


juliusmarminge:
Closing this one, and thank you for the careful rebuild of #7220; the sandboxing and copy-on-run-end design here is well thought through.

The reason is direction rather than quality. We're going with MCP Apps as the way agents render interactive UI inside a reply: the agent's tools return UI resources through MCP and the client hosts them, instead of T3 injecting an artifact protocol into every provider's runtime prompt and capturing HTML files off the workspace. I have a draft of that which will land shortly after v2 (`t3code/codex-turn-mapping`) merges into main. Once it's up I'll link it here.

If you're interested in that model, the parts of this PR that would carry over are the sandbox posture (`allow-scripts` without `allow-same-origin`, CSP meta) and the mobile WebView host, so keep an eye out for the draft.

PR #11642, opened 2026-09-13
fix(server): stop borrowing capabilities for custom Codex models

samaluk:
A manually entered Codex model currently inherits controls from the first discovered model. That can offer a reasoning effort the custom model does not support, and refreshing the provider can bring back stale controls after they were removed.

Keep custom-model capabilities unknown unless explicitly configured, preserve exact discovered matches, and let the refreshed custom entry replace previously borrowed capabilities.

Validation: focused CodexProvider and ProviderRegistry tests, server typecheck, targeted lint and server bundle. This changes shared server capability data; no web, desktop or mobile UI files are changed.

Prepared with GPT-6 Astra in Codex.


juliusmarminge:

PR #11684, opened 2026-09-14
perf: cut allocation churn in revert fallback ordering

Priyansh4444:
## Why

Reverting a thread sorts the messages that were not already kept, oldest first, and keeps the first few. Every comparison in that sort re-read and re-validated both date strings from scratch, so the same dates were parsed hundreds of thousands of times on a large thread. On a 50,000-message thread that parsing was the dominant cost of the revert.

An earlier revision of this branch made that faster by parsing each date once, but it wrapped every message in a small object that stayed alive for the whole sort. That traded parsing for garbage: the collector had to move those objects into the long-lived heap, which costs more per collection. This revision parses each date once and keeps only the messages that will actually be retained, so almost nothing survives to the collector.

## Scope

`packages/shared/src/dateTime.ts` gains `selectFirstByDateTime`. Give it items, a way to read each item's date string, a count, and a tie-breaker, and it returns the requested number of oldest items in the same order the existing `compareDateTimeStrings` defines: malformed dates first, then by instant, with ties broken by the caller. It does not sort the whole input.

Three call sites use it: `retainThreadMessagesAfterRevert` (`apps/server/src/orchestration/projector.ts`), `retainProjectionMessagesAfterRevert` (`apps/server/src/orchestration/Layers/ProjectionPipeline.ts`), and `retainMessagesAfterRevert` (`packages/client-runtime/src/state/threadReducer.ts`). The client also stops doing the work when there is nothing left to keep; the server paths already had that guard.

## Blast Radius

Which messages are kept when a thread is reverted. The selection uses the same ordering rules as before, so results and ties are unchanged.

## Verification

- `vp test run` on the four touched suites: 137 passed.
- Targeted `vp run typecheck` for shared, client-runtime, and server: clean. `vp lint` and `vp fmt --check`: clean.
- Results are identical to the original on a 4,015-value mixed date corpus and 65 hand-written and randomized edge cases.
- 50,000 messages with about 13,400 candidates: 768ms to 49ms warm, 909ms to 67ms cold; GC time 5.6ms to 0.6ms per call; peak memory 177MB to 137MB.
- Nothing to keep (`missingCount` 0): 76ms to 6ms on the same thread.

See the comment for the full comparison of approaches and why this shape was picked.


Priyansh4444:
## What this changes, in plain terms

When you revert a thread, the app decides which messages to keep. For each role it gathers the messages that were not already kept, sorts them oldest first, and keeps the first few.

Sorting compares messages in pairs, over and over. The old code re-read and re-validated both date strings on every single comparison. On a 50,000-message thread that is hundreds of thousands of date parses to pick a few hundred messages, and it was the dominant cost of the revert.

This PR does three things:

1. Parse each date once, then compare numbers.
2. Keep the few oldest messages directly, instead of sorting every candidate and then taking the first few.
3. Skip the work entirely when there is nothing left to keep. The web and mobile path did not have that check; the server paths did.

## Why it is lighter, not just faster

The first version of this branch only did (1). It was much faster, but it wrapped each message and its parsed date in a small object so the sort could use it. Those objects stayed alive for the whole sort, so the garbage collector moved them into the long-lived heap. That means *fewer* collections but *more expensive* ones, including two full collections and a longer pause (see the table). The current version keeps only the messages that will actually be retained, so there is almost nothing left for the collector to clean up.

## Measurements

50,000-message thread, about 13,400 candidates per role, about 800 kept per role. One copy of the code per process, same inputs, median of repeated runs.

| version | warm (JIT on) | cold (JIT off) | GC time per call | longest GC pause | full GCs | peak memory |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| original | 768ms | 909ms | 5.6ms | 2.2ms | 0 | 177MB |
| first try: parse once + wrapper objects | 61ms | 73ms | 2.0ms | 3.5ms | **2** | **215MB** |
| **this PR** | **49ms** | 67ms | **0.6ms** | **0.45ms** | **0** | **137MB** |

Normal threads are much smaller, and there the fallback is a few milliseconds in every version, so the visible win is the "nothing to keep" check: on the 50,000-message thread it goes from 76ms to 6ms.

Other approaches we measured and did not pick:

| approach | warm | cold | GC per call | full GCs | peak memory |
| --- | ---: | ---: | ---: | ---: | ---: |
| one wrapper object per candidate | 55ms | 68ms | 0.8ms | 0 | 174MB |
| plain loop, sort all candidates | 54ms | 66ms | 0.5ms | 0 | 158MB |
| sort positions into a numeric array | 51ms | **62ms** | 0.6ms | 0 | 172MB |
| **keep the oldest few directly (this PR)** | **49ms** | 67ms | 0.6ms | 0 | **137MB** |

## The close call

Sorting into a numeric array is about 8% faster when the code is cold, and a revert is a one-off call so it may never get warm. It was the runner-up. We picked "keep the oldest few directly" because it uses about 25% less peak memory, produces no full collections, and is faster once warm. If reviewers care more about cold-start latency than footprint, the numeric-array sort is a drop-in swap behind the same helper.

## What to expect in production

- Normal threads: no visible change. The fallback is already a few milliseconds; there will be fewer GC pauses.
- Very large threads: roughly 10-16x less time spent picking messages, about 10x less time in the garbage collector, and no full collections from this path.
- No behavior change: 137 tests pass, including every revert test, and 65 hand-written and randomized edge cases (malformed dates, equal timestamps, ties) return identical results.

## How this was measured

Node 24 with V8 13.6 on Linux x64. Each version runs in its own process so one cannot affect the other's JIT, and the median of repeated runs is reported. "Warm" is default V8; "cold" runs with the optimizing compiler disabled (`--no-opt`). GC numbers come from Node's GC event observer; memory is peak process memory. The "first try" row is the earlier revision of this PR, kept as a cautionary data point: it was faster than the original and still the worst choice for memory and pauses.

Priyansh4444:
Superseded by a smaller rewrite: parse each date once, keep the oldest messages directly, and skip the fallback when nothing is missing. This one is being closed and a new draft will follow.

PR #11707, opened 2026-09-14
perf: parse revert dates once and keep only the oldest messages

Priyansh4444:
## What changed

The revert fallback used to re-parse both date strings on every comparison, and the client sorted every candidate even when nothing was missing. Now each date is parsed once, only the messages that will be kept are held (in a small heap), and the client skips the work entirely when nothing is missing, matching the server.

## Why it is better

Median time per revert, one process per arm, same inputs:

| case | before | after | |
| --- | ---: | ---: | ---: |
| 200 messages (about 50 candidates, 4 kept) | 1.16ms | 0.25ms | 4.7x |
| 20 messages (about 7 candidates, 1 kept) | 0.16ms | 0.07ms | 2.3x |
| 200 messages, nothing to keep | 1.12ms | 0.03ms | 35x |
| 20 messages, nothing to keep | 0.15ms | 0.01ms | 22x |

With the optimizing compiler disabled (cold) the shape holds: 200 messages 1.38ms to 0.17ms, 20 messages 0.16ms to 0.04ms. Allocation rate at 200 messages falls from 0.04 to 0.01 GC collections per call.

Output is unchanged: 137 tests pass, and 65 hand-built and randomized edge cases (malformed dates, equal timestamps, ties, empty input, limit 0, limit greater than the candidate count) return identical results.

## Files

- `packages/shared/src/dateTime.ts`: adds `selectFirstByDateTime`.
- `packages/client-runtime/src/state/threadReducer.ts`: uses it and skips the fallback when `missingCount === 0`.
- `apps/server/src/orchestration/projector.ts` and `apps/server/src/orchestration/Layers/ProjectionPipeline.ts`: use it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved message restoration after reverting changes, consistently selecting the earliest applicable user and assistant messages.
  - Added more reliable handling for messages with identical or invalid timestamps.
  - Preserved chronological ordering and tie-breaking when restoring messages.

- **Performance**
  - Improved efficiency when selecting messages from larger conversation histories.


pc-style:
@shivamhwp hello our friend needs a review (ik ur busy af)

Priyansh4444:
The main thing it does:
in the sorting function we keep re-parsing the date.
which causes the sorting algorithm to re-parse dates,
this effectively does a single parse, and then the algorithm to optimize on top of that similar to LeetCode 973 :D

juliusmarminge:

PR #11717, opened 2026-09-14
feat(settings): per-project thread title instructions

Radu028:
Implements the consolidated Ideas request in #6836 (see also #7298, where @juliusmarminge scoped the minimum as "a global title-instructions setting for initial titles and Regenerate title, with reset to defaults").

## Problem

Generated and regenerated thread titles only follow the built-in prompt. Users who want a different shape, for example `[COMPASS 4437] Description` whenever the request names a ticket, shorter titles, or another language, have no place to say so. The `threadTitleInstructions` field on the text generation policy existed in code but nothing ever set it (#7298 noted the same gap).

## What changed

- New `threadTitleInstructions` server setting, empty by default. A **Title instructions** row under General > Text generation edits it, with the usual reset.
- The reactor resolves it and passes it into the thread title prompt for both the first turn and Regenerate. The prompt appends it after the message as project instructions that outrank the editorial rules, so a ticket prefix can beat the default length and "no numbers" guidance.
- The key is in `PROJECT_SCOPED_SERVER_SETTING_KEYS`, so the same row also works as a project override through the existing settings-scope mechanism. That is the whole per-project opt-in, no extra machinery.
- The dead `threadTitleInstructions` policy field is removed in favour of a plain `instructions` input on thread title generation.

Empty keeps the default prompt, so nothing changes for anyone who does not set it.

## Before / after

Before: General > Text generation only had the model row. After, at environment scope:

Title instructions row at environment scope

With a project selected in the breadcrumb, the row stores a project override (layers icon reads "Overridden for this project", reset returns to the environment value):

Title instructions overridden for one project

## Testing

- `TextGenerationPrompts.test.ts`: instructions appear in the first-turn and regeneration prompts and are omitted when blank.
- `ProviderCommandReactor.test.ts`: a project override reaches the title generator on the first turn and on Regenerate.
- Typecheck for contracts, shared, server, web and mobile; lint and format on the touched files.
- Real run with Codex as the text generation provider: with the instructions above set on one project, "Investigate compass-4437: after SSO sign-in the app bounces users back to /login in a redirect loop" produced the title `[COMPASS 4437] SSO Redirect Loop`. The value was stored in `settings.json`, not in the repository.

Model: Claude Fable 5.1. Harness: Claude Code in T3 Code.


juliusmarminge:

PR #11721, opened 2026-09-14
fix(server): Claude rewind no longer orphans its history worker on shutdown

faw01:
Rewinding a Claude thread with a custom home reads and forks history in a child process. That process was started outside the rewind's fiber, so shutting the server down mid-rewind left it running on its own. It is now part of the rewind and stops with it. The custom home still reaches the worker as env. Server env is untouched.

Created with Cursor Grok 4.6, implementation reviewed with Fable 5.1.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability for rollback history and fork operations when using alternate Claude configuration directories.
  - Added clearer error reporting for missing sessions, request failures, timeouts, worker crashes, and invalid responses.
  - Interrupted rollback operations now cleanly stop background processing, reducing the risk of lingering work or inconsistent state.

- **New Features**
  - Mobile builds now support selectable `production` and `v2-preview` profiles, with validation for supported preview builds.


juliusmarminge:

PR #11770, opened 2026-09-14
chore(server): keep the legacy service entry point to the npm package only

juliusmarminge:
## What changes

Follow-up to #11732, which fixed a real regression: a boot-service launcher installed before the executable runs `node <versions>/<v>/node_modules/t3/dist/bin.mjs`, and #11607 had removed that file from the `t3` npm package, so every fleet update from a pre-1700 nightly failed after a successful install.

That PR fixed it in two places. This keeps one and removes the other:

- **Kept:** `dist/bin.mjs` in the `t3` npm package, a forwarder to the platform executable that preserves arguments, stdio, the launcher's IPC channel, signals, and exit status. This is the transition an old launcher actually goes through: it npm-installs the new `t3`, runs the old path, the executable starts, and the first server started this way rewrites the service unit to run the executable directly. After that nothing touches the file. The doc comment now says when to delete it (after the first stable release with the executable).
- **Removed:** `ensureLegacyEntry` in `pinnedRuntime.ts`, which wrote the same shim into every *archive* install, including cached ones on every status check, so an old launcher could start an archive-layout version through the old path. An old launcher never installs an archive (it predates archive installs), so that case does not occur, and it left a permanent fake `node_modules/t3/dist` in the runtime layout plus a write to cached installs on read. The `"archive"` variant of the script and its two tests go with it.

Net: 25 insertions, 92 deletions; `legacyCliLauncherScript` has no parameter now.

## Verification

- `legacyCliLauncher.test.ts` (IPC, args, termination through the npm shim), `pinnedRuntime.test.ts`, and `build-npm-platform-packages.test.ts` (which exercises `dist/bin.mjs` from the built tarball): 10 pass.
- Typecheck clean for `apps/server`, `packages/shared`, and `scripts`; knip clean.

Claude Fable 5 via Claude Code.

---

<a href="https://app.devin.ai/review/pingdotgg/t3code/pull/11770" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **Changes**
  - Simplified runtime installation and validation for greater consistency across published runtimes.
  - Updated the command-line launcher to use platform-specific packages directly.
  - Removed support for launching legacy archive-based runtimes and related compatibility behavior.
  - Updated launcher documentation to clarify its compatibility role and planned future removal.


akhial:
## What happened

A server update initiated from T3 Desktop on a MacBook failed twice with:

```text
Server update failed: The requested target runtime is missing or incomplete.
```

Desktop was connected to both its local server and a remote Linux server. The matching failures were recorded by the Linux background service.

## Diagnosis

This machine demonstrates the archive compatibility case removed in #11770. At the time of failure, its old service launcher was still running, while its server child had already moved to an archive runtime.

The old `runtime/service-launcher.mjs` checks for `node_modules/t3/dist/bin.mjs` plus `.install-complete`. The active `0.0.41-nightly.20260914.1722` runtime had that compatibility entry point. The prepared `0.0.41-nightly.20260915.1752` runtime had a working `t3` executable and the correct completion marker, but lacked the old entry point. The launcher therefore rejected it before activation.

The launcher was the same Node process started September 11. The archive download is performed by the newer server child, so an old launcher can encounter an archive installed by a subsequent server version. Both use launcher protocol 2; the target's preflight does not detect this layout incompatibility.

## Steps to reproduce

Observed upgrade sequence, inferred from the installed files and service logs; the remote failure was not deliberately reproduced during triage:

1. Keep an npm-era background service launcher installed.
2. Update remotely through a version that preserves the archive compatibility entry point. This machine successfully reached `0.0.41-nightly.20260914.1722`.
3. From Desktop, update that server to `0.0.41-nightly.20260915.1752`.
4. The new server updater stages the archive without the compatibility entry point; the old launcher rejects it. Retrying produces the same error.

## Version

Server at failure: `0.0.41-nightly.20260914.1722`. Failed target and triage CLI: `0.0.41-nightly.20260915.1752` (source commit `50ff4c371eab927a9650c114975241999f4cd7b1`).

## Environment

Linux x64, kernel `6.12.107+deb13-amd64`, systemd user service. Old launcher used Node `24.21.0`; triage CLI was invoked using Node `26.8.2`. Client: T3 Desktop on a MacBook; macOS and Desktop versions were not collected.

## Evidence

```text
2026-09-15 10:51:57 UTC: server.updateServerWithProgress -> ServiceLauncherRejectedError
2026-09-15 10:52:17 UTC: server.updateServerWithProgress -> ServiceLauncherRejectedError
Reason: The requested target runtime is missing or incomplete.

Target runtime 0.0.41-nightly.20260915.1752:
  t3 --version                         -> t3 v0.0.41-nightly.20260915.1752
  .install-complete                     -> matches target version
  node_modules/t3/dist/bin.mjs          -> absent

Active runtime 0.0.41-nightly.20260914.1722:
  node_modules/t3/dist/bin.mjs          -> present
```

## Related issues

#11732 originally covered subsequent archive updates under an older launcher. #11770 removed that coverage and archive entry point. Posting here to provide the concrete installed state that contradicts the removal's assumption. No separate matching bug report or newer fix was found in the searches performed.

## Fix applied or workaround

With the user's approval, repaired the Linux service using:

```sh
npx t3@0.0.41-nightly.20260915.1752 service install
```

The command succeeded. Verified that systemd now runs the target executable with `__service-launcher`, the service is active, `service-state.json` selects `0.0.41-nightly.20260915.1752`, and the local HTTP endpoint returns 200. Desktop reconnection and a subsequent remote upgrade were not tested.

## Filed by

Prepared by Codex (GPT-6) via `t3 triage`.

PR #11782, opened 2026-09-14
fix(server): sync worktree branch before turns

Pastequee:
The sidebar reads the branch stored in thread metadata. Renaming the checked-out branch outside T3 left that metadata stale until a provider turn completed, and sending another message could not repair it when no provider session was available.

Refresh local VCS status from the thread worktree when a real user message starts the next turn, then reuse the guarded worktree drift reconciliation to update the thread shell. The existing ownership checks still reject shared worktrees, detached HEADs, and temporary checkout targets.

Tests:
- pnpm --filter t3 exec vitest run src/orchestration/Layers/CheckpointReactor.test.ts
- pnpm --filter t3 typecheck
- node_modules/.bin/vp lint --report-unused-disable-directives apps/server/src/orchestration/Layers/CheckpointReactor.ts apps/server/src/orchestration/Layers/CheckpointReactor.test.ts

Generated with gpt-5.6-sol using the Codex harness.

## Summary by CodeRabbit

* **Bug Fixes**
  * Threads without an active provider session now recognize when their worktree branch has changed locally.
  * The updated branch is adopted when the next user message starts a turn, keeping thread activity aligned with the current checkout.
  * Missing worktrees or branch-status refresh failures no longer interrupt message processing.


juliusmarminge:

PR #11791, opened 2026-09-14
feat(omp): Oh My Pi provider with full terminal parity

Peteralm:
Adds Oh My Pi (`omp`) as a first-class provider, driven through its ACP server (`omp acp`).

## What works

- **Skills and commands.** `$name` lists every skill omp discovered (74 here) and `/` lists its own commands (114 here), including plugin and project commands. Both come from one `omp --mode rpc` probe per refresh, cached per cwd and invalidated by the live `available_commands_update`.
- **Models.** The picker is omp's own catalog (121 models here), grouped by upstream provider, with per-model thinking ladders. The model omp has selected is marked default, which is what lets the composer resolve a model and show the thinking control at all.
- **Context and compaction.** `usage_update` feeds the context meter with omp's own window size; **Compact** runs `/compact`.
- **Usage and auth.** Provider card reads `omp usage --json` (its `generatedAt` is epoch-millis, not a string) and shows the authenticated accounts.
- **Self-update.** `omp update --check` drives the advisory; omp is its own updater.
- **Session import.** The onboarding scan reads omp's transcripts and offers its directories as projects; threads resume in place through `session/load`.
- **Onboarding.** Oh My Pi sits next to Claude Code and Codex in **Setup**, with its own installer and `omp setup` sign-in.
- **`/rename`.** omp's session title becomes the thread title.
- **`/fresh`.** omp swaps its provider session id on the same connection; the runtime adopts the new id instead of going silent for the rest of the thread (opt-in per provider).
- **Argument autocomplete.** Commands advertise hints like `[soft|remote|snapcompact]`; the composer offers those choices.

## Fixes that are not omp-specific

- Terminal escape sequences in agent message text are stripped on the omp path, so `/context` bars render as text instead of `[38;2;107;114;128m` noise.
- A `session/load` that dies now fails the start instead of leaving the turn hanging forever with no error.
- The Windows Electron runtime extraction no longer requires `python3` (uses `Expand-Archive`).

## Limits, documented in `docs/user/providers-oh-my-pi.md`

- No rewind: ACP exposes no branch-from-entry call.
- Some omp commands (`/instinct-*`) only draw in its terminal; the turn carries a note naming the silent command.
- `/wt` and `/move` change omp's directory, not the thread's project.

## Verification

Server suites around the provider, ACP runtime and session import pass, plus the composer suites on the web side; typecheck is clean. Everything above was also exercised in the desktop app against a real omp 18.1.18: turns, approvals, subagents, image input, interrupt, resume after restart, `/review`'s interactive picker, and compaction.

## Summary by CodeRabbit

* **New Features**
  * Added Oh My Pi as a supported provider with setup, authentication, model selection, usage status, skills, slash commands, and session import.
  * Added Oh My Pi onboarding, provider branding, installation guidance, permission-mode mappings, and text-generation support.
  * Added slash-command argument suggestions in the composer.
* **Bug Fixes**
  * Explicit provider-renamed thread titles now correctly replace generated titles.
  * Improved provider display names and skill instruction link handling.
* **Documentation**
  * Added Oh My Pi installation, setup, usage, feature, and limitation guidance.


Peteralm:
Addressed the review:

- `OmpAnsi.ts`: the partial-tail pattern now requires an *unterminated* OSC, so text printed after a closed `ESC ]…BEL` is no longer withheld and dropped. Two tests cover both halves.
- `composer-logic.ts`: an argument completion now replaces the whole token, not just the part before the caret.
- `OmpTextGeneration.ts`: a cancelled prompt fails before decoding, so partial JSON can no longer be returned as a result.
- `OmpDriver.ts`: a cache hit keeps the probe's own timestamp, so the 30s freshness window still expires for a polled cwd. New test polls at 20s/29s (cached) then 31s (re-probe).
- Docs (`providers-oh-my-pi.md`, `internals/providers.md`, `OmpMaintenance.ts`): `omp update` delegates to the installer it detects (Homebrew, mise, Bun, npm, direct binary); the old 'no other installer is involved' wording was wrong.
- `ComposerPromptEditor.test.ts`: the case now tests the `skill://` URI it meant to; the unescaped backslashes tested nothing.
- README: omp listed in the same order as the prose, and sign-in is `omp setup` (omp has no `login` subcommand).

Not changed, with reasons:
- `OmpTextGeneration.ts:231` (branch policy): Claude, Codex and Antigravity also call `buildBranchNamePrompt` without `policy`; matching the repo's existing behaviour here rather than diverging in one provider.
- `ProviderRuntimeIngestion.ts:2059` (title-update ordering): no provider has ordering for title updates today, so this is a new mechanism rather than a regression in this PR — happy to add it if you want it in scope.

PR #11792, opened 2026-09-14
fix(mobile): scale inline pills with Dynamic Type

juliusmarminge:
iOS inline context pills could clip when the surrounding text used a smaller Dynamic Type size. Pills kept their original bitmap size while the paragraph scaled.

Carry the paragraph's font-size multiplier in the native chip payload so measurement, drawing, and image caching use the same scaled size. This covers skill, file, and other context pills. Android already applies this scaling.

Validation: the iOS EAS Preview [build](https://expo.dev/accounts/pingdotgg/projects/t3-code/builds/7af66ec1-acf6-4b14-b2b6-ca6105c77ed6) compiled and signed successfully. Julius tested that build on his iPhone running iOS 27 beta and confirmed the clipping was fixed. Offscreen UIKit checks passed at 12 Dynamic Type sizes; the existing iOS 26.5 and 27 simulator builds did not reproduce the original clipping.

Before, from the affected phone:

Clipped top border of an inline skill pill on iPhone

Julius explicitly waived an additional after screenshot in favor of his successful on-device verification.

Model: GPT-6. Harness: Codex.

## Summary by CodeRabbit

* **Bug Fixes**
  * Context chips in Markdown text now scale consistently with the surrounding text.
  * Chip sizing and rendering now stay aligned when a font-size multiplier is applied.
  * Improved support for larger or customized text settings on iOS.


juliusmarminge:
Reviewed the nonblocking docstring-coverage warning. The existing internal font helper is straightforward, and the attachment code now explains why measurement and drawing must share the paragraph scale. Additional API documentation would not clarify this fix. No actionable review findings remain.

PR #11853, opened 2026-09-15
feat(server): spike orchestration code mode

juliusmarminge:
Draft spike based on [#2829](https://github.com/pingdotgg/t3code/pull/2829).

This lets an agent run multi-step orchestration as a JavaScript async function body through `t3_code_mode_exec`. Existing orchestration and thread tools are exposed as `t3.<tool_name>(args)`, so a script can list threads, read them concurrently with `Promise.all`, and return a compact result without another model turn for each step.

The evaluator is one persistent Rust/native QuickJS process per environment, with up to eight reusable threads and a fresh guest heap for each script. The TypeScript server keeps tool permissions, schema validation, admission limits, cancellation, and durable call journals. Execution handles support wait/cancel and retained submission-key deduplication. Interrupted scripts are never replayed, and cancellation does not undo accepted commands or stop delegated tasks. Development and desktop/CLI/npm release packaging include the native binary.

Whether this is worth shipping is still an open question. Many harnesses already provide their own code mode, including Codex, and can overlap these same MCP calls. T3-owned code mode adds another execution layer, a native artifact, and build dependencies. Its possible value is consistent behavior across providers plus server-owned execution handles and recovery. Keeping this as a separate draft lets us evaluate that tradeoff before including it in the orchestrator work.

Validation:

- 40 focused backend tests passed, including production orchestrator dispatch, concurrent calls, cancellation, revocation, host replacement, and durable recovery without replay.
- Three Rust tests, 29 desktop tests, and 73 packaging tests passed. One Windows cross-architecture probe test also fails on unchanged HEAD and remains separate from this spike.
- Scoped lint/typechecks and native Clippy passed. The full T3 executable builds, and its actual code mode broker works with five concurrent scripts without Node on PATH.
- Manual end-to-end agent testing completed a thread list and two parallel thread reads through `t3_code_mode_exec`; its durable journal records all three calls as completed.
- On macOS arm64, the native release binary is about 1.6 MiB. Host RSS measured 6.3 MiB idle, 8.4 MiB with five scripts waiting on tool calls, and 9.3 MiB with eight. The TS host measured about 207, 293, and 332 MiB respectively on that workload. Heap-heavy scripts use more memory.

The client currently uses ordinary tool rows. A dedicated script view with nested call grouping is outside this spike, and the manual run still displayed raw code mode tool names.


DaleLJefferson:
Codex has code mode built in `codex features enable code-mode`
Not sure why you want to replicate their hard work?

PR #11578, opened 2026-09-13
fix(server): align MCP delegation support with live provider adapter registry

saphid:
## Summary

Production orchestration resolves each provider's V2 adapter dynamically through `ProviderAdapterRegistry.layerFromProviderInstanceRegistry` — any live `ProviderInstance` with an `orchestrationAdapter` is dispatchable. The MCP orchestrator server instead gated `orchestrator_capabilities` and `delegate_task`/`create_threads` target resolution on a static `isBuiltInProviderAdapterDriverV2` list that omitted Antigravity, so MCP advertised and enforced "unsupported" for instances production could actually run.

- `OrchestratorMcpService` now takes `ProviderAdapterRegistryV2` and derives per-instance orchestration capability from `list()` — the same registry dispatch consults — shared between the capability view and target resolution so they cannot diverge.
- `server.ts` wires the same `layerFromProviderInstanceRegistry` facade into `McpHttpServer.layer` that the V2 orchestrator internals use.
- All existing enabled/installed/availability/auth constraints and MCP credential scope checks are unchanged; an instance whose adapter resolves but whose provider state blocks use still reports its real constraints.
- `isBuiltInProviderAdapterDriverV2` (now unused) is removed; the built-in driver-kind set remains for adapter-driver registration tests.

## Test plan

- New `OrchestratorMcpService provider resolution` tests cover: capability advertisement for Codex/Claude/Pi/ACP-registry/Antigravity, a disabled Antigravity instance (adapter resolves, provider disabled → `canRunChildTask: false`), an unknown-driver shadow provider, `delegate_task` to Antigravity by instance id and by `driverKind`, and `provider_unavailable` for drivers without a registered adapter.
- `vp test run` on the touched MCP/orchestration test files: 19 tests pass, including the existing cross-provider `delegate_task`/`create_threads` integration replay.
- `apps/server` `tsc --noEmit`: clean on the changed files; `vp lint` on touched files: clean.
- Independent review (GPT-6 Astra, medium, read-only, frozen diff): no actionable findings.

Done by swe-2-high via Devin (Cursor harness)

Coordination trace: T3 thread 64303217-1d3f-44b1-83d1-a41a7b4e76fd; campaign spine saphid/t3code-personal#298.

PR #11870, opened 2026-09-15
fix(server): select an available provider for delegated tasks

saphid:
## Summary

A driver-only delegation target (`target.driverKind` without `providerInstanceId`) could fail on an unavailable inherited provider instance even when another instance of that driver was healthy: `resolveTarget` preferred the parent's instance whenever it was registered and orchestration-capable, and the constraint check then rejected it.

The fix requires the inherited candidate to pass the same `providerConstraints` availability check as any other candidate before it can win; otherwise the first constraint-free instance of the requested driver is selected. When no healthy candidate exists, resolution falls through to the parent instance id so the constraint check reports why the inherited instance cannot serve. Explicit `providerInstanceId` selection is unchanged and stays strict — an unavailable explicit target fails instead of silently rerouting.

Adapted from @kiiwee's earlier proposal in #11154 (closed unmerged) onto the live adapter-registry capability lookup from #11578. Credit to kiiwee for identifying the defect and the original fix shape.

## Test plan

- [x] Focused regression: `vp test run apps/server/src/mcp/OrchestratorMcpService.test.ts` — the new table-driven test proves healthy inheritance (parent model selection including options preserved verbatim), unavailable parent + healthy peer fallback, no-available-peer rejection, explicit-unavailable rejection without rerouting, and explicit healthy selection. The fallback case failed before the fix (`Provider codex cannot run a child task: Provider instance is disabled.`).
- [x] `vp test run apps/server/src/mcp/` — all 170 MCP tests pass, including the cross-provider delegation integration test.
- [x] Targeted `vp lint`, `vp fmt --check`, and `tsc --noEmit` clean for the changed scope.

Independent cross-provider review: attempted `codex exec -m gpt-6-astra -c model_reasoning_effort="xhigh" --sandbox read-only` on the frozen diff; the reviewer account was at its usage limit (reported retry Sep 19), so the review was skipped — disclosed, not substituted.

Engineering: SWE-2 Max through the Devin/T3 harness.
Coordination: T3 thread 3aa91588-f393-4df8-9d47-c8c769d27899, campaign saphid/t3code-personal#298.


saphid:
@juliusmarminge Please review and merge this OV2 fix when you are happy with it.

Choose an available provider instance for delegated tasks.

Checked 18 September at 10:58 UTC: current head `54329acd33c3493f735843eeb947e30349b6a7c7` is mergeable, all checks are green or intentionally skipped, and there is no new review feedback since the owner's completed scan. Current diff: +208/−7.

The latest-head independent agent review is still pending capacity; no fresh clean review is claimed. This request is for this contribution PR. The two scheduling PRs (#11592/#11635) are being held for a separate integration finding.

juliusmarminge:
Rebased onto `t3code/codex-turn-mapping` at `9d6feae1ff`. All seven commits applied cleanly with no conflicts.

One follow-up commit on top (`46353a3d7c`):

- **Dropped the replay-fixture `worktreePath` plumbing.** The three fixture commits (`declare replay fixture workspaces as thread worktrees`, the `exactOptionalPropertyTypes` fix, and `declare fixture worktrees only for rollback scenarios`) worked around a rollback rejection that v2 has since fixed at the source with `restoreFiles: false` in `testkit/fixtures/shared.ts`. `OrchestratorReplayFixtures.integration.test.ts` and `fixtures/shared.ts` are back to their v2 versions; the net diff is now the 13-line change in `OrchestratorMcpService.ts`, its test, and the doc paragraph.
- **Test fix.** The new delegation test mocked `ThreadManagementService.getThreadProjection`, but `OrchestratorMcpService.loadProjection` on v2 calls `getThreadRecords`, so it failed with `UnimplementedError`. Pointed the mock at the right method, matching the other cases in that file.
- The test was already one table-driven loop over six scenarios, so there was nothing to fold.

Verified in `apps/server`: `vp test run src/mcp/OrchestratorMcpService.test.ts` (9 passed), `vp test run src/orchestration-v2/testkit/OrchestratorReplayFixtures.integration.test.ts` (73 passed, no hangs without the fixture worktree branch), and `vpr typecheck` (clean).

PR #11896, opened 2026-09-15
fix(server): include Codex archived_sessions in usage scan

bulbulogludemir:
Codex moves rolled-out sessions to `archived_sessions`, but the usage scanner only reads `sessions/`, so any archived history silently contributes zero cost to the Usage page.

This change:
- resolves `archived_sessions` alongside `sessions` for every Codex home (deduped per directory, so aliased/shadow homes still scan once),
- gives Codex usage events a stable `dedupeKey` (session, timestamp, model, token totals) so a rollout present in both directories is charged exactly once; genuinely different sessions/events keep distinct keys,
- bumps `USAGE_SCAN_CACHE_VERSION` 3 → 4 because v3 Codex records carry null keys and would double-count archived copies.

Tests:
- `usageTranscripts.test.ts`: identical events parsed from two copies share one dedupeKey.
- `UsageService.test.ts`: an identical archived rollout copy is scanned (sources 4 → 5, files 4 → 5) with totals unchanged.
- `usageScanCache.test.ts`: previous-version rejection now targets v3.
- `apps/server` typecheck passes; usage test files pass (`usageTranscripts/usageScanCache/usageAggregation` green, modified `UsageService` account test green — 2 pre-existing `it.live` timeouts in that file also fail on clean main in this environment).

## Summary by CodeRabbit

- **Bug Fixes**
  - Usage reporting now includes archived Codex sessions, improving completeness when sessions are moved or copied.
  - Archived and active copies of the same session are counted only once, preventing duplicate usage charges.
  - Usage scan data is refreshed safely to account for updated duplicate-detection behavior.
- **Tests**
  - Added coverage confirming consistent identification of copied Codex rollout records and correct usage totals.


bulbulogludemir:
Ready for human review (per the Macroscope approability note — this touches usage metering, so auto-approval is off by policy).

Verification summary:
- New behavior covered by tests: identical Codex rollout copies across `sessions/` and `archived_sessions/` are scanned but charged once; archive-only sessions now contribute usage; moving a rollout into archives keeps totals stable.
- `apps/server` typecheck passes; new/updated tests in `usageTranscripts`, `usageScanCache` and the modified `UsageService` account test pass. (2 pre-existing `it.live` timeouts in `UsageService.test.ts` also fail on clean `main` in this environment, unrelated to this change.)
- Also validated the parser/aggregator against ~2.5k real Codex transcript files (sessions + archives): 0 read errors.

PR #11951, opened 2026-09-15
fix(circe): stateless acoustic repair and frame I/O for live voice

Absterrg0:
Transport repair: a live delegation that receives no speech within the hold window now speaks one acoustic repair prompt instead of dropping silently. No dispatch, no frame, no retry loop.

Frame I/O: session instructions now order the live model to relay backend questions verbatim and delegate replies immediately with the original request. Client and server frame-pin behavior locked with regression tests proving answers carry the exact frame id with no second supervisor call.

Tests: CirceLiveVoice.logic, server CirceLiveVoice, CirceVoiceRuntime, liveVoice, bridge suites green. Typecheck and lint clean for server and web.


Absterrg0:
Opened against the wrong repo by automation; this change belongs to the Circe fork, not upstream. Closing.

PR #11958, opened 2026-09-15
fix(desktop): paste as text no longer doubles the pasted text

### LINKED ISSUE #11955: [Bug]: Ctrl/Cmd+Shift+V pastes the text twice in the terminal and the composer
### What happened

Pressing <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>V</kbd> (paste as text) in the desktop app inserts the clipboard contents **twice**. Reproduced in both the integrated terminal and the chat composer.

### Steps to reproduce

1. Copy any text to the clipboard.
2. Focus the integrated terminal (or the chat composer) in the desktop app.
3. Press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>V</kbd>.
4. The text appears twice.

Plain <kbd>Ctrl</kbd>+<kbd>V</kbd> is unaffected. Clicking **Edit → Paste as Text** from the menu bar is also unaffected.

### Cause

The Edit menu binds `CmdOrCtrl+Shift+V` to the "Paste as Text" item added in #11442. Chromium already performs paste-as-plain-text for that chord natively in editable contexts, so adding an accelerator-bound menu item does not *replace* that handling — it *adds* to it. The menu click handler additionally round-trips through the renderer to `webContents.paste()`, which lands a second paste event on top of Chromium's.

Both surfaces already arm their "skip attachment folding" flag from the same keydown, so the injected paste was never needed for the accelerator path — only for an actual menu click, which produces no keystroke for the renderer to observe.

### Fix

No-op the menu click handler when `event.triggeredByAccelerator` is true, leaving the chord entirely to Chromium. Menu clicks keep the renderer round trip.

```ts
const pasteAsTextClick = (
  _item: Electron.MenuItem,
  _window: Electron.BaseWindow | undefined,
  event: Electron.KeyboardEvent,
) => {
  if (event.triggeredByAccelerator === true) return;
  runMenuEffect("paste-as-text", dispatchMenuAction("paste-as-text"));
};
```

Relevant files: `apps/desktop/src/window/DesktopApplicationMenu.ts`, `apps/desktop/src/ipc/methods/window.ts`, `apps/web/src/lib/desktopPasteAsText.ts`.

I have this fixed locally with a regression test in `DesktopApplicationMenu.test.ts` and can open a PR if useful.

### Environment

Linux desktop app, Electron 44.

TonybynMp4:
Closes #11955.

Pressing <kbd>Ctrl</kbd>/<kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>V</kbd> inserted the clipboard contents twice, in both the integrated terminal and the chat composer.

Chromium already performs paste-as-plain-text for that chord in editable contexts. The **Edit → Paste as Text** item added in #11442 binds the same accelerator, and binding it does not replace Chromium's handling — it stacks on top. The click handler then round-tripped through the renderer to `webContents.paste()`, landing a second paste on Chromium's.

The handler now no-ops when `event.triggeredByAccelerator` is true, leaving the chord entirely to Chromium. Both surfaces already arm their skip-attachment-folding flag from the same keydown (`ChatComposer.tsx` for the composer, `surface.ts` for the Linux terminal shortcut), so the injected paste was only ever needed for a real menu click, which produces no keystroke for them to observe. Menu clicks keep the renderer round trip unchanged.

The `_window` parameter is typed `Electron.BaseWindow` rather than `BrowserWindow` to match Electron 44's `MenuItemConstructorOptions["click"]` signature.

## Verification

- New regression case in `DesktopApplicationMenu.test.ts` asserts the accelerator path dispatches nothing. Negative control: it fails on unmodified source, passes with the fix.
- `vp test run src/window/ src/ipc/` in `apps/desktop` — 117 passing.
- `vp test run src/lib/desktopPasteAsText.test.ts src/terminal/ghostty/surface.test.ts` in `apps/web` — 56 passing. The renderer is untouched; these confirm no fallout.
- Typecheck and lint clean on the changed files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed duplicate text insertion when using the “Paste as Text” keyboard shortcut.
  - Directly selecting “Paste as Text” from the application menu continues to work as expected.
  - Improved consistency between keyboard shortcuts and menu-based paste actions.


TonybynMp4:
Followed up on the paste-to-focus check. **Result: the fix takes that path from two insertions to one, not to zero.**

I could not drive the real app end to end here (no C toolchain to build `node-pty`, and no key-injection tooling), so I isolated the question — which is really about Chromium, not T3 Code — in a minimal Electron 44 app mirroring the same structure: an Edit menu with `CmdOrCtrl+Shift+V`, a non-editable `tabindex` div standing in for the timeline, a textarea for the composer, and a window-level **capture** `paste` listener like the one in `ChatView.tsx`.

One harness caveat worth stating: Playwright's synthetic keypress does not fire the native Electron accelerator (the menu handler never ran, and the doubling did not reproduce). So the run below issues a real Chromium `Control+Shift+V` and then invokes the menu handler explicitly with the event object Electron would pass. The keypress itself is a genuine Chromium paste — it inserts into the textarea — so only the accelerator-to-menu hop is emulated.

| build | focus | paste events at window listener | textarea |
|---|---|---|---|
| current `main` | editable | **2** | `CLIPCLIP` |
| current `main` | non-editable (timeline) | **2** | — |
| with fix | editable | **1** | `CLIP` |
| with fix | non-editable (timeline) | **1** | — |

Two things this settles:

1. **Chromium's native `Ctrl+Shift+V` does fire a `paste` ClipboardEvent at a window-level capture listener even when focus is non-editable.** That is exactly the precondition `pasteTextToFocusComposer` needs, so paste-to-focus keeps its event after the injected paste goes away.
2. On current `main` the non-editable path was *also* doubling — the capture listener received the clipboard twice there too, so this fixes paste-to-focus rather than merely sparing it.

I also ran the fixed build with `triggeredByAccelerator: false` and confirmed `INJECTING_PASTE` still fires, so a real **Edit → Paste as Text** click keeps its renderer round trip.

Still worth one human press of the real chord on a real build before merge, since the accelerator hop is the one link I emulated — but the mechanism it was standing in for is the one already demonstrated in #11955.

juliusmarminge:
CI failed

PR #11608, opened 2026-09-13
fix(client-runtime): map cancelled and interrupted subagent terminal statuses

Exotic209093:
Codex subagents that complete with 'cancelled' or 'interrupted' status were falling through to 'running' because TASK_COMPLETED_STATUS only covered completed, failed, and stopped. This adds both missing entries so the fold correctly marks them as terminal.

Fixes #11164

## Summary by CodeRabbit

* **Bug Fixes**
  * Subagent tasks that are cancelled or interrupted now display the correct terminal status instead of being marked as completed.
  * Task summaries are preserved as results, and completion times are recorded for cancelled and interrupted tasks.

Issue #11164, opened 2026-09-11
[Bug]: Codex subagent terminal statuses fall through to `running` in orchestrator v2


eduardozf:
### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I included enough detail to reproduce or investigate the problem.

### Area

apps/server

### Steps to reproduce

1. Check out PR #2829.
2. Start a Codex thread using orchestrator v2.
3. Spawn a subagent.
4. Interrupt or shut down the subagent, or process a collaboration state where Codex reports `notFound`.
5. Inspect the subagent status recorded by T3.

This can also be reproduced directly by passing each Codex collaboration status through `updateSubagentStates`.

### Expected behavior

T3 should translate each Codex status into the corresponding internal status:

| Codex status | Expected T3 status |
|---|---|
| `pendingInit` | `pending` |
| `running` | `running` |
| `interrupted` | `interrupted` |
| `completed` | `completed` |
| `errored` | `failed` |
| `shutdown` | `cancelled` |
| `notFound` | `failed` |

These are the statuses in Codex’s [current `AgentStatus` definition](https://github.com/openai/codex/blob/28f43b0417ab19632f7e53d17806b98151d6b9a0/codex-rs/protocol/src/protocol.rs#L1824-L1847).

### Actual behavior

While reviewing issue #8499, I found that PR #2829 already fixes the reported registration problem. The completed `spawnAgent` event exposes the child thread ID through `receiverThreadIds`, the PR registers it, and the adapter also handles the newer `subAgentActivity` path while buffering child turns that arrive before registration.

I verified those paths against Codex CLI `rust-v0.150.1`, the version reported in #8499, and the latest `openai/codex` main branch.

During that review, I found a separate bug in [`updateSubagentStates`](https://github.com/pingdotgg/t3code/blob/89459e384ad7ffa3b9dda613e9cfebf32a8979e1/apps/server/src/orchestration-v2/Adapters/CodexAdapterV2.ts#L2561-L2584). It checks for `failed`, `cancelled`, and `closed`, but Codex never sends those values. Several real Codex statuses therefore fall through to `running`:

| Codex status | Current T3 status |
|---|---|
| `pendingInit` | `running` |
| `running` | `running` |
| `interrupted` | `running` |
| `completed` | `completed` |
| `errored` | `failed` |
| `shutdown` | `running` |
| `notFound` | `running` |

The effects are:

- An interrupted child continues to appear active.
[...]

PR #11996, opened 2026-09-16
fix(server): preserve native Codex subagent lifecycle in v2

lnieuwenhuis:
Codex collaboration events use native statuses that the V2 adapter was treating as `running`, leaving interrupted, shut-down, and missing children active and initializing children incorrectly marked as executing.

Map all seven native statuses explicitly. Keep initializing children protected from idle release and promote them on their first native turn, including when the child start arrives before its spawn snapshot. Settle both registration execution nodes for children that terminate before their first native turn; native-turn events retain ownership once execution starts.

Targets `t3code/codex-turn-mapping` for #2829. Refs #11164. #11608 addresses the separate client-runtime path on `main`; this fixes the V2 provider adapter.

Validation: 91 adapter replay tests pass, including regressions observed failing before the fix; scoped server typecheck and targeted lint/format pass. Independent adversarial reviews found lifecycle gaps that were fixed with regression coverage. A subsequent Macroscope finding about children terminating before their first native turn was reproduced and fixed; a fresh adversarial review of that fix reported no findings. Changed files pass the secrets scan. No live-provider verification.

The repository scanner config retains default rules and exempts only the exact synthetic `idempotencyKey` fixture in `packages/effect-codex-app-server/src/protocol.test.ts`, which otherwise flags twice in the inherited V2 history. Negative controls confirm other values, keys, paths, and credential rules remain detected.

Implemented with GPT-6 Astra via Codex.


juliusmarminge:
Rebased this PR onto the current `t3code/codex-turn-mapping` at `4a4c22b29c`.

**What changed on the branch**

- The two "reconcile" merge commits (`e194c16dca`, `6603dbc876`) were dropped. They carried a stale copy of the v2 branch and would have reverted recent v2 work (plus a `.gitleaks.toml` change that does not belong here). Only your three content commits remain, cherry-picked in order: map native statuses, retain initializing subagents, settle nodes before native turns. All three applied cleanly against the current `CodexAdapterV2.ts`; no manual conflict resolution was needed.
- No code edits of my own. I checked that `"pending"` is still a valid `OrchestrationV2Subagent["status"]` in `packages/contracts/src/orchestrationV2.ts`, and that v2's "Map nested Codex subagent threads correctly" change does not overlap with your `agentsStates` mapping (it only changed node parent ids), so each native status is mapped exactly once via `codexSubagentStatus`.

**Verified**

- `vp test run src/orchestration-v2/Adapters/CodexAdapterV2.test.ts` in `apps/server`: 108 passed.
- `vpr typecheck` in `apps/server`: clean (only pre-existing Effect suggestions).

**Left for you / a maintainer**

- If you need to pick up future v2 changes, please rebase rather than merging the base branch in; the merge-based reconcile is what caused the revert risk.

apps/server/src/orchestration-v2/Adapters/CodexAdapterV2.test.ts
```diff
@@ -5440,6 +5440,239 @@ describe("CodexAdapterV2 post-settle continuation", () => {
...
+    ["pendingInit", "pending", false],
+    ["running", "running", false],
+    ["interrupted", "interrupted", true],
+    ["completed", "completed", true],
+    ["errored", "failed", true],
+    ["shutdown", "cancelled", true],
+    ["notFound", "failed", true],
+  ] as const) {
+    for (const nativeTurnStarted of [false, true]) {
+      it.effect(
```
juliusmarminge, line 5453: `it.effect.each`

apps/server/src/orchestration-v2/Adapters/CodexAdapterV2.test.ts
```diff
@@ -5440,6 +5440,239 @@ describe("CodexAdapterV2 post-settle continuation", () => {
...
+                assert.deepEqual(event.turnItem.completedAt, terminal ? now : null);
+              }
+            }).pipe(Effect.provide(Layer.merge(idAllocatorLayer, NodeServices.layer))),
+          ),
+      );
+    }
+  }
+
+  for (const childStartsFirst of [false, true]) {
+    it.effect(
```
juliusmarminge, line 5579: it.effect.each

juliusmarminge (changes requested):

juliusmarminge:
Superseded by merged PR #12974, which includes the Codex V2 native subagent-status lifecycle fixes and broader settlement/reopen handling. Closing this older implementation.

PR #12043, opened 2026-09-16
feat(web): add find-in-conversation (Cmd+F)

IVainqueur:
## Summary
- Adds a browser-style find bar to the conversation pane, triggered by Cmd+F (new `conversation.find.toggle` command).
- Matches are found over the virtualized timeline's row data (not the DOM), so it works across the whole thread regardless of what's currently mounted.
- Highlighting reuses the DOM-text-walk + CSS Custom Highlight API machinery already built for assistant citations, extended to support multiple simultaneous matches plus a distinct active match, and correctly spans matches across markdown-rendered inline elements (bold, links, code).
- Enter/Shift+Enter and prev/next buttons step through matches with a live `n/total` counter; Escape closes the bar and clears highlights from anywhere within it.

## Test plan
- [x] Unit tests for match indexing (`conversationFindMatches.test.ts`) and DOM range/highlight building (`conversationFindHighlight.test.ts`)
- [x] `tsc --noEmit`, lint, and format clean on all touched files
- [x] Manually verified in a real browser against a seeded conversation: open/close/toggle via Cmd+F, live query highlighting across bold/inline-code spans, next/prev navigation via keyboard and buttons, Escape closing from any focus state within the bar
- [x] Verified no Cmd+F conflict in the Electron app menu (no `find` role/accelerator defined)


IVainqueur:
Opened by mistake against the wrong repo (meant to target my fork). Closing and reopening against IVainqueur/t3code.

PR #12073, opened 2026-09-16
feat(provider): add DeepSeek Harness over ACP

ac0rdj:
Adds DeepSeek Harness as a first-party provider.

## Why this provider

DeepSeek Harness already ships an ACP profile (`dsh --profile acp`), and T3 Code already has a
shared ACP runtime, so this is one driver on an existing seam rather than a second protocol.

DeepSeek's own framing is "**everything is a plugin**": models, tools, skills, sessions, sandboxes,
storage, loops, scheduling and the UI are all Cordis plugins that a deployment recomposes in
configuration, without forking the harness. It also records "everything the model sees" in an
append-only session log, and resume, fork and replay all run on that one event stream. That is an
unusually good fit for a control surface, because the harness keeps the parts a client should not own,
and exposes the parts a client can drive.

The harness is unusually well suited to a control surface like T3 Code:

- **The request prefix is stable, so the cache is too.** DSH's own docs state the ACP persona is
  stable "for a fixed profile, provider, model, and tool roster", and the only per-session values in
  it are the model id and the working directory. Nothing in this binding varies the prefix between
  turns. Measured on a live session: with 11,127 tokens already in context, a follow-up turn moved
  the total to 11,172 — **45 tokens for a whole turn**, i.e. the conversation was reused rather than
  re-sent. That is the property that makes long, tool-heavy threads cheap, and it is the main reason
  I built this.
- **It is configurable well past this binding.** `dsh-llm-pi-ai` is mounted dormant and "activates
  when user settings add them", routing to pi-ai catalogs or any OpenAI-compatible gateway with no
  code change. Today T3 reads the advertised ACP catalog, so those routes are not yet surfaced —
  that is the natural follow-up, and it is the difference between one more provider and a provider
  that can front several.

## This is the ACP path, not another dedicated provider integration

Two provider PRs were closed on this repo on the same day with the same reasoning:

> we are closing this because we are not adding another dedicated provider integration ahead of
> shared ACP support already integrated into Orchestrator V2

and, for the one that reused ACP transport:

> although this reuses ACP transport, it introduces a separate adapter lifecycle for session recovery
> and approvals [...] we want to assess Auggie against the shared registry approach

That is the bar, so it is worth being precise about how this PR clears it.

- **The shared ACP runtime is the implementation.** `DshAdapter` drives
  [`AcpSessionRuntime`](../../apps/server/src/provider/acp/AcpSessionRuntime.ts) — the runtime already
  on `main` — for initialization, session create/resume, prompt dispatch, cancellation, configuration
  writes, event streaming, and teardown. There is no second protocol client.
- **No private protocol surface.** It registers **zero** extension handlers, permission handlers, or
  elicitation handlers. `dsh --profile acp` is pure ACP v1: no `_meta`, no private method, no extra
  transport field. Antigravity, Cursor and Grok each carry a private extension; this needs none.
- **No parallel lifecycle.** Turn settlement, cancellation and the liveness watchdog are the same shape
  the other ACP drivers use. Approvals are not reimplemented — the harness never asks, so the adapter
  registers nothing and ACP's unhandled-request path fails closed.
- **Session recovery is the shared one.** The resume cursor is the ACP session id, handled by the
  shared runtime's resume path.

The provider-specific code is two files: the ACP binding (spawn + the `[provider, model]` option
translation) and a driver that wires the shared pieces. That is closer to the "shared registry
approach" than a dedicated adapter managing its own approvals and recovery.

## What

`dsh --profile acp` streams standard ACP v1 `session/update` notifications, advertises no
authentication methods, and selects its model through the standard `model` configuration option. A
model value on the wire is a JSON `[provider, model]` pair, so a selection resolves against the
advertised options and T3 stores the model id itself: the picker shows **one row per model**, not
one per encoding. `reasoning_effort` is a standard select (`off`, `low`, `high`, `max`) and rides on
the model's capabilities, which is where the picker reads it.

**One limitation worth reviewing.** The ACP binding never asks the client to approve a tool call, so
T3's permission modes cannot apply to this provider. I deliberately did not register a
`session/request_permission` handler: an unhandled ACP request answers `method not found`, which
fails closed if a future harness build starts asking. The provider card carries a **No approvals**
badge and hides the interaction-mode toggle instead of offering modes that cannot be enforced. The
user guide says so plainly, and I verified the behaviour rather than assuming it: a live turn wrote
and read a file with zero approval requests.

Sessions resume: the harness session id is the resume cursor, so a restarted server reconnects to the
same conversation. Verified by resuming in a fresh process and having the agent recall a codeword
from before the restart.

### Out of scope, deliberately

- **Agent presets.** `dsh-agent-presets` swaps the tool/prompt composition per session, but the ACP
  profile does not mount it and ACP v1 has no surface for it. ACP ships one fixed composition by
  design, so presets are a deployment concern, not a client one.
- **A custom-model surface.** DSH rejects a model its ACP option does not advertise (`unknown model
  option`), so a user-supplied entry could never be selected. Rather than ship a field that silently
  does nothing, there is no custom-model field at all.
- **The plugin catalog and the trajectory view.** Those are DSH's own surfaces today. ACP v1 has no
  method for them, so a client cannot list a profile's plugins or read its session log. If DSH grows
  that surface, or a deployment exposes it, T3 can read it without a new adapter.

## Screenshots

Provider card, disabled and enabled. The badge is the no-approval warning:

dsh-provider-disabled

dsh-provider-ready

One row per model in the picker, and the selection in the composer:

dsh-model-picker

dsh-selected-in-composer

A real turn driven from the web client, against a real key. The composer shows the selected
DeepSeek model, the turn completed, and the agent created the file it was asked for:

dsh-turn-result

The card in the provider list beside the others:

dsh-card-in-list

Screenshots are hosted on a throwaway `pr-assets/dsh-provider` branch and are not part of this diff.
The turn screenshot shows the selected model and the finished work; the provider-level proof that the
turn ran on DSH is the environment's own request log, quoted under Verification below.

## Verification

- `vp run -r typecheck` clean, `vp lint` clean on every changed file.
- `vp test run` on the touched suites: 168 passed, including the new `DshAcpSupport` and
  `DshProvider` suites and the all-drivers registry slice, which now boots six drivers.
- End to end against a real `dsh` 0.1.5-rc.1 and a real DeepSeek key: provider check `ready` with the
  live catalog, a real tool-using turn, and a resume across processes. I ran that as a throwaway
  local test and deleted it, since it needs a binary and credentials CI does not have.
- The turn was also driven from the web client, and the environment's provider request log records it
  as DSH end to end — the ACP request, then the canonical events, then settlement:

  ```
  NTIVE: {"kind":"request","provider":"dsh","...}
  CANON: {"type":"item.completed","provider":"dsh","threadId":"5307d528-...","...}
  CANON: {"type":"turn.completed","provider":"dsh","threadId":"5307d528-...","...}
  ```

## Review findings

An earlier revision of this PR drew review from Macroscope and CodeRabbit. I verified each finding
against the source and fixed the real ones in a follow-up commit:

- a turn that produced no ACP update stayed `running` forever, because the liveness deadline was set
  from the first update instead of turn start;
- `sendTurn` claimed the turn before validating and leaked the claim on failure, which left a thread
  active with no terminal event and made later turns look like steers;
- a terminated ACP connection left a dead session registered;
- the catalog listed every model twice, which is the picker defect the screenshots above show fixed;
- the custom-model and `unauthenticated` branches could not be reached and are gone;
- reasoning effort is re-validated after a model write, and the model editor's DSH choices now match
  what DSH advertises.

---

I read `CONTRIBUTING.md` before opening this, so I know it is a large diff and that the process asks
for a discussion first. I am opening it as a PR anyway because the work is done and verified, and a
running branch is easier to judge than a proposal. Happy to close it, split it, or move this to an
Ideas discussion instead.

Model and harness: `deepseek-flash` in DeepSeek Harness.

## Summary by CodeRabbit

* **New Features**
  * Added DeepSeek Harness as a supported provider for chat and text generation.
  * Added automatic discovery of available models and reasoning levels.
  * Added session persistence across server restarts.
  * Added DeepSeek branding and provider settings, including executable and home-directory configuration.
  * Added status checks, authentication detection, and version update guidance.
  * Added a no-approval execution model with a corresponding provider badge.

* **Documentation**
  * Added installation, authentication, configuration, model selection, reasoning-level, session persistence, and no-approval execution guidance.


juliusmarminge:

PR #12087, opened 2026-09-16
feat(web): edit saved review comments in place

cmaciasjimenez:
Once a review comment is submitted in the diff panel, its text is frozen: hovering it offers only a delete button, so fixing a typo means deleting the comment and writing it again from scratch. Comments already posted to a pull request host have been editable all along (`pullRequests.updateComment`), which makes the local ones the odd surface out.

Give a saved comment a pencil beside its delete action. It reopens the same inline composer seeded with the comment's text, and buffers the rewrite locally so cancelling leaves the saved comment untouched. Saving upserts the record by id, so the draft keeps one inline reference and the comment keeps its place among the draft's context.

`addReviewComment` already replaced a record with a matching id, but by filtering and re-appending, which moved an edited comment to the end. It now replaces in place, so an edit no longer reorders the draft's context.

No contract, server, or mobile change: review comments are composer-draft state, and mobile dissolves them straight into the draft text rather than keeping an editable bubble. Pending (unsubmitted) pull-request review comments are still delete-only — a separate surface, left for a follow-up.

### Before

Hovering a saved comment offers only delete.

Before: hovering a saved review comment in the diff panel reveals only a delete button

### After

The same comment now offers edit beside delete, and the pencil reopens the composer with the comment's text, Cancel and Save.

After: hovering the same comment reveals an edit button beside delete

After: the inline composer reopened on the saved comment, prefilled with its text, with Cancel and Save

https://github.com/user-attachments/assets/53d618dc-0575-4332-a7cd-aa9f1966450e

### Verification

- `vp test run apps/web/src/composerDraftStore.test.ts apps/web/src/components/diffs/DiffCommentAnnotation.test.tsx` — 143 passed, including a new case covering in-place replacement.
- `vp test run apps/web/src/components/files/FilePreviewPanel.test.ts apps/web/src/components/diffs/StyledDiffCodeView.test.tsx apps/web/src/components/diffs/commentSubmitShortcut.test.ts` — 17 passed.
- Web typecheck and targeted lint clean on the changed files.
- Walked the flow in the running web app on both the thread diff panel and the file preview: add, hover, edit, cancel, save, delete.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to edit saved review comments directly within code annotations.
  - Added Save and Cancel controls, keyboard escape handling, and an edit action for existing comments.

- **Bug Fixes**
  - Updated edited comments without changing their original position in the review comment list.
  - Preserved comment ordering when replacing existing review comments.


cmaciasjimenez:
Leaving the CodeRabbit **Docstring Coverage** pre-merge warning (33.33% vs its 80% default) unaddressed on purpose.

Satisfying it would mean adding a docstring to `saveEdit` in `apps/web/src/components/diffs/AnnotatableCodeView.tsx`. None of its four sibling callbacks in that file carries one:

```
removeEntry    = useCallback(...)
submitEntry    = useCallback(...)
saveEdit       = useCallback(...)   <- added here
beginComment   = useCallback(...)
```

Documenting only the new one would make it the single commented callback in the file. The convention here is that the exported surface gets a doc comment and internal callbacks do not — `DiffCommentAnnotation`'s `/** The shared inline comment treatment for file previews, thread diffs, and pull-request diffs. */` is untouched, and this change adds comments only where the reasoning is not visible from the code: why a rewrite is buffered before saving, and why `addReviewComment` now replaces in place.

The 80% threshold is CodeRabbit's generic default rather than anything this repository asks for, so I have treated it as a false positive rather than changing the code to match it. Happy to reverse that if a maintainer disagrees.

The CodeRabbit check itself reports **Review completed** with no actionable comments, and the review body puts merge risk at Minimal.

PR #12122, opened 2026-09-16
feat(server): run a teardown script when a thread settles

dtdetwiller:
## Problem

Settled threads keep their worktrees and dependencies forever. On a project with a heavy install that is several gigabytes per thread, and it piles up until the user deletes worktrees by hand. On one machine the athlix worktree directory grew from 31 GB to 161 GB in two days.

## Fix

A project action can be marked `runOnThreadSettle`, in `t3.json` or the action editor, mirroring `runOnWorktreeCreate`. A new `WorktreeTeardownReactor` listens for the settled and un-settled events:

- **On settle** it runs the teardown action headlessly in the thread's worktree, with the same `T3CODE_PROJECT_ROOT` and `T3CODE_WORKTREE_PATH` environment actions already get. Because automatic settlement dispatches `thread.auto-settle` and emits the same `thread.settled` event, auto-settled threads are covered too.
- **On un-settle** it runs the setup action again through the existing `ProjectSetupScriptRunner`, so a settled thread never comes back to a broken checkout.
- It skips teardown when another active thread shares the worktree, when the thread runs in the project checkout instead of a worktree, or when the worktree is already gone. Failures are logged, never surfaced as thread errors.

Actions resolve through `resolveProjectScripts` so environment defaults and per-project overrides apply the same way they do for setup.

Supporting changes:

- `ProcessRunInput` gains a `shell` option so the action line runs as written (`/bin/sh -c` on POSIX, `cmd.exe /c` on Windows) instead of being resolved as an executable.
- Web: the action editor has a "Run automatically when the thread settles" switch, `t3.json` import carries the flag, and lists tag the action "teardown". Setup and teardown are exclusive roles, enforced by one `withExclusiveScriptRoles` helper instead of the same map duplicated in the chat view and settings hook.
- Mobile: the actions menu labels the action "(teardown)".
- Docs: new `docs/user/project-scripts.md` covering setup and teardown actions, linked from the docs index and the threads page.

## Surfaces

- Entry points: action editor, `t3.json` import, settings actions list, chat actions menu.
- Clients: web and desktop share the editor; mobile only labels the action (it has no editor).
- Providers: not provider-shaped.
- Contracts: `ProjectScript.runOnThreadSettle` and `T3ProjectFileScript.runOnThreadSettle`, both optional so old data still decodes.
- Reverse state: un-settle reruns setup.

## Verification

- New `WorktreeTeardownReactor.test.ts` covers settle, un-settle, no teardown action, shared worktree, and checkout-or-missing worktree. New process runner test covers shell mode. Existing action helper and schema tests extended.
- Targeted `vp test run`, typecheck, lint, and `vp fmt --check` on every changed package.

Built with Claude Fable 5.1 in T3 Code.

## Summary by CodeRabbit

- **New Features**
  - Added teardown project actions that run automatically when a thread settles in its worktree.
  - Setup actions run again when a settled thread is reopened.
  - Added a project-script editor option for configuring teardown actions.
  - Script menus and settings now identify Setup and Teardown roles clearly.
  - Teardown actions are skipped when unsafe or unnecessary, such as shared or missing worktrees.

- **Documentation**
  - Added documentation covering project actions, lifecycle behavior, configuration, and available environment variables.


dtdetwiller:
Closing: opened by mistake from a downstream checkout. Keeping the change on the fork.

PR #12184, opened 2026-09-17
fix(mobile): allow changing provider in a started thread

juliusmarminge:
## Problem

On mobile, the thread model picker built its catalog from the thread's own provider instance, so an existing thread could only change models *within* the provider it started on. Web has allowed leaving that provider since v2: it unlocks the picker whenever the thread's session reports `supportsProviderSwitchingViaHandoff`, which every v2 adapter does.

## Fix

Derive the same capability on mobile from the thread projection — via `resolveThreadProviderSession`, the helper web reads — and scope the catalog to the thread's instance only when the session cannot hand the conversation off. A thread that never ran a turn is never bound, matching web's `deriveLockedProvider`.

- `apps/mobile/src/features/threads/thread-provider-switching.ts` holds the rule, with tests.
- `use-thread-composer-state` derives it where the projection is already subscribed, so the composer takes a boolean instead of its own projection subscription.

## Evidence

Android emulator (Pixel 9, API 36) on a dev client built from this branch, against an isolated server with Codex and Claude configured. Same thread in both shots: started on Codex (one completed turn), same viewport, same state.

| Before (base) | After (this PR) |
| --- | --- |
| Thread settings picker listing only the Codex section for a started Codex thread | Same picker listing both Codex and Claude sections |

Switching is not just visible, it runs: picking Claude Opus 5 in that Codex thread and sending the next turn hands the conversation over and Claude answers it.

Thread that started on Codex answering its next turn as Claude Opus 5

Tests: `vp test run apps/mobile/src/features/threads apps/mobile/src/lib/modelOptions.test.ts apps/mobile/src/state` — 56 files, 534 tests, all passing. `tsc --noEmit` clean in `apps/mobile`.

Model: Claude Opus 5 (1M context) · Harness: Claude Code in T3 Code


juliusmarminge:
**`Test Server 3` failure is pre-existing on the base branch, not from this PR.**

The four failures are all `checkpoint.rollback` dispatches:

- `OrchestratorReplayFixtures.integration.test.ts` — `thread_rollback/codex`, `thread_rollback/claudeAgent`
- `ThreadFork.integration.test.ts` — Claude native fork rollback (source + fork-local)

They fail at `apps/server/src/orchestration-v2/Orchestrator.ts:6868`, the `if (!isolated) return new OrchestratorDispatchError(...)` guard added by 1ee1d04642 ("fix(server): reject v2 file restore in shared workspaces"), the commit this branch sits directly on top of. Those fixtures roll back in a shared workspace, so the new guard rejects them; the commit updated `CheckpointRollbackService.test.ts` but not these integration fixtures.

Evidence it is not this PR:

- `git diff --name-only t3code/codex-turn-mapping...HEAD` is `apps/mobile/**` only — nothing outside the mobile app changed, so an `apps/server` test cannot observe this diff.
- Reproduced locally at this head with the identical error; the run's inputs outside `apps/mobile` are byte-identical to the base.
- The last green CI run on `t3code/codex-turn-mapping` (f08294a1) predates 1ee1d04642, so the base head has not been through CI with that guard.

Not fixing it here — separate concern, separate PR.

juliusmarminge:
Cross-check: #12200, a one-line `app.config.ts` version bump with no other changes, fails the same four `checkpoint.rollback` tests on `Test Server 3`. That independently confirms the failure is the base branch's, not this PR's.

PR #12223, opened 2026-09-17
fix(source-control): stop Forgejo status refresh from scanning every pull request

loispostula:
## What Changed

Pass `head=<branch>` to Forgejo's pulls endpoint instead of paging the whole repository and matching the branch client-side. The client-side match stays, since a server too old for the filter ignores it and answers unfiltered.

## Why

`vcs.refreshStatus` hangs past 15s on any Forgejo or Gitea repository with real pull request history.

`findLatestPrForHeadContext` asks for the branch's pull requests with `limit: 20`. GitHub and GitLab narrow server-side; Forgejo did not, so it looped `while (results.length < limit)` over `pulls?state=all&limit=50&page=N` filtering in JS. A branch has one pull request at most, so the loop only stopped on an empty page, walking every pull request in the repository. That sits inside `refreshStatus`'s remote write lock.

`head` is documented on `GET /repos/{owner}/{repo}/pulls`. With it, a branch with one pull request costs two requests, a branch with none costs one.

Introduced in #11436.

## Checklist

Written with Claude Opus 5 in T3 Code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Forgejo pull-request listings now correctly filter results by the requested source branch.
  * Pagination stops once the server returns no additional results, avoiding unnecessary requests.


kossoy:
Confirming this on self-hosted Forgejo `16.0.5+gitea-1.22.0` with nightly `0.0.43-nightly.20260923.2150`.

The repo has ~200 PRs. Two of them are huge merged PRs: a package store committed by mistake, then removed (~5.7M lines / ~31k files each). Forgejo computes `additions`/`deletions`/`changed_files` for every item in the list response without caching, so each of those two adds ~6.5 s to any listing page that contains it. The scan always walks to the empty page, so every PR lookup pays for both, even when the branch's PR is on page 1.

From the server trace:

- `vcs.refreshStatus`: 22.75 s, of which `findLatestPrForHeadContext` is 21.6 s (4 pages; page 1 alone 15.6 s)
- the same query with `head=<branch>`: 0.09 s, returns the same PR
- `head=owner:branch` returns an empty list on this version. The bare branch name works, which is what this PR sends.

For us this is what raises the "Some requests are slow" toast when the window regains focus.

PR #12241, opened 2026-09-17
fix(devices): keep agent-device from spawning a second T3 app

adinvadim:
## What Changed

Two guards, same failure:

1. Generated `agent-device` shims always set `ELECTRON_RUN_AS_NODE=1`, so desktop `resolveNodeExecutable` (the Electron binary) runs as Node instead of as a second app. Provider device env also passes that flag.
2. Packaged desktop no longer scans past port 3773. If 3773 is taken, startup fails instead of binding 3774 against the same `~/.t3`.

## Why

OpenCode/Cursor threads died with "Provider session did not survive a server restart" whenever an agent ran `agent-device`. The shim exec'd `T3 Code.app` without `ELECTRON_RUN_AS_NODE`, Electron started a second GUI, sequential port scan picked 3774, that backend reconciled the shared sqlite and settled every live provider session. It also overwrote `server-runtime.json`, so discovery kept pointing at the dead nested instance.

```text
agent-device (PATH shim)
  exec T3 Code.app launcher.mjs     # no ELECTRON_RUN_AS_NODE
    DesktopApp starts
      scan 3773 busy -> 3774
        reconcileProviderSessions
          kill OpenCode / Cursor
```

After:

```text
agent-device
  export ELECTRON_RUN_AS_NODE=1
  exec T3 Code.app as Node -> agent-device.mjs

if a second GUI still appears:
  3773 busy -> refuse, do not start 3774
```

## UI Changes

N/A. Session-kill is backend lifecycle.

## Verification

- `vp test run apps/server/src/device/AgentDeviceTarget.test.ts apps/desktop/src/app/DesktopAppErrors.test.ts` → 5 passed
- Targeted lint on the touched files: clean
- Live: restored `server-runtime.json` to the real 3773 backend (pid 12624); nested 3774 is gone; installed shims are `uchg`'d with `ELECTRON_RUN_AS_NODE=1`

## Checklist

Model: grok-4.6. Harness: Cursor in T3 Code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved desktop command launching to run through Node correctly, preventing unintended duplicate application or backend processes.
  * Preserved existing session handling and exit-code behavior across supported platforms.
  * Desktop backend startup now uses the designated port consistently and provides clearer guidance when that port is unavailable.

* **Tests**
  * Added coverage for corrected desktop launch behavior on macOS- and Windows-style environments.
  * Added coverage for reporting the designated port as unavailable.


adinvadim:
Closing — local machine fix only, not intended for upstream.

PR #12252, opened 2026-09-17
fix(codex): sync generated thread titles

bustakar:
## What changed

T3 now copies its generated thread title to the matching Codex thread.

The server adds an optional thread-naming method to provider adapters. The Codex adapter implements it with `thread/name/set`.

Title synchronization runs when T3 generates or refines a title. It also runs when a Codex session becomes ready, in case T3 generated the title before the provider session was available.

The synchronization code checks the stored title before sending it. It skips manual titles, stale events, and titles that came from the provider. This avoids overwriting user changes and prevents update loops.

A failed title update writes a warning to the server log. It does not interrupt the thread or the active turn.

This PR does not change thread import, session continuation, or manual title editing.

## Why

T3 stored generated titles in its own thread data but did not send them to Codex. The Codex app therefore kept the original thread name, which was often the first prompt or an attachment placeholder.

Sending the generated title through Codex's native naming method keeps the thread name consistent in both applications.

## UI changes

| Before | After |
| --- | --- |
| Before: Codex thread named after the image attachment | After: Codex thread uses the title generated by T3 |

Manual verification with a real Codex session confirmed that T3 generated the title after the first turn and Codex applied the same title to its thread. The thread remained available after the rename.

## Verification

- 284 focused server tests passed across seven test files.
- The orchestration test checks that T3 sends the generated title.
- The Codex adapter test checks that it calls the native naming method.
- The server typecheck passed.
- Targeted lint passed with existing warnings only.
- Formatting and `git diff --check` passed.
- The title sync worked in a real T3 and Codex session.

## Checklist

- [x] I included before and after screenshots for the Codex app
- [x] This PR has no animation changes

## Summary by CodeRabbit

* **New Features**
  * Generated thread titles are now synchronized with supported provider sessions.
  * Thread title updates are applied during active sessions and after new turns generate a title.
  * Codex sessions now support updating the provider-side thread name, with clearer handling when a session or provider thread is unavailable.

* **Bug Fixes**
  * Improved resilience when title synchronization encounters non-interrupting provider errors.


juliusmarminge:

PR #12254, opened 2026-09-17
feat(usage): report OpenCode spend and tokens from its local store

jaikhuranna:
## What changed

The Usage report covered Claude, Codex, and Grok Build but not OpenCode, so OpenCode-only users saw no spend or tokens. This adds OpenCode as a usage source following the same playbook the grok addition (v5) used:

- **Contracts**: `UsageProviderKind` gains `"opencode"`. `USAGE_CONTRACT_VERSION` bumps to 6; `USAGE_MERGE_COMPATIBLE_SINCE` stays at 4 because the change is additive.
- **Server**: a new `opencodeUsageStore.ts` reads assistant rows from OpenCode's own SQLite store (`$XDG_DATA_HOME/opencode/opencode.db`, opened `readOnly`) through a pure `parseOpenCodeMessage` in `usageTranscripts.ts`. The store rides the existing `(size, mtime)` scan cache wholesale — there is no transcript byte stream to resume from.
- **Normalization**: OpenCode reports disjoint tokens (`input` excludes cache reads, `output` excludes reasoning), so reasoning folds into `outputTokens` per the contract's subset invariant; a zero reported cost falls through to the rate table, matching Claude Code subscription turns.
- **Clients**: one provider entry in web and mobile usage presentation (icon already existed).

Deliberately deferred to keep this thin: the per-thread context-window meter, cursor/antigravity sources, and reading `session_message` (v2 projection, empty in the wild; `dedupeKey` is the message id, so dual-reading later is additive-safe).

521 insertions across 14 files (~250 production, ~270 tests), verified with the touched suites: parser, store round-trip, one integrated service scan, scan-cache round-trip, and chart bands.

Worked by GLM (glm-flash-latest) via OpenCode.


jaikhuranna:
## Evidence: OpenCode usage on a real store

**Before** (stable install, same machine): the Usage page shows $0.00 / no activity — the user's entire history is OpenCode sessions, which the usage contract ignored.

Before

**After** (this branch): OpenCode appears with spend, tokens, cache savings, and a per-model breakdown read from the same `opencode.db`.

After

17 sessions · $23.12 · 141M tokens (133M cached), cache savings $5.56. Unpriced models (muse-spark-*) render honestly as Unpriced.

jaikhuranna:
## Evidence: OpenCode usage on a real store

Both screenshots come from the same machine and the same `opencode.db`.

**Before** — the stable install's Usage page: $0.00, no activity. The user's history is entirely OpenCode sessions, which the usage contract ignored.

Before

**After** — this branch: OpenCode appears with spend, tokens, cache savings, and a per-model breakdown, read from the same store (90-day window).

After

31 sessions · $30.27 · 254M tokens (231M cached input), cache savings $7.32. Models without rates (muse-spark-*, nvidia/nemotron) render as Unpriced and are excluded from cost totals.

jaikhuranna:
## Mobile verification

The phone app (T3 Code Dev, this branch, Android emulator) paired to a server running this branch. The Usage widget renders the OpenCode series from the v6 contract buckets with no provider-specific mobile code — it is driven entirely by the shared `PROVIDER_ORDER`, label, and color definitions:

Mobile Usage

OpenCode — $29.79 · 100.0% of cost · 245M tokens (90-day window), confirming the contract addition covers the mobile surface as-is.

PR #12266, opened 2026-09-17
fix(server): stop probing Codex usage for accounts without a subscription

chledowski:
## Problem

A Codex instance pointed at a custom `model_provider` (OpenRouter here) authenticates itself, so `account/read` answers `{"account": null, "requiresOpenaiAuth": false}`. The probe asks for subscription limits anyway, and Codex rejects it:

```
>> {"jsonrpc":"2.0","id":2,"method":"account/read","params":{}}
>> {"jsonrpc":"2.0","id":3,"method":"account/rateLimits/read"}
<< {"id":2,"result":{"account":null,"requiresOpenaiAuth":false}}
<< {"id":3,"error":{"code":-32600,"message":"codex account authentication required to read rate limits"}}
```

(codex 0.154.0, a `CODEX_HOME` holding only a `[model_providers.openrouter]` config and no `auth.json`.)

The instance runs models fine, but the usage view reports a failed probe for a metric that backend does not have.

## Fix

Decide from the account rather than from whether the request happened to succeed. An API key and a self-authenticating backend have no subscription limits: skip the request and publish `unsupported`, which the usage view already renders as "no subscription limits" instead of a warning. `apiKey` was already classified `unsupported` here — it just paid for the request first.

A ChatGPT account that could not authenticate this probe is deliberately *not* included: it has a subscription we failed to read, so it still degrades to `probeFailed` and keeps its last good windows until the next login. A real `-32600` on a ChatGPT account stays visible too, so this does not blanket-suppress failures.

No UI code changed.

## Before / after

The same environment with two Codex instances — a ChatGPT one and the OpenRouter one — plus Claude, on `/usage`.

| Before | After |
| --- | --- |
|  |  |
| "OpenRouter: Codex could not read usage (JSON-RPC -32600)." | No warning; supported bars unchanged. |

## Tests

`vp test run apps/server/src/provider/Layers/ProviderRegistry.test.ts` — added cases for a backend without a subscription (`unsupported`), a signed-out ChatGPT account (stays `probeFailed`), and a ChatGPT `-32600` (stays `probeFailed`, message intact). `tsc --noEmit` for `apps/server` is clean.

`"re-probes when settings change the codex binaryPath"` in that file fails on my machine on an unmodified `main` as well — it picks up a local PATH entry — so it is unrelated to this change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Codex account status detection for accounts without subscription usage.
  - API-key and unsupported custom-provider accounts now report usage limits as unavailable with an appropriate reason.
  - Preserved clear status reporting for accounts requiring authentication.
  - Improved error details when usage information cannot be retrieved, including clearer reporting when usage requests fail.


chledowski (commented):
## Independent Claude review

Reviewed at `ace735b139` against `main`. Ran the two touched test files and `tsc --noEmit` for `apps/server`.

## Findings

**1. Low — a logged-out ChatGPT account now reports usage as `unsupported`, silently dropping its bars** — `apps/server/src/provider/Layers/CodexProvider.ts:688`

`codexSupportsSubscriptionUsage` returns `false` for *any* snapshot without `account.type === "chatgpt"`, including the early-return path at `CodexProvider.ts:474` where `account === null && requiresOpenaiAuth === true` (auth.json missing, or a refresh that failed). That account does have a subscription; it just could not be read this probe.

Failing case: a ChatGPT user whose Codex token stops refreshing. Before this diff the probe published `probeFailed`, and `resolveUsageLimitsAfterProbe` (`apps/server/src/provider/providerUsageLimits.ts:130`) deliberately kept the last good windows. Now the probe publishes `unsupported`, which that same function treats as authoritative and which `collectLimitNotices` (`packages/shared/src/usageLimits.ts:232`) skips outright — so the user's usage bars vanish from the composer and Limits view with no notice explaining why, until they re-authenticate.

Fix: keep the unauthenticated case out of `unsupported`. In `checkCodexProviderStatus`, branch on it first:

```ts
const usageLimits =
  !snapshot.account.account && snapshot.account.requiresOpenaiAuth
    ? makeUnavailableUsageLimits({ checkedAt, reason: "probeFailed" })
    : !codexSupportsSubscriptionUsage(snapshot.account)
      ? makeUnavailableUsageLimits({ checkedAt, reason: "unsupported" })
      : /* existing rateLimits branches */;
```

(equivalently, make the predicate `account.account?.type === "chatgpt" || (!account.account && account.requiresOpenaiAuth)` and let the existing `rateLimits === undefined → probeFailed` branch handle it). A `ProviderRegistry.test.ts` case for `{ account: null, requiresOpenaiAuth: true }` would pin the distinction, which no test currently does — both new suites only cover `requiresOpenaiAuth: false`.

**2. Low — Amazon Bedrock is skipped on assumption, not evidence** — `apps/server/src/provider/Layers/CodexProvider.ts:112`

The PR's reproduction covers a custom `model_provider` (`account: null`); `apiKey` was already classified `unsupported` before this diff. `amazonBedrock` is newly skipped with no reproduction, and the schema carries `usesCodexManagedCredentials` (`packages/effect-codex-app-server/src/_generated/schema.gen.ts:14202`), i.e. a Bedrock account whose credentials Codex manages. If such an account does answer `account/rateLimits/read`, this diff hides limits that previously rendered, and the user has no way to see them. Failing case: Bedrock with `usesCodexManagedCredentials: true` that returns a valid snapshot — the request is never issued and the Limits view shows nothing.

Fix: either confirm against the Codex CLI that `account/rateLimits/read` rejects Bedrock and note that in the comment at `CodexProvider.ts:105-112`, or narrow the predicate to skip only `apiKey` and `account === null && !requiresOpenaiAuth` (the two modes with evidence) and let Bedrock keep its existing failure-degrades path.

## Checked

- Call sites: `account/rateLimits/read` is issued from exactly one place (`probeCodexAppServerProvider`); no other usage path lost a read. Ordering unchanged — `account/read` still precedes the `Effect.all`, so the new dependency adds no round trip and the probe is strictly cheaper for skipped modes.
- Account union is closed (`apiKey | chatgpt | amazonBedrock`), so the predicate is exhaustive today; a future subscription-bearing type would default to `unsupported` (acceptable, flagged implicitly by finding 2's fix).
- Downstream rendering matches the PR's claim: `unsupported` → `limitsNotice` returns "This account has no subscription limits." (`packages/shared/src/usageLimits.ts:396`) and `collectLimitNotices` drops the row entirely — no warning line, no UI change needed.
- `resolveUsageLimitsAfterProbe` semantics for the intended cases (custom/apiKey/Bedrock never had published windows to wipe) — only the logged-out path above is affected.
- ChatGPT path preserved: 3s `timeoutOption`, `Option.getOrElse` "did not answer" message, `Effect.catch` → `codexRateLimitsFailureMessage`, and a real `-32600` still surfaces as `probeFailed` with its message; the `CodexDriver` reset-credit confirmation (`Drivers/CodexDriver.ts:327`) still keys on `probeFailed` for ChatGPT accounts and is unaffected.
- Contract/type surface: `rateLimits?: CodexRateLimitsProbe | undefined` is required by the explicit `undefined` return under `exactOptionalPropertyTypes`; nothing crosses the wire that changed shape, so no `packages/contracts` or client update is implied.
- Tests: `vp test run apps/server/src/provider/Layers/CodexProvider.test.ts apps/server/src/provider/Layers/ProviderRegistry.test.ts` → 67 passed, 1 failed. The failure is `"re-probes when settings change the codex binaryPath"`, which asserts on spawned commands and picks up `/opt/homebrew/bin/brew` from the local PATH; it is environment-dependent and unrelated to this diff, as the PR body states.
- `vp run --filter t3 typecheck` → clean.
- Repo rules: server behavior change ships with focused tests, tests assert observable behavior (the recording client proves the RPC is not issued) rather than mirroring implementation, no docs are made inaccurate, no UI/contract surfaces missed.


---

AI-generated review by Claude using `claude-opus-5` at `medium` effort.
Requested model: `claude-opus-5`.
Configured fallback: `none`. Fallback used: no.

chledowski:
Both findings accepted and fixed in `2455a9690c`.

**1. Signed-out ChatGPT account losing its bars — fixed.** Correct, and it is a real regression: `account: null, requiresOpenaiAuth: true` reached the same `unsupported` branch, and `resolveUsageLimitsAfterProbe` treats that as authoritative, so a user whose Codex token stopped refreshing would have lost their last good windows with no notice. The predicate is now `codexAccountHasNoSubscriptionUsage` and answers `false` for that path, which falls through to `rateLimits === undefined → probeFailed` exactly as before. Pinned by a new `ProviderRegistry.test.ts` case asserting `unauthenticated` auth with `probeFailed` usage.

**2. Amazon Bedrock skipped on assumption — fixed by narrowing.** Also correct; I had no reproduction for Bedrock and `usesCodexManagedCredentials` makes the assumption worse than idle. The predicate now covers only the two modes the reproduction supports — `apiKey`, and `account === null && !requiresOpenaiAuth` — and Bedrock keeps its existing request-and-degrade path. A reader test asserts the request is still issued for it, named so the reason is visible rather than looking like an oversight.

Net effect on the original bug is unchanged: the OpenRouter-style custom backend issues no `account/rateLimits/read` and publishes `unsupported`.

Verified: `vp test run apps/server/src/provider/Layers/CodexProvider.test.ts apps/server/src/provider/Layers/ProviderRegistry.test.ts` → 67 passed, 1 failed (`"re-probes when settings change the codex binaryPath"`, the pre-existing PATH-dependent failure you also reproduced). `tsc --noEmit` for `apps/server` clean.

chledowski:
Rewrote the branch as a single commit (`37375b9`) after the review above, and shrank the change to the minimum that fixes the bug:

- Dropped the extracted `requestCodexRateLimits` helper, its exported surface, and the recording-client tests. The guard is now one conditional at the existing call site, and the behavior is covered where it is observable — the provider status.
- Both review findings stay fixed: the predicate covers only `apiKey` and `account === null && !requiresOpenaiAuth`, so a signed-out ChatGPT account keeps its last good windows on `probeFailed`, and Amazon Bedrock keeps its existing request-and-degrade path rather than being skipped on an assumption I had no reproduction for.

Diff is now 2 files, +90/-33, most of it re-indentation under the new conditional. Added before/after screenshots of `/usage` to the description.

chledowski (commented):
## Independent Claude review

## Findings

No defects found in this diff.

Notes that did not rise to findings:

- **`amazonBedrock` accounts still take the probe path** (`apps/server/src/provider/Layers/CodexProvider.ts:115`). A Bedrock account is not an `apiKey` and is not `account === null`, so it still issues `account/rateLimits/read` and will surface `probeFailed` if Codex rejects it. This is pre-existing behavior, unchanged by the diff, and the helper's doc comment scopes itself to API keys and self-authenticating backends, so it is not a regression introduced here — but it is the same class of bug the PR fixes, and a plausible follow-up.
- **The skip branch itself is only tested indirectly.** The three new tests drive `checkCodexProviderStatus` with injected snapshots, which covers the classification (the user-visible outcome), but nothing asserts that `account/rateLimits/read` is no longer sent when the account has no subscription. There is no existing test harness for `probeCodexAppServerProvider` (no test drives it against the app-server mock peer), so adding one would be new scaffolding rather than a gap in the pattern the repo already follows. Acceptable as-is under "focused tests for that behavior".

## Checked

- **Full diff at `37375b9`** against `main` — only `CodexProvider.ts` and `ProviderRegistry.test.ts`; no UI, contract, or schema changes, consistent with the PR body's "No UI code changed."
- **Classification correctness.** `codexAccountHasNoSubscriptionUsage` (`CodexProvider.ts:114`) reproduces the old `apiKey → unsupported` rule exactly and adds only `account === null && requiresOpenaiAuth === false`. Confirmed against the generated schema (`packages/effect-codex-app-server/src/_generated/schema.gen.ts:38503-38510`) that `requiresOpenaiAuth` is a **required** `Schema.Boolean` on `V2GetAccountResponse`, so an older/partial response cannot silently fall into the new branch by omitting the field. The account union is `apiKey | chatgpt | amazonBedrock` (`schema.gen.ts:14195`).
- **The signed-out ChatGPT case is unreachable by the new branch.** `!account && requiresOpenaiAuth` returns early from `probeCodexAppServerProvider` (`CodexProvider.ts:440`) before the `Effect.all`, and `accountProbeStatus` (`:565`) keeps it `unauthenticated` / `probeFailed`. Verified the new helper agrees (returns `false`), so the two call sites cannot disagree about one snapshot.
- **`Effect.all` shape.** The ternary is evaluated synchronously on `accountResponse` before the concurrent block; the skip arm's `Effect.succeed(undefined)` widens the tuple element to `CodexRateLimitsProbe | undefined` and cannot introduce an uncaught error channel (the request arm still has its `timeoutOption` + `Effect.catch`). The `rateLimits?: CodexRateLimitsProbe | undefined` widening at `:74` is what the `satisfies` at `:489` needs under `exactOptionalPropertyTypes`; nothing else constructs this snapshot in production.
- **Downstream usage-limit semantics.** `resolveUsageLimitsAfterProbe` (`providerUsageLimits.ts:124`) treats `unsupported` as authoritative, so a previously-published window set is replaced — correct for an account that genuinely has none, and recoverable: a later probe after `codex login` republishes real windows. `applyUsageLimitsUpdate` (`:66`) makes `unsupported` sticky against *runtime* turn updates only; for a custom `model_provider` backend that emits no rate-limit notifications this is inert, and it matches the existing `apiKey` treatment.
- **Other consumers.** `checkCodexProviderStatus` has exactly one production caller (`Drivers/CodexDriver.ts:200`). The reset-credit confirmation path (`CodexDriver.ts:327`) keys off `probeFailed` and only applies to ChatGPT accounts, so it is unaffected.
- **Tests run locally:** `vp test run apps/server/src/provider/Layers/ProviderRegistry.test.ts` → 54 passed, 1 failed. The single failure is `"re-probes when settings change the codex binaryPath"`, which asserts on spawned commands and picks up `/opt/homebrew/bin/brew` from the machine's PATH — environment-dependent and unrelated to this diff, matching the PR body's claim. All three new cases pass.
- **Repo rules:** backend behavior change ships with focused tests; no repo-wide checks run; no docs change warranted (no new architectural constraint or user-facing procedure); no plan/scratch artifacts committed; commit title follows conventional format.


---

AI-generated review by Claude using `claude-opus-5` at `medium` effort.
Requested model: `claude-opus-5`.
Configured fallback: `none`. Fallback used: no.

chledowski:
Second review pass on `37375b9` came back with no defects. Both of its non-findings are deliberate and I am leaving them: Amazon Bedrock keeps its existing request-and-degrade path because I have no reproduction showing Codex rejects it (happy to extend if a maintainer knows the answer), and the skip branch is covered through the provider status rather than through new app-server probe scaffolding, to keep this change small.

juliusmarminge:

chledowski:
Thanks @juliusmarminge - let me know if there are any changes where you'd need some help :)

PR #12289, opened 2026-09-17
fix(web): keep thread notification tags within the Windows toast limit

### LINKED ISSUE #12287: [Bug]: Completion notifications never appear on Windows: 73-character notification tag exceeds the platform tag limit
## Environment

- T3 Code `0.0.42` stable (desktop, Windows install; verified from `resources/app.asar` and the exe version)
- Electron `44.1.0` (Chrome `152.0.7977.65`)
- Windows 11 25H2, build `26200.9457`
- Backend in WSL2 (Ubuntu), client served to the desktop window via `t3code://app`
- Notification mode `notifications-and-sound` (`inAppNotificationsEnabled: false`)

## Summary

When a turn completes while the desktop window is minimized or unfocused, Windows never shows the toast. The app still plays the notification sound and sets the red dot on the taskbar icon, and the Action Center stays empty.

The cause is the notification `tag` length. The coordinator builds ([ThreadNotificationCoordinator.tsx:181](https://github.com/pingdotgg/t3code/blob/v0.0.42/apps/web/src/components/ThreadNotificationCoordinator.tsx#L181)):

```
tag: `${environmentId}:${thread.id}`
```

Both IDs are UUIDs, so the tag is 36 + 1 + 36 = 73 characters. On Windows, Chromium silently drops renderer notifications when the tag exceeds a platform limit, and the budget is shared with the origin, so the threshold varies per app and Electron version. The `Notification` constructor still succeeds and no `error` event fires, which is why the app-side effects (sound, taskbar badge) still happen while the toast never appears.

## Steps to reproduce

1. Settings -> General -> Thread notifications: choose a mode that includes notifications.
2. Minimize the T3 Code window (or focus another app).
3. Send a message to an agent and let the turn finish.

## Expected

A Windows toast appears and lands in the Action Center.

## Actual

Sound plays, the red dot appears on the taskbar icon, no toast, nothing in the Action Center, and `LastNotificationAddedTime` for `com.t3tools.t3code` never advances.

## Evidence

Reproduced with a minimal Electron `44.1.0` app that mirrors T3's setup: scheme `t3code` with the same `registerSchemesAsPrivileged` privileges and host `app`, the same AppUserModelID (`com.t3tools.t3code`), `silent: true`, title "Thread completed", a real thread title as the body, and the window minimized. Checked two ways: the renderer's `show` event and Windows' own notification database (`wpndatabase.db`).

| tag | length | delivered |
|---|---|---|
| `control1` | 8 | yes (`show` fired, DB row) |
| `<threadId>` alone | 36 | yes |
| `<envUuidPrefix>:<threadId>` | 45 | yes |
| `<envUuid>:<threadId>` (the actual format) | 73 | **no** (no `show`, no DB row) |

Boundary measurement with a slightly longer origin (`repro-app://app`): 46-character tags deliver, 47 and up are dropped. With T3's shorter origin (`t3code://app`), 45 delivers and 73 drops. The threshold moving with the origin matches Electron's behavior where the Windows toast tag is derived from a notification ID that includes the origin.

On the affected machine, the Windows notification database has no row from the main window origin at all. Every T3 notification that ever arrived had a short tag (manual test notifications created from a preview pane). The app-side creation path is confirmed to run: the taskbar badge is only set after `new Notification(...)` succeeds.

Related: electron/electron#40433 ("Windows 11 web notifications don't work when tag length is greater than 34") and electron/electron#42239, an attempted fix that hashes the tag. Still reproducible on Electron 44.1.0.

## Suggested fix

Keep the tag short:

- `thread.id` alone (36 characters, thread IDs are already unique), or
- a short digest of `${environmentId}:${thread.id}` (for example 16 hex characters)

Both the 36 and 45 character variants above were verified to deliver. I would pick the 36-character one for margin, since the limit varies with the origin and Electron version. Per-thread replacement semantics are preserved either way. A unit test asserting the composed tag stays under, say, 32 characters would guard against regressions.

This also affects the web client in Chromium browsers on Windows, since it is the same tag and platform path. macOS and Linux are not affected.

I can share the full minimal repro app and the raw logs if useful.

satyalyadav:
## What Changed

- Added `threadNotificationTag(environmentId, threadId)` in `apps/web/src/threadNotifications.ts`: a stable 16-character digest of the `environmentId:threadId` pair.
- `ThreadNotificationCoordinator` now uses that tag for `new Notification({ tag })` and as the pending-notification map key, so per-environment/thread replacement behavior is unchanged.

## Why

Windows silently drops renderer notifications when the tag plus origin exceeds the platform toast budget (electron/electron#40433), and the composite tag is 73 characters because `environmentId` and `threadId` are both UUIDs. The app-side path still ran, which is why affected users heard the sound and saw the taskbar badge while the toast never appeared and the Action Center stayed empty. The digest keeps environment and thread uniqueness while staying well below the documented 34-character bound and the origin-dependent budget.

Fixes #12287

## Tests

- `apps/web/src/threadNotifications.test.ts`: the tag is 16 characters, stable, and distinguishes environments and threads.
- Existing coordinator assertions now expect the digest. `vp test run --project unit` over the new test plus both coordinator test files: 32 passed.
- `tsc --noEmit` in `apps/web` passes.

Independent Windows verification by @ScottN-PV (Windows 11 25H2, Electron 44.4.2, source build): the shortened tag reaches Windows Notification Center, the 73-character tag fails with the documented Electron error, banners and sound work after reload, and notification history shows distinct tags per conversation with same-conversation replacement while the other conversation's entry stays unchanged. Click-to-open did not bring the app forward, and that is not a regression from this PR: unpatched main rejects the tag, so no notification exists there to click.

## UI Changes

Not applicable; no visual changes.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved desktop thread notification handling so notifications for the same thread are more reliably grouped, updated, or dismissed.
  - Added consistent notification identification across environments and threads, reducing the risk of duplicate or incorrectly replaced notifications.
- **Reliability**
  - Notification identifiers now remain within platform limits while preserving distinction between different threads and environments.


ScottN-PV:
I tested this patch on Windows. Completion banners and sound worked after reloading the app, but clicking a notification did not bring T3 to the foreground.

- Setup: Windows 11 Home 25H2, build 26200.9457; T3 built from source with Electron 44.4.2. The app loaded at `t3code-dev://app`, using separate test settings and conversations.
- Code tested: the changes from PR commit `13ebd2589a2970ecd759fc90061e2e3d2d226f30`, applied to upstream commit `5781b5240bd5d2e21c651f6b228975ac40cbd67b`.
- Tag-length comparison: I created two notifications through the development app's JavaScript Console. A notification with tag `t3-control` appeared in Windows Notification Center. A notification with a tag consisting of 73 `x` characters failed: Electron logged a Windows error, `The size of the notification tag is too large. (HRESULT: -2143420138)`. Windows' notification-history API contained the short-tag notification and no long-tag notification. Do not disturb was On for both tests; I was checking storage in Notification Center, not whether a banner appeared.
- With the patch and Do not disturb Off, real Codex completion notifications displayed banners. Electron logged the shortened app tag `828d2b52f503fb00`. The first test had no audible sound. After reloading T3 with Ctrl+R, both banner and sound worked in two further tests: one with T3 minimized, and one with T3 unminimized behind another application. I have not established why sound was absent on the first attempt.
- Automated checks: before applying the patch, 16 notification-coordinator tests passed. With the patch, all 19 tests in the coordinator and tag-helper test files passed. Lint on the changed files and the web package's typecheck also passed.

Clicking the banner did not bring T3 to the foreground in either window state. Clicking its Notification Center entry also failed to bring T3 forward in the minimized test. Electron logged `Notification clicked`, but that does not prove T3's click-handling code ran. I had to click T3's taskbar icon. The agent then appeared to continue for about one second before the final response became visible. I do not know whether the agent was still working or the interface was catching up. This PR does not directly change the code that handles notification clicks. I haven't established whether the click failure existed before this patch.

These results support the tag-length fix and confirm banner/sound delivery with the patch in this development setup. Click-to-open remains unresolved. I have not tested a patched Windows installer build, whether notifications from different conversations remain separate, macOS/Linux, or T3 in a standalone Windows browser. I have no video recording; the evidence consists of manual observations and local diagnostic logs.

AI assistance: OpenAI Codex, using Codex CLI 0.155.1 through its app-server integration in the T3 Code desktop app; model: `gpt-6-astra`; reasoning effort: `medium`. Codex assisted with test setup, running automated checks, interpreting diagnostic logs, and drafting this report. I performed the manual Windows checks, reported the visible and audible outcomes, and reviewed the report before posting.

satyalyadav:
Thanks for the detailed Windows verification, and for isolating the tag failure with the 73-character tag. On click-to-open: this PR only changes the tag value and the pending-notification map key, so it does not touch click handling, and your report does not establish it as a regression. It is worth checking on unpatched main, and if it reproduces there it belongs in its own issue with that repro. If you have time, the behavior this PR actually changes is replacement, so a check that two different conversations stay separate and that a second completion replaces the first would be useful evidence.

ScottN-PV:
Thanks for the suggested checks. I ran an automated native Windows follow-up using the same isolated source build: PR `13ebd2589a2970ecd759fc90061e2e3d2d226f30` applied to baseline `5781b5240bd5d2e21c651f6b228975ac40cbd67b`, Electron 44.4.2.

Real Codex turns were started through the isolated server's local API while the desktop client stayed in the background. Windows' notification-history API showed:

- First completion in conversation A: one entry for A.
- Completion in conversation B: two entries, A and B, with distinct tags.
- Second completion in A: still two entries. A reused its original tag and replaced its previous entry; B's tag and notification XML were unchanged. I changed A's conversation title before the second turn so the replacement could be distinguished in the stored notification body.

The application tags were `7fd41c906aca1aa6` for A and `1436fb8f1893e7c7` for B. Electron's log also recorded the old notification being hidden when A completed again. A local coordinator test covering A → B → A passed alongside the 16 existing coordinator tests.

This verifies separation and replacement in native Windows notification history for that development build. It does not add a visual banner/sound check.

I also tested clean, unpatched main at `c26119ada3565dda55bc3e86fdf47c7f9088e4b5`, in a separate worktree and copied disposable profile, with Electron 44.4.2. A real turn completed while the renderer was unfocused and notification permission was granted. Electron attempted the original 73-character application tag, and Windows rejected it with `The size of the notification tag is too large. (HRESULT: -2143420138)`. Notification history was unchanged: no entry for that completion appeared. Consequently there was no new notification to click, so this does not establish click-to-open behavior on unpatched main. I agree that the earlier click observation is not evidence of a regression from this PR. No Windows notification settings were changed for these checks.

AI assistance: OpenAI Codex; harness: Codex CLI 0.155.1 app-server integration; host/interface: T3 Code desktop app; model: `gpt-6-astra`; reasoning/effort: `medium`. Contribution: local API test setup, automated test execution, Windows notification-history inspection, evidence comparison, and drafting this reply. Human involvement: requested the follow-up and automatic testing, and confirmed that the model, effort, runtime version, and host remained the same as in the earlier verification. These follow-up results are automated observations; no new manual Windows checks were performed. Attribution settings are user-confirmed; runtime metadata was not independently captured again for this follow-up.

satyalyadav:
Thanks for the thorough follow-up, Scott. The A to B to A history check plus the unpatched-main control is exactly the evidence this needed, and I appreciate you closing the click-to-open question with it. I have added a summary of your verification to the PR description.

PR #12290, opened 2026-09-17
fix(web): paste markdown with inline code inside bold, italic, or strikethrough

### LINKED ISSUE #12247: Pasting markdown with inline code inside bold, italic or strikethrough silently does nothing
### What happened

Pasting into the desktop chat composer silently does nothing — no text appears, no error, no toast. It reproduces with any markdown containing inline code wrapped in bold, such as an assistant response full of ``**`path/to/File.cs`**`` headings. Other pastes into the same composer work, which initially made it look like a size limit.

### Diagnosis

The composer's paste handler calls `event.preventDefault()` before it inserts anything (`apps/web/src/components/ComposerPromptEditorTiptap.tsx:948`), so any failure in the insert step swallows the paste with no feedback.

The insert step builds Tiptap JSON via `buildTiptapContent` (`apps/web/src/composer-rich-text-doc.ts:166`). When inline code sits inside bold, italic or strikethrough, `parseInlineMarkdown` nests the marks and the builder emits a text node carrying both, e.g. `marks: [{type:"bold"},{type:"code"}]`.

That is an invalid mark set. `@tiptap/extension-code@3.31.3` declares `excludes: "_"`, so the `code` mark cannot coexist with any other mark. ProseMirror rejects the node instead of inserting it, and because `preventDefault()` already ran, the paste is lost silently.

Disabling the rich text composer makes the same paste succeed, which confirms the mechanism: plain mode passes `styling: false`, so `parseInlineMarkdown` never runs and no marks are emitted.

Size and line count are not factors. An 8 KB single line and a 110-block multi-line paste both insert fine as long as no mark set combines `code` with another mark.

Suggested direction (not a patch): have `buildTiptapContent` avoid emitting `code` alongside other marks — either drop the sibling marks on a code span, or leave the surrounding markers literal, matching the "unmatched markers stay literal" rule already stated in `composer-rich-text.ts:10`.

### Steps to reproduce

1. Open the desktop app with a local server and focus the chat composer, with the rich text composer enabled (the default).
2. Copy this exact 8-character string to the clipboard: ``**`x`**``
3. Paste into the composer.

Nothing is inserted. Disabling the rich text composer and repeating step 3 inserts the text correctly.

Controls that all paste correctly, confirming the trigger is the mark combination and not size or line count:

| Pasted | Marks produced | Result |
|---|---|---|
| `https://github.com/pingdotgg/t3code` | none | inserts |
| `alpha` + newline + `bravo` | none | inserts |
| single line, 8,000 chars | none | inserts |
| `alpha **bold** bravo` | `bold` | inserts |
| `` `x` `` | `code` | inserts |
| ``**`x`**`` | `bold+code` | **nothing inserted** |

`` *`x`* `` and `` ~~`x`~~ `` produce `italic+code` and `strike+code` and are expected to fail identically.

### Version

`0.0.43-nightly.20260917.1851` (tag `v0.0.43-nightly.20260917.1851`, commit `0150c6a5`)

### Environment

macOS (darwin arm64 27.0.0), Node v26.8.2, launched as `npx t3@nightly`, desktop app against a local server on `127.0.0.1:3773`, `composerRichTextEnabled` at its default of `true`

### Evidence

```
Mark sets emitted by buildTiptapContent() for the user's 6,483-byte payload
(run against the pure builder at tag v0.0.43-nightly.20260917.1851):

  code           x 43
  bold+code      x  8   <-- invalid; `code` declares excludes: "_"
  bold           x  6
  italic         x  1

Minimal cases:
  "**bold**"     marks=[bold]
  "`code`"       marks=[code]
  "**`x`**"      marks=[bold+code]     <-- nothing inserted
  "~~`x`~~"      marks=[strike+code]
  "*`x`*"        marks=[italic+code]

@tiptap/extension-code@3.31.3/dist/index.js:
  54: name: "code",
  58: excludes: "_",
  59: code: true,
(unpatched in this build; patches/ contains no @tiptap entry)

~/.t3/userdata/logs/server.trace.ndjson   — no failures around the paste
~/.t3/userdata/logs/desktop.trace.ndjson  — all spans exit Success; renderer-side
                                            exceptions are not captured here, which
                                            is why this failure left no trace at all
```

### Related issues

#8331 (pasting code containing `@` creates invalid mentions) touches the same paste path but fails in the mention tokenizer, not the mark schema. #5626 (composer lag with large input) is a performance issue, not a silent drop. Neither is a duplicate.

### Fix applied or workaround

Nothing was written to the machine. Verified workaround: disabling the rich text composer setting makes the paste succeed. Also available: save the text as a `.txt` file and attach it, or rely on pastes ≥32 KiB auto-folding into a `pasted-text.txt` attachment (`packages/client-runtime/src/textPaste.ts:1`), which bypasses the inline insert entirely.

### Filed by

Claude Code (`claude-opus-5`) via `t3 triage`

Gigioxx:
Pasting markdown with inline code inside bold, italic, or strikethrough (``**`x`**``, ``*`x`*``, ``~~`x`~~``) into the rich text composer inserted nothing and showed no error. Tiptap's code mark declares `excludes: "_"`, so the `bold+code` spans the markdown parser produces fail `node.check()` inside `insertContent`. The paste handler had already called `preventDefault()`, so the clipboard was dropped silently.

The composer's code mark now only excludes itself, matching the other marks and the serializer, which already emits ``**`x`**`` for `bold+code`. Typing ``**a `code` c**`` also keeps bold across the code span now instead of dropping it. Plain mode is unchanged. The round-trip tests run `doc.check()` so schema-invalid mark sets fail the way `insertContent` does.

Fixes #12247.

Verified with 87 focused rich text tests (4 fail before the fix), web typecheck, targeted lint, and isolated Chromium against a local dev server.

Before, pasting ``Review **`apps/web/src/App.tsx`** and *`utils.ts`*, then ~~`legacy.ts`~~``:

Before: composer stays empty after the paste

After:

After: the paste inserts bold, italic, and struck code spans

Model: Claude Opus 5. Harness: Claude Code via T3 Code.

## Summary by CodeRabbit

* **New Features**
  * Improved rich-text code formatting in the composer.
  * Code formatting now works correctly alongside bold, italic, and strikethrough styles.
  * Preserved combined formatting when entering or converting supported Markdown content.

* **Bug Fixes**
  * Prevented formatted code content from being rejected or lost during editing and conversion.


Gigioxx:
[claude-opus-5] Responding on behalf of Guille
-----

@coderabbitai Not adding the editor-level paste test. `doc.check()` in the round-trip helper is the same validation `insertContentAt` runs on every pasted node (`@tiptap/core` calls `node.check()` before `tr.replaceWith`), so the four new cases fail with the exact `RangeError` from the issue whenever the schema rejects code inside emphasis. What it would add on top is a guard that `ComposerCodeExtension` stays in the component's extension list, and this repo does not test wiring. No test in `apps/web` mounts the Tiptap editor today and the unit project has no DOM environment, so the mount would be new infrastructure for a nitpick. The paste path was verified in Chromium instead (screenshots in the description).

EMOEMOJAI:
Confirming another desktop occurrence: with the rich text composer enabled, pasting **`xxxx`** inserts nothing. We independently reproduced the underlying failure with the current document schema: validation throws `RangeError: Invalid collection of marks for node text: bold,code`. The same input validates and preserves the literal Markdown in plain-text mode. This matches the issue addressed by this PR. Thanks for the fix!

PR #12294, opened 2026-09-17
fix(server): keep a failed usage probe marked when a sparse update lands

### LINKED ISSUE #12170: Claude usage limits: a rate_limit_event after a failed probe is published as a complete window list
Reading `apps/server/src/provider/providerUsageLimits.ts` at 052c7ae: when the boot probe fails, the published limits are `{ windows: [], unavailable: { reason: "probeFailed" } }`. A `rate_limit_event` during a turn then goes through `applyUsageLimitsUpdate`, which merges its one window onto the empty list and rebuilds the snapshot without `unavailable`. The result looks like a full read with only `seven_day`, and `resolveUsageLimitsAfterProbe` keeps it over later failed probes because the marker is gone.

Seen on 0.0.43-nightly.20260916.1825 after the update restarted the app: the Claude cache held only the weekly window for about 6 minutes, until a refresh-all re-ran the probe. A second Claude instance with no active turn had no windows the whole time.

Possible fix: keep `unavailable`, or mark the snapshot partial, when merging onto a snapshot that had it. Related: `hydrateCachedProvider` in `providerStatusCache.ts` does not carry `usageLimits` across a restart, so every restart starts from nothing.

Gigioxx:
## Problem

Closes #12170.

When the Claude boot probe cannot read `get_usage`, the published limits are `{ windows: [], unavailable: { reason: "probeFailed" } }`. A `rate_limit_event` during the next turn names one window (`seven_day`), and `applyUsageLimitsUpdate` merged it onto the empty list and rebuilt the snapshot without `unavailable`. That lone weekly window then looked like a complete read: the Limits view drew only the weekly bar, and `resolveUsageLimitsAfterProbe` kept it as the last good set through every later failed probe until a probe finally succeeded. A merge onto no snapshot at all (a probe that timed out at a higher level, or one that had not finished yet) had the same shape.

## Fix

`applyUsageLimitsUpdate` now carries `previous.unavailable` onto the rebuilt snapshot, and marks a merge onto `previous === undefined` as `probeFailed`. The sparse windows are still stored, but the snapshot stays incomplete until a successful probe replaces it, so clients keep showing "Could not read limits." and a later failed probe no longer keeps a partial set as last-good. `unsupported` is still rejected outright, and the no-op identity path is unchanged for complete snapshots.

The `partial` reason with bars plus a notice from the triage is out of scope here, as is hydrating `usageLimits` from the status cache on restart. A marked snapshot still gives way to the next failed probe (the triage's "leave that" option): no client draws windows while `unavailable` is set, so there is nothing visible to keep.

## Evidence

Isolated worktree environment, boot probe forced to drop `usage` locally (not committed), then one Claude turn that streamed a `rate_limit_event` for `seven_day` at 0.97. Usage → Limits after that turn:

| Before | After |
| --- | --- |
| Before: Claude shows a single weekly bar with no notice | After: Claude still reports "Could not read limits." |

Before the fix, "Refresh limits" with the probe still failing kept the weekly-only card. After the fix, the same refresh replaces the marked snapshot with a fresh `probeFailed`.

Tests: `vp test run src/provider/providerUsageLimits.test.ts src/provider/makeManagedServerProvider.test.ts src/provider/Layers/ProviderRegistry.test.ts src/provider/Layers/ProviderAdapterRegistry.test.ts` in `apps/server`, 70 tests passing, including a new regression for the merge onto `probeFailed` and onto `undefined`, plus the existing "keeps live usage windows across a failed probe". Server typecheck and targeted lint clean.

Model: Claude Opus 5 (1M context) · Harness: Claude Code in T3 Code

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved usage-limit handling when availability checks fail or return incomplete data.
  - Failed checks now remain clearly marked while partial limit updates are received.
  - Usage-limit information is preserved when updates contain only a subset of available windows.
  - Repeated identical updates no longer produce unnecessary changes.
  - Failed availability results now take precedence over incomplete published limit data.


juliusmarminge:

PR #12376, opened 2026-09-18
fix(web): show the scroll-to-end pill when the chat is not at the newest message

### LINKED ISSUE #12372: [Bug]: Scroll-to-end pill stays hidden after thread switch strands viewport above the end
### Before submitting

Related: #5903 (this is the stale-pill half split out per discussion there).

### Area

apps/web

### Steps to reproduce

1. Open a thread with enough history to overflow the viewport.
2. Send a prompt, then switch to a different thread while the turn runs.
3. Return to the first thread after its response has landed.
4. Observe the viewport: it can be stranded above the newest message, often farther up than where it was left.
5. Observe the "Scroll to end" pill: it does not appear. Nudge the scroll up/down a little and the pill appears.

### Expected behavior

If the viewport is not at the live edge, the "Scroll to end" pill is visible. The end-state tracking reflects the actual viewport position after a thread switch settles, including content whose height changes after rendering.

### Actual behavior

The pill stays hidden on the stranded viewport, as if the state believes it is already scrolled to the end when it is not. Only a manual scroll nudge (which fires real scroll events) makes the pill appear.

### Impact

The stranded viewport looks settled, and the missing pill removes the visible recovery path; END key is the only recourse.

### Version or commit

Local checkout @ dc49767ae5

### Environment

T3 Code web/desktop client (Linux). Provider-independent; affects the shared web timeline.

### Logs or stack traces

No associated exception or console error.

_No response_

### Workaround

Nudge the scroll wheel up/down until the pill appears, or press END to jump to the live edge.

### Suspected mechanism (from code reading, not runtime-verified)

- The position cached for the background thread was saved with `atEnd: true` (`timelineScrollAnchoring.ts`), so the thread-switch effect in `ChatView.tsx` takes the `scrollToEnd` branch and hides the pill.
- Transient `isAtEnd=false` scroll events during settle are suppressed by the post-open live-follow guard (`onIsAtEndChange` early return when the scroll generation still matches) plus the 150ms show-debounce.
- If row heights settle after that (streamed markdown/tool output measuring late), the viewport strands with `isAtEndRef` stale-`true` and no further scroll event corrects it.
- A manual wheel gesture bumps the generation via `cancelTimelineLiveFollowForUserNavigation`, letting the next `isAtEnd=false` through so the pill appears.

A post-settle reconciliation (re-resolve `isAtEnd` from the list state a frame or two after position restore completes and sync the pill) would cover this class of stranding regardless of which layout pass caused it.

bompus:
## What Changed

- After chat content finishes changing size (messages or images loading late, a response finishing, reopening a thread), the timeline now double-checks whether you are actually at the newest message. If not, the "scroll to end" pill appears.
- The check never moves the view by itself — it only controls the pill.
- Two follow-ups simplified the check itself (fewer moving parts, same behavior) and fixed a narrow case where switching threads at just the wrong moment could skip the new thread's check.

## Why

Fixes #12372. The chat view could end up stuck above the newest message with no pill to jump down — only nudging the scroll or pressing END helped. This happens because content can grow without firing a scroll event, so the app still believed it was already at the end.

## Verification

- Tests: 70/70 pass (62 existing + 8 new covering the pill behavior).
- Typecheck clean; lint matches baseline (22 warnings, 0 errors, 0 new).

## UI Changes

<table>
  <tr>
    <th>Before (Pill-sync harness (deterministic))</th>
    <th>After (Pill-sync harness (deterministic))</th>
  </tr>
  <tr>
    <td><video src="https://github.com/user-attachments/assets/d521621e-4fe3-4fb8-b6a9-33ac62cd4ab9" width="100%" controls></video></td>
    <td><video src="https://github.com/user-attachments/assets/1a5ae864-548e-4bc6-869c-34ca7c47db4d" width="100%" controls></video></td>
  </tr>
</table>

The bug only shows up with unfortunate timing, so it can't be captured in a live recording on demand. The clips above are the next best thing, rendered from real test runs: before — the new tests fail on the old code (the pill never appears); after — the full suite passes.

## Checklist

- [x] I included before/after evidence — test-harness clips above (live race not capturable on demand)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved chat timeline positioning after messages resize, responses finish, threads are restored, or conversations change.
  - More reliably detects when the view is away from the latest message and updates the navigation state.
  - Prevents automatic scrolling when users have intentionally moved away from the conversation end.
  - Rechecks timeline position after anchored spacing or disclosure settling completes.
  - Improved live-follow behavior so the timeline settles correctly without disrupting manual navigation.


bompus:
Update: this is handled — when the send-time anchoring or a folding animation finishes, the timeline re-checks whether you are at the newest message, so the pill can't get stuck hidden. Nothing scrolls on its own.

bompus:
Follow-up pushed (rebased onto upstream/main): simplified the re-check itself — removed a redundant watcher and merged two cleanups. No behavior change; tests and checks green.

bompus:
Follow-up pushed: review caught a narrow case — switching threads at just the wrong moment could skip the new thread's check. The scheduler now re-arms instead of skipping, plus a regression test for it. 70/70 tests pass.

bompus:
Fusion tie-break follow-up pushed (851e4c7898): it caught a real hole in my thread-guard change.*

*The hole:* the scheduler deduped on a pending handle regardless of thread key — a thread-A frame pending across a switch swallowed thread B's post-restore schedule, then self-dropped on key mismatch. B lost its mount verification (a narrow regression of the E3 cleanup vs the old every-render effect, which cancelled at commit).

*The fix (one line):* cancel-then-schedule in `scheduleSettledEndVerification` — the reschedule lands in the same rAF phase so bursts still share one frame, and the key guard still covers a stale frame with no fresh schedule behind it. Plus a cross-thread handoff regression test (schedule under A, switch to B mid-frame, flush → B verifies), proven to fail without the fix and pass with it.

*Also in this push:* PR body evidence re-rendered at 70/70 (62 existing + 8 new); fusion also confirmed the two declined eliminations (restoring effect, useEffectEvent) were rightly declined. Verification: 70/70 tests, lint 22w/0e (= baseline), tsc clean.

PR #12378, opened 2026-09-18
fix(web): make PR badge hover lists easy to click

Bil0000:
Moving the mouse from a PR badge into its list crossed a 4px gap. In the sidebar, the thread tooltip could take over there and close the list before a PR could be clicked.

Remove the gap in the shared sidebar/composer PR control. Use the same clickable list row and glass popup for a single PR, including branch PRs with no saved links. Keep the existing badge actions and modified-click behavior.

### Verification

- Reproduced the mouse failure in the real app before the change; the same automated Chromium path passes after it.
- Checked all five linked PRs and a single branch PR at 1280px and 900px widths, plus the composer footer, keyboard focus/Escape, and Ctrl-click.
- All 53 focused ThreadStatusIndicators tests pass; targeted lint, format check, and web typecheck pass.

Scope: one shared web component, used by web and desktop. No server, contract, or mobile changes.

### Single PR

| Before | After |
| --- | --- |
| Single PR before | Single PR after |

### Multiple PRs

| Before | After |
| --- | --- |
| Multiple PRs before | Multiple PRs after |

### Mouse path before and after

https://github.com/user-attachments/assets/705dce4e-8a8f-4b62-b78e-eda1a47ae869

Model: GPT-6. Harness: Codex.


Bil0000:

PR #12404, opened 2026-09-18
fix(server): retain editors when Windows discovery times out

### LINKED ISSUE #4697: [Bug]: No installed editors found
### Before submitting

### Area

Not sure

### Steps to reproduce

1. Open the app
2. Can't press the Open button on the top right

Image

### Expected behavior

Dropdown shows VS Code, Cursor, Explorer, etc.

### Actual behavior

Dropdown shows No installed editors found

### Impact

### Version or commit

Nightly

### Environment

Windows 11 - 0.0.30-nightly.20260728.928

### Logs or stack traces

```shell

```

_No response_

### Workaround

No workaround found. Did not have this issue previously.

Salloxy:
## What Changed

Windows editor discovery can find VS Code or Cursor, then discard them when a later PATH probe exceeds the five-second deadline. Preserve the editors found before timeout and give Windows scans ten seconds. Fast scans still return immediately; macOS and Linux retain their five-second budget.

Only completed scans enter the existing discovery cache. Timeouts can be retried on the next request, and client cancellation still propagates without poisoning the cache.

Run editor, file-reveal, and remote-target discovery concurrently so their deadlines do not add up against the client's 15-second connection-establishment limit. File reveal is still advertised only when a usable file manager was discovered. The service API and client protocol are unchanged.

## Why

Fixes #4697.

Related approaches: #10694 and #11071. This change is limited to retaining partial results, the Windows timeout, and keeping config discovery within its connection budget; it does not redesign PATH lookup or add background refresh machinery.

## Validation

- Windows launcher suite: 15 passed; 15 platform-specific tests skipped.
- Targeted config tests: 4 passed, including concurrent discovery with a ten-second editor scan and a stalled remote-target probe.
- Regression coverage uses the test clock: partial results, immediate retry after timeout, completion after five but before ten seconds, stalled first probes on Windows/macOS/Linux, cache reuse, and interruption.
- Server `tsc --noEmit` passed.
- Targeted lint passed with existing warnings in unrelated portions of `server.test.ts`; `git diff --check` passed.
- Read-only discovery against the actual Windows installation returned Cursor, VS Code, Zed, and File Explorer in 2.2 seconds. No installed-app changes, build, or browser verification.

## Checklist

- [x] Small, focused backend fix
- [x] Explained the failure and the timeout tradeoff
- [x] Focused regression tests and server typecheck

Made using GPT-6 Astra on high reasoning, through Codex.

## Summary by CodeRabbit

* **Bug Fixes**
  * Editor discovery now stops waiting after a defined timeout, preventing stalled scans from blocking server startup.
  * Timed-out editor scans are not cached, allowing subsequent connection attempts to retry and find newly available editors.
  * File-manager and remote-target discovery now complete within the expected startup time, even when a provider is unresponsive.

* **Performance**
  * Editor, file-manager, and remote-target discovery now run concurrently for faster configuration loading.


MajesteitBart:
﻿Thanks for addressing this. I reviewed `d27eff9` while investigating #4697 on an affected Windows desktop. Preserving partial results and running config discovery concurrently are useful improvements. Two cases still appear to leave installed Open targets missing, including File Explorer:

1. **An earlier stalled probe can prevent Explorer from being checked.** The scan is still sequential, and `file-manager` is last in `EDITORS`. Extending the Windows budget to ten seconds helps slow scans, but a probe that stalls until that deadline can still leave later targets undiscovered. If nothing has been found yet, the result is still `[]`. The new stalled-first-probe tests explicitly cover and accept that empty result.

2. **An incomplete refresh can replace a previously complete list.** After the 60-second cache expires, the scan starts with an empty accumulator. On timeout, `cachedAvailableEditors` leaves the old cache entry untouched but returns the new partial or empty `editors` list. Keeping the old entry in memory therefore does not preserve those targets in the config response.

Could this PR cover those cases too? The behavior I would look for is:

- A stalled editor probe does not prevent other healthy targets, including Explorer, from being discovered.
- A previously detected target stays available when its new probe times out, but is removed when a completed probe confirms it is missing.
- Incomplete discovery remains eligible for retry, as this PR already allows.

Two useful regression scenarios would be an early stalled editor probe with a resolvable file manager, and a successful scan followed by cache expiry and a stalled refresh. The latter should also check that a later completed scan can remove a genuinely missing target.

This is based on code and test review of this PR; I have not run this branch in the desktop app. I reproduced the equivalent gaps with a focused module test against #10694. The affected user's released Windows app lost the entire Open menu, including Explorer, so retaining only targets reached before the deadline does not fully cover the reported behavior.

Reviewed and written with Codex (GPT-6), at the affected user's request.

PR #12425, opened 2026-09-18
feat(server): let t3.json skip worktree submodule init

BrinkhaT:
## Problem

After #7674, creating a worktree always runs `git submodule update --init --recursive`. That is the right default when a repo keeps skills or shared tooling in a submodule, but it is expensive for repositories with many submodules that only need a subset in each worktree. There is no project-level way to skip the built-in step.

## Change

Add optional `initSubmodulesOnWorktreeCreate` to `t3.json`.

- omitted or `true`: keep today's behavior
- `false`: skip the automatic submodule checkout in `createWorktree`

Repos that want a subset can initialize it from a `runOnWorktreeCreate` action. Set that action's `async` field to `false` if the agent should wait until those submodules exist.

The flag is read from the new worktree's `t3.json`, so thread worktrees, PR worktrees, and the raw RPC all honor it.

## Tests

- schema decode and published JSON schema keys
- worktrees still check out submodules by default
- `t3.json` with `false` leaves the submodule empty

Model: Grok 4.6 via T3 Code / Grok Build.

## Summary by CodeRabbit

- **New Features**
  - Added the `initSubmodulesOnWorktreeCreate` project setting to control whether submodules are initialized automatically when creating a worktree.
  - Submodule initialization remains enabled by default; setting the option to `false` skips the automatic checkout.

- **Documentation**
  - Documented the new setting and how to initialize submodules through worktree creation actions when automatic initialization is disabled.


juliusmarminge:
Superseded by [#12953](https://github.com/pingdotgg/t3code/pull/12953) (with [#12954](https://github.com/pingdotgg/t3code/pull/12954) / [#12955](https://github.com/pingdotgg/t3code/pull/12955)), which landed `worktreeSubmodules` (`recursive` / `top-level` / `none`) on `main` — the same skip-submodule-init goal, with top-level and settings UI on top. Closing as leftover hygiene.

PR #7882, opened 2026-08-22
fix(provider): resolve skills and slash commands against the workspace root

CurlyPeter:
## What changed

Both composer pickers are empty or incomplete in a packaged desktop build. This resolves skills and slash commands against the **workspace root** instead of the server's single global cwd.

## Why

`DesktopEnvironment.ts` sets the backend cwd to the user's home directory when packaged:

```ts
backendCwd: input.isPackaged ? homeDirectory : appRoot,
```

`ServerProvider.skills` and `ServerProvider.slashCommands` are produced once per provider instance against that one cwd — but both are **project-scoped**:

- `ClaudeSkills` scans `<configDir>/skills` and `<cwd>/.claude/skills`. With cwd pinned to `$HOME`, the "project" root resolves to `~/.claude/skills` for *every* project, so a user who keeps no skills there sees none at all and `$` reports "No skills found". It reports the mis-scoped scan accurately; it is not a UI bug.
- the CLI's init handshake reports project-scoped slash commands, so `/` lists only the built-ins. Measured on one real workspace: the CLI reports **140** commands from the project root versus **48** from a neutral directory.

Running sessions are unaffected — they get the right scope through the CLI's own settings resolution — so this is a discovery gap, not a loss of capability.

## Approach

A provider is a machine-level installation in this model, and version/auth/models genuinely are per-machine. So rather than making the snapshot project-aware, this adds a workspace-scoped surface **beside** it:

- `ProviderInstance` gains optional `discoverSkillsForCwd` / `discoverSlashCommandsForCwd`. Optional, so the other four drivers are untouched.
- `ProviderRegistry` gains matching `*ForInstance` methods, following the existing `getProviderMaintenanceCapabilitiesForInstance` delegation pattern, resolving empty when a driver cannot enumerate per directory.
- new `providers.workspaceSkills` RPC resolves `thread.worktreePath ?? project.workspaceRoot` — the same idiom `assets.createUrl` already uses — mapped to `orchestration:read`.
- the composer prefers the workspace result for `$` and `/`, **falling back to the snapshot**, so an older server behaves exactly as today.

It is keyed on the **project** with an optional thread, because the composer is frequently on a draft thread with no persisted row, and that is precisely when the pickers matter.

## The cache

`ClaudeDriver`'s capabilities cache is now keyed by cwd (capacity 16) rather than holding a single entry probed against the server's cwd.

[...]

PR #12426, opened 2026-09-18
fix(server): discover Claude commands in the active workspace

### LINKED ISSUE #12214: [Bug]: Project `.claude/commands/*.md` never appear in the `/` menu for Claude; the per-workspace refresh only scans `.claude/skills`
### Area

`apps/server`

### Steps to reproduce

1. Use the Claude provider on the desktop app with a local project that has a command file at `<project>/.claude/commands/start-session.md`.
2. Open a thread in that project.
3. Type `/start`.

### Expected behavior

The `/` menu lists `/start-session`, the same as the Claude Code TUI when launched in `<project>`.

### Actual behavior

The menu shows user and plugin commands plus the project's `.claude/skills/*` entries, but no project `.claude/commands/*` entries. Claude Code does load them: the agent in the same thread lists `start-session` among its available commands.

Mechanism, from reading `server.asar` (`apps/server/dist/bin.mjs`) in 0.0.42:
- `probeClaudeCapabilities` runs with `cwd` from `ServerConfig`, which is the server's own working directory. `slashCommands` comes from that probe's `init.commands`, so it never includes a project's commands.
- `snapshotForCwd` for the Claude driver returns `{ ...machineSnapshot, skills }`. `discoverClaudeSkills(config, cwd)` reads only `<configDir>/skills` and `<cwd>/.claude/skills`. `slashCommands` stays the machine-wide list, and `<cwd>/.claude/commands` is never read.
- `persistProvider` strips `workspaceSnapshots`, so `~/.t3/caches/claudeAgent.json` holds only the machine list (253 commands here, none of them project commands).

Suggested fix: have the Claude `snapshotForCwd` also scan `<cwd>/.claude/commands/**/*.md` (subfolders become `dir:name`) and merge the results into `slashCommands`. Or run the capability probe with the workspace cwd.

### Impact

Commands defined in a project can't be found from the menu. Users with a team or workspace command set have to know each name and type it in full.

### Version or commit

T3 Code desktop 0.0.42, Claude Code 2.1.273

### Environment

Windows 11, local connection

### Workaround

Type the full command name without autocomplete. The command should still reach Claude Code, but I have not confirmed this on this build.

yashranaway:
Claude workspace snapshots discover project skills but keep the server-wide slash-command list. Commands loaded by the CLI from the project therefore remain absent from the composer, particularly in packaged desktop installs whose server starts in the home directory.

Read the CLI initialization commands in the requested workspace through the existing snapshot path. Preserve the machine account and model metadata, retain the built-in compact command, and return a failed snapshot when initialization fails so the registry does not cache an incomplete menu. The registry already deduplicates and bounds workspace snapshots.

Fixes #12214. Related #7882 addresses this through a broader discovery API; this change uses the workspace snapshot API already on main.

Validation: 24 focused capability-probe and skill-discovery tests pass, including isolated command lists for two workspaces and recovery after a failed probe. Server typecheck and targeted lint pass. Tests use the SDK boundary without model calls.

Model: GPT-6
Harness: Codex in T3 Code

## Summary by CodeRabbit

- **New Features**
  - Claude capabilities and skills are now tailored to the active workspace.
  - Workspace-specific slash commands remain separate from machine-level commands.
  - Standard compacting support remains available alongside discovered workspace commands.

- **Bug Fixes**
  - Failed capability checks no longer permanently replace available command menus.
  - Subsequent checks can recover and return to a ready state after a temporary failure.


juliusmarminge:

yashranaway:

PR #12427, opened 2026-09-18
fix(server): retry failed OpenCode workspace command discovery

yashranaway:
A failed or timed-out OpenCode workspace command request is converted into an empty command list. The registry then caches that successful snapshot, leaving commands absent until the workspace cache is cleared.

Propagate command discovery failures through the existing driver error path so the registry leaves the workspace eligible for retry. A driver regression test exercises an HTTP 503 followed by a successful response and verifies that workspace commands return.

Validation: focused driver test, server typecheck, and targeted lint pass. No provider process or model call is needed.

Model: GPT-6
Harness: Codex in T3 Code

## Summary by CodeRabbit

- **Bug Fixes**
  - Workspace discovery now reports an error when command loading times out, instead of silently presenting an empty command list.
  - Command discovery can recover from temporary service failures and successfully identify available workspace commands once the service responds again.


juliusmarminge:

yashranaway:

PR #12439, opened 2026-09-18
fix: detect installed editors outside PATH

Bil0000:
Installed editors were missing from Open in when their launchers were not on PATH.

Share discovery between server launches and the desktop remote-editor probe. Check standard macOS app bundles, Windows user/system installs, and Linux user/system and Toolbox launcher folders after PATH. Preserve file positions, Cursor IDE mode, Windows escaping, and existing file-manager behavior.

Validation: 77 focused tests pass; targeted lint, formatting, and server/desktop/shared typechecks pass. Platform tests use temporary install fixtures and mocked spawning on Linux. Native macOS/Windows GUI runs were not available; one native Windows PowerShell test is skipped. Custom install locations still require PATH.

Implemented with GPT-6 in the native Codex harness.

## Summary by CodeRabbit

- **New Features**
  - Improved detection of supported code editors installed outside the system PATH.
  - Added broader discovery across macOS, Windows, and Linux installation locations, including JetBrains Toolbox and editor-specific app locations.
  - Improved launching of discovered editors with platform-specific executables and launch arguments.
  - PATH-based editor launchers are prioritized when available.

- **Bug Fixes**
  - Unusable editor installations are now ignored during discovery.
  - Improved handling of Windows command-line launchers.

### Before and after

Same install fixtures and PATH. The unchanged production menu receives discovery results from before the fix (`52e4b44293`) and after (`1e3291b2ef`). VS Code is on PATH; Cursor, Zed and WebStorm are installed in macOS app bundles outside PATH.

| Before | After |
| --- | --- |
| Before: only VS Code and Finder detected | After: Cursor, Zed and WebStorm are detected too |

These are component captures using executable install fixtures on Linux with the macOS platform injected, not native macOS GUI tests.


Bil0000:
The Windows Electron launcher finding is fixed, with a regression test for Cursor under resources/app/bin. The test failed before the fix; all 77 focused tests now pass.

The advisory docstring-coverage warning is left unchanged because the task instructions explicitly prohibit adding comments or docstrings. The resolver has behavior tests, and no documented API contract changed.

PR #12454, opened 2026-09-18
feat(client): keep several addresses for one paired environment

rene-roid:
A machine that is reached over the LAN at home and over a tailnet elsewhere needed its saved pairing edited every time the client moved networks. This keeps every address on one saved environment and picks the working one automatically.

## What changed

- A bearer profile now stores `alternateHttpBaseUrls` and `pinnedRoute` next to its preferred URL. Both keys are optional, so existing saved catalogs decode unchanged.
- Pairing an already saved environment from a new address keeps the previous address as an alternate instead of replacing it. This works from the existing **Add environment** flow on web, desktop, and mobile.
- On connect, the resolver fetches the environment descriptor over every address at once and dials the first one that answers with the expected environment ID. If none answers it falls back to the preferred address so the usual error surfaces. A pinned profile skips the race.
- Web and desktop: saved environments get an **Edit…** action for the label, preferred URL, other URLs, and the pin. Mobile: the environment row gains the same two fields. Both rows show the address actually in use while connected.
- User docs: a short section under remote access.

## Why

Advertised endpoints are only reachability hints and the connecting device is the only one that can prove a route works, so the choice has to happen on the client at connect time. Racing the descriptor fetch reuses the identity check the resolver already performs and needs no server change.

## Not covered

- Live route switching while connected. An established socket stays on its route; the next connect attempt races again.
- Relay and SSH environments, which own their own route.

## Verification

- `vp test run` on `resolver.test.ts`, `onboarding.test.ts`, and `storageDocument.test.ts` in `packages/client-runtime`: 37 passed. New tests cover the route race, the all-fail fallback, the pin, route normalization on edit, and route merging on re-pair.
- Typecheck for `packages/client-runtime`, `apps/web`, and `apps/mobile`.
- Lint and format on the touched files.
- Before/after screenshots of the edit dialog and mobile row are not attached yet.

Model: Claude Fable 5.1. Harness: Claude Code in T3 Code.

## Summary by CodeRabbit

* **New Features**
  * Saved connections can now include multiple alternative URLs and an option to always use the preferred URL.
  * Connections automatically select the first available route and preserve saved routes when re-pairing.
  * Web and mobile users can edit saved connection labels, preferred URLs, alternate URLs, and route preferences.
  * Web settings now show an Edit action for eligible saved connections and display the active endpoint when routes differ.

* **Documentation**
  * Added guidance for configuring multiple LAN or tailnet URLs.


rene-roid:
Opened prematurely; closing while the change is reviewed locally.

PR #12458, opened 2026-09-18
feat: continue a stopped thread when its usage limit resets

AryaBuddha:
## What Changed

When a turn stops on a provider usage limit, the thread error banner now offers **Continue when limit resets**. Clicking it schedules a server-side continuation: once the limit's reset instant passes, the server starts a turn by sending "Continue" to the agent — with every client closed, and across server restarts. The pending continuation is visible in the same banner (web) / above the composer (mobile) and cancellable there; any manual message, settle, or archive also clears it.

How it works, end to end:

- **Runtime contract**: failed `turn.completed` events gain an optional `usageLimitResetsAt`. Codex sets it from the exhausted rate-limit window it already names in the error sentence; Claude tracks the reset instants of the windows that rejected the turn (`rate_limit_event`) and stamps the latest one. Grok/Cursor/OpenCode/Antigravity report no reset instant at failure, so they don't offer the button.
- **Session read model**: ingestion carries the instant into `session.lastErrorLimitResetsAt` beside `lastError`, clearing both together.
- **Schedule**: new `thread.auto-continue.set` / `.clear` client commands persist `autoContinueAt` on the thread (event-sourced, mirroring snooze, with a `projection_threads` column + migration). The decider enforces a future-time invariant, idempotent re-sets, and clears the schedule on any turn start, settle, or archive.
- **Firing**: a new `AutoContinueReactor` (same periodic-sweep pattern as `ThreadSettlementReactor`) dispatches an internal `thread.auto-continue.fire` for due threads; the decider re-validates the schedule compare-and-set style at fire time, so cancels/reschedules that race a sweep win, then expands into an ordinary `thread.turn.start`.
- **Version skew**: a `threadAutoContinue` capability gates the UI; clients never send the commands to older servers.

## Why

Subscription limits (Claude's 5-hour window, Codex weekly) regularly stop long-running work mid-task, and both providers literally tell the user to "send the message again once the limit resets". Today that means checking back hours later. The reset instant was already parsed structurally for the usage bars — this threads it through the turn-failure path and lets the server do the waiting. The schedule is persisted thread state rather than a timer so it works remote-first and survives restarts.

## UI Changes

Web: the existing thread error banner gains a "Continue when limit resets · &lt;time&gt;" button, which becomes "Continuing when the limit resets · &lt;time&gt; · Cancel" once scheduled (still visible after the error text is dismissed, so the pending turn is never invisible). Mobile: an equivalent card above the composer. Screenshots require provoking a real provider limit stop; I can add them on request.

## Checklist

- [x] This PR is small and focused (one concern: limit-reset continuation, across the surfaces it must touch)

Verified with focused tests: decider set/clear/fire invariants, the sweep reactor, ingestion plumbing of the reset instant, and the Codex reset-instant helper, plus targeted typecheck/lint on every touched package.

Built by Claude Fable 5 via Cursor.

## Summary by CodeRabbit

* **New Features**
  * Added automatic thread continuation when supported usage limits reset.
  * Added banners on web and mobile to schedule, view, and cancel pending continuations.
  * Continuations run automatically even when clients are closed and persist across server restarts.
  * Added support for provider-reported reset times from Claude and Codex.
  * Scheduled continuations are cleared when users send messages, settle threads, or archive them.

* **Documentation**
  * Documented scheduled continuation behavior and provider availability.


juliusmarminge:

adrianoresende:
Nice! This feature is highly valuable for users who do not choose the maximum plan.

PR #12481, opened 2026-09-18
fix(web): show Auto balance in the workspace host menu

Bil0000:
The Workspace card omitted Auto balance, even when it was selected below the composer.

Pass the existing label and action to the shared host selector. Both card layouts now match the composer. The same draft, lock, host-count, and settings checks apply. Nine lines added in two files; routing stays unchanged.

Verified: 75 focused tests, web typecheck, targeted lint, formatting, and browser checks at 1800px and 1000px against two isolated hosts. Keyboard host selection and mouse Auto balance selection keep both labels in sync. Removing the fix made the browser check fail. Targeted lint has existing warnings and no errors.

Before:

Workspace host menu before

After:

Workspace host menu with Auto balance

Built with GPT-6 in Codex.


Bil0000:

PR #12491, opened 2026-09-18
fix(web): show subagent models and running count in Lineage

Bil0000:
V2 sub-agent tooltips showed "Unknown" when the sub-agent record omitted its model, even when the child thread had a model selection.

Use the child thread as a fallback and resolve model aliases with the existing composer name helpers. Both Lineage and timeline tooltips open after 200 ms instead of the default 600 ms. The Lineage title shows the running sub-agent count across all pages, includes agents awaiting a child thread, and hides the count at zero.

If neither record provides a model, the tooltip says "Not reported" instead of guessing. This changes the shared web/desktop UI; mobile has no hover tooltip.

Verified with 47 focused tests, web typecheck, and lint. Reproduced the bug and checked the fix at 1280 px and 820 px in an isolated app; measured the tooltip opening at 226 ms.

Rebased onto the current `t3code/codex-turn-mapping` tip after the base branch was force-pushed. The previously noted `Test Server 3` failure in the unchanged `queued_cancelled_while_active/codex` replay scenario is resolved by `ea483d1b9` ("fix(server): unblock orchestrator CI checks") on the new base.

Before:
Unknown model before the fix

After:
Model name and running count after the fix

Hover timing:
[Hover timing](https://github.com/user-attachments/assets/ff65699e-2a9f-4852-8e69-3a273837d430)

Model: GPT-6. Harness: Codex.


Bil0000:

PR #12493, opened 2026-09-18
feat(server): export log records over OTLP

yordis:
## What Changed

Log records can now be exported over OTLP, configured the way traces and metrics already are: `T3CODE_OTLP_LOGS_URL`, the desktop bootstrap envelope, or Settings, in that order of precedence. Logs share `T3CODE_OTLP_HEADERS`, `T3CODE_OTLP_PROTOCOL`, and the export interval with the other two signals, and the resource each exporter tags its signal with is now one `otlpResource` helper instead of a literal repeated per signal.

With no log endpoint configured, nothing changes. The logger set is what it was, and stdout output and SSH-managed log persistence are untouched either way. When one is configured, the OTLP logger takes the place of `Logger.tracerLogger` rather than joining it, so an in-span message leaves once as a log record instead of twice.

## Why

Traces and metrics could reach a collector and logs could not, which left the signal people reach for first as the one the server kept to itself. Debugging a remote environment meant having spans that showed a turn took too long and no way to read what the server said while it happened, short of asking someone to find a file and paste it.

Partial coverage is also the confusing kind of gap. Two signals arriving looks like all of them arrived, so the missing third is invisible from the dashboard rather than obviously absent.

The placement in `serverLogger.ts` is the part worth reviewing. `Logger.layer` writes the whole logger set rather than adding to it, so the OTLP log exporter has to be declared beside the console logger instead of sitting in the observability layer with its sibling signals. A second layer that installs a logger independently silently drops whichever set loses the merge, which is why this one is not there.

That same call is where the overlap with `Logger.tracerLogger` gets settled. The tracer logger only emits for messages logged inside a recorded span, and files them under traces rather than logs, so it covers less than the log records do. Recording events on spans is also the shape OpenTelemetry is deprecating, in favor of log-based events correlated to the current span: https://opentelemetry.io/blog/2026/deprecating-span-events/. Keeping both would export every in-span message twice, so the log signal replaces it once an endpoint is set. The tradeoff is that the local trace file stops carrying log messages in that mode.

## Checklist

- [x] I included before/after screenshots for any UI changes (not applicable, no UI change)

## Summary by CodeRabbit

* **New Features**
  * Added OTLP log export alongside metrics and trace export.
  * Configure log export with `T3CODE_OTLP_LOGS_URL`.
  * Environment, bootstrap, and persisted settings now resolve observability endpoints consistently.
  * Server status shows whether log export is enabled.
  * Exported logs include trace context and avoid duplicate trace logging.
  * Logs continue appearing in stdout and SSH-persisted output.
  * Observability settings are trimmed and persisted consistently.

* **Documentation**
  * Updated observability guidance with log configuration, examples, and fallback behavior.

* **Compatibility**
  * Older server configurations decode with log export disabled.


yordis:
Following up on the heads up that most server logs already reach a collector as span events, since it changed what this PR does.

That is accurate, and to avoid sending the same message twice I pushed a commit so the two paths are mutually exclusive: when a logs endpoint is configured the server installs the OTLP logger in place of `Logger.tracerLogger`, and with no endpoint set the logger set is exactly what it is on `main` today.

Two reasons the log signal still earns its place. Coverage: `Logger.tracerLogger` returns early when the fiber has no recorded internal span, so the orchestration reactors, the WS layer, the provider adapters, the terminal manager, and most of startup and shutdown emit nothing at all right now. In `apps/server/src` that is 306 log call sites against 51 `withSpan` sites, with the spans concentrated in auth, checkpoint diffing, and the RPC wrapper. Those uncovered paths are stdout only, which is the one place you cannot reach on a remote environment. The diagnostics RPC methods make it sharper, since they set `TracerEnabled: false` deliberately, so the handlers you would reach for to debug a sick server drop their logs from telemetry too.

Destination: a span event lands in the trace store, not the log store, so it is not searchable by severity (`effect.logLevel` is a span attribute string rather than OTLP `severityNumber`), log-based alerting never fires on it, and log retention does not apply. `OtlpLogger` stamps `traceId` and `spanId` on every record, so correlation with the trace survives in the log signal.

This is also the direction OpenTelemetry is heading. They are deprecating the span event recording API in favor of log-based events correlated to the current span: https://opentelemetry.io/blog/2026/deprecating-span-events/. Nothing urgent, and the post is explicit that existing span event data stays part of the supported OTLP trace model, but it does mean the choice is ours in our own logger list rather than something Effect has to change first.

The tradeoff worth flagging: with log export configured the local NDJSON trace file stops carrying log messages, which changes what trace diagnostics shows for those users. If you would rather keep both locally, the alternative is to keep `Logger.tracerLogger` installed and skip only the forward to the OTLP delegate in `LocalFileSpan.event`. Happy to go that way instead, it is just a larger change and it leans on Effect's internal attribute naming.

PR #12494, opened 2026-09-18
fix(server): backfill login shell for consistent agent env

JoeyEamigh:
## What Changed

backfill the login-shell environment at startup instead of just `PATH`

## Why

depending on how the t3code environment was started, agents end up with different variables and functions in their shell. this can break things like `nvm`, `fnm`, custom shell functions, env vars you expect agents to have, etc. this shows up worst when i am using t3code over ssh to a main development box that i also use myself. i expect the agent to have access to shell and env that i set up for myself and work in claude code, codex, etc, but its missing in t3code.

<details>
<summary>claude's reasoning</summary>

## What Changed

- `apps/server` now hydrates the whole login-shell environment at startup instead of `PATH` alone
- add `readFullEnvironmentFromLoginShell` to `@t3tools/shared/shell`, alongside the existing
  name-list reader, using the same `-ilc` probe with a NUL-delimited `env -0` capture
- hydration is backfill-only: a variable the process already has is never overwritten
- `PATH` keeps its current `mergePathEntries` + launchctl behaviour, unchanged
- drop `readPathFromLoginShell`, which has no remaining callers

## Why

On the same machine, as the same user, agents get a different environment depending on how the
server was started.

`fixPath()` recovers `PATH` and nothing else (`apps/server/src/os-jank.ts`). That is enough to
*find* an agent binary, which is what it was written for, but not enough to run one: anything the
user exports from their shell rc is evaluated by that login shell and then discarded.

It shows up worst on the remote path. `packages/ssh/src/tunnel.ts` launches the remote server with

    nohup env T3CODE_NO_BROWSER=1 "$RUNNER_FILE" serve --host 127.0.0.1 --port "$REMOTE_PORT" ...

No login shell is involved, so the daemon holds only what sshd's non-interactive shell produced,
and every agent it spawns inherits that. On my machine the daemon has 110 variables where the
login shell has 161. Missing: `EDITOR`, `NVM_DIR`, `NVM_BIN`, `GVM_ROOT`, `RBENV_SHELL`, and the
rest of the version-manager state, because those are configured below the interactive guard in the
shell rc. `19c214645` describes that exact failure mode for the WSL backend and fixes it there.

The desktop side already solved its half of this: `DesktopShellEnvironment` captures 17 names.
The server never received the equivalent work — #972 says so directly, "the server-side startup
path is left unchanged so this PR stays desktop-scoped." This is that deferred half.

## Why backfill rather than overwrite

An inherited value is the one the service was deliberately launched with, and for session-scoped
handles it is the correct one. #972 established this for `SSH_AUTH_SOCK` (a Terminal-launched
session must keep its own socket rather than take the login shell's), and the locale group in
`DesktopShellEnvironment` follows the same principle.

Backfill also means this composes with the desktop rather than fighting it. In a desktop install
the server inherits Electron's already-hydrated environment and only fills what is still missing,
so every precedence decision in `DesktopShellEnvironment` survives untouched.

Six names stay runtime-owned (`HOME`, `PATH`, `PWD`, `OLDPWD`, `SHLVL`, `_`) plus `T3_*`/`T3CODE_*`.
The login shell inherits this process's environment, so for those it reports its own values rather
than the user's configuration.

## Scope

Server only, on purpose. `DesktopShellEnvironment` is untouched: it hydrates the Electron main
process for Electron's own needs, and agents are spawned by the server, so this reaches both
deployments without changing desktop behaviour.

## Testing

- `vp run --filter @t3tools/shared --filter t3 test` — 5088 passed
- typecheck, lint, `fmt --check`, `knip:check` all clean
- new server tests mirror the desktop contract, including the `SSH_AUTH_SOCK` preservation case
  from `DesktopShellEnvironment.test.ts`

`env -0` is used for the capture. It is supported by Apple's BSD `env`; verified on macOS 26.3,
and the app's floor is `minimumSystemVersion: 13.0`.
</details>

## UI Changes

None.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - POSIX environments now import a more complete set of variables from login shells.
  - Existing environment values are preserved, while missing or empty values can be populated automatically.
  - Runtime-managed variables remain protected from unintended changes.
  - Shell output is parsed more reliably, including values containing newlines, equals signs, and marker text.
  - Environment detection now handles shell lookup failures gracefully and retains the inherited environment.


JoeyEamigh:
one thing to note on the macroscope feedback here: while it is technically correct that it changes behavior, it actually brings it in line with `npx t3`. so any argument against this is also an argument against allowing `npx t3` to continue inheriting the env vars of the parent shell.

PR #12504, opened 2026-09-18
feat(cli): add t3 thread start and send to drive threads from scripts

kakaponick:
Now we can orchestrate the orchestrations 😅

## What Changed

Adds two CLI commands that drive threads on the running T3 Code server:

    t3 thread start <project> <prompt | -> [--model <slug>] [--provider <instance>] [--title <title>] [--json]
    t3 thread send <thread> <prompt | -> [--json]

`thread start`
- Project: id or path, matched the same way as `t3 app`. Prompt: argument, or `-` for stdin.
- Default model and runtime mode come from `resolveProjectSettings`, same as a new thread in the UI.
- `--model` finds its provider in the provider status cache. Only enabled, installed instances count, and aliases match. `--provider` on its own picks that provider's default model.
- Title: `--title`, otherwise the start of the prompt, seeded for title generation the way the composer does it.
- If the turn fails to start, the new thread is deleted.

`thread send`
- Starts the next turn with the thread's own model, runtime mode, and interaction mode.
- If the agent is still working, it waits for the turn to finish instead of steering. On some providers a steer interrupts the running agent.
- Refuses threads that are waiting on an approval or an answer.

Both call the existing `/api/orchestration/shell` and `/api/orchestration/dispatch` endpoints with a session limited to orchestration read and operate scopes, revoked on exit. This is the same pattern `t3 project` uses. `--json` prints ids for scripts. No server, contract, or client changes.

Limitation: new threads start in the project checkout. HTTP dispatch has no worktree bootstrap, so a project that defaults to worktrees still gets a local thread.

## Why

Agents and scripts can only hand work to T3 Code through the UI. With this, an orchestrator agent can start a thread and send follow-ups itself:

    id=$(t3 thread start ~/repo - --model claude-opus-5 --json < task.md | jq -r .threadId)
    t3 thread send "$id" "Now add tests"

## Checklist

Tested in `apps/server/src/bin.test.ts` against a live HTTP server (model resolution, settings defaults, stdin, send keeping thread modes, pending-approval and unknown-thread errors) and in `apps/server/src/cli/thread.test.ts` (waiting for a running turn, on a test clock). Verified manually against a running desktop app on Windows with Claude Opus 5: `send` right after `start` waited for a 15-second command to finish, then continued the thread.

Implemented with Claude Opus 5 in Claude Code.

## Summary by CodeRabbit

* **New Features**
  * Added `thread start` to create threads and begin prompts from the command line.
  * Added `thread send` to continue existing threads, including prompts provided through standard input.
  * Supports project and provider model selection, default models, and text or JSON output.
  * Waits for active turns to finish before sending new messages, with an option to send immediately.
  * Clearly reports unavailable servers, projects, threads, models, empty prompts, and pending approvals.


gshaw:
I would love to have this capability.

PR #12529, opened 2026-09-19
fix(server): scheduler writes no longer overwrite concurrent task edits

saphid:
Stacked on #11592. Until that merges, this diff also shows its commit; review only `fix(server): scheduler writes no longer overwrite concurrent task edits`.

## Problem

Run completion, stuck-run release, and dispatch each re-read the task and then wrote in a separate statement. An edit committed between the two had its `next_run_at` overwritten from the older schedule, and a pause, postpone, or delete landing after the dispatch re-read could still fire.

## Fix

Each re-read and its write now share one transaction, with the contention retry from #11592. The missed-run reschedule is guarded on the `next_run_at` it was computed from, so it loses to an edit instead of overwriting it.

## Tests

`ScheduledTaskService.test.ts`: the due-read to dispatch gap (pause, postpone, delete, and the no-edit case), an edit committed mid-run on both the success and interrupted paths, and a contended completion write that retries instead of stranding the task as running. The retry test fails with the retry disabled. Server typecheck: 0 errors.

Split from the original #11592 by Claude Fable 5.1 via Claude Code; original implementation with SWE-2 High via Devin CLI. One complete independent review by SWE-2 Max; its findings were fixed, and the fixes were then reviewed by Claude Fable 5.1 only.


juliusmarminge:
Closing. The current head of #11592 (f19067f3fc) re-absorbed this split, so the scheduler-writer transactions, `retryContended`, and the "pause committed after the due read" / "contended completion write retries" tests all live there now. I'd rather review #11592 as one PR than the same change in two shapes. If you would prefer to keep the split, drop the content from #11592 first and re-stack this on it.

Issue #5623, opened 2026-08-07
T3 Connect: environment display name defaults to hostname — how to distinguish multiple environments on one machine?


yashaka:
**Setup:** one machine running the desktop app plus three isolated web instances (`npx t3@nightly start --base-dir ~/.t3-<slot> --port 3151x --no-browser`, one per account slot). All four are linked to T3 Connect (`t3 connect link --base-dir ...`) and show up in the mobile app — which is exactly the multi-instance use case T3 Connect shines at.

**Problem:** all four environments display as the same name (the machine hostname), making the list indistinguishable. I found no rename affordance in the mobile app, and no `--name`/label flag on `t3 connect link` or `t3 start` (swept the `--help` surfaces on `0.0.33-nightly.20260807`).

**Question:** is there an existing way to set a per-environment display name (env var / config key I missed)? If not — consider this a feature request: e.g. `t3 connect link --name <label>`, or defaulting to something derived from `--base-dir` when it isn't the default dir. With per-slot base-dirs the stem (`t3-automician`, `t3-claudex`, …) would already disambiguate nicely.

Artem-B:
I think there are two parts that need to be addressed.

First is being able to specify environment name explicitly when we start t3 server. Second is being able to assign a meaningful name to existing instances. Those could be different entities -- immutable instance ID, and user-friendly user editable display name.

PR #12559, opened 2026-09-19
feat(server): allow environment label override

osmocomet:
## What Changed

Adds an optional `T3CODE_ENVIRONMENT_LABEL` environment variable that overrides the environment name shown to connected clients.

The value is trimmed before use. Missing, empty, or whitespace-only values preserve the existing automatic label detection.

This only changes startup-time server configuration.

## Why

Multiple T3 environments running on the same machine can resolve to the same host-derived label, making them difficult to distinguish in connected clients.

This provides a small declarative override for environments where the server process is configured externally.

This is intentionally narrower than #7267, which tackles the broader interactive renaming experience across clients and persistence.

This PR only covers the declarative startup-time case. I hope it can be useful as a small complementary option for headless or externally managed environments.

Related: #5623

## Verification

- `ServerEnvironmentLabel.test.ts`: 10/10 passed
- Targeted lint passed
- Server typecheck passed
- `git diff --check` passed
- Built server bundle manually verified with `T3CODE_ENVIRONMENT_LABEL="Development VM"` and `/.well-known/t3/environment`

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Server administrators can set a custom environment label displayed to connected clients.
  - Labels are trimmed before display; empty or whitespace-only values continue using the automatically detected label.

- **Documentation**
  - Added installation documentation for configuring the custom environment label.


christhomas:
Another use case for this, from a slightly different angle: multiple Claude Code accounts on a single server.

I run two `t3 serve` instances on one machine, each with its own `T3CODE_HOME` and its own `CLAUDE_CONFIG_DIR`, so each environment is signed into a different Claude account. It works well — but both are derived from the same hostname, so they appear identically in the clients and there is no way to tell which one I am about to open.

`T3CODE_ENVIRONMENT_LABEL` as proposed here would solve that cleanly for this setup. Both instances are systemd units, so the label becomes one more `Environment=` line: declarative, no UI step, and it survives a rebuild. That is a real advantage over a UI-only rename for anyone running headless.

Two things I would add:

**1. Icon override too.** `environmentIcon` is currently only settable from inside the app, and both of mine defaulted to `server` — so the name *and* the icon are identical. If the label is overridable at startup, the icon should be as well. Otherwise a headless setup still has nothing to differentiate on.

**2. First-class Claude profiles.** Today, getting two Claude accounts onto one machine means hand-rolling it in the terminal: a separate `CLAUDE_CONFIG_DIR` per service, plus a script to symlink the shared parts (`projects`, `sessions`, `session-env`, `plugins`, `settings.json`) back to a single backend so that only credentials and per-account state stay separate. That pattern works extremely well, and it would be far more powerful as a feature than as something everyone reinvents.

Concretely, it would be useful to:

- create and manage multiple Claude profiles from the app rather than the terminal
- choose, per profile, whether sessions and workspaces are **shared** across profiles (one backend, different credentials) or kept as **fully separate** accounts

The shared-backend-with-separate-credentials case is the interesting one. Plenty of people run more than one Claude account simply for more usage headroom, and in that situation you do not want two disconnected worlds — you want one set of projects and sessions, with the ability to switch which credential is doing the work.

Happy to test this branch against a two-environment headless setup if that would help.

PR #12564, opened 2026-09-19
feat(mobile): remove projects from the thread list and project settings

bawisniewski:
## Problem

Mobile has no way to remove a project once it has been added. The only delete path is the clone banner on the new-task screen, which disappears as soon as the clone finishes. A mis-added project (for example a home directory picked by accident) stays in the list for good, and the only fix is to open the web or desktop app. Web offers removal in the sidebar project menu and in Settings > Project.

## Fix

One confirm-and-remove hook (`useConfirmRemoveProjects`) shared by two entry points:

- **Project picker** ("Choose project" sheet, the one place mobile lists every project): swipe a row left on iOS for **Remove**, matching thread rows; long-press it on Android for a **Remove project** menu. A tap that lands during or after a swipe only closes the row, so a swipe never falls through into selecting the project.
- **Settings > Project overview**: a **Remove project** section, plus a per-checkout **Remove** button when the grouped project spans several environments (mirrors web's project settings). Removing the whole project clears the project scope and pops back to the settings index.

The confirmation copy mirrors the web dialog: thread count, path, environment, and whether it's a whole-project or single-checkout removal. Removal uses `force: true` like web, so the project's threads are deleted with it.

Thread List v2 (the default) has no project rows, so the Home list itself deliberately gets no entry point. Web and desktop already have removal. No contract changes; `project.delete` already existed.

## Verification

- `vp test run apps/mobile/src/features/projects/remove-project.test.ts` (confirmation copy)
- `tsc --noEmit` for `apps/mobile`
- `vp lint` on the touched files
- Signed Release build installed on a physical iPhone (iOS 26); swipe-to-remove in the picker exercised by hand. Screenshots to follow before marking ready.

Claude Fable 5.1 via Claude Code in T3 Code.


bawisniewski:
Superseded by a squashed, single-commit PR with the final shape (project picker swipe/long-press actions + settings). Closing to keep history clean.

bawisniewski:
Superseding PR: #12579

PR #12575, opened 2026-09-19
fix(web): keep typed text when a question option is clicked

### LINKED ISSUE #12569: [Bug]: Clicking a question option discards text typed into the composer
### Area

apps/web (shared desktop composer)

### Summary

While an agent is running, a structured question (single-choice options plus an editable answer field) takes over the composer. If I type into the composer while that question is active and then click one of the options, the text I typed is discarded immediately. Nothing brings it back: it is not in the thread draft, not in the answer, not in the message sent to the agent.

### Steps to reproduce

1. Start a Claude Code task (desktop app, Windows) and wait for the agent to work.
2. Start typing a follow-up in the composer.
3. The agent issues an `AskUserQuestion`-style question with several options.
4. Type (or continue typing) some text into the composer. With the question active, the composer edits the question's custom answer.
5. Click one of the options.

### Expected behavior

Choosing an option must not destroy text the user has typed. Either the text stays visible (as clarification or as the pending follow-up message), or it moves back to the thread draft. The user decides what to do with it.

### Actual behavior

The composer is emptied the moment an option is clicked. The typed text is gone with no recovery path.

### Source

Confirmed from source at `dfbb11bdd7c3f1a5575cb55d3e3abb12be025727` (main), same code as the installed build.

- Selecting an option rebuilds the draft answer with `customAnswer: ""`, so any text typed into the answer field is dropped by design: https://github.com/pingdotgg/t3code/blob/dfbb11bdd7c3f1a5575cb55d3e3abb12be025727/apps/web/src/pendingUserInput.ts#L96-L106
- The option handler in `ChatView` also resets the shared prompt ref to an empty string: https://github.com/pingdotgg/t3code/blob/dfbb11bdd7c3f1a5575cb55d3e3abb12be025727/apps/web/src/components/ChatView.tsx#L8835
- The editor is bound to `activePendingProgress.customAnswer` while a question is active, so the emptied answer is what the user sees: https://github.com/pingdotgg/t3code/blob/dfbb11bdd7c3f1a5575cb55d3e3abb12be025727/apps/web/src/components/chat/ChatComposer.tsx#L6806-L6811

The behaviour dates back to the original user-input implementation (`12edc3455`) and is covered by unit tests as intended (`togglePendingUserInputOptionSelection` returns `customAnswer: ""`). The design reason is that a non-empty custom answer takes precedence over selected options in `resolvePendingUserInputAnswer`, so the two are treated as mutually exclusive. That makes sense for the answer model, but it turns a click into silent data loss for text the user was composing.

### Relation to existing issues

- #11116 (accepted) covers the wider takeover: the composer switches to the question answer and Send routes to the question. This report is the destructive sub-case: even under the current takeover model, clicking an option should not delete typed text. A small fix is possible without the redesign (for example, move a non-empty custom answer back into the thread draft when an option is selected, instead of discarding it).
- #808 (closed) reported the reset when the question appears; that text is preserved in the thread draft. This case is different: the text is lost.

### Impact

Major degradation or frequent failure. Typing while the agent works is normal use, and a single click deletes the message.

### Version or commit

Installed desktop build 0.0.42 (`t3codeCommitHash` `719a76ca1dbf`). Source checked at `dfbb11bdd`.

### Environment

Windows 11 Home 10.0.26200, T3 Code desktop (Alpha), provider Claude Code.

### Workaround

Do not type while a question is active, or press Send with the typed text as the custom answer instead of clicking an option.

<sub><sup>lovely composed in pair with Claude Fable 5.1 in Claude Code</sup></sub>

myfunc:
## What Changed

Clicking an option in a pending user-input question no longer discards text typed into the composer.

- `pendingUserInput.ts`: new pure helper `carryDisplacedCustomAnswerIntoPrompt` with tests.
- `ChatView.tsx`: `onSelectActivePendingUserInputOption` moves a non-empty custom answer into the thread draft before the option replaces it.
- `ChatComposer.tsx`: when the question clears, the shared prompt ref is resynced from the thread draft.

## Why

While a question is active the composer edits the question's custom answer. Selecting an option rebuilds that answer with `customAnswer: ""`, because a custom answer outranks selected options. Anything the user typed in the meantime is gone: not in the answer, not in the draft, not in any message. Fixes #12569.

The second change closes a related gap: after the question resolved, the composer showed the thread draft but the prompt ref still held the last answer text, so Send without further typing read the wrong value.

Text moved into the draft comes back in the composer once the question resolves and is sent as the next message, same as text typed before the question arrived. Mobile keeps the answer in its own input on the question card and is not affected.

## UI Changes

No visual change. Interaction below: typing while a question is open, then clicking an option.

Before:

before

After:

after

## Checklist

- [x] I included before/after screenshots for any UI changes (not applicable, no visual change)
- [x] I included a video for animation/interaction changes (not applicable, behaviour covered by the steps below)

Validation:
- Before, on a dev build from `dfbb11bdd`: type while a single-choice question is open, click an option. The composer empties and the text is not in the draft, the answer, or any message.
- After: same steps. The answer goes out, the typed text is back in the composer, and Send submits it in full as the next message.
- Unit tests for the helper, `typecheck` for `apps/web`, targeted lint with only pre-existing warnings.

Tests cover the new helper only. There is no regression test that fails on `main`: the drop happens inside the `ChatView` handler, and a test there costs more than the fix. We chose to skip it on purpose for now, not sure yet whether it would be overengineering for a change this size. Feedback welcome.

<sub><sup>lovely composed in pair with Claude Fable 5.1 in Claude Code</sup></sub>

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved typed custom answers when selecting options for pending questions.
  * Restored the conversation draft after a pending question is completed or removed, preventing stale answers from being sent.
  * Combined displaced custom answers with existing draft text so no user input is lost.
* **Tests**
  * Added coverage for empty answers, empty drafts, and appending answers to existing drafts.


juliusmarminge:
Superseded by #12577 (merged) which also fixes #12569.

PR #12614, opened 2026-09-19
fix(web): hovering an expanded command row no longer shifts its text

atharvadeosthale:
## What Changed

On an expanded work entry row, the hover timestamp now always reserves its width and stays invisible until the row is hovered or focused. Collapsed rows are unchanged: the timestamp still takes no space until hover.

One added `className` prop in `MessagesTimeline.tsx`.

## Why

Since #8641, the timestamp switches from `absolute` to `static` on hover so it never overlaps the row text. On a collapsed row that only truncates the one-line preview a little earlier. On an expanded row the command text wraps, so taking width away on hover re-wraps the whole command, and the block jumps on every hover in and hover out.

Reserving the width up front on expanded rows means hover only changes opacity.

Checked in the dev web app against a copy of real thread data, on an expanded 510-character command row. Text width (575.39px), text height (136.5px), and row height (177.625px) are identical hovered and not hovered. Collapsed rows still compute `position: absolute` for the hidden timestamp.

## UI Changes

Before:

https://github.com/user-attachments/assets/cfacfaca-af4a-4f03-af88-043996866317

After:

https://github.com/user-attachments/assets/136e6af6-b41b-4fd7-b444-3fff84ac40ca

## Checklist

Claude Fable 5.1 via Claude Code in T3 Code

## Summary by CodeRabbit

* **Style**
  * Updated the timestamp styling for expanded work entries in the chat timeline.


atharvadeosthale:
Hey @juliusmarminge, slight bump for this PR, this fixes a problem that irritates me everyday :D

Sorry for the ping, I will avoid pinging for this PR from now on.

PR #12656, opened 2026-09-20
fix(mobile): dismiss ended Live Activities on foreground

EMOEMOJAI:
## What Changed

Dismiss ended T3 Code Live Activities when the iOS app refreshes its activity registration, before arming an activity for local work, and when signing out. The cleanup leaves active and stale activities available for their existing update-token registration.

Extend the existing `expo-widgets` patch with a factory-scoped native dismissal method. Await cleanup before discovering the current activity so a card created by another foreground path during that wait is reused. Cleanup errors are logged without preventing registration of the active card. Local arming captures the current registration generation and cancels if sign-out, account replacement, or provider teardown occurs while preferences or native cleanup are pending. A same-account token refresh still allows arming to finish.

## Why

A reported iPhone showed a small “Agent work completed” card above the current aggregate card. The small card remained for more than five minutes and survived reopening T3 Code, while new updates continued to arrive on the large card.

`expo-widgets` only returns active/stale activities from `getInstances()`. Consequently, the existing duplicate-card cleanup cannot see an ended activity that iOS still displays on the Lock Screen. Expo documents this deliberate exclusion in https://github.com/expo/expo/pull/48489. The native dismissal pass enumerates matching ended activities directly and ends them with immediate dismissal.

This change requires a new native iOS build; an OTA JavaScript update alone cannot add the native method. The failure-handling path keeps registration working when the native method is unavailable.

## Validation

- 50 tests pass in `remoteRegistration.test.ts` and `liveActivityPreferences.test.ts`.
- The two new cleanup-order regression tests fail against the original registration implementation and pass with this change.
- Six session-cancellation regression cases fail with the generation guards removed and pass with them restored; a seventh confirms same-account token refresh still permits arming.
- Mobile TypeScript check and lint of the four changed TypeScript files pass.
- Native iOS development-client build and installation pass with Xcode 27.0 (27A266a).
- Simulator verification passes on iPhone 18 Pro / iOS 27.0: the original registration code leaves the completed card visible; the PR code dismisses it while preserving the active ActivityKit ID. A subsequent local content update succeeds and appears on the remaining card.

The change applies to iOS registration, foreground refresh, local arming, and activity shutdown. The non-iOS adapter supplies a no-op; no server, provider, or wire-contract changes are needed.

## UI Changes

Verified with a temporary simulator fixture that creates real activities using the production T3 widget and invokes the production registration code. The before run uses `remoteRegistration.ts` from the PR parent; the after run uses the original fix commit `0b2006d`. The subsequent session-cancellation guard is covered by the focused regression tests above; the native patch is unchanged. Both runs use the rebuilt native binary (the original JS does not call the new native cleanup method). No test-only code is committed.

| Before: completed card remains alongside aggregate | After: one aggregate, updated after cleanup |
| --- | --- |
| Before | After |

Videos: [before clip](https://github.com/EMOEMOJAI/t3code/releases/download/pr-12656-verification/before-clip.mp4), [cleanup and successful update](https://github.com/EMOEMOJAI/t3code/releases/download/pr-12656-verification/after-clip.mp4). [Verification notes and fixture source](https://github.com/EMOEMOJAI/t3code/releases/tag/pr-12656-verification) are hosted as evidence assets on the contributor fork; no PR assets are committed.

The active ActivityKit ID is unchanged across both runs and after the successful update. Returning to the app exercises foreground registration without replacing the active card. The fixture uses a token provider returning null, so this verifies local ActivityKit cleanup and update continuity, not remote APNs delivery or cloud authentication.

The build used a temporary checkout without spaces because the existing Expo Widgets build script fails on the original `Coding Projects` path. CocoaPods 1.17.0 was installed as a prerequisite. No build workaround is included in this PR.

## Checklist

Model: GPT-6. Harness: Codex.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ended Live Activities are automatically dismissed before activities are started, refreshed, or when signing out.
  * Prevents duplicate Live Activities during refreshes and concurrent updates.
  * Active Live Activity registration continues even if cleanup fails.
  * Pending activity setup is canceled when preferences are disabled or the session changes.
  * Live Activity presentation is improved across supported widget rendering modes and device capabilities.


EMOEMOJAI:
@coderabbitai resume

EMOEMOJAI:
@coderabbitai full review

PR #12657, opened 2026-09-20
refactor(observability): hold OTLP export settings per signal

yordis:
- A single interval, header set, and wire format shared by all three signals cannot answer a question that was asked about one signal, so any future per-signal configuration has nowhere to land.
- Holding them per signal makes it structurally impossible to pair a credential or wire format from one source with an endpoint that came from another.
- Split out of #12621 so that pull request can shrink to the behavior it actually adds.

## Summary by CodeRabbit

- **New Features**
  - Added separate observability export settings for traces, metrics, and logs.
  - Each signal can now use its own protocol, headers, and export interval.
  - Added default export settings for consistent telemetry configuration.

- **Bug Fixes**
  - Updated trace, metric, and log delivery to use their corresponding signal-specific settings.


juliusmarminge (approved):
Clean split out of #12621; traced every consumer (tracer, metrics, logger, browser-trace proxy) and each reads its own signal, tests and typecheck pass on 55cc015. Two non-blocking nits: `Config.withDefault(10_000)` in `cli/config.ts` could read `DEFAULT_SIGNAL_EXPORT.exportIntervalMs` so the default lives once, and the `SignalExport` doc comment refers to a per-signal source that only arrives in #12621.

Issue #856, opened 2026-03-10
Bug: pending plan questions become stuck after app restart


marvin-bitterlich:
On `v0.0.9`, if T3 Code is fully closed while a thread is waiting on a plan-mode / `request_user_input` question, reopening the app leaves that thread stuck.

## Repro

1. Start a thread that triggers a plan-mode question.
2. Close T3 Code while the question is still pending.
3. Reopen the app.
4. Return to the same thread and try to answer or continue.

## Expected

The thread should recover cleanly and still allow the pending question to be answered.

## Actual

The question remains visible, but the thread is effectively stuck:
- submitting answers does not work
- there is no clear way to dismiss or recover the session

## Notes

This looks related to restart/resume handling for pending user-input requests. I did not find an existing open issue/PR for this exact case.

dlg0:
still an issue in 0.0.17

Issue #503, opened 2026-03-08
QnA prompts persist after app restart but can no longer be answered, session breaks


UtkarshUsername:
If the model asks questions through the QnA tool, and I close and restart the app before answering, the pending questions still appear after restart, but there is no way to answer them. Options can be selected but answers can't be submitted. This leaves the session stuck.

Expected behavior:
After restart, pending QnA prompts should either:
- be answerable again, or
- if that's not possible, be cleared so that the session can continue.

Actual behavior:
The old questions remain visible, but they cannot be answered, and the session becomes unusable.

Steps to reproduce:
1. Start a session, and ask the model to ask questions using the QnA tool.
2. Before answering the questions, close the app.
3. Reopen the app and return to the same session.
4. Observe that the pending questions are still shown.
5. Try to answer or continue the session.
6. Notice that the answers can't be submitted and the session is stuck.

shivamhwp:
@UtkarshUsername fixed ?
if yes close the issue, otherwise let me know !!

Sparky983:
This issue should be reopened as it can still be recreated on 0.0.21:
https://files.catbox.moe/39uho6.mp4
(funnily enough, I hit another bug where it says "Working for [x]", but that's unrelated to this specific issue)

Issue #479, opened 2026-03-08
Dismiss question/plan UI


notkainoa:
In the question/plan UI popup, there’s currently no way to close or dismiss it without answering.

This can get annoying because sometimes I want to skip a question and just prompt the agent.

Expected behavior:
There should be a clear way to exit the popup (for example an “X” button, a “Skip”/“Cancel” action, or closing on outside click / Esc).

basilk15:
Yeah same issue, ideally it should be like a drop down, showing that ok this is the plan its working on or even a clear way to exit or skip as the OP said.

Issue #469, opened 2026-03-08
[BUG] Stale agent questions after app restart


notkainoa:
After codex/the agent asks a follow-up question, the UI restores that question from thread history. The actual submit target for the answer, however, only lives in the in-memory provider session.

So if the user quits and reopens the app, the question card is restored visually, but the backing provider state is gone. The prompt still looks active, but submitting no longer works because the underlying request is invalid.

This leaves a stale question card that appears interactive but is effectively dead after restart, and the app does not detect or surface this state.

To fix, the app should either correctly restore pending question state across session recovery, or detect stale pending questions on reconnect and remove/disable them with a clear error.

PR #12664, opened 2026-09-20
fix(server): report dead-session approval and user-input answers as stale failures

bompus:
When the provider process dies while an approval or question card is pending (OOM-killed OpenCode ACP server, app/system restart), answering loops `provider.*.respond.failed` forever and settle/archive cannot recover the thread.

## Root cause

`processApprovalResponseRequested` / `processUserInputResponseRequested` in `apps/server/src/orchestration/Layers/ProviderCommandReactor.ts` emitted the literal detail `No active provider session is bound to this thread.` when no live session exists. That string matches none of the stale fragments consumed downstream, so the pending counts never clear and every submit just appends another identical failure:

- `derivePendingUserInputCountFromActivities` / `isStalePendingApprovalFailureDetail` (`ProjectionPipeline.ts`)
- the pinned-activity SQL (`ProjectionSnapshotQuery.ts`)
- `derivePendingRequests` (`packages/client-runtime`, shared by web/desktop/mobile)
- the decider's settle gating (`decider.ts`, whose comment mandates parity with the pipeline)

## Change

Both dead-session branches now emit the existing canonical `stalePendingRequestDetail(kind, requestId)` text (old: `No active provider session is bound to this thread.` / new: `Stale pending <kind> request: <id>. Provider callback state does not survive app restarts or recovered sessions. Restart the turn to continue.`). No contract, schema, or pipeline changes — all four consumers already match that string.

Healing mechanics: one more submit appends the stale-marked failure, and the next projection refresh (full replay over the request lifecycle, not an increment) drops the count to 0, so the card clears. No migration needed.

Known limit: this heals sessions marked stopped/missing. A provider dying while the session row stays live still routes through `respondToUserInput`/`respondToRequest` and depends on the adapter error text matching the unknown-pending fragments.

## Orchestrator V2 (`t3code/codex-turn-mapping`, PR #2829, still open)

Partially better there, but the hole remains: session release and startup recovery expire dead `runtimeRequests` (card drops from pending), and message-mode questions have a working Dismiss (`thread.user-input.dismiss`). But native provider-callback questions are undismissable by design — the Dismiss button renders only when `dismissible` (message mode; `ComposerPendingUserInputPanel.tsx`), answers on `not_resumable` requests are rejected (`dispatchRuntimeRequestRespond`), and `run.interrupt` does not clear runtime requests. So a dead native question missed by release/recovery wedges with no manual escape — the same stuck state as V1. This fix cannot merge there literally (different engine files); the V2 follow-up is to allow dismiss (`decision: cancel`) on `not_resumable` native requests and/or clear pending requests on run interrupt.

## Test plan

- `vp test run src/orchestration/Layers/ProviderCommandReactor.test.ts` (apps/server): 71 passed, including 2 new tests — stopped session + pending `approval.requested`/`user-input.requested` + respond yields the stale failure, never calls the provider, and the thread shell reads `hasPendingApprovals`/`hasPendingUserInput` as false.
- Server `typecheck`: 0 errors.

Follow-up debt (not in this PR): replace substring-matched failure text with a structured stale marker; manual dismiss UX for native question cards (#479).

Related #856, #503, #469.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error messages for approval and user-input responses when a provider session is no longer active.
  - Error details now identify stale pending requests and clarify that restarting the turn is required to continue.
  - Stale pending approval and user-input indicators are cleared after the failed response is processed.


f4llenz:
Reproduced #5454 on `0.0.43-nightly.20260924.2187` with the Claude adapter. The trigger was a nightly auto-update restart. `AskUserQuestion` was pending at 13:01Z, and the server restarted at 13:06Z. The thread auto-resumed and finished a turn in a new session. That session was idle-stopped at 13:38Z. Every later submit returned `No active provider session is bound to this thread.`, and `pending_user_input_count` stayed at 1.

This confirms the mechanism this PR relies on. I started a turn through `orchestration.dispatchCommand` so that the session was `running`. Then I sent the same `thread.user-input.respond` again. The Claude adapter rejected it as an unknown request, the reactor recorded the `stalePendingRequestDetail` text, and `pending_user_input_count` went from 1 to 0 at once. With this PR, the dead-session branch produces that same text, so one submit would have closed the card without the manual turn.

PR #12667, opened 2026-09-20
feat: Keep Awake toggle plus Esc stops agent turn

vishalx0707:
Adds two small features in one branch (two commits, one concern each).\n\nKeep Awake (commit 1): remote agent work dies when the host sleeps or the lid closes. Settings → Connections → This-machine now has a Keep Awake toggle. On Windows it holds prevent-display-sleep plus SetThreadExecutionState (SYSTEM+DISPLAY+AWAYMODE) and forces lid-close to Do Nothing with sleep/hibernate timeouts at never, saving all six original values on enable and restoring them on disable, quit, or next launch after a crash (userData recovery file, 25s re-assert loop). macOS/Linux hold the display-sleep blocker only.\n\nEsc stops turn (commit 2): Esc now triggers thread.stop, the same interrupt path as the round Stop button. Dialogs and menus close first, terminal focus is ignored, and it stays remappable in Keybindings settings.\n\nVerification: typecheck clean for contracts, desktop, web and shared; targeted web unit tests pass (keybindings, settingsSearch, KeybindingsSettings logic — 198 tests). No UI screenshots yet — desktop GUI was not launched in this environment.\n\nBuilt with Muse Spark via T3 Code.

## Summary by CodeRabbit

- **New Features**
  - Added a desktop-only **Keep Awake** setting to prevent the device from sleeping during active sessions.
  - Added support for checking availability and enabling or disabling Keep Awake.
  - Added Keep Awake to settings search.

- **Bug Fixes**
  - Pressing Escape now dismisses open menus, pickers, and floating panels before stopping a conversation.
  - Escape stops a conversation only when the terminal is not focused.


vishalx0707:
Split into two PRs: #12669 (Keep Awake) and #12670 (Esc stops agent turn).

PR #12696, opened 2026-09-20
fix(desktop): stop WSL backends when they are disabled

### LINKED ISSUE #12642: [Bug]: Disabling the WSL backend in a running desktop does not stop it; supervisor respawns it on exit
### Before submitting

### Area
apps/desktop

### Steps to reproduce
1. Windows desktop app with the WSL backend enabled and running (`desktop.backendInstance.start {"id":"wsl:Ubuntu"}` in desktop.trace.ndjson).
2. In settings, disable the WSL backend. The trace logs `desktop.settings.writeSettings`, `desktop.wslBackend.reconcile` with event `tearing down WSL backend`, `desktop.ipc.wsl.setEnabled`.
3. Inside WSL: `pgrep -fa wsl-runtime`. The `~/.t3/wsl-runtime/sha256-…/t3 --bootstrap-fd 0` process is still running and still listening.
4. Kill that process inside WSL.

### Expected behavior
Step 2 stops the WSL runtime process and unregisters the instance from the backend pool. Nothing respawns it while the setting is off.

### Actual behavior
The process survives the teardown (checked 90 seconds later). When it is killed, the pool supervisor restarts it:

    desktop.backendInstance.scheduleRestartFiber  "backend exited unexpectedly; restart scheduled"
    desktop.backendInstance.start {"id":"wsl:Ubuntu"}

This happened twice in a row (00:19:12 and 00:20:25). The setting itself persisted (`desktop-settings.json` no longer contains `wslBackendEnabled`), and a full app restart does not start the WSL backend, so only the live reconcile path is affected.

Side effect: the respawned WSL runtime shares `~/.t3/userdata` with a `t3 service install` server in the same distro, and each one re-registers its own port with the relay on startup (last writer wins), which repeatedly broke T3 Connect for the Linux environment.

### Impact
Degrades the experience (needs a full app restart to actually disable WSL; breaks remote access when combined with a WSL service).

### Version or commit
Desktop 0.0.42 (Windows).

### Environment
Windows 11, WSL2 Ubuntu (`networkingMode=mirrored`), `t3 service install` inside the same distro.

### Logs or stack traces

    00:17:14 desktop.settings.writeSettings
    00:17:14 desktop.wslBackend.reconcile  ["tearing down WSL backend"]
    00:17:14 desktop.ipc.wsl.setEnabled
    00:19:12 desktop.backendInstance.scheduleRestartFiber ["backend exited unexpectedly; restart scheduled"]
    00:19:12 desktop.backendInstance.start {"id": "wsl:Ubuntu"}
    00:20:25 desktop.backendInstance.scheduleRestartFiber ["backend exited unexpectedly; restart scheduled"]
    00:20:26 desktop.backendInstance.start {"id": "wsl:Ubuntu"}

### Workaround
Restart the desktop app after disabling the WSL backend.

yashranaway:
Disabling the WSL backend removes its pool entry but leaves the server running. Its supervisor can restart it afterward, even with WSL disabled, and compete with another service in the same distro for remote connectivity.

Registered instances accidentally inherit the pool's application scope from the captured Effect context. Apply the instance scope inside that context so unregistering an instance runs its stop finalizer, releases its process, and cancels scheduled restarts. The primary backend stays running, and the same WSL instance can be registered again.

Fixes #12642.

Validation:
- Two regression tests fail on main before the fix: running-instance teardown and pending-restart cancellation.
- All 42 focused backend pool, backend manager, WSL reconciliation, and WSL IPC tests pass after the fix.
- Coverage includes primary-process isolation, re-enabling the secondary, and application shutdown.
- Desktop typecheck and targeted lint pass.

The tests exercise the real pool and supervisor with scoped process doubles on Linux. A Windows/WSL end-to-end run was not performed. This changes desktop process ownership only; it does not touch the server provider or orchestration layers being rewritten for V2.

Model: GPT-6
Harness: Codex in T3 Code

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved desktop backend lifecycle handling during shutdown and restart scenarios.
  - Prevented pending secondary backend restarts from continuing after cancellation.
  - Ensured backend processes are cleaned up correctly when the desktop application shuts down.

- **Tests**
  - Expanded coverage for backend registration, re-registration, cancellation, process cleanup, and successful HTTP responses.


yashranaway:
Reviewed CodeRabbit's non-blocking docstring-coverage warning. The changed function is the test layer builder, and this PR adds no public API. The production scope-ownership trap is explained beside the fix; adding a comment that repeats the test setup would not add useful guidance. Keeping the focused behavioral coverage and the existing documentation style.

NawzBlaze:
I hit a scope issue while working on the same fix independently and wanted to share a finding that may help with the test approach here.

**Problem:** \Layer.effect\'s inner \Scope.Scope\ closes as soon as its build effect completes under per-test \Effect.provide\. This means \pool.register\ (which calls \Scope.fork(layerScope)\) always forks a closed scope in \it.effect\ tests — so the forked process never spawns, making it impossible to test \
egister\/\unregister\ against the real pool layer.

In production this doesn't happen because the pool layer survives via the massive shared \provideMerge\ build in \main.ts\ (multiple consumers hold observers > 0, keeping the layer scope alive).

**Solution:** Use \it.layer\ from \@effect/vitest\ instead of \Effect.provide\:

\\\	s
it.layer(poolLayerWithFakes)('suite-name', (it) => {
  it.effect('test', () => Effect.gen(function* () {
    const pool = yield* DesktopBackendPool.DesktopBackendPool;
    const instance = yield* pool.register(spec);
    yield* instance.start;
    // ... assert spawnCount, desiredRunning, etc.
  }));
});
\\\

\it.layer\ builds the layer once into a suite-lifetime \Scope.make()\ (kept open until the last test in the block runs), exactly mirroring production's shared-layer semantics. The \pool.register\ fork gets a live scope, processes spawn, and \pool.unregister\ correctly runs the auto-stop finalizer.

I confirmed this empirically: a \Layer.effect\ with \ddFinalizer\ only prints its close log when the \it.layer\ suite finishes — not when the test gen completes.

This lets the regression test exercise the real \pool.register\ / \makeBackendInstance\ / \pool.unregister\ path with a stubbed \ChildProcessSpawner\ (counting \spawnCount\), rather than mirroring the pipe externally.

yashranaway:
Thanks @NawzBlaze; here `Effect.provide` wraps the entire register/start/unregister operation, keeping the real pool alive throughout, and I reverified that reverting the scope fix fails two regressions while the current PR passes all six pool tests.

PR #12700, opened 2026-09-20
fix(server): keep usage-limits incomplete after sparse post-probe updates

### LINKED ISSUE #12170: Claude usage limits: a rate_limit_event after a failed probe is published as a complete window list

macodev00:
## What Changed

In `applyUsageLimitsUpdate` (`providerUsageLimits.ts`): copy `unavailable` on rebuild the same way `resetCredits` is copied; merge onto `previous === undefined` as `probeFailed`; keep identity on no-op so a `probeFailed` snapshot is not rebuilt and stripped.

Adds regression tests so a sparse `{ id: "seven_day" }` update after a failed/missing probe stays marked incomplete (not a clean one-window “complete” list).

Fixes #12170

## Why

A turn `rate_limit_event` after a failed Claude boot probe was upserted onto `[]` / `probeFailed` and published as an unmarked weekly-only snapshot. Clients treated that as a full limits read. `resolveUsageLimitsAfterProbe` is unchanged — a still-marked snapshot is not last-good.

## UI Changes

No UI change. Existing Limits notice (“Could not read limits.”) stays correct while the incomplete marker is on. Screenshots/video not applicable.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved usage-limit tracking for partial updates when provider availability information is incomplete.
  - Preserved existing unavailable status across unchanged or partial updates.
  - Ensured sparse updates without prior usage data remain marked as unverified until a complete probe succeeds.
  - Prevented incomplete failed-probe results from being treated as the latest valid usage snapshot.

- **Tests**
  - Added coverage for sparse updates, unavailable status preservation, repeated identical updates, and failed probe resolution.


macodev00:
@coderabbitai review

macodev00:
@coderabbitai review

PR #12701, opened 2026-09-20
fix(server): tolerate slow OpenCode version probes

### LINKED ISSUE #11080: [Bug]: OpenCode provider flagged Unavailable when CLI version probe exceeds hardcoded 4s

macodev00:
## What Changed

Raise OpenCode CLI version-probe timeout from 4s to 10s (`OPENCODE_VERSION_PROBE_TIMEOUT` in `OpenCodeProvider.ts`). Process-group kill on timeout is unchanged.

In `mergeProviderSnapshot`: if a probe times out after a ready, versioned snapshot, keep that version/auth/ready status (plus cached models/skills) instead of flipping to error. First-boot timeouts with no prior version still error. Missing binary / inventory failures still error.

Fixes #11080

## Why

The 4s budget was added for hung wrappers (#8750), not slow-but-healthy OpenCode startups (e.g. ARM under load). A single timeout was clearing ready status and making the provider look Unavailable.

## UI Changes

No UI change. Provider health behavior only. Screenshots/video not applicable.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * OpenCode version checks now allow up to 10 seconds, improving detection on slower or heavily loaded systems.
  * When a refresh times out, previously confirmed OpenCode provider details remain available instead of being cleared.
  * Providers without a previously confirmed version continue to show an error when version detection times out.


macodev00:
@coderabbitai review

macodev00:
@coderabbitai review

PR #12714, opened 2026-09-20
fix(web): shorten Windows notification tags under platform limit

### LINKED ISSUE #12287: [Bug]: Completion notifications never appear on Windows: 73-character notification tag exceeds the platform tag limit

macodev00:
## What
- Hash Windows completion notification tags to a short stable hex digest so Chromium does not silently drop toasts over the platform tag length limit.

## Why
Fixes https://github.com/pingdotgg/t3code/issues/12287

`environmentId:threadId` is 73 characters; Windows discards renderer notifications over the documented tag budget. A 16-char digest keeps uniqueness / per-thread replacement while staying under the limit.

## UI
N/A — notification plumbing only (no visible chrome change beyond the toast appearing on Windows).

## Checklist
- [x] I read CONTRIBUTING.md and kept this small / bug-only
- [x] What + why explained above
- [x] Followed AGENTS.md documentation rules (no new docs)
- [x] No unrelated changes
- [x] CLA / AI-reviewed standing override for macodev00 auto-contrib

## Summary by CodeRabbit

* **New Features**
  * Added compact, consistent identifiers for desktop thread notifications.
  * Kept notification identifiers within supported length limits for improved compatibility.
  * Ensured notifications for the same environment and thread remain grouped, while distinct combinations receive separate identifiers.
* **Bug Fixes**
  * Improved desktop notification tracking across environments and threads.


macodev00:
@coderabbitai review

macodev00:
@coderabbitai review

macodev00:
@coderabbitai review

macodev00:
@macroscope-app review

macodev00:
@coderabbitai review

macodev00:
@macroscope-app review

macodev00:
@coderabbitai review

macodev00:
@coderabbitai review

PR #12727, opened 2026-09-20
fix(mobile): keep size restoration while live-following to stop iOS scroll teleport

### LINKED ISSUE #12072: [Bug]: iOS: scrolling up in a long thread teleports the viewport and draws rows outside the list bounds
https://github.com/user-attachments/assets/f2979916-a525-4192-a0b3-2a895895b05f

### Before submitting

### Area

apps/mobile

### Steps to reproduce

This one is intermittent and I cannot trigger it on demand. It hits me several times a week. The screen recording below captures a full episode; here is the shape of it:

1. Open a long thread that contains at least one very tall tool-output row (mine has a skill listing of roughly 100 rows).
2. Let the agent finish a reply while the viewport is parked somewhere above the bottom.
3. Tap the scroll-to-bottom button. This works — the feed lands at the end.
4. Drag up a short distance to re-read the reply.

### Expected behavior

A short upward drag moves the viewport by the dragged distance, and rows stay inside the list bounds.

### Actual behavior

Three distinct symptoms, all visible in the recording:

1. **The viewport teleports.** A short upward drag jumps to the start or the middle of the thread instead of moving by the dragged distance. It flips between the conversation region and the large tool-output table many screens away.
2. **Rows render outside the list bounds.** Conversation text is drawn over the navigation header and below the composer, so the list stops clipping its rows.
3. **The scroll-to-bottom chevron floats mid-screen** instead of sitting above the composer, which suggests the computed layout is wrong rather than just the scroll offset.

**Hypothesis, from reading the source:** `apps/mobile/src/features/threads/ThreadFeed.tsx` already documents this failure mode above `getFixedItemSize`:

> every row above the viewport is assumed to be `estimatedItemSize` tall, and scrolling up through unmeasured content corrects each row's height as it mounts — the feed visibly jumps

The mitigation from #4874 only assigns fixed heights to the small chrome rows. Message rows return `undefined` and fall back to LegendList's per-type running average against `estimatedItemSize={180}`. A single tool-output row an order of magnitude taller than that average would produce a large offset correction the moment it mounts while scrolling up, which matches symptom 1. Symptom 2 looks adjacent to #11813.

**Question:** #11813 landed after the 1.1.0 App Store build I am on. Does the 1.2.0 train already change this behavior? I am happy to re-test and close this if so — I just have no way to run that build yet.

### Impact

### Version or commit

App Store 1.1.0

### Environment

iPhone 15 Pro Max, iOS 27

### Logs or stack traces

```shell

```

Screen recording attached in a follow-up comment.

### Workaround

Tapping the scroll-to-bottom button re-anchors the feed, but the next upward drag can teleport again.

macodev00:
## What
- Keep size restoration on while live-follow is armed (`maintainVisibleContentPosition` size:true, data:false) so the first upward drag after scroll-to-bottom cannot measure tall rows in a render gap and teleport the viewport.

## Why
Fixes https://github.com/pingdotgg/t3code/issues/12072

With MVCP fully off during follow, an unmeasured tall assistant row could jump the list and paint over header/composer.

## UI
iOS long-thread scroll should stay stable when leaving live-follow (no new chrome).

## Checklist
- [x] I read CONTRIBUTING.md and kept this small / bug-only
- [x] What + why explained above
- [x] Followed AGENTS.md documentation rules (no new docs)
- [x] No unrelated changes
- [x] CLA / AI-reviewed standing override for macodev00 auto-contrib

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live-follow behavior in thread feeds to keep the view stable as new content arrives.
  * Preserved scroll position more reliably when reading history or while expanded and collapsed content settles.
  * Reduced unexpected jumps when switching between live updates and previously loaded content.

* **Tests**
  * Added coverage for thread-feed position restoration during live-follow, history reading, and disclosure-settling scenarios.


macodev00:
@coderabbitai review

macodev00:
@coderabbitai review

macodev00:
@coderabbitai review

macodev00:
@macroscope-app review

macodev00:
@coderabbitai review

macodev00:
@coderabbitai review

PR #12736, opened 2026-09-20
fix(web): discarding a draft from the sidebar asks first

### LINKED ISSUE #12735: [Bug]: Hover-only "Discard draft" X in the sidebar deletes the draft on one click, no confirmation or undo
### Area

apps/web (sidebar, shared with the desktop app)

### Summary

Each thread row with an unsent draft shows an X ("Discard draft") on hover, at the right edge next to Snooze and Settle. One click on it deletes the draft text and attachments. There is no confirmation, no undo, no toast. The X only appears on hover, so the first time you learn it exists is when your message is gone. I have lost drafts to it several times already.

### Steps to reproduce

1. Open a thread, type a message in the composer, do not send.
2. Open another thread. The first row now shows the amber draft marker.
3. Move the pointer onto the first row to go back to it. The hover strip at the right edge shows X, Snooze and Settle over the time/status label.
4. Click while the pointer is on the X.

### Expected behavior

A destructive action on user text needs a guard: a confirmation before, or an undo (toast) after. Alternatively the X should not sit inside the row's own click target, so a click meant to open the thread cannot delete its draft.

### Actual behavior

The draft is cleared at once and the row loses its marker. Nothing brings the text back. The standalone new-thread draft rows above the pinned block have the same X with the same result.

Hover state right before the click. The X sits where the pointer lands when going back to the thread:

Sidebar thread row on hover: Discard draft X next to Snooze and Settle

### Source

Checked at `main` @ `7445aa733`.

- Thread row: X is rendered only on hover, inside the row, next to Snooze/Settle: https://github.com/pingdotgg/t3code/blob/7445aa733ada33e45289e5aa5055f79142556513/apps/web/src/components/Sidebar.tsx#L1860-L1888
- Its handler releases uploads and clears the composer, nothing else: https://github.com/pingdotgg/t3code/blob/7445aa733ada33e45289e5aa5055f79142556513/apps/web/src/components/Sidebar.tsx#L1076-L1084
- New-thread draft row: same X, calls `clearDraftThread`: https://github.com/pingdotgg/t3code/blob/7445aa733ada33e45289e5aa5055f79142556513/apps/web/src/components/Sidebar.tsx#L790-L803 and https://github.com/pingdotgg/t3code/blob/7445aa733ada33e45289e5aa5055f79142556513/apps/web/src/components/Sidebar.tsx#L897-L906
- Introduced in #9658 (thread rows) and #5777 (draft rows). `docs/user/thread-sidebar.md` does not mention the control.

### Relation to existing issues

- Discussion #10349 (Ideas, 2026-09-06) reports the same loss and asks for removal or a confirmation. No response so far. Filing as a bug because the outcome is silent loss of typed text, the same class as #12569 (accepted) and #8963, not a feature proposal.

### Impact

### Version or commit

Desktop 0.0.42, source checked at `main` @ `7445aa733`

### Environment

Windows 11 Home 10.0.26200, T3 Code desktop (Alpha), provider Claude Code

### Workaround

None after the click. Prevention only: click the left part of the row, or copy the text before leaving the thread.

<sub><sup>lovely composed in pair with Claude Fable 5.1 in Claude Code</sup></sub>

Mnigos:
Fixes #12735.

## Problem

In the sidebar, a thread row with an unsent draft and a new-thread draft row both show a hover-only X ("Discard draft"). One click wiped the prompt, attachments, and contexts with no confirmation, undo, or toast. The X sits on the row's own click target and `composerDraftStore` keeps no snapshot, so a click meant to open the thread could lose the message for good.

## Fix

The triage offered a confirm or an undo toast. I went with the confirm, because it matches what already exists: mobile guards the same action with `Alert.alert("Discard draft?", …)`, and sidebar delete and archive use `api.dialogs.confirm`.

- `confirmDiscardDraft` in `Sidebar.tsx` opens the destructive confirm dialog ("Discard draft? The unsent message and its attachments will be removed.").
- Both discard paths await it before touching the store: `handleDiscardDraftClick` on thread rows (`clearComposerContent`) and `handleDiscard` on draft rows (`clearDraftThread`). Uploads are released only after confirming.

No new setting: unlike delete and archive, there is no way to get a discarded draft back, so an opt-out did not seem worth it. Moving the X out of the row's click target is left out to keep this to one concern; with the confirm in place a stray click no longer loses anything. `docs/user/thread-sidebar.md` does not mention the control, so there is nothing to update.

## Verification

Web typecheck and targeted lint are clean. No unit test: the change routes two click handlers through the existing dialog, and a test would only assert that wiring.

Recorded in a real browser against the dev server, on a new-thread draft row. After: the X opens the dialog, Cancel keeps the draft, Confirm removes it.

| Before: one click, the draft is gone | After: the click asks first |
| --- | --- |
| before | after |
| before still | after still |

Implemented with Claude Code (Claude Fable 5.1).


myfunc:
What do you think about "double-click" on it without modal? I don't remember where I've seen this practice, but I found it nice.

image

Mnigos:
> [!NOTE]

Nice idea, and lighter than a modal. I kept the confirm dialog here because the triage on #12735 asked for a confirm or an undo toast, and the dialog matches what mobile and thread delete/archive already do. An inline "discard?" second step would be a new pattern in the sidebar (extra state, reset on mouse leave, and a fast accidental double-click still loses the draft), so I would rather leave that call to the maintainers. Happy to switch if they prefer it.

PR #12765, opened 2026-09-20
fix(devices): stream physical Android devices

ttmx:
## What Changed

Physical Android devices can now be viewed and controlled from the Device panel. :)

- Select `scrcpy` as the stream source when opening a physical Android device, while preserving the existing emulator behavior.
- Decode the SPS keyframe immediately after configuring WebCodecs instead of discarding it and requesting another keyframe.

## Why

Physical Android devices do not support the `grpc-screenshot` stream mode, preventing their video streams from working. `scrcpy` is supported.

This also exposed a decoder bug: the first keyframe was used for codec configuration and then discarded. The `grpc-screenshot` path worked with the subsequent keyframe request, but the `scrcpy` path did not. The initial packet already contains a valid keyframe, so it can be decoded immediately after configuration.

## Testing

- `vp test run apps/server/src/device/DeviceService.test.ts apps/web/src/components/device/deviceStream.test.ts`
- Verified that Android emulator video and controls continue to work through the Device panel.
- Tested that a physical device can now be interacted with in the devices tab, instead of showing as stuck on "Connecting video..."

## Checklist

https://github.com/user-attachments/assets/4971d6b4-1ca3-4098-9433-fe13a9e26d22

- [x] I included before/after screenshots for any UI changes — not applicable
- [x] I included a video for animation/interaction changes — not really applicable but included regardless for the fix

Model: GPT-5.6 Sol/ GPT-6 Astra 
Harness: Codex in T3 Code

## Summary by CodeRabbit

- **New Features**
  - Physical Android devices now automatically select the scrcpy stream source when streaming begins.

- **Bug Fixes**
  - Improved Android video startup by decoding the initial keyframe immediately when available, reducing unnecessary keyframe requests and helping streams begin more reliably.
  - Android streams with SPS configuration data now configure video decoding without resetting the active stream.


ttmx:
> _🟠 Major_ · Invalidate pending decoder configuration when the video session changes. · `stream.ts:566-584`
> > `packages/client-runtime/src/device/stream.ts:566-584`
> > _🩺 Stability & Availability_ | _🟠 Major_ | _⚡ Quick win_
> > **Invalidate pending decoder configuration when the video session changes.**
> > If `"video-session"` arrives while `configureDecoder()` is pending, `closeDecoder()` clears `videoDecoder` but does not invalidate the pending callback. If the captured packet is a keyframe, the callback can configure a decoder from the old SPS and decode that old packet. New-session SPS/keyframe packets received while `configuring` is true are discarded. This can leave the Android decoder unable to render the new session until another keyframe arrives.
> > Track a video-session generation. Capture it before calling `configureDecoder()`. If the generation changes before the callback completes, close the stale decoder and call `requestKeyframe()` instead of decoding the captured packet. Increment the generation when handling `"video-session"`.

This was a preexisting issue, but I can add it to the PR if the maintainers wish.
I don't think this issue will actually appear in practice though, and handling it might be too much code for what it prevents.

PR #12791, opened 2026-09-20
fix(server): launch contained ACP providers from packaged runtimes

pekth:
Packaged Linux T3 can launch its own CLI instead of the ACP provider when cgroup containment is enabled. The wrapper calls `process.execPath -e`, but in the packaged runtime that executable is T3 rather than Node. Preview `0.0.43-preview.20260919.1974` emitted CLI help on ACP stdout, and Grok rejected it as invalid JSON.

Use `/bin/sh` builtins to enter and verify the cgroup, clear wrapper environment markers, and exec the provider with its original arguments. This retains failure exits 125 and 126 without requiring a Node evaluator.

This is a T3 Code packaged-runtime bug, not a Fleet configuration bug. The actual preview binary rejects `-e`, exits 1, writes `DESCRIPTION` CLI help to stdout, and reports `Unrecognized flag: -e` on stderr. A separate source-derived comparison used the same fake lease and arguments without importing or executing Fleet. The original T3 wrapper did not start the target; the patched wrapper exited 0, preserved the exact arguments, cleared the wrapper markers, and wrote its PID only to the fake cgroup file. Fleet has a separate local push-gate defect, which is outside this PR.

Validation: a regression with `process.execPath` stubbed to a non-Node executable failed against the original code. All 23 focused process-tree tests passed, including argument/marker preservation, membership mismatch, and a missing executable. The real cgroup containment test also passed separately. Server typecheck, scoped lint, formatting, and diff checks passed. The running preview was not replaced, so successful installed-candidate Grok startup remains unverified.

Targets the V2 branch tracked by #2829.

Models: GPT-5.6 Luna for implementation, GPT-5.6 Sol for independent review, and GPT-6 Astra for controller verification. Harness: Codex through T3 Code preview.


juliusmarminge:
Rebased onto the current v2 head (no code changes). Locally one assertion in `AcpSessionRuntime.processTree.test.ts` (`resolveLinuxCgroupTargetCommand("node", …, { PATH: undefined })`) fails on my box because no `node` lives in `/usr/bin` or `/bin`; it fails identically on v2 without this PR, so it is environmental and not yours. Will merge once CI is green.

PR #12799, opened 2026-09-20
fix(server): Grok accounts with no usage yet no longer vanish from Limits

jakeleventhal:
xAI omits `creditUsagePercent` from its billing response entirely until an account has accrued metered usage — it sends the field late rather than sending `0`. We mapped that absence to `unavailable: { reason: "unsupported" }`, which the contract reserves for accounts that can *never* report quota (API keys, Bedrock). `collectLimitAccounts` drops those accounts, and `collectLimitNotices` deliberately mutes their notice because there is nothing to act on. The result: a freshly signed-in Grok account disappears from the Limits view with no bar and no explanation, until it happens to be used. `applyUsageLimitsUpdate` also refuses mid-turn windows for an `unsupported` snapshot, so nothing could recover it in the meantime.

I hit this with a second Grok account: it was signed in, listed in the model picker, and running threads, but simply absent from Usage with nothing to indicate why.

`grokUsageResponseToLimits` now reports no windows and leaves the `unavailable` marker off when the billing read succeeds but carries no percentage. The account keeps its row and picks up the existing "No limits reported." notice, and the bar appears on its own once usage registers. The genuine can-never-report paths in `readGrokUsageLimits` (API key, alternate auth deployments, custom config layers) still return `unsupported` and are untouched.

## Before

The second Grok account is signed in and working, but nothing on the page mentions it.

Before

## After

After

Captured against a dev server with two Grok instances, with the billing response for the second stubbed to the real payload xAI returns for an unmetered account (`currentPeriod` present, no `creditUsagePercent`). That stub was only for the screenshots and is not part of the diff.

`vp test run apps/server/src/provider/Layers/GrokProvider.test.ts packages/shared/src/usageLimits.test.ts` — 62 passing. The shared side already covered this shape: a provider with empty windows and no marker yields "No limits reported." while an `unsupported` one is skipped, so no client change was needed.

Model: Claude Opus 5 (1M context). Harness: Claude Code in T3 Code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Accounts with temporarily unavailable or incomplete billing usage data now remain visible in the Limits view.
  * Missing or invalid usage measurements no longer classify accounts as unsupported.


apps/server/src/provider/Layers/grokUsageLimits.ts
```diff
@@ -41,7 +41,15 @@ export function grokUsageResponseToLimits(
...
   if (usedPercent === undefined || !Number.isFinite(usedPercent)) {
-    return makeUnavailableUsageLimits({ checkedAt, reason: "unsupported" });
+    // A billing read that succeeded but carries no percentage is an account
+    // with nothing metered yet, not one that can never report: xAI omits the
+    // field entirely (rather than sending 0) until usage registers, then fills
+    // it in. Calling that `unsupported` would strand the account — the Limits
+    // view drops unsupported entries and deliberately mutes their notice, so a
+    // freshly signed-in Grok account would vanish with no explanation until it
+    // happened to be used, and `applyUsageLimitsUpdate` would refuse the
+    // mid-turn windows that could have recovered it.
```
jakeleventhal, line 51: since the t3 team has collaborated with grok, maybe something to ask them to fix on their end

PR #12815, opened 2026-09-21
fix: render images whose destination has spaces

ValeraZSD:
## Problem

When an agent writes an image whose path contains a space unquoted, the whole line renders as literal text:

```
![Settings → General](C:\Users\me\My Pictures\before after.png)
```

CommonMark ends an unquoted link destination at the first space, so this never parses as an image — the renderer shows the raw markdown line instead (observed on mobile and web; desktop serves the web client, so it inherits the behaviour).

The machinery for the destination itself already works — once the syntax parses, the workspace-path classification relays the image through a signed asset URL on mobile and resolves it on web. Only the parser never sees it.

## Fix

`repairMarkdownImageDestinations` (new export in `@t3tools/client-runtime/markdown-links`) rewrites such destinations to the angle-quoted form — the CommonMark syntax that holds spaces — before parsing:

```
![shot](C:\dir with spaces\a.png)   →   ![shot](<C:\\dir with spaces\\a.png>)
```

The backslashes are doubled because a backslash before punctuation is an escape inside a destination too: written once, `<D:\shots\.cache\a b.png>` parses back as `D:\shots.cache\a b.png` and resolves to nothing. Both clients already unwrap `<...>` when classifying, so the rewrite is invisible downstream. It runs at the two chat entry points: web `ChatMarkdown` (which desktop reuses) and mobile `ThreadFeed` (message rows and assistant content). The clients share no renderer — web parses with `react-markdown`, mobile with `react-native-nitro-markdown` — so the text before parsing is the only seam both can use.

Everywhere Markdown is shown literally is left exactly as written: fenced code (including a fence whose info string is not a valid one), indented code, code spans (scanned per block, since a span can close on a later line), raw HTML blocks, and an escaped `\![`. So are destinations that already parse — relative, angle-quoted, quoted-title — and prose that is not a path.

Scope is images only; link syntax is deliberately left for a separate pass.

## Tests

- 23 unit tests in `packages/client-runtime/src/markdownLinks.test.ts`: Windows, UNC and POSIX paths with spaces, a path segment starting with punctuation, an escaped parenthesis inside the destination, bracketed and escaped-bracket alt text, several images on one line with balanced parens, and each context that must stay untouched — parseable destinations, prose, fenced code, an unclosed fence, a fence-looking line that opens no fence, a code span closing on a later line, an unmatched backtick, indented code, a raw HTML block, an escaped image marker
- three render tests in `apps/web/src/components/ChatMarkdown.workspace-images.test.tsx`: the unquoted spaced path reaches the signed asset URL instead of "Image unavailable", a path through a `.cache` segment arrives with its separators intact, and copy-as-markdown puts a destination on the clipboard that another Markdown reader parses

Required CI is held for a maintainer's approval on this PR, so it has not run. Locally, on Windows: `vp check` 0 errors; `packages/client-runtime` 1,586 tests; `apps/web` unit project 5,323 of 5,325; `apps/mobile` 1,677 of 1,679; `tsc` clean in all three packages. The four failures are pre-existing and environmental — two assert an English OS locale, two compare `git` output that arrives CRLF here — and none are in files this PR touches. Nothing in the diff reaches Rust, mobile native or release smoke.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed workspace images with spaces, Windows-style paths, hashes, parentheses, and other special characters failing to render in chat and thread feeds.
  - Improved Markdown image handling across fenced code, HTML blocks, links, headings, and indented code without altering valid content.
  - Fixed copied image Markdown so destinations remain compatible when pasted into other Markdown tools.
  - Improved assistant message rendering for repaired image references, reducing unnecessary “Image unavailable” fallbacks.


ValeraZSD:
Thanks — the bot findings were right about the direction of the risk: the repair was reaching text that is meant to be read literally. `ac5d878` takes all seven Macroscope findings and the three from CodeRabbit, each with a test in `markdownLinks.test.ts`.

**Corruption of literal Markdown (repair now skips it)**

- **Code span crossing lines** — spans are scanned over a whole block instead of a line, so a span opened on one line and closed on another still shields what is between them.
- **Unmatched backtick** — an opener that never closes is literal text, so it shields nothing; the rest of the line is still repaired. (Both come from one rewritten `inlineCodeSpans`, which also now closes on a run of *exactly* its own length, as CommonMark requires.)
- **Escaped `\`** — skipped, by counting the backslash run before the marker.
- **Raw HTML block** — a line opening `<pre|script|style|textarea>` puts the walker inside a raw block until the closing tag; `rehypeRaw` gets it as written.
- **Indented code** — four spaces or a tab where a block can start is code. Where the same indent continues a paragraph it is text, and it is still repaired.

**Wrong output where it did repair**

- **Escaped `)` in the destination** — the scan steps over escapes, so `![x.png)` keeps the whole filename instead of ending at `\>`.
- **Bracketed alt text** — `IMAGE_OPEN_PATTERN` now accepts escaped brackets and one level of balanced brackets, so `![see \[this\]](…)` and `![a [b] c](…)` are repaired rather than ignored.
- **Fence opener with a backtick in its info string** — that opens no fence, so the Markdown under it is live and is repaired. (CodeRabbit's phrasing asked for "no non-whitespace after the marker", which would reject ```` ```js ````; the CommonMark rule is the narrower one.)
- **`markdownImageCopy`** — angle-quotes a destination carrying whitespace, so copy-as-markdown produces something another reader parses.

**One the review did not find.** Backslash escapes are processed *inside* a destination, so the angle-quoted form still lost a separator before punctuation: `<D:\my projects\.cache\a b.png>` parses to `D:\my projects.cache\a b.png`, which resolves to nothing. The emitted destination now doubles its backslashes, which is also correct for a UNC path. Measured against `mdast-util-from-markdown` (the parser `react-markdown` uses here), before and after:

```
![a](<D:\shots\.cache\_x\a b.png>)          => D:\shots.cache_x\a b.png
![a](<D:\\shots\\.cache\\_x\\a b.png>)      => D:\shots\.cache\_x\a b.png

![a](<\\server\share\my shots\a b.png>)     => \server\share\my shots\a b.png
![a](<\\\\server\\share\\my shots\\a b.png>) => \\server\share\my shots\a b.png
```

One helper, `markdownImageDestination`, now writes that destination for both the repair and the copy path. The web app also memoises the repair on the text it is handed.

**On the approvability note** — why this is shared text-level code rather than an isolated renderer fix. The two clients share no renderer: web is `react-markdown`, mobile is `react-native-nitro-markdown`, so a remark plugin would fix one of them. It would not fix either, in fact: when the destination fails to parse there is no `image` node to visit, and no text-node rewrite can be safer than this one — it would face the same contexts with less information, since by then `\!` has already become `!`. The text before parsing is the only seam both clients have, and the repair is a pure function with the block rules stated in tests.

Scope is unchanged: images only, no link syntax, and a destination that already parses is left exactly as written.

Verified locally: `packages/client-runtime` 1,579 tests pass (`tsc` clean); `apps/web` unit project 5,323 of 5,325, the workspace-images suite covering both the end-to-end render and the clipboard; mobile `tsc` clean. The two `apps/web` failures are not this change. Unrelated, but worth knowing: `apps/web/src/components/Sidebar.snooze.test.ts` fails on a machine whose OS locale is not English — it asserts `/Mon/` against a `toLocaleDateString` result (mine is `ru-RU`, so it reads `пн`). Nothing in this PR touches it.

ValeraZSD:
Round two, `76ad136` — all three taken, each with a test, each checked against `mdast-util-from-markdown` first.

- **HTML blocks beyond the raw-text four.** Real, and the same corruption class as `<pre>`: measured, an image inside `<div>`, `<!-- -->`, `<![CDATA[` or a `<table>` block produces no image node, so rewriting there changes what the reader is meant to see. The walker now tracks all seven CommonMark forms with their own end conditions — the raw-text tags to their closing tag, a comment to `-->`, `<?` to `?>`, `<!NAME` to `>`, CDATA to `]]>`, the block tag list and a complete tag alone on a line to the next blank line. A tag alone on a line only opens a block where a block can start, so `<em>hi</em> x` is still a paragraph and is still repaired — that one is a test too.
- **Indented code not ending at an unindented line.** Real. The state was carried to the next blank line, so the paragraph after a code block was skipped. It now clears on the first unindented line.
- **CRLF.** Real, with the opposite effect to the one described: `FENCE_LINE_PATTERN` ends `(.*)$`, so `\r` is absorbed and the opener *is* recognised — it is the closing pattern that rejects it, so a fence opened and never closed and everything after it went unrepaired. Nothing inside a fence was ever modified. Fixed at the source rather than in each pattern: the walker reads each line without its terminator and keeps the terminator on the line it rewrites, so a CRLF document stays CRLF.

Worth stating explicitly, since it decides how these findings read: the repair only ever **skips** or **angle-quotes**. Every miss leaves text exactly as the model wrote it, which is today's behaviour, so the failure mode of this walker being too cautious is the status quo and the failure mode of it being too eager is a corrupted message. Where the two conflict — the standalone-tag rule, an indent inside a list item — it is written to be cautious.

`packages/client-runtime` 1,584 tests pass, `tsc` clean, formatter clean; `apps/web` workspace-images 36/36.

ValeraZSD:
Round three, `18c4ad6` — taken, correct, and the reasoning holds up against the parser.

`blockLines.length > 0` was standing in for "a paragraph is open", and a heading satisfies it while leaving no paragraph open. Measured on `mdast-util-from-markdown`, with the image angle-quoted so only the block context decides:

```
# H\n<x>\n![x](</tmp/a b.png>)      => no image   (HTML block)
---\n<x>\n![x](</tmp/a b.png>)      => no image   (HTML block)
H\n===\n<x>\n![x](</tmp/a b.png>)   => no image   (HTML block)
text\n<x>\n![x](</tmp/a b.png>)     => image      (paragraph, tag cannot interrupt)
- item\n<x>\n![x](</tmp/a b.png>)   => image      (lazy continuation)
```

So an ATX heading, a thematic break and a setext underline now end the block they are in, and the two that must stay repairable — a paragraph and a list item above the tag — are tests alongside them, as is an image *inside* a heading, which the same change must not stop repairing.

`packages/client-runtime` 1,586 pass, `tsc` and formatter clean.

ValeraZSD:
Related: #12615 fixes the other half of this, and touches the same files — worth reading together, since neither covers the other's case.

- #12615 rewrites `\` → `/` inside drive-letter destinations, so `C:\Users\dara\.t3\shot.png` stops losing its separator to CommonMark's backslash escapes. Its `bareDestination` scan ends at the first whitespace and its tests carry no spaced path, so `shot` is still not an image after it.
- This PR angle-quotes image destinations that contain whitespace, which is the only form that survives a space. `isRepairableImageDestination` returns false for a destination with no whitespace, so the `.t3` case above is untouched by it.

Complementary rather than competing, then — but they will conflict in `markdownLinks.ts`, `ChatMarkdown.tsx` and `ThreadFeed.tsx`. If #12615 lands first I'm happy to rebase onto `markdownWindowsPaths.ts` and reduce this to the whitespace case: the forward-slash rewrite is a cleaner way to survive the escapes than the backslash doubling here, and this PR could drop that part entirely. Tell me which order suits you and I'll do it.

One gap neither closes: #12615 covers links and reference definitions as well as images; this one is images-only by design, so a *link* whose destination contains spaces still breaks. Easy to widen if you want it.

The workflows here are still held for approval (CI, Web Preview, Mobile EAS Preview, Mobile Fingerprint Check), so nothing has run on this branch — the local equivalents are in the description.

PR #12846, opened 2026-09-21
perf(server): stop decoding unrelated events during startup

extoci:
Attachment cleanup replay read and decoded every event after its retry cursor during bootstrap, even though it only needs thread revert/delete metadata and the newest sequence. On a 1.25 GB snapshot of a real T3 database, that path decoded 4,954 events and 23.8 MB of JSON after an ordinary restart.

Query those cleanup rows directly, plus the newest event needed to advance the cursor. Cleanup ordering, deduplication, and retry behavior stay the same without making restart cost grow with unrelated event payloads.

Across three fresh copies of the same snapshot, the database-dependent phase from migrations complete through projection bootstrap fell from a 971 ms median to 512 ms (47.3%). The full launch remained noisy because provider startup runs in the same window, so this measurement isolates the repeated database work changed here.

| Projection bootstrap | Before | After | Change |
| --- | ---: | ---: | ---: |
| Run 1 | 1,018 ms | 512 ms | 49.7% faster |
| Run 2 | 780 ms | 535 ms | 31.4% faster |
| Run 3 | 971 ms | 411 ms | 57.7% faster |
| **Median** | **971 ms** | **512 ms** | **47.3% faster** |

Validation: all 34 legacy projection pipeline tests passed before retargeting. On the Orchestrator v2 base, server typecheck passes and targeted lint and formatting pass.

Model: GPT-5.6 Sol. Harness: Codex in T3 Code.


juliusmarminge (changes requested):
Retarget against v2 pls

kridaydave:
@extoci both of these are sitting on the same request from juliusmarminge: "Retarget against v2 pls." They have not been actioned since the review came in.

Both branches are already based on `t3code/codex-turn-mapping`, so I am not sure which v2 branch you mean. If it is a newer branch than that, give me the name and I can rebase both onto it. If it is codex-turn-mapping itself, then the retarget is already done and these just need the review findings addressed instead.

Two notes from the reviews while you are here. #12847 has a one-time migration that removes four redundant SQLite indexes, and juliusmarminge asked that no more migrations land on main because the numbers conflict. Worth keeping that in mind if this one moves forward. #12846 replaces part of the event-replay path, and Macroscope flagged it as not approvable on that ground alone, so it likely needs a different shape before it is ready.

juliusmarminge:
Rebased onto `t3code/codex-turn-mapping` at `9d6feae1ff`. The commit applied cleanly with no conflicts.

One follow-up commit on top (`b027642515`): added `apps/server/src/orchestration/Layers/ProjectionPipeline.test.ts` (v2 has no pipeline test file yet; it uses the same `OrchestrationProjectionPipelineLive` + in-memory SQLite layer shape as main's) with a single focused case for the narrowed replay query:

- the cleanup cursor advances to `MAX(sequence)` when the latest event is a plain `thread.archived`, not a revert or delete;
- two `thread.reverted` events for one thread collapse into one prune;
- a `thread.deleted` whose cleanup cannot remove its files leaves the cursor where it was, and the next bootstrap after the files are fixable moves it forward.

The test also passes against the v2 baseline, so it pins behavior rather than the implementation.

Verified in `apps/server`: `vp test run src/orchestration/Layers/ProjectionPipeline.test.ts` (1 passed) and `vpr typecheck` (clean).

Nothing else changed; the production diff is as you wrote it.

PR #12847, opened 2026-09-21
perf(server): stop maintaining redundant projection indexes

extoci:
Four projection indexes are strict prefixes of later ordering indexes. SQLite can use the longer indexes for the same lookups, but every projection write still maintained both B-trees.

Drop the redundant indexes in migration 55, after Orchestrator v2 migration 54, and keep the covering indexes that serve the existing read plans. The migration is safe to rerun and a focused test verifies both sides of that contract.

The benchmark used a 1.25 GB snapshot of a real T3 database and representative activity rows with approximately 3.9 KB payloads.

| Benchmark | Before | After | Change |
| --- | ---: | ---: | ---: |
| Insert 10,000 activity rows | 580.8 ms | 520.6 ms | 10.4% faster |
| Read 1,000 activity rows | 14.3 ms | 13.8 ms | No regression |
| Database storage | — | 5.48 MB reclaimed | Smaller |

The representative read kept the same covering-index query plan after the migration.

Validation: the focused migration test runs through Orchestrator v2 migration 54 and the cleanup migration 55. The v2 server typecheck passes, and targeted lint and formatting pass. The copied-database benchmark exercised the same index drops before retargeting.

Model: GPT-5.6 Sol. Harness: Codex in T3 Code.


juliusmarminge (changes requested):
Retarget against v2 pls. Would prefer not introducing more migrations on main cause numbers conflicts

PR #12851, opened 2026-09-21
fix(web): point V2LifecycleRow's AgentElapsed import at its new module

saphid:
## What Changed

One-line import fix: `V2LifecycleRow.tsx` imports `AgentElapsed` from `../AgentsPanel`, but #12835 moved the component to `chat/AgentElapsed.tsx`. #12842, which added the row, merged after the move and kept the old path.

## Why

The base branch tip currently fails the `Check` typecheck (`TS2307: Cannot find module '../AgentsPanel'`) and the `MessagesTimeline.test.tsx` suite for every open branch, including its own commits.

## Verification

- `pnpm typecheck` (web): zero `error TS` at this commit; the same checkout without the fix reports the `TS2307` above.
- `vp test run apps/web/src/components/chat/MessagesTimeline.test.tsx`: 66 passed.


saphid:
Closing: the base branch was force-pushed with the same repair (the stale `../AgentsPanel` import is gone at the new tip), so this is no longer needed. Keeping the report: the tip between #12835 and this rewrite failed `Check` and `Test` for every open branch.

PR #9638, opened 2026-09-04
feat(web only): find in chat with mod+f

imabdulazeez:
## What Changed

Adds a find bar over the open thread.

- New `chat.find` keybinding command, default `mod+f` with `when: "!terminalFocus"`, plus a "Find in chat" command palette action.
- Matches are counted from the timeline's row model, so messages that the virtualized list has not mounted are still found.
- Matches in mounted rows are painted with the CSS Custom Highlight API. No markdown is rewritten and no DOM nodes are wrapped.
- Enter and Shift+Enter step through matches and scroll each one into view. Esc closes.
- `Aa` and `ab` toggles for match case and whole word. One shared pattern feeds both the counter and the highlighter, so counts and painted ranges always agree.
- Selected text becomes the query when the bar opens, as in browser find.
- The pull requests page now resolves the same `chat.find` command for its list search instead of matching Mod+F by hand.

Composed from the shared UI primitives: `InputGroup`, `InputGroupAddon`, `InputGroupInput`, and `InputGroupText` for the field, `Button` (`ghost-muted`, `icon-xs`) for the controls, `Toggle` (`ghost`, `xs`) for the options, and the shared keybinding registry via `resolveShortcutCommand` for the shortcut. No new surfaces or variants are introduced.

Supersedes #5562 with a smaller footprint, and picks up where #4599 left off. Resolves the request in discussion #6709 (formerly #1486).

## Why

Long threads have no way to search the conversation. The timeline is virtualized, so the browser's native find only sees mounted rows, and the desktop shell binds no find-in-page at all. Users asked for Codex-style Cmd+F in #1486 and have kept +1ing it since.

Searching the row model rather than the DOM is what makes off-screen matches work. Painting with the Highlight API avoids the code-block corruption that earlier attempts hit when they injected markup into markdown.

### Relation to orchestration V2 (#2829)

Earlier find PRs (#3539, #4599, #3779) were closed ahead of the orchestration V2 rewrite. This PR is web-only: it touches no orchestration, adapter, or provider code, and its single server change is one assertion in `keybindings.test.ts`. Replaying it on the current #2829 head applies cleanly apart from two adjacent-line conflicts in `CommandPalette.tsx` and `keybindings.test.ts`, so it should survive that merge with a trivial rebase.

## UI Changes

image

https://github.com/user-attachments/assets/33f5045b-a490-4a59-8afe-19727b089c8d

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
[...]

PR #3779, opened 2026-07-07
feat(web): add find-in-thread search with Cmd+F

davidmashburn:
## Summary

- Restore thread-local find (Cmd+F) in the chat view with highlighted matches
- Add next/previous navigation and scroll-to-active-match in the message timeline
- Highlight search terms in markdown-rendered assistant/user content

## Test plan

- [ ] Open a thread with several messages, press Cmd+F (Ctrl+F on non-Mac)
- [ ] Search for a term that appears in multiple messages; verify match count and highlights
- [ ] Use next/previous controls; verify active match scrolls into view
- [ ] Close find bar with Escape; verify highlights clear
- [ ] Confirm proposed plan cards and markdown code blocks highlight correctly

Made with [Cursor](https://cursor.com)

> [!NOTE]
> ### Add find-in-thread search triggered by Cmd+F in chat view
> - Adds a `FindInThread` UI bar that opens with Cmd/Ctrl+F and closes with Escape, with next/previous match navigation via buttons or Enter/Shift+Enter.
> - Introduces [`searchHighlight.ts`](https://github.com/pingdotgg/t3code/pull/3779/files#diff-91a745f132aa98bedfab018483529d82f52c3c88b2c378f15110414c486794fb) with `findAllMatches` and `highlightHtml` utilities for case-insensitive substring detection and `<mark>` tag injection.
> - Propagates `findQuery` and match state through `MessagesTimeline` context to highlight matches in assistant messages, user messages, and proposed plan cards via `ChatMarkdown`.
> - Extends the markdown sanitize schema in [`ChatMarkdown.tsx`](https://github.com/pingdotgg/t3code/pull/3779/files#diff-70b5bb003244414202733c39403081a84415e222927a1ef415599958115c0c05) to allow `<mark>` tags, and adds CSS in [`index.css`](https://github.com/pingdotgg/t3code/pull/3779/files#diff-70ad657a634ce55957729ff8d1c93f697cc7e3ca2a5cb5fe6b00c90d7d515fcd) to style highlighted and active matches with scroll-margin support.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 3175584. 6 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

PR #3539, opened 2026-06-24
feat(web): add Ctrl/Cmd+F find-in-chat

hendrillara:
## What

Adds a browser-style **Ctrl/Cmd+F find-in-chat** to the conversation view. The shortcut opens a floating find bar that searches the current thread — user/assistant messages, tool/work entries, and proposed plans, including collapsed content — shows a live `N/total` counter, navigates with Enter/Shift+Enter (or ↑/↓), and highlights matches.

## Why

The transcript is virtualized (`LegendList`), so off-screen messages aren't in the DOM and native browser/Electron find misses most of them.

## How

- **Search the data model, not the DOM.** `chatSearch.ts` projects each timeline entry to plain text (markdown → text via the same remark pipeline `ChatMarkdown` uses) and builds an ordered match list. The counter and the highlight offsets share this one text basis, so the count stays honest.
- **Reveal + scroll.** Navigating expands a match's turn-fold / work-group / collapsed-long-user-message if hidden, then `scrollToItem`s it into view. Streaming row updates don't re-scroll (the reveal effect keys on the active match and reads rows via refs).
- **Highlight via the CSS Custom Highlight API** (`CSS.highlights` + `Highlight` ranges over a `TreeWalker` of content-only nodes) — no `<mark>` injection, no DOM mutation of message content. A rAF-coalesced `MutationObserver` re-applies as virtualized rows materialize and as a message streams. Where projected text and rendered DOM diverge (raw HTML, custom chips), it falls back to revealing + flashing the row.
- **Keybinding** `find.toggle` (`mod+f`, gated `!terminalFocus`) registered in `@t3tools/contracts` + `@t3tools/shared`. New translucent tokens `--find-match` / `--find-match-current` (light + dark).

## Tests

Pure logic is unit-tested (`vite-plus/test`): projection, match-building/ordering/case-fold (incl. length-changing folds like ß), active-match reconcile, row-location across grouped work entries, offset→range mapping, and keybinding gating. Repo-wide: typecheck clean, `vp check` 0 errors, all suites green (web/contracts/shared/server).

## Not yet verified / scope

Visual behavior (highlight painting, scroll landing, fold/work-group/long-message reveal, streaming stability) has **not been exercised in a running GUI** — please try it manually. v1 leaves out regex/whole-word, terminal/preview search, and cross-thread search. In a multi-field tool entry where the same term appears in more than one field, the "current" highlight may land on a different occurrence within that row (the counter stays correct).

[...]

PR #12959, opened 2026-09-21
feat(web): find text in the current thread with mod+f

rodrigohpalmeirim:
## What Changed

`mod+f`, or **Find in thread** in the command palette, opens a find bar over the open conversation on web and desktop.

- Matches are counted from the loaded user and assistant messages and proposed plans, so text in folded turns and in rows the virtualized list has not mounted still counts.
- Highlights are painted with the CSS Custom Highlight API onto the mounted rows, reusing the text walker the assistant citations already use. No DOM nodes are wrapped and no markdown is rewritten.
- Enter and Shift+Enter step through matches with an n/total counter. Stepping unfolds the match's turn, pins its row, and scrolls the hit into view. Escape closes and clears.
- The bar covers the turns loaded in the view. When older turns exist it offers **Load earlier**, the same paging the timeline header uses.
- Thinking and tool output are not searched. Thinking is grouped with tool activity behind its own disclosure, so counting it would point at nothing visible. Same scope as the thread search in the palette.
- Text selected in the timeline prefills the query.

Wiring: one new `chat.find` command in contracts, a default `mod+f` binding gated on `!terminalFocus` so a focused terminal keeps the chord, a palette action, and a small store so the shortcut, palette and bar share open state. The palette's close handler now leaves focus alone when the action it ran already moved it, the same guard the composer menus use.

No server, RPC or provider changes. Mobile is unchanged.

## Why

There is no way to find text inside an open thread. `mod+f` does nothing in the desktop app, and native browser find only sees the rows currently mounted, so a long agent thread means scrolling and eyeballing. Tracker: #6709.

Earlier attempts (#1501, #3539, #3779, #4599, #7434, #9638, #9660, #10439, #11526) were closed for size or for touching the orchestration layer ahead of V2. This one is deliberately client-only and self-contained, so it does not conflict with that rewrite. Whole-thread search can move server-side later without changing the UI.

## UI Changes

Screenshot 2026-09-21 232209

https://github.com/user-attachments/assets/57d16e20-d02d-4f01-8ab0-b9bad7d9f599

## Verification

- `vp test run` in `apps/web` on `ChatFind.logic` and `chatFindHighlight`: 15 new cases cover pattern escaping and whitespace, occurrence counting across messages and plans, exclusion of thinking and system messages, active-match retention when history prepends, wrap-around stepping, and range mapping across inline markup and block boundaries. Neighbouring suites pass: assistant text selection, web keybindings, keybinding settings, command palette logic, contracts keybindings, mobile hardware keyboard commands.
- `tsc --noEmit` is clean in `apps/web`, `apps/server`, `apps/desktop`, `apps/mobile`, `packages/contracts` and `packages/shared`. Lint and format are clean on all touched files.
- Verified in the browser on web: shortcut, palette entry, counter, stepping into folded turns, load earlier on a 36-turn thread, Escape, and focus handling.

## Checklist

Implemented by Claude Fable 5.1 through the Claude Code harness in T3 Code.

## Summary by CodeRabbit

- **New Features**
  - Added find-in-thread search across chat messages and proposed plans.
  - Highlight matches with result counts, active-match scrolling, and previous/next navigation.
  - Open search with Cmd/Ctrl+F or the command palette; use Enter/Shift+Enter to navigate and Escape to close.
  - Selected text can prefill the query, and earlier history can load when needed.
  - Search excludes thinking and tool output and covers currently loaded thread content.
  - Search results remain visible when matching collapsed content is revealed.

- **Documentation**
  - Added guidance for using thread search.


apps/web/src/components/chat/useChatFind.ts
```diff
@@ -0,0 +1,207 @@
...
+  query: string;
+  entries: ReadonlyArray<TimelineEntry>;
+  rows: ReadonlyArray<MessagesTimelineRow>;
+  listRef: RefObject<LegendListRef | null>;
+  viewport: HTMLElement | null;
+  onExpandTurn: (turnId: TurnId) => void;
+  onManualNavigation: () => void;
+}) {
+  const pattern = useMemo(() => (enabled ? buildChatFindPattern(query) : null), [enabled, query]);
+  const matches = useMemo(() => collectChatFindMatches(entries, pattern), [entries, pattern]);
```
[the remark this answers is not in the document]
rodrigohpalmeirim: Fixed in 4bbf95535. Matches are now counted from a markdown-stripped copy of the source: emphasis delimiters, link and image targets, code fences, headings and list markers are dropped while their text is kept. A query like `**` no longer counts, and `bold` maps to the rendered word. Exact parity with the DOM is not attainable without rendering; the remaining gap is non-content chrome such as a code block's language label.

apps/web/src/components/chat/ChatFind.logic.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Thinking is grouped with tool activity behind its own disclosure and system
+ * messages never render, so counting them would point at nothing.
+ */
+export function chatFindEntrySource(
+  entry: TimelineEntry,
+): { text: string; turnId: TurnId | null } | null {
+  switch (entry.kind) {
+    case "message":
+      if (entry.message.role !== "user" && entry.message.role !== "assistant") return null;
+      return { text: entry.message.text, turnId: entry.message.turnId };
```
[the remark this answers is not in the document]
rodrigohpalmeirim: Fixed in 4bbf95535. The row context now carries the entry that holds the active match, and the collapsed user message body stays open while the match is inside it, so the highlight is visible and the reveal scroll can reach it.

apps/web/src/components/chat/ChatFind.logic.ts
```diff
@@ -0,0 +1,107 @@
...
+    case "proposed-plan":
+      return { text: entry.proposedPlan.planMarkdown, turnId: entry.proposedPlan.turnId };
```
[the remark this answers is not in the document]
rodrigohpalmeirim: Fixed in 4bbf95535 with the same mechanism: while the active match is inside a collapsed plan, the card renders the full plan, and the fine scroll waits for the range to exist before positioning.

apps/web/src/components/chat/useChatFind.ts
```diff
@@ -0,0 +1,207 @@
...
+  // A new query restarts from the first match; the same query keeps its place.
+  const [selection, setSelection] = useState<{ query: string; match: ChatFindMatch } | null>(null);
```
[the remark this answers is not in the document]
rodrigohpalmeirim: Fixed in 4bbf95535. The hook owns the query now, and setting it clears the stepped-to selection, so returning to an earlier query starts at its first match again. No effect involved.

apps/web/src/components/chat/MessagesTimeline.tsx
```diff
@@ -1276,6 +1297,18 @@ export const MessagesTimeline = memo(function MessagesTimeline({
...
             />
           ) : null}
+          {findOpen ? (
+            <ChatFindBar
+              query={findQuery}
+              onQueryChange={setFindQuery}
+              matchCount={chatFind.matches.length}
+              activeIndex={chatFind.activeIndex}
+              onStep={chatFind.step}
+              onClose={hideFind}
```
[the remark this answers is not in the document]
rodrigohpalmeirim: Intentional. Browser find in Chrome, Firefox and Safari keeps the last query when the bar is reopened, and this mirrors that so a repeat search is one keystroke away. Highlights are cleared on close and only return with the bar.

apps/web/src/components/chat/ChatFind.logic.ts
```diff
@@ -0,0 +1,131 @@
...
+ * with what highlighting finds in the DOM: delimiters, link and image targets,
+ * fences, headings and list markers are dropped while their text is kept.
+ */
+export function markdownSearchText(markdown: string): string {
+  return markdown
+    .replace(/^[ \t]*(```|~~~)[^\n]*$/gm, "")
+    .replace(/!\[([^\]]*)\]\([^)]*\)/g, "$1")
+    .replace(/\[([^\]]+)\]\([^)]*\)/g, "$1")
+    .replace(/<((?:https?|mailto):[^>\s]+)>/g, "$1")
+    .replace(/<\/?[a-zA-Z][^>\n]*>/g, "")
```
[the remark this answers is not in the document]
rodrigohpalmeirim: Fixed in 8c1d74ffc. Fenced blocks and inline code are lifted out before the HTML and emphasis rules run and put back verbatim afterwards, so `<div>` written in code stays searchable while real tags outside code are still dropped. Covered by a new test.

apps/web/src/components/chat/ChatFind.logic.ts
```diff
@@ -0,0 +1,139 @@
...
+ * with what highlighting finds in the DOM: delimiters, link and image targets,
+ * fences, headings and list markers are dropped while their text is kept.
+ */
+export function markdownSearchText(markdown: string): string {
+  // Code renders literally, so its contents skip the HTML and emphasis rules.
+  const literals: string[] = [];
+  const keep = (text: string) => `\uE000${literals.push(text) - 1}\uE000`;
+  return markdown
+    .replace(/^[ \t]*(```|~~~)[^\n]*\n([\s\S]*?)\n[ \t]*\1[ \t]*$/gm, (_, _fence, body: string) =>
+      keep(body),
```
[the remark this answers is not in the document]
rodrigohpalmeirim: Fixed in 75e16028f. Fences now open on three or more backticks or tildes and close on the same marker at least as long as the opening run, so four-delimiter fences keep their body literal too. Regression test added for both markers.

apps/web/src/components/ChatView.tsx
```diff
@@ -6855,6 +6856,13 @@ export default function ChatView(props: ChatViewProps) {
         return;
       }
 
+      if (command === "chat.find") {
```
NickOverstreet, line 6859: **The new default `mod+f` -> `chat.find` binding (guarded only by `!terminalFocus`) is claimed by ChatView's window capture-phase keydown handler with preventDefault + stopPropagation no matter where focus is or whether the bar can be seen, so it steals Cmd/Ctrl+F from the Files-panel code editor and opens an invisible, focus-grabbing bar when the chat column is hidden.**

Open a workspace file in the Files panel (@pierre/diffs Editor), click into the code and press Cmd/Ctrl+F: before this PR the editor's search panel opened; now the window-capture handler stops the event before the editor's contentEl listener, and the chat find bar opens and takes focus (only Mod+Alt+F still reaches the editor). With the right panel maximized (the chat column is `w-0` but MessagesTimeline stays mounted), shown as the <=980px modal Sheet, or with a Base UI modal open (outside content is only aria-hidden, not inert), Cmd/Ctrl+F moves focus into a find input the user cannot see; typing searches and scrolls the hidden timeline and turns live-follow off, and on web the browser's native find is gone on those surfaces.

apps/web/src/components/chat/useChatFind.ts
```diff
@@ -0,0 +1,218 @@
...
+      const matchedEntryIds = new Set(state.matches.map((match) => match.entryId));
+      const ranges: Range[] = [];
+      let activeRange: Range | null = null;
+      for (const rowElement of scrollNode.querySelectorAll<HTMLElement>(ROW_SELECTOR)) {
+        const rowId = rowElement.dataset.timelineRowId;
+        if (rowId === undefined || !matchedEntryIds.has(rowId)) continue;
+        const rowRanges = collectChatFindRanges(rowElement, state.pattern);
+        if (state.activeMatch !== null && rowId === state.activeMatch.entryId) {
+          activeRange =
+            rowRanges[Math.min(state.activeMatch.occurrence, rowRanges.length - 1)] ?? null;
```
NickOverstreet: **The active DOM range is `rowRanges[Math.min(occurrence, rowRanges.length - 1)]`, but `rowRanges` come from `readAssistantText` over the whole `[data-timeline-row-id]` element, which includes chrome the source-text count never sees (the sr-only author `<h3>`: "T3 Code" on every assistant row, "You" on every user row, and the plan card's "Plan" badge), so the active match is shifted by one in those rows.**

Assistant reply "I updated the code in two places; the code now compiles." with query "code": the counter says 1/2 but the row yields three ranges [hidden "Code", code#1, code#2]. At 1/2 the active highlight is painted on the visually hidden heading and revealRange targets the row top (a hit deep in the message stays off-screen); at 2/2 code#1 is active; code#2 can never become active. The same happens for "you" in user rows and "plan" in plan cards (reproduced with the verbatim functions over a mocked row). Scoping the walk to the body markers the citation code already uses (`[data-assistant-citation-source]`, `[data-user-message-body]`) avoids it.

apps/web/src/components/chat/chatFindHighlight.ts
```diff
@@ -0,0 +1,70 @@
+import { readAssistantText, type RenderedTextChunk } from "../../lib/assistantTextSelection";
+import { findPatternSpans, type TextSpan } from "./ChatFind.logic";
+
+export const CHAT_FIND_HIGHLIGHT_NAME = "t3-chat-find";
+export const CHAT_FIND_ACTIVE_HIGHLIGHT_NAME = "t3-chat-find-active";
+
+export function supportsChatFindHighlights(): boolean {
```
NickOverstreet: **`CHAT_FIND_HIGHLIGHT_NAME`, `CHAT_FIND_ACTIVE_HIGHLIGHT_NAME` and `supportsChatFindHighlights` are exported but only used inside this file, so CI's `knip --exports` step ("Check unused code") will fail.**

`vp run knip:check` runs `knip --workspace apps/web ... --exports` and knip.jsonc does not set `ignoreExportsUsedInFile`, so knip 6.34.0 reports "Unused exports (3)" at chatFindHighlight.ts:4, :5 and :7 and exits 1 (reproduced on a fixture with this PR's files). That fails the job and skips the later check, typecheck and build steps; CI has not run yet because the fork PR is awaiting approval. Dropping `export` from the three declarations fixes it.

```suggestion
const CHAT_FIND_HIGHLIGHT_NAME = "t3-chat-find";
const CHAT_FIND_ACTIVE_HIGHLIGHT_NAME = "t3-chat-find-active";

function supportsChatFindHighlights(): boolean {
```

apps/web/src/components/chat/MessagesTimeline.tsx
```diff
@@ -956,8 +961,28 @@ export const MessagesTimeline = memo(function MessagesTimeline({
     onExpandTurn: expandCitedTurn,
     onManualNavigation,
   });
+  const findOpen = useChatFindStore((store) => store.open);
+  const findFocusRequestId = useChatFindStore((store) => store.focusRequestId);
+  const hideFind = useChatFindStore((store) => store.hide);
+  // Find is scoped to one thread; switching threads closes it.
+  const findThreadKeyRef = useRef(listIdentityKey);
```
NickOverstreet: **Find's `open` flag lives in a module-global store that is cleared only when an already-mounted MessagesTimeline sees `listIdentityKey` change (the ref starts at the current key on mount and nothing resets it on unmount), while a freshly mounted ChatFindBar treats its own mount as a focus request, so find reopens on threads where the user never opened it and can steal focus from the composer.**

Press Cmd/Ctrl+F in a new, empty draft: the key is swallowed and nothing shows (the empty timeline returns before rendering the bar), but `open` is now true. Leaving find open and starting a new thread gives the same state (draft ChatViews are keyed by draftId and remount). Type the first prompt and press Enter: rows appear, ChatFindBar mounts and calls focus() and select() on its input, and the composer autofocus does not re-run (the thread id is unchanged), so the next keystrokes go into the find box and Enter steps through matches instead of sending. Returning from Settings or Pull requests also brings the bar back on whatever thread opens.

apps/web/src/components/chat/MessagesTimeline.tsx
```diff
@@ -956,8 +961,28 @@ export const MessagesTimeline = memo(function MessagesTimeline({
...
+  useEffect(() => {
```
NickOverstreet, line 982: **On an in-place thread switch, find is closed only by this passive effect's `hideFind()`, so in the same effect flush `useChatFind`'s reveal effect still runs with `enabled: true`, the old query and the new thread's entries and rows, calling `onManualNavigation()` (which cancels the new thread's position restore and turns live-follow off) and scrolling to, or unfolding the turn of, the stale query's first match.**

Find is open with "error" in thread A; click server thread B (same ChatView and MessagesTimeline instance) whose history contains "error". B opens scrolled to its oldest "error" row instead of its end or remembered position (the find scrollToIndex supersedes the restore scroll), handleScroll then saves that spot as B's remembered position, live-follow can stay off, and the bar is already gone so nothing explains the jump. Resetting find in the existing render-phase `listIdentityKey` block, or scoping `open` to a thread key, prevents the stale render.

apps/web/src/components/chat/ChatFind.logic.ts
```diff
@@ -0,0 +1,145 @@
...
+  return (
+    markdown
+      // A fence closes on the same marker at least as long as the opening one.
+      .replace(/^[ \t]*(`{3,})[^\n]*\n([\s\S]*?)\n[ \t]*\1`*[ \t]*$/gm, (_, _fence, body: string) =>
+        keep(body),
+      )
+      .replace(/^[ \t]*(~{3,})[^\n]*\n([\s\S]*?)\n[ \t]*\1~*[ \t]*$/gm, (_, _fence, body: string) =>
+        keep(body),
+      )
+      .replace(/(`+)([^`]+?)\1/g, (_, _ticks, body: string) => keep(body))
```
NickOverstreet: **`markdownSearchText` keeps inline-code and link text verbatim, but ChatMarkdown renders path-like inline code and file links as chips whose only text is the basename (plus `" · L<line>"`), and mention/context chips (buttons), `<details>` summaries (buttons; closed bodies unmounted) and image alt text contribute no searchable DOM text, so counted hits have no range and the `Math.min` clamp stacks them onto another hit.**

In a project thread the assistant writes "I changed `apps/web/src/components/ChatView.tsx` and the components list." and the user searches "components": the counter says 2 but the row has one DOM hit (the chip shows "ChatView.tsx"), so 1/2 and 2/2 highlight the same word and Enter looks stuck. Searching a full path such as "apps/server/src/wsServer.ts" against "`apps/server/src/wsServer.ts:120`" shows 1/1 with no highlight at all. Coding agents cite paths this way in most replies. The new test at ChatFind.logic.test.ts:84 also locks image alt text into the count.

apps/web/src/components/chat/ChatFind.logic.ts
```diff
@@ -0,0 +1,145 @@
...
+ * Thinking is grouped with tool activity behind its own disclosure and system
+ * messages never render, so counting them would point at nothing.
+ */
+export function chatFindEntrySource(
+  entry: TimelineEntry,
+): { text: string; turnId: TurnId | null } | null {
+  switch (entry.kind) {
+    case "message":
+      if (entry.message.role !== "user" && entry.message.role !== "assistant") return null;
+      return { text: markdownSearchText(entry.message.text), turnId: entry.message.turnId };
```
NickOverstreet: **Entries are counted from raw `message.text`, but user rows render `resolveUserMessageContext(message).text` through ChatMarkdown with `parseRawHtml={false}` (raw HTML shown literally; legacy `<terminal_context>`/`<element_context>` blocks lifted into chips) and assistant rows render Codex `:codex-file-citation{...}` directives as basename links, so visible text reports "No results" while hidden text is counted.**

User message `Why does <Suspense> not catch this error?` with query "suspense" shows "No results" (the tag rule at line 67 deletes it) even though it is on screen; the same goes for `Promise<void>` or `Array<T>`. A legacy user message ending in a terminal_context block counts words from the hidden terminal output ("boom" gives 1/1 with nothing highlighted). In a Codex thread `:codex-file-citation{path="/repo/apps/web/src/lib/bar.ts"}` renders as "bar.ts", yet "src" counts 3 hits with 0 in the DOM. Normalizing through `resolveUserMessageContext` and `renderCodexDirectivesForCopy` (existing helpers) and not stripping tags for user rows would line them up.

apps/web/src/components/chat/useChatFind.ts
```diff
@@ -0,0 +1,218 @@
...
+          const pending = pendingRevealRef.current;
+          if (pending?.rowId === rowId) {
+            const listState = list.getState();
+            const index = listState.indexByKey(rowId);
+            const settled = index !== undefined && listState.sizeAtIndex(index) > 0;
```
NickOverstreet, line 174: **The pending reveal counts a row as settled once `sizeAtIndex(index) > 0` and any active range exists, then scrolls once and clears `pendingRevealRef` with no retry and no cancellation on user input; right after `revealed` expands a plan or long user message that size is still the last measured collapsed height, and a code block still showing its aria-hidden Suspense fallback yields no range or a clamped neighbour.**

A thread ends with a long, collapsed proposed plan; search a word about 40 lines into it: the plan expands in the same commit, but the rAF paint runs before LegendList's ResizeObserver re-measures, so `scrollToOffset` is clamped to the stale total size and the match stays below the viewport with no retry. For a hit inside a code block whose Shiki grammar has not loaded yet, the loop either settles on a different hit in that row (never revealing the real one) or keeps retrying for up to 60 frames and then yanks the viewport back after the user has already scrolled away.

apps/web/src/components/chat/useChatFind.ts
```diff
@@ -0,0 +1,218 @@
...
+    step,
+    alwaysRender,
+  };
+}
+
+function revealRange(range: Range, scrollNode: HTMLElement, list: LegendListRef) {
+  const rect = range.getBoundingClientRect();
+  const scrollRect = scrollNode.getBoundingClientRect();
+  if (rect.height <= 0 || scrollNode.clientHeight <= 0) return;
+  const visible =
```
NickOverstreet: **`revealRange` treats the active range as visible when it sits 48px inside the scroll node's rect vertically, ignoring the composer overlay that covers the bottom of that node and any horizontal clipping by nested scrollers, so the active match can stay hidden while the counter advances.**

With the composer expanded (a glass overlay of roughly 140-190px positioned over the list), stepping to a hit deep in a long message lands it near y=700 of an 800px viewport; `rect.bottom <= scrollRect.bottom - 48` passes, revealRange returns, and the active highlight sits under the composer. A hit in the last column of a wide markdown table (tables scroll horizontally inside a ScrollArea) or at the end of an unwrapped code line is never scrolled into horizontal view. Other scroll code already subtracts the composer inset (timelineScrollAnchoring.ts:63, ChatView.tsx:5370).

apps/web/src/components/chat/ChatFind.logic.ts
```diff
@@ -0,0 +1,145 @@
...
+    case "proposed-plan":
+      return {
+        text: markdownSearchText(entry.proposedPlan.planMarkdown),
```
NickOverstreet: **Plans are counted over raw `planMarkdown`, but ProposedPlanCard renders a "Plan" badge, a title from `proposedPlanTitle(...) ?? "Proposed plan"` and a body from `stripDisplayedPlanMarkdown` (which drops the leading title and a following "Summary" heading), so plan rows have phantom and extra hits and the occurrence-to-range mapping is off.**

Plan "# Fix login / ## Summary / Do X" with query "summary" shows 1/1 with no DOM range, so nothing is highlighted and the reveal retries for 60 frames, then gives up. This PR's own fixture "# Plan / 1. Reproduce the login bug ..." with query "plan" counts 1 but renders 2 (badge and title), so the active highlight lands on the badge. A plan without a heading counts "proposed" once but also renders it in the fallback title.

apps/web/src/components/chat/useChatFind.ts
```diff
@@ -0,0 +1,218 @@
...
+    scheduleRef.current();
+  }, [pattern, matches, activeMatch]);
+
+  const step = useCallback(
+    (direction: 1 | -1) => {
+      const next = stepChatFindIndex(activeIndex, matches.length, direction);
+      const match = next >= 0 ? matches[next] : undefined;
+      if (!match) return;
+      // Re-reveal even when the match is unchanged, as with a single result.
+      navigatedKeyRef.current = null;
```
NickOverstreet: **Re-revealing depends on refs and `targetKey` changes: `step()` nulls `navigatedKeyRef` but nothing re-runs the effect when the match is unchanged, `expandedKeyRef` is never cleared by `step()`, and until the user steps the active match is simply `matches[0]`, so navigation silently fails or fires later without a user action.**

(a) With a single result, scroll away and press Enter: nothing happens; the next unrelated update (a streamed chunk, a disclosure toggle) re-runs the effect with the nulled ref, yanks the view back and turns live-follow off. (b) Match A is in folded turn T: find unfolds T, you step to B, T re-folds (fold toggle or a new turn starting), Shift+Enter back to A: `expandedKeyRef.current === targetKey`, so T is never re-expanded and nothing is revealed while the counter says 1/2. (c) Unstepped at 1/3 in turn 30, click the bar's "Load earlier": `matches[0]` becomes a hit in the prepended page and the view jumps there, contrary to `resolveActiveMatchIndex`'s own doc comment.

apps/web/src/components/chat/ChatFind.logic.ts
```diff
@@ -0,0 +1,145 @@
...
+      .replace(/<((?:https?|mailto):[^>\s]+)>/g, "$1")
+      .replace(/<\/?[a-zA-Z][^>\n]*>/g, "")
+      .replace(/^[ \t]{0,3}#{1,6}[ \t]+/gm, "")
+      .replace(/^[ \t]*>[ \t]?/gm, "")
+      .replace(/^[ \t]*(?:[-*+]|\d+[.)])[ \t]+(?:\[[ xX]\][ \t]+)?/gm, "")
+      .replace(/^[ \t]*\|?[ \t]*:?-{3,}:?[ \t]*(\|[ \t]*:?-{3,}:?[ \t]*)*\|?[ \t]*$/gm, "")
+      .replace(/`+/g, "")
+      .replace(/(\*\*|__|~~)(?=\S)([\s\S]*?\S)\1/g, "$2")
+      .replace(/(^|[^\w*])[*_](?=\S)([^*_\n]*?\S)[*_](?![\w*])/g, "$1$2")
+      .replace(/\uE000(\d+)\uE000/g, (_, index: string) => literals[Number(index)] ?? "")
```
NickOverstreet: **`markdownSearchText` restores its U+E000 placeholders in one non-recursive pass, so a placeholder captured inside another literal is never expanded, and it never decodes backslash escapes or character references or strips table-cell pipes, so whole code blocks and ordinary visible text drop out of the count.**

A ~~~ fence wrapping a ``` fence (the backtick pass at line 56 runs before the tilde pass at line 59) leaves a raw placeholder, so "npm install" inside it shows "No results"; a stray backtick before a fence that pairs with a later inline code span makes "TypeError" in that fence count 0. Likewise `**MAX\_RETRIES**` (renders MAX_RETRIES), `&lt;div&gt;`, and "alpha beta" across table cells all count 0 while visible, and "|" counts 6 phantom hits per table row (checked against the real react-markdown pipeline).

apps/web/src/components/chat/MessagesTimeline.tsx
```diff
@@ -4009,11 +4050,14 @@ const CollapsibleUserMessageBody = memo(function CollapsibleUserMessageBody(prop
...
   footer?: ReactNode;
+  /** A find match landed inside; open the clipped body so it can be seen. */
+  revealed?: boolean;
 }) {
   const [expanded, setExpanded] = useState(false);
   const hasVisibleBody = props.text.trim().length > 0;
   const canCollapse = hasVisibleBody && shouldCollapseUserMessage(props.text);
-  const isCollapsed = canCollapse && !expanded;
+  // The find bar holds the body open while its active match is inside.
+  const isCollapsed = canCollapse && !expanded && props.revealed !== true;
```
NickOverstreet: **`revealed` is a temporary override OR-ed into the collapse state instead of setting the component's own `expanded` state, so while find holds a long user message or plan open its "Show less"/"Collapse plan" button (`setExpanded(isCollapsed)`, i.e. `setExpanded(false)`) does nothing, and as soon as find closes or steps away the body re-collapses and hides the text that was just found.**

Search a word that appears only past the 10-line preview of a long plan: the plan expands and the hit scrolls into view, but the visible "Collapse plan" button is a no-op. Press Escape to read: `chatFindRevealEntryId` becomes null, the card swaps back to the preview so the matched text leaves the DOM, and since the list anchors on the plan itself the viewport can land on unrelated content. The same applies to long user messages here and in ProposedPlanCard.tsx:76/208.

apps/web/src/components/chat/ChatFind.logic.ts
```diff
@@ -0,0 +1,145 @@
...
+      .replace(/!\[([^\]]*)\]\([^)]*\)/g, "$1")
+      .replace(/\[([^\]]+)\]\([^)]*\)/g, "$1")
+      .replace(/<((?:https?|mailto):[^>\s]+)>/g, "$1")
+      .replace(/<\/?[a-zA-Z][^>\n]*>/g, "")
+      .replace(/^[ \t]{0,3}#{1,6}[ \t]+/gm, "")
+      .replace(/^[ \t]*>[ \t]?/gm, "")
+      .replace(/^[ \t]*(?:[-*+]|\d+[.)])[ \t]+(?:\[[ xX]\][ \t]+)?/gm, "")
+      .replace(/^[ \t]*\|?[ \t]*:?-{3,}:?[ \t]*(\|[ \t]*:?-{3,}:?[ \t]*)*\|?[ \t]*$/gm, "")
+      .replace(/`+/g, "")
+      .replace(/(\*\*|__|~~)(?=\S)([\s\S]*?\S)\1/g, "$2")
```
NickOverstreet: **`markdownSearchText` has quadratic rules for unclosed `**`/`__`/`~~` (line 73) and `[` (line 65) openers, and `collectChatFindMatches` re-runs it over every loaded message on every keystroke and every `entries` change with no per-message cache, so one large message stalls the main thread on each update.**

A user message of 64K-120K chars of unfenced C (`void f(char **argv, int **grid)`, sendable via paste-as-text) costs 150-560 ms per find keystroke and per streamed or activity update while a query is active; colored `git log` output (unclosed ESC[ sequences) costs about 320 ms at 120K. Even normal text costs about 13-19 ms per update at 150 loaded turns. A WeakMap cache keyed by the message/plan object (about 0.2 ms) plus line-bounded rules removes both.

apps/web/src/components/chat/ChatFindBar.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const hasQuery = query.trim().length > 0;
+
+  return (
+    <div
+      role="search"
+      aria-label="Find in thread"
+      className="surface-glass absolute top-2 right-4 z-30 flex items-center gap-0.5 rounded-lg border border-border/60 p-1 shadow-sm"
+      onKeyDown={(event) => {
+        if (event.nativeEvent.isComposing) return;
```
NickOverstreet: **The IME guard checks only `event.nativeEvent.isComposing`, not the `event.keyCode === 229` check the rest of the repo uses, so in Safari 26.x and earlier the Enter that commits an IME candidate (delivered after compositionend with `isComposing` false) steps to the next match.**

On app.t3.codes or `npx t3` web in macOS Safari 26.x, a Japanese or Chinese IME user types a query and presses Enter to commit the candidate: `onStep(1)` runs, so the counter shows 2/N instead of 1/N and the view scrolls on every commit. Sidebar.tsx:1303/2944, ThemeSearchSection.tsx:277 and others use `if (event.nativeEvent.isComposing || event.keyCode === 229) return;` for exactly this (#2817, #6281, #10262).

```suggestion
        if (event.nativeEvent.isComposing || event.keyCode === 229) return;
```

apps/web/src/components/ChatView.tsx
rodrigohpalmeirim: Fixed in ac2b0eb. `chat.find` now returns without `preventDefault` when the event target is inside the file editor (`.file-preview-virtualizer`) or a `[role=dialog]`, and when the right panel is maximized, so the editor's search panel or the browser's native find runs instead. The `!terminalFocus` when-clause is unchanged.

apps/web/src/components/chat/useChatFind.ts
rodrigohpalmeirim: Fixed in ac2b0eb. Ranges are collected from the row bodies only: `[data-user-message-body]`, `[data-assistant-citation-source]`, and a `data-chat-find-body` marker on the plan card's title and body. The hidden author heading and the Plan badge no longer enter the stream, so occurrence N in the source maps to range N in the row.

apps/web/src/components/chat/chatFindHighlight.ts
rodrigohpalmeirim: Fixed in ac2b0eb. The three names are module-private now. `knip --workspace apps/web --exports` passes locally.

apps/web/src/components/chat/MessagesTimeline.tsx
rodrigohpalmeirim: Fixed in ac2b0eb. MessagesTimeline records which focus request it consumed and binds find to the thread key that had rows on screen at that moment. A request that predates the mount is consumed without opening, and a request that arrives while nothing is rendered (empty draft) is dropped and the store closed by effect. A draft's first turn, a new-thread remount, or a return from Settings can no longer bring the bar back or take focus from the composer.

apps/web/src/components/chat/MessagesTimeline.tsx
rodrigohpalmeirim: Fixed in ac2b0eb. `findOpen` is derived during render as `store.open && scope.threadKey === listIdentityKey`, so on an in-place switch the hook is disabled in the very render that shows the new thread; the `hide()` is only cleanup. No stale `scrollToIndex`, turn expand, or `onManualNavigation` reaches thread B.

apps/web/src/components/chat/ChatFind.logic.ts
rodrigohpalmeirim: Fixed in ac2b0eb for the common cases. Inline code and markdown links that resolve to a file (same `resolveInlineCodeFileLinkMeta` / `resolveMarkdownFileLinkMeta`, with the thread cwd) are replaced by the chip label (basename plus ` · L<line>`), images contribute no text, and user context chips are removed. Not replicated: the parent-directory suffix ChatMarkdown adds when two files in one message share a basename, `<details>` summaries, and mention chips. For any residual mismatch the active range now prefers the exact occurrence and only falls back to the last painted hit after the retry window, so stepping still lands in the right row.

apps/web/src/components/chat/ChatFind.logic.ts
rodrigohpalmeirim: Fixed in ac2b0eb. User rows normalize `resolveUserMessageContext(message).text` with tag stripping off, so `<Suspense>` counts and highlights; legacy terminal/element context blocks are lifted into chips and dropped. Assistant text goes through `renderCodexDirectivesForCopy` first, so a Codex citation counts as its basename.

apps/web/src/components/chat/useChatFind.ts
rodrigohpalmeirim: Fixed in ac2b0eb. The reveal requires the exact range for the occurrence rather than a clamped neighbour, `revealRange` reports when the wanted offset was clamped by an unmeasured row so the paint loop retries next frame, and `wheel`/`touchmove`/`pointerdown` on the scroll node cancel a pending reveal so it cannot yank the viewport after the user has moved on.

apps/web/src/components/chat/useChatFind.ts
rodrigohpalmeirim: Fixed in ac2b0eb for the composer: the visibility check and target offset subtract the composer inset (`contentInsetEndAdjustment`, the same value the timeline footer uses). Horizontal clipping inside a table's ScrollArea or an unwrapped code line is not handled; the highlight is painted but not scrolled horizontally. Left as is for now; happy to follow up if maintainers want it.

apps/web/src/components/chat/ChatFind.logic.ts
rodrigohpalmeirim: Fixed in ac2b0eb. Plan text is now `proposedPlanTitle(plan) ?? "Proposed plan"` followed by `stripDisplayedPlanMarkdown(plan)` normalized, which is what the card renders; the badge sits outside the body markers so it is not painted.

apps/web/src/components/chat/useChatFind.ts
rodrigohpalmeirim: Fixed in ac2b0eb. Each step bumps a reveal request that is part of the target key, so Enter on a single result re-reveals and a turn that folded again is re-expanded. The first match is adopted as the selection during render, so Load earlier cannot move an unstepped active match.

apps/web/src/components/chat/ChatFind.logic.ts
rodrigohpalmeirim: Fixed in ac2b0eb. Placeholders are restored until none remain, backslash escapes are kept literal and cannot open markup, character references decode, and pipes in table rows become spaces. The emphasis, link, and image rules are line-bounded now.

apps/web/src/components/chat/MessagesTimeline.tsx
rodrigohpalmeirim: Fixed in ac2b0eb. `revealed` now flips the component's own `expanded` state when it changes, so Collapse works while find is open and the text stays visible after Escape.

apps/web/src/components/chat/ChatFind.logic.ts
rodrigohpalmeirim: Fixed in ac2b0eb. Emphasis, link, and image rules stop at line ends, and normalized text is cached per message/plan object in a WeakMap, invalidated when the text or cwd changes. Locally the 111 KB unfenced-C case went from about 900 ms to about 2 ms per pass and the unclosed-`[` case from about 230 ms to about 1 ms.

apps/web/src/components/chat/ChatFindBar.tsx
rodrigohpalmeirim: Fixed in ac2b0eb using the suggested check.

PR #12963, opened 2026-09-22
feat(web): add first-use hint primitive

extoci:
T3 Code has several contextual features worth teaching once, but each caller currently has to invent its own persistence and presentation.

This adds a compact first-use hint dialog with one imperative entry point. The manager durably claims hint IDs, deduplicates same-session races, queues overlapping hints, and falls back to session-only behavior when browser storage is unavailable. The host is mounted once in the authenticated app shell and is independent from toast infrastructure.

Verification:
- `vp test run src/firstUseHints.test.ts --project unit`
- targeted `vp lint` on changed files (only pre-existing root-route warnings)

Built with gpt-5.6-sol in Codex via T3 Code.

## Summary by CodeRabbit

- **New Features**
  - Added first-use hints in dismissible dialogs with titles, descriptions, and optional primary and secondary actions.
  - Hints are displayed once per identifier and can be dismissed or completed through the available actions.
  - Added support for queuing multiple hints and remembering previously shown hints across sessions.
  - Added automatic first-use hint display within the authenticated application experience.


juliusmarminge:
Demo pls

extoci:
Here’s the primitive running in the real app with placeholder content, so the presentation can be reviewed independently of its first consumer:

First-use hint demo

I also exercised dismissal in browser automation and reloaded; the same hint did not reappear. The placeholder trigger used for this capture is not part of the PR.

PR #12984, opened 2026-09-22
refactor(web): drop className overrides that repeat the base styles

juliusmarminge:
Removes classes on components/ui exports that the component already applies,
and folds three defaults into the base where every consumer set the same
value: SidebarFooter padding, ComboboxListVirtualized sizing, and
MenuShortcut alignment. No rendered change.

Lowers the no-restyle ceiling from 1247 to 1213.

## Summary by CodeRabbit

* **Style**
  * Refined sidebar, dialog, empty-state, combobox, picker, and settings layouts to use more consistent default spacing, sizing, typography, and color styling.
  * Adjusted sidebar footer padding and virtualized list presentation for a more uniform interface.
  * Simplified background and text styling across application views while preserving existing navigation, resizing, scrolling, and content behavior.
  * Standardized dialog sizing and menu shortcut alignment.


juliusmarminge:
Confirmed with tailwind-merge: `p-0` does not override `not-empty:px-1`, so two of the three lists were padded before. Fixed in the follow-up commit by keeping the padding in the base and letting ModelPickerContent opt out explicitly with `not-empty:p-0`, which is what it was doing before.

PR #12986, opened 2026-09-22
refactor(web): give TooltipPopup width and code variants

juliusmarminge:
Every multi-line tooltip set its own max-width and line-height through
className, in eight different combinations, and every path or command
tooltip re-declared the same monospace treatment. Expose width="sm..2xl"
(max-w-64 through max-w-120, wrapping enabled) and variant="code"
(monospace, 11px, breaks anywhere) and migrate the 60 call sites.

Visible effect: wrapped tooltips now share one line-height (leading-snug)
where sites previously picked tight or relaxed, and code tooltips share one
size. Widths map to the nearest step of the scale.

Lowers the no-restyle ceiling from 1206 to 1146.

## Summary by CodeRabbit

* **New Features**
  * Added standardized tooltip width options for more consistent sizing across the app.
  * Added a code-styled tooltip variant for paths, commands, IDs, and other technical content.
  * Long dialog and alert titles now wrap more reliably.

* **Style**
  * Updated tooltips throughout the app to use consistent sizing, wrapping, alignment, and typography.
  * Improved readability of technical tooltip content with standardized monospace styling and wrapping behavior.


juliusmarminge:
Good catch. Fixed in 0c0c925 by putting `wrap-anywhere` on the `DialogTitle`/`AlertDialogTitle` base rather than restoring it per site, since any dialog title can interpolate a path or branch name.

juliusmarminge:

PR #12988, opened 2026-09-22
refactor(web): fold repeated overrides into ui defaults

juliusmarminge:
- MenuItemLabel truncates by default; every label is a single line and the
  four sites that asked for it were the ones long enough to notice.
- MenuCheckboxItem's label column is minmax(0,1fr) so labels can truncate;
  two consumers patched this per site.
- PopoverTitle is text-sm; three of four consumers shrank it from text-lg.
- Empty title and description drop text-foreground / text-sm, which the
  base already applies, and the /78 description tint and mt-2 nudge, so the
  three hero empty states match every other Empty.

Lowers the no-restyle ceiling from 1124 to 1105.

## Summary by CodeRabbit

* **Style**
  * Standardized empty-state and onboarding typography by relying on consistent default styling.
  * Added consistent spacing between empty-state descriptions and preceding content.
  * Updated popover titles to use a smaller, consistent text size.
  * Improved menu label handling and available text space across project, editor, usage, and pull request controls.
  * Refined checkbox and filter menu layouts for more reliable label alignment.


juliusmarminge:
Fixed in the base instead of per site: `EmptyDescription` now also adds `mt-1` when it follows another description.

juliusmarminge:

PR #12996, opened 2026-09-22
refactor(web): add a tone axis to Button

juliusmarminge:
Icon-only buttons that signal state set text-* through className together
with [--control-icon-color:currentColor] so the glyph follows the label
color; twelve sites, each with its own pairing. Expose that as
tone="muted|current|primary|warning|destructive" and migrate them.

Visible effect: the favorite star in provider models uses the warning
token instead of text-yellow-500, and the muted tones share one
muted-foreground value where two sites used /40 and /60 opacity.

Lowers the no-restyle ceiling from 1038 to 1004.

## Summary by CodeRabbit

* **Style**
  * Standardized button tones across sidebars, settings, filters, attachments, and browser controls for a more consistent interface.
  * Improved visual states for active filters, favorited items, inherited settings, and environment indicators.
  * Clarified destructive actions, including provider and theme removal controls, with consistent warning styling.
  * Refined muted and current-color button appearances while preserving existing layout, sizing, and interactions.

* **Tests**
  * Added coverage to verify button tone styling and state-specific color behavior.


juliusmarminge:
All four findings had the same root cause: `tone` was declared before `variant` in the cva map, so tailwind-merge let the variant's text color win. Fixed in e29a6ec by declaring tone last, giving every tone an explicit hover color (so `current` keeps foreground, `primary`/`warning` keep their color on hover, `destructive` hovers to destructive), and adding `button.test.tsx` to pin the order. Verified the merged output for each variant/tone pair.

PR #13002, opened 2026-09-22
chore(lint): exempt CollapsibleTrigger from no-restyle

juliusmarminge:
CollapsibleTrigger is a bare button and a disclosure row is not a Button,
so there is no styled counterpart to render through; all 15 consumers
style it fully and no two share a shape. Its className is the API. Every
other headless trigger has a Button to render, so they stay covered.

Lowers the no-restyle ceiling from 911 to 842.

## Summary by CodeRabbit

- **Chores**
  - Updated styling validation thresholds to reflect the current state of the interface.
  - Expanded the supported styling options for collapsible controls, including layout, color, typography, spacing, shape, effects, and motion.


juliusmarminge:

PR #13009, opened 2026-09-22
refactor(web): render menu and popover triggers through Button

juliusmarminge:
Six headless triggers hand-rolled a full button recipe in className. Two
were the same accent pill, which is Button variant="secondary" size="xs";
four were bare text links, which is InlineButton. Render through those
instead so the trigger inherits focus rings and hover from one place.

Visible effect: the diff-scope pickers use the secondary token (zinc-50)
instead of accent (zinc-100), one step lighter at rest.

Lowers the no-restyle ceiling from 784 to 735.

## Summary by CodeRabbit

- **Style**
  - Updated dropdowns, menus, popovers, and banners to use consistent shared button styling.
  - Preserved existing labels, sizing, layout, and interaction behavior.
  - Refined muted, secondary, underlined, and default button appearances across the interface.
- **Chores**
  - Tightened the styling consistency threshold for ongoing UI restyling.


juliusmarminge:

PR #13020, opened 2026-09-22
refactor(web): use ghost-muted where ghost buttons restyled to muted

juliusmarminge:
Fourteen ghost buttons re-declared text-muted-foreground and
hover:text-foreground through className. That combination is exactly the
ghost-muted variant (same icon tint, same hover surface), so swap them over
and drop the classes. No rendered change.

Lowers the no-restyle ceiling from 1146 to 1124.

Layer 1 of 24 in stack #13044, ordered least to most opinionated. Part of the `components/ui` className cleanup ([audit and guardrail: #12982](https://github.com/pingdotgg/t3code/pull/12982)). Each layer lowers the `shadcn/no-restyle` ceiling; CI fails if the count rises. For every override cluster the question was "is this a real design-system concept, or should the primitive get stricter and the consumers change?"; a prop was only added for the former.

## Summary by CodeRabbit

* **Style**
  * Updated secondary and utility buttons throughout the app with more consistent muted styling.
  * Refined controls for copying file paths and commands, managing tabs and settings, editing projects, dismissing notices, handling comments, expanding content, and opening pull requests.
  * Existing button actions, visibility, accessibility, and interaction behavior remain unchanged.


juliusmarminge:
Right; dropped the hover override. The copy button is a plain ghost-muted now.

PR #13026, opened 2026-09-22
refactor(web): add radius="none" to ScrollArea

juliusmarminge:
Eight scroll areas opted out of the inherited radius with rounded-none
because they sit flush to a panel edge. Expose radius="none" for that.
The onboarding wizard also forced its scrollbars visible; scrollbars reveal
on hover and scroll everywhere else, and now there too.

Lowers the no-restyle ceiling from 792 to 782.

Layer 7 of 24 in stack #13044, ordered least to most opinionated. Part of the `components/ui` className cleanup ([audit and guardrail: #12982](https://github.com/pingdotgg/t3code/pull/12982)). Each layer lowers the `shadcn/no-restyle` ceiling; CI fails if the count rises. For every override cluster the question was "is this a real design-system concept, or should the primitive get stricter and the consumers change?"; a prop was only added for the former.

## Summary by CodeRabbit

* **Style**
  * Improved scroll-area corner styling consistency across chat, files, onboarding, diagnostics, settings, and navigation views.
  * Scroll areas now support inherited or explicitly square corners for more consistent layouts.
  * Restored default scrollbar opacity behavior in onboarding lists.
  * Refined visual styling standards across the application.


juliusmarminge:
The ceiling is set per layer to its measured count (this layer sits lower in the reordered stack than the commit message's original numbers implied). Every layer's value is re-measured after each rebase; CI's gate is exact.

PR #13028, opened 2026-09-22
refactor(web): add SidebarInput

juliusmarminge:
The sidebar thread search and the settings search both rendered an
unstyled Input and then reached into its slot with seven utilities to
make it bare sidebar text. That is a sidebar part, not an Input variant:
SidebarInput lives in sidebar.tsx with the sidebar tokens it depends on.
The placeholder now uses sidebar-muted-foreground at both sites; one used
the dimmer icon colour.

Lowers the no-restyle ceiling from 804 to 792.

Layer 9 of 24 in stack #13044, ordered least to most opinionated. Part of the `components/ui` className cleanup ([audit and guardrail: #12982](https://github.com/pingdotgg/t3code/pull/12982)). Each layer lowers the `shadcn/no-restyle` ceiling; CI fails if the count rises. For every override cluster the question was "is this a real design-system concept, or should the primitive get stricter and the consumers change?"; a prop was only added for the former.

## Summary by CodeRabbit

* **Style**
  * Updated search fields in Settings and thread headers with consistent sidebar styling.
  * Preserved existing search functionality, layout behavior, and accessibility support.
  * Improved visual consistency and alignment for inputs displayed within sidebar areas.
  * Standardized sidebar input appearance across supported fields.

## Before / after

**sidebar**

|---|---|


juliusmarminge:
Confirmed: `{...props}` spreads after `data-slot="input"`, so the forwarded slot clobbered it. Removed.

PR #13031, opened 2026-09-22
refactor(web): one wrap width for tooltips, plus a code variant

juliusmarminge:
Multi-line tooltips set their own max-width and line-height through
className in eight combinations, and path or command tooltips each
re-declared the same monospace treatment. TooltipPopup now wraps prose at
max-w-80 with one line-height, and variant="code" renders monospace at
max-w-120 breaking anywhere. The 60 call sites drop their classes.

Dialog and alert-dialog titles also gain wrap-anywhere in the base, since
any of them can interpolate a path.

Visible effect: tooltips that wrapped at 16, 18, 24 or 30rem now wrap at
20rem; wrapped tooltips share one line-height.

Lowers the no-restyle ceiling from 1207 to 1147.

Layer 12 of 24 in stack #13044, ordered least to most opinionated. Part of the `components/ui` className cleanup ([audit and guardrail: #12982](https://github.com/pingdotgg/t3code/pull/12982)). Each layer lowers the `shadcn/no-restyle` ceiling; CI fails if the count rises. For every override cluster the question was "is this a real design-system concept, or should the primitive get stricter and the consumers change?"; a prop was only added for the former.

## Summary by CodeRabbit

* **Style**
  * Standardized tooltip presentation across the app with shared default and code-specific styles.
  * Removed inconsistent component-specific sizing, wrapping, spacing, and typography overrides from tooltips.
  * Improved wrapping for long dialog titles.

* **Chores**
  * Tightened styling checks to enforce updated tooltip presentation standards.

## Before / after

**tooltip**

|---|---|
| tooltip before | tooltip after |


juliusmarminge:
All five point at the same thing: the default variant needs to break long tokens. Added `wrap-anywhere` to the default variant once in tooltip.tsx, so paths, URLs and status strings wrap in every tooltip.

PR #13032, opened 2026-09-22
refactor(web): one vertical rhythm for dialog bodies

juliusmarminge:
Dialog bodies set their own rhythm through className: fourteen at
space-y-4, three at 3, two at 5, one at 0 with custom padding, and seven
with none. That is one rhythm plus drift. DialogPanel now applies
space-y-4 in the base and the 27 dialogs drop their classes.

Visible effect: three dialogs (SSH password, link pull request, proposed
plan save) loosen by 4px between sections; two (theme import, add
environment) tighten by 4px; the seven that had no spacing now space
their direct children like every other dialog; the shared-policy settings
dialog regains the standard 24px side padding.

Lowers the no-restyle ceiling from 1089 to 1071.

Layer 13 of 24 in stack #13044, ordered least to most opinionated. Part of the `components/ui` className cleanup ([audit and guardrail: #12982](https://github.com/pingdotgg/t3code/pull/12982)). Each layer lowers the `shadcn/no-restyle` ceiling; CI fails if the count rises. For every override cluster the question was "is this a real design-system concept, or should the primitive get stricter and the consumers change?"; a prop was only added for the former.

## Summary by CodeRabbit

- **Style**
  - Standardized spacing between content elements across dialogs for a more consistent layout.
  - Removed redundant panel-level spacing from individual dialogs while preserving spacing within their content.
  - Refined padding and spacing in settings, project, pull request, connection, activity, and usage dialogs.
  - Improved loading-state presentation in the browser import wizard with clearer padded content areas.

## Before / after

**dialog-git**

|---|---|
| dialog-git before | dialog-git after |


juliusmarminge:
Agreed those two rows were fighting the rhythm. They're a plain flex div inside a standard DialogPanel now, no override.

PR #13034, opened 2026-09-22
refactor(web): InlineButton underlines on hover and takes a tone

juliusmarminge:
Six of eight InlineButton sites re-declared font-medium underline-offset-2
hover:underline plus a tone, each slightly differently; the other two
drew a bottom border instead. An inline text action has one look now:
surrounding text size, medium weight, underline on hover, with
tone="muted|destructive" for the two tones in use. It also accepts
render, so a link can be an inline action.

The thread pull-request badge drops buttonVariants and renders through
Button or InlineButton, and the tabular-nums/11px tweaks on it go.

Visible effect: the sidebar undo link underlines instead of drawing a
border; the two process signal links and the expandable-text toggle read
at surrounding size instead of 11px; the sidebar PR badge no longer forces
tabular figures.

Lowers the no-restyle ceiling from 1038 to 999.

Layer 15 of 24 in stack #13044, ordered least to most opinionated. Part of the `components/ui` className cleanup ([audit and guardrail: #12982](https://github.com/pingdotgg/t3code/pull/12982)). Each layer lowers the `shadcn/no-restyle` ceiling; CI fails if the count rises. For every override cluster the question was "is this a real design-system concept, or should the primitive get stricter and the consumers change?"; a prop was only added for the former.

## Before / after

**settings-provider-instance**

|---|---|
| settings-provider-instance before | settings-provider-instance after |

## Summary by CodeRabbit

* **Style**
  * Standardized inline action controls with default, muted, and destructive visual tones.
  * Refined hover, typography, and pull-request detail styling across settings and sidebar notices.
  * Improved presentation and interaction of pull-request and stack badges.

* **Bug Fixes**
  * Improved click handling for pull-request actions from different interactive elements.
  * Inline action controls now default to button behavior where appropriate, helping prevent unintended form submissions.


juliusmarminge:
Verified with renderToStaticMarkup: `type="button"` on the render element survives (mergeProps only overrides with defined values), so the stack badge can't become a submit button. No change.

PR #13035, opened 2026-09-22
refactor(web): one minimum width for menus, three widths for popovers

juliusmarminge:
MenuPopup decided whether to apply its minimum width by scanning the
incoming className for w-/min-w- tokens, and 37 menus picked widths from
fourteen values. Menus size to their content; they now share one minimum
(10rem) and a viewport cap, and no width prop. Popovers hold prose and
forms, so a fixed width is real: width="sm|md|lg" (16, 20, 24rem). Select
lists match their trigger, as SelectPopup already does.

Visible effect: menus that had forced a width of 14-24rem now fit their
content from a 10rem floor; popovers snap to the nearest of three widths
(e.g. 18rem becomes 20rem, 28rem becomes 24rem); the keybinding selects
match their trigger instead of a fixed floor.

Layer 16 of 24 in stack #13044, ordered least to most opinionated. Part of the `components/ui` className cleanup ([audit and guardrail: #12982](https://github.com/pingdotgg/t3code/pull/12982)). Each layer lowers the `shadcn/no-restyle` ceiling; CI fails if the count rises. For every override cluster the question was "is this a real design-system concept, or should the primitive get stricter and the consumers change?"; a prop was only added for the former.

## Summary by CodeRabbit

* **New Features**
  * Added consistent small, medium, and large width presets for popovers.
  * Popups and menus now adapt their sizing to content and available screen space.

* **Style**
  * Updated menus, dropdowns, and popovers throughout the app to remove rigid width constraints.
  * Improved responsive behavior across chat, settings, pull requests, navigation, and development workflows while preserving existing alignment, positioning, and interactions.

## Before / after

**menu-diff-scope**

|---|---|
| menu-diff-scope before | menu-diff-scope after |

**popover-model-picker**

|---|---|
| popover-model-picker before | popover-model-picker after |

**settings-provider-update-advisory**

|---|---|
| settings-provider-update-advisory before | settings-provider-update-advisory after |


juliusmarminge:
Restored `max-w-56` on both profile menus; a max-width is layout and stays on className, the comment is accurate again.

PR #13037, opened 2026-09-22
refactor(web): stacked sidebar groups share one inset

juliusmarminge:
Sidebar groups trimmed the edge facing a neighbour by hand (px-2 pt-2
pb-0 and variants) because SidebarContent stacked them with a gap and each
group carried a full inset. The container owns that rhythm now: no gap,
and a group that follows a group drops its top inset. Seven groups drop
their padding classes.

Visible effect: none where groups already trimmed themselves; the settings
sidebar nav and the legacy sidebar's project list lose a doubled 8px seam
between groups.

Lowers the no-restyle ceiling from 827 to 808.

Layer 18 of 24 in stack #13044, ordered least to most opinionated. Part of the `components/ui` className cleanup ([audit and guardrail: #12982](https://github.com/pingdotgg/t3code/pull/12982)). Each layer lowers the `shadcn/no-restyle` ceiling; CI fails if the count rises. For every override cluster the question was "is this a real design-system concept, or should the primitive get stricter and the consumers change?"; a prop was only added for the former.

## Summary by CodeRabbit

- **Style**
  - Refined spacing and alignment across the main, legacy, and settings sidebars.
  - Reduced excess gaps, padding, and insets between sidebar sections for a more consistent, compact layout.
  - Improved spacing around sidebar headers and content groups, including areas with fixed headers.
  - Improved sidebar header layering to keep search and navigation elements displayed correctly.
  - Standardized the appearance of sidebar sections across navigation and settings views.

## Before / after

**sidebar**

|---|---|

**settings-keybindings**

|---|---|
| settings-keybindings before | settings-keybindings after |


juliusmarminge:
Fixed in SidebarContent: when a fixedHeader is present the first content group drops its top inset too, so the seam is handled by the container rather than a per-site class.

PR #13039, opened 2026-09-22
refactor(web): show more / show less are ordinary sidebar sub-rows

juliusmarminge:
The two list-expander rows in the thread list restyled SidebarMenuSubButton
into a taller, dimmer row. They are rows in the list and now look like
the rows around them.

Visible effect: the show more / show less rows in the legacy sidebar go
from 32px at 75% muted text to the standard 28px sub-row.

Lowers the no-restyle ceiling from 643 to 633.

Layer 20 of 24 in stack #13044, ordered least to most opinionated. Part of the `components/ui` className cleanup ([audit and guardrail: #12982](https://github.com/pingdotgg/t3code/pull/12982)). Each layer lowers the `shadcn/no-restyle` ceiling; CI fails if the count rises. For every override cluster the question was "is this a real design-system concept, or should the primitive get stricter and the consumers change?"; a prop was only added for the former.

## Summary by CodeRabbit

- **Style**
  - Removed hover styling from the “Show more” and “Show less” thread-list buttons.
  - Button labels, layout, and expand/collapse behavior remain unchanged.

- **Chores**
  - Updated styling validation thresholds to reflect the latest interface changes.

## Before / after

**sidebar-bottom**

|---|---|
| sidebar-bottom before | sidebar-bottom after |


juliusmarminge:
SidebarGroup's base already has `relative`, so `z-[1]` establishes the stacking context; no change needed.

PR #13040, opened 2026-09-22
refactor(web): Empty has three sizes

juliusmarminge:
Empty states set their own min-height, padding and title scale per site.
Two of those are real: a card-sized notice (compact) and a route-filling
hero. The rest were how much air a given screen wanted, which is not a
kind of empty state. Expose size="compact|default|hero"; the two account-
page empty states keep their top border on a wrapper because the border
belongs to the list slot they replace, and the pull-request action rows
are a plain flex div under the Empty.

Visible effect: the source-control and provider settings empty states
lose a forced 22rem min-height; the four pull-request empty states lose
16px of extra vertical padding.

Lowers the no-restyle ceiling from 965 to 936.

Layer 21 of 24 in stack #13044, ordered least to most opinionated. Part of the `components/ui` className cleanup ([audit and guardrail: #12982](https://github.com/pingdotgg/t3code/pull/12982)). Each layer lowers the `shadcn/no-restyle` ceiling; CI fails if the count rises. For every override cluster the question was "is this a real design-system concept, or should the primitive get stricter and the consumers change?"; a prop was only added for the former.

## Summary by CodeRabbit

- **UI Improvements**
  - Updated empty states across projects, client connections, environments, pull requests, provider settings, and source control for a more consistent layout.
  - Added compact and hero display options to better match available space and context.
  - Adjusted spacing, borders, minimum heights, and typography for clearer, more balanced presentations.
  - Preserved existing pull request actions, including retry and GitHub-related options.

## Before / after

**empty-pull-requests**

|---|---|
| empty-pull-requests before | empty-pull-requests after |

**empty-source-control**

|---|---|
| empty-source-control before | empty-source-control after |


juliusmarminge:
Deliberate: an action row under an empty state is a layout concern of that screen, not a slot on EmptyContent. Three sites use a plain flex div.

PR #13042, opened 2026-09-22
refactor(web): render menu and popover triggers through Button

juliusmarminge:
Six headless triggers hand-rolled a full button recipe in className. Two
were the same accent pill, which is Button variant="secondary" size="xs";
four were bare text links, which is InlineButton. Render through those
instead so the trigger inherits focus rings and hover from one place.

Visible effect: the diff-scope pickers use the secondary token instead of
accent, one step lighter at rest; the device-tool version links read at
surrounding text size.

Lowers the no-restyle ceiling from 762 to 713.

Layer 23 of 24 in stack #13044, ordered least to most opinionated. Part of the `components/ui` className cleanup ([audit and guardrail: #12982](https://github.com/pingdotgg/t3code/pull/12982)). Each layer lowers the `shadcn/no-restyle` ceiling; CI fails if the count rises. For every override cluster the question was "is this a real design-system concept, or should the primitive get stricter and the consumers change?"; a prop was only added for the former.

## Summary by CodeRabbit

* **Style**
  * Updated dropdown and popover controls across diff views, pull requests, device tools, settings, usage, and chat to use consistent shared button styles.
  * Improved visual consistency for compact and inline controls while preserving existing labels, sizing, spacing, and behavior.
  * Refined trigger styling for clearer, more cohesive navigation and menu interactions.
  * Standardized secondary and muted appearances across menus and contextual controls.

## Before / after

**menu-diff-scope**

|---|---|
| menu-diff-scope before | menu-diff-scope after |


juliusmarminge:
Fixed: both UsagePage test mocks export InlineButton.

PR #13059, opened 2026-09-22
fix(web): context strip labels no longer loop between compact and expanded

### LINKED ISSUE #12891: [Bug]: React #185 loop in the composer context strip label compaction (BranchToolbar measure -> setOverflows), 0.0.43-nightly.20260921.2044
### Before submitting

- [x] I searched existing issues and did not find an open duplicate.
- [x] I included the available details needed to investigate the crash.

### Area

apps/web

### Steps to reproduce

No deterministic repro. The renderer crashed once to the crash screen while a thread was open on `0.0.43-nightly.20260921.2044`. Retry recovered and it has not recurred since.

Because the stack decodes cleanly, the mechanism is reproducible by reasoning even though the trigger is not: it is the composer context strip's label compaction loop in `BranchToolbar.tsx`, not `@legendapp/list` as in #11308.

### Expected behavior

The context strip settles on one of the two label states and the renderer stays up.

### Actual behavior

React error #185 (`Maximum update depth exceeded`).

I extracted the client assets out of the installed `app.asar` for this exact build and matched the stack offsets against `origin/main`:

| Minified frame | Source |
|---|---|
| `dist-BVnGsFIt.js:26:27485` / `:27010` | react-dom `getRootForUpdatedFiber` / `enqueueConcurrentHookUpdate` (the `throw Error(185)` site) |
| `dist-BVnGsFIt.js:26:58533` / `:58155` | react-dom `dispatchSetStateInternal` / `dispatchSetState` |
| `_chat-DOCEAHoD.js:8:59203` | `setOverflows(nextOverflows)`, the last statement of `measure()` in `BranchToolbar.tsx` |
| `_chat-DOCEAHoD.js:8:59884` | `measure()` inside `useLayoutEffect(() => { measure(); })`, the one with **no dependency array** |
| `dist-BVnGsFIt.js:26:91668` / `:96139` / `:104933` / `:96697` | react-dom commit / work loop |

So the loop is: every render runs the dependency-free layout effect, `measure()` recomputes `needed` vs `available` and calls `setOverflows`, and when the two states disagree about each other the flip-flop is synchronous until React gives up at 50 nested updates. The only guard is `CONTEXT_STRIP_COMPACT_EXPAND_HYSTERESIS_PX = 16` in `BranchToolbar.logic.ts:81`, which cannot absorb a disagreement larger than 16px.

Note that the label width animation is not involved here: this machine has Reduce Motion on, so the `[overflows]` layout effect returns at the `prefers-reduced-motion` check before animating anything. The loop is measurement only.

**Suspect, stated as a hypothesis rather than a measurement:** #12805 (`feat(web): truncate branch names and paths in the middle`, merged 2026-09-20 23:43 -0300) is in this build. I confirmed it by finding the minified `splitForMiddleTruncate` in the shipped chunks. That PR changed the branch label that `measure()` measures, from

```jsx
<span data-composer-label-motion className="block w-full min-w-0 max-w-[240px] truncate ...">{triggerLabel}</span>
```

to a `MiddleTruncate`, which renders two spans (`min-w-0 truncate` head plus `shrink-0` tail) inside an `inline-flex overflow-hidden` wrapper. But `measure()` still reserves hidden text as

```ts
let textWidth = label.scrollWidth;
for (const inner of label.querySelectorAll<HTMLElement>("*")) {
  textWidth = Math.max(textWidth, inner.scrollWidth);
}
needed += Math.max(0, textWidth - label.getBoundingClientRect().width);
```

a **max** over the subtree, which describes one clipped span, not a head/tail split whose real width is closer to head + tail. The shipped bundle also shows the strip in a mixed state: the branch label is already `MiddleTruncate`, while the "Run on" and environment labels still render the old single truncating span, so the two label kinds are now measured by the same formula with different accuracy. Combined with the `max-w-[240px]` cap on the expanded label and `max-w-0` in compact, the reconstructed `needed` can differ between the two states by much more than 16px.

I have not reproduced the numeric path, so treat the causal link as dated correlation plus a plausible mechanism, not proof.

### Impact

### Version or commit

T3 Code (Nightly) 0.0.43-nightly.20260921.2044

### Environment

macOS 26.6.2 (arm64), MacBook Pro built-in Liquid Retina XDR plus a 5120x2880 external display, DPR 2. Reduce Motion is enabled system-wide. Window size at crash time was not captured. The thread path from the crash screen is omitted because it contains private identifiers.

### Logs or stack traces

```shell
Error: Minified React error #185; visit https://react.dev/errors/185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at fi (t3code://app/assets/dist-BVnGsFIt.js:26:27485)
    at li (t3code://app/assets/dist-BVnGsFIt.js:26:27010)
    at Is (t3code://app/assets/dist-BVnGsFIt.js:26:58533)
    at Fs (t3code://app/assets/dist-BVnGsFIt.js:26:58155)
    at t3code://app/assets/_chat-DOCEAHoD.js:8:59203
    at t3code://app/assets/_chat-DOCEAHoD.js:8:59884
    at Uc (t3code://app/assets/dist-BVnGsFIt.js:26:91668)
    at sl (t3code://app/assets/dist-BVnGsFIt.js:26:96139)
    at xl (t3code://app/assets/dist-BVnGsFIt.js:26:104933)
    at sl (t3code://app/assets/dist-BVnGsFIt.js:26:96697)
```

This is the full stack as shown on the crash screen; it is not truncated.

### Workaround

Retry from the crash screen recovered immediately and it has not recurred. Since the decision depends on strip width and on the current branch, model and environment label text, widening the window should move the strip out of the flapping band.

riccardopll:
Fixes #12891

## Problem

Scrolling up in a thread crashed the desktop app to the "Something went wrong" screen with React error # 185 (maximum update depth exceeded) on `0.0.43-nightly.20260922.2096`. The minified frames decode to `setOverflows(nextOverflows)` in `measure()` and the dependency-free `useLayoutEffect(() => { measure(); })` in `useLabelsOverflow` (`apps/web/src/components/BranchToolbar.tsx`), the loop triaged in #12891.

`measure()` reserved hidden label text as the largest `scrollWidth` in the label subtree. Since #12805 the branch label is a `MiddleTruncate`, a clipped head beside an unclipped tail, so the compact pass reserved the head only while the expanded pass laid out head and tail. The two passes disagreed by about one tail, more than the 16px hysteresis, and each layout effect flipped the other until React gave up.

Scrolling up hits it reliably: the composer collapses into the resting layout, its model and mode controls move into the context strip, and the strip lands on that boundary.

## Fix

Sum the leaf spans of each label instead, capped at the motion element's max-width (the one width the compact override leaves alone), so compact and expanded reserve the same room. The math lives in `resolveContextStripLabelHiddenWidth` with tests, including a sweep over strip widths asserting the compact decision settles in one pass.

No visual change. Verified with the touched test file, targeted lint, and the web typecheck.

Made with Claude Fable 5.1 in Claude Code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved context-strip label sizing when space is limited.
  - Labels now reserve the appropriate amount of hidden text width, preventing inaccurate overflow calculations.
  - Improved transitions between compact and expanded label states, avoiding visual oscillation as available space changes.
  - Label sizing now remains consistent across different strip widths and available space.


riccardopll:
hi @juliusmarminge is there any way you can review this? unfortunately t3code is borderline unusable with this constant react error popup

PR #13067, opened 2026-09-22
fix(orchestration): report subagent models and use live model choices

Bil0000:
Fix subagent model labels so children show their reported model instead of the parent's. Use the composer's live model catalog for delegation, including custom models; native CLI lists remain provider-controlled.

Verified on current v2: 367 focused tests, six subagent replays, server/web type checks, lint, and formatting. Grok's replay startup failure also occurs on unchanged v2.

Model: GPT-6 Astra. Harness: Codex.


Bil0000:
CI status on `f7ff936427`: all test jobs pass. `Check` fails only at the `shadcn/no-restyle` ceiling (659 findings, limit 627). The [v2 base job](https://github.com/pingdotgg/t3code/actions/runs/35817304175/job/107041480892) fails with the same count and limit; the [PR job](https://github.com/pingdotgg/t3code/actions/runs/35817520959/job/107042138393) has the same result. This PR's web changes do not add style overrides.

Bil0000:
Update for `08bbfd8dc1`: the PR is now 2 commits and 12 files against the current V2 base. `Check`, `Test`, all three server test shards, and the other required checks pass. My earlier no-restyle CI failure note applied to the old head and is resolved.

PR #13075, opened 2026-09-22
feat(web): morph the terminal icon into a chevron when a command group opens

dBianchii:
## What Changed

A collapsed command group in the timeline shows a static terminal icon. It now morphs into a chevron-down while the group is expanded, and back on close.

Command groups only — every other summary icon (hammer, eye, square-pen, MCP favicons…) keeps its static glyph. `prefers-reduced-motion` degrades it to an instant swap.

## Why

Nothing signalled that the row opens, and nothing changed once it did. Morphing rather than swapping keeps the affordance and the "this is a command" signal on the same 24×24 grid, so the row never reflows.

Adds `morphicons` (6.5 KB gzipped, zero runtime deps) and `lucide` — the icon *data* package, which `morphicons` needs because `lucide-react` exports components. Pinned to the installed `lucide-react` version so both draw the same shapes; tree-shakes to 290 B gzipped under rolldown.

## UI Changes

https://github.com/user-attachments/assets/29fa09ad-5492-4644-9821-c0d396f19c92

## Checklist

---

Opus 5 in Claude Code, driven from T3 Code.

## Summary by CodeRabbit

* **UI Improvements**
  * Updated tool-group controls with clearer expand and collapse indicators.
  * Added an animated transition between disclosure states for a smoother experience.
  * The interface now respects reduced-motion preferences and avoids unnecessary animation when enabled.
* **Bug Fixes**
  * Improved the fallback icon shown for tool groups without a custom icon, making the control’s purpose clearer.


dBianchii:
If you look at the video, you will see that there are other icon expansions that are not changed by this. 
image
For example, the individual bash calls themselves ^

I didn't change that because it would require keeping the icon in the top left. Maybe that's desirable. If so, let me know and I will edit the code accordingly.

Additionally, maybe there are other places where we would want to do this morph icons thing. 

I opened this PR to start that discussion and see what other changes we need

apps/web/package.json
```diff
@@ -47,7 +47,9 @@
     "jose": "catalog:",
     "jsonc-parser": "3.3.1",
     "jszip": "3.10.1",
+    "lucide": "0.564.0",
```
dBianchii, line 50: The reason why I added 'lucide' is because it includes the raw data needed to make morphicons work. 

I tested doing this with just lucide-react and the same couldn't be accomplished without hardcoding additional data for the icons

PR #13085, opened 2026-09-22
fix(mobile): preserve Android search and filters when folding

PixPMusic:
Folding an Android device switches the Threads list between the compact route and the expanded sidebar. Each kept its own project filter, and the compact route kept a separate search query. Unfolding could therefore show an empty query and all projects; folding back could restore an older selection.

The Android compact route now uses the workspace search state already used by the sidebar. Both lists use the existing shared list-options provider for project selection, alongside environment and sort options. Clearing a filter updates the same shared state. iOS project filters benefit from the shared selection; its native search behavior is unchanged because its fields do not restore text from this state. This is independent of #11057 and #10629.

## Before / After

Android 17 Pixel 9 Pro Fold emulator, dark theme, identical native build and isolated showcase data. Select **Linux**, enter **Make** while folded, then unfold:

| Before: search and project filter lost | After: query and Linux selection retained |
| --- | --- |
| Before unfolding resets search and shows all projects | After unfolding retains Make and shows only Linux threads |

[Before recording](https://github.com/PixPMusic/t3code/releases/download/pr-13085-assets/android-fold-before-evidence.mp4) · [After recording: three round trips](https://github.com/PixPMusic/t3code/releases/download/pr-13085-assets/android-fold-after-evidence.mp4) · [Folded after the third return](https://github.com/PixPMusic/t3code/releases/download/pr-13085-assets/android-after-folded-return.png)

Native screenshots are 1080 × 2424 folded and 2076 × 2152 unfolded. The recording keeps a fixed canvas, with black margins around the folded screen. Evidence is hosted on the fork's evidence-only release; no assets are committed to this PR.

## Verification

- Reproduced on production source from `aff9318bf4`; verified the fix on `6f9dfd0352` using the same native client and emulator.
- Three folded → unfolded → folded round trips retain **Make**, the **Linux** selection, and matching results after every transition.
- Changed to the **T3 Code** project and **remote** from the expanded sidebar; three unfolded → folded → unfolded round trips retain the updated state. Clearing search still leaves only that project's threads after folding.
- Cleared search and selected **All projects**; three more round trips keep both cleared and show threads from all three projects.
- The legacy thread list also retains **Make** and **Linux** through three folded → unfolded → folded round trips.
- 60 focused tests pass across home list options, filter menus, thread grouping, and layout derivation. The lifecycle regression starts in either layout and checks project, environment, and sort state through repeated select/change/clear transitions.
- Mobile typecheck and scoped formatting pass. Scoped lint has no new warnings compared with the base revision. `git diff --check` passes.
- Changes are limited to mobile client state. Web/desktop, providers, contracts, and connection protocols are unchanged. Runtime verification covers Android's current and legacy thread lists; iOS native search remains covered by its separate PR.

Model: GPT-6 Astra | Harness: Codex in T3 Code


PixPMusic:
Reviewed CodeRabbit's advisory docstring-coverage warning. The repository does not configure an 80% docstring requirement, and AGENTS.md asks for comments that explain non-obvious behavior rather than narrating code. The shared provider already explains its layout lifetime, and the Android/iOS search-state distinction has a nearby comment. Additional docstrings for the small setters and list components would repeat their types and implementation, so I am leaving those unchanged.

CodeRabbit reported no actionable code findings, and Macroscope approved `6f9dfd0352`. The superseded fingerprint run was cancelled when the PR description changed; the newer fingerprint run passed on the same head.

PR #2642, opened 2026-05-11
[codex] Add remote provider compatibility map

juliusmarminge:
## What changed

- Added provider compatibility advisory contract fields, including supported/broken/graceful status, severity, recommended range, and recommended version.
- Added a bundled fallback compatibility document and a checked-in `provider-compatibility.v1.json` intended for GitHub raw hosting.
- Added best-effort remote compatibility map enrichment with timeout, TTL cache, schema validation, and bundled fallback behavior.
- Wired compatibility enrichment into provider snapshot background enrichment for Codex, Claude, Cursor, and OpenCode.
- Surfaced provider compatibility warnings in the provider settings UI with a warning/error popover.

## Why

Released T3 Code builds need compatibility policy updates without requiring users to update the app. This lets maintainers update a static GitHub-hosted map to mark provider harness versions as supported, graceful, unsupported, or broken for specific T3 Code release ranges.

## Validation

- `bun fmt`
- `bun lint` (passes with existing unrelated warnings)
- `bun typecheck`
- `bunx vitest run packages/contracts/src/server.test.ts apps/server/src/provider/providerSnapshot.test.ts apps/server/src/provider/providerCompatibility.test.ts apps/web/src/components/settings/providerStatus.test.ts`

> [!NOTE]
> ### Add remote provider compatibility map with advisory enrichment and UI update actions for Codex
> - Introduces [provider-compatibility.v1.json](https://github.com/pingdotgg/t3code/pull/2642/files#diff-e5482c000e4911b60aa516288c47b6e308fedd1fd70873a7d639b965920ee150) as a versioned compatibility policy document covering Codex, Claude, OpenCode, and Cursor drivers, with semver ranges and recommended versions gated by T3 Code version.
> - Adds a new `ProviderCompatibility` service ([ProviderCompatibility.ts](https://github.com/pingdotgg/t3code/pull/2642/files#diff-fc81737f210236d13e8d3d45edb71f74b1484a13ce568b1564d717dc3aefebdc)) that loads bundled policies or fetches a remote document, enriches provider snapshots with `compatibilityAdvisory` fields, and may override provider status and message.
> - All four provider drivers (Codex, Claude, OpenCode, Cursor) now run targeted compatibility advisory enrichment in their snapshot pipelines and require `ProviderCompatibilityService` in their Effect environments.
[...]

PR #13131, opened 2026-09-22
feat(server): flag incompatible provider versions from the model manifest

juliusmarminge:
Independent Opus reimplementation of the closed [#2642](https://github.com/pingdotgg/t3code/pull/2642), built on `main`.

## Problem

T3 Code can't tell a user that their installed provider CLI is too old for this build. A Claude Code older than 2.1.111, for example, can't run any current Claude model. Today the only settings signal is a generic "Update available" arrow, and the real failure shows up mid-turn.

## Fix

- **Data.** The model manifest gets an optional `compatibility` section. It reuses the existing remote → disk cache → bundled pipeline, TTL, and `enableProviderUpdateChecks` gate, so there is no second fetcher or cache.
  - Each driver lists policies selected by T3 Code version range.
  - The first matching provider range classifies the installed version as `supported`, `graceful`, `unsupported`, or `broken`.
  - A policy can name a `recommendedVersion` for when "update to latest" is the wrong fix.
  - Range syntax the matcher can't evaluate (`~`, `x`) fails decoding instead of silently never matching.
  - A remote or cached manifest with no `compatibility` key keeps the bundled ranges. An explicit `{}` clears them.
- **Evaluation.** A pure function runs in `ProviderRegistry.upsertProviders`. Every driver's snapshot passes through that function, so every adapter is covered.
  - Only non-`supported` verdicts are attached, as a new optional `ServerProvider.compatibility` field.
  - The verdict is recomputed from the snapshot's current version on every upsert, so it can't go stale.
  - It is advisory only: it never changes `status` or blocks a session.
- **Targeted update.** `server.updateProvider` accepts an optional `targetVersion`.
  - Package-manager installs (npm, pnpm, bun, Vite+) are rewritten to `pkg@<version>`.
  - Native updaters and Homebrew can only reach latest, so they refuse a pin with a clear error.
  - A pinned install counts as succeeded when it lands on the requested version, even if that is behind latest.
- **Web.**
  - The settings update popover leads with the verdict: "Unsupported version", "Known broken version", or "Limited support".
  - When a version is recommended, the button reads "Install vX" and the copyable latest command is hidden.
  - The composer banner warns about `unsupported` and `broken` versions on providers that are otherwise ready.
  - A dismissed banner comes back if the provider moves to a different bad version.

## Bundled ranges

These were checked against `main`'s adapters rather than copied from #2642.

| Driver | Range | Status | Why |
|---|---|---|---|
| claudeAgent | `>=2.1.280` | supported | Newest `minVersion` in the bundled model manifest |
| claudeAgent | `>=2.1.111` | graceful | `ClaudeModelCatalog` hides newer models on older CLIs |
| claudeAgent | `<2.1.111` | unsupported | Oldest model `minVersion`, so no current model runs |

The other providers get no entry today:

- **OpenCode** already refuses versions older than `MINIMUM_OPENCODE_VERSION` at probe time.
- **Codex** has no known version floor on `main`. The 0.141 floor from the V2 adapter does not apply here.
- **Cursor, Grok, and Antigravity** have no known threshold.

The schema and evaluator are provider-agnostic, so any of these can gain ranges through the remote manifest without a release.

## Surfaces

- **Web and desktop.** Desktop wraps web, so both get the settings popover and the composer banner.
- **Remote and relay.** The verdict travels in the existing provider snapshot, so remote clients see it with no extra calls.
- **Mobile.** Mobile has no provider update UI today. It ignores the new optional field and doesn't change.

## Verification

- `vp test run` passes for:
  - server: `providerCompatibility`, `providerMaintenance`, `providerMaintenanceRunner`, `ProviderRegistry`, `ModelManifest`
  - web: `ProviderStatusBanner`, `providerStatus`, `ProviderUpdateLaunchNotification.logic`
- Exception: two npm-prefix cases in `providerMaintenance.test.ts` fail on my machine, and fail the same way on unmodified `main` (d7819c18813). They pass in CI, so they are environment-specific and not introduced here.
- The new tests cover:
  - first-match ordering and T3 Code policy selection
  - missing, `0.0.0`, and unmatched versions
  - `recommendedVersion` rules
  - rejection of unsupported range syntax
  - replacement of a stale verdict, and fallback to the bundled ranges
  - pinned command rewriting per installer
  - the runner's pinned-success and native-refusal paths
  - the banner re-showing on a version change
  - the popover preferring the recommended version
- `typecheck` is clean for `packages/contracts`, `apps/server`, `apps/web`, and `apps/mobile`.

## Evidence

All shots use the same seeded state and a 1280×800 viewport, with Claude Code 2.1.100 installed from npm. Before is `main` (d7819c18813); after is this branch.

**Settings → Providers, Claude update popover**

| Before | After |
|---|---|
| Before: Claude 2.1.100 shows a generic "Update available" popover | After: the popover reads "Unsupported version" and gives the Claude reason |

**Composer on a Claude thread**

| Before | After |
|---|---|
| Before: no warning above the composer | After: a banner reads "Claude 2.1.100 is unsupported" |

Made with Claude Opus 5.5 (`claude-opus-5-5`) in Claude Code, running in T3 Code.

---

<a href="https://app.devin.ai/review/pingdotgg/t3code/pull/13131" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added provider compatibility checks with supported, graceful, unsupported, and broken statuses.
  * Added compatibility warnings and guidance in provider banners and settings.
  * Added recommendations to install specific compatible provider versions when available.
  * Provider updates can now target an exact compatible version where supported.

* **Bug Fixes**
  * Preserved compatibility information when refreshing or loading provider manifests.
  * Improved update commands to correctly pin package-manager installations.

* **Documentation**
  * Documented provider compatibility policies and version recommendations.


juliusmarminge:
Superseded by [#13130](https://github.com/pingdotgg/t3code/pull/13130), now merged into main. It carries over this PR's compact settings warning icon, tooltip and version popover, with short inline status labels and full health diagnostics on hover or focus. It retains the compatibility refresh and update guardrails from that implementation. Closing this alternate implementation as requested.

PR #13146, opened 2026-09-22
chore(lint): keep mobile theme escape-hatch allowlist honest

juliusmarminge:
## Problem

The mobile lint rule `t3code/no-mobile-uniwind-theme-escape-hatches` keeps theme styling on semantic Uniwind classNames and only exempts reviewed native/third-party interop boundaries via an `allowUniwindTheme` override list in `vite.config.ts`. The audit of that boundary found it had drifted:

- `features/terminal/ThreadTerminalRouteScreen.tsx` was still allowlisted even though the shared terminal header refactor removed its last theme-bridge usage.
- `ThreadFilesTreeScreen` still read the JS palette (`useUniwindTheme()["--color-header"]`) to background an ordinary Android header `View`, which Uniwind handles with a semantic class.
- Review follow-up: `AppearancePreviews.tsx` used the JS palette only for two ordinary `Text` nodes (line numbers / keywords in the settings code preview) — a real ordinary-screen escape hatch, not an interop boundary.

## Fix

- Drop the stale `ThreadTerminalRouteScreen.tsx` entry.
- Style the files-screen Android header with `bg-header` (the same semantic token home screens already use) and delete the theme read.
- Style the appearance code preview with `text-icon-subtle` and a conditional `text-md-link` on the nested keyword spans (same tokens; the nested-`Text` conditional class pattern already ships in `thread-search-match.tsx`), keep only dynamic font metrics in `style`, and drop `AppearancePreviews.tsx` from the allowlist so future JS palette reads there are linted.

The allowlist now matches the actual `useUniwindTheme` importers 1:1 (24 = 24).

## Audit claim checked: "the rule fails on a clean tree because `src/lib/useUniwindTheme.ts` is missing from the allowlist"

Not reproducible; no entry is needed for that file.

- `apps/mobile/src/lib/useUniwindTheme.ts` **exists**; it is the hook *definition*, not an importer. The rule reports *imports* of `…/useUniwindTheme`, `useCSSVariable`, `useThemeColor`, and `dark:`/`light:` literals — the definition file triggers none of these (it only imports `AppearancePreferencesProvider`).
- Repo-wide lint on a clean tree at the base commit passes: exit 0, zero diagnostics from this rule. A probe file importing `useUniwindTheme` outside the allowlist does error, confirming the rule and its overrides are actually wired.
- Counting 26 grep hits for `useUniwindTheme` includes the definition file; the remaining files are exactly the allowlisted importers. Adding an allowlist entry for the definition would be dead config, so none was added.

## Audit claim checked: "the allowlist includes ordinary screens" (ArchivedThreadsScreen, GitOverviewSheet, NewTaskDraftScreen, ThreadSettingsSheet, thread-list items ×2, HomeHeader, SettingsEnvironmentsRouteScreen, ConnectionsNewRouteScreen)

Each of these still has a genuine JS-color consumer that cannot take a `className`; verified per file against current source:

| File | Consumer of the theme color |
| --- | --- |
| ArchivedThreadsScreen, thread-list-items, thread-list-v2-items | `ThreadSwipeable backgroundColor` → RNGH/Reanimated swipeable container styles (`thread-swipe-actions.tsx`); v2 routes through `getThreadListV2RowAppearance` and also feeds the swipeable |
| GitOverviewSheet, ReviewSheet, thread-file-navigator-pane, ThreadFilesRouteScreen | native react-native-screens `Screen` `style` / `ScreenStackHeaderConfig` colors, `SearchBar`, SVG gradient stops |
| NewTaskDraftScreen, ThreadComposer | `T3ComposerEditor` native `textStyle` / `fallbackColor` |
| ThreadSettingsSheet, HomeHeader, ThreadFilesRouteScreen (`headerTintColor`) | react-navigation JS `screenOptions` (`contentStyle`, `headerStyle`, `headerTintColor`) |
| SettingsEnvironmentsRouteScreen, ConnectionsNewRouteScreen | header action `tintColor` → `NativeHeaderToolbar` extracts `child.props.tintColor` and hands it to the native header item |

Partial class migrations for the row-background reads (e.g. `bg-thread-selected` on `RowPressable`) were considered and skipped: those files keep their allowlist entries because of the swipeable/navigation consumers, so the change would alter high-traffic selected-row visuals (interaction class vs inline style precedence in `RowPressable`) without removing any escape hatch. Not safely scoped for a lint-enforcement PR.

## Verification (non-visual; screenshots not applicable)

- `vp test run` in `oxlint-plugin-t3code`: 50/50 pass (rule behavior unchanged).
- Full repo lint (`vpr lint`, includes `--report-unused-disable-directives`): exit 0 before and after; zero `no-mobile-uniwind-theme-escape-hatches` diagnostics; baseline probed with a temp importer to confirm rule + allowlist are applied.
- `apps/mobile` `tsc --noEmit`: clean.
- No visual change: `bg-header`, `text-icon-subtle`, and `text-md-link` resolve to the same `--color-header`, `--color-icon-subtle`, and `--color-md-link` tokens the removed JS reads returned.

---

Model: callstack/Apex via pi

## Summary by CodeRabbit

* **Style**
  * Android thread file screens now use the app’s standard header background. Other platforms are unchanged.
  * Appearance previews now use the app’s standard text colors for code line numbers and highlighted keywords, keeping these elements consistent with the selected appearance.


juliusmarminge:
Fixed the reviewer-audit P2: `AppearancePreviews.tsx` rendered its code-preview line numbers/keywords with JS palette reads on ordinary `Text` nodes. Migrated to `text-icon-subtle` / conditional `text-md-link` (nested-Text conditional classes already ship in `thread-search-match.tsx`), removed its allowlist entry, and verified allowlist == importers 1:1. Head: f27a865730a4279d06f11eb068ddcf2d885b1af1.

PR #13151, opened 2026-09-22
refactor(mobile): break module cycles with focused extractions

juliusmarminge:
## The problem

Mobile audit item #13: the app has real circular imports. Composer attachment preview retention lived in `state/use-composer-drafts` while `lib/attachmentUpload` / `lib/localAttachmentPreview` called it, and composer-draft cleanup reached back into `lib/attachmentUpload` through an `await import()` escape hatch — a bidirectional state↔lib dependency that only avoided a static cycle via dynamic-import gymnastics (madge counts the dynamic edge and reports it as cycles 1–4). On top of that, four component pairs shared a prop/request type in one direction while importing the component in the other, and a fifth cycle (`ThemedSwitch` ↔ `MaterialSwitch`) hides behind Android platform resolution:

- `FilePreviewModal` ↔ `FilePreview` ↔ `MediaImagePreview`
- `ConfirmDialogHost` ↔ `MaterialConfirmDialog`
- `SegmentedControl` ↔ `MaterialSegmentedControl`
- `reviewWordDiffs` ↔ `shikiReviewHighlighter`

The audit's other entries (`state/thread-order` → thread-list model, `lib/runtime` → cloud/observability, `components`/`native` → appearance, etc.) are one-directional upward edges, not cycles — this PR pins them instead of untangling them.

## The fix

Smallest honest seam per cycle:

- **`lib/composerAttachmentPreviewRetention.ts`** now owns the preview-retention lease. Composer draft state registers the owner-side cleanup hook at module load, removing the `lib → state/use-composer-drafts` upward edges that forced the dynamic-import round trip. Behavior is unchanged: the file stays readable until the last preview/upload release, which still retries the unused-file sweep. The registered hook mirrors the existing `ConfirmDialogHost` module-level registration idiom; while unregistered a release simply skips the retry, which is exactly the pre-existing state (no draft store loaded ⇒ no owners to sweep). A callback-parameter API instead would have re-introduced the `lib → state` edge at every call site.
- **`.types.ts` extractions** (following the existing `ControlPillMenu.types.ts` precedent): `ConfirmDialog.types.ts`, `SegmentedControl.types.ts`, `FilePreviewModal.types.ts`, `reviewHighlightedToken.types.ts`. The old modules re-export the types so external call sites (including `features/threads` and `features/files`) are untouched.
- **`legacy-plan-mode.ts` → `state/`**: it is pure contract-shaped model logic that `state/thread-outbox-model` and `state/use-thread-composer-state` imported out of `features/threads`.
- **`ThemedSwitchProps` → `MaterialSwitch.types.ts`** for the Android-hidden cycle flagged in review; acyclic under any resolution order.
- **`src/dependency-graph.test.ts`** is the regression check. It models Metro's per-platform resolution (platform ext before `.native` before generic; each platform graph excludes the other platform's variant files) and asserts (1) zero circular static imports **under both iOS and Android resolution** — dynamic `import()` stays the documented async escape hatch, so `state/use-composer-drafts → lib/attachmentUpload` remains a lazy edge, not a static cycle — and (2) shrink-only ceilings on the remaining upward edges (unique module pairs across platforms): state→features 6, lib→features 7, components→features 33, native→features 8, lib→state 11 (`lib/runtime.ts` is the composition root and counts toward its ceiling). Verified to fail on a seeded Android-only type cycle and on a seeded lib→features import.

**Follow-up (tracked, not in this PR):** the pinned ceiling edges themselves — thread-order→thread-list model, appearance-provider imports from `lib`/`components`, terminal preferences from `lib`, native→features glue — each need an owner-side seam and should be lowered opportunistically.

No HomeScreen, thread-list, or `package.json` changes. The shiki-touch is a 4-line type extraction only — no startup behavior — so it shouldn't collide with the shiki-startup audit item.

## Verification

- `vp test run` in `apps/mobile`: full package suite green (1688 tests)
- `tsc --noEmit` clean; `vp lint` clean on changed files (remaining warnings are pre-existing on untouched lines)
- Dependency test verified to catch a seeded Android-only type cycle and a seeded lib→features import

Done by callstack/Apex in pi.

---

**Review follow-ups addressed in `8e33b05`:** Sol audit's Android platform-resolution gap (`MaterialSwitch.android.tsx` type cycle) — types extracted, guard now resolves per-platform (iOS + Android) and was verified against that exact seeded mutation; Claude audit's wording note — this PR removes `lib → state` upward edges and the dynamic-import round trip rather than claiming a purely static cycle there.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved composer attachment preview handling so files remain available while previews or uploads are still using them.
  - Preserved attachment cleanup after the final preview or upload completes.

- **Tests**
  - Added dependency checks to help prevent circular relationships and enforce mobile architecture boundaries.
  - Updated attachment-related test coverage for the revised preview-retention behavior.

- **Refactor**
  - Consolidated shared component and dialog type definitions without changing runtime behavior.


juliusmarminge:
**Audit responses (new head `8e33b05`)**

GPT-6 Sol audit (Android-hidden cycle at `3597143`) — confirmed and fixed:
- Root cause correct: the guard checked generic `.tsx` before `.android.tsx`, and `MaterialSwitch.android.tsx` type-imports `ThemedSwitchProps` back from `ThemedSwitch`, which `ThemedSwitch.tsx` imports as a value.
- `ThemedSwitchProps` moved to `components/MaterialSwitch.types.ts` (host re-exports; Android variant imports the types module), so the pair is acyclic under any resolution order.
- `dependency-graph.test.ts` now builds one graph per platform with Metro's candidate order (platform → `.native` → generic) and excludes the other platform's variant files per graph, asserting zero static cycles under **both** iOS and Android resolution. Verified by seeding the exact mutation shape (`X.tsx` → `Y` ← `Y.android.tsx` type-back): fails on Android, passes on iOS. Ceilings now dedupe unique `from -> to` pairs across platforms (same pinned values).

Claude audit (low-severity scope note) — wording corrected: the `state/use-composer-drafts → lib/attachmentUpload` edge was already a dynamic `import()`, so the original state was a bidirectional dependency held apart by the dynamic escape hatch, not a static cycle. The extraction is described as removing the `lib → state` upward edges (and the round trip it forced); the guard's static-cycle claim now explicitly excludes dynamic edges, both here and in the test header.

On the mutable registration hook: kept deliberately. It mirrors the existing `ConfirmDialogHost` module-level registration idiom, the null-handler window is behaviorally identical to the pre-extraction state (no draft store loaded ⇒ the sweep would find no owners), and a callback-parameter API would have put `lib → state` imports back at every call site, which is the edge this extraction exists to remove.

Remaining audit #13 upward edges (thread-order→thread-list model, appearance provider from `lib`/`components`, terminal preferences from `lib`, native→features glue) stay pinned by the ceiling constants and are called out as follow-up in the description; each needs its own owner-side seam.

Verification at `8e33b05`: `vp test run` in `apps/mobile` fully green (1688 tests), `tsc --noEmit` clean, lint clean on changed files.

Done by callstack/Apex in pi.

juliusmarminge:
Final comment sweep at head `5ebd0c9` — no outstanding feedback:

- Fresh independent whole-PR audits (GPT-6 Sol, Claude Opus 5.5): no blockers; platform-aware iOS/Android cycle graph and the dynamic-edge wording correction both verified.
- Inline review comments: none. Reviews: Macroscope approvability **approved** on this head; no change requests.
- Bot reports: thread-transfer impact within every ceiling; all CI checks green on the current head.
- Known optional improvement (not a blocker, deliberately no code change): an explicitly cross-platform import (e.g. a `.ios` file importing a `.android` variant) resolves to nothing in the opposite platform's graph. An all-files check finds zero such cycles today; if one is ever introduced, the guard may miss it. Not worth a follow-up until it earns one.

Marking ready; leaving unmerged for a human to land.

Done by callstack/Apex in pi.

PR #13158, opened 2026-09-22
perf(mobile): enable React Compiler

juliusmarminge:
## The problem

Mobile performance currently depends on manual `memo`/`useMemo`/`useCallback` discipline across the app. That discipline is easy to lose in review and hard to audit. Web already runs `babel-plugin-react-compiler@1.0.0` (`apps/web/vite.config.ts`); mobile has been stuck relying on hand-written memoization.

## The change

Flip `experiments.reactCompiler` on in `apps/mobile/app.config.ts`. That is the entire diff:

- `babel-plugin-react-compiler@1.0.0` is already a hard dependency of `babel-preset-expo@57`, so no package, babel, or metro changes are needed.
- React 19.2 ships the built-in `react/compiler-runtime`, so the target resolves with no extra dependency.
- The compiler only transforms app code, never `node_modules`.
- Manual memoization (~830 `useCallback`/`useMemo`/`memo` call sites) is intentionally untouched — the compiler preserves manual memoization where it can.

Mobile prep for hooks/refs diagnostics already landed on main. It does not resolve every diagnostic — it ensures the remaining diagnostics land in the compiler's non-fatal skip path (a function that cannot be compiled safely is left as written), which is what makes this flip safe.

## What happens to functions the compiler can't handle

`babel-preset-expo` runs the compiler with skip-rather-than-fail semantics in **both** dev and production (prod sets `panicThreshold: NONE`; dev leaves it at the plugin default, which also never panics on per-component diagnostics — only config-level errors throw). A green export therefore proves "no config errors", not "zero diagnostics"; functions with diagnostics are silently left uncompiled. Their original hand-written code ships unchanged — safe, just not optimized.

Measured by running the same plugin (`babel-plugin-react-compiler@1.0.0`, `target: '19'`, `compilationMode: 'infer'`) directly over all 589 non-test `.ts`/`.tsx` files in `apps/mobile/src`:

- **562** functions compile (memoized).
- **45** functions skip on Error-severity diagnostics: 23 Refs (refs accessed during render), 17 PreserveManualMemo, 5 Immutability.
- **29** more skip on unsupported syntax (`try/finally`, dynamic `import()`, `??=` — compiler "Todo" hints).
- 0 babel errors; counts are approximate (standalone scan vs. the full Metro pipeline), but directionally exact.

The remaining refs/manual-memo skips are exactly the non-fatal class the main-side prep work designed for; they can be fixed incrementally afterward and will start compiling on their own once fixed.

## Evidence (iOS, measured on the same machine across on/off checkouts)

- Dev (unminified) bundle: **16,600,271 B → 16,991,312 B** (+391 KB, **+2.35%**). **938** `memo_cache_sentinel` marker occurrences in the shipped bundle. Note: marker occurrences count memo-cache slots and bundle copies, not necessarily distinct functions (the bundle also covers workspace packages beyond `apps/mobile/src`); the occurrence count reproduced identically in a second checkout. The per-function compile/skip counts above are the authoritative function-level numbers.
- Production Hermes bytecode bundle (`.hbc`): **19,686,518 B → 20,006,259 B** (+310 KB, **+1.62%**). Both sizes reproduced byte-identically on re-verification.
- Cold full-export compile time: **unmeasurable delta**. Baseline runs 53/83 s (min/median) vs 57/136 s with the compiler on, inside a {53…136} s spread on this contended shared machine — the noise swamps the signal. Warm builds are unaffected via the Metro transform cache.
- Verification: `expo export --platform ios` green in both dev and prod modes, `tsc --noEmit` clean, lint clean.

Closes the mobile audit item: "enable React Compiler for mobile".

---

Model and harness that did the work: Claude (Anthropic), running as Callstack's Apex agent on the pi harness, orchestrated through T3 Code.


juliusmarminge:
Two corrections to the PR body (code unchanged, still ec1197e):

1. **Attribution** — the implementation was done by Apex (Callstack) on the pi harness, not plain T3 Code; footer fixed.
2. **Diagnostics claim** — the body previously claimed a green dev export proves *zero* compiler diagnostics. That's wrong: `babel-preset-expo` passes `panicThreshold: NONE` in prod and leaves the default (skip, never panic) in dev, so per-component diagnostics never fail the export — only config errors do. I re-measured with a standalone run of the same plugin/target over all 589 non-test `apps/mobile/src` files: 562 functions compile, 45 skip on Error diagnostics (23 Refs, 17 PreserveManualMemo, 5 Immutability), 29 skip on unsupported-syntax Todo hints (try/finally, dynamic import, `??=`). Skipped functions ship their original code unchanged. The body now states this precisely.

juliusmarminge:
Body swept again after the Opus 5.5 whole-PR audit (code unchanged, still ec1197e):

- The 938 figure is restated as **`memo_cache_sentinel` marker occurrences** (slot/bundle-copy count), no longer equated with distinct component/hook functions; the 562-compiled / 45 Error-skipped / 29 unsupported-syntax-skipped per-function counts are flagged as the authoritative function-level numbers.
- The hooks/refs prep sentence no longer implies all diagnostics are resolved — it now says prep routes remaining diagnostics into the non-fatal skip path.

Remaining gates before this is ready to merge (holding; not merging):
1. Human review/approval — Macroscope approvability is 'Not approved' by design for default behavior changes.
2. On-device smoke pass of feed / composer / thread selection / review, requested by the auditor before readiness — the primary agent is running the integrated device pass.
3. CI on the latest commit stays green.

PR #13161, opened 2026-09-22
refactor(mobile): git sheets use uniwind platform variants instead of className ternaries

juliusmarminge:
## Problem

The mobile audit found 389 `Platform.OS` checks in `apps/mobile` while Uniwind's `ios:`/`android:` className variants were used only a handful of times — including 65 inline `Platform.OS === "android" ? "..." : "..."` className ternaries. The git sheet family was one of the densest clusters.

## Change

Converts every inline platform className ternary in the git sheet family to a single class string using Uniwind `ios:`/`android:` variants, following the pattern already used in `apps/mobile/src/features/settings`:

- `GitCommitSheet`, `GitBranchesSheet`, `GitConfirmSheet`, `GitOverviewSheet`, `gitSheetComponents`
- Shared classes stay in base; classes that differ per platform move under `ios:`/`android:` prefixes, so the two platforms never fight over the same utility. Where the converted call site renders an `AppText`/`AppTextInput`, the component's own defaults (`font-sans text-foreground`) stay in the base layer and the new `ios:`/`android:` utilities override them through Uniwind's variant cascade — the same mechanism the existing settings-screen `android:` usage relies on.
- Structural platform branches (android-only headers, checkbox views, dividers), style objects, sheet options, and icon-size props are untouched — those are real platform logic, not styling.

## Measurements

| Metric | Before | After |
| --- | --- | --- |
| `Platform.OS` occurrences in `apps/mobile/src` | 385 | 336 |
| className ternaries in the five git-sheet files | ~49 | 0 |
| `ios:`/`android:` variant utilities in `apps/mobile/src` | 11 | 222 |

HomeScreen and ThreadNavigationSidebar were deliberately left alone (active recycler/error work). ReviewSheet/AddProjectScreen can follow the same pattern in a separate pass.

## Verification

- `tsc --noEmit` (apps/mobile): clean
- `vp lint apps/mobile/src/features/threads/git/`: 0 errors; the four react-compiler warnings are pre-existing (identical on the HEAD versions of the same files)
- `vp test run src/features/threads/git`: passing
- `expo export --platform ios --dev false`: succeeds, and the exported bundle contains compiled selector artifacts for the new variant classes (e.g. `ios:border-border-subtle`, `android:active:bg-subtle`), confirming Uniwind compiles them rather than dropping them
- Before/after device evidence per platform (iOS + Android) through the overview, branches, commit, and confirm sheets is still required before merge and will come from the integrated mobile pass. One item to inspect explicitly there: the `GitConfirmSheet` description line, where base `leading-normal` coexists with per-platform text sizes (`android:text-base` / `ios:text-sm`) — Tailwind 4 emits a paired line-height on `text-*`, so the effective line-height (original: multiplier 1.5 from `leading-normal`) needs a visual confirmation on both platforms.

## Surface checklist

- Clients: mobile only; no wire contract, no web/desktop impact.
- Providers: n/a.
- Reverse states / connection modes: n/a (pure presentational class refactor).

---

Model: Claude (Apex by Callstack, pi harness).

---

<a href="https://app.devin.ai/review/pingdotgg/t3code/pull/13161" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

* **Refactor**
  * Updated Git-related mobile sheets to use platform-specific styling consistently across Android and iOS.
  * Preserved existing layouts, visual appearance, validation, loading states, navigation, and Git actions.


juliusmarminge:
## Root cause found — closing this PR (revert pushed: `f317a434`)

Both device-pass regressions (iOS commit-footer shift, Android blank sheets) are real and share one root cause: **`ios:`/`android:` variants are not platform-guarded in this repo's uniwind 1.11.0 pipeline**, so the premise of this refactor does not hold.

### Evidence

Compiled stylesheet embedded in an `expo export --platform ios` bundle of `5c191ec` (production-like Metro pipeline) contains the Android entries **in the iOS bundle**, with their platform metadata stripped:

```json
"android:p-4": [{ "entries": [["padding", … * 4]], "native": false, "complexity": 0, … }]
"android:bg-sheet-solid": [ … "native": false … ]
"ios:flex-1": [ … "native": false … ]
```

- `addMetaToStylesTemplate` is supposed to drop styles where `platform !== currentPlatform`, but the entries arrive with `platform: null` (the `native: false` field is literally `platform !== null`), so nothing is dropped.
- The runtime store (`node_modules/uniwind/src/core/native/store.ts`) has **no platform check** in its style filter — guards cover only minWidth/maxWidth/theme/orientation/rtl/active/focus/disabled/dataAttributes.
- All variant tokens compile with `complexity: 0`, so the only remaining tie-break for a shared property is **last token in the className string wins — identically on both platforms**.

### Why each device pass saw what it saw

- **iOS commit block +14px down/right, gray input:** commit-message wrapper's `android:p-4` leaks (no iOS counterpart to overtake it; with `polyfills: { rem: 14 }` in `metro.config.js`, `p-4` = 14px) and TextInput's `android:bg-sheet-solid` leaks. Branch/overview/confirm matched only because the leaked `android:p-4`/`android:gap-3` was masked by later iOS utilities for the same properties.
- **Android blank sheets:** `className="android:shrink android:grow-0 ios:flex-1"` — `ios:flex-1` is the last token, so on Android the ScrollView also gets `flex-basis: 0% + grow 1`, collapsing content to zero height in the `fitToContents` sheet. Same for the root `bg-sheet ios:flex-1`.

### Impact beyond this PR

This also affects already-shipped `android:` usages (they apply on iOS today, masked or subtle): `SettingsControlRow` (`android:gap-1`, `android:text-base`), `SettingsServerControlsRouteScreen` (`android:px-4`), `SettingsControlRow` layout (`android:min-h-14 android:py-3`). Worth its own repro + fix (uniwind platform tagging, likely in the MQ/mediaType handling of the 1.11.0 pipeline or the lightningcss version it runs against) before any `ios:`/`android:`-based migration is attempted again.

Revert commit restores the exact ternaries; branch history is kept so the conversion can be re-attempted once platform variants are actually guarded. Net PR diff after the revert is zero — there is nothing to dual-audit here anymore.

PR #13167, opened 2026-09-23
fix(web): preserve nested scroll behavior in chat timeline

Yash-Singh1:
Scrolling inside an expanded tool group could collapse the composer even though the chat viewport had not moved. The composer listened for wheel events at the document level and treated any event inside the chat as timeline navigation.

The scroll handlers now check whether a nested scroll area can consume the gesture. The composer and live-follow behavior respond only when the gesture can reach the chat viewport. The same check covers keyboard navigation and scroll containment.

| Before | After |
| --- | --- |
| Composer collapsed after scrolling a tool group | Composer stays expanded after scrolling a tool group |

[Before recording](https://github.com/user-attachments/assets/8d8f33cf-ff09-496c-91d6-ebacc9d8e390) · [After recording](https://github.com/user-attachments/assets/2fe81fa9-88ba-46b9-93fa-4bd7f0aed0f9)

The recordings use the same dispatched wheel event and 30 px nested scroll because the Browser panel's scroll control changes position without emitting a wheel event. In the baseline, the composer shrinks from 178 to 84 px while the chat scroll position stays fixed. With the fix, it stays at 178 px.

Validation: 170 focused web tests pass; web typecheck passes. The baseline and fixed behavior were recorded in an isolated web client.

Generated with GPT-6 in Codex.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved timeline scrolling and live-follow behavior for wheel and keyboard navigation.
  - Prevented browser zoom gestures and unrelated targets from affecting timeline interactions.
  - Improved handling of nested scrollable areas, scroll boundaries, and overscroll behavior.
  - Prevented nested scrolling from incorrectly triggering composer collapse gestures.

- **Tests**
  - Added coverage for timeline target detection, nested scrolling, edge chaining, overflow behavior, and gesture handling.


Yash-Singh1:
Before: scrolling inside the expanded tool group collapses the composer while the chat viewport stays put.

https://github.com/user-attachments/assets/8d8f33cf-ff09-496c-91d6-ebacc9d8e390

Before screenshot

After: the same nested scroll leaves the composer expanded while the chat viewport stays put.

https://github.com/user-attachments/assets/2fe81fa9-88ba-46b9-93fa-4bd7f0aed0f9

After screenshot

Both recordings apply the same wheel event and 30 px nested scroll. The Browser panel's scroll control changes scroll position without sending a wheel event, so the event was dispatched in the page for each recording.

PR #13179, opened 2026-09-23
fix(cli): t3 update points apt and dnf installs to their package manager

t3dotgg:
When a system package owns `t3` (for example `/usr/bin/t3 -> /opt/t3code-cli/t3`), `t3 update` still downloads a copy into `~/.t3/runtime/versions`. It cannot repoint `/usr/bin/t3`, and it can move the background service onto the copy. The command and the service then silently run different versions.

`t3 update` now reads a one-line `package-type` file next to the real executable. If the file says `deb`, `rpm`, or any other value, the command stops before any network or disk change and prints the right command, for example `sudo apt update && sudo apt upgrade`. Archive and npm installs have no marker, so nothing changes for them.

This makes a future `t3code-cli` Debian and RPM package safe. [#13177](https://github.com/pingdotgg/t3code/pull/13177) does not ship that package yet. Remote self-update needs no change: the service always runs its pinned copy under the T3 home, not the package's executable.

Tests: `update.test.ts` passes (5 tests). The new tests cover detection for `deb`, `rpm`, unknown, and missing markers, and a refusal that makes no HTTP, process, or service calls.

Created with Claude Fable 5.1 in Claude Code.

## Summary by CodeRabbit

* **Bug Fixes**
  * The update command now detects installations managed by Debian or RPM package managers.
  * Package-managed installations are no longer updated through the application’s built-in updater.
  * Users receive package-manager-specific instructions to update these installations safely.


t3dotgg:
> [!NOTE]
> 🤖 **Claude Opus 5.5 responding on behalf of Theo**

Closing. This only mattered for a `t3code-cli` system package, and [#13177](https://github.com/pingdotgg/t3code/pull/13177) does not ship one. The branch stays. If we add that package later, this needs the open review note fixed: a marker file that exists but cannot be read must not count as "no marker".

PR #13183, opened 2026-09-23
refactor(mobile): retire the legacy grouped thread list

juliusmarminge:
## The problem

The flat Thread List v2 has been the default on every app variant for a while; the grouped legacy list survives only behind **Settings → Legacy → Legacy Thread List**. As long as it exists, the mobile home screen and the navigation sidebar each maintain **two** complete list pipelines: two row components, two item builders, two equality functions, plus sort/group menu state that v2 silently ignores. Every perf and behavior fix has to be thought through twice, and one of the two paths is dead code that still ships.

This PR deletes the v1 path outright (following up on #13149, which made the v2 list the fast path — this PR makes it the only path).

## What is removed

- **Home screen + sidebar**: the v1 memo pipeline (project groups, list layout, collapsed-group state), the `threadListV2Enabled` ternaries, and the v1 `LegendList` render path. Only the v2 `LegendList` remains.
- **Files deleted**: `thread-list-items.tsx` (v1 rows), `homeListItems.ts` (v1 item builder) and their tests, `buildHomeThreadGroups` + its tests, the `use-thread-list-v2-enabled` hook and `resolveThreadListV2Enabled`.
- **Preference**: `legacyThreadListEnabled` is removed from the mobile preferences schema; sanitizing drops the stored key, so opting back is gone everywhere (same one-way door we used when the v2 flag flipped to default-on). The Legacy settings section stays for Plan Mode.
- **Menus**: header/filter menus keep environment + project filters. The "Sort projects" / "Sort threads" submenus were already hidden while v2 was on (v2 uses fixed creation order and never honored them), so the controls and their handler plumbing go away. `threadSortOrder` no longer exists on the mobile side at all; **`projectSortOrder` and `projectGroupingMode` remain** and keep ordering/grouping project scopes exactly as v2 uses them today.
- Keyboard thread-jump items are now just the v2 item union.

## Behavior notes

- No visual change for anyone on the default (v2) list; the home list, shelves, swipe menus, search, filters, and sidebar are byte-identical in behavior.
- A device that had explicitly opted into the legacy list will land on the default flat list after this ships (the stored preference key is dropped). That is the intended one-way door.
- The "customized" filter-icon state previously counted custom sort orders; it now keys off the environment/project filters only — matching what the menus can actually change.

## Evidence

- `apps/mobile`: **1682 tests pass** (183 files), `tsc --noEmit` clean, lint has no new warning classes (file-level count parity checked against the branch base; one heuristic `react(memo-dependencies)` extra-deps warning remains in `ThreadNavigationSidebar.tsx` alongside ten pre-existing ones of the same class).
- Removed-test accounting: v1 group/layout builder tests were deleted with their code; the surviving `homeThreadList` tests (project scopes/sorting) were preserved and renamed; the `hasCustomHomeListOptions` test file went with its function.
- Device verification (iPhone 16 Pro iOS 18.6 + iPad Pro 11 M5 iOS 26.5 dev client, same native fingerprint `cd6f4eb6` reused — no native delta): home list renders and scrolls to end without recycle artifacts; settled/snoozed shelves toggle and render; row swipe reveals Settle/Snooze actions on a recycled `LegendList` cell; filter menu opens with only Project + Environment (no sort/group items); Thread behavior settings show the Legacy section with Plan Mode only — the Legacy Thread List row and toggle are gone; iPad split shell sidebar list renders, scrolls, and selection opens the detail pane. No redboxes or list-pool warnings during the pass. Functional pass, not exhaustive stress testing.
- Follow-up in the same PR (`f9ff0fe`): the three filter-button accessibility labels that still said "Filter and sort threads" were renamed to "Filter threads" since sorting is no longer offered.

## Stack

This stacks on #13149 (both rewrite the same list-render sections); merge that one first, or merge this one against `main` after it lands.

---

Model: Claude (Apex via pi), doing the work; base branch `agent/mobile-audit-list-recycler` (PR #13149).

---

<a href="https://app.devin.ai/review/pingdotgg/t3code/pull/13183" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

- **Updates**
  - The mobile home screen and sidebar now use a flat thread list, with environment and project filters.
  - Project and thread sorting controls and the “Legacy Thread List” setting are no longer available.
  - Keyboard shortcuts navigate the flat list, and filter controls are labeled “Filter threads.”
  - Search, project scoping, pending tasks, and existing thread actions remain available; archived threads remain excluded.


juliusmarminge:
Follow-up for the audit cleanup landed as #13203 (both layers of this stack merged while it was being prepared, so the cleanup targets `main` directly): deletes the `homeThreadList.ts` unused imports and recency-window constants, the `threadSortOrder` preference + unused setter in `home-list-options.ts` (making this PR body's claim true end to end), the stale test helpers/import, and the legacy-list comment. `no-unused-vars` clean, tsc clean, mobile suite 186/1702.

PR #13201, opened 2026-09-23
test(server): stop ACP soft-steer tests racing the prompt settle

saphid:
## What Changed

Added an `afterPromptSettledWithBackgroundWork` test hook to `AcpAdapterV2`. It fires once the adapter records a settled prompt while background work is still running. Three `AcpAdapterV2` tests now wait on that hook before they send the soft-steer interrupt. Before, they waited for the raw `stopReason` frame and then did two `Effect.yieldNow` calls. Test-only change: the hook defaults to `Effect.void`.

## Why

`Stop after settled soft steer contains the orphan runtime and respawns without subagent carryover` is an `it.live` test against the mock ACP agent subprocess. The raw `stopReason` frame is logged before the adapter records the settle. Two yields do not guarantee the adapter has caught up. When the interrupt arrives first, the adapter still treats the prompt as running and sends `session/cancel`, and the test fails:

```
AssertionError: settled soft steer must not send session/cancel before the later Stop: expected true to be false
```

This happened on 2 of 2 recent Linux CI runs of our fork's nightly build of this branch. On macOS it passed 14 of 14 runs, including 8 runs under CPU load. Two `it.effect` tests use the same wait pattern, so they get the same fix.

## Verification

- **Reproduced the failure:** added a scratch 100 ms delay before the adapter signals the settle. The unchanged test then fails every time with the assertion above.
- **Fix holds with the delay:** with the fix and the same delay, the test passes.
- **Full file, delay removed:** `vp test run src/orchestration-v2/Adapters/AcpAdapterV2.test.ts` passes 110/110.
- `apps/server` typecheck and `vp fmt --check` pass. `vp lint` reports no new warnings.

## Checklist

- [ ] I included before/after screenshots for any UI changes (no UI changes)
- [ ] I included a video for animation/interaction changes (no UI changes)

Fixed by Claude Opus 5.5 in Claude Code, running in T3 Code.


saphid:
@juliusmarminge could you merge this when you get a chance? It's test-only: three `AcpAdapterV2` tests wait on a settle hook instead of `yieldNow`. The `it.live` one failed on Linux on its own.

All server test shards pass. The one red check, `Check`, fails the same way on `t3code/codex-turn-mapping` at `060756de5ad` without this change: `lint:restyle-ceiling` reports 659 web `className` overrides against a ceiling of 628. It was still green on #13113, so the breach landed on the branch after that.

PR #13224, opened 2026-09-23
fix(server): show usage limits for every Claude instance

### LINKED ISSUE #13053: Second Claude account's usage never appears in the Limits tab — silently marked unsupported, never recovers
### What happened

> In the pricing tab, if i have 2 claude instances the token limit/usage doesn't show for both of them.

Two Claude provider instances are configured, each a separate Anthropic account, each with its own `CLAUDE_CONFIG_DIR` (`homePath`). Only one of the two ever shows a bar in Usage → Limits. The second instance never appears there — no bar, no warning, no error — even though it's actively used and has hit its own real rate limits.

### Diagnosis

Traced through `apps/server/src/provider/Layers/claudeUsageLimits.ts` (`claudeUsageResponseToLimits`) at commit `719a76ca` (tag `v0.0.42`):

```js
if (!response.rate_limits_available || !response.rate_limits) {
  return { limits: makeUnavailableUsageLimits({ checkedAt, reason: "unsupported" }), ... };
}
```

For the second instance, the boot-time `get_usage` SDK probe evidently reports `rate_limits_available: false` (inferred from client behavior — the raw probe response body isn't logged anywhere, so I could not confirm exactly which condition on that instance triggers it; the instance uses a distinct `CLAUDE_CONFIG_DIR` and also has a `CLAUDE_CODE_OAUTH_TOKEN` environment override configured, but the reporter's actual sign-in for that instance was interactive, not token-based, so the trigger is unconfirmed). Whatever the trigger, once it fires, T3 marks the instance `"unsupported"` — the same bucket used for API-key accounts that genuinely have no subscription. That is silent by design in two places:

- `collectLimitAccounts` (`packages/shared/src/usageLimits.ts`) skips any provider whose `usageLimits` carries a notice, including `"unsupported"` — no bar drawn.
- `collectLimitNotices` explicitly `continue`s past `reason === "unsupported"` — no warning shown either.

This traps the instance permanently: `applyUsageLimitsUpdate` (`apps/server/src/provider/providerUsageLimits.ts`) refuses to apply later updates once a snapshot is marked `"unsupported"`, on the assumption "an account that cannot have subscription windows will not start reporting them mid-turn." That assumption doesn't hold here — the provider event log shows genuine mid-turn `rate_limit_event` data from this same instance (`five_hour` at 109% utilization, `seven_day` at 45%), proving the account does have real, trackable quota. That data is discarded on arrival because the snapshot is already locked to `"unsupported"`.

Net effect: a second Claude account with real subscription limits is indistinguishable, in this code path, from an account with no subscription at all, and disappears from the Limits tab with zero user-facing feedback — regardless of which auth method it used to sign in.

Checked upstream issue #12170, which covers a related but distinct bug (a transient `probeFailed` state after a failed boot probe, which self-heals within ~5 minutes via periodic refresh). This case doesn't self-heal — every fresh probe re-derives the same `"unsupported"` state, so it persists indefinitely across days of use. Not a duplicate.

### Steps to reproduce

1. Configure a second Claude provider instance signed into a different Anthropic account than the first, using its own `homePath` (separate `CLAUDE_CONFIG_DIR`).
2. Open Usage → Limits.
3. Observe: only one of the two instances ever gets a bar or a notice, before or after refreshing.
4. Run a turn on the instance with no bar until it hits a rate limit. The mid-turn warning carries real utilization data, but the Limits tab still shows nothing for that instance.

### Version

0.0.42 (also confirmed present in the currently-running desktop build, `0.0.43-nightly.20260919.1978` — no relevant commits landed between the two)

### Environment

Linux x64, Node v26.8.2, claude CLI 2.1.278

### Evidence

```
[2026-09-20T18:58:27.751Z] CANON: {"type":"runtime.warning","provider":"claudeAgent","payload":{"message":"Claude usage limit reached. This turn is paused until the 5-hour limit resets in 2h 12m.","detail":{"status":"rejected","resetsAt":1789938600,"rateLimitType":"five_hour","overageStatus":"rejected","overageDisabledReason":"org_level_disabled","isUsingOverage":false,"unifiedWindows":{"five_hour":{"utilization":1.09,"resetsAt":1789938600},"seven_day":{"utilization":0.45,"resetsAt":1789938000}}}},"providerInstanceId":"claudeAgent_bagri"}
```

### Related issues

#12170 — related mechanism (probeFailed → sparse window merge), but transient/self-healing, unlike this permanent `"unsupported"` lock. Not a duplicate.

### Fix applied or workaround

None found.

### Filed by

claude (sonnet-5) via t3 triage

macodev00:
## What Changed

A second Claude provider instance is no longer dropped from Usage → Limits when its boot `get_usage` probe comes back without windows.

- `rate_limits_available: true` with a null `rate_limits` body is `probeFailed`, not `unsupported`.
- A subscription login (including a second account on its own `CLAUDE_CONFIG_DIR`) that comes back `rate_limits_available: false` is also `probeFailed`. API-key, Bedrock, and Vertex logins stay `unsupported`.
- A later `rate_limit_event` fills those windows in. The snapshot stays marked `probeFailed` so one window is not treated as a finished read, and the next empty probe does not wipe the windows.
- The Limits tab draws whatever windows are already known. An empty failed probe still shows “Could not read limits.” The probe logs `rate_limits_available` and whether `rate_limits` was present.

Fixes https://github.com/pingdotgg/t3code/issues/13053

## Why

Two Claude accounts each have their own quota. The second one could be classified `unsupported` at boot — the same bucket as an API key — and then omitted from Limits with no bar and no notice. Mid-turn utilization was discarded, and every later probe locked that state in again.

## UI Changes

No layout change. The existing Limits bars and notice are used. A second subscription instance with a failed usage probe shows the notice until a turn reports windows, then those windows are drawn. Screenshot needs two signed-in Claude accounts, which this change does not have locally.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Eligible Claude subscription accounts reporting unavailable or unreadable usage data now show a failed limits check instead of being treated as having no quota.
  - Claude usage checks now distinguish recognized subscription accounts from unsupported account types.
  - Previously reported usage windows remain visible when a later limits check fails, and new window updates can recover limits previously marked unsupported.
  - Accounts with available usage windows no longer show an unnecessary limits notice, even if a subsequent check fails.
  - When limits cannot be read and no usage windows are available, the account is surfaced with a clear limits-read error.


macodev00:
@coderabbitai review

@macroscope-app review

macodev00:
@coderabbitai review

@macroscope-app review

macodev00:
@coderabbitai review

macodev00:
@macroscope-app review

macodev00:
@coderabbitai review

@macroscope-app review

macodev00:
@coderabbitai review

macodev00:
@macroscope-app review

macodev00:
@coderabbitai full review

PR #13097, opened 2026-09-22
fix(server): keep Claude sessions up while child tasks run

macodev00:
## What
`ClaudeAdapter.startSession` refuses to replace an existing Claude session while child tasks are still running (emits `runtime.warning`, returns `ProviderAdapterRequestError`, leaves the open query up).

`ProviderService.startSession` snapshots the running thread’s MCP session and issues the replacement credential with `retainExisting` so `revokeThread` does not run first. If the adapter rejects and the live session remains, only the new `providerSessionId` is revoked and the previous session config is restored. An accepted replacement revokes the previous credential after the new session returns.

## Why
Closes #12694. Redo after #13086 (Macroscope Not approved — High: rejected `startSession` still revoked the MCP credential used by the live Claude query). Also supersedes earlier #12723 / #13080 eligibility/breadth closes.

## UI
N/A (server provider / MCP session)

## Linked Issues
- Closes #12694

## Checklist
- [x] I read CONTRIBUTING.md and AGENTS.md documentation requirements
- [x] Scope is a small bug/reliability fix
- [x] Tests updated for ClaudeAdapter + ProviderService / MCP registry
- [x] No Co-authored-by / Cursor branding in commits

## Summary by CodeRabbit

- **New Features**
  - MCP credentials can remain active while a provider session replacement is pending.
  - Provider session replacements now preserve the existing active session when live tasks are running.
  - Session replacement automatically switches to the new session after existing tasks finish.

- **Bug Fixes**
  - Prevented active sessions and credentials from being lost when a replacement is rejected.
  - Improved credential cleanup after successful, failed, or refused session starts.
  - Added warnings when replacement is blocked by running tasks.

PR #13225, opened 2026-09-23
fix(server): keep live Claude sessions authorized during replacement

### LINKED ISSUE #12694: [Bug]: Claude session replacement suppresses exit and silently stops live subagent tasks
## Before submitting

- [x] I searched open and closed issues for the replacement warning, suppressed
  exit behavior, stopped live tasks, and `thread.turn.start`; I found no
  matching report.
- [x] This report includes the observed event chain and source locations needed
  to investigate it.

## Area

`apps/server`

## Steps to reproduce

1. Start a Claude provider session and leave one or more background/subagent
   tasks live.
2. Submit `thread.turn.start` in a condition where
   `ProviderCommandReactor.ensureSessionForThread` chooses its restart branch.
   At commit `c7fa1e96`, this happens when runtime mode, cwd, provider instance,
   an unsupported model switch, or Claude model-selection equality requires a
   restart.
3. The reactor calls `providerService.startSession` while the adapter still has
   an existing context for that thread.
4. Inspect orchestration events and provider session state.

The narrower predicate that triggered each observed replacement had already
rotated from the local trace, so this report does not claim which predicate was
true. The replacement/stop chain itself was observed twice: once while the old
session was `running` with four live tasks, and once while it was `ready` but
still owned one live task.

## Expected behavior

An incoming turn should reuse the active provider session when compatible. If a
replacement is required, T3 should emit a durable, reasoned session-boundary
event (or reject/defer the replacement) so the orchestration timeline and the
seat can detect that the old session ended and that live child tasks were
stopped.

## Actual behavior

`ClaudeAdapter.startSession` detects the existing context, logs
`claude.session.replacing`, and calls
`stopSessionInternal(existingContext, { emitExitEvent: false })`.
`stopSessionInternal` closes the query, records every `liveTaskId` as
`task.completed(status=stopped)`, and suppresses `session.exited`. A replacement
session then starts, but there is no `thread.settled`,
`thread.session-stop-requested`, stopped session row, provider error, or other
durable boundary explaining why the live tasks disappeared. In the observed
cases, the seat learned about the replacement only because its provider harness
name changed.

## Impact

Major degradation or frequent failure. An ordinary incoming turn can silently
kill in-flight subagents while the parent thread continues, and the durable
session projection does not expose the boundary needed for recovery.

## Version or commit

`0.0.43-nightly.20260918.1895`, inspected source commit
`c7fa1e96943f380a8893e89460d03a362285ab83`.

## Environment

Self-hosted Linux container; Claude provider (CLI 2.1.277). The container stayed
healthy and did not restart during either event.

## Sanitized evidence

```text
thread.turn-start-requested
claude.session.replacing existingSessionStatus=<running|ready>
task.completed status=stopped  # four tasks in one event, one in the other
replacement session starting/running

Absent from both chains:
thread.settled
thread.session-stop-requested
session.exited / durable stopped session row
provider error
container restart
```

Relevant source at the inspected commit:

- [`ProviderCommandReactor` restart predicates and decision log](https://github.com/pingdotgg/t3code/blob/c7fa1e96943f380a8893e89460d03a362285ab83/apps/server/src/orchestration/Layers/ProviderCommandReactor.ts#L747-L812)
- [`ClaudeAdapter.startSession` replacement with `emitExitEvent: false`](https://github.com/pingdotgg/t3code/blob/c7fa1e96943f380a8893e89460d03a362285ab83/apps/server/src/provider/Layers/ClaudeAdapter.ts#L4394-L4414)
- [`stopSessionInternal` stops live tasks and conditionally suppresses `session.exited`](https://github.com/pingdotgg/t3code/blob/c7fa1e96943f380a8893e89460d03a362285ab83/apps/server/src/provider/Layers/ClaudeAdapter.ts#L4262-L4364)

## Workaround

No safe server-side workaround is known. Operationally, background work must be
made resumable and reconciled after an incoming turn. Avoiding replacement-capable
turn delivery prevents the symptom but is not a product fix.

macodev00:
## What Changed

`ClaudeAdapter.startSession` refuses to replace an open Claude query while child tasks are still running. It emits `runtime.warning` and returns `ProviderAdapterRequestError`, and leaves that query up.

`ProviderService.startSession` reads the thread's current MCP provider session before it probes `hasSession`. When that session belongs to a different provider instance, it probes that instance's adapter. The replacement credential is issued with `retainExisting`, so the live query's bearer token stays registered until the new start is accepted.

Settlement then depends on the outcome:

- A rejected or failed start that leaves the previous session up revokes only the new credential and restores the previous config.
- A `hasSession` failure after the old session was already stopped is treated as no active session, so the old bearer token is revoked too.
- An accepted replacement revokes every credential for the thread except the one just issued.
- Starts for the same thread take a lock from the credential snapshot through settlement, so a later success cannot leave an earlier token authorized.

This supersedes #13097, which Macroscope did not approve. That revision treated a post-replacement `hasSession` failure as "session still active" (so the stopped query's token stayed registered) and revoked only the snapshotted previous credential (so overlapping `startSession` calls could leave the original token authorized).

## Why

Closes #12694. An ordinary turn was able to replace a live Claude session, stop its child tasks, and suppress `session.exited`. Refusing that replacement keeps the tasks running. The credential changes are what make the refusal safe: the open query must keep the MCP token it already baked in, and a stopped or superseded query must not keep authorizing `/mcp`.

## UI Changes

No UI changes.

## Linked Issues

- Closes #12694

## Checklist

- [x] No UI changes, so no screenshots or video
- [x] I read CONTRIBUTING.md and AGENTS.md documentation requirements
- [x] Scope is a small bug/reliability fix
- [x] Tests cover Claude live-task replacement, refused and accepted MCP rotation, a failed session probe after the old session stopped, overlapping starts, and a rejected switch to another provider instance
- [x] No Co-authored-by / Cursor branding in commits

## Summary by CodeRabbit

* **Bug Fixes**
  * Claude sessions with active child tasks can no longer be replaced; the replacement is rejected while the existing session remains active. Replacement is allowed after those tasks finish.
  * MCP credentials are now handled more safely during session starts and replacements: valid credentials are preserved when a replacement is rejected, and superseded credentials are revoked when appropriate.
  * Concurrent session starts for the same thread are serialized to prevent credential conflicts.


macodev00:
@coderabbitai review

@macroscope-app review

macodev00:
Superseding: Macroscope **Not approved** (High — when previous provider instance is gone, failed start restores stale `McpProviderSession` / keeps stopped process bearer via `keepLiveMcpCredential` on `!ownerLocated`; must revoke+clear instead).

Closing in favor of a fresh READY redo from upstream `main` addressing that High (issue #12694). This is MS close+redo **2/2** for #12694 today (Europe/Ljubljana).

PR #13226, opened 2026-09-23
fix(mobile): Android terminal keystrokes paint immediately

### LINKED ISSUE #12756: [Bug]: Android: terminal typing echo is abnormally slow (per-keystroke delay)
### Before submitting

### Area

apps/mobile

### Steps to reproduce

1. Open a thread in the Android client.
2. Open the terminal view and tap into the input.
3. Type a sequence of characters.
4. Observe the delay between each key press and the character appearing on screen.

### Expected behavior

Keystrokes should echo on screen immediately — imperceptibly, comparable to a native or desktop terminal. Typing in the terminal should feel instant.

### Actual behavior

There is a consistently noticeable delay between tapping a key and the character appearing in the terminal. It happens on every keystroke, so typing feels abnormally slow overall.

Worth checking as likely causes: whether every keystroke waits for a full network round-trip before rendering (no local echo / no optimistic echo), and per-keystroke re-render overhead in the terminal view itself.

### Impact

### Environment

- Device: Xiaomi 15 on HyperOS 4 beta 3 (CN ROM)
- Platform: Android client (current build as of 2026-09-20)

---
🤖 GLM 5.3 Flash on behalf of iliasgws

macodev00:
## What Changed

Android terminal keystrokes now paint in the native surface before the remote pty echoes them. Printable input is fed to Ghostty immediately, then reconciled with the authoritative pty byte stream so a matching echo is drawn once. If the pty does not echo those bytes, the grid is rebuilt from the pty buffer and local echo stays off until a later keystroke is echoed verbatim. Password and passphrase prompts are left for the pty.

Fixes https://github.com/pingdotgg/t3code/issues/12756

## Why

Characters were drawn only after a full attach round trip: the hidden Android field emitted the key, the client wrote it to the pty, and the glyph appeared when that output came back and the whole buffer was fed into Ghostty. On a phone that delay is visible on every key, so typing feels slow. Web and desktop use the same remote-echo model and usually sit next to the server, which is why they still feel instant.

## UI Changes

This changes when typed characters appear in the Android terminal. Layout is unchanged. This environment has no Android emulator (`/dev/kvm` is unavailable) and no device, so there is no before/after recording. Reconciliation is covered by `TerminalLocalEchoTest`.

iOS still paints from the remote buffer. This report is the Android client.

## Checklist

Screenshots do not show the latency. A recording needs a device build of the native terminal module.

## Summary by CodeRabbit

* **New Features**
  * Terminal keystrokes now appear immediately while the remote session responds, making typing feel more responsive.
* **Bug Fixes**
  * Prevented typed characters from appearing twice when the terminal echoes them back.
  * Improved recovery when remote output differs from the locally displayed input, keeping the terminal display in sync.
  * Avoided locally displaying input at password prompts, where characters should remain hidden.


macodev00:
@coderabbitai review

@macroscope-app review

macodev00:
Superseding: Macroscope **Not approved** (High×2 on Android local echo — unbounded trusted `predicted` growth; `PASSWORD_PROMPT` misses Passcode:/PIN: so secrets can paint locally).

Closing in favor of a fresh READY redo from upstream `main` for issue #12756. MS close+redo **1/2** for #12756 today (Europe/Ljubljana).

PR #13233, opened 2026-09-23
fix(mobile): Android terminal keystrokes paint immediately

macodev00:
## What Changed

Android terminal keystrokes now paint in the native surface before the remote pty echoes them. Printable input is fed to Ghostty immediately, then reconciled with the authoritative pty byte stream so a matching echo is drawn once. If the pty does not echo those bytes, the grid is rebuilt from the pty buffer and local echo stays off until a later keystroke is echoed verbatim.

Trusted prediction stops at the same pending-input cap as the untrusted probe, and further keystrokes fall back to remote echo. Password, passphrase, passcode, and PIN prompts, including ones with a URL or a parenthetical, are left for the pty. A control key holds local echo until the pty answers, and a rebuild does not resend device replies for history the previous session already answered.

Fixes https://github.com/pingdotgg/t3code/issues/12756

Supersedes https://github.com/pingdotgg/t3code/pull/13226

## Why

Characters were drawn only after a full attach round trip. On a phone that delay is visible on every key. The previous attempt (#13226) was not approved because a trusted session could grow the predicted buffer without limit, and the prompt check missed `Passcode:` and `PIN:`.

## UI Changes

This changes when typed characters appear in the Android terminal. Layout is unchanged. This environment has no Android emulator (`/dev/kvm` is unavailable) and no device, so there is no before/after recording. Reconciliation is covered by `TerminalLocalEchoTest`.

iOS still paints from the remote buffer. This report is the Android client.

## Checklist

- [x] This PR is small and focused
- [x] I explained what changed and why
- [ ] I included before/after screenshots for any UI changes
- [ ] I included a video for animation/interaction changes

Screenshots do not show the latency. A recording needs a device build of the native terminal module.

## Summary by CodeRabbit

* **New Features**
  * Terminal input now appears immediately as you type, rather than waiting for the remote session to respond. The display updates when the session confirms your input and corrects itself if the remote output differs.
  * Local echo is held back for control input and detected password, passcode, or PIN prompts to avoid showing sensitive typing.

PR #13238, opened 2026-09-23
fix(mobile): Android terminal keystrokes paint immediately

### LINKED ISSUE #12756: [Bug]: Android: terminal typing echo is abnormally slow (per-keystroke delay)

macodev00:
## What Changed

Android terminal keystrokes now paint in the native surface before the remote pty echoes them. Printable input is fed to Ghostty immediately, then reconciled with the authoritative pty byte stream so a matching echo is drawn once. If the pty does not echo those bytes, the grid is rebuilt from the pty buffer and local echo stays off until a later keystroke is echoed verbatim.

Trusted prediction stops at the same pending-input cap as the untrusted probe, and further keystrokes fall back to remote echo. Password, passphrase, passcode, and PIN prompts, including ones with a URL or a parenthetical, are left for the pty. A control key holds local echo until the pty answers, and a rebuild does not resend device replies for history the previous session already answered.

This supersedes #13233 (Macroscope not approved) and #13226. The three blocking findings from #13233 are baked in:

1. Secret-prompt detection inspects the current line after the last newline or carriage return, then strips ANSI. It does not `takeLast(240)` first, so a long `Password for 'https://…':` line cannot drop the leading `Password` token while keeping the trailing colon.
2. Trusted keys refused at the pending-input cap stay in `pendingInput`. Painting stays off until those bytes are echoed, so a partial echo cannot paint the next key and later reset the grid.
3. `awaitingRemote` stays set while a new snapshot is only the outstanding printable echo (for example `ls` after `ls` + Enter). It clears once CR/LF or other control processing is visible.

Fixes https://github.com/pingdotgg/t3code/issues/12756

Supersedes https://github.com/pingdotgg/t3code/pull/13233 and https://github.com/pingdotgg/t3code/pull/13226

## Why

Characters were drawn only after a full attach round trip. On a phone that delay is visible on every key. #13226 was not approved because a trusted session could grow the predicted buffer without limit, and the prompt check missed `Passcode:` and `PIN:`. #13233 fixed those, then Macroscope rejected long-prompt detection, cap overflow after a partial echo, and clearing `awaitingRemote` on a printable-only suffix.

## UI Changes

This changes when typed characters appear in the Android terminal. Layout is unchanged. This environment has no Android emulator (`/dev/kvm` is unavailable) and no device, so there is no before/after recording. Reconciliation is covered by `TerminalLocalEchoTest`.

iOS still paints from the remote buffer. This report is the Android client.

## Checklist

Screenshots do not show the latency. A recording needs a device build of the native terminal module.

## Summary by CodeRabbit

* **New Features**
  * On Android, printable terminal input can appear immediately as you type. The display reconciles it with terminal responses and avoids duplicating echoed text.
  * Input is not painted ahead for password prompts or control keys that require a terminal response.
* **Bug Fixes**
  * Improved handling of terminal output that differs from predicted input, including preserving scrollback during reconciliation.


macodev00:
@coderabbitai review

@macroscope-app review

macodev00:
Closing — Macroscope **Not approved** (High: `emitTypedInput` paints printable secrets for echo-disabled prompts like `read -s -p 'API token: '`).

Issue #12756 daily MS-redo cap is **2/2 HIT** today (prior redos already used). No further auto-redo this sweep; noting for morning digest. A fresh attempt can resume tomorrow from upstream `main` addressing the High (gate local echo on actual echo-disabled/termios state, not only password-like prompt labels).

PR #13248, opened 2026-09-23
feat(settings): choose the branch prefix for new worktrees

yekta:
## What Changed

Every worktree branch is renamed to `t3code/<generated name>` on its first turn, and the prefix is a constant. This adds a project-scoped `worktreeBranchPrefix` server setting, shown as **Branch prefix** in **Settings → General** next to Workspace and Submodules. Type a namespace such as `yekta` or `team/yekta`, or leave it empty to name the branch without one. The default stays `t3code`.

- `packages/shared/src/git.ts`: `buildGeneratedWorktreeBranchName` moves here from the reactor and takes the prefix; `normalizeWorktreeBranchPrefix` sanitizes what the user typed into a valid ref namespace (keeps case and slashes, drops characters git refuses, `""` means no prefix).
- `packages/contracts/src/settings.ts`: the key in `PROJECT_SCOPED_SERVER_SETTING_KEYS`, `ProjectSettingsOverrides`, `ServerSettings` (lenient `TrimmedString`, default `"t3code"`) and `ServerSettingsPatch`. Kept lenient on purpose so a hand-edited value can never fail the whole settings snapshot.
- `ProviderCommandReactor`: the first-turn rename reads the prefix through the existing `projectSettingsForThread`, so a project override wins over the environment value.
- Web: a `DraftInput` row in `ProjectDefaultsSettings` with the usual reset arrow, mixed state and layers popover (`No prefix` label for the empty value), plus a searchable catalog entry.
- Docs: one paragraph in `docs/user/project-settings.md`, including that a prefix must not be the name of a git remote. A local branch whose first segment matches a remote is ambiguous in git, and T3 Code's existing publish path strips that segment for any branch, not only generated ones; the setting does not try to second-guess it.

The placeholder branch (`t3code/<8 hex>`) that web and mobile create at bootstrap is unchanged. Clients generate it before the server can read settings, older mobile builds keep sending it, and `isTemporaryWorktreeBranch` gates the rename on it, so the prefix applies at the rename instead. The placeholder is only visible until the first turn's name generation finishes, which is today's behavior.

## Why

People who push worktree branches and open PRs from them end up with `t3code/` in every branch name and merge commit, which fights personal or team naming conventions. There is no way around it short of renaming the branch by hand after each thread.

## UI Changes

| Before (origin/main) | After |
|---|---|
| No Branch prefix row | Branch prefix row showing t3code |

Environment value changed to `Yekta`; the row gains the standard reset arrow:

Branch prefix set to Yekta with reset arrow

Project scope with an override; the layers popover shows the chain (project, environment, default):

Layers popover: Project team/yekta, Environment Yekta, Default t3code

Empty value, shown as **No prefix**:

Branch prefix empty with No prefix placeholder

Searchable from the settings search:

Settings search result for branch prefix

Not captured: mobile. The server settings pages there only have choice rows, so like **Add project starts in** this setting has no mobile row; mobile decodes the key and ignores it.

## Verification

- `packages/shared/src/git.test.ts`: normalizer cases (case kept, invalid characters, `.lock` and separators in any order, empty, 64-char cap) and the builder with a custom, nested and empty prefix.
- `packages/contracts/src/settings.test.ts`: default, trimmed value, empty value, patch and project override decode.
- `ProviderCommandReactor.test.ts`: the first-turn rename targets `yekta/add-safer-backoff` with a prefix and `add-safer-backoff` without one, and the thread branch follows. The harness now accepts a server settings override.
- Targeted lint and typecheck clean on contracts, shared, client-runtime, server, web and mobile. Existing settings tests for web, mobile, client-runtime and server pass.
- Manually: value persists to `settings.json` and to `projectSettingsOverrides` for the project scope (screenshots above).

## Checklist

- [ ] I included a video for animation/interaction changes (none)

Written with Claude Fable 5.1 via Claude Code.

## Summary by CodeRabbit

* **New Features**
  * Configure a branch prefix in project settings for worktree branches. The default is `t3code`; an empty prefix disables it, and changes apply to worktrees created afterward.
  * Generated branches use the configured prefix when renamed after the first turn. Prefixes are cleaned up to meet Git branch naming rules.
* **Documentation**
  * Project settings guidance now explains branch prefix behavior and cautions that using a remote name such as `origin` can cause branches to be published without the prefix.


yekta:
Closing in favor of a fresh PR.

PR #13258, opened 2026-09-23
fix(models): defer effort and context to provider settings

Bil0000:
Claude and Codex effort/context dropdowns could override provider settings and show a window the thread did not use. Remove those choices, including stale saved values. Codex now reads `~/.codex/config.toml`; Claude Code reads `~/.claude/settings.json`. The context meter uses the window reported by the provider (for example, Codex may report 872K for a 1M request). Service tier and fast mode remain.

Verified: 326 focused tests, typechecks, lint, web build, and CI. The built-in Browser panel cannot reach this isolated dev server, so a new live after image is unavailable.

Before (the removed Claude context picker):

Claude context choice before the fix

GPT-6 Astra (Codex)

## Summary by CodeRabbit

* **Changes**
  * Codex reasoning effort and context length are now configured through Codex settings rather than T3 Code controls. Service-tier selection remains available.
  * Claude model options now rely on provider configuration; T3 Code no longer exposes provider-managed effort or context-window choices.
  * Updated guidance explains where to configure these options and how provider-reported context and service tiers are displayed.


Bil0000:
@coderabbitai review

Bil0000:
Checked the latest CodeRabbit Stop finding against `main`: `effectiveTurnId = turnId ?? session.activeTurnId` is unchanged by this PR. This PR no longer changes Codex turn queueing, so the finding is an existing Stop issue, not a regression from the provider-settings change. It needs a separate focused Stop fix.

PR #13261, opened 2026-09-23
fix: allow 100 MB file uploads without raising native media limits

Bil0000:
## What Changed

Raise the ordinary-file upload cap to 100 MB (100,000,000 bytes) across the shared contract, server, web, and mobile clients. Keep the existing image limits and Antigravity's 50 MiB native attachment budget. New clients still honor a smaller limit advertised by an older server.

## Why

The former 50 MiB cap blocked larger files before an agent could inspect them by path.

## Verification

- 311 focused tests passed across server, contracts, client runtime, and mobile.
- Targeted type checks for those four packages passed.
- Targeted lint and formatting passed. Lint reported one pre-existing warning outside the changed lines.

Provider and model limits still apply after upload. This PR does not change how Antigravity handles unsupported file types.

Model: GPT-6. Harness: Codex.

## Summary by CodeRabbit

* **New Features**
  * Increased the maximum size for non-image file attachments to 100 MB.
* **Documentation**
  * Updated the composer guide to reflect the 100 MB per-file limit.


Bil0000:
@coderabbitai review

apps/server/src/provider/acp/AntigravityAcpSupport.ts
```diff
@@ -303,24 +303,28 @@ export const buildAntigravityPrompt = Effect.fn("buildAntigravityPrompt")(functi
...
-          `Could not read attachment '${attachment.name}'.`,
-        );
-      }
-      continue;
+    if (info.type !== "File") {
+      return yield* EffectAcpErrors.AcpRequestError.invalidParams(
+        `Could not read attachment '${attachment.name}'.`,
+      );
     }
+    if (isPathOnly) continue;
```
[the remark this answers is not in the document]
Bil0000: Fixed in 77dfd04aa9. ProviderService now rejects any attachment path context that cannot fit, before adapter dispatch. The existing six-driver test now covers an exact 120,000-character prompt and rejection one character over; all six cases reproduced the bug before the fix. All 93 ProviderService tests, server type checking, targeted lint, and formatting passed.

Bil0000:
No code change is planned for the docstring coverage warning. The FileReader helper was moved for reuse, the new size calculation is small, and comments for deleted base64 helpers were removed. Adding docstrings solely to satisfy a percentage would not clarify these operations; focused tests cover the size boundaries, output preservation, fallback encoding, and read failures.

juliusmarminge (changes requested):
please split up into PRs with a single concern

Bil0000:
@juliusmarminge I split this as requested. #13261 now changes only the 100 MB ordinary-file upload cap and keeps Antigravity's 50 MiB native limit. The Antigravity path handling is in #13339; its tests cover a 75 MB PDF using the saved-path fallback. The web image-preparation change is also off this branch and is being prepared separately. Could you re-review this narrowed PR?

CodeRabbit's new oversized-PDF note concerns the provider path fallback in #13339. It is outside this upload-only diff.

Bil0000:
The image-preparation change is now in its own PR: #13342. The three concerns are separated: #13261 for the upload cap, #13339 for Antigravity file paths, and #13342 for web image preparation.

packages/contracts/src/orchestration.ts
```diff
@@ -166,7 +166,7 @@ export const PROVIDER_SEND_TURN_MAX_INPUT_CHARS = 120_000;
 export const PROVIDER_SEND_TURN_MAX_ATTACHMENTS = 100;
 export const PROVIDER_SEND_TURN_MAX_IMAGE_BYTES = 10 * 1024 * 1024;
 const PROVIDER_SEND_TURN_MAX_TOTAL_IMAGE_BYTES = 80 * 1024 * 1024;
-export const PROVIDER_SEND_TURN_MAX_FILE_BYTES = 50 * 1024 * 1024;
+export const PROVIDER_SEND_TURN_MAX_FILE_BYTES = 100_000_000;
```
juliusmarminge: Blocker (see review summary): this constant also bounds `ChatFileAttachment.sizeBytes` on **decode**. Every shipped client, and any server downgraded to `v0.0.42`, still rejects anything over 52,428,800 bytes, so one 75 MB attachment makes the thread undecodable for them. Please drop the read-side bound first, enforce the cap at upload issuance (`assets.ts:103`), and raise this constant only after the relaxed reader ships.

juliusmarminge (changes requested):
Audited the full PR against `main` at head `2fa6fd8c00e`. The split into [#13339](https://github.com/pingdotgg/t3code/pull/13339) and [#13342](https://github.com/pingdotgg/t3code/pull/13342) worked, and this PR is now narrow. It merges cleanly into current `main`, CI was green on `b65124359c9`, and the 311 focused tests pass locally. There is one compatibility blocker.

### Blocker: existing clients cannot load a thread that has a file over 50 MiB

`PROVIDER_SEND_TURN_MAX_FILE_BYTES` is both the upload cap and a **read** bound. `ChatFileAttachment.sizeBytes` (`packages/contracts/src/orchestration.ts:316-330`) rejects anything over the constant, and that schema decodes every stored message, `thread.message-sent` event, thread stream item, and projection row. `ChatUnknownAttachment` excludes `type: "file"`, so an oversized file cannot fall through to it either. `orchestration.test.ts:386` pins that strictness on purpose.

This PR lets a new server persist a file of 50–100 MB. Every shipped client still decodes with the 50 MiB bound: stable `v0.0.42`, the current nightly and preview tags, and the mobile store builds.

Reproduced with the contracts from the PR base (`894d33419dc`, the same bound as `v0.0.42`) against a 75,000,000-byte `application/zip` attachment:

| Payload | Base contract | This PR |
|---|---|---|
| `OrchestrationThreadStreamItem` (`subscribeThread` event) | FAIL: `Expected a value less than or equal to 52428800` | OK |
| `OrchestrationEvent` (event-store replay) | FAIL | OK |
| projection `attachments_json` row | FAIL | OK |

**Who hits this:** someone updates their desktop or `npx t3` server and sends a 75 MB ZIP. Any older client on that environment then fails to decode the thread: the mobile app, another desktop install connecting remotely, or a cached web bundle. Tracing the code (not observed in a client): the HTTP snapshot decode fails and falls back to the socket. The socket chunk decode goes through `RpcClient`'s `orDie`, the subscription's `onDefect` shows **"Could not synchronize the thread."**, and the stream stops without retrying. The event is permanent, so the thread stays broken on that client until it updates. Mobile is the long tail here, because store builds lag the server.

The server downgrade path also breaks (for example `t3 update 0.0.42` after a nightly). The older server fails to decode the same event on replay (`OrchestrationEventStore.ts` `rowToEvent`) and the projection row.

**Smallest fix I see:**
1. Separate the read bound from the write bound. Drop the upper `sizeBytes` bound from the persisted/read `ChatFileAttachment` shape, and keep the known-type discrimination. Enforce the cap where bytes enter: `AttachmentCreateUploadUrlInput` (`packages/contracts/src/assets.ts:103`) already signs the size. The Normalizer's on-disk size check (`apps/server/src/orchestration/Normalizer.ts:190`) ties the turn's `sizeBytes` to that upload, so the server stays protected. Update `orchestration.test.ts:386` to assert that an oversized file still **decodes**, and keep the upload-input test that rejects `100_000_001`.
2. Land (1) and let it reach stable desktop and the mobile store. Then raise the cap. Alternatively, a maintainer can explicitly accept the skew for this release. That decision belongs to a maintainer, not to this PR.

### Pushed fix

Browser recordings reuse this constant (`apps/web/src/browser/browserRecordingUpload.ts:26`, `apps/server/src/mcp/toolkits/preview/handlers.ts:169`), so their cap rose to 100 MB as well. The agent-facing `preview_recording_stop` description and `PreviewAutomationRecordingTooLargeError` still said 50 MiB. [`2fa6fd8c00e`](https://github.com/Bil0000/t3code/commit/2fa6fd8c00e028d95f1284b93fa3f9408de62cdf) updates both strings. The preview broker and toolkit tests pass, and lint and format are clean.

### Non-blocking

- **T3 Connect tunnels (unverified):** managed endpoints run through Cloudflare. Cloudflare documents a 100 MB request-body limit on Free and Pro zones, so uploads near 100,000,000 bytes may get a 413 through a tunnel. Please confirm the relay zone's limit before advertising exactly 100 MB to tunnel clients.
- The web upload XHR still times out after 5 minutes (`apps/web/src/lib/attachmentUploadQueue.ts:33`). A 100 MB file needs about 2.7 Mbit/s of sustained upstream to finish, which is plausible to miss on remote links.
- The six-driver `ProviderService` loop runs the same generic path-line code six times against the same fake Codex adapter, so one case covers it. The new 100 MB `AttachmentUpload` test writes 100 MB to disk to prove behavior that the contract boundary test already covers. Both are optional trims.

No approval from me. The final call belongs to maintainers.

_Audit: Claude Opus 5.5 via Claude Code._

packages/contracts/src/orchestration.ts
Bil0000: Agreed. I opened #13429 against `main` to remove the read-side file-size bound while keeping the 50 MiB upload cap. Its focused decoder test accepts a 75 MB file, rejects zero bytes, and the existing upload-cap test still passes. I added the rollout dependency to this PR description. I will leave this thread open and hold #13261 until that reader change reaches stable desktop and mobile, unless you explicitly accept the version-skew risk.

PR #13287, opened 2026-09-23
Add Hermes Agent ACP provider

crossthebridgetpa:
## Summary

- Hermes Agent speaks ACP (Agent Client Protocol), so it can plug into T3 the same way Grok, Claude Code, and the other built-in providers do.
- Adds a Hermes Agent built-in ACP provider: contracts settings, server driver/adapter/provider layers, and the web UI registration (icon, provider metadata, settings search).
- Mirrors the existing Grok provider's shape end to end. No new dependencies.

## Design notes

- Auth method is resolved dynamically from `initialize.authMethods[0]` instead of a hardcoded id, because Hermes can offer a different method depending on which upstream provider it is currently running against. The resolved id is cached and re-probed once if it goes stale.
- Models are discovered live from `session/new`, not a static list, since Hermes's available models depend on the user's own provider configuration.
- The adapter sets `cancelBehavior: "wait-for-prompt"`. Hermes does not reject a `session/prompt` that arrives while a turn is running: it queues it internally and answers the request immediately with a placeholder. With the default `interrupt` cancel, the adapter would redispatch a steer before Hermes had actually stopped, and the follow-up would be swallowed by that queue. `wait-for-prompt` makes the adapter wait for Hermes's real response to the cancelled prompt before dispatching the next one. A guard also fails any turn whose reply is that placeholder text, so the condition can never be silent.
- Provider status and install detection use `hermes acp --check` as the probe command, with CLI probe timeouts widened to 10s after early testing showed the default timeout was too tight for cold starts.

## Limitations

- Hermes 0.21.x never emits `session/elicitation` under `hermes acp` (its interactive ask-user tool is unavailable in that mode), so clarifying questions arrive as plain assistant text. The adapter still wires `session/elicitation` to T3's user-input flow so it works once Hermes exposes it.
- Text generation (commit messages, PR titles, branch names, thread titles) is unsupported. Hermes has no one-shot, out-of-session completion API, only the interactive ACP session used for chat turns, so every `HermesTextGeneration` operation fails closed and callers fall back the same way they would for any provider instance that does not support this capability.

## Test plan

- `pnpm --filter @t3tools/contracts typecheck`
- `pnpm --filter t3 typecheck`
- `cd apps/server && pnpm exec vp test run src/provider/Layers/HermesAdapter.test.ts src/provider/Layers/HermesProvider.test.ts src/provider/acp/HermesAcpSupport.test.ts src/provider/Layers/ProviderRegistry.test.ts src/provider/Layers/ProviderInstanceRegistryLive.test.ts`
- `pnpm --filter @t3tools/contracts test`
- Manual: enable the Hermes provider in Settings, send a prompt and confirm it streams, trigger a tool call and confirm the permission flow, send two follow-up messages while a turn is running and confirm both are answered after it completes, stop a turn mid-stream and confirm the session recovers cleanly.

## Summary by CodeRabbit

* **New Features**
  * Added experimental support for Hermes as an AI provider. Enable it in Settings and configure its executable path and optional home directory.
  * Hermes supports custom models, permission requests, interactive input, and in-progress turn steering.
  * Provider status and available models are checked and displayed in the app.
* **Bug Fixes**
  * Improved handling of interrupted turns and authentication retries, with clearer warnings when a model change or turn cannot be completed.


crossthebridgetpa:
Closing this myself: I missed CONTRIBUTING.md before opening and this is far larger than what you accept. Moving the idea to a Discussion; the branch stays on my fork if it is ever useful.

